PR #6, opened 2014-06-20
Using Zero-Copy API to read

dsy88:
Replace all the byte array based API with Byte Buffer based API in the reading path to avoid byte array copy.
1.Introduce a Compatible layer to keep compatible with old Hadoop API
2.Make the whole read path use ByteBuffer directly


dvryaboy:
@julienledem @tsdeng is this still outstanding?

julienledem:
@dvryaboy this has not been merged yet. Merge planed immediately after next release.

julienledem:
This was merged as part of #267

PR #7, opened 2014-06-23
PARQUET-7: [thrift] avoid thrift amender if all fields are optional.

dvryaboy:
No perf tests... I did some rudimentary testing which seems to bear out the idea but nothing scientific yet.


julienledem:
this was merged as part of https://github.com/apache/parquet-mr/pull/28

PR #17, opened 2014-07-11
PARQUET-16: Avoid calling getFileStatus() on all part-files

liancheng:
JIRA issue: [PARQUET-16](https://issues.apache.org/jira/browse/PARQUET-16)

This PR improves performance of `ParquetInputFormat.getSplit(JobContext)`, especially when reading large files from S3.

When calling `getSplits(JobContext)`, all the `FileStatus` objects are already fetched via `listStatus`, but abandoned immediately. PR #2 added an LRU cache for `Footer` objects, and fortunately corresponding `FileStatus` objects are cached together. In this PR, two private methods are added to leverage the LRU cache and prevent retrieving `FileStatus` objects sequentially when calling `getSplits(JobContext)`.

[Review on Reviewable](https://reviewable.io/reviews/apache/incubator-parquet-mr/17)

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
@@ -89,7 +89,7 @@ public class ParquetInputFormat<T> extends FileInputFormat<Void, T> {
    * key to configure the filter
    */
   public static final String UNBOUND_RECORD_FILTER = "parquet.read.filter";
-  
+
   /**
    * key to configure type checking for conflicting schemas (default: true)
    */
@@ -102,7 +102,7 @@ public class ParquetInputFormat<T> extends FileInputFormat<Void, T> {
 
   private static final int MIN_FOOTER_CACHE_SIZE = 100;
 
-  private LruCache<FileStatusWrapper, FootersCacheValue> footersCache;
+  private LruCache<FileStatusWrapper, CachedFooter> footersCache;
 
   private Class<?> readSupportClass;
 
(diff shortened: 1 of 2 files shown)


Myasuka:
Hi, guys.
We store parquet files with size 110GB, 6000+ parts on Tachyon, and query Spark sql such as `select count (*)`. However, before launching Spark tasks, there exists 18~20 seconds getting file status, I can see a lot of log info printling on the console, these info looks like `INFO   - getFileStatus(tachyon://****/**): HDFS Path: hdfs://****/** TPath: tachyon://****/**` These behaviors cost a lot of time, while we query the same data on HDFS in parquet format, there not exist these `getFileStatus` calls, which results in quering on HDFS gain a better performance.

Our Spark version is 1.2.0, Tachyon version is 0.6-SNAPSHOT, and Hadoop version is 2.4.0, I think this PR is related to our problem.

liancheng:
Hey @Myasuka, Spark SQL already worked around this issue by providing a customized input format. Since you only observe those `getFileStatus` calls on Tachyon, I guess it's Tachyon rather than Parquet who issued these calls. However, I'm rather unfamiliar with Tachyon. Please correct me if I'm wrong here.

Myasuka:
Hi, @liancheng , sorry for late reply. It seems that it's Tachyon's bug, you can see the call stack tracks below, the 1st pic is quering directly from HDFS while the other one is quering from Tachyon. Moreover, I have created an [issue](https://tachyon.atlassian.net/browse/TACHYON-275) in Tachyon.

fromhdfs
withtachyon

liancheng:
Closing this since #45 fixed this issue from another angle.

PR #49, opened 2014-09-04
PARQUET-77 zero copy improvements

parthchandra:
Adds to dsy88's pull request - adds the Direct decompressor interface to Snappy and adds a compatibilty utils function to read into a ByteBuffer allocated by the caller.

PR #50, opened 2014-09-04
Parquet 78 byte buffer write 20140904

parthchandra:
Adds to dsy88's pull request and pull request #49. Adds a ByteBuffer api to the write path. 
The API's pass an allocator to allocate a byte buffer to read/write into. I've provided two allocators, one to allocate from Heap and the other to allocate from Direct memory.
The Allocator interface has a 'release' method to allow allocators that use ref counting to know when to release memory. 
Actual memory allocation happens in CapacityByteArrayOutputStream.


parthchandra:
Subsumed in PR 267 (committed as 6b605a4)

PR #62, opened 2014-09-18
PARQUET-93: Add support for protobuf to parquet-cascading

matt-martin:
The parquet-cascading module currently includes code that lets Cascading write and read Thrift-based data to Parquet files. Protobufs is another popular alternative to Thrift which can be explicitly supported by leveraging the existing code in parquet-protobuf.


matt-martin:
@julienledem I know it's been a while, but I thought I'd give this PR a bump.  Any thoughts?

julienledem:
This looks good to me.

julienledem:
@qinghui-xu @matt-martin @lukasnalezenec @costimuraru @kgalieva: comments?
Otherwise I think we should merge this.

PR #70, opened 2014-09-30
PARQUET-110: Pig schema bug

rdblue:
Pig is removing some bags if they only have one element. To compensate
for this, when filtering tuples against the Pig schema, we have to
detect whether the bag is missing and filter the inner Parqet type
instead. This detects the situation when matching a bag and the inner
Parquet type is a repeated group with 1 element. If the inner Pig type
does not have a matching name, then this assumes it has been removed.


rdblue:
This was fixed in Pig

PR #73, opened 2014-10-18
PARQUET-116: Pass a filter object to user defined predicate in filter2 api

saucam:
Currently for creating a user defined predicate using the new filter api, no value can be passed to create a dynamic filter at runtime. This reduces the usefulness of the user defined predicate, and meaningful predicates cannot be created. We can add a generic Object value that is passed through the api, which can internally be used in the keep function of the user defined predicate for creating many different types of filters.
For example, in spark sql, we can pass in a list of filter values for a where IN clause query and filter the row values based on that list.

### DIFF
--- a/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
@@ -12,7 +12,7 @@ import java.io.Serializable;
  */
 // TODO: consider avoiding autoboxing and adding the specialized methods for each type
 // TODO: downside is that's fairly unwieldy for users
-public abstract class UserDefinedPredicate<T extends Comparable<T>> {
+public abstract class UserDefinedPredicate<T extends Comparable<T>, S extends Serializable> {
 
   /**
    * A udp must have a default constructor.
@@ -26,7 +26,7 @@ public abstract class UserDefinedPredicate<T extends Comparable<T>> {
    * Return true to keep the record with this value, false to drop it.
    * o is a filter object that can be used for filtering the value.
    */
-  public abstract boolean keep(T value, Serializable o);
+  public abstract boolean keep(T value, S o);
 
   /**
    * Given information about a group of records (eg, the min and max value)
--- a/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
@@ -10,7 +10,7 @@ package parquet.filter2.predicate;
  */
 // TODO: consider avoiding autoboxing and adding the specialized methods for each type
 // TODO: downside is that's fairly unwieldy for users
-public abstract class UserDefinedPredicate<T extends Comparable<T> > {
+public abstract class UserDefinedPredicate<T extends Comparable<T>> {
   /**
    * A udp must have a default constructor.
    * The udp passed to {@link FilterApi} will not be serialized along with its state.
--- a/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
@@ -22,8 +22,9 @@ public abstract class UserDefinedPredicate<T extends Comparable<T>> {
 
   /**
    * Return true to keep the record with this value, false to drop it.
+   * o is a filter object that can be used for filtering the value.
    */
-  public abstract boolean keep(T value);
+  public abstract boolean keep(T value, Object o);
 
   /**
    * Given information about a group of records (eg, the min and max value)
--- a/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
@@ -11,6 +11,7 @@ package parquet.filter2.predicate;
 // TODO: consider avoiding autoboxing and adding the specialized methods for each type
 // TODO: downside is that's fairly unwieldy for users
 public abstract class UserDefinedPredicate<T extends Comparable<T>> {
+
   /**
    * A udp must have a default constructor.
    * The udp passed to {@link FilterApi} will not be serialized along with its state.
--- a/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
@@ -1,7 +1,5 @@
 package parquet.filter2.predicate;
 
-import java.io.Serializable;
-
 /**
  * A UserDefinedPredicate decides whether a record should be kept or dropped, first by
  * inspecting meta data about a group of records to see if the entire group can be dropped,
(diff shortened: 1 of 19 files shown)


saucam:
Hi @isnotinvain any updates on this one ? 
This is useful for many different use cases, for example , we can pass in a bloom filter and get all rows not passing...

isnotinvain:
@rdblue makes some pretty good arguments for using String, and it would simplify a lot of this.

@julienledem @egonina do you have any strong opinions on how to store this kind of state?

parquet-column/src/main/java/parquet/filter2/predicate/UserDefinedPredicate.java
```diff
@@ -20,8 +26,17 @@
...
+   * This method is used to set the object that is used in the keep method for filtering.
+   * Called before returning the new instance of this class.
+   */
+  public void configure(S udpConfigParam) { 
+    this.udpConfig = udpConfigParam; 
+  }
+
   /**
    * Return true to keep the record with this value, false to drop it.
+   * o is a filter object that can be used for filtering the value.
```
julienledem: what does o refer to here?

parquet-scala/src/main/scala/parquet/filter2/dsl/Dsl.scala
```diff
@@ -26,10 +27,10 @@ import parquet.io.api.Binary
  */
 object Dsl {
 
-  private[Dsl] trait Column[T <: Comparable[T], C <: Operators.Column[T]] {
+  private[Dsl] trait Column[T <: Comparable[T], C <: Operators.Column[T], S <: java.io.Serializable] {
     val javaColumn: C
 
-    def filterBy[U <: UserDefinedPredicate[T]](clazz: Class[U]) = FilterApi.userDefined(javaColumn, clazz)
+    def filterBy[U <: UserDefinedPredicate[T, S]](clazz: Class[U], o: S) = FilterApi.userDefined(javaColumn, clazz, o)
```
julienledem: S could be defined as a parameter of this method rather then in Column. That would avoid having to add Serializable everywhere bellow

julienledem:
Hi @saucam 
Adding a serializable configuration to the current UserDefinedPredicate is a little awkward because the point of using a class name was not make explicit there was no serialized state.
Now if we want to add this capability we may as well pass directly an instance of a UserDefinedPredicate that is serializable.
I would recommend the following:
- leave UserDefined as is.
- similar to UserDefined, define a new ```ConfiguredUserDefined<T extends Comparable<T>, U extends UserDefinedPredicate<T> & Serializable>```
with a constructor: ```ConfiguredUserDefined(Column<T> column, U udp)```
- add in FilterApi:
```
public static <T extends Comparable<T>, U extends UserDefinedPredicate<T> & Serializable>
   UserDefined<T, U> userDefined(Column<T> column, U udp) {
     return new ConfiguredUserDefined<T, U>(column, udp);
  }
}
```
- add the new ConfiguredUserDefined in the visitor pattern and add the appropriate implementation that will be similar as UserDefined (you can even make it common by provided a base class to both)
I hope this helps

saucam:
Hi @julienledem ,

Thanks for the comments, added the changes mentioned. How does it look now ?

isnotinvain:
@julienledem is there value in making ConfiguredUserDefinedPredicate compose another UDP, or instead just make:
 ```
asbstract class ConfiguredUserDefinedPredicate extend UserDefinedPredicate<...> implements Serializable
```

**Comment by isnotinvain** [comment]
Sorry to keep piling on this review, one more question though. It seems like we shouldn't need to add a new operator to the filter tree, and then update every single visitor for filter trees.

We just want 2 ways of serializing a UDP, either by class name, or by serializing an instance. I think we can handle this just at the layer that deals with building a filter predicate + serializing it. The rest should be transparent.

I think we can do this by adding a method to FilterApi that accepts an Instance of a UDP that is Serializable, similar to what you have, eg:

```
/**
 * Keeps records that pass the provided {@link UserDefinedPredicate}
 */
public static <T extends Comparable<T>, U extends UserDefinedPredicate<T>>
 UserDefined<T, U> userDefined(Column<T> column, Class<U> clazz) {
 return new UserDefined<T, U>(column, clazz);
}

/**
 * Similar to above but allows to pass Serializable {@link UserDefinedPredicate}
 */
public static <T extends Comparable<T>, U extends UserDefinedPredicate<T>> & Serializable
 UserDefined<T, U> userDefined(Column<T> column, U udp) {
 return new UserDefined<T, U>(column, udp);
}
```

Note that we are still instantiating a `UserDefined`, not a new type like ConfigurableUserDefined.

The reason is, UserDefined is already serializable. If we make it act like a union type that holds either a class, or a reference to a serializable UDP, it should be fine, eg:

```
public static final class UserDefined<T extends Comparable<T>, U extends UserDefinedPredicate<T>> implements FilterPredicate, Serializable {
    private final Column<T> column;
    private final Class<U> udpClass;
    private final U udpInstance;
    ...

    public U getUserDefinedPredicate() {
      if (udpInstance != null) { 
        return udpInstance; 
      } else {
      try {
        return udpClass.newInstance();
      } catch (InstantiationException e) {
        throw new RuntimeException(String.format(INSTANTIATION_ERROR_MESSAGE, udpClass), e);
      } catch (IllegalAccessException e) {
        throw new RuntimeException(String.format(INSTANTIATION_ERROR_MESSAGE, udpClass), e);
      }
    }
   }
  }
```

Now this is the only change we have to make, none of the visitors need to care about a new type to visit. Does that make sense?

saucam:
But how is it enforced that 

private final U udpInstance

is serializable in the above class ?

julienledem:
@isnotinvain UserDefined could be abstract with just the column field and we would have 2 subclasses. One with the field udpClass and the other with the field udpInstance. The visitor only needs to know about UserDefined and an abstract definition of getUdpInstance() that would either return the serialized instance or instantiate it from the class depending on the implementation.

isnotinvain:
@julienledem yes, that would work too. Either way the main thing is it's a new getUdpInstance() method behavior, but none of the visitors need to be updated (I don't feel strongly whether we do this through subclassing, or just an if statement in UserDefined).

saucam:
@isnotinvain , @julienledem , made UserDefined abstract, added subclasses SimpleUserDefined , ConfiguredUserDefined

please suggest...

isnotinvain:
oops, I was playing around with this too. I'll take a look, thanks!

isnotinvain:
Cool, looks like we did the same thing :) 

I sent you a PR: https://github.com/saucam/incubator-parquet-mr/pull/2

This combines what you just submitted, plus updates to the scala DSL to support this new form of UDP, and some tests that exercise the serialization codepath. Let me know what you think.

saucam:
Thanks for the PR @isnotinvain  :)  merged.

julienledem:
I like it
+1

isnotinvain:
OK, let's merge once the tests pass. I'm trying to reproduce the test failure locally (we can't hit the rebuild button in travis unfortunately)

isnotinvain:
Tests passed for me locally.

isnotinvain:
I'm re-running the tests via this PR: https://github.com/apache/incubator-parquet-mr/pull/116, if those pass, I'll merge this PR instead.

isnotinvain:
OK, I ran the tests successfully via my github fork here:
https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/50141393

This is ready merge once @saucam merges: https://github.com/saucam/incubator-parquet-mr/pull/3

saucam:
Done

isnotinvain:
Thanks! I'll merge this now (finally :p)

saucam:
Thanks :)

PR #82, opened 2014-11-08
PARQUET-128: Optimize the parquet RecordReader implementation when: A. filterpredicate is pushed down , B. filterpredicate is pushed down on a flat schema

saucam:
The RecordReader implementation currently will read all the columns before applying the filter predicate and deciding whether to keep the row or discard it.
We can have a RecordReader which will only assemble the columns on which filters are applied (which are usually a few), then apply the filter and decide whether to keep the row or not , and then goes on to assemble the remaining columns or skip the remaining columns accordingly.
Also for applications like spark sql , the schema usually applied is a flat one with no repeating or nested columns. In such cases, its better to have a light-weight, faster RecordReader.
The performance improvement by this change is seen to be significant , and is better in case smaller number of rows are returned by filtering (which is usually the case) and there are many number of columns
This PR:
             1. Added OptFilterRecordReaderImplementation which skips reading columns for the rows rejected by the filter on which filter is never applied.
             2. Added a light-weight FlatSchemaRecordReaderImplementation with filtering, which only works with a flat schema (non repeated and non nested columns), which uses the above optimization as well. Usually for application like spark sql, the schema is flat and hence this reader results in much faster queries.

### DIFF
--- a/parquet-column/src/main/java/parquet/filter2/predicate/Operators.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/Operators.java
@@ -99,7 +99,7 @@ public final class Operators {
   }
 
   // base class for Eq, NotEq, Lt, Gt, LtEq, GtEq
-  static abstract class ColumnFilterPredicate<T extends Comparable<T>> implements FilterPredicate, Serializable  {
+  public static abstract class ColumnFilterPredicate<T extends Comparable<T>> implements FilterPredicate, Serializable  {
     private final Column<T> column;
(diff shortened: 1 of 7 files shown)


phraktle:
Any plans on merging this?

isnotinvain:
Sorry, I'll try to read through this this week.

saucam:
Closing this in favor of https://github.com/apache/parquet-mr/pull/288 (rebased to master)

PR #83, opened 2014-11-11
PARQUET-113: Update parquet-avro for new LIST and MAP annotation specs

rdblue:
This also fixes null values in arrays (PARQUET-83) based on #47 by @vallurisathish and null map values.

This does not implement unannotated repeated groups, which should be done in a follow-up commit.

There are currently two behavioral changes caused by this PR when reading old data:
1. Column projection inside a LIST structure probably fails because the Avro projection schema is translated to a Parquet Type using the new conversion rules, then matched against the old Type. This should be fixed before committing.
2. List structures in Parquet files not written by parquet-avro are interpreted differently. Previously, the type of the repeated group immediately inside a LIST-annotation was converted as the list's element type. Now, this is assumed to be a synthetic group if it is a single-field group. For files written with parquet-avro, the write schema is used to determine whether or not that assumption is correct.


rdblue:
This was merged in #128.

PR #89, opened 2014-12-01
PARQUET-142: add path filter in ParquetReader

nevillelyh:
Currently parquet-tools command fails when input is a directory with _SUCCESS file from mapreduce. Filtering those out like ParquetFileReader does fixes the problem.

```
parquet-cat /tmp/parquet_write_test
Could not read footer: java.lang.RuntimeException: file:/tmp/parquet_write_test/_SUCCESS is not a Parquet file (too small)

$ tree /tmp/parquet_write_test
/tmp/parquet_write_test
├── part-m-00000.parquet
└── _SUCCESS
```


julienledem:
+1 this sounds good to me.

rdblue:
I've merged this. Thanks @nevillelyh!

PR #91, opened 2014-12-04
PARQUET-139: Avoid reading footers when using task-side metadata

rdblue:
This updates the InternalParquetRecordReader to initialize the ReadContext in each task rather than once for an entire job. There are two reasons for this change:

1. For correctness, the requested projection schema must be validated against each file schema, not once using the merged schema.
2. To avoid reading file footers on the client side, which is a performance bottleneck.

Because the read context is reinitialized in every task, it is no longer necessary to pass the its contents to each task in ParquetInputSplit. The fields and accessors have been removed.

This also adds a new InputFormat, ParquetFileInputFormat that uses FileSplits instead of ParquetSplits. It goes through the normal ParquetRecordReader and creates a ParquetSplit on the task side. This is to avoid accidental behavior changes in ParquetInputFormat.


parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -147,17 +152,19 @@ public float getProgress() throws IOException, InterruptedException {
...
-  public void initialize(MessageType requestedSchema, MessageType fileSchema,
-      Map<String, String> extraMetadata, Map<String, String> readSupportMetadata,
+  public void initialize(MessageType fileSchema,
+      Map<String, String> fileMetadata,
       Path file, List<BlockMetaData> blocks, Configuration configuration)
       throws IOException {
-    this.requestedSchema = requestedSchema;
+    // initialize a ReadContext for this file
+    ReadSupport.ReadContext readContext = readSupport.init(new InitContext(
+        configuration, toSetMultiMap(fileMetadata), fileSchema));
```
julienledem: you should probably change the Deprecated annotations in ReadSupport.init

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -253,7 +283,23 @@ public ParquetInputFormat() {
...
-    return new ArrayList<InputSplit>(getSplits(configuration, getFooters(jobContext)));
+    List<InputSplit> splits = new ArrayList<InputSplit>();
+
+    if (isTaskSideMetaData(configuration)) {
+      // Although not required by the API, some clients may depend on always
+      // receiving ParquetInputSplit. Translation is required at some point.
+      for (InputSplit split : super.getSplits(jobContext)) {
+        Preconditions.checkArgument(split instanceof FileSplit,
+            "Cannot wrap non-FileSplit: " + split);
+        splits.add(ParquetInputSplit.from((FileSplit) split));
```
julienledem: Are you referring to Hive here? We should also modify Hive so that the ParquetInputSplit goes away altogether at some point.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -232,14 +240,36 @@ public ParquetInputFormat() {
...
+    return getReadSupportInstance(
+        (Class<? extends ReadSupport>) getReadSupportClass(configuration));
+  }
+
+  /**
+   * @param readSupportClass to instantiate
+   * @return the configured read support
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> ReadSupport<T> getReadSupportInstance(
```
julienledem: it does not look like the getReadSupportInstance methods should be public.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -232,14 +240,36 @@ public ParquetInputFormat() {
...
    * @return the configured read support
+   * @deprecated use getReadSupportInstance static methods instead
    */
+  @Deprecated
   public ReadSupport<T> getReadSupport(Configuration configuration){
+    if (readSupportClass != null) {
+      return getReadSupportInstance(readSupportClass);
+    } else {
+      return getReadSupportInstance(configuration);
+    }
```
julienledem: what about:
```
return getReadSupportInstance(readSupportClass == null ? getReadSupportClass(configuration) : readSupportClass)
```

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -201,19 +203,25 @@ public static Filter getFilter(Configuration conf) {
...
+   * Constructor for subclasses, such as AvroParquetInputFormat, or wrappers.
+   * <p>
+   * Subclasses and wrappers may use this constructor to set the ReadSupport
+   * class that will be used when reading instead of requiring the user to set
+   * the read support property in their configuration.
+   *
+   * @param readSupportClass a ReadSupport subclass
    */
-  public <S extends ReadSupport<T>> ParquetInputFormat(Class<S> readSupportClass) {
+  public <S extends ReadSupport> ParquetInputFormat(Class<S> readSupportClass) {
```
julienledem: Sorry I'm not following. Why are you dropping T?

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -253,7 +283,23 @@ public ParquetInputFormat() {
   @Override
   public List<InputSplit> getSplits(JobContext jobContext) throws IOException {
     Configuration configuration = ContextUtil.getConfiguration(jobContext);
-    return new ArrayList<InputSplit>(getSplits(configuration, getFooters(jobContext)));
+    List<InputSplit> splits = new ArrayList<InputSplit>();
+
+    if (isTaskSideMetaData(configuration)) {
```
julienledem: this whole block could replace the content of TaskSideMetadataSplitStrategy. Why is it not? It is a little weird to keep the ClientSideMetadataSplitStrategy but not the other one. Either we keep the strategy way of organizing the code or we don't but it should not be in the middle.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -486,118 +532,8 @@ public String toString() {
...
-    }
-  }
-
-  abstract List<ParquetInputSplit> getSplits(
-      Configuration configuration,
-      List<Footer> footers,
-      final long maxSplitSize, final long minSplitSize,
-      ReadContext readContext) throws IOException;
-}
-class TaskSideMetadataSplitStrategy extends SplitStrategy {
```
julienledem: I'm good with going with the FileInputFormat splits. I was probably trying to be too clever with this logic.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -108,32 +99,46 @@ private static long end(List<BlockMetaData> blocks) {
...
+  /**
+   * Builds a {@code ParquetInputSplit} from a mapred
+   * {@link org.apache.hadoop.mapred.FileSplit}.
+   *
+   * @param split a mapreduce FileSplit
+   * @return a ParquetInputSplit
+   * @throws IOException
+   */
+  public static ParquetInputSplit from(org.apache.hadoop.mapred.FileSplit split) throws IOException {
+    return new ParquetInputSplit(split.getPath(),
```
julienledem: those don't need to be public.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -108,32 +99,46 @@ private static long end(List<BlockMetaData> blocks) {
...
    * @param file the path of the file for that split
    * @param start the start offset in the file
    * @param end the end offset in the file
    * @param length the actual size in bytes that we expect to read
    * @param hosts the hosts with the replicas of this data
    * @param rowGroupOffsets the offsets of the rowgroups selected if loaded on the client
-   * @param requestedSchema the user requested schema
-   * @param readSupportMetadata metadata from the read support
    */
   public ParquetInputSplit(
```
julienledem: We need to keep the one the Hive depends on for compatibility.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
julienledem: By the way, I thing Hive depends on the split constructor but not the getSplits method.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -135,7 +135,18 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext context)
...
+    if (inputSplit instanceof ParquetInputSplit) {
+      parquetSplit = (ParquetInputSplit) inputSplit;
+    } else if (inputSplit instanceof FileSplit) {
+      parquetSplit = ParquetInputSplit.from((FileSplit) inputSplit);
+    } else {
+      throw new IllegalArgumentException(
+          "Invalid split (not a FileSplit or ParquetInputSplit): " + inputSplit);
+    }
+
+    initializeInternalReader(parquetSplit, ContextUtil.getConfiguration(context));
```
julienledem: ParquetInputSplit extends FileSplit so we probably can just depend on FileSplit.
On the other end we should really receive only what was returned by getSplits which always returns ParquetInputSplits in your implementation.
In both cases the logic above can be simplified.

parquet-hadoop/src/main/java/parquet/hadoop/mapred/DeprecatedParquetInputFormat.java
```diff
@@ -39,11 +41,15 @@
...
-    return new RecordReaderWrapper<V>(realInputFormat, split, job, reporter);
+    return new RecordReaderWrapper<V>(split, job, reporter);
   }
 
   @Override
   public InputSplit[] getSplits(JobConf job, int numSplits) throws IOException {
+    if (isTaskSideMetaData(job)) {
+      return toParquetSplits(super.getSplits(job, numSplits));
+    }
+
```
julienledem: I don't get this? Why the need for the special case?
I would say this is in the wrong place and this wrapper should not have to care about isTaskSideMetaData

pom.xml
```diff
@@ -330,6 +330,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.10</version>
         <configuration>
+          <redirectTestOutputToFile>true</redirectTestOutputToFile>
```
julienledem: is it in addition to std out? How does that work with Travis?

julienledem:
thank @rdblue
I did a first pass of comments.
Overall this looks good to me.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
rdblue: I wasn't talking about Hive, and you're right that Hive uses its own getSplits method.

I think it's a common practice to assume the type of the splits returned by getSplits, as we do here assuming that the FileInputFormat returns FileSplits. Originally, I had this return the FileSplit itself and handled both types in the RecordReader, but this would needlessly break any client that assumed ParquetInputSplit. This seemed unnecessary until we're ready to get rid of ParquetInputSplit, but that depends on deprecating the client-side behavior and is definitely a separate issue.

pom.xml
rdblue: This was an accidental addition. I like to be able to go directly to the output of a test without scrolling so much locally, but it doesn't work with Travis. I'll remove it.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
rdblue: I'm putting it back. I dropped it in other changes, and was mistaken when I said above it wasn't correct.

parquet-hadoop/src/test/java/parquet/hadoop/TestInputFormat.java
```diff
@@ -105,17 +104,6 @@ public void testThrowExceptionWhenMaxSplitSizeIsSmallerThanMinSplitSize() throws
...
   @Test
-  public void testThrowExceptionWhenMaxSplitSizeIsSmallerThanMinSplitSizeTaskSide() throws IOException {
-    try {
-      generateTSSplitByMinMaxSize(50, 49);
-      fail("should throw exception when max split size is smaller than the min split size");
-    } catch (ParquetDecodingException e) {
-      assertEquals("maxSplitSize and minSplitSize should be positive and max should be greater or equal to the minSplitSize: maxSplitSize = 49; minSplitSize is 50"
-              , e.getMessage());
-    }
-  }
```
rdblue: I've gone through and verified that split calculation is thoroughly tested. Because we no longer do the split calculations tested here, these tests are no longer useful. However, we do want to make sure that the relevant parts are tested:
* Mapping row groups to splits is very important and possibly impacted by changes to how splits are calculated. I've gone through and verified that the tests for this are thorough and still apply to using FileSplits. I'm confident that this doesn't need any more testing and is correct.
* Split calculation is now handled by FileInputFormat, so testing should be done upstream. We don't want to test this because it creates an unnecessary reliance on a library's behavior.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
rdblue: Hive depends on the Deprecated one above that says "For compatibility only".

https://github.com/apache/hive/blob/trunk/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/read/ParquetRecordReaderWrapper.java#L249

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
rdblue: Ideally, I'd like to move Hive over to using this and remove the deprecated constructor. In that case, it would need to be public. I also kept the mapreduce version for similar uses. I'm happy to change it, but I think it would be better to keep them exposed.

rdblue: Actually, I took another look at this. I think the right thing to do is to update Hive so that it doesn't need an internal Parquet split class at all. It should pass `FileSplit` to Parquet instead now that Parquet's splits are basically file splits.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetRecordReader.java
rdblue: I think this should be left as-is so we can start passing `FileSplit` instead of `ParquetInputSplit`. That would eliminate the need for Hive to do anything other than translate a mapred `FileSplit` to a mapreduce `FileSplit`. Then we can deprecate `ParquetInputSplit` from the public API and make it entirely internal and possibly remove it, once the client-side split strategy is gone.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
rdblue: I chose to do this because I think the eventual goal is to remove the client-side strategy. In that case, we would remove the `isTaskSideMetaData` check and the `getSplits` code will be just this method, without the `getSplits(conf, footers)`. As you note, there isn't much sense in just one split strategy so that would be removed as well.

But, the easiest way to get this commit done was to keep the split strategy for now so that I didn't have to change much in that code and we can be confident it works as it did before this commit. I saw no reason to move it all back into the static space of `ParquetInputFormat` and get rid of the strategy abstraction before we could get rid of the last remaining strategy.

I think this is the cleanest way to get to where we can eventually remove the strategy entirely. But, I should definitely deprecate methods under this one that will be removed.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
rdblue: The problem I'm trying to fix here is the "weird caching" referred to in the comment below. Caching the `readSupportClass` could cause problems when reusing an instance of the `InputFormat`, which happens in HiveServer 2. I think the cleanest way to do this is to remove `getReadSupport` from the API and replace it with static methods.

Here is my reasoning behind making those public:
* `getReadSupportInstance(Configuration)` is used by the `mapred.DeprecatedParquetInputFormat` which means we either have to duplicate it or make it public in place of `getReadSupport`.
* `getReadSupportInstance(Class)` makes sense considering you could do this before by using:

```java
new ParquetInputFormat(MyReadSupport.class).getReadSupport()
```

These really just replace the method that I'm deprecating above.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
rdblue: Done.

parquet-hadoop/src/main/java/parquet/hadoop/mapred/DeprecatedParquetInputFormat.java
rdblue: This method reads the file footers before delegating to `getSplits(Configuration, List<Footer>)` to avoid needing to construct a mapreduce `JobContext` object. But now that we only want to read footers if we are using client-side metadata, this needs to either 1) delegate to `ParquetInputFormat` earlier or 2) know when to read or not read the footers.

Delegating earlier requrires constructing the `JobContext` mapreduce object, while knowing whether or not to read the footer requires `isTaskSideMetadata` (using `ParquetInputFormat.TASK_SIDE_METADATA`). I think the second option is cleaner so that's what I did here.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -232,14 +240,36 @@ public ParquetInputFormat() {
   /**
    * @param configuration to find the configuration for the read support
    * @return the configured read support
+   * @deprecated use getReadSupportInstance static methods instead
```
rdblue: I don't see any other uses of this class when I grep through the code. Did you mean outside this repository? The reference in `mapred.DeprecatedParquetInputFormat` has been updated to use `getReadSupportInstance(Configuration)`.

rdblue:
@julienledem and @tomwhite I think I've addressed all of the review issues, though there were a few that I explained better rather than changing. I'd like to get this in shortly and push for a 1.6.0 release, so please take another look when you have time. Thanks!

tomwhite:
My feedback has been addressed, so I'm +1 on this once the travis run passes.

rdblue:
@julienledem I'd like to commit these changes this week, if you have a chance to take a second look. Thanks!

julienledem:
LGTM +1
Let's plan a follow-up PR to simplify the code a little by removing the SplitStrategy remaining class.

rdblue:
Thanks for the reviews! I've rebased and will merge this when tests pass.

rdblue:
It looks like Travis has a huge backlog, so I tested this locally instead of waiting for the job.

PR #92, opened 2014-12-09
When an avro read schema is present use it as the basis of the merged Schema

dguy:
In the current implementation the merge of on-disk schemas are determined to be incompatible as decided by the merge logic. However, this doesn't take into account the read-schema. 
This patch means that as long as the on-disk schemas can be merged into the read schema then we can successfully read/merge the on-disk data for all splits.


julienledem:
@tomwhite what do you think?

PR #95, opened 2014-12-22
added ParquetProtoStorer support for pig.

linyunfeng:
This is a simple implementation by following the thrift module to add the pig support for ParquetProtoStorer.

### DIFF
--- a/parquet-protobuf/pom.xml
+++ b/parquet-protobuf/pom.xml
@@ -53,9 +53,26 @@
       <version>${elephant-bird.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


lukasnalezenec:
Hi, 
Do you really need to write Prototobuffers in Pig ? I think that if you want to write Parquet in pig and read protobufers elsewere, you can use Thrift Storer and then read Protobufer using project parquet-proto.

Sorry for late response, I was on holidays.

linyunfeng:
We are using protobufer as standard schema protocol instead of thrift. so we dont use thrift at all. For the user like us would use parquet-pig + parquet-protobuf only.

PR #96, opened 2015-01-04
[PARQUET-159]: fixed broken test

tigerquoll:
Assertion exception (in oracle jdk6 at least) does not appear to have a (String, Exception) constructor.

### DIFF
--- a/parquet-hadoop/src/test/java/parquet/format/converter/TestParquetMetadataConverter.java
+++ b/parquet-hadoop/src/test/java/parquet/format/converter/TestParquetMetadataConverter.java
@@ -244,8 +244,10 @@ public class TestParquetMetadataConverter {
       try {
         verifyAllFilters(metadata(rgs), splitSize);
       } catch (AssertionError e) {
-        AssertionError e2 = new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " + splitSize + ")");
-        e2.initCause(e);
+        AssertionError aException = new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " +
+            splitSize + ")");
+        aException.initCause(e);
+        throw aException;
       }
     }
   }
--- a/parquet-hadoop/src/test/java/parquet/format/converter/TestParquetMetadataConverter.java
+++ b/parquet-hadoop/src/test/java/parquet/format/converter/TestParquetMetadataConverter.java
@@ -244,7 +244,8 @@ public class TestParquetMetadataConverter {
       try {
         verifyAllFilters(metadata(rgs), splitSize);
       } catch (AssertionError e) {
-	  throw new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " + splitSize + ")");
+	      throw new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " + splitSize + "): " +
+          Arrays.toString(e.getStackTrace()));
       }
     }
   }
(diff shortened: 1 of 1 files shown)


rdblue:
@tigerquoll, do you know more about why this is happening on your system? It seems odd that you are missing the constructor that accepts a throwable for AssertionError. Plus, the current [Travis CI build](https://travis-ci.org/apache/incubator-parquet-mr/builds) is passing and my build on ubuntu 14.04 works.

tigerquoll:
Your Travis build process uses jdk 1.7. I am using jdk 1.6. What is on your Ubuntu box?

rdblue:
I'm using JDK 1.7 from the [WebUpd8 java ppa](https://launchpad.net/~webupd8team/+archive/ubuntu/java). Isn't 1.6 [no longer supported](http://www.oracle.com/technetwork/java/javase/eol-135779.html)? While we build to 1.6 for those that are still using it, I don't think it is necessarily a requirement for the build to run on 1.6.

That said, I'm fine rewriting this to avoid the error on 1.6, but I'd rather not just drop the exception that was caught. Can you preserve the Exception's stack trace and message?

tigerquoll:
https://issues.apache.org/jira/browse/PARQUET-159 contains the error message

You are correct that JDK6 is no longer formally supported by oracle, but your pom.xml contains:

  <properties>
    <targetJavaVersion>1.6</targetJavaVersion>
    <maven.compiler.source>1.6</maven.compiler.source>
    <maven.compiler.target>${targetJavaVersion}</maven.compiler.target>
....
So currently the maven build process expect 1.6 compatible code and produce bytecode for 1.6, but requires 1.7 libraries. I could only find the one spot in your test code where this happened, but it will probably happen more and more in the future as developers stop using JDK6.

The most sensible options I see from here are:
1. Drop support for 1.6 JDK, update pom files to use 1.7, and update docco to say jdk 1.7 is the minimum requirement, or
2. Continue supporting 1.6 JDK, and add a test build using a 1.6 JDK to catch these type of errors in the future.

Thanks for looking into this.

rdblue:
Let's go for option 2. It looks like an easy enough fix for this issue, though I would say we need to preserve the content from the exception that is no longer being propagated to merge this.

tigerquoll:
@rdblue I've added a call to initcause() to preserve the exception. I tried logging on to the Travis CI system, but I do not have enough access to the Travis CI system to setup a JDK6 build for you.

parquet-hadoop/src/test/java/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -244,7 +244,8 @@ public void randomTestFilterMetaData() {
       try {
         verifyAllFilters(metadata(rgs), splitSize);
       } catch (AssertionError e) {
-	  throw new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " + splitSize + ")", e);
+        AssertionError e2 = new AssertionError("fail verifyAllFilters(metadata(" + Arrays.toString(rgs) + "), " + splitSize + ")");
+        e2.initCause(e);
```
rdblue: Still need to throw e2?

rdblue:
Thanks! I think you forgot to throw the new Exception, but the approach is good. I'll add a the jdk6 build to Travis in a separate PR.

tigerquoll:
@rdblue - all fixed up as per request

julienledem:
thanks @tigerquoll 
Sorry I missed your PR.
This has been fixed in https://github.com/apache/incubator-parquet-mr/pull/101
same fix you did.

PR #98, opened 2015-01-06
PARQUET-160: avoid wasting 64K per empty buffer.

julienledem:
This buffer initializes itself to a default size when instantiated.
This leads to a lot of unused small buffers when there are a lot of empty columns.


parquet-column/src/main/java/parquet/column/impl/ColumnWriterV2.java
```diff
@@ -58,13 +58,14 @@ public ColumnWriterV2(
...
       int initialSizePerCol,
-      ParquetProperties parquetProps) {
+      ParquetProperties parquetProps,
+      int pageSize) {
     this.path = path;
     this.pageWriter = pageWriter;
     resetStatistics();
-    this.repetitionLevelColumn = new RunLengthBitPackingHybridEncoder(getWidthFromMaxInt(path.getMaxRepetitionLevel()), initialSizePerCol);
-    this.definitionLevelColumn = new RunLengthBitPackingHybridEncoder(getWidthFromMaxInt(path.getMaxDefinitionLevel()), initialSizePerCol);
-    this.dataColumn = parquetProps.getValuesWriter(path, initialSizePerCol);
```
julienledem: We should tweak the initialSize here.
levels should get a tiny initial size (100 bytes?) in case they are always null or always defined.

julienledem:
The initial size here should be tweaked as well to something smaller:
https://github.com/julienledem/incubator-parquet-mr/blob/1df4a71d5a8f6e7c0adae142ce16bfccd34de999/parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordWriter.java#L111

parquet-encoding/src/main/java/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -38,66 +42,73 @@
...
-        if (Log.DEBUG) LOG.debug(String.format("slab size %,d too small for value of size %,d. replacing slab", currentSlab.length, minimumSize));
-        byte[] newSlab = new byte[minimumSize];
-        capacity += minimumSize - currentSlab.length;
-        this.currentSlab = newSlab;
-        this.slabs.set(currentSlabIndex, newSlab);
-      }
+    int nextSlabSize;
+    if (size == 0) {
+      nextSlabSize = initialSize;
+    } else if (size > pageSize / 5) {
```
isnotinvain: should 5 be configurable too?

we could also make CapacityByteArrayOutputStream abstract or take as an argument a slab size calculator etc. so that we can plug in different behaviors here. what do you think?

parquet-encoding/src/main/java/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -38,66 +42,73 @@
...
-      if (slabSize < minimumSize) {
-        if (Log.DEBUG) LOG.debug(String.format("slab size %,d too small for value of size %,d. Bumping up slab size", slabSize, minimumSize));
-        this.slabSize = minimumSize;
-      }
-      if (Log.DEBUG) LOG.debug(String.format("new slab of size %d", slabSize));
-      this.currentSlab = new byte[slabSize];
-      this.slabs.add(currentSlab);
-      this.capacity += slabSize;
+      // double the size every time
+      nextSlabSize = size;
```
isnotinvain: when did this get doubled?

isnotinvain: nevermind, I get it, nextSlabSize = the total size so far.

parquet-encoding/src/main/java/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -156,38 +167,12 @@ public int getCapacity() {
...
-        capacity += slab.length;
-      }
-    }
-    this.currentSlabIndex = 0;
-    this.currentSlabPosition = 0;
-    this.currentSlab = slabs.get(currentSlabIndex);
-    this.size = 0;
+    // readjust slab size.
+    // 7 = 2^3 - 1 so that doubling the initial size 3 times will get to the same size
+    initSlabs(max(size / 7, initialSize));
```
isnotinvain: do we still want this? or do we want to start all over again?
this means 1 large column chunk will trigger always having more mem overhead right? even if it's only 1 out of many small ones.

isnotinvain:
Do you want to tweak the initial size here as well?
Do you think we should try this out internally and see if it's an improvement first?

isnotinvain:
@julienledem ping!

isnotinvain:
Sent a PR against this PR here: https://github.com/julienledem/incubator-parquet-mr/pull/2

isnotinvain:
@tsdeng ok, this PR is now ready to review, it's got both @julienledem's changes and mine as well.

parquet-hadoop/src/main/java/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -40,6 +42,7 @@
 
 class ColumnChunkPageWriteStore implements PageWriteStore {
   private static final Log LOG = Log.getLog(ColumnChunkPageWriteStore.class);
+  private static final int COLUMN_CHUNK_WRITER_MAX_SIZE_HINT = 64 * 1024;
```
isnotinvain: rm this, not used.

isnotinvain:
+1, lets merge when the tests are green

isnotinvain:
I'm running these tests here:
https://github.com/isnotinvain/incubator-parquet-mr/pull/2

in case we have to wait a long time for the travis CI apache queue.

isnotinvain:
Tests passed! merging now...

PR #99, opened 2015-01-07
PARQUET-136: NPE thrown in StatisticsFilter when all values in a string/binary column trunk are null

saucam:
In case of all nulls in a binary column, statistics object read from file metadata is empty, and should return true for all nulls check for the column. Even if column has no values, it can be ignored.

The other way is to fix this behaviour in the writer, but is that what we want ?

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
+++ b/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
@@ -72,9 +72,9 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     return (column.getStatistics().isEmpty()) || (column.getStatistics().getNumNulls() == column.getValueCount());
   }
 
-  // are there any nulls in this column chunk?
+  // are there any nulls in this column chunk? 
   private boolean hasNulls(ColumnChunkMetaData column) {
-    return column.getStatistics().getNumNulls() > 0;
+    return (column.getStatistics().getNumNulls() > 0) || (isAllNulls(column));
   }
 
   @Override
--- a/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
+++ b/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
@@ -67,8 +67,9 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
   }
 
   // is this column chunk composed entirely of nulls?
+  // in case of all nulls, the stats object read from file metadata is empty
   private boolean isAllNulls(ColumnChunkMetaData column) {
-    return column.getStatistics().getNumNulls() == column.getValueCount();
+    return (column.getStatistics().isEmpty()) || (column.getStatistics().getNumNulls() == column.getValueCount());
   }
 
   // are there any nulls in this column chunk?
--- a/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
+++ b/parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
@@ -81,6 +81,16 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     Column<T> filterColumn = eq.getColumn();
     T value = eq.getValue();
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
+    }
 
     if (value == null) {
       // we are looking for records where v eq(null)
@@ -94,8 +104,6 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
       return true;
     }
 
-    Statistics<T> stats = columnChunk.getStatistics();
-
     // drop if value < min || value > max
     return value.compareTo(stats.genericGetMin()) < 0 || value.compareTo(stats.genericGetMax()) > 0;
   }
@@ -105,6 +113,16 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     Column<T> filterColumn = notEq.getColumn();
     T value = notEq.getValue();
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
+    }
 
     if (value == null) {
       // we are looking for records where v notEq(null)
@@ -118,8 +136,6 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
       return false;
     }
 
-    Statistics<T> stats = columnChunk.getStatistics();
-
     // drop if this is a column where min = max = value
     return value.compareTo(stats.genericGetMin()) == 0 && value.compareTo(stats.genericGetMax()) == 0;
   }
@@ -129,14 +145,22 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     Column<T> filterColumn = lt.getColumn();
     T value = lt.getValue();
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
+    }
 
     if (isAllNulls(columnChunk)) {
       // we are looking for records where v < someValue
       // this chunk is all nulls, so we can drop it
       return true;
     }
-
-    Statistics<T> stats = columnChunk.getStatistics();
 
     // drop if value <= min
     return  value.compareTo(stats.genericGetMin()) <= 0;
@@ -147,14 +171,22 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     Column<T> filterColumn = ltEq.getColumn();
     T value = ltEq.getValue();
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
+    }
 
     if (isAllNulls(columnChunk)) {
       // we are looking for records where v <= someValue
       // this chunk is all nulls, so we can drop it
       return true;
     }
-
-    Statistics<T> stats = columnChunk.getStatistics();
 
     // drop if value < min
     return value.compareTo(stats.genericGetMin()) < 0;
@@ -165,14 +197,22 @@ public class StatisticsFilter implements FilterPredicate.Visitor<Boolean> {
     Column<T> filterColumn = gt.getColumn();
     T value = gt.getValue();
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
+    }
 
     if (isAllNulls(columnChunk)) {
       // we are looking for records where v > someValue
       // this chunk is all nulls, so we can drop it
       return true;
     }
-
-    Statistics<T> stats = columnChunk.getStatistics();
 
     // drop if value >= max
     return value.compareTo(stats.genericGetMax()) >= 0;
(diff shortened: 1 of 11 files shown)


parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -67,13 +67,14 @@ private ColumnChunkMetaData getColumnChunk(ColumnPath columnPath) {
   }
 
   // is this column chunk composed entirely of nulls?
+  // in case of all nulls, the stats object read from file metadata is empty
   private boolean isAllNulls(ColumnChunkMetaData column) {
-    return column.getStatistics().getNumNulls() == column.getValueCount();
+    return (column.getStatistics().isEmpty()) || (column.getStatistics().getNumNulls() == column.getValueCount());
```
liancheng: Is it possible to have an empty row group without any records in it? In this case, a binary column also has null statistics  values, and `isAllNulls`/`hasNulls` both return `true`, which is at least semantically wrong.

liancheng:
Would be great if a test case can be added :-)

isnotinvain:
Thanks for fixing!

So currently, in the case of writing a row group that is all nulls, we don't record the number of nulls?
That does sound like a bug in the write path to me. @julienledem what do you think?

A test would be great as well.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
isnotinvain: If we decide this is a bug in the write path, lets put a TODO here with a reference to a jira.
Hopefully the only way the stats can be empty is when the column is all nulls, but I wonder if that is true?
What about files that were written pre-statistics?

@julienledem do you know if there can be an empty rowgroup? Hopefully not...

isnotinvain:
ping @egonina who might know about whether we should be writing stats for all null column chunks

isnotinvain:
So actually, when the stats object is empty, I think we should just do no filtering -- we don't know if it's empty because of this corner case, or because the file simply has no stats.

I think an aggressive early check for stats.isEmpty is all that we need here?

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
julienledem: I think we should fix it in the write path and have column.getStatistics().getNumNulls() == column.getValueCount() in that case.
missing statistics should mean "the writer did not have that feature"
We should not have empty row groups. There was a bug in the past that produced empty row groups in some cases but this was fixed.

isnotinvain:
I've filed: https://issues.apache.org/jira/browse/PARQUET-161 for the write path issue.

I think the fix here on the read path is to check isEmpty and keep all chunks that have no statistics.

egonina:
For a column that is all nulls, we would count the number of nulls and put that in the stats object but the stats object is still marked as empty (markAsNotEmpty is not called down [this path](https://github.com/apache/incubator-parquet-mr/blob/ccc29e4dde24584118211f27c71bb01bacc39326/parquet-column/src/main/java/parquet/column/impl/ColumnWriterV1.java#L113)). Looks like a bug, we should mark the stats as non-empty if the values are all nulls, since the column is non-empty. Then empty stats mean that we didn't collect any stats at all for the column.

saucam:
Hi all,

Thanks for the feedback. 
@isnotinvain  yes the statistics object is not written in case there are only nulls. I agree it is better to not filter in case of statistics object being empty. 
@egonina thanks for pointing out the omits.
Made the relevant changes.

Will add test case soon.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -81,6 +81,16 @@ private boolean hasNulls(ColumnChunkMetaData column) {
...
     ColumnChunkMetaData columnChunk = getColumnChunk(filterColumn.getColumnPath());
+    Statistics<T> stats = columnChunk.getStatistics();
+
+    if (stats.isEmpty()) {
+     // Do not filter in case of empty statistics
+     // pre-statistics files will have no such info
+     // best not to filter these.
+     // Also, in case of all nulls in a column, empty statistics object 
+     // is returned. This fixes the read path because of that bug
+      return false;
```
isnotinvain: since this comment is repeated, and we are going to fix the bug in the write path, lets just make this comment something like:
`// we have no statistics available, we cannot drop any chunks`

parquet-hadoop/src/test/java/parquet/filter2/statisticslevel/TestStatisticsFilter.java
```diff
@@ -279,10 +279,10 @@ public void testUdp() {
   @Test
   public void testClearExceptionForNots() {
     List<ColumnChunkMetaData> columnMetas = Arrays.asList(
-        getIntColumnMeta(new IntStatistics(), 0L),
-        getDoubleColumnMeta(new DoubleStatistics(), 0L));
+        getDoubleColumnMeta(new DoubleStatistics(), 0L),
```
isnotinvain: why the changes to this test?

isnotinvain:
I've sent you a PR for your 'npe' branch:
https://github.com/saucam/incubator-parquet-mr/pull/1/files

I think that gets us to a place where we don't need #103 and we'll have fixed both PARQUET-136 and PARQUET-161 in a way that is compatible with old files that were both written a) before stats were collected and b) while this bug was in effect.

Let me know what you think. Thanks!

parquet-hadoop/src/test/java/parquet/filter2/statisticslevel/TestStatisticsFilter.java
saucam: This test case starts failing because earlier, when we did not exit with stats.isEmpty() , filter condition will be evaluated for the IntColumn, and will return true, now since for the 'and' predicate we have : 

@Override
  public Boolean visit(And and) {
    return and.getLeft().accept(this) && and.getRight().accept(this);
  }

It will try to filter for double column and get an exception. 

With the stats.isEmpty() check in place, IntColumn filter will itself return a false (since the test is using an empty stats object), there would be no need to evaluate the doubleColumn filter check, and hence no exception contrary to what we are expecting !

saucam:
sorry for not thinking it through the first time! 
@isnotinvain , thanks for all the help. Have merged your changes (reverted the test case revert because of the reasons mentioned in the in line comment). Its much cleaner now
how does it look ?

saucam:
Added a test case.

saucam:
@isnotinvain , any updates on this one ?

parquet-hadoop/src/test/java/parquet/filter2/statisticslevel/TestStatisticsFilter.java
isnotinvain: thanks, now I understand!

parquet-hadoop/src/test/java/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -310,6 +316,39 @@ public boolean accept(Path p) {
...
+    ParquetWriter<Group> writer = new ParquetWriter<Group>(path, configuration, new GroupWriteSupport());
+   
+    Group r1 = new SimpleGroup(schema);
+    writer.write(r1);
+    writer.close();
+    
+    ParquetMetadata readFooter = ParquetFileReader.readFooter(configuration, path);
+    for (BlockMetaData block : readFooter.getBlocks()) {
+      for (ColumnChunkMetaData col : block.getColumns()) {
+        col.getPath();
```
isnotinvain: what's this loop for if all we do is call getPath() and ignore the return value?

parquet-hadoop/src/test/java/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -310,6 +316,39 @@ public boolean accept(Path p) {
...
+      }
+    }
+    { // assert the number of nulls are correct for the first block
+      assertEquals(1, (readFooter.getBlocks().get(0).getColumns().get(0).getStatistics().getNumNulls()));
```
isnotinvain: maybe we should also assert that isEmpty() is false?

parquet-hadoop/src/test/java/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -310,6 +316,39 @@ public boolean accept(Path p) {
...
+    writer.write(r1);
+    writer.close();
+    
+    ParquetMetadata readFooter = ParquetFileReader.readFooter(configuration, path);
+    for (BlockMetaData block : readFooter.getBlocks()) {
+      for (ColumnChunkMetaData col : block.getColumns()) {
+        col.getPath();
+      }
+    }
+    { // assert the number of nulls are correct for the first block
```
isnotinvain: do we need the extra braces here?

isnotinvain:
@saucam sorry for the slow reply thanks a lot!

LGTM, minor comments but aside from that +1

Long term we probably need more comprehensive tests around statistics, but probably not in this PR :)

saucam:
Added changes for the comments.

@isnotinvain , thanks !

isnotinvain:
+1, merge when the tests pass (to trigger the tests running again you might have to commit a whitespace change and then revert it, Travis can be flakey)

cc @julienledem @tsdeng who can actually merge this

saucam:
cc  @julienledem @tsdeng

isnotinvain:
I should be able to merge this now, I'll try to do it today.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -221,6 +254,19 @@ public Boolean visit(Not not) {
...
+    if (stats.isEmpty()) {
+      // we have no statistics available, we cannot drop any chunks
+      return false;
+    }
+
+    if (isAllNulls(columnChunk)) {
+      // there is no min max, there is nothing
+      // else we can say about this chunk, we
+      // cannot drop it.
+      return false;
```
egonina: why can't we drop it if it is all nulls in this case?

egonina:
+1 besides one clarification question

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -118,7 +130,12 @@ private boolean hasNulls(ColumnChunkMetaData column) {
       return false;
     }
 
-    Statistics<T> stats = columnChunk.getStatistics();
+    if (isAllNulls(columnChunk)) {
+      // there is no min max, there is nothing
+      // else we can say about this chunk, we
+      // cannot drop it.
+      return false;
+    }
```
julienledem: comparing null to anything else than null should always drop it. right?

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
julienledem: same question.

julienledem:
Same question as Katya. Otherwise this looks good to me.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
saucam: this filter is notEq filter , where we are looking for columnValue != (some non null value), so  any row with null column should be returned and cannot be dropped.

Also I think the above check is redundant anyways, since we have a hasNulls() check above this one, so if there are any nulls, we will be returning 'false' and this check would never be reached.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
saucam: we cannot just drop it, because the user defined predicate can be anything, we don't know beforehand. Also , if we don't keep this check, it will proceed to get min/max for the predicate, which can be nulls in case all values are nulls, so its best to just keep the chunk in case of all nulls.

isnotinvain: we should have (and could) extend the UDP interface to receive not just the min/max, but also the null count. Currently the UDP interface only accepts min/max, and in the case of all nulls, there is no min/max.

isnotinvain: See here: https://github.com/apache/incubator-parquet-mr/blob/master/parquet-column/src/main/java/parquet/filter2/predicate/Statistics.java

also note that if we add num nulls there, then existing UDP's will have to check for null themselves or their behavior might change.

parquet-hadoop/src/main/java/parquet/filter2/statisticslevel/StatisticsFilter.java
isnotinvain: Lets see...

case 1: v notEq(null) (line 121)
  we are trying to keep all non-null records. If this is a chunk of only-nulls, it can be dropped, otherwise, it can't. This case returns either way (does not fall through).

case2 : v notEq(nonNull) (line 127)
  we are trying to keep all records where v is not a particular (non-null) value. For example, v notEq(7). If this chunk contains nulls, there are some records to keep, because we decided that:
`null notEq(7)` is true, eg, we chose to make predicates work the way a programming language like java works (null != 7) instead of the way SQL or Pig works (nulls are viral, and dropped in all scenarios). It boils down to in programming languages we think of null as a value, and SQL think of it as "unknown."
This case falls through to the next unlike the case above.

@saucam is right, Line 134 to 137 (which I added to every operator) is unreachable in this case, we can delete it because to get here we've already confirmed that `!stats.isEmpty() && !stats.hasNulls()`

case3: v notEq(nonNullTarget) && there's a min + max
only case where we can drop the chunk is: min = max = nonNullTarget
eg, v notEq(7) and this column is a column of all 7s.

Thanks for catching the unreachable block!

isnotinvain: FYI, docs describing how nulls are handled: https://github.com/apache/incubator-parquet-mr/blob/master/parquet-column/src/main/java/parquet/filter2/predicate/FilterApi.java#L90

saucam:
removed the unreachable block. @julienledem how is it now ?

egonina:
LGTM

isnotinvain:
@saucam or merge process is a little strict, can you rename this PR:

"PARQUET-136: NPE thrown in StatisticsFilter when all values in a string/binary column trunk are null"

(note the all caps PARQUET)

Our merge script requires this to deal with JIRA / how apache git vs github integration works.

Thanks!

saucam:
done :)

isnotinvain:
Done! Thanks!

PR #101, opened 2015-01-07
PARQUET-174: Replaces AssertionError constructor introduced in Java7

laurentgo:
AssertionError(String, Throwable) was introduced in Java7. Replacing it with AssertionError(String) + initCause(Throwable)


julienledem:
+1

PR #102, opened 2015-01-08
PARQUET-157: Divide by zero fix

jimfcarroll:
There is a divide by zero error in logging code inside the InternalParquetRecordReader. I've been running with this fixed for a while but everytime I revert I hit the problem again. I can't believe anyone else hasn't had this problem. I submitted a Jira ticket a few weeks ago but didn't hear anything on the list so here's the fix.

This also avoids compiling log statements in some cases where it's unnecessary inside the checkRead method of InternalParquetRecordReader.

Also added a .gitignore entry to clean up a build artifact.

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
@@ -102,13 +102,16 @@ class InternalParquetRecordReader<T> {
   private void checkRead() throws IOException {
     if (current == totalCountLoadedSoFar) {
       if (current != 0) {
-        long timeAssembling = System.currentTimeMillis() - startedAssemblingCurrentBlockAt;
-        totalTimeSpentProcessingRecords += timeAssembling;
-        LOG.info("Assembled and processed " + totalCountLoadedSoFar + " records from " + columnCount + " columns in " + totalTimeSpentProcessingRecords + " ms: "+((float)totalCountLoadedSoFar / totalTimeSpentProcessingRecords) + " rec/ms, " + ((float)totalCountLoadedSoFar * columnCount / totalTimeSpentProcessingRecords) + " cell/ms");
-        long totalTime = totalTimeSpentProcessingRecords + totalTimeSpentReadingBytes;
-        long percentReading = 100 * totalTimeSpentReadingBytes / totalTime;
-        long percentProcessing = 100 * totalTimeSpentProcessingRecords / totalTime;
-        LOG.info("time spent so far " + percentReading + "% reading ("+totalTimeSpentReadingBytes+" ms) and " + percentProcessing + "% processing ("+totalTimeSpentProcessingRecords+" ms)");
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -102,13 +102,14 @@ public InternalParquetRecordReader(ReadSupport<T> readSupport, UnboundRecordFilt
...
-        LOG.info("time spent so far " + percentReading + "% reading ("+totalTimeSpentReadingBytes+" ms) and " + percentProcessing + "% processing ("+totalTimeSpentProcessingRecords+" ms)");
+        totalTimeSpentProcessingRecords += (System.currentTimeMillis() - startedAssemblingCurrentBlockAt);
+        if (Log.INFO) {
+            LOG.info("Assembled and processed " + totalCountLoadedSoFar + " records from " + columnCount + " columns in " + totalTimeSpentProcessingRecords + " ms: "+((float)totalCountLoadedSoFar / totalTimeSpentProcessingRecords) + " rec/ms, " + ((float)totalCountLoadedSoFar * columnCount / totalTimeSpentProcessingRecords) + " cell/ms");
+            long totalTime = totalTimeSpentProcessingRecords + totalTimeSpentReadingBytes;
+            long percentReading = totalTime == 0 ? 0 : (100 * totalTimeSpentReadingBytes / totalTime);
+            long percentProcessing = totalTime == 0 ? 0 : (100 * totalTimeSpentProcessingRecords / totalTime);
+            LOG.info("time spent so far " + percentReading + "% reading ("+totalTimeSpentReadingBytes+" ms) and " + percentProcessing + "% processing ("+totalTimeSpentProcessingRecords+" ms)");
```
rdblue: Does this need to be logged if the total time processing and reading is 0ms? I think it would be better to do this:

```java
if (totalTime > 0) {
  long percentReading = ...;
  long percentProcessing = ...;
  LOG.info("time spent so far"...);
}
```

rdblue:
Just one comment and I'll commit this.

In the future, could you also explain any changes you include that aren't related to the fix? Wrapping the info messages in a check for `Log.INFO` makes sense, but an explanation would have helped. It looked like a big change for a divide-by-zero bug. Thanks!

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
jimfcarroll: I'll make that change.

I have a .gitignore change in a subsequent commit. Do you want that removed?

jimfcarroll:
Okay. I updated the PR so it incorporates your suggestion. It makes sense to me.

If you want me to pull the .gitignore commit, let me know.

julienledem:
+1 LGTM

rdblue:
Thanks @jimfcarroll!

jimfcarroll:
Thanks! I'm looking forward to 1.6.0.

PR #105, opened 2015-01-12
Parquet 167 - Snappy Compression codec - Optimize freeing of DirectByteBuffers

tigerquoll:
This PR changes the memory handling of the Snappy codec, by allowing the codec to explicitly free memory when no longer required.

Freeing of DirectByteBuffers is usually performed by GC routines calling the 'cleaner' routine registered for the buffer.  We can access the same routines by:
a. Utilising the sun proprietry api java.nio.ByteBuffer interface to access the cleaner routine
b. Using reflection to access the routine.
In this PR, option (a) was used as it was considered the cleaner/more performant approach.

Given the capability to explicitly free DirectByteBuffers, the snappy codec was changed so that when DirectByteBuffers are being replaced with larger buffers, the old buffers are explicitly freed rather then waiting for GC to free them.

Optional behaviour was added controlled by the configuration option: ''org.apache.hadoop.io.compress.SnappyCodec.keepBuffersForReuse" (defaults to true).
If this option is true, then the existing snappy codec memory use behaviour of keeping buffers available continues to occur.

If this option if false, then a new behaviour occurs that frees buffers once the snappy codec has finished with them, and reallocates the buffers when required again.  Some overhead is expected for this, but should be negligible versus the cost of disk IO.

A new interface (parquet.hadoop.codec.buffers.CodecByteBuffer) has been added to abstract out the difference between these two options.  The module includes associated implementations and a factory object.

### DIFF
--- a/parquet-hadoop/pom.xml
+++ b/parquet-hadoop/pom.xml
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>org.xerial.snappy</groupId>
       <artifactId>snappy-java</artifactId>
-      <version>1.0.5</version>
+      <version>1.1.1.6</version>
       <type>jar</type>
       <scope>compile</scope>
     </dependency>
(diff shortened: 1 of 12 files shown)


julienledem:
Thanks.
Adding @	parthchandra who has been looking at the bytebuffer allocations and @nongli who implemented the Codec.

PR #107, opened 2015-01-25
PARQUET-171: Support Avro schema resolution across Parquet-Avro files

jeffreyolchovy:
This pull request provides a derived version of `AvroReadSupport` that supports Avro schema resolution.

Currently, `AvroReadSupport` will throw a runtime exception if there exist Parquet-Avro files with different Avro schemas.

This is a problem our organization has encountered when running MapReduce jobs over a large range of Parquet-Avro files as Avro schemas may change over time.

Differing Avro schemas may be "compatible" -- a feature of Avro's schema resolution/evolution -- and if schemas in a given input set are compatible, they can be resolved to a single Avro schema.

The provided solution extends `AvroReadSupport` (for backwards compatibility) and uses the utilities provided by the Apache Avro library for checking if all Avro schemas present in the Parquet-Avro data files are compatible with the given "reader" schema.

See https://issues.apache.org/jira/browse/PARQUET-171 for a full description of the problem and solution.

/cc @rdblue

### DIFF
--- a/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
@@ -34,7 +34,6 @@ import parquet.io.api.RecordMaterializer;
 import parquet.schema.MessageType;
 
 import static org.apache.avro.SchemaCompatibility.*;
-import static parquet.avro.AvroReadSupport.*;
 
 /**
  * Avro implementation of {@link ReadSupport} for Avro {@link IndexedRecord}s which cover both Avro Specific and
--- a/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
@@ -51,7 +51,7 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
 
   static final String AVRO_SCHEMA_METADATA_KEY = "avro.schema";
 
-  private static final String AVRO_READ_SCHEMA_METADATA_KEY = "avro.read.schema";
+  static final String AVRO_READ_SCHEMA_METADATA_KEY = "avro.read.schema";
 
   public static String AVRO_DATA_SUPPLIER = "parquet.avro.data.supplier";
 
@@ -179,11 +179,7 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
       return mergeKeyValueMetadata(context, readerSchema);
     // otherwise, do not attempt to perform schema resolution/evolution
     } else {
-      Map<String, String> metadata = context.getMergedKeyValueMetaData();
-      if (readerSchemaString != null) {
-        metadata.put(AVRO_READ_SCHEMA_METADATA_KEY, readerSchemaString);
-      }
-      return metadata;
+      return context.getMergedKeyValueMetaData();
     }
   }
 
@@ -192,8 +188,11 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
     MessageType fileSchema = context.getFileSchema();
     MessageType schema = fileSchema;
     Map<String, String> metadata = getMergedKeyValueMetadata(context);
-    // replace Parquet schema with one created from an Avro projection, iff one is supplied
     Configuration configuration = context.getConfiguration();
+    String readerSchemaString = configuration.get(AVRO_READ_SCHEMA);
+    if (readerSchemaString != null) {
+      metadata.put(AVRO_READ_SCHEMA_METADATA_KEY, readerSchemaString);
+    }
     String requestedProjectionString = configuration.get(AVRO_REQUESTED_PROJECTION);
     if (requestedProjectionString != null) {
       Schema avroRequestedProjection = new Schema.Parser().parse(requestedProjectionString);
--- a/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
@@ -40,10 +40,6 @@ import static parquet.avro.AvroReadSupport.*;
  * Avro implementation of {@link ReadSupport} for Avro {@link IndexedRecord}s which cover both Avro Specific and
  * Generic. Users should use {@link AvroParquetReader} or {@link AvroParquetInputFormat} rather than using
  * this class directly.
- * <p>{@code AvroReadSupport} that will perform schema evolution, if possible.</p>
- * <p>It is required to set the {@link AvroReadSupport#AVRO_READ_SCHEMA} in the context {@link Configuration}.</p>
- * <p>Avro schemas found in input files will be checked for compatibility against the {@code AVRO_READ_SCHEMA}.</p>
- * <p>A {@code RuntimeException} will be thrown if schemas are not compatible.</p>
  */
 public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
 
@@ -78,10 +74,17 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
     configuration.set(AVRO_DATA_SUPPLIER, clazz.toString());
   }
 
+  /**
+   * Enable the resolution of Avro schemas in the metadata against the {@link #AVRO_READ_SCHEMA}.
+   */
   public static void enableAvroSchemaCompatibilityCheck(Configuration configuration) {
     configuration.setBoolean(AVRO_SCHEMA_COMPATIBILITY_CHECK, true);
   }
 
+  /**
+   * Disable the resolution of Avro schemas in the metadata against the {@link #AVRO_READ_SCHEMA}.
+   * Avro schema compatibility checks are disabled by default.
+   */
   public static void disableAvroSchemaCompatibilityCheck(Configuration configuration) {
     configuration.setBoolean(AVRO_SCHEMA_COMPATIBILITY_CHECK, false);
   }
@@ -115,7 +118,7 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
    * Attempt to merge the Avro schemas listed under the {@code AVRO_SCHEMA_METADATA_KEY} by checking
    * if each schema is compatible with the requested "reader" schema.
    *
-   * <p>Avro schemas in the metadata may differ across our Parquet files, but they may be compatible in terms of Avro schema evolution.</p>
+   * <p>Avro schemas in the metadata may differ, but they may be compatible in terms of Avro schema evolution.</p>
    * <p>For other keys in the metadata, follow the conventions of {@link InitContext#getMergedKeyValueMetaData}.</p>
    *
    * @param context init context that contains unmerged key-value metadata
@@ -159,8 +162,9 @@ public class AvroReadSupport<T extends IndexedRecord> extends ReadSupport<T> {
   }
 
   /**
-   * To initialize, {@link #AVRO_READ_SCHEMA} must be defined in the context configuration.
-   * @see ReadSupport#init(InitContext)
+   * If an {@link #AVRO_READ_SCHEMA} is provided and the {@link #AVRO_SCHEMA_COMPATIBILITY_CHECK}
+   * flag has been set, all Avro schemas found in the metadata will be checked for compatibility
+   * against the {@code AVRO_READ_SCHEMA}.
    */
   @Override
   public ReadContext init(InitContext context) {
--- a/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
@@ -25,6 +25,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import org.apache.avro.Schema;
+import org.apache.avro.SchemaCompatibility.SchemaCompatibilityType;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.util.ReflectionUtils;
@@ -33,7 +34,7 @@ import parquet.hadoop.api.ReadSupport;
 import parquet.io.api.RecordMaterializer;
 import parquet.schema.MessageType;
 
-import static org.apache.avro.SchemaCompatibility.*;
+import static org.apache.avro.SchemaCompatibility.checkReaderWriterCompatibility;
 
 /**
  * Avro implementation of {@link ReadSupport} for Avro {@link IndexedRecord}s which cover both Avro Specific and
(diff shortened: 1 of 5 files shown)


tomwhite:
This is a great thing to add. I think the logic could be added to AvroReadSupport directly, since it is a net improvement.

jeffreyolchovy:
I originally thought the same thing but was worried about breaking existing code. My main goal was to try to fast track the merging of this.

I don't mind merging this logic into AvroReadSupport, if that's how you'd like me to proceed. Thanks.

jeffreyolchovy:
After attempting to move the schema resolution logic to `AvroReadSupport`, two existing test cases now fail:
```
Results :
Tests in error: 
  testReadWriteChangedCar(parquet.avro.TestSpecificInputOutputFormat): could not merge metadata: key avro.schema contains incompatible schemas
  testAvroReadSchema(parquet.avro.TestSpecificReadWrite): could not merge metadata: key avro.schema contains incompatible schemas

Tests run: 32, Failures: 0, Errors: 2, Skipped: 0
```
Both of these test cases deal with schema resolution of records in the `Cars` protocol (found in the test resources directory).

In one case, `Car` records are trying to be migrated to `NewCar` records, and in the other case `Car` records are trying to be migrated/coerced to `ShortCar` records.

Because `AvroReadSupport` is used "under-the-hood" and an Avro reader schema is set, schema compatibility checks will now be executed between `Car` and `NewCar` and `Car` and `ShortCar` types, however these schemas are not compatible as far as Avro is concerned.

As far as I can tell, it seems that the tests and the test resources are relying on some sort of heuristical coersion between record types as opposed to what I typically think of when it comes to schema resolution/evolution: having the *same* protocol with records of the *same* name change their contents over time.

I'm looking for some feedback on how to proceed, especially because I can't think of a way to resolve the failures in these tests without changing the paradigm of the two larger unit test files, which I think is a somewhat larger undertaking than I (or you) had planned.

My branch with the direct modifications to `AvroReadSupport` can be found at https://github.com/Tapad/incubator-parquet-mr/tree/refactor/PARQUET-171 and here is a quick comparison of it with the HEAD of the feature branch https://github.com/Tapad/incubator-parquet-mr/compare/Tapad:feature/PARQUET-171...refactor/PARQUET-171.

/cc @tomwhite

tomwhite:
Without this change the value of avro.schema in each Parquet file is checked to ensure they are all the same. With this change, the reader schema is checked against all the writer schemas (avro.schema) to ensure that it can read all of them. That check wasn't being done before, but the effect should only be to fail earlier. For safety's sake, I think I'd prefer to add a switch to enable the new behaviour rather than have a different implementation of AvroReadSupport.

I can't see why the two tests are failing, they should both be a compatible schema change. I did notice that NewCar's description doesn't have a default (it should be null) - not sure if that is a problem.

jeffreyolchovy:
>> I think I'd prefer to add a switch to enable the new behaviour rather than have a different implementation of AvroReadSupport.

Agree with this entirely.

>> I can't see why the two tests are failing, they should both be a compatible schema change. I did notice that NewCar's description doesn't have a default (it should be null) - not sure if that is a problem.

Yeah, the description should definitely have a default value, but that's not the only thing that makes them incompatible. The idea of having a `record Car { ... }` be cast/coerced to a `record NewCar { ... }` is the issue. Avro is not OK with having the actual name of the record changed even though the contents of the records adhere to schema resolution guidelines (once you add a default value to the description for `NewCar`). I manually verified this by compiling a bunch of permutations of the IDLs. It's an awful lot of noise so I'll add a link to a pastebin (http://pastebin.com/pzyrQ28e).

Either way, I'm going to update the PR with the switch to enable the new behavior.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -15,16 +15,23 @@
...
 import org.apache.avro.Schema;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.util.ReflectionUtils;
+import parquet.hadoop.api.InitContext;
 import parquet.hadoop.api.ReadSupport;
 import parquet.io.api.RecordMaterializer;
 import parquet.schema.MessageType;
 
+import static org.apache.avro.SchemaCompatibility.*;
```
tomwhite: Nit: try to avoid wildcard imports.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -33,9 +40,13 @@
...
   public static String AVRO_REQUESTED_PROJECTION = "parquet.avro.projection";
-  private static final String AVRO_READ_SCHEMA = "parquet.avro.read.schema";
 
-  static final String AVRO_SCHEMA_METADATA_KEY = "avro.schema";
+  public static final String AVRO_READ_SCHEMA = "parquet.avro.read.schema";
+
+  public static final String AVRO_SCHEMA_COMPATIBILITY_CHECK = "parquet.avro.schema.compatibility";
+
+  public static final String AVRO_SCHEMA_METADATA_KEY = "avro.schema";
+
```
tomwhite: Is there a need to make these public? Properties are set using the static methods.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -59,6 +70,115 @@ public static void setAvroDataSupplier(Configuration configuration,
...
+        Set<Schema> writerSchemas = new HashSet<Schema>();
+        /* It is necessary to instantiate a Schema.Parser for each schema
+         * because we are parsing for the same type with different, "evolved" definitions!
+         */
+        for (String s : writerSchemaStrings) {
+          Schema.Parser parser = new Schema.Parser();
+          Schema writerSchema = parser.parse(s);
+          writerSchemas.add(writerSchema);
+        }
+        if (!areSchemasCompatible(readerSchema, writerSchemas))
```
tomwhite: Nit: please add braces even for a one-line body.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -59,6 +70,115 @@ public static void setAvroDataSupplier(Configuration configuration,
...
+    boolean checkSchemaCompatibility = configuration.getBoolean(AVRO_SCHEMA_COMPATIBILITY_CHECK, false);
+    // attempt to resolve/evolve writer schemas to the reader schema
+    // iff a reader schema is provided and the compatibility check flag is flipped
+    if (readerSchemaString != null && checkSchemaCompatibility) {
+      Schema.Parser parser = new Schema.Parser();
+      Schema readerSchema = parser.parse(readerSchemaString);
+      return init(configuration, mergeKeyValueMetadata(context, readerSchema), context.getFileSchema());
+    // otherwise, do not attempt to perform schema resolution/evolution
+    } else {
+      return init(configuration, context.getMergedKeyValueMetaData(), context.getFileSchema());
```
tomwhite: InitContext#getMergedKeyValueMetaData() is deprecated, so it would be better to avoid calling it. How about always calling a private mergeKeyValueMetadata(context) method that only checks compatibility if checkSchemaCompatibility is true, otherwise just falls back to the behaviour in InitContext#getMergedKeyValueMetaData().

parquet-avro/src/test/java/parquet/avro/TestAvroReadSupport.java
```diff
@@ -0,0 +1,198 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static parquet.avro.AvroReadSupport.*;
+
+public class TestAvroReadSupport {
+
+  /* Avro Schemas */
+
+  /* equivalent to message Foo { required binary field_1 (UTF8); } */
+  static final String RECORD_WITH_REQUIRED_STRING = "{\"type\":\"record\",\"name\":\"Foo\",\"fields\":[{\"name\":\"field_1\",\"type\":{\"type\":\"string\"}}]}";
```
tomwhite: Please break these across multiple lines so they are more readable. (Max line length is 80.)

parquet-avro/src/test/java/parquet/avro/TestAvroReadSupport.java
```diff
@@ -0,0 +1,198 @@
...
+import org.junit.rules.ExpectedException;
+import parquet.hadoop.api.InitContext;
+import parquet.schema.MessageType;
+import parquet.schema.MessageTypeParser;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static parquet.avro.AvroReadSupport.*;
+
+public class TestAvroReadSupport {
```
tomwhite: These are great tests. It would be good to have one (in TestReadWrite?) that additionally checks that records can actually be read correctly when there are different writer schemas.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -59,6 +70,115 @@ public static void setAvroDataSupplier(Configuration configuration,
...
+    }
+  }
+
   @Override
   public ReadContext init(Configuration configuration, Map<String, String> keyValueMetaData, MessageType fileSchema) {
```
tomwhite: The logic in this method could be moved to init(InitContext context) and this method deleted (since it is deprecated in the superclass).

jeffreyolchovy: Can you clarify? This is the init(InitContext context) method, unless I am misunderstanding you.

parquet-avro/src/test/java/parquet/avro/TestAvroReadSupport.java
jeffreyolchovy: Will do.

parquet-avro/src/test/java/parquet/avro/TestAvroReadSupport.java
jeffreyolchovy: Ok.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
jeffreyolchovy: Nope, no need, I think the original change to the access modifier is now moot -- will change.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
jeffreyolchovy: Ok, I think I understand, I'll implement and solicit your feedback to see if we're on the same page, thanks.

jeffreyolchovy:
@tomwhite re: "It would be good to have one (in TestReadWrite?) that additionally checks that records can actually be read correctly when there are different writer schemas."

I'm having trouble getting this feature to work correctly when using the AvroParquetReader, there seems to be more stuff happening behind the scenes than with the AvroParquetInputFormat.

I can verify that all of this works well and good with AvroParquetInputFormat, as that is where this PR originates from. My organization has been using a version of "EvolvableAvroReadSupport" I originally wrote in Scala, and we thought other people writing MapReduce jobs could benefit from our work -- thus began the Java port and the pull request.

We are not direct users of AvroParquetReader (moreover, ParquetReader) and I am unfamiliar with the code behind it. I'll have to do a deep dive into the logic to figure out why it doesn't behave similarly to how AvroParquetInputFormat integrates with AvroReadSupport, but unfortunately, it's not a high priority.

rdblue:
> The idea of having a record Car { ... } be cast/coerced to a record NewCar { ... } is the issue

This is something we hit in Kite with schema union resolution. [AVRO-1590](https://issues.apache.org/jira/browse/AVRO-1590) fixed it for that case, but I think this would require a different fix.

Another thing to consider here is the change brought in by PARQUET-139. Parquet used to need to read all of the footers in the MR driver to submit a job, which could take a very long time. Now, schema resolution is done in each task for that task's file. This means that the merge code for multiple files' metadata will be going away because the read preparation will be done on the client only. In that case, does it still make sense to add a check for each file schema in the MR driver process? That would defeat the purpose of PARQUET-139 and require reading all of the footers in a single process (bottleneck) right?

jeffreyolchovy:
I was unaware of the structural type of matching, but that's a pretty cool feature.

Unfortunately -- and I agree with you -- I think more is necessary to fix the problems I was experiencing in trying to get this to work with the `AvroParquetReader`. The solution in the PR always worked for the MR process, as the schema in each file was resolved to the schema supplied by the user. I feel like this would still have to happen, regardless if this evolution was done in each task or in the centralized initialization phase. I guess the issue here is that just single tasks would fail lazily if the schemas are not compatible?

We still rely on our custom implementation which is most similar to the first PR I took out for this. I assume other organizations must have experienced the same issue as well. And just to reiterate the use-case:

- Collect data with schema `Foo` on a daily basis for 15 days
- Modify `Foo` schema (in a compatible manner),
- Continue collecting for another 15 days
- Run a MR job across the 30 days of collected data

This will currently fail with a RuntimeException (could not merge metadata: key avro.schema has conflicting values)

ottomata:
Oo, sounds heavy.  I'm evaluating using Parquet + Avro for the Wikimedia Foundation.  This would certainly affect us as well if we decided to use Avro with our Parquet data.

rdblue:
Hi @ottomata, thanks for pinging this issue. It is a bit out of date and I haven't come back to it lately. Now is a good time.

@jeffreyolchovy, the latest releases of Parquet, both 1.6.0 and 1.7.0 include PARQUET-139. That fixes the "cannot merge metadata" error by always resolving the requested schema with each file schema individually. That behavior now matches what Avro does, so you shouldn't have another problem with the merging. In fact, we're planning to remove the metadata merging code path in the next major release.

For the issue where you want to use a different root record name in the Schema, I'll have a look later today and try to build a test case that fails. If there is a problem, I think it will require a fix on the Avro side, but we've been planning to do a 1.7.8 release there and I can make sure a fix for it makes it in.

@ottomata, I wouldn't worry too much about this particular schema evolution issue. Avro Schema evolution is used pretty widely and these issues are rare. If you need any more information on Parquet or the Avro support, let us know what we can do to help.

rdblue:
@jeffreyolchovy, I just wrote a test to [verify Parquet matches Avro's String behavior](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/test/java/org/apache/parquet/avro/TestStringBehavior.java). It writes with a generic schema, then reads with a schema reflected from a Java class with a different name. That appears to work just fine and is able to read the data correctly. Isn't that what we wanted to make sure works? Is your test different?

jeffreyolchovy:
@rdblue Hey, glad this is getting some attention as well. Just for clarification, I have no desire to "use a different root record name in the Schema" but that was one of the tests in the Parquet code base which failed with my changeset. I think that this is the wrong behavior altogether (it will fail the Avro library schema compatibility test, as documented above in the pastebin links). As far as I was concerned, I was not aware you can create some record under one name then read it back with another name just because the fields are compatible. That is what the existing test in Parquet expected (perhaps for ease of testing two compatible-but-different schemas?).

Moreover, the general issue that I was addressing with this changeset seems to be resolved using version >= 1.6.0rc5! It passed my integration tests and I will attempt to use the provided AvroReadSupport class more extensively in our staging environment. Feel free to close out this PR since I don't think it really applies anymore. Thanks!

rdblue:
Thanks for the update @jeffreyolchovy! Could you close this issue? We can't actually close them without a "closes X" commit message. The submitter can, though.

Also, be careful with the RC releases. Because those weren't verified as official releases, there were some minor API incompatibilities in there. I'm ensuring those are kept out of the real releases, which should be happening regularly again now that we are done with the Apache setup and are a top-level project.

PR #111, opened 2015-01-31
PARQUET-134 patch - Support file write mode

masokan:
Julien,
   I changed the integer constants to enum as you requested.  Please review the patch.

Thanks.


rdblue:
+1

Since Travis had an unrelated error, I tested this locally and it is passing both profiles.

rdblue:
Merged. Thanks @masokan!

PR #115, opened 2015-02-04
PARQUET-177: Added lower bound to memory manager resize

danielcweeks:
PARQUET-177

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
@@ -19,6 +19,7 @@
 package parquet.hadoop;
 
 import parquet.Log;
+import parquet.ParquetRuntimeException;
 
 import java.lang.management.ManagementFactory;
 import java.util.HashMap;
@@ -39,16 +40,19 @@ import java.util.Map;
 public class MemoryManager {
   private static final Log LOG = Log.getLog(MemoryManager.class);
   static final float DEFAULT_MEMORY_POOL_RATIO = 0.95f;
+  static final long DEFAULT_MIN_MEMORY_ALLOCATION = 0;
   private final float memoryPoolRatio;
 
   private final long totalMemoryPool;
+  private final long minMemoryAllocation;
   private final Map<InternalParquetRecordWriter, Long> writerList = new
       HashMap<InternalParquetRecordWriter, Long>();
 
-  public MemoryManager(float ratio) {
+  public MemoryManager(float ratio, long minAllocation) {
     checkRatio(ratio);
 
     memoryPoolRatio = ratio;
+    minMemoryAllocation = minAllocation;
     totalMemoryPool = Math.round(ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax
         () * ratio);
     LOG.debug(String.format("Allocated total memory pool is: %,d", totalMemoryPool));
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
```diff
@@ -39,16 +40,19 @@
 public class MemoryManager {
   private static final Log LOG = Log.getLog(MemoryManager.class);
   static final float DEFAULT_MEMORY_POOL_RATIO = 0.95f;
+  static final long DEFAULT_MIN_MEMORY_ALLOCATION = 0;
```
julienledem: We could have a better minimum than 0 which is in any case a bad value.
Possibly something related to the pageSize or the initial blockSize?

julienledem:
This looks good to me. I made a comment regarding the default value for this.

parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
danielcweeks: We could set it to something like 10MB, which is my best guess (without any empirical evidence).  We give tasks 1.5 GB which only safely allows for about 10 writers per task with the default 128MB Row Group setting.  A 10MB lower bound would probably push us to 100 writers, which would likely account for most situations.

Any other suggestions are welcome.

danielcweeks:
Ok, I updated the property name to be more clear and set the min to be
DEFAULT_BLOCK_SIZE / 10, which should allow for a lot of writers, but still
maintain a reasonable (though not most efficient) block size.

-Dan

rdblue:
This looks good to me.

I'm wondering if later we should add a minimum in terms of column chunk size. I know we can't guarantee it, but if we can take the number of columns into account we would have a more accurate minimum.

For example, if you have 10 columns and 128 MB, you get on average ~12MB column chunks. But if you have 100 columns, then you get on average 1.2MB column chunks. That's a big difference, so if we can make this check schema-dependent, it might be worth doing.

danielcweeks:
That's an excellent point, but would make it a lot harder to estimate the
number of writers (and apply it globally).

When you get into this situation, you've already started to degrade your
performance.  This is just a minor speed bump to keep someone from doing
something catastrophically bad.

I do like the idea for making column chunk min size an alternate setting
for better tuning where you need it, but we should probably make that a
separate JIRA.

julienledem:
This looks good to me

danielcweeks:
OK, I'll create a JIRA for capture Ryan's enhancement.

Throw me a plus and I'll give merging a shot.

rdblue:
I don't think you would need to change how the number of writers is affected. Just track the schema with the most columns and use that count to estimate the smallest chunk size. I agree this can be a separate JIRA issue.

You're right that when you hit this, you're already in a bad state. That's why I think it is important to produce a prominent warning and tell everyone to change their writer architecture instead of allowing this to dial down row group size.

danielcweeks:
Ryan, I agree fully that using the column approach would be more accurate (and not really harder to implement), but I'm having a harder time coming up with a sane default value.   Do we have any idea what a good minimum is for a column chunk.  Also, this would effectively be an average across the table.  You could still have tiny column chunks if your column data is significantly skewed.

I'll take a quick look at switching the implementation.

danielcweeks:
@rdblue is this more inline with what you were thinking.  I just used the page size for the default (i.e. average chunk size should never be smaller than a single page).

That might be too low, but possibly a good place to start.

rdblue:
@danielcweeks, page size is a great idea. That's already configurable and it makes no sense to have a column chunk less than a page.

rdblue:
+1

danielcweeks:
Thanks @rdblue!  I like this a lot better.

PR #97, opened 2015-01-04
[PARQUET-133]: upgrade snappy-java to 1.1.1.6

tigerquoll:
Upgraded snappy-java as requested by PARQUET-133.  Needed to do a global mvn clean before the tests passed cleanly though. Is there some cross contamination of test data?

PR #85, opened 2014-11-14
PARQUET-133: Upgrade snappy-java to 1.1.1.6

xerial:
Upgrade snappy-java to 1.1.1.6 (the latest vesrion), since 1.0.5 is no longer maintained in https://github.com/xerial/snappy-java, and 1.1.1.6 supports broader platforms including PowerPC, IBM-AIX 6.4, SunOS, etc. And also it has a better native coding loading mechanism (allowing to use snappy-java from multiple class loaders)

PR #116, opened 2015-02-04
[Don't Merge This] -- Just re-running some tests.

isnotinvain:


isnotinvain:
@julienledem have you seen these kinds of CI failures? Says it can't find some snappy classes?

julienledem:
is it caused by this? : https://github.com/apache/incubator-parquet-mr/pull/97
maybe try reverting this PR

isnotinvain:
Hmm, looks like #85 (duplicate of #97) came after this, so this is actually running something older than that. I'll merge and see if that helps.

PR #117, opened 2015-02-05
[PARQUET-36] Add support for dictionaries to FilteringPrimitiveConverter

MickDavies:
https://issues.apache.org/jira/browse/PARQUET-36

Please take a look. I have made some changes to Dictionary which need careful consideration and perhaps an alternative approach is better.

Also I made some changes to example implementation.

I am going to think about adding a few more tests.

Unfortunately I am having real problems building all project as I can get thrift 0.7 for my Mac. Does anyone know how to do this?


julienledem:
https://github.com/Parquet/parquet-mr/wiki/Developer-documentation#version-of-the-thrift-compiler-to-install

MickDavies:
Thanks - I think I tried these instructions a few weeks ago and I couldn’t compile thrift. I can’t remember why not, I think it may have been to do with yosemite.

I’ll give it another go.

Mick

MickDavies:
oops test is wrong - fixing now

MickDavies:
I have fixed the test I wrote which exercises the dictionary filtering for eq and neq cases. The generator produced some code for udfs. I'll look for some tests of this and add to those also.

MickDavies:
I am running into the issue that is detailed here - https://issues.apache.org/jira/browse/THRIFT-2229 <https://issues.apache.org/jira/browse/THRIFT-2229>. This prevents me building thrift 0.7, I’m not sure if there is a workaround.

Have other people got thrift 0.7 running on Mac OS 10.9 or above?

Mick

MickDavies:
Is there a reason why parquet continues to depend on this old version of thrift? I think I read somewhere it was used just for testing.

Would it be easy to upgrade?

Thanks

Mick

rdblue:
@MickDavies: I don't think you need to use thrift 0.7.0. I've built with 0.9.0 without problems, and newer versions (before 0.9.2 at least) should work.

parquet-column/src/main/java/parquet/column/Dictionary.java
```diff
@@ -29,10 +30,15 @@
...
 
   private final Encoding encoding;
+  private final PrimitiveType.PrimitiveTypeName primitiveTypeName;
 
-  public Dictionary(Encoding encoding) {
+  public Dictionary(PrimitiveType.PrimitiveTypeName primitiveTypeName, Encoding encoding) {
+    this.primitiveTypeName = primitiveTypeName;
     this.encoding = encoding;
   }
+  public PrimitiveType.PrimitiveTypeName getPrimitiveTypeName() {
```
rdblue: The pattern we use elsewhere is that the concrete classes include an implementation rather than having the abstract class carry an instance variable. That would be better for two reasons:

1. It would not require changing the constructor's signature
2. It ensures that the concrete classes always match their reported primitive type

Otherwise, I could instantiate a `PlainLongDictionary` and pass in `PrimitiveTypeName.BINARY` without triggering a complaint.

MickDavies: Thanks for the review, I think that the suggestion is good. It would mean creating a couple of new dictionary classes e.g. PlainInt96Dictionary. 

Are you happy with Dictionary carrying the type information, or at least having abstract method to get information. I did this to make adding dictionary support to SimplePrimitiveConverter simpler, and I added dictionary support to SimplePrimitiveConverter so I could test using existing framework.

parquet-column/src/main/java/parquet/example/data/Group.java
```diff
@@ -72,6 +73,8 @@ public Group getGroup(String field, int index) {
     return getGroup(getType().getFieldIndex(field), index);
   }
 
+  abstract public  void add(int fieldIndex, Primitive value);
+
```
julienledem: How is this change related to "Add support for dictionaries to FilteringPrimitiveConverter" ?

parquet-column/src/main/java/parquet/example/data/simple/convert/SimplePrimitiveConverter.java
```diff
@@ -18,13 +18,18 @@
...
+import parquet.io.ParquetDecodingException;
 import parquet.io.api.Binary;
 import parquet.io.api.PrimitiveConverter;
+import parquet.io.api.RecordConsumer;
 
 class SimplePrimitiveConverter extends PrimitiveConverter {
 
   private final SimpleGroupConverter parent;
   private final int index;
+  private DictionaryPrimitive[] dictionaryPrimitives;
```
julienledem: Shouldn't we reuse the already existing type specific dictionaries?

julienledem: https://github.com/apache/incubator-parquet-mr/blob/master/parquet-column/src/main/java/parquet/column/values/dictionary/PlainValuesDictionary.java

parquet-column/src/main/java/parquet/filter2/recordlevel/FilteringPrimitiveConverter.java
```diff
@@ -40,23 +40,25 @@ public FilteringPrimitiveConverter(PrimitiveConverter delegate, ValueInspector[]
...
   }
 
-  // TODO: this works, but
-  // TODO: essentially turns off the benefits of dictionary support
-  // TODO: even if the underlying delegate supports it.
-  // TODO: we should support it here. (https://issues.apache.org/jira/browse/PARQUET-36)
   @Override
   public boolean hasDictionarySupport() {
-    return false;
+    return delegate.hasDictionarySupport();
```
julienledem: this seems independent

parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
```diff
@@ -61,6 +64,24 @@
 
     private boolean result = false;
     private boolean isKnown = false;
+    private BitSet dictionaryResults;
+
+    /** Populates array of BitSet based on whether values pass filter */
+    void setDictionary(Dictionary dictionary) {
+      dictionaryResults = new BitSet(dictionary.getMaxId()+1);
+      for (int i = 0, end = dictionary.getMaxId()+1; i < end; i++) {
```
julienledem: How about the following?
```for (int i = 0; i <= dictionary.getMaxId(); i++) {```

File: parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java [hunk 6]
```diff
@@ -61,6 +64,24 @@
...
+      }
+    }
+
+    /** Update based on element from dictionary */
+    void updateFromDictionary(int dictionaryId) {
+      setResult(dictionaryResults.get(dictionaryId));
+    }
+
+    /** Does the element in the dictionary pass the filter */
+    abstract protected boolean update(Dictionary dictionary, int dictionaryId);
```
julienledem: this is strange to me. I think the goal it to make things more efficient by know the dictionary in advance.

julienledem: this is redundant with the update(type) methods below

parquet-column/src/main/java/parquet/example/data/Group.java
MickDavies: I added support for dictionaries to SimplePrimitiveConverter in order to aid testing. This change meant that  I could do the below. The changes to Simple... are not essential for the dictionary support but seemed like a good way to test

  public void addValueFromDictionary(int dictionaryId) {
    parent.getCurrentRecord().add(index, dictionaryPrimitives[dictionaryId]);
  }

parquet-column/src/main/java/parquet/example/data/simple/convert/SimplePrimitiveConverter.java
MickDavies: I think this was Ryan's point. I need to add abstract method to Dictionary, and probably a couple of new dictionaries covering Int96 and FixedLenByteArray

julienledem:
Hi @MickDavies 
Thanks for looking into this. I made some comments above.
There are two different features in there:
 - allow the dictionary ids to go through when the underlying converter supports them
 - make the filtering more efficient when there's a dictionary by knowing all the values before hand.
I'd like someone to add in the description of PARQUET-36 a short description of how we achieve these.
Does it sound good?
Thanks again!
Julien

parquet-column/src/main/java/parquet/filter2/recordlevel/FilteringPrimitiveConverter.java
MickDavies: Maybe. Its more work for FilteringPrimtive converter, it will need to call different methods on delegates but seems like a sensible idea

parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
MickDavies: Ill make this change

parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
MickDavies: Not sure I understand this point. SetDictionary will evaluate all filters when dictionary set. Then update sets the value for the column based on lookup against bitset

MickDavies:
Hi @julienledem 

I guess you could separate these two features, I hadn't done so in my mind. The former is important from measurements I have done for some datasets using Spark. For data I am using a query over a table with around 100M rows can run 25% faster.

I guess the second one just seemed like an obvious optimization

julienledem:
both features are useful and they will both bring gains for different reasons.
We don't need to split the JIRA in two but we should keep in mind those 2 distinct goals when reviewing.
I think there are several ways the second optimization could be done and that's why I think a short spec in the JIRA would be useful.

parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
julienledem: I'm looking at the API: in the examples below the callee either gets the value from the dictionary or ignores the dictionary. It sounds like we could make the specialization directly on the setDictionary call.
this is similar to the dictionary support in the converter: we look at the dictionary as a whole (isNull will always return false). If the implementation needs to lookup the value in dictionary we may as well directly call the update(type) corresponding method (That's what happens in converters). It does not need to go one level below.

parquet-column/src/main/java/parquet/example/data/simple/convert/SimplePrimitiveConverter.java
MickDavies: I have simplified this code and removed DictionaryPrimitive class. I think I still need a factory method to create correct type of Primitive (this is why I added type info to Dictionary). I have moved the factory method to Primitive. Let me know if you think there is a preferred approach.

parquet-column/src/main/java/parquet/column/Dictionary.java
MickDavies: I have made the change suggested above

parquet-column/src/main/java/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
MickDavies: I think there may be some confusion because I reused the update method name for this method, which is silly.
   /** Does the element in the dictionary pass the filter */
    abstract protected boolean update(Dictionary dictionary, int dictionaryId);
I will change it. It does not update anything, it is only called from setDictionary and is used to evaluate the predicate so that the result can be stored in the BitMap. I will change the name to evaluateFilter, or something simpler. 

The FilteringPrimitiveConverter then calls updateFromDictionary to make the update, which just accesses the value in the bitmap.

MickDavies:
Thanks for all the feedback, they comments were very useful.

 I have responded to most of the comments in the last check-in, with:
* Change method name update to evaluateFilterForDictionaryElement to clarify intent 
* Change Dictionary classes so that type information is abstract method associated with subclass, not carried as attribute 
* Change FilteringPrimitiveConverter so that it supports dictionaries even if delegates do not 
* Simplify implementation of SimplePrimitiveConverter by using type converter 

I still have to do the following:
* Update JIRA as requested by @julienledem 
* Add tests for dictionary filtering where delegate does not support dictionaries
* Ensure build works completely - try to get build going with thrift 0.9.1 working - looks positive
* Look at version conformance issues.

MickDavies:
I've added a unit test for FilteringPrimitiveConverter. I wanted to use a mocking library and saw that mockito and eaymock referenced in poms though only mockito seems to be being used, so I used that.

Also I upgraded to junit 4.12 from 4.10, so that I could set the name in Parameterized junit annotation, which makes the test output look a bit nicer, but is not essential.

Would you like me to tidy up poms a bit wrt to these dependencies. E.g. centralise version specification, and maybe remove easymock.

MickDavies:
I'm going to simplify this  work removing changes to example simple implementation

MickDavies:
Is anyone free to review, I would like to get this change in or rejected. Thanks

mpeleshenko:
Can this MR be reviewed please? This is a big performance blocker when reading a large parquet file with many String columns using a Filter as new String instances are created for each record read even though cardinality is low.

parquet-column/src/main/java/parquet/filter2/recordlevel/FilteringPrimitiveConverter.java
```diff
@@ -34,29 +35,38 @@
...
   }
 
-  // TODO: this works, but
-  // TODO: essentially turns off the benefits of dictionary support
-  // TODO: even if the underlying delegate supports it.
-  // TODO: we should support it here. (https://issues.apache.org/jira/browse/PARQUET-36)
   @Override
   public boolean hasDictionarySupport() {
-    return false;
+    return true;
```
mpeleshenko: Shouldn't this return `delegate.hasDictionarySupport()`?

MickDavies: Hi

I'm not sure. It's 4 years since this PR, so it's pretty stale now.

I could try to resurrect it but I don't know if there is much interest from the project.

Mick

PR #118, opened 2015-02-05
PARQUET-180: Update use of TBinaryProtocol#setReadLength.

rdblue:
This is no longer supported in thrift 0.9.2 and was only used
defensively. The reason to remove it now is to avoid linker errors when
the wrong version of thrift is found in the classpath.


julienledem:
I commented on PARQUET-180
I'd like to keep this, possibly using reflection (once, not every time) to gate the call.

rdblue:
Okay, I've updated this as suggested and verified that it compiles with libthrift 0.9.0 and 0.7.0, but can use 0.9.2 at runtime.

julienledem:
LGTM.
Maybe keep the comment that explains what this does?

PR #119, opened 2015-02-10
PARQUET-164: Add warning when scaling row group sizes.

rdblue:


parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
```diff
@@ -55,7 +55,7 @@ public MemoryManager(float ratio, long minAllocation) {
     minMemoryAllocation = minAllocation;
     totalMemoryPool = Math.round(ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax
         () * ratio);
-    LOG.debug(String.format("Allocated total memory pool is: %,d", totalMemoryPool));
+    LOG.info(String.format("Allocated total memory pool is: %,d", totalMemoryPool));
```
julienledem: I'm not sure we want this to show up in the case where the memory manager does not do anything special (scale == 1).
It will add noise for user that have only one writer.

julienledem:
Just one comment otherwise, this looks good to me.

parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
rdblue: Good point, especially when the warning I added when the allocation is changed now includes this information. I've changed this back to debug.

PR #121, opened 2015-02-13
PARQUET-187: Replace JavaConversions.asJavaList with JavaConversions.seqAsJavaList

colinmarc:
The former was removed in 2.11, but the latter exists in 2.9, 2.10 and 2.11. With this change, I can build on 2.11 without any issue.


julienledem:
SGTM
+1 once travis has run.

PR #123, opened 2015-02-23
PARQUET-162: ParquetThrift should throw when unrecognized columns are passed to the column projection API

tsdeng:
ParquetThrift should throw when unrecognized columns are passed to the column projection API

### DIFF
--- a/parquet-thrift/src/main/java/parquet/thrift/projection/FieldProjectionFilter.java
+++ b/parquet-thrift/src/main/java/parquet/thrift/projection/FieldProjectionFilter.java
@@ -28,21 +28,42 @@ import java.util.List;
  */
 public class FieldProjectionFilter {
   public static final String PATTERN_SEPARATOR = ";";
-  List<PathGlobPattern> filterPatterns;
+  List<PathGlobPatternStatus> filterPatterns;
+
+  /**
+   * Class for remembering if a glob pattern has matched anything.
+   * If there is an invalid glob pattern that matches nothing, it should throw.
+   */
+  private class PathGlobPatternStatus{
+    PathGlobPattern pattern;
+    boolean hasMatchingPath = false;
+    PathGlobPatternStatus(String pattern){
+      this.pattern =new PathGlobPattern(pattern);
+    }
+
+    public boolean matches(FieldsPath path) {
+      if (this.pattern.matches(path.toString())) {
+        this.hasMatchingPath = true;
+        return true;
+      }else{
+        return false;
+      }
+    }
+  }
 
   public FieldProjectionFilter() {
-    filterPatterns = new LinkedList<PathGlobPattern>();
+    filterPatterns = new LinkedList<PathGlobPatternStatus>();
   }
 
   public FieldProjectionFilter(String filterDescStr) {
-    filterPatterns = new LinkedList<PathGlobPattern>();
+    filterPatterns = new LinkedList<PathGlobPatternStatus>();
 
     if (filterDescStr == null || filterDescStr.isEmpty())
       return;
 
     String[] rawPatterns = filterDescStr.split(PATTERN_SEPARATOR);
     for (String rawPattern : rawPatterns) {
-      filterPatterns.add(new PathGlobPattern(rawPattern));
+      filterPatterns.add(new PathGlobPatternStatus(rawPattern));
     }
   }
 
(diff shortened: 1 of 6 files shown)


isnotinvain:
Right now we are treating the glob path as a glob over a string and allowing users to do string math on their paths. 

For example, we'd allow:

```
foo/bar/{a,b,c}
```

which would match any columns `foo/bar/a` if it existed, `foo/bar/b` if it existed, and `foo/bar/c` if it existed, but would throw if *none* of them existed.

So there are two things that @tsdeng and I discussed about this.

1) We should definitely take any expression with `{}` sections in them, and explode to all posibilities, and verify that all of them are valid columns.

2) We currently allow users to do globs over partial column names, which I don't think is useful. For example:

```
foo/bar/baz{a,b,c}
```
will match `foo/bar/baza`, `foo/bar/bazb`, `foo/bar/bazc`
and
```
foo/bar/aaa{x,y}/z
```

will match `foo/bar/aaax/z` and `foo/aaay/z`

I think that it would make sense to restrict our syntax to not be arbitrary string patterns, but instead to be patterns of tokens, where tokens are column names. So anything in a `{}` must itself be a column name, not a portion of one, eg:
```
foo/{bar,baz}/z
foo/bar/{x,y,z}
```
where bar, baz, x, y, and z are all column names, but
```
foo/b{ar,az}/z
```
would not be allowed because `ar` and `az` are not column names.

Thoughts? // @tsdeng @julienledem

isnotinvain:
Oh, at the same time, we could also drop support for `?` and other regex things like that.

isnotinvain:
So the simplest form of the new syntax would be:

1) tokens = split the input on `/`
2) each token must be `*`, `**`, a column name, or `.startWith("{")` and `.endWith("}")`, in which case
2.1) sub_tokens = split the `{...}` section on `,` and all of those must be column names

sriramkrishnan:
I am all in favor of keeping it simple. So +1 to @isnotinvain's previous proposal.

Another gotcha is that the convention for filter and projection pushdown are inconsistent. IIRC you need to use a.b for the filter predicate, and a/b for projection pushdown. Would be great if we can standardize on one.

isnotinvain:
There's even 1 more, the filters are done on the parquet schema, but the projections are done on the thrift schema...

isnotinvain:
@tsdeng this PR looks good, can you add logging that says "Schema after projection: %s" or something like that so we can see what columns are being kept? Or even just print out the full schema annotated, like:
```
Foo {
  bar **skipped by projection push down**
  baz
}
```

parquet-thrift/src/main/java/parquet/thrift/projection/FieldProjectionFilter.java
```diff
@@ -28,21 +28,43 @@
...
 public class FieldProjectionFilter {
   public static final String PATTERN_SEPARATOR = ";";
-  List<PathGlobPattern> filterPatterns;
+  List<PathGlobPatternStatus> filterPatterns;
+
+  /**
+   * Class for remembering if a glob pattern has matched anything.
+   * If there is an invalid glob pattern that matches nothing, it should throw.
+   */
+  private class PathGlobPatternStatus {
```
isnotinvain: this should be a static class

parquet-thrift/src/test/java/parquet/thrift/TestThriftSchemaConverter.java
```diff
@@ -202,9 +203,29 @@ public void testProjectOnlyKeyInMap() {
...
 
+
+  private void shouldThrowWhenProjectionFilterMatchesNothing(String filters, String unmatchedFilter, Class<? extends TBase<?, ?>> thriftClass) {
+    try {
+      getFilteredSchema(filters, thriftClass);
+    } catch (ThriftProjectionException e) {
+      assertEquals("unmatched projection filters: [" + unmatchedFilter + "]", e.getMessage());
+      return;
+    }
+    fail("should throw projection exception when filter matches nothing");
```
isnotinvain: you can just move this call to fail() into the try block, eg:

```
try {
  getFilteredSchema(filters, thriftClass);
  fail("should throw projection exception when filter matches nothing");
} catch {
  assertEquals("unmatched projection filters: [" + unmatchedFilter + "]", e.getMessage());
}

and then you don't need a return inside of the catch block

isnotinvain:
Two minor comments, but LGTM / +1
We can revisit the new syntax / strictness / consistency with the filter2 API later.

isnotinvain:
Oh, if the logging is easy to add, lets add that here too. If not, we can follow up separately.

PR #124, opened 2015-02-25
PARQUET-189:Support building parquet with thrift 0.9.0

winningsix:
Summary:
   add one profile to enable thrift with version 0.9.0

### DIFF
--- a/parquet-thrift/pom.xml
+++ b/parquet-thrift/pom.xml
@@ -115,6 +115,12 @@
       <version>3.4</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
+      <version>${thrift.version}</version>
+      <scope>provided</scope>
+    </dependency>
 
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -114,6 +114,7 @@
     <scala.maven.test.skip>false</scala.maven.test.skip>
     <pig.version>0.11.1</pig.version>
     <pig.classifier></pig.classifier>
+    <thrift.version>0.7.0</thrift.version>
   </properties>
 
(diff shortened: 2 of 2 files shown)


jfarrell:
thanks for the patch @winningsix, why 0.9.0 and not the latest 0.9.2?

winningsix:
Thank you for your comments @jfarrell I tried 0.9.2 but failed since <noformat>ThriftBytesWriteSupport</noformat> is using a old API <noformat>setReadLength</noformat> from <noformat>TBinaryProtocol</noformat> which does not exist in 0.9.2.

winningsix:
Please refer to PARQUET-180

rdblue:
Why does this add the thrift dependency to parquet-thrift? Was it a transitive dependency before?

winningsix:
Hi @rdblue, it is a transitive dependency. By adding this dependency with version specified, I can build the parquet project with thrift 0.9.

parquet-thrift/pom.xml
```diff
@@ -115,6 +115,11 @@
       <version>3.4</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
+      <version>${thrift.version}</version>
```
rdblue: Good to add this because there are direct dependencies on it. Usually it's provided by the application with thrift classes though, so could you change the scope to provided?

rdblue:
#118 fixes the problem with 0.9.2. Although it won't build, it passes tests. I just had one more comment, then I can merge this. Thanks @winningsix!

winningsix:
Thanks @rdblue for your comments. I have update my patch in the latest pull request.

rdblue:
I merged this into master by hand since it had a minor conflict, if anyone has a close button, please close this.

PR #125, opened 2015-02-25
PARQUET-190: fix an inconsistent Javadoc comment of ReadSupport.prepareForRead

choplin:
ReadSupport.prepareForRead does not return RecordConsumer but RecordMaterializer


rdblue:
Thanks, @choplin!

PR #127, opened 2015-02-26
PARQUET-192: Fix map null encoding

rdblue:
This depends on PARQUET-191 for the correct schema representation.

PR #128, opened 2015-02-26
PARQUET-193: Implement nested types compatibility rules in Avro

rdblue:
This depends on PARQUET-191 and PARQUET-192.

This replaces #83.


rdblue:
Resolved issues from Tom's review of the previous PR:

* ~~The configuration property should have a better name~~
  * Renamed to `parquet.avro.add-list-element-records`, which describes the result
* Use `SchemaBuilder` in tests
  * I started to switch, but the result was longer, harder to read, and error prone so I would prefer to keep the current implementation
* ~~There are still write-side changes~~
  * Moved into #126 and #127 because they are separate issues
* ~~Remove changes to ConversionPatterns~~
  * Removed; will add for write-side changes.

Not resolved:
* Configuration property should be documented
  * Anyone know a good place for these? Should we start a doc for them?

rdblue:
Rebased on master to pull in PARQUET-191 and PARQUET-192.

tomwhite:
+1 this looks good to me. Thanks for addressing my earlier comments.

PR #129, opened 2015-03-03
PARQUET-188: Change column ordering to match the field order.

rdblue:
This was the behavior before the V2 pages were added.


julienledem:
+1 LGTM
thanks for fixing this.

PR #130, opened 2015-03-03
PARQUET-41 Add bloom filters to parquet statistics

winningsix:
This patch is not ready yet and is used for the discussion. 
1. Currently bloom filter supports long type only and will add other types support later. 
2. The expected number of entries is set to a fixed number. It will be configurable as a follow-up jiras.
3. The bloom filter will be row group level only.

For the item#2, there are two approaches that one is constructing the bloom filter when flushing the data to the page and another way is using the initialSizePerCol when creating the dataColumn.


isnotinvain:
I made some comments here:
https://issues.apache.org/jira/browse/PARQUET-41

I think we'll need some planning around how we want to approach new complex types of statistics in parquet. We need to make sure the serialization format is stable / durable, well specified for use outside of the jvm. We've also avoided adding a dependency on guava to parquet so far.

PR #131, opened 2015-03-03
PARQUET-197 : fix parquet-cascading not writing parquet metadata file

tsdeng:
Repro: run a scalding job that writes parquet files to a folder. no _metadata and _common_metadata file is created
Impact: potential performance problem if parquet metadata is read from client side, which is the case for sparkSQL
casue: the metatdata writing logic is in the mapreduce API but not the mapred API of parquet.

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/hadoop/mapred/DeprecatedParquetOutputCommitter.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/mapred/DeprecatedParquetOutputCommitter.java
@@ -1,0 +1,23 @@
+package parquet.hadoop.mapred;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapred.*;
+import parquet.hadoop.ParquetOutputCommitter;
+
+import java.io.IOException;
+
+/**
+ *
+ * Adapter for supporting ParquetOutputCommitter in mapred API
+ *
+ * @author Tianshuo Deng
+ */
+public class DeprecatedParquetOutputCommitter extends FileOutputCommitter {
+
+  @Override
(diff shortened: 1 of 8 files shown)


parquet-hadoop/src/main/java/parquet/hadoop/mapred/DeprecatedParquetOutputCommitter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.IOException;
+
+/**
+ *
+ * Adapter for supporting ParquetOutputCommitter in mapred API
+ *
+ * @author Tianshuo Deng
+ */
+public class DeprecatedParquetOutputCommitter extends FileOutputCommitter {
```
julienledem: I'd prefer we say "MapredParquetOutputCommitter"
using "Deprecated" to refer to the deprecated but still heavily in use mapred api is confusing to users. "mapred" is better qualifier.

julienledem:
Just one comment otherwise this looks good to me (assuming the build passes).
Please check why the build fails

tsdeng:
Thanks @julienledem , I just updated the PR with adding a new method setAsOutputFormat to set both the output format class and output committer. Use this method then we don't forget to set the ParquetOutputCommiter when using DeprecatedParquetOutputFormat. Could you take a look again with the new changes?

julienledem:
LGTM.
+1

isnotinvain:
+1, thanks!

isnotinvain:
testing here: https://github.com/isnotinvain/incubator-parquet-mr/pull/3

isnotinvain:
Tested failed:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/53133792/log.txt

PR #133, opened 2015-03-05
PARQUET-186: Fix Precondition performance problem in SnappyUtil.

rdblue:
This fixes the problem by adding string formatting to the preconditions. This avoids any string formatting unless the precondition throws an Exception. We should check for string operations in other tight loops as well.


isnotinvain:
+1

rdblue:
I added tests I forgot in the first version and added the old method signatures back because semver caught that it breaks compatibility.

rdblue:
Tests are passing locally for both profiles so I merged this. Thanks for the review, @isnotinvain!

PR #136, opened 2015-03-06
PARQUET-204: add parquet-schema directory support

nevillelyh:


tsdeng:
This is good. +1
Thanks @nevillelyh

parquet-hadoop/src/main/java/parquet/hadoop/util/HiddenFileFilter.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+
+public class HiddenFileFilter implements PathFilter {
+  @Override
+  public boolean accept(Path p) {
+    return !p.getName().startsWith("_") && !p.getName().startsWith(".");
+  }
+}
```
julienledem: As this is stateless there could be one static instance of it.
HIDDEN_FILE_FILTER = new HiddenFileFilter();

julienledem:
Thanks for cleaning up all this duplicated code.
+1

parquet-hadoop/src/main/java/parquet/hadoop/util/HiddenFileFilter.java
rdblue: Agree

nevillelyh:
I changed it to a static INSTANCE member.

parquet-tools/src/main/java/parquet/tools/command/ShowSchemaCommand.java
```diff
@@ -68,7 +71,21 @@ public void execute(CommandLine options) throws Exception {
...
+    if (fs.isDirectory(path)) {
+      FileStatus[] statuses = fs.listStatus(path, HiddenFileFilter.INSTANCE);
+      if (statuses.length == 0) {
+        throw new RuntimeException("Directory " + path.toString() + " is empty");
+      }
+      file = statuses[0].getPath();
+    } else {
+      file = path;
+    }
+    metaData = ParquetFileReader.readFooter(conf, file);
```
liancheng: Should we take schema evolution into account here? I.e., show the merged schemas of all part-files.

rdblue: I don't think so. There's no guarantee that there is a single schema for the data and merging all of the schemas into one would be misleading: a union strategy can produce a schema that can't be satisfied (as a column projection) by any of the files. I think it's best to return one or all of the unique schemas, but this is already going slightly beyond what Parquet itself should be doing as a file format. Parquet reads and writes files, while Hive or Kite manages the data as a collection.

rdblue:
Thanks @nevillelyh!

PR #137, opened 2015-03-06
PARQUET-205: skip _SUCCESS file in RAT

nevillelyh:


nevillelyh:
Not a real issue. Can be avoided by dropping test files in target/

PR #140, opened 2015-03-09
PARQUET-210: add JSON support for parquet-cat

nevillelyh:
JSON output with this patch:
```
{"int_field":99,"long_field":1099,"float_field":2099.5,"double_field":5099.5,"boolean_field":true,"string_field":"str99","nested":{"numbers":[100,101,102,103,104],"name":"name99","dict":{"a":100,"b":200,"c":300}}}
```

Current output format:
```
int_field = 99
long_field = 1099
float_field = 2099.5
double_field = 5099.5
boolean_field = true
string_field = str99
nested:
.numbers:
..array = 100
..array = 101
..array = 102
..array = 103
..array = 104
.name = name99
.dict:
..map:
...key = a
...value = 100
..map:
...key = b
...value = 200
..map:
...key = c
...value = 300
```

### DIFF
--- a/parquet-tools/src/main/java/parquet/tools/command/CatCommand.java
+++ b/parquet-tools/src/main/java/parquet/tools/command/CatCommand.java
@@ -18,6 +18,9 @@ package parquet.tools.command;
 import java.io.PrintWriter;
 
 import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.OptionBuilder;
+import org.apache.commons.cli.Options;
 import org.apache.hadoop.fs.Path;
 
 import parquet.hadoop.ParquetReader;
@@ -31,6 +34,15 @@ public class CatCommand extends ArgsOnlyCommand {
     "where <input> is the parquet file to print to stdout"
   };
 
+  public static final Options OPTIONS;
+  static {
+    OPTIONS = new Options();
+    Option help = OptionBuilder.withLongOpt("json")
+                               .withDescription("Show records in JSON format.")
+                               .create('j');
+    OPTIONS.addOption(help);
+  }
+
   public CatCommand() {
     super(1, 1);
   }
@@ -38,6 +50,11 @@ public class CatCommand extends ArgsOnlyCommand {
   @Override
   public String[] getUsageDescription() {
     return USAGE;
+  }
+
+  @Override
+  public Options getOptions() {
+    return OPTIONS;
   }
 
   @Override
@@ -52,7 +69,11 @@ public class CatCommand extends ArgsOnlyCommand {
       PrintWriter writer = new PrintWriter(Main.out, true);
       reader = new ParquetReader<SimpleRecord>(new Path(input), new SimpleReadSupport());
       for (SimpleRecord value = reader.read(); value != null; value = reader.read()) {
-        value.prettyPrint(writer);
+        if (options.hasOption('j')) {
+          value.prettyPrintJson(writer);
+        } else {
+          value.prettyPrint(writer);
+        }
         writer.println();
       }
     } finally {
--- a/parquet-tools/src/main/java/parquet/tools/read/SimpleRecord.java
+++ b/parquet-tools/src/main/java/parquet/tools/read/SimpleRecord.java
@@ -15,17 +15,18 @@
  */
 package parquet.tools.read;
 
+import java.io.IOException;
 import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.Maps;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.node.BinaryNode;
 
 public class SimpleRecord {
   public static final int TAB_SIZE = 2;
-  private final List<NameValue> values;
+  protected final List<NameValue> values;
 
   public SimpleRecord() {
     this.values = new ArrayList<NameValue>();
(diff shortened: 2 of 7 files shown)


rdblue:
This is a great addition, I agree that it is nice to have the ability to have both a detailed view and a JSON-ish view. The main problem I see is that we are claiming that this outputs JSON.

I'd rather output something like JSON but call it a "pretty" view instead to avoid making making guarantees that the data is valid JSON. If you want to output valid JSON, then this should instead use a JSON library like Jackson and we should define mappings from the various primitive and logical types to a JSON representation.

For example, this represents binary as a String that contains all of the byte values. This should be documented and we should think about the most reasonable representation (base64?).

parquet-tools/src/main/java/parquet/tools/read/SimpleRecord.java
```diff
@@ -97,6 +99,92 @@ public void prettyPrint(PrintWriter out, int depth) {
...
+    } else if (double[].class == val.getClass()) {
+      out.print(Arrays.toString((double[])val));
+    } else if (boolean[].class == val.getClass()) {
+      out.print(Arrays.toString((boolean[]) val));
+    } else if (val.getClass().isArray()) {
+      out.print(Arrays.deepToString((Object[])val));
+    } else if (SimpleRecord.class.isAssignableFrom(val.getClass())) {
+      ((SimpleRecord) val).prettyPrintJson(out);
+    } else if (val.getClass() == String.class) {
+      out.print("\"" + String.valueOf(val) + "\"");
```
rdblue: JSON requires certain characters, like newline, to be escaped. This is done automatically with Jackson, which I would recommend using instead of building JSON manually.

julienledem:
Thanks @nevillelyh !
I agree with @rdblue this would be a great addition and we should use a JSON library (jackson is already a dependency) to generate JSON, not concatenate strings.
binary representation could be configurable (if it is of OriginalType String it should be the string, otherwise probably an array of numbers?)
It should either be valid JSON or not look like JSON in my opinion.

rdblue:
bq. binary representation could be configurable

Jackson standardizes this using base64, which I think makes the most sense. I also want to avoid this being used as a Parquet-to-JSON converter . . . but maybe we shouldn't worry too much about that.

nevillelyh:
@rdblue @julienledem I changed it to use Jackson and the code actually looks cleaner now. Do we want any special treatment for binary?

parquet-tools/src/main/java/parquet/tools/command/CatCommand.java
```diff
@@ -52,7 +69,11 @@ public void execute(CommandLine options) throws Exception {
       PrintWriter writer = new PrintWriter(Main.out, true);
       reader = new ParquetReader<SimpleRecord>(new Path(input), new SimpleReadSupport());
       for (SimpleRecord value = reader.read(); value != null; value = reader.read()) {
-        value.prettyPrint(writer);
+        if (options.hasOption('j')) {
+          value.prettyPrintJson(writer);
```
julienledem: I think that we should generate the JSON based on the parquet schema (It is passed to the SimpleReadSupport from the file's metadata) rather than inferring isArray and isMap from the fields in each object.
I would make prettyPrintJson not part of SimpleRecord but instead:
prettyPrintJson(value, schema, writer)
and traverse schema at the same time as value and use schema to decide if it should be represented as a map, an array or a string (OriginalType is List, Map or UTF8).

julienledem: Another approach is to use different subtypes of SimpleRecord (ListSimpleRecord, MapSimpleRecord, ...) for this.

julienledem:
Thanks @nevillelyh, that looks great. I made a comment about isArray and isMap. I think we are getting close.

parquet-tools/src/main/java/parquet/tools/command/CatCommand.java
nevillelyh: Sure I'll go for the first approach and factor this out to a util class then?
Don't have a lot of time this week so it might take a bit longer.

julienledem: At your preference. This sounds good. Thanks again for contributing. The main thing I care about is the isArray and isMap decision to be driven by the schema more explicitly. 

The second approach could be as simple as creating subclasses of SimpleRecordConverter for Map and List that override start() to create a new Simple{List,Map}Record() with toJsonObject overridden to do the right thing. They would be used there:
https://github.com/apache/incubator-parquet-mr/blob/a0c77b6a442e2c4a355a4b145898bed976f23bb4/parquet-tools/src/main/java/parquet/tools/read/SimpleRecordConverter.java#L64

rdblue:
> Do we want any special treatment for binary?

Yes, I think Julien is right that we should convert to JSON according to the Parquet schema. That means binary should be handled consistently as binary. With Jackson, you can tell it that a value is binary and it will encode it for you.

nevillelyh:
@julienledem I went with 2nd approach after all but here it is.
@rdblue can you point me to the Jackson binary settings docs?

rdblue:
The Jackson representation you want is [`BinaryNode`](https://fasterxml.github.io/jackson-databind/javadoc/2.5/com/fasterxml/jackson/databind/node/BinaryNode.html)

nevillelyh:
@rdblue sorry for the late response. I changed it to use BinaryNode/base64 for both default and JSON formats.

rdblue:
This looks good to me, but when I went to test it my cluster's version of parquet-tools is getting loaded first. Can someone test that this works as expected?

nevillelyh:
@rdblue try `hadoop jar target/parquet-tools*.jar cat`?

rdblue:
Got it working with `mvn dependency:copy-dependencies` and copying the jar into target/dependency. Then run with: `java -cp 'target/dependency/*' parquet.tools.Main cat`.

I'd need to play with classpath ordering on my cluster. Good to know the current jar isn't compatible with CDH, which is still doing packaging with the tarball.

PR #142, opened 2015-03-09
PARQUET-214: Fix Avro string regression.

rdblue:
At some point, parquet-avro converted string fields to binary without
the UTF8 annotation. The change in PARQUET-139 to filter the file's
schema using the requested projection causes a regression because the
annotation is not present in some file schemas, but is present in the
projection schema converted from Avro.

This reverts the projection change to avoid a regression in a release.
Fixing the projection as in PARQUET-139 will need to be done as a
follow-up.


parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -81,7 +83,7 @@ public ReadContext init(Configuration configuration,
     }
     // use getSchemaForRead because it checks that the requested schema is a
     // subset of the columns in the file schema
```
tomwhite: Remove the comment too, so it's not out of sync with the code.

tomwhite:
+1

PR #143, opened 2015-03-10
PARQUET-217 Use simpler heuristic in MemoryManager

isnotinvain:
We found that the heuristic of throwing when:
```
minMemoryAllocation > 0 && newSize/maxColCount < minMemoryAllocation
```
in MemoryManager is not really valid when you have many (3k +) columns, due to the division by the number of columns.
This check throws immediately when writing a single file with a 3GB heap and > 3K columns.

This PR introduces a simpler heuristic, which is a min scale, and we throw when the MemoryManager's scale gets too small. By default I chose 25%, but I'm happy to change that to something else.

For backwards compatibility I've left the original check in, but it's not executed by default anymore, to get this behavior the min chunk size will have to be set in the hadoop configuration. I'm also open to removing it entirely if we don't think we need it anymore.

What do you think?
@danielcweeks @rdblue @dongche @julienledem

### DIFF
--- a/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
@@ -122,8 +122,8 @@ public class MemoryManager {
     for (Map.Entry<InternalParquetRecordWriter, Long> entry : writerList.entrySet()) {
       long newSize = (long) Math.floor(entry.getValue() * scale);
       if(scale < 1.0 && minMemoryAllocation > 0 && newSize < minMemoryAllocation) {
-          throw new ParquetRuntimeException(String.format("New Memory allocation %d" +
-          " is smaller than the minimum allocation size of %d",
+          throw new ParquetRuntimeException(String.format("New Memory allocation %d bytes" +
+          " is smaller than the minimum allocation size of %d bytes.",
               newSize, minMemoryAllocation)){};
       }
       entry.getKey().setRowGroupSizeThreshold(newSize);
--- a/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
+++ b/parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
@@ -40,39 +40,27 @@ import java.util.Map;
 public class MemoryManager {
   private static final Log LOG = Log.getLog(MemoryManager.class);
   static final float DEFAULT_MEMORY_POOL_RATIO = 0.95f;
-  static final float DEFAULT_MIN_SCALE = 0.25f;
+  static final long DEFAULT_MIN_MEMORY_ALLOCATION = ParquetWriter.DEFAULT_PAGE_SIZE;
   private final float memoryPoolRatio;
 
   private final long totalMemoryPool;
   private final long minMemoryAllocation;
-  private final float minScale;
   private final Map<InternalParquetRecordWriter, Long> writerList = new
       HashMap<InternalParquetRecordWriter, Long>();
 
-  public MemoryManager(float ratio, float minScale) {
-    this(ratio, -1, minScale);
-  }
-
-  @Deprecated
   public MemoryManager(float ratio, long minAllocation) {
-    this(ratio, minAllocation, DEFAULT_MIN_SCALE);
-  }
+    checkRatio(ratio);
 
-  @Deprecated
-  MemoryManager(float ratio, long minAllocation, float minScale) {
-    checkRatio(ratio, "memory pool ratio");
-    checkRatio(minScale, "min scale");
-
-    this.memoryPoolRatio = ratio;
-    this.minMemoryAllocation = minAllocation;
-    this.minScale = minScale;
-    totalMemoryPool = Math.round(ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax() * ratio);
+    memoryPoolRatio = ratio;
+    minMemoryAllocation = minAllocation;
+    totalMemoryPool = Math.round(ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax
+        () * ratio);
     LOG.debug(String.format("Allocated total memory pool is: %,d", totalMemoryPool));
   }
 
-  private void checkRatio(float ratio, String name) {
+  private void checkRatio(float ratio) {
     if (ratio <= 0 || ratio > 1) {
-      throw new IllegalArgumentException("The configured " + name + " " + ratio + " is " +
+      throw new IllegalArgumentException("The configured memory pool ratio " + ratio + " is " +
           "not between 0 and 1.");
     }
   }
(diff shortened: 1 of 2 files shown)


dongche:
Hi @isnotinvain , thanks for figuring out this problem. How about we combine these two heuristic together?

Using chunk size is more accurate normally and make sense. For many(3K+) columns case, I guess it might be sparse table or skewed data, and we can use the min scale you write. If so, we could set a default columns number boundary (1K?) for switching between the chunk size limits and your block size limit. I am not sure 25% is ok or not. Let's see other guys' opinion. 

How does this sound?

isnotinvain:
@dongche what do we gain by having the `newSize/maxColCount < minMemoryAllocation` check?
My understanding is that as the chunk size is scaled down, at some point we want to stop and say "ok, that is too much scaling down" and stop writing (we are out of memory at this point essentially).

The check I added does that by inspecting the scale factor, with the assumption that it is fairly easy to describe "what scale factor is too small" based on percent of the original size.

What worries me about the `newSize/maxColCount < minMemoryAllocation` is two things:
1) We shouldn't be doing any checks at all when we are writing a single file, and are not constrained for memory (scale factor of 1). This feature is here to support a different use case, but should not interfere with the more common case of writing a single file.

2) I'm not convinced that `newSize/maxColCount` is a fair assumption even with a smaller number of columns, it essentially assumes every column is of equal size, which I think is often not the case. I think it's reasonable to expect that schemas with optional fields will often have columns that are often empty. This skew gets even more dramatic when the optional field itself has a large schema.

By not taking into account the number of columns at all, I think we get something that more directly measures what we care about, namely "are we scaling back too much?" 

What do you think? Thanks!

julienledem:
As Alex is explaining, the problem with the way the current code is checking for min size is that it is expecting all columns to have the same amount of data in it which is rarely true and makes it just wrong in practice.
As the goal is to prevent row groups to become too small I think it is probably better to test for that instead as is proposed in this PR.
The scaling down of row groups will only happen if there are sufficiently many writers at the same time that they fill the memory. The point is we want to prevent writing files that we know will not be usable because performance characteristics will be heavily degraded by tiny row groups. 1/4th of what was expected is already pretty small.

In short I like the proposed change. Possibly we can have a more conservative minimum but probably not.

isnotinvain:
Tests passed locally for me (h1 and h2), running them in my own queue here too:
https://github.com/isnotinvain/incubator-parquet-mr/pull/4

isnotinvain:
Anyone want to weigh in on whether we should:
1) keep the old heuristic but have it off by default (current state of this PR)
2) drop the old heuristic
3) combined the old and new heuristic with an additional check for number of columns

I'm leaning towards 2) but if we want backwards compatibility then 1) seems OK too.

@danielcweeks @rdblue @dongche @julienledem

rdblue:
> the problem with the way the current code is checking for min size is that it is expecting all columns to have the same amount of data

I was interpreting this another way: if there isn't enough space for one page per column in memory then there is probably an issue. I guess what you're saying is that for sparse schemas in practice, you end up with some columns aren't even a full block. So maybe a full block isn't a reasonable limit, but I think the logic of some minimum per column is sound. Half a block? 20% of a block? 200K? Is there something reasonable here we could use?

> We shouldn't be doing any checks at all when we are writing a single file

Agreed. Let's make sure all of the logic only applies when there is more than one open file.

rdblue:
> I'm not convinced that newSize/maxColCount is a fair assumption even with a smaller number of columns, it essentially assumes every column is of equal size, which I think is often not the case.

The idea is to assume that each column has *some* size that grows with the number of records. If that size isn't big enough then we have a problem. In practice, we might not get columns that have even one entire block, but I think basing this on some per-column minimum is still the right thing.

So for reasonably well-encoded columns what are we looking at for a minimum?

Another way to fix this is to set the default block size to something better. I talked with some Impala folks and found out they're using 64k blocks. That would increase the number of columns that would pass this check by 16x.

isnotinvain:
I guess my question is, why do we need an absolute minimum instead of using the scale factor? The scale factor is being applied to an absolute value that's been supplied, so we can assume that absolute value is the anchor, and when we try to scale that value too low proportionally, then we've gone too far?

julienledem:
I think thhe only reasonable minimum for a column is 0 because it will happen pretty often that a column defined in the schema will actually have no data in it. Assuming each column has one page is actually pretty far from reality (as we realized now that we tried this).
As we have a row group byte size goal, it seems reasonable to me to express the minimum in terms of minimum row group size goal and deprecate any column chunk based estimate.
This PR proposes to expresses this as a ratio of the original row group. We could have an absolute setting instead but that would basically amount to the same in my opinion.
I would propose to either keep the current minimum row group byte size ratio (proposed in this PR) or replace it with an absolute minimum row group byte size.
I'm +1 on this PR unless someone else disagrees. What do you think @rdblue @dongche @danielcweeks ?

isnotinvain:
I discussed this some more with @rdblue, we came up with applying a minimum size check on the entire rowgroup size -- we need to support the (probably not uncommon) case where a rowgroup has only 1 column populated. What we really care about is that when reading the big (or biggest) columns, the rowgroup is sufficiently large.

rdblue:
I agree with the changes, considering some columns can be as small as a few bytes. If most columns are that sparse, then making the minimum size scale with the number of columns no longer makes sense.

At the same time, X% of 128MB and X% of 8MB are vastly different, so I'm not sure scale factor makes a lot of sense either.

@isnotinvain and I were talking about what *would* make sense for a minimum and I think it should be based on reasonable use. What is the point of a columnar format if none of your columns can actually save I/O? So then we assume that you want some limited overhead, like < 2%, and have an OS imposed minimum read size, 4k. At least one column, but be a little more generous and say 2 column chunks = 50 * 4k * 2 = 400k minimum. That's on the order of the current page size so let's say at least one full page = 1MB.

parquet-hadoop/src/main/java/parquet/hadoop/MemoryManager.java
```diff
@@ -121,10 +121,10 @@ private void updateAllocation() {
 
     for (Map.Entry<InternalParquetRecordWriter, Long> entry : writerList.entrySet()) {
       long newSize = (long) Math.floor(entry.getValue() * scale);
-      if(minMemoryAllocation > 0 && newSize/maxColCount < minMemoryAllocation) {
-          throw new ParquetRuntimeException(String.format("New Memory allocation %d"+
-          " exceeds minimum allocation size %d with largest schema having %d columns",
-              newSize, minMemoryAllocation, maxColCount)){};
+      if(scale < 1.0 && minMemoryAllocation > 0 && newSize < minMemoryAllocation) {
+          throw new ParquetRuntimeException(String.format("New Memory allocation %d" +
```
rdblue: Nit: we should include units in the error message.

rdblue:
+1

Any comments @danielcweeks or @dongche?

dongche:
+1. The change looks good! Thanks @isnotinvain

PR #144, opened 2015-03-10
PARQUET-212: Add thrift support for list reading rules

rdblue:
This updates parquet-thrift so that it implements the list reading rules specified by PARQUET-113. Null values are not allowed in Thrift, and the protocol cannot be amended to include them. Optional list elements are handled in this PR by throwing an exception (default) or by ignoring null values if `parquet.thrift.ignore-null-elements` is set to true in the `Configuration`.

### DIFF
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -81,6 +81,13 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-hadoop</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
@@ -20,7 +20,9 @@ package org.apache.parquet.avro;
 
 import java.lang.reflect.Constructor;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericArray;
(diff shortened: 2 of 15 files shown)


rdblue:
Rebased and fixed all test failures. I think this is ready for review. Would you mind taking a look @julienledem or @isnotinvain? Thanks!

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -82,6 +85,37 @@ private void checkUnmatchedProjectionFilter(FieldProjectionFilter filter) {
...
+   * checking whether a potential synthetic group matches the expected
+   * ThriftField.
+   * <p>
+   * This method never guesses because the expected ThriftField is known.
+   *
+   * @param repeatedType a type that may be the element type
+   * @param thriftElement the expected Schema for list elements
+   * @return {@code true} if the repeatedType is the element schema
+   */
+  static boolean isElementType(Type repeatedType, ThriftField thriftElement) {
```
tsdeng: rename to isListElementType ?

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -82,6 +85,37 @@ private void checkUnmatchedProjectionFilter(FieldProjectionFilter filter) {
...
+    if (repeatedType.isPrimitive() ||
```
tsdeng: for synthetic field, should we also check if the name equals to "element" ?

rdblue: There's no need for this check because we know the expected type, `thriftElement`. This method determines whether the thrift structure matches the repeated type or if the repeated type must contain an element that matches the thrift structure. Matching by the structure itself avoids making decisions based on field names.

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
rdblue: Done.

rdblue:
@tsdeng, could you have another look at this? I think it is ready to go.

PR #145, opened 2015-03-10
PARQUET-172: Add parquet-thrift binary tests.

rdblue:
These tests validate that there is no encoding problem with parquet-thrift or parquet-scrooge. See https://github.com/laurencer/parquet-mr-bug


isnotinvain:
+1 LGTM

rdblue:
Will

PR #146, opened 2015-03-11
PARQUET-215 Discard records with unrecognized union members in the thrift write path

isnotinvain:
Fixes Parquet-215, adds a test case for it, and fixes some tests that were quietly not doing anything previously to actually exercise the code they were intended to exercise. (they were tests that catch exceptions and make assertions about them, but never enforced that the exception was actually thrown, and in one case, it never was).


parquet-thrift/src/test/resources/parquet/hadoop/thrift/AddressBook.json
```diff
@@ -1,5 +1,6 @@
 {
   "type" : "STRUCT",
+  "isUnion": false,
```
isnotinvain: actually, now that I think about it -- I need to support old metadata files that don't have this field set.

pom.xml
```diff
@@ -235,6 +235,7 @@
                      <exclude>parquet/org/**</exclude>
                      <exclude>parquet/column/**</exclude>
                      <exclude>parquet/hadoop/ParquetInputSplit</exclude>
+                     <exclude>parquet/thrift/struct/**</exclude>
```
isnotinvain: @julienledem thoughts on this? It's easy to keep the old constructor and mark it deprecated, but it really shouldn't be used, there's no reasonable default for isUnion other than figuring out if it really is a union.

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
```diff
@@ -358,11 +361,28 @@ public String toDebugString() {
...
+            // this is an unrecognized field in a struct, not a union
+            notifyIgnoredFieldsOfRecord(field);
+            hasFieldsIgnored |= true;
+            //read the value and ignore it, NullProtocol will do nothing
+            new ProtocolReadToWrite().readOneValue(in, new NullProtocol(), field.type);
+            continue;
+          case YES:
+            // this is a union with an unrecognized member -- this is fatal for this record
+            // in the write path, because it will be unreadable in the read path.
+            // throwing here means we will skip this record entirely.
```
julienledem: or fail

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -93,7 +94,7 @@ private StructType toStructType(TStructDescriptor struct) {
                         Requirement.fromType(field.getFieldMetaData().requirementType);
         children.add(toThriftField(field.getName(), field, req));
       }
-      return new StructType(children);
+      return new StructType(children, TBaseUtil.isUnion(struct.getThriftClass()));
```
julienledem: I don't see why we need a TBaseUtil at all and don't declare "isUnion" private in this class.
I usually try to avoid "Util" classes. It could always be moved if some other class needs it.

parquet-thrift/src/main/java/parquet/thrift/struct/ThriftType.java
```diff
@@ -206,6 +237,32 @@ public ThriftField getChildById(short id) {
...
+      switch (isUnion) {
+        case YES:
+          return true;
+        case NO:
+          return false;
+        case UNKNOWN:
+          return null;
+        default:
+          throw new ParquetRuntimeException("This should never happen"){};
+      }
```
julienledem: possibly move this logic to the IsUnion enum.
public enum IsUnion {
  abstract Boolean toBool();
      YES { Boolean toBool() { return true} },
     NO { Boolean toBool() { return false} },
     UNKNOWN { Boolean toBool() { return null} },
+   }

parquet-thrift/src/main/java/parquet/thrift/struct/ThriftType.java
```diff
@@ -175,9 +180,35 @@ final public void visit(StringType stringType) {
...
+   * Whether a struct is a union is not always known, because it was not always
+   * written to the metadata files.
+   *
+   * We should always know this in the write path, but may not in the read path.
+   */
+   public enum IsUnion {
+      YES,
+      NO,
+      UNKNOWN
+   }
```
julienledem: another possible naming for this.
enum StructType { UNION, STRUCT, UNKNOWN }
I don't feel very strongly about it.

rdblue: I agree with this. It seems odd to have the boolean naming convention, `is_`, used for a non-boolean. Julien's suggestion would make this cleaner.

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
isnotinvain: You mean it depends on the configuration? Some modes fail on the first bad record?

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
isnotinvain: Yeah, I only made this thinking the detection was going to be more involved. I'll just move it here.

parquet-thrift/src/main/java/parquet/thrift/struct/ThriftType.java
isnotinvain: yeah that works thanks

parquet-thrift/src/main/java/parquet/thrift/struct/ThriftType.java
isnotinvain: Yeah, I also didn't like having "Yes" an "No" in the mix, UNION, STRUCT, UNKNOWN sounds good to me. I will update the JSON serialization as well to use the format:

`"structType" : "UNION"`

instead of:
`"isUnion": true`

isnotinvain: StructType is already used (this is a member of class StructType), having trouble coming up with a reasonable name... `StructKind`? seems like a different way of saying StructType and doesn't make it clear what it's for.  `StructSubType`? I went with `IsUnion` because it's there only to explain whether it's a union or not. `StructOrUnion`? I'd like Union to be in the name so that readers see that it's basically here to describe "is this struct a union?"

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
julienledem: I mean the comment could say: "throwing here means we will skip this record entirely *or fail*" as the decision to skip or fail the entire file happens higher up the stack.

parquet-thrift/src/main/java/parquet/thrift/struct/ThriftType.java
julienledem: maybe just type? Naming is hard, I'll just defer to you.

pom.xml
julienledem: In this instance the most important thing is to keep compatibility regarding deserializing old JSON that did not have this field. It sounds like the old  constructor is not necessary. I don't feel strongly about this. But the unknown case will happen anyway so maybe keep the old constructor deprecated? Do we need to remove the annotation anyway?

parquet-thrift/src/test/resources/parquet/hadoop/thrift/AddressBook.json
julienledem: yep

julienledem:
I made a few comments above.
Otherwise, this looks good to me.
+1

pom.xml
isnotinvain: I don't need this exclusion anymore, I kept the deprecated constructor, which is never used internally but if someone uses it externally it'll still work.

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
isnotinvain: got it

isnotinvain:
Tests passed here: https://github.com/isnotinvain/incubator-parquet-mr/pull/5
Going to merge this now.

PR #147, opened 2015-03-12
Log projected schema changes

MansurAshraf:
Logs discarded and projected fields after projection pushdown. Fixes https://issues.apache.org/jira/browse/PARQUET-213

### DIFF
--- a/parquet-thrift/pom.xml
+++ b/parquet-thrift/pom.xml
@@ -115,6 +115,25 @@
       <version>3.4</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.scala-lang</groupId>
+          <artifactId>scala-library</artifactId>
+          <version>${scala.version}</version>
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>org.scalacheck</groupId>
+          <artifactId>scalacheck_2.10</artifactId>
+          <version>1.12.1</version>
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>org.scalatest</groupId>
+          <artifactId>scalatest_2.10</artifactId>
+          <version>2.2.4</version>
+          <scope>test</scope>
+      </dependency>
+
 
   </dependencies>
 
@@ -150,5 +169,5 @@
         <artifactId>maven-shade-plugin</artifactId>
       </plugin>
     </plugins>
-  </build> 
+  </build>
 </project>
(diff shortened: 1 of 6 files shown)


parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -146,11 +153,23 @@ public void visit(ThriftType.ListType listType) {
...
-    if (currentType == null)
-      return new MessageType(currentName, new ArrayList<Type>());
+      final MessageType result;
+      if (currentType == null) {
+          result = new MessageType(currentName, new ArrayList<Type>());
+      } else {
+          GroupType rootType = currentType.asGroupType();
+          result = new MessageType(currentName, rootType.getFields());
+      }
+      logger.info("projected fields");
```
isnotinvain: maybe we can make a static method that builds the string to log, and then we can log it as a single call to logger.info.

That way we can write a unit test that tests the static method w/o having to test a mock logger or anything funky like that.

MansurAshraf:
Hive module failing for some reason
 [INFO] Parquet Hive Storage Handler ....................... FAILURE [  0.824 s]

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -56,221 +58,232 @@
  */
 public class ThriftSchemaConvertVisitor implements ThriftType.TypeVisitor {
```
julienledem: indentation is two spaces. This diff is much bigger than it should because of the re-indentation.

parquet-thrift/pom.xml
```diff
@@ -115,6 +115,25 @@
...
+          <artifactId>scalacheck_2.10</artifactId>
+          <version>1.12.1</version>
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>org.scalatest</groupId>
+          <artifactId>scalatest_2.10</artifactId>
+          <version>2.2.4</version>
+          <scope>test</scope>
+      </dependency>
```
julienledem: are those needed?

julienledem:
Thanks @MansurAshraf 
Could you fix the indentation to 2 spaces?
It is hard to review this way.

isnotinvain:
Just a heads up that this will probably conflict with https://github.com/apache/incubator-parquet-mr/pull/150

PR #148, opened 2015-03-20
PARQUET-223: Add builders for MAP and LIST types

SinghAsDev:
As of now, Parquet does not provide builders for Maps and Lists. This leaves margin for user errors. Having Map and List builders will make it easier for users to build these types.


parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -247,7 +308,8 @@ public P named(String name) {
    * @param <P> The type that this builder will return from
```
rdblue: Should add docs for <Q>

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +628,358 @@ protected GroupType build(String name) {
...
+
+    public P named(String name) {
+      Preconditions.checkState(originalType == null,
+          "LIST is already the logical type and can't be changed");
+      return super.named(name);
+    }
+
+    public GroupBuilder<ListBuilder<P>> groupElement(Type.Repetition repetition) {
+      Preconditions
+          .checkState(elementType == null, "groupElement can be called just once on a ListBuilder");
```
rdblue: It looks like the element builders don't behave like the key and value buliders, but I think they should. I think the `named` method for the `ListBuilder` should throw an exception because the element type must be defined. Then the `named` method for the element builder should build the element and return the grandparent builder like map keys and values do. Then you wouldn't need to name the element level.

Does that make sense?

parquet-column/src/test/java/parquet/schema/TestTypeBuilders.java
```diff
@@ -599,6 +601,479 @@ public Type call() throws Exception {
...
+        typeList));
+    GroupType actual = Types.optionalMap()
+        .key(INT64)
+        .requiredValue(INT64)
+        .named("myMap");
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void testMapWithInt64KeyAnnotations() {
```
rdblue: This doesn't appear to have any annotations on the INT64 key field. It looks like this might be misnamed because it looks like it's testing the group builder's `requiredMap` method.

parquet-column/src/test/java/parquet/schema/TestTypeBuilders.java
```diff
@@ -599,6 +601,479 @@ public Type call() throws Exception {
...
+    valueFields.add(new PrimitiveType(OPTIONAL, INT32, "two"));
+    typeList.add(new GroupType(OPTIONAL, "value", valueFields));
+
+    GroupType map = new GroupType(REQUIRED, "myMap", OriginalType.MAP, new GroupType(REPEATED, "map",
+        typeList));
+
+    MessageType expected = new MessageType("mapParent", map);
+    GroupType actual =
+        Types.buildMessage().requiredMap()
+            .groupKey()
```
rdblue: Nit: I find that indenting helps readability for long schemas.

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,528 @@ protected GroupType build(String name) {
...
+    }
+
+    @Override
+    protected abstract Q self();
+
+    public MapKeyBuilder<P, Q> key(PrimitiveTypeName type) {
+      if (parent != null) {
+        return new MapKeyBuilder<P, Q>(this, type);
+      } else {
+        return new MapKeyBuilder<P, Q>(this, returnType, type);
```
rdblue: I don't think this is needed since the Map builder's parent is no longer explicitly used. The other builders don't have it either.

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,528 @@ protected GroupType build(String name) {
...
+      }
+    }
+
+    public MapGroupKeyBuilder<P, Q> groupKey() {
```
rdblue: Unfortunately, I think this needs a listKey() method as well... I forgot about using lists as keys. Not mapKey though, since I don't think we need to allow maps to use maps as keys.

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,528 @@ protected GroupType build(String name) {
...
+      public MapGroupValueBuilder<P, Q> groupValue(Type.Repetition repetition) {
+        parent.setKeyType(build("key"));
+        return new MapGroupValueBuilder<P, Q>(parent).repetition(repetition);
+      }
+
+      public MapGroupValueBuilder<P, Q> requiredGroupValue() {
+        return groupValue(Type.Repetition.REQUIRED);
+      }
+
+      public MapGroupValueBuilder<P, Q> optionalGroupValue() {
```
rdblue: I think this should allow both maps and lists to be used as the value, so you'll need requiredListValue, requiredMapValue, optionalListValue, and optionalMapValue.

parquet-column/src/main/java/parquet/schema/Types.java
rdblue: Sorry, should add docs for all type parameters.

parquet-column/src/main/java/parquet/schema/Types.java
SinghAsDev: value() inside MapKeyBuilder uses MapBuilder's parent. So, it is required
:). I should have mentioned in the first reply itself.

On Mon, Mar 23, 2015 at 5:16 PM, Ryan Blue <notifications@github.com> wrote:

> In parquet-column/src/main/java/parquet/schema/Types.java
> <https://github.com/apache/incubator-parquet-mr/pull/148#discussion_r26994189>
> :
>
> > +      super(parent);
> > +    }
> > +
> > +    private BaseMapBuilder(Class<P> returnType) {
> > +      super(returnType);
> > +      this.returnType = returnType;
> > +    }
> > +
> > +    @Override
> > +    protected abstract Q self();
> > +
> > +    public MapKeyBuilder<P, Q> key(PrimitiveTypeName type) {
> > +      if (parent != null) {
> > +        return new MapKeyBuilder<P, Q>(this, type);
> > +      } else {
> > +        return new MapKeyBuilder<P, Q>(this, returnType, type);
>
> I don't think this is needed since the Map builder's parent is no longer
> explicitly used. The other builders don't have it either.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/incubator-parquet-mr/pull/148/files#r26994189>.
>



-- 

Regards,
Ashish

parquet-column/src/main/java/parquet/schema/Types.java
SinghAsDev: Adding listKey() is tricky, as we do not want to call named on key and we want list's element to construct list and return list's parent on calling named on list's element. I can think of following solutions, none of which is really clean.
1. Re-Implement all variants of ElementBuilder in MapListKeyBuilder and add all methods that build value to all those variants. Lots of code duplication.
2. Add value builders to existing ListElementBuilders with an assert to make sure its parent is a keyBuilder.

SinghAsDev: @rdblue I have pushed changes that should address most of your reviews, except this one. I think it will be better if we discuss this before I go ahead and implement something. Let me know what you think.

rdblue:
Looks like the test failures are caused by incompatible changes flagged by semver. I don't think they are actually incompatible because you moved the methods to a superclass, but we'll need to add a temporary exception here: https://github.com/apache/parquet-mr/blob/master/pom.xml#L240

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,1024 @@ protected GroupType build(String name) {
...
+
+    private BaseMapBuilder(Class<P> returnType) {
+      super(returnType);
+      this.returnType = returnType;
+    }
+
+    @Override
+    protected abstract Q self();
+
+    public MapKeyBuilder<P, Q> key(PrimitiveTypeName type) {
```
rdblue: I thought we had decided not to add a key builder for maps? I thought that was to be handled by `key(Type)` and `requiredKey(Type)` methods.

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,1024 @@ protected GroupType build(String name) {
...
+    }
+
+    public MapGroupKeyBuilder<P, Q> groupKey() {
+      return new MapGroupKeyBuilder<P, Q>(this);
+    }
+
+    public MapListKeyBuilder<P, Q> listKey() {
+      return new MapListKeyBuilder<P, Q>(this);
+    }
+
```
rdblue: Okay, I think we still need a couple of methods that were overlooked: `key(Type)` so you can make more complex types than are supported by the builders. Then we will also need `value(Type)`, `requiredValue(Type)`, and `optionalValue(Type)` on the key builders that set the value type.

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,1024 @@ protected GroupType build(String name) {
...
+    public ListListElementBuilder<P, Q> requiredListElement() {
+      return listElement(Type.Repetition.REQUIRED);
+    }
+
+    public ListListElementBuilder<P, Q> optionalListElement() {
+      return listElement(Type.Repetition.OPTIONAL);
+    }
+
+    public ListMapElementBuilder<P, Q> mapElement(Type.Repetition repetition) {
```
rdblue: Since we allow lists as keys, we can get into a bad situation where a key is a list of maps. I think lists of maps are a rare thing anyway, so let's just say we don't have to worry about the case where a list contains a map and remove all of that stuff. Does that sound okay? Again, we can handle that case by using `requiredElement(Type)` or `optionalElement(Type)`

parquet-column/src/main/java/parquet/schema/Types.java
```diff
@@ -555,6 +622,1024 @@ protected GroupType build(String name) {
...
+
+    public ListMapElementBuilder<P, Q> requiredMapElement() {
+      return mapElement(Type.Repetition.REQUIRED);
+    }
+
+    public ListMapElementBuilder<P, Q> optionalMapElement() {
+      return mapElement(Type.Repetition.OPTIONAL);
+    }
+
+    public ListKeyElementBuilder<P, Q> keyElement(PrimitiveTypeName type,
```
rdblue: What is this class for? Do we need a ListKeyElement, and if so, do we need it here?

rdblue:
Super close. I think we just need to add a couple of methods and remove some stuff. Still need to add the methods that take `Type` objects as I noted in comments. I also think we should remove the nested maps within lists.

I didn't quite understand the key builders within List either, and what made them distinct from the mapElement builder. I don't think they are needed either.

One last thing: can you rebase on master? Thanks!

parquet-column/src/main/java/parquet/schema/Types.java
SinghAsDev: I think we can allow list of maps. However, the problem is that we do not want to handle lists of maps as keys of a map, this we can catch and advise user to use key(Type) method on MapBuilder instead.

I will make this change and if that does not look good to you then we can get rid of it.

parquet-column/src/main/java/parquet/schema/Types.java
SinghAsDev: True. Adding them.

parquet-column/src/main/java/parquet/schema/Types.java
SinghAsDev: This is to support Maps with List as Keys. I think we need it as we need an override of named that builds the value of map.

Having this here avoids bunch of code duplication.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+
+    public MapListValueBuilder<P, Q> requiredListValue() {
+      return listValue(Type.Repetition.REQUIRED);
+    }
+
+    public MapListValueBuilder<P, Q> optionalListValue() {
+      return listValue(Type.Repetition.OPTIONAL);
+    }
+
+    public Q value (Type type) {
```
rdblue: Nit: style is off for these methods: shouldn't have a space before the open paren.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+
+    public ListMapElementBuilder<P, Q> requiredMapElement() {
+      return mapElement(Type.Repetition.REQUIRED);
+    }
+
+    public ListMapElementBuilder<P, Q> optionalMapElement() {
+      return mapElement(Type.Repetition.OPTIONAL);
+    }
+
+    public ListKeyElementBuilder<P, Q> keyElement(PrimitiveTypeName type,
```
rdblue: Let's remove this method, `keyElement`, and the `keyGroupElement` method. I don't think they should be exposed on the list builder.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+
+      public MapListValueBuilder<P, Q> requiredListValue() {
+        return listValue(Type.Repetition.REQUIRED);
+      }
+
+      public MapListValueBuilder<P, Q> optionalListValue() {
+        return listValue(Type.Repetition.OPTIONAL);
+      }
+
+      public BaseMapBuilder<P, Q> value (Type type) {
```
rdblue: Nit: spacing issue here too.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+        return parent.named(name);
+      }
+
+      @Override
+      protected MapListValueBuilder<P, Q> self() {
+        return this;
+      }
+    }
+
+    public static class MapListKeyBuilder<P, Q extends BaseMapBuilder<P, Q>> extends
```
rdblue: If this requires exposing the `keyElement` methods on the list builder to avoid problems with `named`, then lets not support lists as map keys in the API. We can add it back later, but the implementation it quite complicated and I think it is cleaner to allow callers to simply set a list type instead of having so much code. I think this should allow you to remove all of the key/element code and this will be much shorter.

We can always add this back in later.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+        return this;
+      }
+
+      public void setKey(Type key) {
+        parent.setKeyType(key);
+      }
+    }
+
+    protected void setKeyType(Type keyType) {
+      this.keyType = keyType;
```
rdblue: Should this verify that `keyType` is null? Similar logic in `setValueType`?

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,114 +622,1267 @@ protected GroupType build(String name) {
...
+
+    public ListGroupElementBuilder<P, Q> requiredGroupElement() {
+      return groupElement(Type.Repetition.REQUIRED);
+    }
+
+    public ListGroupElementBuilder<P, Q> optionalGroupElement() {
+      return groupElement(Type.Repetition.OPTIONAL);
+    }
+
+    public ListElementBuilder<P, Q> element(PrimitiveTypeName type, Type.Repetition repetition) {
```
rdblue: You still need to add the versions that take types: `element(Type)`, `requiredElement(Type)`, and `optionalElement(Type)`

rdblue:
Okay, I've done one more look through. There are two things still holding it up:

1. `element(Type)` methods are missing
2. I think we should remove the `keyElement` code and not provide convenience methods for lists inside map keys. This ended up being too messy and exposed extra methods on the list builder.

I also noted a few other things that I haven't seen before. Not blockers, but we may as well fix them:

* All of the inner builder classes have names like MapGroupKeyBuilder. That's okay, but there's no need to prefix the name with Map because there is no other Key. So I think it should be just GroupKeyBuilder. That also fixes the problem of doubled names, like MapMapValueBuilder or ListListElementBuilder.
* The key, value, and element method names are singular. I noticed this when trying out the element builder: `Types.requiredList().requiredElement(INT32).named("myList")`. I think the plural, `requiredElements` would be easier. It's up to you whether we change it or not. Let's just make sure it is consistent between key, value, and element.
* Class and method ordering could be more consistent for readability (very minor)

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
SinghAsDev: As these are not public methods, I did not want to put any contracts. However, it definitely won't hurt. Adding the checks.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
SinghAsDev: Shoot, missed it. Adding it now. Thanks for pointing this.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -555,6 +622,741 @@ protected GroupType build(String name) {
...
+
+    public ListElementBuilder<P, Q> requiredListElement() {
+      return listElement(Type.Repetition.REQUIRED);
+    }
+
+    public ListElementBuilder<P, Q> optionalListElement() {
+      return listElement(Type.Repetition.OPTIONAL);
+    }
+
+    public BaseListBuilder<P, Q> element(Type type, Type.Repetition repetition) {
```
rdblue: It just occurred to me that repetition is ignored because the Type includes a repetition. Maybe we don't need `requiredElement` and `optionalElement` after all. Same with the key and value methods that take a `Type`.

SinghAsDev: Done. Thanks again for catching this.

On Sun, May 17, 2015 at 2:07 PM, Ryan Blue <notifications@github.com> wrote:

> In parquet-column/src/main/java/org/apache/parquet/schema/Types.java
> <https://github.com/apache/parquet-mr/pull/148#discussion_r30474900>:
>
> > +      if (parent != null) {
> > +        return new ListElementBuilder<P, Q>(this).repetition(repetition);
> > +      } else {
> > +        return new ListElementBuilder<P, Q>(this, returnType).repetition(repetition);
> > +      }
> > +    }
> > +
> > +    public ListElementBuilder<P, Q> requiredListElement() {
> > +      return listElement(Type.Repetition.REQUIRED);
> > +    }
> > +
> > +    public ListElementBuilder<P, Q> optionalListElement() {
> > +      return listElement(Type.Repetition.OPTIONAL);
> > +    }
> > +
> > +    public BaseListBuilder<P, Q> element(Type type, Type.Repetition repetition) {
>
> It just occurred to me that repetition is ignored because the Type
> includes a repetition. Maybe we don't need requiredElement and
> optionalElement after all. Same with the key and value methods that take
> a Type.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/parquet-mr/pull/148/files#r30474900>.
>



-- 

Regards,
Ashish

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -205,7 +207,7 @@ public void testFixedLengthEquals() {
     Type f4 = Types.required(FIXED_LEN_BYTE_ARRAY).length(4).named("f4");
     Type f8 = Types.required(FIXED_LEN_BYTE_ARRAY).length(8).named("f8");
     Assert.assertFalse("Types with different lengths should not be equal",
-        f4.equals(f8));
+            f4.equals(f8));
```
rdblue: There are a lot of non-functional changes to the white space that look non-standard.

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -599,6 +601,777 @@ public Type call() throws Exception {
         });
   }
 
+  @Test
+  public void testRequiredMap() {
+    List<Type> typeList = new ArrayList<Type>();
+    typeList.add(new PrimitiveType(REQUIRED, INT64, "key"));
+    typeList.add(new PrimitiveType(REQUIRED, INT64, "value"));
+    GroupType expected = new GroupType(REQUIRED, "myMap", OriginalType.MAP, new GroupType(REPEATED,
+        "map",
```
rdblue: According to spec, the repeated level should be named "key_value".

rdblue: Actually, we'll fix this separately.

PR #150, opened 2015-03-22
PARQUET-229 Add a strict thrift projection API with backwards compat support

isnotinvain:
Currently, the thrift projection API accepts strings in a very general glob format that supports not only wildcards like `*` and `?` and expansions (`{x,y,z}`) but also character classes `[abc]`, and negation.
Because of this flexibility, it's hard to give users good error reporting, for example letting them know that when they requested columns `foo.bar.{a,b,c}` there is actually no such column `foo.bar.c`.

This PR introduces a new syntax that supports a more restrictive form of glob syntax and enforces that all **expansions** of a glob match a column, not just that all globs match a column. The new syntax is very simple and only has four special characters: `{`,`}`,`,`, and `*`

It supports glob expansion, for example:
`home.{phone,address}` or `org.apache{-incubator,}.foo`

And the wildcard `*` which is treated the same way as java regex treats `(.*)`, for example:
`home.*` or `org.apache*.foo`

In the new syntax glob paths mean "keep all the child fields of the field matched by this glob", just like variable access would work in a programming language. For example: `x.y.z` means keep field `z` and all of its children (if any). So it's not necessary to do `x.y.z.*`. However, `x.y.z` would not keep field `x.y.zoo`. If that was desired, then `x.y.z*` could be used instead.

Setting `"parquet.thrift.column.filter"` will result in the same behavior that it does currently in master, though a deprecation warning will be logged. The classes that implement the current behavior have been marked as deprecated, and using this will log a warning.

Setting `"parquet.thrift.column.projection.globs"` will instead use this new syntax, and entry points in the various Builder's in the codebase is added as well.

This PR does a little bit of cleanup as well, moving some shared methods to a `Strings` class and simplifying some of the class hierarchy in `ThriftSchemaConverterVisitor`. There are a few `// TODO Why?` added as well that I wanted to ask about.


isnotinvain:
cc @tsdeng @julienledem

parquet-cascading/src/main/java/parquet/cascading/ParquetValueScheme.java
```diff
@@ -105,9 +119,17 @@ public ParquetValueScheme(Config<T> config) {
     this.config = config;
   }
 
+  // TODO: Why is there thrift references in this file?
```
isnotinvain: just wanted to call this question out

parquet-common/src/main/java/parquet/glob/GlobParser.java
```diff
@@ -0,0 +1,224 @@
...
+    int anchor = 0; // first un-parsed character position
+
+    for (int i = 0; i < pattern.length(); i++) {
+      char c = pattern.charAt(i);
+
+      switch (c) {
+        case ',':
+          if (unmatchedBraces == 0) {
+            // commas not allowed in the top level expression
+            // TODO: maybe turn this check off?
```
isnotinvain: what do you think?

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -49,25 +50,21 @@
...
-  }
-
-  FieldProjectionFilter fieldProjectionFilter;
-  Type currentType;
-  FieldsPath currentFieldPath = new FieldsPath();
-  Type.Repetition currentRepetition = Type.Repetition.REPEATED;//MessageType is repeated GroupType
-  String currentName = "ParquetSchema";
+  private Type currentType;
+  private Type.Repetition currentRepetition = Type.Repetition.REPEATED; // MessageType is repeated GroupType
+  private String currentName = "ParquetSchema"; // TODO: why?
```
isnotinvain: just want to call out this question

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -90,7 +87,7 @@ public void visit(ThriftType.MapType mapType) {
     //=========handle value
     currentFieldPath.push(mapValueField);
     currentName = "value";
-    currentRepetition = OPTIONAL;
+    currentRepetition = OPTIONAL; // TODO: why?
```
isnotinvain: Why are map value optional? If they get projected out, we replace them with null?

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -100,6 +97,7 @@ public void visit(ThriftType.MapType mapType) {
       return;
     }
 
+    // TODO: this can never happen
     if (keyType == null && valueType != null)
```
isnotinvain: Intellij points out that this if statement is unreachable -- is this a bug?

isnotinvain:
Tests running here: https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/55340816

parquet-cascading/src/main/java/parquet/cascading/ParquetValueScheme.java
julienledem: Filters are defined in terms of the thrift schema and T has to be either a Thrift of Scrooge generated class.
Possibly this should be refactored or renamed.

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
tsdeng: yes, the value of a map can be projected out

isnotinvain: can parts of the key be projected out? That wouldn't make any sense right?

isnotinvain: I should double check if we enforce that you can't project fields nested under a key

parquet-cascading/src/main/java/parquet/cascading/ParquetValueScheme.java
isnotinvain: ok

isnotinvain:
tests passed: https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/57598699

tsdeng:
could you also update this doc with the new API? https://github.com/apache/incubator-parquet-mr/blob/master/parquet_cascading.md
It could serve as a tutorial for users to use it

tsdeng:
minor comment, otherwise LGTM

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
isnotinvain: revisit this later

PR #151, opened 2015-03-23
[PARQUET-198] Initial commit of ParquetAvroScheme (WIP)

AndreSchumacher:
This is a first attempt of adding a ParquetAvroScheme to parquet-cascading that can be used to read and write Avro records from Parquet files (say, from within Scalding). The code draws heavily on the Thrift implementation of the respective classes.

Currently only code-generated Avro classes are supported.

This code is also partly based on the work done here: https://github.com/epishkin/scalding/tree/parquet_avro/scalding-parquet


dkincaid:
Any chance this will eventually make it into the codebase? I could really use it.

PR #152, opened 2015-03-25
PARQUET-220: do not log unnecessary warnings on initializing ParquetRecordReader

kostya-sh:
Refactored to replace TaskInputOutputContext with TaskAttemptContext.

ParquetRecordReader used to check that the passed context is instance of
TaskInputOutputContext however the functionality it uses doesn't rely on this
fact.


julienledem:
Thanks!
The build is failing because you changed public methods.
We could add this package (parquet/hadoop/util/counters) to the list of internal apis in:
https://github.com/apache/incubator-parquet-mr/blob/4ed0bdf1c73fd82d3080d15085675de96d5be0aa/pom.xml#L235

rdblue:
I added an exclusion for ContextUtil (definitely not public), added backward-compatible signatures for the other two methods, and merged this in #162 so it could be included in the 1.6.0 release. Thanks, @kostya-sh!

PR #153, opened 2015-03-25
PARQUET-227 Enforce that unions have only 1 set value, tolerate bad records in read path

isnotinvain:
See https://issues.apache.org/jira/browse/PARQUET-227


parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
```diff
@@ -0,0 +1,105 @@
...
+  private static final Log LOG = Log.getLog(CorruptRecordCounter.class);
+
+  private static final float DEFAULT_THRESHOLD =  0f;
+  private static final long DEFAULT_MIN_ERRORS = 0;
+
+  private long numRecords;
+  private long numErrors;
+
+  private final double errorThreshold; // max fraction of errors allowed
+  private final long minErrors; // throw error only after this many errors
```
tsdeng: Why have both errorThreshold and minErrors? Could you explain more about the difference and purpose of each?

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -80,6 +81,7 @@
   private long totalCountLoadedSoFar = 0;
 
   private Path file;
+  private CorruptRecordCounter corruptRecordCounter;
```
tsdeng: final ?

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -202,12 +205,29 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
 
     while (!recordFound) {
       // no more records left
-      if (current >= total) { return false; }
+      if (current >= total) { return false;}
```
tsdeng: why removing space?

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -202,12 +205,29 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
 
       try {
         checkRead();
-        currentValue = recordReader.read();
+
+        corruptRecordCounter.incRecordsSeen();
+        boolean corrupt = false;
+
+        try {
```
tsdeng: Nested try catch block and multiple if...then blocks in this while loop make it hard to read and track. Also the inErrorsSeen depends on the calling of incRecordSeen is error prone. I think your propose is to have an enhanced recordReader that can track the corrupted record. So an Adapter Pattern like CorruptedTrackingRecordReader maybe able to simplify this logic a bit.

parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
```diff
@@ -0,0 +1,105 @@
...
+
+  public void incRecordsSeen() {
+    numRecords++;
+  }
+
+  public void incErrorsSeen(CorruptRecordException cause) throws ParquetDecodingException {
+    numErrors++;
+    if (numErrors > numRecords) {
+      // incorrect use of this class
+      throw new ParquetDecodingException("Forgot to invoke incRecords()?");
```
tsdeng: Error message not clear(print the actual number). Fundamental problems is incErrorsSeen depends on incRecordsSeen. Could use adapter pattern on record reader, and encapsulate this dependency in the adapter

parquet-thrift/src/main/java/parquet/hadoop/thrift/ThriftReadSupport.java
```diff
@@ -73,6 +73,12 @@ public static void setRecordConverterClass(JobConf conf,
     conf.set(RECORD_CONVERTER_CLASS_KEY, klass.getName());
   }
 
+  public static void setRecordConverterClass(Configuration conf,
```
tsdeng: remove the duplication of setRecordConverterClass(JobConf conf) by calling this method there

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
```diff
@@ -394,10 +404,42 @@ public String toDebugString() {
...
+      boolean wasIgnored = readOneValue(in, field.type, buffer, expectedField.getType());
+      if (!wasIgnored) {
+        childFieldsPresent++;
+      }
+      hasFieldsIgnored |= wasIgnored;
       in.readFieldEnd();
       buffer.add(FIELD_END);
     }
+
+    // check that union had exactly 1 (no more no less) child fields.
```
tsdeng: This method is becoming increasingly complicated, can we move the code below to a private method?

parquet-column/src/main/java/parquet/io/api/RecordMaterializer.java
```diff
@@ -45,4 +48,22 @@ public void skipCurrentRecord() { }
    * @return the root converter for this tree
    */
   abstract public GroupConverter getRootConverter();
+
+  public static class CorruptRecordException extends ParquetDecodingException {
```
julienledem: please add javadoc here.

parquet-column/src/main/java/parquet/io/api/RecordMaterializer.java
```diff
@@ -33,6 +35,7 @@
 
   /**
    * @return the result of the conversion
+   * @throws CorruptRecordException to signal that a record is corrupt, but can be skipped
```
julienledem: Is "Corrupt" the correct term here? The data is not corrupted but no valid instance of T can be produced for the current record and this implementation of RecordMaterializer.

parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
```diff
@@ -0,0 +1,105 @@
...
+    LOG.warn(String.format("Error while reading an input record (%s out of %s so far and %s total): ",
+        numErrors, numRecords, totalNumRecords), cause);
+
+    if (numErrors > 0 && errorThreshold <= 0) { // no errors are tolerated
+      throw new ParquetDecodingException("Error while decoding records", cause);
+    }
+
+    double errRate = numErrors/(double)totalNumRecords;
+
+    if (numErrors >= minErrors && errRate > errorThreshold) {
```
julienledem: How to configure this kind of minErrors record count before applying the threshold is always a little tricky.
Metadata has the total number of records and it looks like totalNumRecords is the actual number of records from the metadata and not the number of records seen so far.
I would be in favor of removing minErrors and just use ```if (errRate > errorThreshold)```

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -202,12 +205,29 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+          corruptRecordCounter.incErrorsSeen(e);
+          corrupt = true;
+        }
+
         current ++;
+
+        if (corrupt) {
+          if (DEBUG) LOG.debug("skipping a corrupt record");
+          continue;
+        }
```
julienledem: I would agree that the control flow could be simplified rather than the try{} cacth {} setting a boolean that triggers a continue later.

parquet-thrift/src/main/java/parquet/hadoop/thrift/ThriftReadSupport.java
julienledem: + javadoc

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
```diff
@@ -381,6 +389,8 @@ public String toDebugString() {
             // read path.
             throw new ParquetRuntimeException("This should never happen! "
                 + "Don't know if this field is a union, was the deprecated constructor of StructType used?\n" + type){};
+          default:
+            throw new ParquetRuntimeException("Unrecognized StructOrUnionType: " + type.getStructOrUnionType()){};
```
julienledem: we should avoid those anonymous subclasses and just create a concrete one.
Something like ShouldNotHappenException?

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
```diff
@@ -394,10 +404,42 @@ public String toDebugString() {
...
+        break;
+      case UNKNOWN:
+        // we should never reach here in the write path -- this only happens if the
+        // deprecated constructor of StructType is used, which should only be used in the
+        // read path.
+        throw new ParquetRuntimeException("This should never happen! "
+            + "Don't know if this field is a union, was the deprecated constructor of StructType used?\n" + type){};
+      default:
+        throw new ParquetRuntimeException("Unrecognized StructOrUnionType: " + type.getStructOrUnionType()){};
+    }
```
julienledem: should this be:
```
if (type.isUnion() && childFieldsPresent != 1) {
  if (childFieldsPresent == 0) {
     throw new DecodingSchemaMismatchException("Cannot write a TUnion with no set value in :\n" + type);
  } else {
    throw new DecodingSchemaMismatchException("Cannot write a TUnion with more than 1 set value in :\n" + type);
  }
}
```
with type.isUnion() define as ```type.getStructOrUnionType() == UNION```

parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
isnotinvain: The idea is, you can set both a percent threshold, and also an absolute minimum. Now that you point this out -- this isn't really needed. 

Elephantbird does this because elephantbird doesn't know the total number of records, so it divides the number of errors by the current record count, which is why they need this minErrors. But we don't even need that, nor the incRecords() method at all, so I think we can simply this by just removing the record count, incRecords() and the minErrors all together

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
isnotinvain: typo, thanks. I had moved this around earlier and then put it back.

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
isnotinvain: yes, thanks

isnotinvain: oh, nevermind, I had it as final originally but unfortunately this class is setup by an initialize() method, not the constructor, so this can't be final because it has to get assigned in the initialize() method. An unfortunate consequence of hadoop's pattern of instantiating classes reflectively via their default constructor.

parquet-thrift/src/main/java/parquet/thrift/ThriftRecordConverter.java
```diff
@@ -843,7 +843,9 @@ public T getCurrentRecord() {
       rootEvents.clear();
       return thriftReader.readOneRecord(protocol);
     } catch (TException e) {
-      throw new ParquetDecodingException("Could not read thrift object from protocol", e);
+      protocol.clear();
+      rootEvents.clear();
```
julienledem: why do we have 2 of those?

parquet-hadoop/src/main/java/parquet/hadoop/InternalParquetRecordReader.java
isnotinvain: I agree this loop is complicated. Unfortunately, the checkRead() method does a lot of internal state mutation, so I'm hesitant to add another layer around this class if that layer has to interact with the state that checkRead() does.

But I agree, this whole loop / class could use some cleanup. Actually some of the checks (like supporting 2 kinds of filters) could potentially go away once we are happy w/ the filter2 API.

parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
isnotinvain: removed numRecords entirely

parquet-thrift/src/main/java/parquet/hadoop/thrift/ThriftReadSupport.java
isnotinvain: yeah, Ideally we'd remove the JobConf one because it is less general than the Configuration one.
I'll mark it as deprecated. (a JobConf is a Configuration, but a Configuration is not a JobConf).

isnotinvain: By the way, this tripped me up -- should ScroogeInputFormat call this method automatically? It seems odd to use ScroogeInputFormat and then also have to call this method?

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
isnotinvain: yeah good call

parquet-column/src/main/java/parquet/io/api/RecordMaterializer.java
isnotinvain: done

parquet-column/src/main/java/parquet/io/api/RecordMaterializer.java
isnotinvain: Good point, how about `RecordMaterializationException`

parquet-hadoop/src/main/java/parquet/hadoop/CorruptRecordCounter.java
isnotinvain: yeah agreed

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
isnotinvain: we can do that if we don't worry about being exhaustive and throwing when unrecognized enum members are encountered. I didn't make this class follow the visitor pattern, so we can choose to do this, but if we ever add another type to this enum we may forget to update this line.

parquet-thrift/src/main/java/parquet/thrift/ThriftRecordConverter.java
isnotinvain: I wanted to ask you that -- also is `skipCurrentRecord` doing the right thing or should it clear both as well?

I confirmed that you need to clear protocol for this to work. I think it was formerly never a problem because this exception was fatal so it didn't matter.

parquet-thrift/src/main/java/parquet/thrift/BufferedProtocolReadToWrite.java
isnotinvain: done

isnotinvain:
@tsdeng, @julienledem this is the one we want to get in for 1.6.0

parquet-thrift/src/main/java/parquet/thrift/ThriftRecordConverter.java
isnotinvain: @julienledem do you remember much about this? I can dig into it if not.

tsdeng:
LGTM +1, shipIt once build passes. Thanks @isnotinvain

parquet-thrift/src/main/java/parquet/thrift/ThriftRecordConverter.java
isnotinvain: We can probably refactor this, but the reason is rootEvents are events that we plan to replay through the protocol. In the case of skipping a record, clearing this is enough. In the case of an exception, we have already pushed rootEvents into protocol, which has now only partially consumed those events. So both need to be cleared. I think this is correct, and we can refactor separately.

isnotinvain:
tests passed: https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/57598033

julienledem:
+1

PR #154, opened 2015-03-26
PARQUET-225: INT64 support for Delta Encoding

lunchev:
JIRA: https://issues.apache.org/jira/browse/PARQUET-225

As of now, parquet doesn't support delta encoding for INT64. However it is planned in the format:
https://github.com/Parquet/parquet-format/blob/master/Encodings.md

The benefits of this feature are huge. For timestamps it achieves twice better compression than SNAPPY on plain encoding, and the reading is faster. This feature is actually advertised on the home page of Parquet, even though it is not yet implemented:
http://parquet.incubator.apache.org/
http://image.slidesharecdn.com/hadoopsummit-140630160016-phpapp01/95/efficient-data-storage-for-analytics-with-apache-parquet-20-30-638.jpg?cb=1404162126

This pull request adds full support for INT64. It keeps INT32 binary unchanged, even though it reads it with the INT64 implementation. That might be slower on 32 bit processors, but I don't not consider it a problem.

Regarding testing, all test made for INT32 are ran against the new INT64 implementation and also 2 new unit tests are added specifically for INT64.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/delta/DeltaBinaryPackingValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/delta/DeltaBinaryPackingValuesWriter.java
@@ -24,7 +24,7 @@ import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.bytes.CapacityByteArrayOutputStream;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.values.ValuesWriter;
-import org.apache.parquet.column.values.bitpacking.BytePacker;
+import org.apache.parquet.column.values.bitpacking.BytePackerForLong;
 import org.apache.parquet.column.values.bitpacking.Packer;
 import org.apache.parquet.io.ParquetEncodingException;
 
@@ -52,10 +52,15 @@ import java.io.IOException;
  */
 public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
   /**
+   * Is the current column of type int (INT32)
+   */
+  private boolean isInt;
+
+  /**
    * max bitwidth for a mini block, it is used to allocate miniBlockByteBuffer which is
    * reused between flushes.
    */
-  public static final int MAX_BITWIDTH = 32;
+  public static final int MAX_BITWIDTH = 64;
 
   public static final int DEFAULT_NUM_BLOCK_VALUES = 128;
 
@@ -86,7 +91,7 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
    * stores delta values starting from the 2nd value written(1st value is stored in header).
    * It's reused between flushes
    */
-  private int[] deltaBlockBuffer;
+  private long[] deltaBlockBuffer;
 
   /**
    * bytes buffer for a mini block, it is reused for each mini block.
@@ -97,12 +102,12 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
   /**
    * firstValue is written to the header of the page
    */
-  private int firstValue = 0;
+  private long firstValue = 0;
 
   /**
    * cache previous written value for calculating delta
    */
-  private int previousValue = 0;
+  private long previousValue = 0;
 
   /**
    * min delta is written to the beginning of each block.
@@ -110,7 +115,7 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
    * therefore are all positive
    * it will be reset after each flush
    */
-  private int minDeltaInCurrentBlock = Integer.MAX_VALUE;
+  private long minDeltaInCurrentBlock = Long.MAX_VALUE;
 
   public DeltaBinaryPackingValuesWriter(int slabSize, int pageSize, ByteBufferAllocator allocator) {
     this(DEFAULT_NUM_BLOCK_VALUES, DEFAULT_NUM_MINIBLOCKS, slabSize, pageSize, allocator);
@@ -119,7 +124,7 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
   public DeltaBinaryPackingValuesWriter(int blockSizeInValues, int miniBlockNum, int slabSize, int pageSize, ByteBufferAllocator allocator) {
     this.config = new DeltaBinaryPackingConfig(blockSizeInValues, miniBlockNum);
     bitWidths = new int[config.miniBlockNumInABlock];
-    deltaBlockBuffer = new int[blockSizeInValues];
+    deltaBlockBuffer = new long[config.blockSizeInValues];
     miniBlockByteBuffer = new byte[config.miniBlockSizeInValues * MAX_BITWIDTH];
     baos = new CapacityByteArrayOutputStream(slabSize, pageSize, allocator);
   }
@@ -131,6 +136,21 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
 
   @Override
   public void writeInteger(int v) {
+    /*
+     * isInt is needed. Just calling internalWrite might result in a different binary representation.
+     * Example: [Integer.MIN_VALUE, Integer.MAX_VALUE, Integer.MIN_VALUE, Integer.MAX_VALUE]
+     */
+    isInt = true;
+    internalWrite(v);
+  }
+
+  @Override
+  public void writeLong(long v) {
+    isInt = false;
+    internalWrite(v);
+  }
+
+  public void internalWrite(long v) {
     totalValueCount++;
 
     if (totalValueCount == 1) {
@@ -139,7 +159,10 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
       return;
     }
 
-    int delta = v - previousValue;//calculate delta
+    // Calculate delta. The possible overflow is accounted for. The algorithm is correct because
+    // Java long is working as a mudalar ring with base 2^64 and because of the plus and minus
+    // properties of a ring. http://en.wikipedia.org/wiki/Modular_arithmetic#Integers_modulo_n
+    long delta = v - previousValue;
     previousValue = v;
 
     deltaBlockBuffer[deltaValuesToFlush++] = delta;
@@ -159,6 +182,13 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
       deltaBlockBuffer[i] = deltaBlockBuffer[i] - minDeltaInCurrentBlock;
     }
 
+    if (isInt) {
+      minDeltaInCurrentBlock = (int) minDeltaInCurrentBlock;
+      for (int i = 0; i < deltaValuesToFlush; i++) {
+        deltaBlockBuffer[i] &= (1l << 32) - 1l;
+      }
+    }
+
     writeMinDelta();
     int miniBlocksToFlush = getMiniBlockCountToFlush(deltaValuesToFlush);
 
@@ -170,7 +200,7 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
     for (int i = 0; i < miniBlocksToFlush; i++) {
       //writing i th miniblock
       int currentBitWidth = bitWidths[i];
-      BytePacker packer = Packer.LITTLE_ENDIAN.newBytePacker(currentBitWidth);
+      BytePackerForLong packer = Packer.LITTLE_ENDIAN.newBytePackerForLong(currentBitWidth);
       int miniBlockStart = i * config.miniBlockSizeInValues;
       for (int j = miniBlockStart; j < (i + 1) * config.miniBlockSizeInValues; j += 8) {//8 values per pack
         // mini block is atomic in terms of flushing
@@ -182,7 +212,7 @@ public class DeltaBinaryPackingValuesWriter extends ValuesWriter {
       }
     }
 
-    minDeltaInCurrentBlock = Integer.MAX_VALUE;
+    minDeltaInCurrentBlock = Long.MAX_VALUE;
     deltaValuesToFlush = 0;
   }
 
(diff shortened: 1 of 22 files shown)


julienledem:
@nongli any comment?

parquet-column/src/main/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterForInteger.java
```diff
@@ -0,0 +1,232 @@
...
+  public void writeInteger(int v, final CapacityByteArrayOutputStream baos) {
+    totalValueCount++;
+
+    if (totalValueCount == 1) {
+      firstValue = v;
+      previousValue = firstValue;
+      return;
+    }
+
+    int delta = v - previousValue;//calculate delta
```
nongli: I'm not sure if the previous version had this issue to but doesn't this overflow?
-INT_MAX - INT_MAX

lunchev: That's a direct copy-paste from the previous version. I just duplicated the code and changed some INTs to LONGs. Unfortunately in Java AFAIK if the goal is high performance, duplicating the code is the best approach.

As for the overflow. My understanding is that it is meant to overflow. I am pretty sure that a formal proof can be given that this algorithm is correct.

I will try to give a more informal proof. According to the java specification, "The built-in integer operators do not indicate overflow or underflow in any way.". (http://stackoverflow.com/questions/5131131/what-happens-when-you-increment-an-integer-beyond-its-max-value). Therefore the algorithm practically works in the (mod INT32 signed) ring (http://en.wikipedia.org/wiki/Modular_arithmetic). A known properties of any modular ring is that:
a (mod INT32) + b (mod INT32) = a + b (mod INT32)
a (mod INT32) - b (mod INT32) = a - b (mod INT32)
The algorithm uses only additions and substractions, so by the properties above the decoded value is the same as the encoded, even though it might overflow a few times during the calculations. 

There is actually a unit test that checks exactly the case you mentioned - shouldReadMaxMinValue(). I have also added the INT64 equivalent shouldReadMaxMinValueLong() in this pull request.

One important note (that I was missing for quite some time). 
The INT32 version SHOULD NOT be written by the INT64 version encoder. The difference comes exactly from this overflow. INT_MIN - INT_MAX stored in int would still be a 32 bit integer, however stored in long would be 33 bit integer. Therefore given a list of 32 bit integers, encoded them with INT32 delta encoding might result in a different binary representation than encoding the same list of 32 bit integers with INT64 delta encoding. 

Nevertheless the INT64 decoding works for all possible cases. That's why I am using only one decoder - the INT64 decoder, and I am redirecting the INT32 decoder to it.

nongli:
Thanks! I'm convinced.

+1

parquet-column/src/main/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterForInteger.java
julienledem: It sounds like we should have a short comment in the code to explain this.
It looks like we have unit tests for this. Did you check that we actually go through the worst case scenario in the test?

lunchev: OK, I will add two comments. One for this overflow and one for the INT32 encoding (that should not be done by the INT64 encoder). 

Yes, we have unit tests for [the INT32 case](https://github.com/lunchev/incubator-parquet-mr/commit/bd6d1b398c0e4b43355ae112743500792bdf6edd#diff-7bc3f0a262f0aac6518722769ebcd270R129) and for [the INT64 case](https://github.com/lunchev/incubator-parquet-mr/commit/bd6d1b398c0e4b43355ae112743500792bdf6edd#diff-7bc3f0a262f0aac6518722769ebcd270L142).  I have walked through those two tests with a debugger to check exactly how they works and yes, I think that they check the works case scenario for the overflow. We also have [randomDataTest()](https://github.com/lunchev/incubator-parquet-mr/blob/bd6d1b398c0e4b43355ae112743500792bdf6edd/parquet-column/src/test/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterTest.java#L241) and [randomDataTestLong()](https://github.com/lunchev/incubator-parquet-mr/blob/bd6d1b398c0e4b43355ae112743500792bdf6edd/parquet-column/src/test/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterTest.java#L225) that also hit quite a lot of overflowing cases and they still work perfectly fine.

@julienledem Do you need me to check some other scenario?

julienledem: is the case of -INT_MAX followed by INT_MAX a special case that should be tested specifically? If not then I think we ar fine as is.

julienledem:
Don't you need to change the initialization as well to use it?
https://github.com/apache/incubator-parquet-mr/blob/4ed0bdf1c73fd82d3080d15085675de96d5be0aa/parquet-column/src/main/java/parquet/column/ParquetProperties.java#L163

parquet-column/src/main/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterForInteger.java
lunchev: Integer.MAX_VALUE =  2147483647
Integer.MIN_VALUE = -2147483648
-INT_MAX would be -2147483647

No, the -MAX, MAX test is practically identical to the MIN, MAX test. There is no need to test it specifically.

lunchev:
Yes, you are right. If you want to make it the default encoding for INT64 the initialization should be changed. I just wasn't feeling confident enough to change the default encodings. I will change this as well.

parquet-column/src/main/java/parquet/column/values/delta/DeltaBinaryPackingValuesWriterForInteger.java
julienledem: sorry for being unclear, I meant in the int64 case Long.MIN_VALUE followed by Long.MAX_VALUE and in the int32 case Integer.MIN_VALUE followed by Integer.MAX_VALUE
So if it's already tested then nothing in particular.

julienledem:
It is still behind the PARQUET_2_0 flag.
But yes we want to be sure :)
If we have enough tests then we should add it.
Also probably validate cross coding/decoding with parquet-cpp implementation.

lunchev:
This feature is well tested. I think it's time to make it default especially after it is behind the PARQUET_2_0 flag.
Also added some comments regarding overflow and possible binary incompatibility if INT32 are written with INT64.
I am currently testing the cross coding/decoding with parquet-cpp manually. Is there an automated tool/code for that somewhere?

lunchev:
This build failed because of Travis CI:
"No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself."
It is fine on my machine and it is a trivial change from the last commit that was green. Is there a way to run Travis again?

lunchev:
I cleaned up the code a little bit. It is much smaller now if you compare it to the master.

I spent quite some time trying to validate cross coding/decoding with parquet-cpp implementation, but [parquet-cpp](https://github.com/apache/incubator-parquet-cpp) is still in a very early stage  of development and it doesn't support many things like [Unsupported encoding](https://github.com/apache/incubator-parquet-cpp/blob/master/src/parquet.cc#L226) for DELTA_BINARY_PACKED. 

I am quite confident in this code considering all the unit tests it has and I would be glad to start discussing a possible merge. Of course I will be happy to respond to any comments if anything is still unclear.

rdblue:
@lunchev, could you rebase this on master now that we have renamed to org.apache.parquet packages? I'd like to get this in. Thanks!

lunchev:
@rdblue, I rebased this on the newest master. It is ready to be merged.

lunchev:
Hi @rdblue, is there any progress on merging this PR? It has been ready, waiting to be merged for a while.

rdblue:
@lunchev, I've not had a chance to get to the review backlog in a little while. Sorry about that! I'll try to get this reviewed in the next couple of weeks.

rdblue:
I've been distracted by the 1.8.0 release, but this is the first thing I'll review when it is out. Thanks for being patient!

lunchev:
@rdblue, any news on this PR? More than 2 months passed since the last comment. As far as I can see both 1.8.0 and 1.8.1 are already released.

rdblue:
@lunchev, sorry for the delay. I've just not had the time to look at this.

lunchev:
@rdblue ping

parquet-encoding/src/test/java/org/apache/parquet/column/values/bitpacking/TestByteBitPacking.java
```diff
@@ -62,6 +98,16 @@ private void packUnpack(BytePacker packer, int[] values, int[] unpacked) {
     return values;
   }
 
+  private long[] generateValuesLong(int bitWidth) {
+    long[] values = new long[32];
+    for (int j = 0; j < values.length; j++) {
+      values[j] = ((long)(Math.random() * (1l<<32)) * (long)(Math.random() * (1l<<32))) &
+          ((1l << bitWidth) - 1l);
```
rdblue: Why not use randomLong and a bit mask?

parquet-column/src/main/java/org/apache/parquet/column/values/delta/DeltaBinaryPackingValuesWriter.java
```diff
@@ -130,6 +135,21 @@ public long getBufferedSize() {
 
   @Override
   public void writeInteger(int v) {
+    /*
+     * isInt is needed. Just calling internalWrite might result in a different binary representation.
+     * Example: [Integer.MIN_VALUE, Integer.MAX_VALUE, Integer.MIN_VALUE, Integer.MAX_VALUE]
+     */
+    isInt = true;
```
rdblue: Calling both writeInteger and writeLong would corrupt data. I think that the object model calls to MessageColumnIORecordConsumer determine whether writeInteger or writeLong is used, and that makes it possible for an object model to mix calls.

What about making this writer abstract and implementing each method separately? The code would be nearly identical, but would throw an exception if it is improperly used.

parquet-column/src/test/java/org/apache/parquet/column/values/bitpacking/TestBitPackingColumn.java
```diff
@@ -25,7 +25,6 @@
 import java.io.IOException;
 
 import org.junit.Test;
-
```
rdblue: Nit: white-space change to an otherwise untouched file.

rdblue:
@lunchev, thanks for all your patience on this. I took a look and I'm about ready to commit. I just have the one blocking problem with the writer allowing mixed use of writeInteger and writeLong. If you could fix that, rebase on master, and resubmit, that would be awesome. I think @isnotinvain said he would take a quick look as well.

I rebased locally and it doesn't have any conflicts, so it should be quick. Thanks!

lunchev:
Ok, I will update the pull request by the end of the week. Thanks for looking at it @rdblue

parquet-column/src/main/java/org/apache/parquet/column/values/delta/DeltaBinaryPackingValuesWriter.java
rdblue: While we're at it, why not use int as the internal storage for the integer version? That way we don't use the extra memory (which could matter for staying in cache) and would avoid requiring the mask step just before flushing.

rdblue:
One more thing: I think this should be protected by another writer version since it is technically a new encoding that breaks forward-compatibility with v2.

lunchev:
@rdblue, I did what you requested - DeltaBinaryPackingValuesWriter is now abstract and the INT32 and INT64 versions are implemented separately. I also split the tests for this class into two classes for INT32 and for INT64. The branch is rebased to the newest master.

However I haven't added a new writer version. I think it would be best if you do that, since it is a customer facing change and you can decide best whether PARQUET_2_1 or PARQUET_3_0 or something else would work best.

lunchev:
Thinking about the new writer version, I think it is about time to be able to specify encoding and compression on a per column basis. For example for most of our INT64 columns we want delta + no compression, for some string columns we want RLE + snappy, and for others we want dictionary + snappy. Currently you cannot chose them explicitly, even though the file format allows it.

If parquet supports this, new writer versions like PARQUET_2_0 would no longer be needed, since the difference between two versions would be what encodings and compressions you have at your disposal. You can still use them for defaults though.

rdblue:
Thanks for updating this!

I'm not sure we will need to bump the writer version after all. Julien mentioned something in the last parquet sync-up about the 2.0 stuff not being final yet. I'll post something to the list about it.

I do like having a writer version, though. It helps keep compatibility issues to a minimum and allows us to make guarantees about forward-compatibility. Otherwise we'd have to worry if a particular reader supports some encoding, rather than making sure that the reader implements all encodings for a format version. We can still add the ability to get the granular control you're looking for, though. I think that makes sense.

rdblue:
+1

I just merged this. Thanks for the contribution and your never-ending patience, @lunchev!

l15k4:
Guys I think that it is still not enabled by default in `1.9.0` release. 
```
Types.required(PrimitiveTypeName.INT64).named("foo")
```
I tried to encode Longs that starts from 0 being incremented by 1, but I the resulting parquet file size is identical in 1.9.0 and 1.8.0 versions and `WriterVersion.PARQUET_2_0`.

Is there something I have to enable for Delta Encoding to get applied?

l15k4:
I got it, it is caused by : 
https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultValuesWriterFactory.java#L78-L86

If the `ParquetWriter` used for writing records has `enableDictionary` true, the "Delta Encoding" is not used. 
That means that "Delta Encoding" cannot be used with Dictionary-like encodings? I don't understand how am I supposed to write files with both encodings.

PR #158, opened 2015-04-03
PARQUET-239: Make AvroParquetReader#builder static.

rdblue:
Fixes new API method added since 1.5.0.


julienledem:
SGTM. +1

rdblue:
Tests are failing because this is considered an incompatible change when compared to 1.6.0rc1. The correct previous version is 1.5.0 and this passes locally. #166 updates the previous version to 1.5.0 and fixes other build problems so this will need to be rebased and merged after that one.

julienledem:
Sounds good. I +1'd #166

PR #159, opened 2015-04-03
PARQUET-234: Add ParquetInputSplit methods for compatibility.

rdblue:


julienledem:
What's the expected benefit of returning dummy values like empty collections or null?
If something calls those methods, it will fail or worse, have unexpected behavior.
If we can not reproduce the original behavior, it should either throw an UnsupportedOperationException or not be reinstated.

In short we should decide what should happen when those methods are called before reintroducing them.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -158,6 +168,24 @@ public ParquetInputSplit(
...
+    return null;
+  }
+
+  /**
+   * @return the file schema
+   * @deprecated the file footer is no longer read before creating input splits
+   */
+  @Deprecated
+  public String getFileSchema() {
+    return null;
```
rdblue: TODO: throw UnsupportedOperationException.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -115,6 +116,15 @@ private static long end(List<BlockMetaData> blocks, String requestedSchema) {
   }
 
   /**
+   * @return the block meta data
+   * @deprecated the file footer is no longer read before creating input splits
+   */
+  @Deprecated
+  public List<BlockMetaData> getBlocks() {
+    return Collections.emptyList();
```
rdblue: TODO: throw UnsupportedOperationException

rdblue:
@julienledem, I think you're right that some should throw instead of returning fake values. But for some, I think the empty values work. For example, returning a null requested schema should just cause readers to use the file schema. And I think it should be fine for the metadata accessors to return empty (but valid) metadata. That way there's a chance that older code will succeed.

rdblue:
@julienledem, I added UnsupportedOperationException for all of the methods, as you suggested. I think you're right about it and we can't fake any of that information. We can only make sure that programs that didn't access it at all will succeed, which is what this commit does by making the class binary compatible but throwing exceptions for any behavior changes.

parquet-hadoop/src/main/java/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -26,6 +26,7 @@
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
```
julienledem: not used?

julienledem:
looks good to me.
+1

PR #160, opened 2015-04-04
PARQUET-438: Fix ParquetInputformat.getFooters() order

mingyukim:
ParquetInputFormat.getFooters() should return in the same order as what listStatus() returns


mingyukim:
Closing for now...

PR #161, opened 2015-04-06
PARQUET-242: Fix AvroReadSupport.setAvroDataSupplier.

rdblue:
This should use the supplier class's name, rather than its toString
representation or else loading the class doesn't work.


julienledem:
Do you have a test for this?
otherwise this looks good to me.
+1

PR #163, opened 2015-04-07
PARQUET-220: do not log unnecessary warnings on initializing ParquetReco...

kostya-sh:
...rdReader

Refactored to replace TaskInputOutputContext with TaskAttemptContext.

The warning is now logged only in case if TaskAttemptContext interface doesn't
support counters (e.g. in Hadoop 1.x)


kostya-sh:
Abandoned

PR #164, opened 2015-04-07
PARQUET-241: Fix ParquetInputFormat.getFooters() order

mingyukim:
ParquetInputFormat.getFooters() should return in the same order as what listStatus() returns

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
@@ -29,8 +29,10 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import org.apache.hadoop.conf.Configurable;
@@ -389,7 +391,9 @@ public class ParquetInputFormat<T> extends FileInputFormat<Void, T> {
       return Collections.emptyList();
     }
     Configuration config = ContextUtil.getConfiguration(jobContext);
-    Map<FileStatusWrapper, Footer> footersMap = new HashMap<FileStatusWrapper, Footer>();
+    // Use LinkedHashMap to preserve the insertion order and ultimately to return the list of
+    // footers in the same order as the list of file statuses returned from listStatus()
+    Map<FileStatusWrapper, Footer> footersMap = new LinkedHashMap<FileStatusWrapper, Footer>();
     Set<FileStatus> missingStatuses = new HashSet<FileStatus>();
     Map<Path, FileStatusWrapper> missingStatusesMap =
             new HashMap<Path, FileStatusWrapper>(missingStatuses.size());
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/test/java/parquet/hadoop/TestInputFormat.java
```diff
@@ -378,6 +387,62 @@ public void testDeprecatedConstructorOfParquetInputSplit() throws Exception {
...
+    FileInputFormat.setInputPaths(job, url);
+    List<Footer> footers = new ParquetInputFormat<Object>().getFooters(job);
+    for (int i = 0; i < numFiles; i++) {
+      Footer footer = footers.get(i);
+      File file = new File(tempDir, String.format("part-%05d.parquet", i));
+      assertEquals("file:" + file.getAbsolutePath(), footer.getFile().toString());
+    }
+  }
+
+  private void createParquetFile(File file) throws IOException {
```
mingyukim: This was my first attempt at creating a test parquet file. Please let me know if there's an easier way.

mingyukim:
Ping. Can someone review this?

rdblue:
Hi @mingyukim, sorry about the delay. I didn't see that you had a PR up or that you had asked me to review it. It looks like this won't currently merge because of the rename done in 1.7.0. Could you rebase this patch on master and ping me to review? Thanks!

mingyukim:
Thanks for following up! Instead of merging it with master, I just went ahead and cherry-picked it on top of master.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInputFormat.java
```diff
@@ -378,6 +385,62 @@ public void testDeprecatedConstructorOfParquetInputSplit() throws Exception {
     shouldSplitStartBe(splits, 0, 50);
   }
 
+  @Test
+  public void testGetFootersReturnsInCorrectOrder() throws IOException {
```
rdblue: Nit: this should probably be "predictable order" rather than "correct order".

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -407,33 +407,36 @@ private static void staticAddInputPathRecursively(List<FileStatus> result,
...
-    for (Footer newFooter : newFooters) {
-      // Use the original file status objects to make sure we store a
-      // conservative (older) modification time (i.e. in case the files and
-      // footers were modified and it's not clear which version of the footers
-      // we have)
-      FileStatusWrapper fileStatus = missingStatusesMap.get(newFooter.getFile());
-      footersCache.put(fileStatus, new FootersCacheValue(fileStatus, newFooter));
+    List<Footer> footers = new ArrayList<Footer>(statuses.size());
+    for (FileStatus status : statuses) {
+      FileStatusWrapper fileStatusWrapper = new FileStatusWrapper(status);
```
rdblue: Instead of creating a new FileStatusWrapper for each file here, what about a slightly different way of doing it:
* For footersMap, use a LinkedHashMap to preserve the key insertion order
* Insert the FileStatusWrapper for missing footers pointing to null
* When assembling the list of footers, loop over the map and if the footer is null, look it up in the cache using the existing FileStatusWrapper

That would save allocations as well as so many footersMap lookups, while also making the part where the cache is loaded completely separate from the rest (it could be refactored into a separate method).

mingyukim:
Thanks for the feedback. I addressed both of the comments in the new commit.

PR #165, opened 2015-04-07
PARQUET-243: Add Avro reflect support

rdblue:


parquet-avro/src/main/java/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -36,7 +34,6 @@
 import parquet.io.api.PrimitiveConverter;
 import parquet.schema.GroupType;
 import parquet.schema.MessageType;
-import parquet.schema.OriginalType;
 import parquet.schema.Type;
 
 class AvroIndexedRecordConverter<T extends IndexedRecord> extends GroupConverter {
```
tomwhite: Can you add a bit of documentation to explain the salient differences between this class and AvroRecordConverter? It's quite hard to tell just by looking.

parquet-avro/src/main/java/parquet/avro/AvroWriteSupport.java
```diff
@@ -41,7 +43,14 @@
...
  * this class directly.
  */
-public class AvroWriteSupport extends WriteSupport<IndexedRecord> {
+public class AvroWriteSupport<T> extends WriteSupport<T> {
+
+  public static final String AVRO_DATA_SUPPLIER = "parquet.avro.write.data.supplier";
+
+  public static void setAvroDataSupplier(
+      Configuration configuration, Class<? extends AvroDataSupplier> suppClass) {
+    configuration.set(AVRO_DATA_SUPPLIER, suppClass.toString());
```
tomwhite: getName()?

parquet-avro/src/main/java/parquet/avro/AvroParquetReader.java
```diff
@@ -30,7 +30,7 @@
 /**
  * Read Avro records from a Parquet file.
  */
-public class AvroParquetReader<T extends IndexedRecord> extends ParquetReader<T> {
+public class AvroParquetReader<T> extends ParquetReader<T> {
```
tomwhite: Is this a binary compatible change? If not, then we may need a separate class. Similarly, for AvroParquetWriter, AvroReadSupport and AvroWriteSupport.

parquet-avro/src/main/java/parquet/avro/AvroReadSupport.java
```diff
@@ -33,7 +34,7 @@
  * Generic. Users should use {@link AvroParquetReader} or {@link AvroParquetInputFormat} rather than using
```
tomwhite: Update the comment to cover Reflect.

tomwhite:
This looks great.

I think AvroParquetOutputFormat needs to be parameterized by T (like AvroParquetInputFormat already is). Also can you add a TestReflectInputOutputFormat to test the formats when using reflect.

rdblue:
@tomwhite, I've updated this and rebased it after the rename.

I didn't need to duplicate the read or write classes where the type bound has changed because they are extending classes or interfaces with a weaker type bound. In that case, the JVM writes the method using the parent's descriptor, so they are binary compatible. I also verified this expected behavior locally with test classes, by compiling against &lt;T extends Something&gt; then changing to just &lt;T&gt; and verifying that the dependency binaries are interchangeable. In addition, I made sure that without the superclass or superinterface, binary compatibility is indeed broken.

I think this is ready to go. Thanks for taking a look!

tomwhite:
Thanks for checking the compatibility! I think this is ready now. +1

PR #166, opened 2015-04-07
PARQUET-235: Fix parquet.metadata compatibility.

rdblue:
ColumnPath and Canonicalizer were moved from parquet-hadoop to
parquet-common in parquet.common.{internal,schema}, which broke
compatibility and would require bumping the major version. This moves
the classes back into parquet.hadoop.metadata and adds temporary
exclusions for the move between modules. There are no breaking changes
to the classes themselves, verified by copying them into parquet-hadoop
and building.

This also changes the previous version back to 1.5.0 rather than an RC
(which carries no compatibility guarantees, though this is compatible
with both version). It also adds an exclusions for a false positive in
Binary.

### DIFF
--- a/parquet-column/src/main/java/parquet/filter2/predicate/FilterApi.java
+++ b/parquet-column/src/main/java/parquet/filter2/predicate/FilterApi.java
@@ -20,7 +20,7 @@ package parquet.filter2.predicate;
 
 import java.io.Serializable;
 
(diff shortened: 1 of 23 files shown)


julienledem:
we should probably make the package name match the module again when we rename everything into the org.apache.parquet namespace.

julienledem:
+1

PR #169, opened 2015-04-08
PARQUET-247: Add DATE mapping in ValidTypeMap of filter2

dongche:
When Hive use Parquet filter predicate, the Date type is converted to Integer. In ValidTypeMap, it map the class and Parquet type. It throw exception when checking the data type Date.
We should add the map to support Date.


dongche:
Hi, @rdblue , @isnotinvain , I am working on HIVE-10253 and find we may need this change, otherwise filter predicate with Date type cannot pass the validation check.

What do you think? Thanks!

AdamPD:
We should be adding the mappings for the other types as well (TIME_MILLIS and TIMESTAMP_MILLIS).

I am having the same challenge doing PPD for Apache Drill with those types.  I can certainly create another JIRA item for those, but it seems trivial to add it to this one.

dongche:
Sounds good! Thanks.
Updated to add mapping for those types.

dongche:
Verified the updates in my local env again. Build and test successfully.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ValidTypeMap.java
```diff
@@ -74,9 +74,22 @@ private static void add(Class<?> c, FullTypeDescriptor f) {
     // Both of these binary types are valid
     add(Binary.class, new FullTypeDescriptor(PrimitiveTypeName.BINARY, null));
     add(Binary.class, new FullTypeDescriptor(PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY, null));
+    add(Binary.class, new FullTypeDescriptor(PrimitiveTypeName.INT96, null));
```
isnotinvain: so, for something like this, does it make more sense to use a different java class other than Binary? we could use String for UTF8, and BigNum or something here?

Maybe that could be a higher level API on top of this one though.

isnotinvain:
per: https://issues.apache.org/jira/browse/PARQUET-201 do you mind adding the rest of the primitives too?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ValidTypeMap.java
dongche: Add this here is just to make filter checking pass if user needs INT96 for a filter.
It seems INT96 is designed as Binary in PrimitiveType. Maybe higher level API could bind it with other java class. Not sure which to use here other than Binary?

dongche:
Sure, my pleasure!
Updated code with the mapping for OriginalType INT_* and UINT_*

nandorKollar:
Fixed by removing OriginalType check in #219

PR #175, opened 2015-04-14
PARQUET-252 : Support nests container types for scrooge support

tsdeng:
parquet should support nested container type for scrooge, like list<list> or list<map>

### DIFF
--- a/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
+++ b/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
@@ -297,7 +297,7 @@ public class ScroogeStructConverter {
    * @return
    */
   private String fixNestListOrSetName(String name) {
-    return name.replaceAll("_list_elem|_set_elem","");
+    return name.replaceAll("_list_elem|_set_elem", "");
   }
 
   private ThriftType convertStructTypeField(ThriftStructFieldInfo f) {
@@ -330,11 +330,15 @@ public class ScroogeStructConverter {
         enumValues.add(new ThriftType.EnumValue(enumDesc.id, enumDesc.originalName));
       }
       return new ThriftType.EnumType(enumValues);
-    } catch (Exception e) {
+    } catch (ReflectiveOperationException e) {
+      throw new ScroogeSchemaConversionException("Can not convert enum field " + fieldName, e);
+    } catch (RuntimeException e) {
       throw new ScroogeSchemaConversionException("Can not convert enum field " + fieldName, e);
     }
+
   }
 
+  //In scrooge generated class, if a class is a union, then it must have a field called "Union"
   private boolean isUnion(Class klass){
     for(Field f: klass.getDeclaredFields()) {
       if (f.getName().equals("Union"))
--- a/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
+++ b/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
@@ -192,7 +192,7 @@ public class ScroogeStructConverter {
     ThriftType elementType = convertClassToThriftType(fieldName + "_set_elem", requirement, valueManifest);
     //Set only has one sub-field as element field, therefore using hard-coded 1 as fieldId,
     //it's the same as the solution used in ElephantBird
-    ThriftField elementField = generateFieldWithoutId(fixNestListOrSetName(fieldName), requirement, elementType);
+    ThriftField elementField = generateFieldWithoutId(fieldName, requirement, elementType);
     return new ThriftType.SetType(elementField);
   }
 
@@ -202,7 +202,7 @@ public class ScroogeStructConverter {
 
   private ThriftType convertListTypeField(String fieldName, Manifest<?> valueManifest, Requirement requirement) {
     ThriftType elementType = convertClassToThriftType(fieldName + "_list_elem", requirement, valueManifest);
-    ThriftField elementField = generateFieldWithoutId(fixNestListOrSetName(fieldName) , requirement, elementType);
+    ThriftField elementField = generateFieldWithoutId(fieldName , requirement, elementType);
     return new ThriftType.ListType(elementField);
   }
 
@@ -277,27 +277,6 @@ public class ScroogeStructConverter {
     } else {
       return convertStructFromClass(typeClass);
     }
-  }
-
-  /**
-   * There is a bug/inconsistency in ThriftStructConverter:
-   * The field name of a nested list/set has the same name as its parent. for example:
-   *  1: required list<list<int>> lm
-   * the name of the inner list is lm, but is should be lm_list_elem.
-   *
-   * This bug/inconsistency does not affect the converted parquet schema because ThriftSchemaConverter ignores the name
-   * of inner list.
-   *
-   * Here in ScroogeStructConverter the same behavior is replicated to generate the same ThriftStruct with
-   * ThriftStructConverter
-   *
-   * TODO: This fix might not be necessary since parquet mostly operates on MessageType, the ThriftStruct is an intermediate
-   * layer for generating MessageType from a thrift/scrooge class. The fix could be removed once we are confident it's
-   * safe for doing so.
-   * @return
-   */
-  private String fixNestListOrSetName(String name) {
-    return name.replaceAll("_list_elem|_set_elem", "");
   }
 
   private ThriftType convertStructTypeField(ThriftStructFieldInfo f) {
--- a/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
+++ b/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
@@ -290,6 +290,10 @@ public class ScroogeStructConverter {
    *
    * Here in ScroogeStructConverter the same behavior is replicated to generate the same ThriftStruct with
    * ThriftStructConverter
+   *
+   * TODO: This fix might not be necessary since parquet mostly operates on MessageType, the ThriftStruct is an intermediate
+   * layer for generating MessageType from a thrift/scrooge class. The fix could be removed once we are confident it's
+   * safe for doing so.
    * @return
    */
   private String fixNestListOrSetName(String name) {
--- a/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
+++ b/parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
@@ -69,7 +69,7 @@ public class ScroogeStructConverter {
   }
 
   private ThriftType.StructType convertCompanionClassToStruct(Class<?> companionClass) {
-    ThriftStructCodec companionObject = null;
+    ThriftStructCodec<?> companionObject;
     try {
       companionObject = (ThriftStructCodec<?>)companionClass.getField("MODULE$").get(null);
     } catch (ReflectiveOperationException e) {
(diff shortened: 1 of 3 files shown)


parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
```diff
@@ -273,11 +261,45 @@ private ThriftType convertClassToThriftType(Class typeClass) {
...
+   * Here in ScroogeStructConverter the same behavior is replicated to generate the same ThriftStruct with
+   * ThriftStructConverter
+   *
+   * TODO: This fix might not be necessary since parquet mostly operates on MessageType, the ThriftStruct is an intermediate
+   * layer for generating MessageType from a thrift/scrooge class. The fix could be removed once we are confident it's
+   * safe for doing so.
+   * @return
+   */
+  private String fixNestListOrSetName(String name) {
+    return name.replaceAll("_list_elem|_set_elem","");
```
julienledem: It sounds like we should fix the bug instead of replicating the buggy behavior.

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
```diff
@@ -295,36 +317,56 @@ private List getEnumList(String enumName) throws ClassNotFoundException, Illegal
...
       for (Object enumObj : enumCollection) {
-        ScroogeEnumDesc enumDesc = ScroogeEnumDesc.getEnumDesc(enumObj);
-        //be compatible with thrift generated enum which have capitalized name
+        ScroogeEnumDesc enumDesc = ScroogeEnumDesc.fromEnum(enumObj);
         enumValues.add(new ThriftType.EnumValue(enumDesc.id, enumDesc.originalName));
       }
       return new ThriftType.EnumType(enumValues);
-    } catch (ReflectiveOperationException e) {
-      throw new ScroogeSchemaConversionException("Can not convert enum field " + f, e);
+    } catch (Exception e) {
```
julienledem: are there Exceptions other than RuntimeExceptions?

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
```diff
@@ -295,36 +317,56 @@ private List getEnumList(String enumName) throws ClassNotFoundException, Illegal
...
+      throw new ScroogeSchemaConversionException("Can not convert enum field " + fieldName, e);
+    }
+  }
+
+  private boolean isUnion(Class klass){
+    for(Field f: klass.getDeclaredFields()) {
+      if (f.getName().equals("Union"))
+        return true;
```
julienledem: I don't understand this logic.
Is it a union if one of the fields is called "Union"?

parquet-scrooge/src/test/java/parquet/scrooge/ScroogeStructConverterTest.java
```diff
@@ -18,106 +18,147 @@
...
-import parquet.scrooge.test.TestMapComplex;
-import parquet.scrooge.test.TestMapPrimitiveKey;
-import parquet.scrooge.test.TestMapPrimitiveValue;
-import parquet.scrooge.test.TestOptionalMap;
-import parquet.scrooge.test.TestPersonWithAllInformation;
-import parquet.scrooge.test.TestSetPrimitive;
-import parquet.scrooge.test.TestUnion;
-import parquet.scrooge.test.StringAndBinary;
+import parquet.schema.MessageType;
+import parquet.scrooge.test.*;
```
julienledem: keep the individual imports, it makes it easier to read on web.

parquet-scrooge/src/test/java/parquet/scrooge/ScroogeStructConverterTest.java
```diff
@@ -18,106 +18,147 @@
...
+    try {
+      Class<? extends TBase<?, ?>> thriftClass = (Class<? extends TBase<?, ?>>)Class.forName(scroogeClass.getName().replaceFirst("parquet.scrooge.test", "parquet.thrift.test"));
+      ThriftType.StructType structFromThriftSchemaConverter = new ThriftSchemaConverter().toStructType(thriftClass);
+      ThriftType.StructType structFromScroogeSchemaConverter = new ScroogeStructConverter().convert(scroogeClass);
+
+      assertEquals(structFromThriftSchemaConverter, structFromScroogeSchemaConverter);
+      assertEquals(toParquetSchema(structFromThriftSchemaConverter), toParquetSchema(structFromScroogeSchemaConverter));
+    } catch (ClassNotFoundException e) {
+      fail("ClassNotFoundException: " + e.getMessage());
+    }
```
julienledem: why not just let the exception propagate up? At least we would have the correct stacktrace

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
tsdeng: if it's a union, then it must have a field called "Union"

julienledem:
Overall this looks good to me. I made a few comments.

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
tsdeng: we have a unit test for union covering this

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
```diff
@@ -98,8 +98,7 @@ private Class getCompanionClass(Class klass) {
       //each struct has a generated fieldInfos method to provide metadata to its fields
       try {
         Object r = klass.getMethod("fieldInfos").invoke(c);
-        Iterable<ThriftStructFieldInfo> a = JavaConversions$.MODULE$.asJavaIterable((scala.collection.Iterable<ThriftStructFieldInfo>)r);
-        return a;
+        return JavaConversions$.MODULE$.asJavaIterable((scala.collection.Iterable<ThriftStructFieldInfo>)r);
       } catch (ReflectiveOperationException e) {
```
isnotinvain: catch ClassCastException too?

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
```diff
@@ -273,11 +261,45 @@ private ThriftType convertClassToThriftType(Class typeClass) {
...
+   * Here in ScroogeStructConverter the same behavior is replicated to generate the same ThriftStruct with
+   * ThriftStructConverter
+   *
+   * TODO: This fix might not be necessary since parquet mostly operates on MessageType, the ThriftStruct is an intermediate
+   * layer for generating MessageType from a thrift/scrooge class. The fix could be removed once we are confident it's
+   * safe for doing so.
+   * @return
+   */
+  private String fixNestListOrSetName(String name) {
+    return name.replaceAll("_list_elem|_set_elem", "");
```
isnotinvain: I worry that this is unsafe -- what about a thrift schema like:
`1: required list<bool> _list_elem`

How is the TBase implementation able to accomplish this behavior without a string.replace?

parquet-scrooge/src/test/thrift/test.thrift
```diff
@@ -169,6 +169,95 @@ struct TestFieldOfEnum{
...
 
 struct StringAndBinary {
-  1: required string s;
-  2: required binary b;
+  1: required string s
+  2: required binary b
+}
+
+#fixture fox nested structures
+struct NestedList {
```
isnotinvain: do we need copies of these for tbase as well? How are we comparing w/ tbase?

tsdeng: What do you mean? For the thrift file, we generate both scrooge and tbase classes. And they use the same same thrift definition, so we can compare the schema derived from a thrift class and a scrooge class

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
tsdeng: There is a bug in TBase implementation that's causing it not using the element names...
It's a bug in TBase implementation. I can either remove this method or keep it consistent with current TBase implementation. I'm creating a job to test if removing it is safe, if it's safe then I can delete this method.

parquet-scrooge/src/test/thrift/test.thrift
isnotinvain: ok, I didn't realize we generated both tbase and scrooge for these.

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
isnotinvain: If we can fix the bug I agree, lets do that.

But if we want this behavior, what I mean is, instead of doing a replace, can't we just use `fieldName` instead of `fieldName + "_list_elem"` above? I assume that's what the TBase impl does?

isnotinvain: Meaning, instead of injecting "_list_elem" and then removing it later, just don't add it at all?

tsdeng: no we can't, since if the leaf node is a map, then the _list_elem stuff need to be passed all the way down to the leaf node.

tsdeng: This bug might be trickier than you thought... ThriftSchema converter use the name that is converted by elephant_bird. but for the list element, it's using the container's name. But for map, it's using the correct key name. And the tricky case comes here: when you have list<list<list<map>>>  the name of map key is ls_list_elem_list_elem_list_elem_map_key, but the parent's name is ls. Basically all the parents do not have the _list_elem suffixies due to the bug in TBase implementation, but the leaf node should have the _list_elem suffixes.  I have unit tests for this which you can reference.  As I said, this fix is purely for being conservative and consistent with TBase implementation. Hopefully I can do a more thorough test and remove this fix/hack

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
tsdeng: fixed

isnotinvain: while you're in here, can you explicitly catch all the children of ReflectiveOperationException instead of ReflectiveOperationException? ReflectiveOperationException was added in java 7, so java 6 users get errors when using this (I heard from someone in the slack room)

isnotinvain:
with the hax/fix removed, why does it still produce the same thing as the TBase one? Did you fix it there too (I don't see that, maybe I missed it?)

isnotinvain:
also, maybe we should encode the rules about _list_elem and so on in a shared helper class (like the one Ryan set up?)

tsdeng:
@isnotinvain after removing the hack/fix, the ThriftStruct generated from ThriftStructConverter and ScroogeSchemaConverter are different, but he MessageType derived from the ThriftStruct are still the same. It's because when converting from ThriftStruct to MessageType, the name of nested element are renamed by ThriftSchemaConverter. So I modified the test to only compare at the MessageType level

parquet-scrooge/src/main/java/parquet/scrooge/ScroogeStructConverter.java
tsdeng: fixed

tsdeng:
@isnotinvain the encode rules about _list_elem is not really shared by others: It's specific to thrift/scrooge since it's on used at the ThriftStruct level, NOT the MessageType level, it's easy to misunderstand and miss this difference. The ThriftStruct serves as a intermediate level before converting to MessageType(parquet schema), both Thrift and Scrooge are converted to ThriftStruct first and then get converted to MessageType. The behavior of appending _list_elem to the nested element is only specific to elephant-bird. As you can see, in ThriftSchemaConverter.ThriftStructConverter, it does not has any code to appending the _list_elem suffix, and the reason for that is it's done in elephant bird's code. But scrooge schema converter does not use elephant bird therefore I need to replicate this behavior manually. So instead of putting it in to a shared module, I would put them as private static method inside ScroogeSchemaConverter.

isnotinvain:
Thanks, that clarifies that a lot.

isnotinvain:
+1

PR #179, opened 2015-04-17
PARQUET-23: Rename to org.apache.parquet.

rdblue:
This includes all of the code updates by module (so that we can use github for the review). I think this is currently all of the changes needed.


isnotinvain:
Does git understand these to be moves, or do we lose the history for these files?

rdblue:
Yes, git understands the moves, though it doesn't track them explicitly like svn would. In order to see the full history, you need to use the `--follow` option to tell git to look for moves when it sees new files and deletions.

isnotinvain:
I spot checked some files. I fought with git but after lots of googling I still can't find a way to get a sane diff to work. `git log --follow` understands the file renames no problem, but I can't find the magic that will get `git diff` to do the same, it always shows files deleted + added instead of just showing me what changed int the moved files, even with various rename following settings turned on.

isnotinvain:
+1, looked at the individual commits which were much easier to diff.

tsdeng:
+1 checked individual commits

PR #180, opened 2015-04-22
PARQUET-213: Logs discarded and projected fields after projection pushdown.

MansurAshraf:
Reviving the old PR after clean up. I noticed that @isnotinvain has added new projection filter API so not sure if this change still applies

### DIFF
--- a/parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
+++ b/parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
@@ -35,6 +35,7 @@ import static parquet.schema.Types.primitive;
 import java.util.ArrayList;
 import java.util.List;
 
+import parquet.Log;
 import parquet.schema.GroupType;
 import parquet.schema.MessageType;
 import parquet.schema.OriginalType;
@@ -44,6 +45,7 @@ import parquet.schema.Type;
 import parquet.schema.Types.PrimitiveBuilder;
 import parquet.thrift.projection.FieldProjectionFilter;
 import parquet.thrift.projection.FieldsPath;
+import parquet.thrift.projection.ProjectionFilter;
 import parquet.thrift.projection.ThriftProjectionException;
 import parquet.thrift.struct.ThriftField;
 import parquet.thrift.struct.ThriftType;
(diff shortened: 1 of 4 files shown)


parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -56,17 +58,19 @@
  */
 public class ThriftSchemaConvertVisitor implements ThriftType.TypeVisitor {
 
-  public FieldProjectionFilter getFieldProjectionFilter() {
-    return fieldProjectionFilter;
-  }
+    private static final Log logger = Log.getLog(ThriftSchemaConvertVisitor.class);
```
rdblue: Looks like your indentation was accidentally off, causing this change to be larger than you intended.

rdblue: In fact, this is a problem throughout, which makes the PR hard to review. I'd appreciate it if you could fix this and force-push to update the PR.

parquet-thrift/src/main/java/parquet/thrift/projection/FieldProjectionFilter.java
```diff
@@ -18,17 +18,21 @@
...
  * Filter thrift attributes using glob syntax.
  *
  * @author Tianshuo Deng
  */
-public class FieldProjectionFilter {
+public class FieldProjectionFilter implements ProjectionFilter {
   public static final String PATTERN_SEPARATOR = ";";
   List<PathGlobPatternStatus> filterPatterns;
+    private final Set<String> matchedPaths = new HashSet<String>();
+    private final Set<String> unmatchedPaths = new HashSet<String>();
```
rdblue: Do we expect paths to have duplicates? It seems like a List would work just as well here.

parquet-thrift/src/main/java/parquet/thrift/projection/ProjectionFilter.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package parquet.thrift.projection;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Mansur Ashraf.
+ */
+public interface ProjectionFilter {
```
rdblue: What is the purpose of this interface? It seems like we don't really need it when we could just continue to use FieldProjectionFilter directly.

parquet-thrift/src/main/java/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -82,7 +83,7 @@ private void checkUnmatchedProjectionFilter(FieldProjectionFilter filter) {
     return new ThriftStructConverter().toStructType(thriftClass);
   }
 
-  private static class ThriftStructConverter {
+  protected static class ThriftStructConverter {
```
rdblue: Is this change needed?

parquet-thrift/src/main/java/parquet/thrift/projection/FieldProjectionFilter.java
```diff
@@ -68,16 +72,26 @@ public FieldProjectionFilter(String filterDescStr) {
     }
   }
 
+    @Override
   public boolean isMatched(FieldsPath path) {
```
rdblue: Can tracking the paths that have matched or not be done where this is called? I think it would be better to not keep state since this class functions as a predicate.

isnotinvain:
So actually, this PR is going to conflict in a big way with https://github.com/apache/parquet-mr/pull/150 which has been merge-able for a while (but not merged due to the freeze). I'd like to get that one in first, and then we can add this too.

PR #181, opened 2015-04-27
PARQUET-261: Extend classloader fix to rest of codebase.

alexrobbins:
This applies the fix from https://github.com/Parquet/parquet-mr/pull/289 to the rest of the codebase.

Anywhere a Hadoop Configuration object is available, the utility method .getClassByName is used. Otherwise, a new utility class ClassLoading is used.


alexrobbins:
Unfortunately, this doesn't fix the issue for me.

PR #182, opened 2015-04-27
PARQUET-211: Update version for 1.8.0 development.

rdblue:


isnotinvain:
+1

What's the new process for releasing to maven central though? Will we still do that through sonatype, or does apache provide something we can use? Do we need to cut a 1.7.0 maven release as well?

rdblue:
With this in, I will need to update the release profile to use maven like parquet-format does. I'll do that in the 1.7.0 branch and backport them to master so we don't need to block master any longer than necessary.

PR #186, opened 2015-04-29
PARQUET-265: Update POM files for Parquet TLP.

rdblue:


isnotinvain:
All the travis builds seem to be failing for the same reason as this one. There was a travis incident yesterday, but it seems that that is over and this issue remains. I'm not really sure what we can do to debug this since we don't have that much control over the travis builds.

isnotinvain:
OK, think we have a guess at the issue -- one of the scrooge dependencies transitively pulled in a snapshot that has since been purged. Seems like that dependency was improperly published. Upgrading the scrooge plugin should fix this.

isnotinvain:
This will take a long time for travis to get to, so queued it up here as well:
https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/60607157

rdblue:
@isnotinvain or @julienledem can you take a look at this? It is needed for the release.

isnotinvain:
+1

PR #187, opened 2015-04-29
PARQUET-268: Downgrade scrooge-maven-plugin.

rdblue:
The 3.17.0 version no longer works because a transitive dependency has
been purged. 3.18.0 is the natural upgrade, but fails with a
configuration problem. The latest documentation on updates is for moving
to 3.17.0, so the easiest solution that works is to downgrade to 3.16.0.
The build is working.

### DIFF
--- a/parquet-scrooge/pom.xml
+++ b/parquet-scrooge/pom.xml
@@ -185,7 +185,7 @@
         <plugin>
             <groupId>com.twitter</groupId>
             <artifactId>scrooge-maven-plugin</artifactId>
-            <version>3.17.0</version>
+            <version>3.16.3</version>
             <configuration>
                 <outputDirectory>${project.build.directory}/generated-test-sources/scrooge</outputDirectory>
                 <thriftNamespaceMappings>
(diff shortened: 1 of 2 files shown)


rdblue:
@isnotinvain I think this fixes the problem. Can you take a look?

isnotinvain:
Lets do 3.16.3 (latest one that works) and add a TODO and pointer to a jira for upgrading to 3.19 when it's ready

But yes, +1 this is what is needed for now

isnotinvain:
filed: https://issues.apache.org/jira/browse/PARQUET-269
can you reference it in the pom file? Thanks

rdblue:
Updated. I'll merge this when the tests pass.

isnotinvain:
Thanks!

isnotinvain:
This will take a long time for travis to get to, so queued it up here as well:
https://travis-ci.org/isnotinvain/incubator-parquet-mr/builds/60607157

rdblue:
@isnotinvain I've ignored the failing tests. I think that's the best way to fix master. What do you say?

isnotinvain:
+1 to unblock the build, and I'll try to get the upgraded version in as soon as it's available.

rdblue:
Fixed. Thanks for your help, @isnotinvain!

PR #189, opened 2015-05-06
PARQUET-287: Keep a least 1 column from union members when projecting thrift unions

isnotinvain:
Currently, the projection API allows you to select only some "kinds" of a union, or to drop a required union entirely. This becomes an issue when assembling these records, as they will be appear to be unions of an unknown type (how do you coerce an empty struct into a union?). The way this case is handled for primitives is by supplying a default value (like 0, or null). However, with a union, you have to choose what "kind" of the union it will act as, and in the interest of not being misleading, this PR reads one column to figure out what the correct "kind" is.

In the future, the better solution is to filter these records out -- a projection is really a request for a filter in this case. But for now, this should get us correctness without involving the filter API.

I think this PR needs some more tests before merging, but I wanted to get it out and get some feedback now. 

I also refactored how ThriftSchemaVisitor works to not be stateful, by explicitly passing state through the recursion -- this makes it much easier to reason about.

*edit* This PR also includes a fix for PARQUET-275 because I encountered it during testing.

*edit 2* This PR also includes a fix for PARQUET-283


parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -55,203 +73,239 @@
...
+    ThriftField keyField = mapType.getKey();
+    ThriftField valueField = mapType.getValue();
+
+    ConvertedField convKey = keyField.getType().accept(this,
+        new State(state.path.push(keyField), REQUIRED, "key"));
+
+    ConvertedField convValue = valueField.getType().accept(this,
+        new State(state.path.push(valueField), OPTIONAL, "value"));
+
+    if (!convKey.isKeep()) {
```
isnotinvain: Note: add tests for `Map<Union, T>`, `Map<T, Union>`, `Map<Union, Union>`, `List<Union>` etc.

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -18,36 +18,54 @@
  */
```
isnotinvain: this file changed enough that you might just want to hit "view" above and read the new copy of it

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -55,203 +73,266 @@
...
+    State state = new State(new FieldsPath(), REPEATED, "ParquetSchema");
 
-  public ThriftSchemaConvertVisitor(FieldProjectionFilter fieldProjectionFilter) {
-    this.fieldProjectionFilter = Preconditions.checkNotNull(fieldProjectionFilter, "fieldProjectionFilter");
+    ConvertedField conv = struct.accept(new ThriftSchemaConvertVisitor(filter, true), state);
+
+    MessageType messageType;
+
+    if (!conv.isKeep()) {
+      // TODO: should this be fatal? Projecting away everything?
```
isnotinvain: this should be fatal, right?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ConvertedField.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+package org.apache.parquet.thrift;
+
+import org.apache.parquet.ShouldNeverHappenException;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.thrift.projection.FieldsPath;
+
+import static org.apache.parquet.Preconditions.checkNotNull;
+
+public interface ConvertedField {
```
tsdeng: Java doc on the responsibility and usages of this interface?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
isnotinvain: fixed

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
isnotinvain: @tsdeng what do you think?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
```diff
@@ -55,203 +73,266 @@
...
 
-  private Type currentType;
-  private Type.Repetition currentRepetition = Type.Repetition.REPEATED; // MessageType is repeated GroupType
-  private String currentName = "ParquetSchema";
+  public static MessageType convert(StructType struct, FieldProjectionFilter filter) {
+    State state = new State(new FieldsPath(), REPEATED, "ParquetSchema");
 
-  public ThriftSchemaConvertVisitor(FieldProjectionFilter fieldProjectionFilter) {
-    this.fieldProjectionFilter = Preconditions.checkNotNull(fieldProjectionFilter, "fieldProjectionFilter");
+    ConvertedField conv = struct.accept(new ThriftSchemaConvertVisitor(filter, true), state);
```
tsdeng: no need to save typing here. conv can mean converter and other things. convertedField is fine

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java
tsdeng: Should be fatal, but in this PR https://github.com/apache/parquet-mr/pull/195
It enforces GroupType to have at least one child.. But may be throwing here in your code is more clear for debugging purposes?

tsdeng:
Minor comments, otherwise LGTM

tsdeng:
+1

PR #194, opened 2015-05-11
PARQUET-279 : Check empty struct in compatibility checker

tsdeng:
Add the empty struct check in the CompatibilityChecker util.
Parquet currently does not support empty struct/group without leaves

### DIFF
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
@@ -124,11 +124,11 @@ public class ThriftSchemaConverter {
         break;
       case SET:
         final Field setElemField = field.getSetElemField();
-        type = new ThriftType.SetType(toThriftField(name, setElemField, requirement));
+        type = new ThriftType.SetType(toThriftField(setElemField.getName(), setElemField, requirement));
         break;
       case LIST:
         final Field listElemField = field.getListElemField();
-        type = new ThriftType.ListType(toThriftField(name, listElemField, requirement));
+        type = new ThriftType.ListType(toThriftField(listElemField.getName(), listElemField, requirement));
         break;
       case ENUM:
         Collection<TEnum> enumValues = field.getEnumValues();
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
@@ -213,7 +213,7 @@ class CompatibleCheckerVisitor implements ThriftType.TypeVisitor<Void, Void> {
     short oldMaxId = 0;
 
     if (newStruct.getChildren().size() == 0) {
-      report.emptyStruct("new struct has no child: " + path);
+      report.emptyStruct("new struct is an empty struct: " + path);
     }
 
     for (ThriftField oldField : currentOldType.getChildren()) {
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
@@ -107,7 +107,7 @@ class State {
   }
 }
 
-class CompatibleCheckerVisitor implements ThriftType.TypeVisitor<Void, State> {
+class CompatibleCheckerVisitor implements ThriftType.StateVisitor<Void, State> {
 
   CompatibilityReport report = new CompatibilityReport();
 
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -33,6 +33,7 @@ import org.apache.parquet.thrift.struct.ThriftType.I16Type;
 import org.apache.parquet.thrift.struct.ThriftType.I32Type;
 import org.apache.parquet.thrift.struct.ThriftType.I64Type;
 import org.apache.parquet.thrift.struct.ThriftType.StringType;
+import org.apache.parquet.Strings;
 
 /**
  * A checker for thrift struct to enforce its backward compatibility, returns compatibility report based on following rules:
@@ -81,12 +82,8 @@ class CompatibilityReport {
   }
 
   public String prettyMessages() {
-    StringBuffer message = new StringBuffer();
-    for(String m: messages) {
-      message.append(m);
-      message.append("\n");
-    }
-    return message.toString();
+
+    return Strings.join(messages, "\n");
   }
 
   @Override
@@ -117,13 +114,7 @@ class CompatibleCheckerVisitor implements ThriftType.TypeVisitor<Void, Void> {
 
     @Override
     public String toString() {
-      StringBuffer buffer = new StringBuffer("/");
-      Iterator<String> it = path.descendingIterator();
-      while(it.hasNext()) {
-        buffer.append(it.next());
-        buffer.append('/');
-      }
-      return buffer.toString();
+     return Strings.join(path.descendingIterator(), "/");
     }
   }
 
@@ -212,8 +203,8 @@ class CompatibleCheckerVisitor implements ThriftType.TypeVisitor<Void, Void> {
     ThriftType.StructType currentOldType = ((ThriftType.StructType) oldType);
     short oldMaxId = 0;
 
-    if (newStruct.getChildren().size() == 0) {
-      report.emptyStruct("new struct is an empty struct: " + path);
+    if (newStruct.getChildren().isEmpty()) {
+      report.emptyStruct("encountered an empty struct: " + path);
     }
 
     for (ThriftField oldField : currentOldType.getChildren()) {
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
@@ -19,7 +19,10 @@
 package org.apache.parquet.thrift.struct;
 
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
+import java.util.Iterator;
 import java.util.List;
 
 import org.apache.parquet.thrift.struct.ThriftType.BoolType;
@@ -32,7 +35,7 @@ import org.apache.parquet.thrift.struct.ThriftType.I64Type;
 import org.apache.parquet.thrift.struct.ThriftType.StringType;
 
 /**
- * A checker for thrift struct, returns compatibility report based on following rules:
+ * A checker for thrift struct to enforce its backward compatibility, returns compatibility report based on following rules:
  * 1. Should not add new REQUIRED field in new thrift struct. Adding optional field is OK
  * 2. Should not change field type for an existing field
  * 3. Should not delete existing field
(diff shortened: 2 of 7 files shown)


parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -43,24 +46,81 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
...
+    messages.add(message);
+    hasEmptyStruct = true;
+  }
+
   public List<String> getMessages() {
     return messages;
   }
+
+  public String prettyMessages() {
+    StringBuffer message = new StringBuffer();
```
isnotinvain: use StringBuilder instead. StringBuffer is thread safe (and slower)

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -43,24 +46,81 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
...
+  public void emptyStruct(String message) {
+    messages.add(message);
+    hasEmptyStruct = true;
+  }
+
   public List<String> getMessages() {
     return messages;
   }
+
+  public String prettyMessages() {
```
isnotinvain: actually, just do `parquet.Strings.join(messages, "\n")`

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -43,24 +46,81 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
...
+        '}';
+  }
 }
 
 class CompatibleCheckerVisitor implements ThriftType.TypeVisitor {
+
+  /**
+   * log the path when a incompatible field is found, so it's easier to track down the incompatible field
+   */
+  private static class FieldPath {
```
isnotinvain: can we use the FieldPath that we already have? (the one in my PR would work I think)

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -43,24 +46,81 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
...
+      path.push(fieldName);
+    }
+
+    String pop() {
+      return path.pop();
+    }
+
+    @Override
+    public String toString() {
+      StringBuffer buffer = new StringBuffer("/");
```
isnotinvain: Strings.join()

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -142,6 +204,11 @@ private boolean firstIsMoreRestirctive(ThriftField.Requirement firstReq, ThriftF
   public void visit(ThriftType.StructType newStruct) {
     ThriftType.StructType currentOldType = ((ThriftType.StructType) oldType);
     short oldMaxId = 0;
+
+    if (newStruct.getChildren().size() == 0) {
```
isnotinvain: .isEmpty() ?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -142,6 +204,11 @@ private boolean firstIsMoreRestirctive(ThriftField.Requirement firstReq, ThriftF
...
+      report.emptyStruct("new struct is an empty struct: " + path);
```
isnotinvain: this might be a little misleading, maybe "encountered an empty struct" or something like that? users don't know that this variable is newStruct so new doesn't probably mean much to them

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityRunner.java
```diff
@@ -73,6 +73,13 @@ private static void compareJson(LinkedList<String> arguments) throws IOException
       System.err.println(report.getMessages());
       System.exit(1);
     }
+
+    if (!report.hasEmptyStruct()) {
```
isnotinvain: isn't this backwards? 

`if (report.hasEmptyStruct())` ?

isnotinvain: are there tests for this file?

parquet-thrift/src/test/java/org/apache/parquet/thrift/struct/CompatibilityCheckerTest.java
```diff
@@ -107,14 +108,26 @@ public void testList() {
     verifyCompatible(ListStructV1.class, ListStructV2.class, true);
   }
 
+  @Test
+  public void testEmptyStruct() {
+    CompatibilityReport report = getCompatibilityReport(NestedEmptyStruct.class, NestedEmptyStruct.class);
+    assertTrue(report.prettyMessages().contains("new struct is an empty struct: /required_empty/"));
```
isnotinvain: can we assertEquals() instead? assertTrue doesn't provide good error messages

isnotinvain:
minor comments but overall LGTM

parquet-thrift/src/test/java/org/apache/parquet/thrift/struct/CompatibilityCheckerTest.java
tsdeng: prettyMessages contains multiple error messages, so equals may not work

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
tsdeng: Yeah, need to wait for your PR to be merged first

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
tsdeng: fixed

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
tsdeng: fixed, would you mind taking a look again? @isnotinvain

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -124,11 +124,11 @@ private static ThriftField toThriftField(String name, Field field, ThriftField.R
         break;
       case SET:
         final Field setElemField = field.getSetElemField();
-        type = new ThriftType.SetType(toThriftField(name, setElemField, requirement));
+        type = new ThriftType.SetType(toThriftField(setElemField.getName(), setElemField, requirement));
```
isnotinvain: why did this work before? we don't really read the names ever?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -52,93 +57,120 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
...
+    return "CompatibilityReport{" +
+        "isCompatible=" + isCompatible +
+        ", hasEmptyStruct=" + hasEmptyStruct +
+        ", messages=\n" + prettyMessages() +
+        '}';
+  }
 }
 
-class CompatibleCheckerVisitor implements ThriftType.TypeVisitor<Void, Void> {
+class State {
```
isnotinvain: private static final class State

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -52,93 +57,120 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
 
 class CompatibilityReport {
```
isnotinvain: {private/public} static final class CompatibilityReport?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -43,8 +48,8 @@
 public class CompatibilityChecker {
 
   public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, ThriftType.StructType newStruct) {
-    CompatibleCheckerVisitor visitor = new CompatibleCheckerVisitor(oldStruct);
-    newStruct.accept(visitor, null);
+    CompatibleCheckerVisitor visitor = new CompatibleCheckerVisitor();
+    newStruct.accept(visitor, new State(oldStruct,new FieldsPath()));
```
isnotinvain: we can make the return type of this visitor CompatibilityReport right? then you don't need a getter for it right?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -52,93 +57,120 @@ public CompatibilityReport checkCompatibility(ThriftType.StructType oldStruct, T
 
 class CompatibilityReport {
   boolean isCompatible = true;
+  boolean hasEmptyStruct = false;
```
isnotinvain: make these private if they have getters

isnotinvain:
Made some comments, LGTM overall thanks!

isnotinvain:
+1

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -32,7 +34,7 @@
 import org.apache.parquet.thrift.struct.ThriftType.StringType;
 
 /**
- * A checker for thrift struct, returns compatibility report based on following rules:
+ * A checker for thrift struct to enforce its backward compatibility, returns compatibility report based on following rules:
  * 1. Should not add new REQUIRED field in new thrift struct. Adding optional field is OK
  * 2. Should not change field type for an existing field
  * 3. Should not delete existing field
```
rdblue: I think this should be "should not delete an existing REQUIRED field". Removing an optional field is fine, right?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
```diff
@@ -32,7 +34,7 @@
 import org.apache.parquet.thrift.struct.ThriftType.StringType;
 
 /**
- * A checker for thrift struct, returns compatibility report based on following rules:
+ * A checker for thrift struct to enforce its backward compatibility, returns compatibility report based on following rules:
  * 1. Should not add new REQUIRED field in new thrift struct. Adding optional field is OK
  * 2. Should not change field type for an existing field
```
rdblue: Also shouldn't change optional to required?

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
rdblue: Nevermind, this is a forward-compatibility requirement. I don't think this is required for backward-compatibility though. Deleting an existing field will cause the code to ignore it. Things might get weird when you try to add another field with the same ID, but that's a separate issue.

rdblue:
If this isn't used anywhere, then why was it part of the public API? I don't think it would be perfectly reasonable to use this so I'm not sure I agree that we can change it. Also, the API breaking changes aren't related to PARQUET-279 so I don't understand why these need to go in at all.

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
tsdeng: We consider deleting a field as an incompatible change. This is more of a internal requirement for us. For example when you move a field from one struct to another one, it requires data consumer to explicitly deal with the situation to avoid data loss and we don't want that to happen

parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/CompatibilityChecker.java
tsdeng: "4. Should not make requirement type more restrictive for a field in new thrift struct"
addresses that

tsdeng:
@rdblue to address your concern, I separated the refactor and API breaking change to another PR

PR #195, opened 2015-05-11
PARQUET-278 : enforce non empty group on MessageType level

tsdeng:
As columnar format, parquet currently does not support empty struct/group without leaves. We should throw when constructing an empty GroupType to give a clear message.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
@@ -88,7 +88,7 @@ public class GroupType extends Type {
    */
   GroupType(Repetition repetition, String name, OriginalType originalType, List<Type> fields, ID id) {
     super(name, repetition, originalType, id);
-    if (fields.size() == 0) {
+    if (fields.isEmpty()) {
       throw new InvalidSchemaException("A group type can not be empty. Parquet does not support empty group without leaves. Empty group: " + name);
     }
     this.fields = fields;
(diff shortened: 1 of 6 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -88,6 +88,9 @@ public GroupType(Repetition repetition, String name, OriginalType originalType,
    */
   GroupType(Repetition repetition, String name, OriginalType originalType, List<Type> fields, ID id) {
     super(name, repetition, originalType, id);
+    if (fields.size() == 0) {
```
isnotinvain: isEmpty()

parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
```diff
@@ -145,4 +145,18 @@ public void testIDs() throws Exception {
...
     assertEquals(schema.toString(), schema2.toString());
   }
+
+  @Test
+  public void testEmptyGroup() {
+    try {
+      MessageType t5 = new MessageType("root1",
+          new GroupType(REQUIRED, "g1"),
+          new GroupType(REQUIRED, "g2",
+              new PrimitiveType(OPTIONAL, BINARY, "b")));
```
isnotinvain: you need to put:
`fail("this should throw")` 
in this try block -- if this never throws the test will still pass

isnotinvain:
LGTM except for the `fail()` part

PR #197, opened 2015-05-16
PARQUET-251: Binary column statistics error when reuse byte[] among rows

SinghAsDev:


SinghAsDev:
@rdblue could you take a look.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -72,13 +72,13 @@ else if (!this.isEmpty())
   }
 
   public void updateStats(Binary min_value, Binary max_value) {
-    if (min.compareTo(min_value) > 0) { min = min_value; }
-    if (max.compareTo(max_value) < 0) { max = max_value; }
+    if (min.compareTo(min_value) > 0) { min = Binary.fromByteArray(min_value.getBytes()); }
```
rdblue: Why not use `Binary.copy`?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
```diff
@@ -285,8 +285,7 @@ public void fallBackDictionaryEncodedData(ValuesWriter writer) {
     }
 
     protected static Binary copy(Binary binary) {
-      return Binary.fromByteArray(
-          Arrays.copyOf(binary.getBytes(), binary.length()));
+      return Binary.fromByteArray(binary.getBytes());
```
rdblue: I think this should be built into `Binary`. There are times when `getBytes` will copy and it doesn't need to, like when the binary is backed by a byte array slice. It would be better if each Binary implementation had its own copy method.

kostya-sh:
A possible optimization: `FromStringBinary` does not need to copy byte array in `getBytes()` as its byte array is already a copy (`String.getBytes()`) and cannot be modified outside of parquet code base.

rdblue:
I think this should implement Julien's suggestion on the JIRA issue: Binary should know the intent of the producer. If the producer doesn't intend to change the binary, then the copy behavior can change. I think this should update the factory methods, `fromByteBuffer` etc., to signal that intent. I wouldn't use "immutable" because this still violates that definition. Maybe `fromReusedByteBuffer` and `fromConstantByteBuffer`?

Don't worry about the naming of `getBytes` and `getBytesUnsafe` for now. We'll be able to rename them easily when we agree what they should be named.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
rdblue: Ah, because the method I was looking at is `DictionaryValuesWriter.copy`. I note below that this should be added to the binary API.

SinghAsDev:
Thanks for the review guys. I have updated the PR.

@kostya-sh, FromStringBinary extends ByteArrayBackedBinary and it is possible for a consumer to modify the underlying bytes if getBytes() were to return the underlying bytes. To avoid this, having getBytes() provide a copy makes sense.

kostya-sh:
@SinghAsDev, yes you are right, the consumer can modify the bytes. However given that consumer is always code in parquet-mr library and this code never modifies bytes my proposal is still valid optimization.

The original reason for `getBytes()` is to handle the case when a producer modifies the underlying bytes.

I quite like the idea to make a producer signal an intent (`fromReusedByteBuffer` and `fromConstantByteBuffer`) as it makes it very clear that copying is required to protect from a producer modifying the underlying array, not a consumer.

kostya-sh:
Also It looks like parquet-mr doesn't always puts a numeric version into file metadata.

E.g for version 1.6.0rc7, the producer string looks like `parquet-mr (build ec6f200b4943cfcbc8be5a8e53fdebf0)`. 1.6.0 also doesn't put the numeric version.

rdblue:
@kostya-sh, the commit hash behavior happens in real releases, not just RCs? If so, that makes this significantly harder.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -219,14 +219,14 @@ private void writeValue(Type type, Schema avroSchema, Object value) {
     } else if (avroType.equals(Schema.Type.UNION)) {
       writeUnion((GroupType) type, nonNullAvroSchema, value);
     } else if (avroType.equals(Schema.Type.FIXED)) {
-      recordConsumer.addBinary(Binary.fromByteArray(((GenericFixed) value).bytes()));
```
rdblue: I don't think the intent is captured correctly here. Avro objects like `Fixed` and `Utf8` will be reused if reading from an Avro file and writing to Parquet. In that case, shouldn't these methods use `Binary.fromReusedByteArray` and similar?

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
```diff
@@ -387,7 +388,7 @@ public void write(Map<String, Object> record) {
         recordConsumer.endField("mymap", index++);
 
         recordConsumer.startField("myfixed", index);
-        recordConsumer.addBinary(Binary.fromByteArray((byte[]) record.get("myfixed")));
+        recordConsumer.addBinary(Binary.fromUnmodifiedByteArray((byte[]) record.get("myfixed")));
```
rdblue: It doesn't matter much for tests, but I think this is also the wrong call.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/DataGenerator.java
```diff
@@ -107,7 +107,7 @@ public void generateData(Path outFile, Configuration configuration, ParquetPrope
           .append("float_field", 1.0f)
           .append("double_field", 2.0d)
           .append("flba_field", new String(chars))
-          .append("int96_field", Binary.fromByteArray(new byte[12]))
+          .append("int96_field", Binary.fromUnmodifiedByteArray(new byte[12]))
```
rdblue: This is correct because there is only one reference to the byte array.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -46,8 +46,8 @@ public void mergeStatisticsMinMax(Statistics stats) {
 
   @Override
   public void setMinMaxFromBytes(byte[] minBytes, byte[] maxBytes) {
-    max = Binary.fromByteArray(maxBytes);
-    min = Binary.fromByteArray(minBytes);
+    max = Binary.fromUnmodifiedByteArray(maxBytes);
+    min = Binary.fromUnmodifiedByteArray(minBytes);
```
rdblue: Do we know that this is always called with constant byte buffers and not reused byte buffers?

parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
```diff
@@ -82,11 +82,11 @@ public String memUsageString(String prefix) {
   @Override
   public void writeBytes(Binary v) {
     int i = 0;
-    byte[] vb = v.getBytes();
+    byte[] vb = v.getBytesUnsafe();
     int length = previous.length < vb.length ? previous.length : vb.length;
     for(i = 0; (i < length) && (previous[i] == vb[i]); i++);
     prefixLengthWriter.writeInteger(i);
-    suffixWriter.writeBytes(Binary.fromByteArray(vb, i, vb.length - i));
+    suffixWriter.writeBytes(Binary.fromUnmodifiedByteArray(vb, i, vb.length - i));
```
rdblue: This looks odd. The contract for unsafe access to the bytes backing another Binary is to use them immediately and forget about them so it doesn't matter if they change. But in this case you create a constant Binary that signals the bytes won't change. So this is taking something that might change and passing on something that won't.

What we need here is the ability to get a slice of the original v that inherits whether it is reused or constant:

```java
Binary slice(int start, int length);
```

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -71,7 +79,19 @@ public boolean equals(Object obj) {
 
   @Override
   public String toString() {
-    return "Binary{" + length() + " bytes, " + Arrays.toString(getBytes()) + "}";
+    return "Binary{" + length() + " bytes, " + Arrays.toString(getBytesUnsafe()) + "}";
```
rdblue: This should also flag whether the binary is reused.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -33,10 +33,12 @@
 
 abstract public class Binary implements Comparable<Binary>, Serializable {
 
+  private boolean isUnmodifiable;
```
rdblue: I think it would be easier to reason about this if this were named `isReused`. That immediately signals that the content might change. `isUnmodifiable` signals that we shouldn't do things that modify the content, but that's not really correct because it might be the caller that has wrapped a byte array that changes the content.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -71,7 +79,19 @@ public boolean equals(Object obj) {
...
   public String toString() {
-    return "Binary{" + length() + " bytes, " + Arrays.toString(getBytes()) + "}";
+    return "Binary{" + length() + " bytes, " + Arrays.toString(getBytesUnsafe()) + "}";
+  }
+
+  public Binary copy() {
+    if (isUnmodifiable) {
+      return this;
+    } else {
+      return Binary.fromReusedByteArray(getBytes());
```
rdblue: Because this allocates a new, safe copy of the bytes and forgets about it, this should instead call the unmodifiable wrapper method.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -71,7 +79,19 @@ public boolean equals(Object obj) {
...
+    }
+  }
+
+  protected void setUnmodifiable(boolean isUnmodifiable) {
+    this.isUnmodifiable = isUnmodifiable;
```
rdblue: Why not update the constructor for the abstract class instead?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -110,6 +131,11 @@ public void writeTo(OutputStream out) throws IOException {
     }
 
     @Override
+    public byte[] getBytesUnsafe() {
+      return getBytes();
```
rdblue: This should add a comment to explain why this always copies the bytes: because the backing array is larger than the slice used for this Binary.

rdblue: Another option is to return the backing bytes and expose the start and length values.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -148,7 +174,7 @@ public void writeTo(DataOutput out) throws IOException {
 
   private static class FromStringBinary extends ByteArrayBackedBinary {
     public FromStringBinary(byte[] value) {
-      super(value);
+      super(value, true);
```
rdblue: This isn't a huge concern because this class is private, but if this is going to statically set whether or not the binary is reused, then it should really do the conversion to a byte array in the constructor so there is no chance for the byte array to be leaked in the future.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -157,15 +183,26 @@ public String toString() {
...
 
-  public static Binary fromByteArray(final byte[] value, final int offset, final int length) {
-    return new ByteArraySliceBackedBinary(value, offset, length);
+  private static Binary fromByteArray(final byte[] value, final int offset, final int length,
+                                      final boolean isUnmodifiable) {
+    return new ByteArraySliceBackedBinary(value, offset, length, isUnmodifiable);
+  }
+
+  public static Binary fromReusedByteArray(final byte[] value, final int offset, final int length) {
+    return fromByteArray(value, offset, length, false);
```
rdblue: Why is there a fromByteArray method? Why not call the constructor directly?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
rdblue: Also, add a `isReused` method. It would be good to be able to check.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -262,12 +313,17 @@ public void writeTo(OutputStream out) throws IOException {
     }
 
     @Override
+    public byte[] getBytesUnsafe() {
+      return getBytes();
```
rdblue: If the buffer isn't reused, then it is safe to cache a copy of the bytes and use it each time unsafe is called rather than copying all the time.

SinghAsDev:
@kostya-sh I agree with you that FromStringBytes always get a copy of bytes. However, the consumer can still modify it. As you mentioned, as of now consumers are not modifying it, but that does not mean it won't be modified in future. May be by mistake. If the consumer does not need to modify the bytes() then it should be using getBytesUnsafe() instead of getBytes(). Let me know if I am missing updating of getBytes() to getBytesUnsafe somewhere in the code.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -341,8 +397,16 @@ private void readObjectNoData() throws ObjectStreamException {
...
-    return new ByteBufferBackedBinary(value);
+  private static Binary fromByteBuffer(final ByteBuffer value, final boolean isUnmodifiable) {
+    return new ByteBufferBackedBinary(value, isUnmodifiable);
+  }
+
+  public static Binary fromReusedByteBuffer(final ByteBuffer value) {
+    return fromByteBuffer(value, false);
+  }
+
+  public static Binary fromUnmodifiedByteBuffer(final ByteBuffer value) {
```
rdblue: I don't think `fromUnmodifiedByteBuffer` is clear enough. Unmodified makes it sound like it is not modified when it is created, but might be afterward. I think const or constant gets closer, but what we want is to signal that this will not be reused or rewritten. Immutable is too strong because it has a formal definition in Java.

What about using `fromReusableByteBuffer` and `fromConstantByteBuffer`?

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
```diff
@@ -384,6 +384,23 @@ public void testBinaryMinMax() {
   }
 
   @Test
+  public void testBinaryMinMaxForReusedBackingByteArray() {
+    BinaryStatistics stats = new BinaryStatistics();
+
+    byte[] bytes = new byte[] { 10 };
+    stats.updateStats(Binary.fromReusedByteArray(bytes));
```
rdblue: You should be able to create the Binary just once, change the backing bytes, and use it in each call to `updateStats`.

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/TestDictionary.java
```diff
@@ -512,12 +513,13 @@ private void writeRepeated(int COUNT, ValuesWriter cw, String prefix) {
     }
   }
 
-  private void writeRepeatedWithReuse(int COUNT, ValuesWriter cw, String prefix) {
-    Binary reused = Binary.fromString(prefix + "0");
+  private void writeRepeatedWithReuse(int COUNT, ValuesWriter cw,
+                                      String prefix) throws UnsupportedEncodingException {
+    byte[] reused = (prefix + "0").getBytes("UTF-8");
```
rdblue: This should still be a Binary, just a reused one backed by a byte array.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -38,21 +39,47 @@
  * no filtering will be performed.
  */
 public class RowGroupFilter implements Visitor<List<BlockMetaData>> {
+  private static final int STATISTICS_FIXED_VERSION = 161; // 1.6.1
   private final List<BlockMetaData> blocks;
   private final MessageType schema;
+  private final Boolean skipStatisticsChecks;
 
-  public static List<BlockMetaData> filterRowGroups(Filter filter, List<BlockMetaData> blocks, MessageType schema) {
+  public static List<BlockMetaData> filterRowGroups(Filter filter, ParquetMetadata footer,
```
rdblue: @isnotinvain, is this a public API? It looks like it to me.

If this is a public API, then we need to keep a version of this method that takes a `List<BlockMetaData>` for compatibility. You can deprecate it though.

parquet-hive/parquet-hive-storage-handler/src/main/java/org/apache/hadoop/hive/serde2/objectinspector/primitive/ParquetStringInspector.java
```diff
@@ -77,7 +77,8 @@ public String getPrimitiveJavaObject(final Object o) {
 
   @Override
   public Object set(final Object o, final Text text) {
-    return new BinaryWritable(text == null ? null : Binary.fromByteArray(text.getBytes()));
+    return new BinaryWritable(text == null ? null : Binary.fromUnmodifiedByteArray(text.getBytes
```
rdblue: If this is on the write side, keep in mind that texts can be reused.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -318,7 +318,7 @@ final void writeRawValue(Object value) {
     @Override
     final void writeRawValue(Object value) {
       ByteString byteString = (ByteString) value;
-      Binary binary = Binary.fromByteArray(byteString.toByteArray());
+      Binary binary = Binary.fromUnmodifiedByteArray(byteString.toByteArray());
```
rdblue: Is ByteString immutable? If it isn't, then I think this should be `fromReusedByteArray`.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -38,21 +39,47 @@
...
+    if (app.equalsIgnoreCase("parquet-mr")) {
+      return Integer.parseInt(version) < STATISTICS_FIXED_VERSION;
+    }
+
+    return true;
+  }
+
   @Override
   public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredicateCompat) {
+    if (skipStatisticsChecks) {
```
rdblue: It looks like this would always skip statistics, but there are a lot of cases where statistics are still valid. I think they only need to be skipped when the underlying column type is binary, fixed, and int96 because the other objects are immutable (string) or primitives. In that case, the checks here should only ignore statistics for the potentially corrupted columns.

rdblue:
@SinghAsDev, this only updates row group filtering. Are we not filtering pages based on this?

It seems like it would be safer to filter out bad statistics in the [metadata converter](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L237) when converting from the Thrift footer to Parquet MR metadata objects.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
SinghAsDev: That is a good point @rdblue. I think you are correct and these should instead be fromReusedByteBuffer. However, in the case of UTF8, we can still use fromUnmodifiedByteBuffer as it passes copy of the buffer, utf8.getBytes().

SinghAsDev: Actually on digging a bit more, looks like Utf8 can be constructed by passing a string, in which case it keeps a copy of bytes, or passing byte[], in which case it keeps the original byte[]. I think this should be fixed on avro side to have consistent behavior for getBytes(). Anyways, for now I think it would be better and safe to use fromResusedByteArray for Utf8 as well.

SinghAsDev: Few more questions. Reading from Avro file is not controlled by parquet project, right? I could not find code related to reading Avro file here. If I am missing something, please point me to the right place. Assuming that reading is not controlled by parquet project, I think it would be safe to assume that objects are being re-used by Avro file reader. With this thought I am changing even the fromUnmodifiedByteBuffer to fromReusedByteBuffer.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
SinghAsDev: Yes, at least in the current version of code this is true. Let me know if you think otherwise.

parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
SinghAsDev: I thought the contract for unsafe access to the bytes backing another Binary is to never update it, as it is owned by some other object. If that is true, then it should be up to the consumer what it wants to do with it as long as it never updates it.
On the same note, I have used fromUnmodifiedByteArray here as the producer is clearly not modifying the backing bytes.
However, I do understand your concern that this particular scenario acts as a producer of binary backed by bytes owned by some other entity and still claims that bytes won't be changed.
Two possible solutions, either change getBytesUnsafe() to getBytes(), or change fromUnmodifiedByteArray to inherit reusable property from original v, as suggested by you. Former will require an extra copy, which I think is not required here. I will update the code as suggested by you to inherit the reusable property from original v. Thanks for pointing this out.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: As different implementations of Binary are used via Binary interface, exposing start and length values will require Binary interface to be aware of it. Correct me if I am not getting your point.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: I am probably missing something here, but I did not understand this.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Ahh.. this was left around while refactoring. Removed now.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Caching will take memory. It makes sense to cache it only if it is being called again and again. If it is being called again and again, then the consumer must take care of caching. That's my understanding. Let me know if you disagree.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Sounds good to me.

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
SinghAsDev: Nice :)

parquet-hive/parquet-hive-storage-handler/src/main/java/org/apache/hadoop/hive/serde2/objectinspector/primitive/ParquetStringInspector.java
SinghAsDev: Do you suggest we use fromReusedBytes() to be safe?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
SinghAsDev: toByteArray return copy of bytes, so we should be safe here.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
rdblue: You're right: there is no code in Parquet that does the reading from Avro as I describe. I'm thinking of the common use case of copying from Avro to Parquet for a conversion. Avro will reuse objects to avoid creation and garbage collection overheads, so we need to assume that the objects passed in are reused. I'm all for making this kind of improvement to Avro, too, but right now I think the right thing to do is assume that the objects are reused and wrap them accordingly.

Other object models should use similar logic. On the write side, we don't know whether the objects are being reused, so we should assume that they are unless they claim to be Immutable.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
rdblue: I think this should either use the "reused" call or require the caller wrap the byte arrays. That way there isn't a way for future callers to use this unsafely.

parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
rdblue: The contract for getBytesUnsafe is to use them immediately so it doesn't matter if they change. If you pass them on, then that's not necessarily an immediate use. The consumer certainly shouldn't be changing the bytes in a Binary, but we mostly care about the consumer's intended use.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
rdblue: Yes. That option would definitely complicate things. But it would allow us to do fewer copies. Probably something we should do if we find out we're making too many copies in the future.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
rdblue: String to byte[] takes an immutable object and returns a mutable one. Then you create an unmodified variant of that mutable byte[]. It would be a better API to have the constructor here take a String rather than byte[]. That way, it cannot be passed a byte array that is referenced -- and possibly changed -- elsewhere.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
rdblue: Maybe. There are quite a few methods that call this internally, like `hashCode` and `equals`. If we are hiding the fact that a copy is made internally, then the caller doesn't know to cache anything.

But, the solution is probably to improve the internal implementations. For example, `hashCode` might be improved to avoid the need for a copy.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Got it. To do that I will have to call getBytes while invoking super, which takes bytes. .getBytes("UTF-8") needs exception handling which can not be done in the super. Let me know if you know of a way to work around this. As you mentioned the class is private, so this is not a very big concern.

rdblue: The conversion should be done in a static method. Then you can add exception handling to that method but still use it in the call to super.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Makes sense!

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Cool, Thanks!

kostya-sh:
Update on extracting version from file meta data. 

There is a bug in parquet.Version class, it doesn't read parquet version properly. It might work sometimes (when parquet-column.jar is the first jar in the classpath for example), but generally it is broken. I've create a JIRA: https://issues.apache.org/jira/browse/PARQUET-297

rdblue:
Thanks @kostya-sh!

SinghAsDev:
@rdblue, I have addressed your comments. Take a look once you get a chance. This PR depends on @kostya-sh PARQUET-291. If you decide to build and run tests, you will have to pull in changes from PARQUET-291.

kostya-sh:
@SinghAsDev, I guess you meant PARQUET-297 (https://github.com/apache/parquet-mr/pull/208)

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -69,6 +69,7 @@
 
 public class ParquetMetadataConverter {
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
+  private static final int STATISTICS_FIXED_VERSION = 161; // 1.6.1
```
kostya-sh: This should probably be 180

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -234,18 +237,38 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
     return Encoding.valueOf(encoding.name());
   }
 
-  public static Statistics toParquetStatistics(org.apache.parquet.column.statistics.Statistics statistics) {
+  public static Statistics toParquetStatistics(
+      String createdBy, org.apache.parquet.column.statistics.Statistics statistics) {
```
kostya-sh: I don't think you can change signature of this public method, as it breaks backwards compatibility

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -234,18 +237,38 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
...
+    final String app = versionTokens[0];
+    final String version = versionTokens[2].substring(0, 5).replaceAll("\\.", "");
+
+    if (app.equalsIgnoreCase("parquet-mr")) {
+      return Integer.parseInt(version) < STATISTICS_FIXED_VERSION;
+    }
+
+    return true;
+  }
+
```
kostya-sh: I would implement it slightly differently. In pseudo code:
```
if (!createdBy.startswith("parquet-mr")) {
  return false; // assume that other implementations got it right
}
return (version can be parsed) && (version >= STATISTICS_FIXED_VERSION)
```
Also make sure that you catch exceptions from `Integer.parseInt`

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -408,14 +408,14 @@ private void writeValue(Type type, Schema avroSchema, Object value) {
...
+      recordConsumer.addBinary(Binary.fromReusedByteArray(((GenericFixed) value).bytes()));
     }
   }
 
   private Binary fromAvroString(Object value) {
     if (value instanceof Utf8) {
       Utf8 utf8 = (Utf8) value;
-      return Binary.fromByteArray(utf8.getBytes(), 0, utf8.getByteLength());
+      return Binary.fromConstantByteArray(utf8.getBytes(), 0, utf8.getByteLength());
```
kostya-sh: `Utf8` is not immutable (unlike `String`). Should use `fromReusedByteArray` here.

SinghAsDev:
@kostya-sh thanks for the review, addressed your concerns.

parquet-column/src/main/java/org/apache/parquet/io/RecordConsumerLoggingWrapper.java
```diff
@@ -112,7 +112,7 @@ public void addBoolean(boolean value) {
      */
     @Override
     public void addBinary(Binary value) {
-      if (DEBUG) log(Arrays.toString(value.getBytes()));
+      if (DEBUG) log(Arrays.toString(value.getBytesUnsafe()));
```
isnotinvain: getBytesUnsafe should always pay attention to the length + offset right? Should this be copy()?

rdblue:
@SinghAsDev can you fix the test failures?

parquet-column/src/main/java/org/apache/parquet/io/RecordConsumerLoggingWrapper.java
rdblue: I think this is correct. There's no need to copy here unless the backing bytes have an offset and length that doesn't match the byte array's bounds. But in that case, getBytesUnsafe will already make a copy.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -147,8 +190,19 @@ public void writeTo(DataOutput out) throws IOException {
   }
 
   private static class FromStringBinary extends ByteArrayBackedBinary {
-    public FromStringBinary(byte[] value) {
-      super(value);
+    public FromStringBinary(String value) {
+      super(getBackingBytes(value), true);
+    }
+
+    private static byte[] getBackingBytes(String value) {
```
rdblue: Minor: This doesn't really return the backing bytes for the string, it returns the string encoded as UTF8 bytes. It would be better to name the method something like `encodeUtf8`. But it's private, so this isn't a big deal.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -71,7 +81,24 @@ public boolean equals(Object obj) {
...
+
+  public Binary copy() {
+    if (isReused) {
+      return Binary.fromConstantByteArray(getBytes());
+    } else {
+      return this;
+    }
+  }
+
+  public boolean isReused() {
```
isnotinvain: can we add comments explaining what it means to be reused?
There's an implicit contract here right? This is for signaling who "owns" the backing byte array right?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -147,8 +190,19 @@ public void writeTo(DataOutput out) throws IOException {
...
-    public FromStringBinary(byte[] value) {
-      super(value);
+    public FromStringBinary(String value) {
+      super(getBackingBytes(value), true);
+    }
+
+    private static byte[] getBackingBytes(String value) {
+      byte[] bytes;
+      try {
+        bytes = value.getBytes("UTF-8");
```
rdblue: Nit: You could just return from here and not have the `bytes` variable.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -48,6 +50,14 @@ private Binary() { }
 
   abstract public byte[] getBytes();
 
+  /**
+   * Variant of getBytes() that avoids copying backing data structure
+   * @return backing byte[] if possible, else returns result of getBytes()
+   */
+  abstract public byte[] getBytesUnsafe();
```
isnotinvain: maybe add to the docs that this will always be a byte[] of the correct size, with an offset of 0. It just might be a reference to a byte array that you aren't supposed to hold on to or modify unless you know what you're doing. right?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -258,16 +338,27 @@ public void writeTo(OutputStream out) throws IOException {
 
       value.mark();
       value.get(bytes).reset();
+      cachedBytes = bytes;
```
rdblue: The bytes can only be cached if the backing buffer isn't reused. If it is reused, then the backing buffer might change and cachedBytes could be stale.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -108,7 +107,8 @@ public void test() throws Exception {
           assertEquals(1.0f, group.getFloat("float_field", 0), 0.001);
           assertEquals(2.0d, group.getDouble("double_field", 0), 0.001);
           assertEquals("foo", group.getBinary("flba_field", 0).toStringUsingUTF8());
-          assertEquals(Binary.fromByteArray(new byte[12]), group.getInt96("int96_field", 0));
+          assertEquals(Binary.fromConstantByteArray(new byte[12]), group.getInt96("int96_field",
+              0));
```
rdblue: Nit: it's better to insert new lines further up the syntax tree. For example, after the first argument to assertEquals rather than in the call to getInt96.

parquet-pig/src/main/java/org/apache/parquet/pig/TupleWriteSupport.java
```diff
@@ -172,7 +172,7 @@ private void writeValue(Type type, FieldSchema pigType, Tuple t, int i) {
           } else {
             throw new UnsupportedOperationException("can not convert from " + DataType.findTypeName(pigType.type) + " to BINARY ");
           }
-          recordConsumer.addBinary(Binary.fromByteArray(bytes));
+          recordConsumer.addBinary(Binary.fromConstantByteArray(bytes));
```
rdblue: I think this is correct for the String case, but how do we know that the DataByteArray case will not be a reused object? You could always change bytes to a Binary and wrap the byte array inside the if blocks.

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoWriteSupportTest.java
```diff
@@ -50,7 +50,7 @@ public void testSimplestMessage() throws Exception {
 
     inOrder.verify(readConsumerMock).startMessage();
     inOrder.verify(readConsumerMock).startField("one", 0);
-    inOrder.verify(readConsumerMock).addBinary(Binary.fromByteArray("oneValue".getBytes()));
+    inOrder.verify(readConsumerMock).addBinary(Binary.fromConstantByteArray("oneValue".getBytes()));
```
rdblue: Could this be fromString?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -318,7 +318,7 @@ final void writeRawValue(Object value) {
     @Override
     final void writeRawValue(Object value) {
       ByteString byteString = (ByteString) value;
-      Binary binary = Binary.fromByteArray(byteString.toByteArray());
+      Binary binary = Binary.fromConstantByteArray(byteString.toByteArray());
```
rdblue: How do you know this won't be reused?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -234,18 +237,53 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
...
     }
     return stats;
   }
 
+  private static boolean shouldIgnoreStatistics(String createdBy) {
+    if (!createdBy.startsWith("parquet-mr")) {
+      return false; // Assume that other implementations got the statistics correct
+    }
+
+    return parseVersion(createdBy) < STATISTICS_FIXED_VERSION;
```
isnotinvain: is this going to work, even with funky version numbers? Don't we need to compare using something that is aware of the "major.minor.patch" syntax? eg: 1.8.0 is > 1.7.11 but 180 is not > 1711

What about versions with stuff like "rc" in them and so on?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
isnotinvain: +1

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -648,7 +689,8 @@ public void writeDataPageHeader(
       org.apache.parquet.column.Encoding dlEncoding,
       org.apache.parquet.column.Encoding valuesEncoding,
       OutputStream to) throws IOException {
-    writePageHeader(newDataPageHeader(uncompressedSize,
+    writePageHeader(newDataPageHeader(createdBy,
```
rdblue: It looks like these changes apply to the write side, but we need to ignore statistics when reading.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -234,18 +237,53 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
...
+
+  private static int parseVersion(String createdBy) {
+    final String[] versionTokens = createdBy.split(" ");
+
+    if (versionTokens.length < 3) {
+      return -1; // Ignore statistics
+    }
+
+    try {
+      return Integer.parseInt(versionTokens[2].substring(0, 5).replaceAll("\\.", ""));
```
rdblue: This parsing seems pretty brittle and doesn't account for changes in the format or for versions like 1.6.0rc2. What about using a regex (with tests) instead?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -234,18 +237,53 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
...
+  }
+
+  public static Statistics toParquetStatistics(
+      String createdBy, org.apache.parquet.column.statistics.Statistics statistics) {
     Statistics stats = new Statistics();
     if (!statistics.isEmpty()) {
       stats.setNull_count(statistics.getNumNulls());
-      if(statistics.hasNonNullValue()) {
+      if (statistics.hasNonNullValue() &&
+          createdBy != null &&
```
rdblue: Why not put the createdBy check in shouldIgnoreStatistics?

rdblue:
@SinghAsDev, this is getting closer. There are a couple things left to do:
* Run benchmarks to make sure the Binary changes aren't slowing things down too much
* Add tests for the new Binary behavior. For example, slice is completely new and needs to be tested and we should validate the copy behavior of different backing implementations.
* The statistics need to be ignored on the read side, not the write side
* We also need a test to verify the statistics now written are correct, and one to verify that stats are correctly ignored with version 1.6.0.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: That's correct. Added comment.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: That is correct. Added bit more info to the docs.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: Good point Ryan! Fixed.

parquet-pig/src/main/java/org/apache/parquet/pig/TupleWriteSupport.java
SinghAsDev: I think you are correct, its better to mark it reused.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
SinghAsDev: I do not think it is reused right now. But, the fact that producer can potentially change it makes "reused" version suitable to be used here.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
SinghAsDev: It will work for the case you are referring to as it takes first five characters of version string. However, the current logic is brittle. Will improve it.

isnotinvain: Maybe we can use this:
https://github.com/jeluard/semantic-versioning/blob/master/api/src/main/java/org/semver/Version.java

We're already using the enforcer from this project, and here is a Version parser that implements `Comparable<Version>`, seems like what we need.

isnotinvain: @SinghAsDev I added a test-time dependency on this library in #218, we could make it a runtime dep, or just grab only the Version file

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
kostya-sh: From ByteString javadocs (https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString): ```Like String, the contents of a ByteString can never be observed to change, not even in the presence of a data race or incorrect API usage in the client code.```

So using `fromConstantByteArray` is correct here.

SinghAsDev: Makes sense. Thanks for pointing this out @kostya-sh.

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoWriteSupportTest.java
SinghAsDev: Yes. Updated.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
SinghAsDev: @isnotinvain thanks for pointing to this. However, I have made changes to parse the version string without using the lib, I think it should be in good shape right now. Can you take a look at it? If you think we should still use the lib you mentioned, I can make the change. It just looked like an over kill, as all we care about is if a version is less than 1.8.0 or not.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
SinghAsDev: I don't know how I ended up making write side changes :). I have added the checks to read path and have reverted write side changes.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
SinghAsDev: Done!

SinghAsDev:
@rdblue, @isnotinvain and @kostya-sh I have updated the PR to address your comments. Let me know if I am still missing something.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -48,6 +50,16 @@ private Binary() { }
 
   abstract public byte[] getBytes();
 
+  /**
+   * Variant of getBytes() that avoids copying backing data structure by returning
+   * backing byte[] of the Binary. Do not modify backing byte[] unless you know what
+   * you are doing.
+   * @return backing byte[] of correct size, with an offset of 0, if possible, else returns result of getBytes()
+   */
+  abstract public byte[] getBytesUnsafe();
```
SinghAsDev: @julienledem you had some concern regarding "getBytesUnsafe" to which @rdblue posted an answer. Are you OK with this now?

SinghAsDev:
the build error is not related, "GET request of: com/twitter/scrooge-generator_2.10/3.17.0/scrooge-generator_2.10-3.17.0.jar from central failed: Connection reset -> [Help 1]"

parquet-column/pom.xml
```diff
@@ -77,6 +77,11 @@
       <version>1.3.149</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
isnotinvain: why does this need to be added? isn't this inherited from the parent pom?
Shouldn't this have `<scope>test</scope>` ? Or is this being added to add it as a requirement at runtime (not only for tests)?

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -71,7 +83,28 @@ public boolean equals(Object obj) {
...
+      return this;
+    }
+  }
+
+  /**
+   * Signals if backing bytes are owned, and can be modified, by producer of the Binary
+   * @return if backing bytes are held on by producer of the Binary
+   */
+  public boolean isReused() {
+    return isReused;
```
isnotinvain: I'd like there to be a more clear name than `isReused` but I can't think of one... maybe `isFinal` or `backingBytesReassignable` or something?

parquet-column/pom.xml
SinghAsDev: My bad. This is not required. Removed.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: How about isBackingBytesReused?

isnotinvain: yeah, I like that, thanks!

SinghAsDev: Done!

parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
```diff
@@ -0,0 +1,110 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+public class TestBinary {
+  @Test
+  public void testSliceForByteArrayBackedBinary() {
+    Binary binary = Binary.fromConstantByteArray("test-123".getBytes());
```
isnotinvain: maybe be explicit about `getBytes("UTF-8")` here and elsewhere

parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
```diff
@@ -0,0 +1,110 @@
...
+    Assert.assertArrayEquals("123".getBytes(), binary.slice(5, 3).getBytesUnsafe());
+  }
+
+  @Test
+  public void testSliceForByteArraySliceBackedBinary() {
```
isnotinvain: how about make some tests that all Binary's must pass, and then run all the Binary instances through them?

eg:

```
interface BinaryFactory { Binary get(String s); }

private void testSlice(BinaryFactory bf) {
  Binary b = bf.get("test-123");
  assertArrayEquals("123".getBytes("UTF-8"), b.slice(5, 3).getBytesUnsafe());
}

private void testCopy(BinaryFactory bf) {
 ...
}

private void testFoo(BinaryFactory bf) {
 ...
}

private void testBinary(BinaryFactory bf) {
  testSlice(bf);
  testCopy(bf);
  testFoo(bf);
}

@Test
pubic void testConstantBinary() {
  testBinary(new BinaryFactory { @Override Binary get(String s) { return Binary.fromConstantByteArray(s.getBytes("UTF-8")) } });
}

@Test
pubic void testBufferBinary() {
  testBinary(new BinaryFactory { @Override Binary get(String s) { return Binary.fromBuffer(...) } });
}

etc.
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -259,6 +274,48 @@ public static Statistics toParquetStatistics(org.apache.parquet.column.statistic
     return stats;
   }
 
+  private static boolean shouldIgnoreStatistics(String createdBy) {
+    if (createdBy == null || !createdBy.startsWith("parquet-mr")) {
```
isnotinvain: did all the buggy parquet-mr versions correctly write the createdBy field?

isnotinvain: they didn't, right? I think we need to handle the case where the createdBy is something like "parquet-mr <sha>" (missing a version number)

isnotinvain: see: https://issues.apache.org/jira/browse/PARQUET-297

parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
rdblue: +1

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Agreed. When a version of the writer can't be determined, this should consider the min/max stats invalid.

SinghAsDev: Unless I am missing something, this is what current code will do, i.e., ignore the stats when a version of "parquet-mr" writer can't be determined.

isnotinvain:
I have some time today and want to get this merged ASAP, so I'm going to send a PR against this PR with some of the tests I mentioned above.

isnotinvain:
https://github.com/SinghAsDev/parquet-mr/pull/2/files

SinghAsDev:
@isnotinvain thanks for the changes! Read your message when I was about to push mine :). Let me know if still something remains unaddressed.

isnotinvain:
@SinghAsDev I added some more tests and pulled some of the version parsing into separate files here:
https://github.com/SinghAsDev/parquet-mr/pull/3

Let me know what you think!

After that, I think all we need is to merge master into your branch, resolve any conflicts, and we should be good to go!

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -69,6 +71,9 @@
 
 public class ParquetMetadataConverter {
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
+  private static final int STATISTICS_FIXED_VERSION_MAJOR = 1;
+  private static final int STATISTICS_FIXED_VERSION_MINOR = 8;
+  private static final int STATISTICS_FIXED_VERSION_BUILD = 0;
```
rdblue: Nit: according to semver, this should be "patch"

rdblue:
Alex's patch looks like the right direction. @SinghAsDev, could you merge his patch and coordinate making sure the comments I made on it are incorporated? (Assuming Alex agrees)

After that I think we're about ready to go.

isnotinvain:
@SinghAsDev can you merge master? I think we're good to go once the tests pass

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -425,6 +436,27 @@ public void testConvertToThriftStatistics() throws Exception {
   }
 
   @Test
+  public void testStatisticsGettingIgnoredFor160Version() throws Exception {
+    long[] longArray = new long[] {39L, 99L, 12L, 1000L, 65L, 542L, 2533461316L, -253346131996L, Long.MAX_VALUE, Long.MIN_VALUE};
+    LongStatistics parquetMRstats = new LongStatistics();
```
rdblue: I have a couple of questions about this:
* Why is this validating that Long stats are no longer used? Shouldn't long stats be okay?
* Why are the stats all 0? How does the reader know not to use the stats?

isnotinvain: should be nulls right?

isnotinvain:
Oh also: I noticed that some files have created_by == "parquet-mr" (that's it, nothing else) as well as "parquet-mr (build abcd" (no version constant, no version value). Both of those will fall under "unparsable" the way this is written, which means we will correctly ignore them. But we could also handle parsing these instead of letting them fall into the unparsable category

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
rdblue: The stats are 0 for LongStatistics because the method returns a primitive, I think. But it scares me that the API allows an empty stats object to be used like this. At a minimum, if the stats aren't set those should through an Exception. I'd prefer not to deserialize a stats object at all and return null. Is there something preventing that?

This should also not be passing tests because LongStatistics shouldn't be filtered. We should clone this and add a test to verify that BinaryStatistics are actually filtered.

isnotinvain: i think there's a method on stats like isEmpty() that has to be inspected before using the object IIRC, let me look....

isnotinvain: yeah:
https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java#L235

I don't remember if we use null objects here or empty ones when there are no stats

rdblue: I think we use empty ones for some reason. Probably fine for now, but I'll file a bug to fix it. That's a misleading API and could lead to incorrect results with predicate push-down.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -566,7 +581,10 @@ public ColumnChunkPageReader readAllPages() throws IOException {
                     this.readAsBytesInput(compressedPageSize),
                     dataHeaderV1.getNum_values(),
                     uncompressedPageSize,
-                    fromParquetStatistics(dataHeaderV1.getStatistics(), descriptor.col.getType()),
+                    fromParquetStatistics(
+                        fileMetaData.getCreatedBy(),
```
isnotinvain: this isn't guarded by a check for fileMetaData being null, but it should be, right? (use of the deprecated constructor).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -586,7 +604,10 @@ public ColumnChunkPageReader readAllPages() throws IOException {
                     ParquetMetadataConverter.getEncoding(dataHeaderV2.getEncoding()),
                     this.readAsBytesInput(dataSize),
                     uncompressedPageSize,
-                    fromParquetStatistics(dataHeaderV2.getStatistics(), descriptor.col.getType()),
+                    fromParquetStatistics(
+                        fileMetaData.getCreatedBy(),
```
isnotinvain: here too?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -51,6 +51,7 @@
   private final Iterator<Footer> footersIterator;
   private final Filter filter;
 
+  private String createdBy;
```
isnotinvain: does this have to be stored in an instance variable? seems like it could be a local variable?

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -46,8 +46,8 @@ public void mergeStatisticsMinMax(Statistics stats) {
 
   @Override
   public void setMinMaxFromBytes(byte[] minBytes, byte[] maxBytes) {
-    max = Binary.fromByteArray(maxBytes);
-    min = Binary.fromByteArray(minBytes);
+    max = Binary.fromReusedByteArray(maxBytes);
+    min = Binary.fromReusedByteArray(minBytes);
```
isnotinvain: I should have asked this earlier, but, should there be a defensive copy here too?

We are assuming that minBytes / maxBytes will be re-used, so what's the lifecycle assumption being made? That these won't be modified until after the last call to updateStats is made? Is that a valid assumption? Seems like this should actually do a copy, right?

isnotinvain: oh, this is only used in the read path right? so the backing bytes should never change, but we use .fromReusedByteArray so that copy() would do the right thing if it were ever used. Is that correct?

isnotinvain:
OK, I made some tiny changes here: https://github.com/SinghAsDev/parquet-mr/pull/4
If that gets merged, I'm +1 on this PR.

@SinghAsDev @rdblue think this this is ready?

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
SinghAsDev: I think it could be debated from either ways. A producer might want to just update the max/ min bytes, instead of explicitly calling updateStats by creating a new Binary every time it has to update. In this case, if we add a comment stating the behavior explicitly will be nice. I am doing that for now. If you disagree, then I will change it copy instead.

rdblue:
I'll review when the PR gets merged. @SinghAsDev, please ping me when it is ready,

SinghAsDev:
@rdblue should be ready for your review now.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
rdblue: We don't have control over where the byte arrays come from, so the safe thing is to use the reused form. If we changed the API and could make guarantees about it, then I think we could do otherwise.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -341,16 +440,16 @@ private void readObjectNoData() throws ObjectStreamException {
 
   }
 
-  public static Binary fromByteBuffer(final ByteBuffer value) {
```
rdblue: This is a breaking change. I just tracked down why semver didn't catch it, but we need to add these methods back and deprecate them.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -79,10 +112,20 @@ public String toString() {
     private final int offset;
     private final int length;
 
+    @Deprecated
```
isnotinvain: This is a constructor in a private class, so I think you can safely delete the deprecated constructor.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -147,8 +205,27 @@ public void writeTo(DataOutput out) throws IOException {
...
 
   private static class FromStringBinary extends ByteArrayBackedBinary {
+    public FromStringBinary(String value) {
+      // reused is false, because we do not
+      // hold on to the underlying bytes,
+      // and nobody else has a handle to them
+      super(encodeUTF8(value), false);
+    }
+
+    @Deprecated
```
isnotinvain: same, you can delete this constructor too

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -157,15 +234,38 @@ public String toString() {
...
+   */
   public static Binary fromByteArray(final byte[] value, final int offset, final int length) {
-    return new ByteArraySliceBackedBinary(value, offset, length);
+    return fromReusedByteArray(value, offset, length); // Assume producer intends to reuse byte[]
   }
 
   private static class ByteArrayBackedBinary extends Binary {
     private final byte[] value;
 
+    @Deprecated
```
isnotinvain: can drop this one too

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -225,15 +339,38 @@ public void writeTo(DataOutput out) throws IOException {
...
   public static Binary fromByteArray(final byte[] value) {
-    return new ByteArrayBackedBinary(value);
+    return fromReusedByteArray(value); // Assume producer intends to reuse byte[]
   }
 
   private static class ByteBufferBackedBinary extends Binary {
     private transient ByteBuffer value;
+    private transient byte[] cachedBytes;
 
+    @Deprecated
```
isnotinvain: this one too

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
SinghAsDev: True, removed

isnotinvain:
+1 -- @rdblue ready to merge?

rdblue:
+1

PR #199, opened 2015-05-19
PARQUET-248: Add ParquetWriter.Builder.

rdblue:
This refactors the builder recently added to parquet-avro so that it can
be used by all object models. The Builder class is abstract and
implementations should extend it.

This changes the API slightly from AvroParquetWriter, renaming
withBlockSize to withRowGroupSize. The Avro builder has not been
released so this isn't a breaking change.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -61,6 +61,7 @@ public class ParquetWriter<T> implements Closeable {
    * @throws IOException
    * @see #ParquetWriter(Path, WriteSupport, CompressionCodecName, int, int, boolean, boolean)
    */
+  @Deprecated
   public ParquetWriter(Path file, WriteSupport<T> writeSupport, CompressionCodecName compressionCodecName, int blockSize, int pageSize) throws IOException {
     this(file, writeSupport, compressionCodecName, blockSize, pageSize,
         DEFAULT_IS_DICTIONARY_ENABLED, DEFAULT_IS_VALIDATING_ENABLED);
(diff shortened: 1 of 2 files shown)


rdblue:
@SinghAsDev, can you do a first round of review?

SinghAsDev:
+1, looks good to me.

rdblue:
@isnotinvain could you take a look? I think this is a good thing to get in now so we have time to get the other object models using builders.

peltekster:
@rdblue, This was already done in https://github.com/apache/parquet-mr/pull/157
I have no idea why it was ignored. Anyway, are you interested in having control on the encoding used per column?

rdblue:
Hi @peltekster, sorry about that. I had no idea it had been done as part of another issue. Since this is a separate issue, it should be a bit easier to merge this one and use it to simplify the other commit. Would you mind reviewing this then?

rdblue:
@isnotinvain, now that #211 is in, I've updated this and added the new options from master. I think it's about ready to go in.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -273,4 +299,189 @@ public void close() throws IOException {
...
+    }
+
+    /**
+     * Set the Parquet format row group size used by the constructed writer.
+     *
+     * @param rowGroupSize an integer size in bytes
+     * @return this builder for method chaining.
+     */
+    public SELF withRowGroupSize(int rowGroupSize) {
+      this.blockSize = rowGroupSize;
```
isnotinvain: should we also rename this.blockSize to be this.rowGroupSize ?

isnotinvain:
+1

PR #200, opened 2015-05-19
PARQUET-262: Restore semver checks.

rdblue:
Because 1.6.0 to 1.7.0 was a breaking rename, semver was turned off
until the 1.7.0 artifacts were released. Now that they are available,
the check needs to be restored.

There were already 2 breaking changes that are fixed by this commit:
* A field in AvroReadSupport was made final
* An accessor method in ThriftSchemaConvertVisitor was removed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,7 @@
     <cascading.version>2.5.3</cascading.version>
     <parquet.format.version>2.3.0-incubating</parquet.format.version>
     <log4j.version>1.2.17</log4j.version>
-    <previous.version>1.5.0</previous.version>
+    <previous.version>1.7.0</previous.version>
     <thrift.executable>thrift</thrift.executable>
     <scala.version>2.10.4</scala.version>
     <!-- scala.binary.version is used for projects that fetch dependencies that are in scala -->
@@ -222,7 +222,6 @@
             </dependency>
          </dependencies>
          <executions>
-           <!--
            <execution>
              <id>check</id>
              <phase>verify</phase>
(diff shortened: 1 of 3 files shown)


rdblue:
@isnotinvain and @julienledem, could you review? This should go in soon to avoid more breaking changes.

isnotinvain:
There are a bunch of excludes listed, but they are all for the old package name. We should swap them to the new package name, and probably prune the list too, though it's difficult because a lot of these classes are private (ThriftSchemaConverter for example) and we should be free to change them around.

rdblue:
Good point. We need to clean that up. Ideally, we'd take the time to define a public API now so we can move everything to public or private packages by 2.0.0. I'm not sure anyone will have time though.

rdblue:
What about merging this as-is and then filing a follow-up issue to fix the exclusions?

isnotinvain:
A follow up is fine, but in this PR, I think we should either: clear all the `<exclude>`s because they are basically no-ops right now, or keep them as is, but make them `org/apache/parquet/...` right?

rdblue:
Fixed. I restored the current exclusions and updated the ones for shaded deps.

pom.xml
```diff
@@ -235,19 +234,19 @@
...
-                     <exclude>parquet/column/**</exclude>
-                     <exclude>parquet/hadoop/ParquetInputSplit</exclude>
+                     <exclude>org/apache/parquet/hadoop/thrift/**</exclude>
+                     <exclude>org/apache/parquet/thrift/projection/**</exclude>
+                     <exclude>org/apache/parquet/thrift/ThriftSchemaConverter</exclude>
+                     <exclude>org/apache/parquet/filter2/**</exclude>
+                     <exclude>org/apache/parquet/column/**</exclude>
+                     <exclude>org/apache/parquet/hadoop/ParquetInputSplit</exclude>
+                     <exclude>shaded/**</exclude> <!-- shaded by parquet -->
+                     <exclude>parquet/**</exclude> <!-- shaded by parquet-format -->
```
julienledem: I don't think we need exclusions for what comes from a separate jar (parquet-format).
Also the shaded packages should go under org/apache/parquet/shaded.

rdblue: If we need to exclude shaded binaries, then we need to make sure we get them all. `parquet/**` covers slf4j and thrift from parquet-format that are present in jars that shade parquet-format. I agree we should move all shaded dependencies to org/apache/parquet/shaded, but that's separate issue, right?

julienledem: the plugin checks for what is in this repo, not dependencies. so parquet-format is not part of it.
agreed renaming shaded packages is separate, just wanted to mention it.

In general I would err on the side of removing excludes when in doubt. They can always be added when needed. Are you seeing errors related to this?

julienledem:
+1 
I made a comment on the current list.

pom.xml
rdblue: I just built parquet-column to confirm that the shade plugin runs before the enforcer plugin. The semver rule checks binaries, right? Then the shaded binaries would be included in the check unless they are excluded like this.

I'm not seeing an error and I don't really expect one without changing dependency versions. But if it should be excluded I see no reason not to.

julienledem: I'm just suspecting that this particular line does nothing. But I may be wrong.
The semver rule should check only the binaries produced by the project, not dependencies (like parquet-format).

rdblue: When the dependencies are shaded, then their shaded dependencies are included. For example, parquet-tools shades parquet-format, which shades things under `parquet/**`. We should update both to use org/apache/parquet/shaded.

isnotinvain:
+1, though maybe in a follow up I can try and remove some of the exclusions.

pom.xml
julienledem: ah OK. Then I don't get why parquet-tools shades parquet-format. But that's another discussion.

rdblue: Yeah, we need to take a look at the packaging for parquet-tools. When it moved to a shaded jar that messed things up for me. We will package the old way, as an assembly tarball.

rdblue:
Merged. Thanks for the reviews!

PR #202, opened 2015-05-26
minor modify:swap the name of twoBlocks and oneBlocks

baishuo:
In original code, the variable **twoBlocks**  actually holds one block but **oneBlock**  holds two blocks,  This pr just change swap the name of  twoBlocks and oneBlocks


rdblue:
@baishuo, I don't think the variable names you changed were referring to the number of blocks. They were referring to the associated file ("one.parquet" or "two.parquet"). While it may make some sense to have made those names align with the number of blocks in the file, I don't see much of a reason to make changes to this code. It's working and tests what it needs to test already. Is there a particular motivation for this PR?

baishuo:
@rdblue I see：），close this pr

PR #203, opened 2015-05-26
PARQUET-289: Allow ParquetReader.Builder subclasses.

rdblue:
This adds a protected constructor for subclasses, a getReadSupport
method for subclasses to override, and exposes the configuration for
subclasses to modify inside of getReadSupport.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
+import org.apache.parquet.Preconditions;
 import org.apache.parquet.filter.UnboundRecordFilter;
 import org.apache.parquet.filter2.compat.FilterCompat;
 import org.apache.parquet.filter2.compat.FilterCompat.Filter;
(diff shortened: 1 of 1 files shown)


kostya-sh:
I wonder if introducing AbstractBuilder class (Builder extends AbstractBuilder, getReadSupport is abstract method of AbstractBuilder) would be a cleaner way to implement this? This will make check that getReadSupport overridden a compile time rather than runtime check.

rdblue:
@kostya-sh, it would definitely be cleaner. It also needs a `self` method, like `ParquetWriter.Builder` in #199. The problem is that changing a public class to an abstract class breaks binary compatibility, so it can't be done until 2.x.

kostya-sh:
I might be wrong but I think `AbstractBuilder` can be introduced without changing `Parquet.Builder` in backward incompatible way:

    class AbstractBuilder {
        // move most of the stuff from Builder here
    }

    class Builder extends AbstractBuilder {
       // only constructor and getReadSupport() needs to be here
    }

rdblue:
We can introduce the extra class to avoid breaking compatibility for `ParquetReader.Builder`, but then we wouldn't be able to use it without breaking compatibility for the builder factory methods, like `AvroParquetReader.builder`. That returns a `ParquetReader.Builder` and returning a subclass of `AbstractBuilder` instead would break compatibility.

kostya-sh:
Thanks for the explanation, it makes sense now. I knew I was missing something. Designing backward compatible APIs is hard.

tsdeng:
LGTM +1

PR #204, opened 2015-05-26
PARQUET-290: Add data model to Avro reader builder

rdblue:
This PR currently includes #203, which will be removed when it is merged.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetReader.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetReader.java
@@ -20,11 +20,14 @@ package org.apache.parquet.avro;
 
 import java.io.IOException;
 
+import org.apache.avro.generic.GenericData;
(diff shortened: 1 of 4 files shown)


tsdeng:
LGTM +1

parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetReader.java
```diff
@@ -66,4 +69,47 @@ public AvroParquetReader(Configuration conf, Path file) throws IOException {
...
+    }
+
+    public Builder<T> withCompatibility(boolean enableCompatibility) {
+      this.enableCompatibility = enableCompatibility;
+      return this;
+    }
+
+    @Override
+    protected ReadSupport<T> getReadSupport() {
+      if (isReflect) {
```
avinash-traceable, line 107: isReflect is always true. Shouldn't it be false by default?
The else below is unusable because of this.

PR #205, opened 2015-06-01
PARQUET-151: Skip writing _metadata file in case of no footers since schema cannot be determined.

saucam:
This fixes npe seen during mergeFooters in such a case.
 For this scenario onus of writing any summary files lies with the caller (It might have some global schema available) So for example spark does it when persisting empty RDD.


rdblue:
Looks good to me. Thanks @saucam!

saucam:
Thanks for the review and the merge @rdblue  !

PR #208, opened 2015-06-02
PARQUET-297: fix reading parquet version

kostya-sh:


SinghAsDev:
@kostya-sh thanks for fixing this. Can we add some tests cases?

kostya-sh:
@SinghAsDev, adding test cases is not very straightforward task as the code relies on parquet-column.jar being available in the classpath and it hasn't been built yet when maven runs tests. Please let me know if you know of any good way to write tests for this functionality.

kostya-sh:
Reading a file from parquet-column.jar is also questionable, why not parquet-common.jar where the Version class resides? What was the original reason for reading parquet-common pom.properties?

SinghAsDev:
@kostya-sh I think we can create test jar files with just a test META-INF, containing version, file in it. Makes sense?

kostya-sh:
@SinghAsDev How exactly do you propose to do it? Commit a test jar file to Git and add it to the test classpath by specifying additionalClasspathElements for maven-surefire-plugin? This might work, if others are happy with this approach I will add a test.

kostya-sh:
BTW reading Git hash from Manifest file doesn't work if manifest file is not available. E.g. in parquet files created with Spark, created_by looks like `parquet-mr version 1.6.0rc7`. Maybe the build should be modified to put a file with a Git hash along with pom.properties

kostya-sh:
I've implemented fix in a different (I believe more robust) way. The way it works:
- META-INF/maven/org.apache.parquet/parquet-common/version.properties is created during the build
- Version reads VERSION_NUMBER and FULL_VERSION values from this file

It also works when Manifest file is no longer present (like in Spark).

SinghAsDev:
+1
@kostya-sh Sorry about the delay on getting to this. Really liked your recent changes. Thanks!

rdblue:
@kostya-sh, these changes are looking good. In the last sync-up, we talked about using a strategy like this to get the information, but then adding a class to parquet-generator that generates `Version` with `VERSION_NUMBER` and `FULL_VERSION` constants. That way, we can always rely on having the information, even if META-INF content has been removed. If the `Version` class isn't present, you can't write data at all. Would you mind updating this to generate a `Version` class? Otherwise we'll get to it before too long.

kostya-sh:
I have created another pull request: https://github.com/apache/parquet-mr/pull/213

I also made one minor improvement to the build related to parquet-generator: https://issues.apache.org/jira/browse/PARQUET-309 (https://github.com/apache/parquet-mr/pull/214)

PR #210, opened 2015-06-09
PARQUET-304: Add an option to make requested schema case insensitive in read path

saucam:
For projects such as hive and spark-sql which use parquet, the schema of the stored tables is always lowercase (because of limitations of hive metastore). It would be great if we can have a configurable option to read data from parquet irrespective of the case of the requested schema , supplied via ReadContext object.

This PR adds configurable option of ParquetInputFormat.CASE_SENSITIVITY which can be set to false. In that case, parquet will resolve requested columns irrespective of case. This alleviates projects like spark-sql to read footers at the driver side and reconcile schema (to be read from footers) with metastore schema before calling parquet read.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
@@ -59,9 +59,13 @@ import static org.apache.parquet.Preconditions.checkNotNull;
 public class SchemaCompatibilityValidator implements FilterPredicate.Visitor<Void> {
 
   public static void validate(FilterPredicate predicate, MessageType schema) {
+   validate(predicate, schema, true);
+  }
+
+  public static void validate(FilterPredicate predicate, MessageType schema, boolean isCaseSensitive) {
     checkNotNull(predicate, "predicate");
     checkNotNull(schema, "schema");
-    predicate.accept(new SchemaCompatibilityValidator(schema));
+    predicate.accept(new SchemaCompatibilityValidator(schema, isCaseSensitive));
   }
 
   // A map of column name to the type the user supplied for this column.
@@ -76,11 +80,14 @@ public class SchemaCompatibilityValidator implements FilterPredicate.Visitor<Voi
   // the original type of a column, keyed by path
   private final Map<ColumnPath, OriginalType> originalTypes = new HashMap<ColumnPath, OriginalType>();
 
-  private SchemaCompatibilityValidator(MessageType schema) {
+  private boolean isCaseSensitive = true;
 
+  private SchemaCompatibilityValidator(MessageType schema, boolean isCaseSensitive) {
+
+    this.isCaseSensitive = isCaseSensitive;
     for (ColumnDescriptor cd : schema.getColumns()) {
-      ColumnPath columnPath = ColumnPath.get(cd.getPath());
-      columnsAccordingToSchema.put(columnPath, cd);
+      ColumnPath columnPath = isCaseSensitive? ColumnPath.get(cd.getPath()):ColumnPath.getLowerCase(cd.getPath());
+        columnsAccordingToSchema.put(columnPath, cd);
 
       OriginalType ot = schema.getType(cd.getPath()).getOriginalType();
       if (ot != null) {
@@ -186,7 +193,8 @@ public class SchemaCompatibilityValidator implements FilterPredicate.Visitor<Voi
   }
 
   private ColumnDescriptor getColumnDescriptor(ColumnPath columnPath) {
-    ColumnDescriptor cd = columnsAccordingToSchema.get(columnPath);
+    ColumnDescriptor cd = isCaseSensitive ? columnsAccordingToSchema.get(columnPath):
+      columnsAccordingToSchema.get(ColumnPath.getLowerCase(columnPath.toArray()));
     checkArgument(cd != null, "Column " + columnPath + " was not found in schema!");
     return cd;
   }
--- a/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 import org.apache.parquet.io.InvalidRecordException;
 
@@ -38,7 +39,6 @@ public class GroupType extends Type {
 
   private final List<Type> fields;
   private final Map<String, Integer> indexByName;
-
   /**
    * @param repetition OPTIONAL, REPEATED, REQUIRED
    * @param name the name of the field
@@ -46,6 +46,10 @@ public class GroupType extends Type {
    */
   public GroupType(Repetition repetition, String name, List<Type> fields) {
     this(repetition, name, null, fields, null);
+  }
+
+  public GroupType(Repetition repetition, String name, List<Type> fields, boolean isCaseSensitive) {
+    this(repetition, name, null, fields, null, isCaseSensitive);
   }
 
   /**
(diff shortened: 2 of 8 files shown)


saucam:
cc @isnotinvain 

cc @liancheng   request your thoughts on this ?

liancheng:
@saucam Thanks for working on this!

To be more specific, Spark SQL itself can be configured to be either case sensitive or case insensitive.  But when using Spark SQL to access Hive tables whose metadata are stored in Hive metastore, it have to be case insensitive because Hive metastore is so. Our current solution is to read both Hive metastore schema and Parquet schema, then try to get an arbitrative schema by having case information from Parquet schema and column type information from Hive schema.

To be honest, I feel kinda complicated about this issue. From the perspective of Spark SQL, schema resolution code can be simpler if Parquet provides this configuration. However, personally I feel that case insensitivity can be a footgun in many cases and a source of bugs. Let's leave this to Parquet committers to decide :)

danielcweeks:
The way Hive resolves this is that the client never reads the footer (only hive metadata), but when the task processes the file it will read the footer and resolve the case sensitivity (see https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/read/DataWritableReadSupport.java#L155).  I'm not opposed to having it configurable, but I'm not clear why it's necessary.

saucam:
@danielcweeks this is possible to do in spark as well (have tried, works for projections) but how do you resolve column names in filter objects which are pushed down in parquet before creating the tasks ? For that it becomes inevitable to have a resolved schema on the driver side ? I wonder if and how hive is able to resolve that ?

danielcweeks:
@saucam it looks like in hive they push the filter expression to the task side as well and evaluate it there.  However, I don't see that they address the case sensitivity issue, but I just took a cursory look at the code.  I assume they could do the same as the do for column projection.

saucam:
@julienledem request your thoughts on this ...

nezihyigitbasi:
Presto also handles case sensitivity itself on the coordinator side before running the tasks, please see [this](https://github.com/facebook/presto/pull/4686) PR.

billonahill:
@nezihyigitbasi Presto works with case sensitivity with the PR you reference but that fix is limited to top-level columns only. It doesn't address querying case sensitive nested data structures, which is currently not supported in presto (see related https://github.com/prestodb/presto/issues/2863). I believe this proposed parquet fix would address that.

julienledem:
If we add an option for case insensitivity in Parquet, we should make sure it is consistent and decide what to do with conflicting names. It sounds like it's case insensitive selection. projections and filters apply to all column that match ignoring the case?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
```diff
@@ -59,9 +59,13 @@
 public class SchemaCompatibilityValidator implements FilterPredicate.Visitor<Void> {
 
   public static void validate(FilterPredicate predicate, MessageType schema) {
+   validate(predicate, schema, true);
+  }
+
+  public static void validate(FilterPredicate predicate, MessageType schema, boolean isCaseSensitive) {
```
julienledem: I would think that isCaseSensitive should be a property of the FilterPredicate.

parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -48,6 +48,10 @@ public GroupType(Repetition repetition, String name, List<Type> fields) {
     this(repetition, name, null, fields, null);
   }
 
+  public GroupType(Repetition repetition, String name, List<Type> fields, boolean isCaseSensitive) {
```
julienledem: this should not be a property of the type itself, but a property of the selection

parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -391,4 +400,16 @@ protected Type union(Type toMerge, boolean strict) {
     }
     return newFields;
   }
+
+  public List<Type> getContainedFields(GroupType base) {
```
julienledem: I'd make this method optionally case sensitive.

parquet-column/src/main/java/org/apache/parquet/schema/MessageType.java
```diff
@@ -50,6 +50,13 @@ public MessageType(String name, List<Type> fields) {
    super(Repetition.REPEATED, name, fields);
  }
 
+ public MessageType(String name, List<Type> fields, boolean isCaseSensitive) {
+     super(Repetition.REPEATED, name, fields, isCaseSensitive);
+ }
+
+ public MessageType(MessageType base, boolean isCaseSensitive) {
+     super(Repetition.REPEATED, base.getName(), base.getFields(), isCaseSensitive);
+ }
```
julienledem: same comment, I don't think the property belong here.

parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/ColumnPath.java
```diff
@@ -48,6 +59,10 @@ public static ColumnPath get(String... path){
     return paths.canonicalize(new ColumnPath(path));
   }
 
+  public static ColumnPath getLowerCase(String... path){
+    return pathsLowerCase.canonicalize(new ColumnPath(path));
+  }
```
julienledem: It seems that the toLowerCase can be applied to path directly and then call the existing get(). There's no need for a separate Canonicalizer

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -56,7 +67,7 @@ private RowGroupFilter(List<BlockMetaData> blocks, MessageType schema) {
     FilterPredicate filterPredicate = filterPredicateCompat.getFilterPredicate();
 
     // check that the schema of the filter matches the schema of the file
-    SchemaCompatibilityValidator.validate(filterPredicate, schema);
+    SchemaCompatibilityValidator.validate(filterPredicate, schema, isCaseSensitive);
```
julienledem: if the predicate has the isCaseSensitive field it can be applied here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -169,13 +169,23 @@ public void initialize(MessageType fileSchema,
     // initialize a ReadContext for this file
     ReadSupport.ReadContext readContext = readSupport.init(new InitContext(
         configuration, toSetMultiMap(fileMetadata), fileSchema));
-    this.requestedSchema = readContext.getRequestedSchema();
```
julienledem: Whether the requestedContext is caseSensitive can com from the readContext. I don't think we need a Hadoop property for this.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -169,13 +169,23 @@ public void initialize(MessageType fileSchema,
...
+    MessageType request = readContext.getRequestedSchema();
+    boolean isCaseSensitive = configuration.getBoolean(ParquetInputFormat.CASE_SENSITIVITY, true);
```
julienledem: I would prefer we don't use configuration deep in the code for such things.
The ReadSupport implementation can pass thing through the ReadContext. You can add a field there

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -169,13 +169,23 @@ public void initialize(MessageType fileSchema,
...
+
+    if (!isCaseSensitive) {
+      MessageType base = new MessageType(request, false);
+      this.requestedSchema = new MessageType(request.getName(), fileSchema.getContainedFields(base));
```
julienledem: instead of adding isCaseSensitive in base, it can be passed to getContainedField

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -114,6 +114,11 @@
   public static final String FILTER_PREDICATE = "parquet.private.read.filter.predicate";
 
   /**
+   * key to configure case sensitivity
+   */
+  public static final String CASE_SENSITIVITY = "parquet.read.case.sensitivity";
```
julienledem: I don't think we need a property in that case.

julienledem:
I made comments inline. This is the kind of feature that is easy to add but hard to get right and to change if not right.
Sorry for letting this drag for long.

PR #211, opened 2015-06-09
PARQUET-306: Add row group alignment

rdblue:
This adds `AlignmentStrategy` to the `ParquetFileWriter` that can alter the position of row groups and recommend a target size for the next row group. There are two strategies: `NoAlignment` and `PaddingAlignment`. Padding alignment is used for HDFS and no alignment is used for all other file systems. When HDFS-3689 is available, we can add a strategy to use that.

The amount of padding is controlled by a threshold between 0 and 1 that controls the fraction of the row group size that can be padded. This is interpreted as the maximum amount of padding that is acceptable, in terms of the row group size. For example, setting this to 5% will write padding when the bytes left in a HDFS block are less than 5% of the row group size. This defaults to 0%, which prevents padding from being added and matches the current behavior. The threshold is controlled by a new OutputFormat configuration property, `parquet.writer.padding-thresh`.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -41,6 +41,7 @@ import org.apache.parquet.Version;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.page.DictionaryPage;
 import org.apache.parquet.column.statistics.Statistics;
 import org.apache.parquet.hadoop.metadata.ColumnPath;
@@ -69,6 +70,20 @@ public class ParquetFileWriter {
   public static final byte[] MAGIC = "PAR1".getBytes(Charset.forName("ASCII"));
   public static final int CURRENT_VERSION = 1;
 
+  // need to supply a buffer size when setting block size. this is the default
+  // for hadoop 1 to present. copying it avoids loading DFSConfigKeys.
+  private static final int DFS_BUFFER_SIZE_DEFAULT = 4096;
+
+  // visible for testing
+  static final Set<String> BLOCK_FS_SCHEMES = new HashSet<String>();
+  static {
+    BLOCK_FS_SCHEMES.add("hdfs");
+  }
+
+  private static boolean supportsBlockSize(FileSystem fs) {
+    return BLOCK_FS_SCHEMES.contains(fs.getUri().getScheme());
+  }
+
   // File creation modes
   public static enum Mode {
     CREATE,
@@ -79,13 +94,13 @@ public class ParquetFileWriter {
 
   private final MessageType schema;
   private final FSDataOutputStream out;
+  private final AlignmentStrategy alignment;
   private BlockMetaData currentBlock;
-  private ColumnChunkMetaData currentColumn;
   private long currentRecordCount;
   private List<BlockMetaData> blocks = new ArrayList<BlockMetaData>();
   private long uncompressedLength;
   private long compressedLength;
-  private Set<org.apache.parquet.column.Encoding> currentEncodings;
+  private Set<Encoding> currentEncodings;
 
   private CompressionCodecName currentChunkCodec;
   private ColumnPath currentChunkPath;
@@ -157,7 +172,7 @@ public class ParquetFileWriter {
    */
   public ParquetFileWriter(Configuration configuration, MessageType schema,
       Path file) throws IOException {
-    this(configuration, schema, file, Mode.CREATE);
+    this(configuration, schema, file, Mode.CREATE, ParquetWriter.DEFAULT_BLOCK_SIZE);
   }
 
   /**
@@ -168,12 +183,55 @@ public class ParquetFileWriter {
    * @throws IOException if the file can not be created
    */
   public ParquetFileWriter(Configuration configuration, MessageType schema,
-      Path file, Mode mode) throws IOException {
-    super();
+                           Path file, Mode mode) throws IOException {
+    this(configuration, schema, file, mode, ParquetWriter.DEFAULT_BLOCK_SIZE);
+  }
+
+  /**
+   * @param configuration Hadoop configuration
+   * @param schema the schema of the data
+   * @param file the file to write to
+   * @param mode file creation mode
+   * @param rowGroupSize the row group size
+   * @throws IOException if the file can not be created
+   */
+  public ParquetFileWriter(Configuration configuration, MessageType schema,
+      Path file, Mode mode, long rowGroupSize) throws IOException {
     this.schema = schema;
     FileSystem fs = file.getFileSystem(configuration);
     boolean overwriteFlag = (mode == Mode.OVERWRITE);
-    this.out = fs.create(file, overwriteFlag);
+
+    if (supportsBlockSize(fs)) {
+      // use the default block size, unless row group size is larger
+      long dfsBlockSize = Math.max(fs.getDefaultBlockSize(file), rowGroupSize);
+
+      this.alignment = PaddingAlignment.get(dfsBlockSize, rowGroupSize);
+      this.out = fs.create(file, overwriteFlag, DFS_BUFFER_SIZE_DEFAULT,
+          fs.getDefaultReplication(file), dfsBlockSize);
+
+    } else {
+      this.alignment = NoAlignment.get(rowGroupSize);
+      this.out = fs.create(file, overwriteFlag);
+    }
+  }
+
+  /**
+   * FOR TESTING ONLY.
+   *
+   * @param configuration Hadoop configuration
+   * @param schema the schema of the data
+   * @param file the file to write to
+   * @param rowAndBlockSize the row group size
+   * @throws IOException if the file can not be created
+   */
+  ParquetFileWriter(Configuration configuration, MessageType schema,
+                    Path file, long rowAndBlockSize)
+      throws IOException {
+    FileSystem fs = file.getFileSystem(configuration);
+    this.schema = schema;
+    this.alignment = PaddingAlignment.get(rowAndBlockSize, rowAndBlockSize);
+    this.out = fs.create(file, true, DFS_BUFFER_SIZE_DEFAULT,
+        fs.getDefaultReplication(file), rowAndBlockSize);
   }
 
   /**
@@ -195,6 +253,9 @@ public class ParquetFileWriter {
     state = state.startBlock();
     if (DEBUG) LOG.debug(out.getPos() + ": start block");
 //    out.write(MAGIC); // TODO: add a magic delimiter
+
+    alignment.alignForRowGroup(out);
+
     currentBlock = new BlockMetaData();
     currentRecordCount = recordCount;
   }
@@ -211,8 +272,7 @@ public class ParquetFileWriter {
                           long valueCount,
                           CompressionCodecName compressionCodecName) throws IOException {
     state = state.startColumn();
-    if (DEBUG) LOG.debug(out.getPos() + ": start column: " + descriptor + " count=" + valueCount);
-    currentEncodings = new HashSet<org.apache.parquet.column.Encoding>();
+    currentEncodings = new HashSet<Encoding>();
     currentChunkPath = ColumnPath.get(descriptor.getPath());
     currentChunkType = descriptor.getType();
     currentChunkCodec = compressionCodecName;
@@ -263,9 +323,9 @@ public class ParquetFileWriter {
   public void writeDataPage(
       int valueCount, int uncompressedPageSize,
       BytesInput bytes,
-      org.apache.parquet.column.Encoding rlEncoding,
-      org.apache.parquet.column.Encoding dlEncoding,
-      org.apache.parquet.column.Encoding valuesEncoding) throws IOException {
+      Encoding rlEncoding,
+      Encoding dlEncoding,
+      Encoding valuesEncoding) throws IOException {
     state = state.write();
     long beforeHeader = out.getPos();
     if (DEBUG) LOG.debug(beforeHeader + ": write data page: " + valueCount + " values");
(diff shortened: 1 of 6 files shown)


rdblue:
@isnotinvain could you look at this before the ParquetWriter builder? If we can get this in, then I'll have to update the builder class to set the padding threshold.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -53,12 +53,12 @@ public void setMinMaxFromBytes(byte[] minBytes, byte[] maxBytes) {
 
   @Override
   public byte[] getMaxBytes() {
-    return max.getBytes();
+    return hasNonNullValue() ? max.getBytes() : null;
```
isnotinvain: should this just throw if !hasNonNullValue?
We already provide hasNonNullValue, shouldn't the caller inspect that before calling getMaxBytes?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -69,6 +72,20 @@
...
   public static final int CURRENT_VERSION = 1;
 
+  // need to supply a buffer size when setting block size. this is the default
+  // for hadoop 1 to present. copying it avoids loading DFSConfigKeys.
+  private static final int DFS_BUFFER_SIZE_DEFAULT = 4096;
+
+  // visible for testing
+  static final Set<String> BLOCK_FS_SCHEMES = new HashSet<String>();
+  static {
+    BLOCK_FS_SCHEMES.add("hdfs");
```
isnotinvain: do we need hftp too?

isnotinvain: should we extract this info from the path, or just make it a required argument to the constructor, and pass the right one in all the hadoop related usages?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -168,12 +185,59 @@ public ParquetFileWriter(Configuration configuration, MessageType schema,
...
+                           float paddingThresh)
+      throws IOException {
     this.schema = schema;
     FileSystem fs = file.getFileSystem(configuration);
     boolean overwriteFlag = (mode == Mode.OVERWRITE);
-    this.out = fs.create(file, overwriteFlag);
+
+    if (supportsBlockSize(fs)) {
+      // use the default block size, unless row group size is larger
+      long dfsBlockSize = Math.max(fs.getDefaultBlockSize(file), rowGroupSize);
```
isnotinvain: by using the default block size, are we ignoring user's requests for non-default block sizes?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -69,6 +72,20 @@
   public static final byte[] MAGIC = "PAR1".getBytes(Charset.forName("ASCII"));
   public static final int CURRENT_VERSION = 1;
 
+  // need to supply a buffer size when setting block size. this is the default
+  // for hadoop 1 to present. copying it avoids loading DFSConfigKeys.
+  private static final int DFS_BUFFER_SIZE_DEFAULT = 4096;
```
isnotinvain: this is normally overridable via the hadoop config right? by hardcoding it, now requests for non default buffer sizes won't be respected right?

isnotinvain:
How do we make sure that row groups are still aligning with HDFS blocks? Everything gets written to an output stream, so it seems like it would be really easy to accidentally shift everything slightly and ruin the alignment -- eg we write a magic header at the beginning of the file right? does that push everything slightly out of alignment? It only has to be slightly off for any wins to be lost right?

I'm just wondering if we're relying on being very careful about that, or if the way this works is not sensitive to that (maybe it's not because it asks the output stream for it's position, which is all that matters maybe?)

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
rdblue: This gets called from within Jackson when converting to JSON, so we don't have much of a choice.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: We weren't exposing a way to set this when we create the file anyway, so this isn't removing an option that users had. It is setting an option that the library had and wasn't using. Hard-coding it to the value we would always use anyway. I need to have a value for this because there isn't a method call that allows us to set the block size and default the buffer size. This default is very stable so I think this is perfectly safe.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: Since we can tell easily, I think it makes sense to check this way. No need to complicate the API when you wouldn't ever change it.

rdblue: What's HFTP?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: No, users control their HDFS block size by setting `dfs.block.size`, which is what this method accesses. So this is getting the user-configured block size, checking that it is valid, and using it.

rdblue:
> How do we make sure that row groups are still aligning with HDFS blocks?

This is based off of the position that the FS OutputStream reports, which is always going to be correct. If some other thread somehow had a reference to it and wrote data, then this would see that the position had changed.

I'm also going to update this so that we don't use a percentage of the row group size to configure it. I'm going to add a max padding setting in bytes instead.

rdblue:
Ok, changed padding thresh to maxPaddingSize. I think that's much better.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
isnotinvain: ok SGTM

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: HFTP is a FileSystem implementation for HDFS over HTTP. It's used to read data across clusters when they have mismatched versions of HDFS.

I guess more of what I was getting at is, there's HDFS, HFTP, and VIEWFS, and (possibly?) more. So locking in to just hdfs maybe is too strict? I wonder if there's a property of FileSystem that describes whether it supports blocks / or if there's some flag that's set (like block size) that indicates that the FS uses blocks?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: ok that's what I wanted to double check, thanks

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: > so this isn't removing an option that users had
Yeah that's what I'm wondering about -- we were using a method before that didn't ask for this info, did that method pull this info from the hadoop configuration? If so -- then users were previously able to change this by messing w/ their hadoop configuration, but now they can't.

On the other hand, if the method we were using before just grabbed the hardcoded default, then this makes sense as it's no different (I'm not worried about duplicating the default, just about whether we've taken away the ability to change this via hadoop configuration)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: Looks like [HFTP is a read-only FS|https://hadoop.apache.org/docs/r2.4.1/hadoop-project-dist/hadoop-hdfs/Hftp.html] so we shouldn't need to add that. But I think we should add viewfs as you suggested and webhdfs. Anything backed by HDFS should probably have it but we can add more over time.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: No, this is something you set through the API, it wasn't exposed before, and the default hasn't changed since Hadoop-1.

rdblue:
I've fixed the tests and added webhdfs and viewfs to the supported block schemes. Should be good to go?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: ok thanks

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -225,6 +234,18 @@ private CompressionCodecName getCodec(TaskAttemptContext taskAttemptContext) {
     return CodecConfig.from(taskAttemptContext).getCodec();
   }
 
+  public static void setMaxPaddingSize(JobContext jobContext, int maxPaddingSize) {
```
isnotinvain: nit, but how about `setMaxRowGroupPaddingBytes`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -225,6 +234,18 @@ private CompressionCodecName getCodec(TaskAttemptContext taskAttemptContext) {
...
+    setMaxPaddingSize(getConfiguration(jobContext), maxPaddingSize);
+  }
+
+  public static void setMaxPaddingSize(Configuration conf, int maxPaddingSize) {
```
isnotinvain: nit, but how about `setMaxRowGroupPaddingBytes`

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -98,56 +127,238 @@ public void testWriteMode() throws Exception {
...
-    w.writeDataPage(3, 4, BytesInput.from(bytes2), stats2, BIT_PACKED, BIT_PACKED, PLAIN);
-    w.writeDataPage(1, 4, BytesInput.from(bytes2), stats2, BIT_PACKED, BIT_PACKED, PLAIN);
+    w.writeDataPage(2, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
+    w.writeDataPage(3, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
+    w.writeDataPage(1, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
     w.endColumn();
     long c2Ends = w.getPos();
     w.endBlock();
+
+    long firstRowGroupEnds = w.getPos(); // should be 109
```
isnotinvain: should you assert this?

isnotinvain:
What do you think about adding one more test, that writes a parquet file (with padding) and then reads it back (there are already lots of tests that do this, but w/o padding). I remember seeing discussion around a different issue where parquet stores an offset for the dictionary header, but never actually reads it, just assuming that the dictionary is the first page, and so we never noticed that the offset was wrong, etc. I guess I'm saying it'd be good to sanity check that parquet still reads files with padding in them w/o issue.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
rdblue: No, so that the test is less likely to fail if we add a metadata field.

rdblue:
@isnotinvain, good idea. I originally set the default padding to non-zero so all the tests tested that, which is why you don't see one. I still think we should change the default to include a reasonable amount of padding, but probably not in this PR.

rdblue:
I added the test, but I didn't change the method names.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
rdblue: I don't think adding row group is more clear, this is the maximum amount of padding per block, not per row group. We just happen to check if we should pad between row groups. I guess it is a little strange because it ends up being the minimum row group size, but we could change the implementation so that isn't the case.

rdblue:
Rebased on top of #221 and tests are passing.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInputOutputFormatWithPadding.java
```diff
@@ -0,0 +1,216 @@
...
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @Test
+  public void testBasicBehaviorWithPadding() throws Exception {
+    ParquetFileWriter.BLOCK_FS_SCHEMES.add("file");
+
+    File inputFile = temp.newFile();
+    FileOutputStream out = new FileOutputStream(inputFile);
+    out.write(FILE_CONTENT.getBytes());
```
isnotinvain: probably want to `getBytes("UTF-8")` right?

isnotinvain:
+1, just one comment about character encodings in the tests

PR #212, opened 2015-06-16
PARQUET-308: Add ParquetWriter#getDataSize accessor.

rdblue:
This returns the current file position plus the amount of data buffered
in the current row group as an estimate of final data size.


tsdeng:
LGTM +1

PR #213, opened 2015-06-17
PARQUET-297: generate Version class using parquet-generator

kostya-sh:


parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingGenerator.java
```diff
@@ -256,4 +254,8 @@ private static String align(int value, int digits) {
     result.append(valueString);
     return result.toString();
   }
+  
+  private static int paddedByteCountFromBits(int bitLength) {
```
rdblue: I don't see why these changes should be included in the version pull request. Maybe this was an accident?

parquet-generator/src/main/java/org/apache/parquet/version/VersionGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+    add("  public static final String VERSION_NUMBER = \"");
+    add(props.getProperty("versionNumber"));
+    add("\";\n");
+
+    add("  public static final String FULL_VERSION = \"");
+    add(props.getProperty("fullVersion"));
+    add("\";\n");
+
+    add("}\n");
+    writer.close();
```
rdblue: Other than the main method, this looks compatible with the old Version class to me. While it's nit-picking and probably over-cautious, I think we should probably include main.

rdblue:
Thanks for building this @kostya-sh! It looks great, just a couple of comments.

parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingGenerator.java
kostya-sh: I had to break circular dependency between parquet-generator and parquet-common. 

This was the only method from parquet-common that was used by parquet-generator, so I've just copied it here and removed dependency on parquet-common (have a look at parquet-generator/pom.xml change).

rdblue: Okay, that makes sense to me. Sounds good.

parquet-generator/src/main/java/org/apache/parquet/version/VersionGenerator.java
kostya-sh: Done

parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingGenerator.java
isnotinvain: yeah sort of a bummer but not too bad. Can you add a comment to this method: 
```
// duplicated from BytesUtils to avoid a circular dependency between parquet-common and parquet-generator
// TODO: fix the circular dependency by moving the shared bits to a 3rd module (PARQUET-XXXX)
```

isnotinvain:
Overall LGTM, just one minor comment

parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingGenerator.java
rdblue: +1

kostya-sh: I will add a comment about ByteUtils. Not sure it is worth fixing though as the method is one-liner, is not parquet specific and easy to understand.

isnotinvain:
+1

PR #214, opened 2015-06-17
PARQUET-309: remove unnecessary compile dependency on parquet-generator

kostya-sh:
parquet-generator is build-time dependency only and shouldn't be listed in
pom.xml dependencies section.


rdblue:
Looks good to me. @isnotinvain or @julienledem, this shouldn't be needed at runtime, right?

isnotinvain:
+1, we don't need it at runtime. If the build continues to work after a `mvn clean` then SGTM

(the generator sneaks some source files into other module's target/generated-source dirs)

rdblue:
Thanks @kostya-sh! Nice work.

PR #216, opened 2015-06-17
PARQUET-264: Remove remaining references to parquet being an incubator project

isnotinvain:
Do we need a new DISCLAIMER file, or can we just rm it?


rdblue:
We can remove the disclaimer. We're grown-ups now.

rdblue:
After a quick read-through, there are a couple more things:
* The contribution section still has a reference to the old google group instead of the dev@ list or the community page of parquet.apache.org
* The license section still says Copyright Twitter, which isn't correct any longer
* Code of conduct can list Twitter OSS, but should link to the Apache one as well: https://www.apache.org/foundation/policies/conduct.html
* Maybe we should remove the releases section, it still lists 1.0.0

There are also some out of date things that would be nice to fix... but we can do that later. The ones above are what *should* be fixed, I think.

rdblue:
Oh, and the license check is automatic, so we can remove that from the contribution section. Rat automatically runs with each build.

isnotinvain:
OK thanks, I'll update

isnotinvain:
@rdblue updated. I removed the CONTRIBUTING file because it was out of sync w/ the README and instead just took the good parts from the two of them and put it in README. We could instead move that section to CONTRIBUTING and remove it from README.

rdblue:
I had one comment on 6debacd, but otherwise +1.

benpence:
+1

rdblue:
I like the pick-me-up link. We'll just have to start using it. And reviewing patches for them. :)

PR #217, opened 2015-06-17
PARQUET-246: Recover data from files written with incomplete DELTA_BYTE_ARRAY state

spena:
Here's a fix for PARQUET-246. I run the patch over a file generated with the issue found with PARQUET-246 ,and it can read the file without problems.

I run all the unit tests as well, and it looks ok.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/Encoding.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/Encoding.java
@@ -189,13 +189,29 @@ public enum Encoding {
    * Suffixes are stored as delta length byte arrays.
    */
   DELTA_BYTE_ARRAY {
+    private DeltaByteArrayReader previous;
+
     @Override
     public ValuesReader getValuesReader(ColumnDescriptor descriptor,
         ValuesType valuesType) {
       if (descriptor.getType() != BINARY) {
         throw new ParquetDecodingException("Encoding DELTA_BYTE_ARRAY is only supported for type BINARY");
       }
-      return new DeltaByteArrayReader();
+
+      DeltaByteArrayReader current = new DeltaByteArrayReader();
+
+      //A bug was found on PARQUET-246 where DeltaByteArrayWriter was not cleaning the previous binary
+      //value before starting a new page reader, causing that the first value of the page being encoded uses
+      //a prefix value from the last page. This issue was fixed by resetting the previous binary value
+      //when writing data to a new page.
+      //However, buggy files were written with this issue. This line passes the previous binary
+      //value of the last page reader to the new page reader so that data can be recovered.
+      if (previous != null) {
+        current.setPreviousBinary(previous.getPreviousBinary());
+      }
+
+      previous = current;
+      return current;
     }
   },
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
@@ -536,15 +536,24 @@ class ColumnReaderImpl implements ColumnReader {
     } else {
       this.dataColumn = dataEncoding.getValuesReader(path, VALUES);
     }
+
+    try {
+      // Here we pass the last value from the last page read in order to read files written with the
+      // issue mentioned on PARQUET-246 where the prefixLength for DELTA_BYTE_ARRAY encoding values
+      // was not resetted when a new page was started.
+      if (path.getType().equals(PrimitiveTypeName.BINARY) && binding != null && binding.getBinary() != null) {
+        dataColumn.initFromPage(pageValueCount, bytes, offset, binding.getBinary().getBytes());
+      } else {
+        dataColumn.initFromPage(pageValueCount, bytes, offset, new byte[0]);
+      }
+    } catch (IOException e) {
+      throw new ParquetDecodingException("could not read page in col " + path, e);
+    }
+
     if (dataEncoding.usesDictionary() && converter.hasDictionarySupport()) {
       bindToDictionary(dictionary);
     } else {
       bind(path.getType());
-    }
-    try {
-      dataColumn.initFromPage(pageValueCount, bytes, offset);
-    } catch (IOException e) {
-      throw new ParquetDecodingException("could not read page in col " + path, e);
     }
   }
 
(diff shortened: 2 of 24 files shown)


isnotinvain:
Thanks for the PR!

I have a few questions:

1) this should only apply to columns that are encoded w/ delta byte array, and only if the file was written with the bug in place, right? So there should be a guard to only apply this for that specific encoding, and only if some config option is toggled, right? 

2) It'd be great if we could solve this without adding a param to a core API (the value reader) just to work around a bug that has since been fixed. Maybe we can just make a new encoding for this case?

spena:
Hi, thanks for the feedback.

I can avoid changing the core API, and setting the previous value only for DELTA_BYTE_ARRAY by using something like this:

dataColumn.initFromPage(pageValueCount, bytes, offset);
if (dataColumn instanceof DeltaByteArrayReader) {
  if (binding != null) {
    ((DeltaByteArrayReader) dataColumn).setPreviousBinary(binding.getBinary());
  }
}

I am thinking about your idea of using a new encoding instead, and how to detect the bug to toggle a config option to begin using the new encoding. But, the only place where I can detect the bug is on DeltaByteArrayReader.readBytes. At this point, the last value from the last page is not available, unless I initialize the reader with it like setPreviousBinary() does. 

The bug happens only with the first value of the page because prefixLength is > 0; and this first value shouldn't have prefixLength != 0, right? So, the previous variable initialized with the last value of the last page should not cause issues with new files written without the bug because prefixLength != 0 will be false. After that, previous will contain the correct previous value, and the rest of the decoding will work fine.

What do you think?

isnotinvain:
I think I need to read this more closely :p

But are you saying we don't need to hide this feature behind a flag because correctly written files will do the right thing, even with this fix in? That would be good :)

Even then, this seems very fragile... without extensive comments I feel like we could easily mess this up in the future, because this bug is not part of the file format. I sort of wonder whether this fix belongs in master, or if we should just cut a branch and provide instructions for using the branch as a recovery tool for corrupted files.

@julienledem @rdblue @tsdeng @SinghAsDev what do you think about baking this fix into the library vs. providing a recovery tool? I worry that this fix won't survive for long unless we're very careful, and it also complicates the library a fair amount.

spena:
Correct. The DeltaByteArrayWriter.reset() is called for every new page is created. This resets the previous binary value, so the first time writeBytes() is called, it will encode the value using 0 as prefix.
Then, when DeltaByteArrayReader.readBytes() is called with the first value of the new page, a 0 prefix will be detected, and this fix won't be called. However, if prefix != 0, then the bug was found, and this fix recovers the data. 

Btw, I made another change, and moved the logic to the Encoding.DELTA_BYTE_ARRAY to avoid  using instanceof.

isnotinvain:
@spena thanks, this update puts the logic in a good place.

I'm still not sure how to feel about putting this fix in master though -- by doing so we're essentially making this a "feature" of parquet, so it should have tests, and a defined contract, etc, or as @rdblue pointed out maybe hidden behind a flag. What do you think @julienledem @rdblue @tsdeng @SinghAsDev ?

parquet-column/src/main/java/org/apache/parquet/column/Encoding.java
```diff
@@ -189,13 +189,29 @@ public ValuesReader getValuesReader(ColumnDescriptor descriptor,
    * Suffixes are stored as delta length byte arrays.
    */
   DELTA_BYTE_ARRAY {
+    private DeltaByteArrayReader previous;
```
isnotinvain: This is a good fix for the bug, but now this class and its getter are stateful, which is why I am a little worried about putting it directly into parquet instead of a branch or repair tool or something.

rdblue:
Like the fix for PARQUET-251, I think this should be guarded by a version check. If we set the previous binary and fix the bug in PARQUET-246, then we have the opposite problem: we will encoding using a byte[0] as the starting previous and using the last page's previous value when decoding. We also need tests for both cases.

Is there a way to detect from the data itself that this encoding bug is present? I can't think of one, but that would be ideal. Otherwise, we have to use a version check. I considered the option of adding a flag to turn on this behavior for recovery, but this is one where users won't necessarily know that the data is corrupt, so we need to detect it and fix it without help.

One last thing: we need to make sure that we aren't skipping any pages when we are setting the previous binary like this. Otherwise we will end up with corrupt data again.

isnotinvain:
This is why I'm not totally sure if this belongs in the library itself? We can provide this in a branch for users that need to recover data, but do we want to carry forward support for corrupt files forever? In the case of parquet-251, we are opting out of an optimization, it seems safe to ignore statistics when the corruption might be present. Here, we need to be certain and choose the correct read strategy, have tests to enforce this (undocumented) behavior forever, *and* make sure that we never skip pages -- but what about in the future when we start applying statistics filters to pages?

Do you think this is a blocker for the release? We have fixed the problem going forward. Is it common for libraries like parquet to support reading corrupted files that were corrupted due to bugs in the library? (I have no idea, wondering if you've seen something like this before).

Thanks!
Alex

rdblue:
> do we want to carry forward support for corrupt files forever?

We have to because users don't necessarily know that the data is corrupt or how it was corrupted. If Parquet writes bad data, we need to always be able to read it correctly.

> we need to be certain and choose the correct read strategy

I think we should flag this in the page's metadata when reading it from disk. We already have the version in there, so we can add it easily enough. Like the stats problem, we assume that all Parquet MR files from versions before 1.8.0 are affected.

> have tests to enforce this (undocumented) behavior forever

Yes, but this isn't that expensive. We need to test that we can read fixed data and corrupted data. We can use the same data sample for it and assert different ways of reading it. As for this being undocumented... good point. We need to document these incompatibilities in the Parquet spec so other readers can implement them.

> make sure that we never skip pages

I think we should add a flag to the page metadata, then we can detect that when trying to determine whether or not to read a page. Can we still filter entire row groups, or does this bug affect pages in different row groups as well?

> Do you think this is a blocker for the release? . . . Is it common for libraries like parquet to support reading corrupted files that were corrupted due to bugs in the library?

Yes, I think it is a blocker, though I could be convinced if there were a good reason not to get this in. I don't think the remaining effort is that much work though.

Yes, it is common for libraries to be backward-compatible with data written in older versions, even if the data was written with a bug. Parquet itself has done this in the past to deal with metadata mistakes.

isnotinvain:
OK, those are good points. I think there's more work here to accurately detecting when the bug happened or not, and also documentation in parquet-format and so on.

Since we have fixed the cause of the bug, I'd like to not block on this for the release. The release is getting very close (only PARQUET-251 is left and it's almost done), so I'd like to begin the apache release process as soon as PARQUET-251 is closed. Does that seem reasonable?

rdblue:
Is there a good reason for getting this release out now rather than waiting a day or two to fix this? I can be convinced if there is, but this issue is pretty much the definition of a release blocker to me.

I would also say that this should be noted in the vote thread and probably discussed on the list before that goes out. I think it would be reasonable to -1 the release vote to get this fixed so I recommend making the case and building a consensus before getting to the release vote.

isnotinvain:
OK looking into this in more detail -- we need to also make sure that the state of previous values etc. does not carry over across columns (we don't want to apply the previous from column A when decoding column B). As far as I can tell, Encoding is an enum (singleton) so that's not the right place to track the previous, right? This may not work when a schema has multiple binary columns.

So this really needs to be tracked somewhere where the pages are being iterated through and decoded (similar to the first pass of the PR).

isnotinvain:
Actually I have one more question -- the pages in a row group may use different encodings, right? So we actually have to keep track of the last value from the last time delta encoding was used for a given column, skipping over pages that weren't delta encoded.

rdblue:
@isnotinvain, looks like I got the wrong issue number in #235 and didn't close this. Could you close this now that the fix and tests are in? Thanks!

PR #219, opened 2015-06-19
PARQUET-201: Fix ValidTypeMap being overly strict with respect to OriginalTypes

isnotinvain:


isnotinvain:
@rdblue @tsdeng what do you think? I just removed the checks around OriginalType, I don't think they served any purpose.

tsdeng:
OriginalType was then type stored in Footer? Could you explain more about why it is not needed anymore?

tsdeng:
LGTM +1

PR #220, opened 2015-06-19
PARQUET-284: Clean up ParquetMetadataConverter

isnotinvain:
makes all method static, removes unused thread-unsafe cache, etc.

Turns out the "cache" was only read from *after* rebuilding what needed to be cached... so no performance gain there (and no loss by getting rid of it)

However, I don't know if this will fix the issue mentioned in PARQUET-284, I don't think concurrent access to a HashMap will cause deadlock, it would just cause undefined behavior in reads or maybe ConcurrentModificationException

UPDATE: I'm wrong, it can cause an infinite loop so this should fix the issue https://gist.github.com/rednaxelafx/1081908

UPDATE2: Put the cache back in, made it static + thread safe

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.format.converter;
 
+import static org.apache.parquet.Preconditions.checkNotNull;
 import static org.apache.parquet.format.Util.readFileMetaData;
 import static org.apache.parquet.format.Util.writePageHeader;
 
@@ -34,6 +35,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.parquet.Log;
 import org.apache.parquet.hadoop.metadata.ColumnPath;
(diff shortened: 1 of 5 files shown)


isnotinvain:
@tsdeng @rdblue @SinghAsDev This one is ready to go too

SinghAsDev:
LGTM! Will probably need a rebase if PARQUET-251 gets in before this.

julienledem:
A lot of this is memory size issue more than perf issue.
the set of encodings is often the same over and over.

isnotinvain:
@julienledem as written this cache was never used though, unless i'm misreading this?

isnotinvain:
I see now, the work is always done, but the cached copy is always returned so, so it avoids having long-lived copies of the same thing.

Just using a concurrent or thread-local map here won't necessarily help, the lists inside of the map / set have to themselves be thread-safe too...

isnotinvain:
OK, I added the cacheing back in, using concurrent map, atomic updates, and immutable lists

julienledem:
+1
Another solution to get rid of the cache is to change the metadata format a bit to reduce redundancy.
The serialized metadata would be smaller as well.
(for example not use the column path to identify a column but its index instead)

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -183,54 +200,35 @@ private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGrou
     return converted;
   }
 
-  private static final class EncodingList {
```
tsdeng: @julienledem  Just wanna make sure we don't miss any information after removing this class. What's the original intention of creating this EncodingList instead of using the Set directly

tsdeng:
LGTM +1

PR #221, opened 2015-06-19
PARQUET-311: Fix NPE when debug logging metadata

nezihyigitbasi:
Fixes the issue reported at https://issues.apache.org/jira/browse/PARQUET-311

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
@@ -19,6 +19,7 @@
 package org.apache.parquet.format.converter;
 
 import static java.util.Collections.emptyList;
+import static org.apache.parquet.schema.MessageTypeParser.parseMessageType;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 import static org.apache.parquet.format.CompressionCodec.UNCOMPRESSED;
@@ -34,11 +35,19 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
 import java.util.TreeSet;
 
+import org.apache.parquet.column.statistics.BinaryStatistics;
+import org.apache.parquet.hadoop.metadata.BlockMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -252,4 +267,47 @@ public void randomTestFilterMetaData() {
     }
   }
 
+  @Test
+  public void testNullFieldMetadataDebugLogging() throws NoSuchFieldException, IllegalAccessException, IOException {
```
rdblue: @nezihyigitbasi, thanks for adding a test! I implemented the same fix in #211, but it's much better to have one.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -252,4 +267,47 @@ public void randomTestFilterMetaData() {
...
+
+    deleteFile(testFile, conf);
+
+    createTestFile(testFile, conf);
+
+    //enable debug logging
+    Field debugFlag = Log.class.getField("DEBUG");
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    modifiersField.setInt(debugFlag, debugFlag.getModifiers() & ~Modifier.FINAL);
```
rdblue: Is it necessary to turn on debug logging like this? Why not just test the method call that is failing? I think that is the conversion of the Parquet MR metadata object to JSON. That would simplify this test and avoid these reflection changes.

nezihyigitbasi: Yep I will simplify this. This attempt was just to create the exact scenario of how a user will hit this problem.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
nezihyigitbasi: thanks for pointing that out, I didn't see that your PR already fixes the same issue. For now I will update the unit test below and once your PR is in I can rebase this PR.

nezihyigitbasi: @rdblue updated the unit test.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -54,13 +32,43 @@
...
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.OriginalType;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
 import org.apache.parquet.schema.Type.Repetition;
 import org.apache.parquet.schema.Types;
+import org.junit.Assert;
+import org.junit.Test;
 
-import com.google.common.collect.Lists;
+import java.io.ByteArrayInputStream;
```
rdblue: Could you revert the non-functional import changes? I know it's annoying, but we try to keep these to a minimum to avoid conflicts when merging.

nezihyigitbasi: @rdblue done.

PR #225, opened 2015-06-23
PARQUET-152: Add validation on Encoding.DELTA_BYTE_ARRAY to allow FIX…

spena:
PARQUET-152: Add validation on Encoding.DELTA_BYTE_ARRAY to allow FIXED_LEN_BYTE_ARRAY types.

  * FIXED_LEN_BYTE_ARRAY types are binary values that may use DELTA_BYTE_ARRAY encoding,
    so they should be allowed to be decoded using the same DELTA_BYTE_ARRAY encoding.

@rdblue @nezihyigitbasi  Could you review this fix? 

I executed a test by writing a file that fall backs to DELTA_BYTE_ARRAY encoding, then read the file, and compare the read values with the written values, and it worked fine.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/Encoding.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/Encoding.java
@@ -19,6 +19,7 @@
 package org.apache.parquet.column;
 
 import static org.apache.parquet.column.values.bitpacking.Packer.BIG_ENDIAN;
+import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT32;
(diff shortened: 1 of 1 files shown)


nezihyigitbasi:
@spena this problem was previously causing failures during read benchmarks, this change at least makes the benchmarks go through. 
@rdblue are there any other types that can use the `DELTA_BYTE_ARRAY` encoding?

spena:
ParquetProperties.writerToFallbackTo() shows that only BINARY  and FIXED_LEN_BYTE_ARRAY are supported for DELTA_BYTE_ARRAY:

```
...
case BOOLEAN:
        return new RunLengthBitPackingHybridValuesWriter(1, initialSizePerCol, pageSize);
case BINARY:
case FIXED_LEN_BYTE_ARRAY:
        return new DeltaByteArrayWriter(initialSizePerCol, pageSize);
case INT32:
...
```

rdblue:
Looks good to me. I don't think anything else should use DELTA_BYTE_ARRAY.

PR #228, opened 2015-06-24
PARQUET-317: Fix writeMetadataFile crash when a relative root path is used

stevencanopy:
This commit ensures the fully-qualified path is used prior to calling mergeFooters(..).

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -484,10 +484,9 @@ public class ParquetFileWriter {
    * @throws IOException
    */
   public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers) throws IOException {
-    Path qualifiedPath = FileSystem.get(configuration).makeQualified(outputPath);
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -484,7 +484,8 @@ private static void serializeFooter(ParquetMetadata footer, FSDataOutputStream o
    * @throws IOException
    */
   public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers) throws IOException {
-    ParquetMetadata metadataFooter = mergeFooters(outputPath, footers);
+    Path qualifiedPath = FileSystem.get(configuration).makeQualified(outputPath);
```
rdblue: Won't this throw an exception if the outputPath is in a FS other than the default?

stevencanopy:
@rdblue Ah, good point. I've noticed that there was already code a couple lines down that creates the fully qualified path for outputPath, but wasn't using it for the call to mergeFooters(...). 

I've pushed a new commit that simply moved the use of fully-qualified path prior to the call to mergeFooters.

rdblue:
Thanks for fixing this, @stevencanopy! I've merged it and it will be in the 1.8.0 release.

PR #230, opened 2015-06-30
PARQUET-320: Fix semver problems for parquet-hadoop.

rdblue:
Re-enables semver checks for Parquet packages by removing the parquet/** exclusion that was matching unexpected classes. This also fixes all of the semver problems that have been committed since the check started excluding all Parquet classes.


isnotinvain:
+1, thanks. Sorry, a lot of that was me "cleaning things up"

I think some of these classes might be considered private / internal, but we can make that distinction more clear in 2.0.

PR #231, opened 2015-06-30
PARQUET-146: Move Parquet to Java 7

nezihyigitbasi:
This has been pending for a while. With the other fix referenced in PARQUET-146, I think Parquet can now move to Java 7.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,8 +66,8 @@
   </pluginRepositories>
 
   <properties>
-    <targetJavaVersion>1.6</targetJavaVersion>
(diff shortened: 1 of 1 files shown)


nezihyigitbasi:
@rdblue Is there anything else that needs to be done to move Parquet to Java 7? Will be happy to help with that.

julienledem:
I believe we should move to java 7 too.
Let me check.

rdblue:
+1 for Java 7!

julienledem:
+1

PR #235, opened 2015-07-03
PARQUET-246: File recovery and work-arounds

rdblue:
This is another way to recover data written with the delta byte array problem in PARQUET-246. This builds on @isnotinvain's strategy for solving the problem by adding a method to the encoding to detect it. This version is more similar to the fix for PARQUET-251 and includes a CorruptDeltaByteArrays helper class that uses the writer version. Most of the file changes are to get the file writer version to Encoding and the ColumnReaderImpl.

This also repairs the problem by using a new interface, RequiresPreviousReader, to pass the previous ValuesReader, which is slightly cleaner because the reader doesn't need to expose getter and setter methods.

The problem affects pages written to different row groups, so it was necessary to detect the problem in parquet-hadoop and fail jobs that cannot reconstruct data. The work-around to recover is to set "parquet.split.files" to false so that files are read sequentially. This could be set automatically in isSplittable, but this would require reading all file footers before submitting jobs, which was recently fixed. I think it is a fair compromise to detect the error case and recommend a solution.

This also includes tests for the problem to verify the fix.

Replaces old pull requests: closes #217 closes #235

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
@@ -29,6 +29,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 
 import org.apache.parquet.Log;
+import org.apache.parquet.VersionParser.ParsedVersion;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.ColumnDescriptor;
@@ -40,6 +41,7 @@ import org.apache.parquet.column.page.DataPageV1;
 import org.apache.parquet.column.page.DataPageV2;
 import org.apache.parquet.column.page.DictionaryPage;
 import org.apache.parquet.column.page.PageReader;
+import org.apache.parquet.column.values.RequiresPreviousReader;
 import org.apache.parquet.column.values.ValuesReader;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridDecoder;
 import org.apache.parquet.io.ParquetDecodingException;
@@ -130,6 +132,7 @@ class ColumnReaderImpl implements ColumnReader {
     }
   }
 
+  private final ParsedVersion writerVersion;
   private final ColumnDescriptor path;
   private final long totalValueCount;
   private final PageReader pageReader;
@@ -327,10 +330,11 @@ class ColumnReaderImpl implements ColumnReader {
    * @param path the descriptor for the corresponding column
    * @param pageReader the underlying store to read from
    */
-  public ColumnReaderImpl(ColumnDescriptor path, PageReader pageReader, PrimitiveConverter converter) {
+  public ColumnReaderImpl(ColumnDescriptor path, PageReader pageReader, PrimitiveConverter converter, ParsedVersion writerVersion) {
     this.path = checkNotNull(path, "path");
     this.pageReader = checkNotNull(pageReader, "pageReader");
     this.converter = checkNotNull(converter, "converter");
+    this.writerVersion = writerVersion;
     DictionaryPage dictionaryPage = pageReader.readDictionaryPage();
     if (dictionaryPage != null) {
       try {
@@ -525,8 +529,11 @@ class ColumnReaderImpl implements ColumnReader {
   }
 
   private void initDataReader(Encoding dataEncoding, byte[] bytes, int offset, int valueCount) {
+    ValuesReader previousReader = this.dataColumn;
+
     this.pageValueCount = valueCount;
     this.endOfPageValueCount = readValues + pageValueCount;
+
     if (dataEncoding.usesDictionary()) {
       if (dictionary == null) {
         throw new ParquetDecodingException(
@@ -545,6 +552,11 @@ class ColumnReaderImpl implements ColumnReader {
       dataColumn.initFromPage(pageValueCount, bytes, offset);
     } catch (IOException e) {
       throw new ParquetDecodingException("could not read page in col " + path, e);
+    }
+
+    if (dataEncoding.requiresSequentialReads(writerVersion) && previousReader != null) {
+      // previous reader can only be set if reading sequentially
+      ((RequiresPreviousReader) dataColumn).setPreviousReader(previousReader);
     }
   }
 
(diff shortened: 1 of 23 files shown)


rdblue:
@tsdeng and @spena, could you review?

parquet-column/src/main/java/org/apache/parquet/column/values/RequiresPreviousReader.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values;
+
+public interface RequiresPreviousReader {
```
tsdeng: Not sure if this interface/generation is actually needed.
The actual problem is the DeltaByteArrayReader needs to reset the 'previous' member to fix the bug.
It needs the previous char array, the previous value, not the previous reader. This interface gives an impression the subclass could hold a reference of the previous reader.

How about:
remove this interface,
create a setPreviousValue method in DeltaByteArrayReader.
and call:
setPreviousValue(deltaByteArrayReader.previous)

parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayReader.java
```diff
@@ -75,4 +82,18 @@ public Binary readBytes() {
...
+
+  /**
+   * There was a bug (PARQUET-246) in which DeltaByteArrayWriter's reset() method did not
+   * clear the previous value state that it tracks internally. This resulted in the first
+   * value of all pages (except for the first page) to be a delta from the last value of the
+   * previous page. In order to read corrupted files written with this bug, when reading a
+   * new page we need to recover the previous page's last value to use it (if needed) to
+   * read the first value.
+   */
+  public void setPreviousReader(ValuesReader reader) {
```
tsdeng: the fix is actually setPreviousValue, not setPreviousReader

parquet-column/src/main/java/org/apache/parquet/column/values/RequiresPreviousReader.java
rdblue: Previous is internal state. This method avoids exposing that state to others, allowing access to all private members in the recipient ValuesReader.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
```diff
@@ -546,6 +573,11 @@ private void initDataReader(Encoding dataEncoding, byte[] bytes, int offset, int
     } catch (IOException e) {
       throw new ParquetDecodingException("could not read page in col " + path, e);
     }
+
+    if (dataEncoding.requiresSequentialReads(writerVersion) && previousReader != null) {
```
tsdeng: Is this code only for the particular case of fixing the DeltaByteArrayReader bug?

If it is, then is the generalization of the RequiresPreviousReader necessary?

rdblue: This is more generic. This allows the data encoding to require sequential reads for any reason, which will result in the previous values reader being passed to the new values reader. This is a generic way to make that happen in case we run into it in the future.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
```diff
@@ -546,6 +573,11 @@ private void initDataReader(Encoding dataEncoding, byte[] bytes, int offset, int
...
+      // previous reader can only be set if reading sequentially
+      ((RequiresPreviousReader) dataColumn).setPreviousReader(previousReader);
```
tsdeng: Is there a better way than doing the type casting here?

rdblue: We don't want this to be a normal part of the API, so using an interface like this to carry the method is a good practice. This also allows the reader to expose just the one method and not expose its internal state.

tsdeng: Discussed offline. If we wanna make it general, we should make encoding to provide a SequentialReader rather than type casting the Reader.
If we wanna make it not that generic, we should not let the "requiresSequentialReads" be part of the normal Encoding's API

tsdeng:
+1

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -187,11 +192,31 @@ private void initializeInternalReader(ParquetInputSplit split, Configuration con
             + " in range " + split.getStart() + ", " + split.getEnd());
       }
     }
+
+    checkDeltaByteArrayProblem(footer.getFileMetaData(), configuration, filteredBlocks.get(0));
```
isnotinvain: Hi @rdblue I'm trying to upgrade scalding to parquet 1.8.0 and I'm seeing an `java.lang.IndexOutOfBoundsException: Index: 0, Size: 0` here in a test that runs on a local mini-cluster. 

Is it a valid assumption that this list always has a 0th element? In the case of task-side filtering, it is possible for a task to discover that there are 0 matching row groups right? I think this is a bug, but I want to make sure I'm not just missing something.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -438,6 +443,9 @@ public void testConvertToThriftStatistics() throws Exception {
 
   @Test
   public void testWriteReadStatistics() throws Exception {
+    // this test assumes statistics will be read
+    Assume.assumeTrue(!shouldIgnoreStatistics(Version.FULL_VERSION, BINARY));
```
isnotinvain: Not sure I'm a fan of Assume, this could easily disable this test by accident right?

why do we need this guard?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
isnotinvain: easy fix:
https://github.com/apache/parquet-mr/pull/242

PR #237, opened 2015-07-07
PARQUET-327. Show statistics in the dump output.

tomwhite:


parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
```diff
@@ -260,6 +261,10 @@ public Void visit(DataPageV1 pageV1) {
                 out.format(" DLE:%s", pageV1.getDlEncoding());
                 out.format(" RLE:%s", pageV1.getRlEncoding());
                 out.format(" VLE:%s", pageV1.getValueEncoding());
+                Statistics<?> statistics = pageV1.getStatistics();
+                if (statistics != null) {
+                  out.format(" ST:[%s]", statistics);
+                }
```
liancheng: I think it would be great to show `ST:[null]` when `statistics` is null.

rdblue:
@tomwhite, sorry to take so long to get this reviewed. I didn't realize it was waiting.

I had a slight problem rebasing this test on master. Could you fix the conflict and push an update? Then I'll get this merged. Thanks!

PR #238, opened 2015-07-07
Requested schema should use read schema

tsdeng:
In ThriftRead support requested schema should use schema for read


rdblue:
@tsdeng, can you open a JIRA issue for this and update the summary?

PR #241, opened 2015-07-15
PARQUET-335: Remove Avro check for MAP_KEY_VALUE.

rdblue:
This is not required by the map type spec. This does not affect data
written by the Avro object model because this bug is in the conversion
from a Parquet schema to an Avro schema. Files written with parquet-avro
do not convert the underlying schema because they use the Avro schema.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -303,7 +303,6 @@ public class AvroSchemaConverter {
             }
             GroupType mapKeyValType = parquetGroupType.getType(0).asGroupType();
             if (!mapKeyValType.isRepetition(Type.Repetition.REPEATED) ||
-                !mapKeyValType.getOriginalType().equals(OriginalType.MAP_KEY_VALUE) ||
                 mapKeyValType.getFieldCount()!=2) {
               throw new UnsupportedOperationException("Invalid map type " + parquetGroupType);
             }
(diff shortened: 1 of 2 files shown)


rdblue:
@liancheng could you verify this looks alright? The spec doesn't require MAP_KEY_VALUE at all, so this shouldn't be there.

@isnotinvain can you +1 this after Cheng reviews?

liancheng:
Would be good to add the following test case to `TestAvroSchemaConverter`:

```java
  @Test
  public void testParquetMapWithoutMapKeyValueAnnotation() throws Exception {
    MessageType parquetSchema = MessageTypeParser.parseMessageType(
        "message myrecord {\n" +
            "  required group mymap (MAP) {\n" +
            "    repeated group map {\n" +
            "      required binary key (UTF8);\n" +
            "      required int32 value;\n" +
            "    }\n" +
            "  }\n" +
            "}\n");
    new AvroSchemaConverter().convert(parquetSchema);
  }
```

Otherwise looks good to me.

liancheng:
Actually this test case is probably better:

```java
  @Test
  public void testParquetMapWithoutMapKeyValueAnnotation() throws Exception {
    Schema schema = Schema.createRecord("myrecord", null, null, false);
    Schema map = Schema.createMap(Schema.create(Schema.Type.INT));
    schema.setFields(Collections.singletonList(new Schema.Field("mymap", map, null, null)));
    String parquetSchema =
        "message myrecord {\n" +
            "  required group mymap (MAP) {\n" +
            "    repeated group map {\n" +
            "      required binary key (UTF8);\n" +
            "      required int32 value;\n" +
            "    }\n" +
            "  }\n" +
            "}\n";

    testParquetToAvroConversion(schema, parquetSchema);
  }
```

isnotinvain:
It should work even without the (MAP) part too right? so long as the shape of the schema is a group containing a repeated group of pairs, right?

rdblue:
No, according to the spec, without the MAP annotation the mymap value should be a struct with a single field called map that is a required list of required elements. Those elements should each be a struct in this case with a key and a value.

isnotinvain:
ok

rdblue:
@liancheng thanks! I've added that as the test case and rebased on master. I think this is about ready to go.

rdblue:
@isnotinvain, can you take another look? Thanks!

isnotinvain:
+1

liancheng:
+1

PR #242, opened 2015-07-16
PARQUET-336: Fix ArrayIndexOutOfBounds in checkDeltaByteArrayProblem

isnotinvain:


rdblue:
@isnotinvain, I think you're right on this one and the fix looks good.

Is it easy to add a test for this case? It seems like we should have something that catches problems like this when all of the row groups are filtered out for a task.

isnotinvain:
Yeah, I am not sure how easy this is to test w/o introducing a mini-cluster test (which can be pretty painful to setup)

I can try and see if a narrow unit-test is do-able.

isnotinvain:
Actually, wasn't hard to add a test that hits the issue, seems this issue is hit in local mode as well.

tsdeng:
+1

PR #243, opened 2015-07-16
PARQUET-337 handle binary fields in set/map/list in parquet-scrooge

jaked:
https://issues.apache.org/jira/browse/PARQUET-337


isnotinvain:
+1, @tsdeng look good to you? I thought we were already testing Binary fields -- did we just miss Binary's inside of collections?

tsdeng:
@isnotinvain  you are right, it's only missing for collections. Thanks @jaked for the fix. +1

PR #244, opened 2015-07-16
PARQUET-338: Fix pull request example in README

isnotinvain:
The example PR has the wrong format, it uses [PARQUET-123] instead of PARQUET-123:


tsdeng:
+1

PR #246, opened 2015-07-17
PARQUET-340: MemoryManager: max memory can be truncated

Zariel:
Using float will cause the max heap limit to be limited to 2147483647
due to math.round(float) if used with a large heap. This should be a double
precision to prevent rounding to an int32 before storing into a long.


julienledem:
+1 thanks for the fix

Zariel:
Thanks!

PR #247, opened 2015-07-21
PARQUET-341 improve write performance for wide schema sparse data

tsdeng:
In write path, when there are tons of sparse data, most of time is spent on writing nulls.
Currently writing nulls has the same code path as writing values, which is reclusive traverse all the leaves when a group is null.
Due to the fact that when a group is null all the leaves beneath it should be written with null value with the same repetition level and definition level, we can eliminate the recursion call to get the leaves

This PR caches the leaves for each group node. So when a group node is null, their leaves can be flushed with null values directly.

We tested it with a really wide schema on one of our production data. It improves the performance by ~20%

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
@@ -185,14 +185,17 @@ public class MessageColumnIO extends GroupColumnIO {
       GroupColumnIO  parent = primitive.getParent();
       do {
         getLeafWriters(parent).add(writer);
-      } while ((parent = parent.getParent()) != null);
+        parent = parent.getParent();
+      } while (parent != null);
     }
 
     private List<ColumnWriter> getLeafWriters(GroupColumnIO group) {
-      if (!groupToLeafWriter.containsKey(group)) {
-        groupToLeafWriter.put(group, new ArrayList<ColumnWriter>());
+      List<ColumnWriter> writers = groupToLeafWriter.get(group);
+      if (writers == null) {
+        writers = new ArrayList<ColumnWriter>();
+        groupToLeafWriter.put(group, writers);
       }
-      return groupToLeafWriter.get(group);
+      return writers;
     }
 
     public MessageColumnIORecordConsumer(ColumnWriteStore columns) {
--- a/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
@@ -18,9 +18,12 @@
  */
 package org.apache.parquet.io;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.parquet.Log;
 import org.apache.parquet.column.ColumnWriteStore;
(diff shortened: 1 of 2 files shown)


tsdeng:
@julienledem @isnotinvain could you take a look at this?

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -172,16 +175,36 @@ public boolean isWritten(int i) {
...
     private final ColumnWriter[] columnWriter;
+    /** maintain a map of a group and all the leaf nodes underneath it. It's used to optimize writing null for a group node
+     * all the leaves can be called directly without traversing the sub tree of the group node */
+    private Map<GroupColumnIO, List<ColumnWriter>>  groupToLeafWriter = new HashMap<GroupColumnIO, List<ColumnWriter>>();
     private final ColumnWriteStore columns;
     private boolean emptyField = true;
 
+    private void buildGroupToLeafWriterMap(PrimitiveColumnIO primitive, ColumnWriter writer) {
+      GroupColumnIO  parent = primitive.getParent();
+      do {
```
isnotinvain: how about:

```
  do {
    getLeafWriters(parent).add(writer);
    parent = parent.getParent();
  } while (parent != null)
```

to make this a little easier to read?

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -172,16 +175,36 @@ public boolean isWritten(int i) {
...
+        getLeafWriters(parent).add(writer);
+      } while ((parent = parent.getParent()) != null);
+    }
+
+    private List<ColumnWriter> getLeafWriters(GroupColumnIO group) {
```
isnotinvain: tiny bit more efficient:

```
   private List<ColumnWriter> getLeafWriters(GroupColumnIO group) {
     List<ColumnWriter>() writers = groupToLeafWriter.get(group);
     if (writers == null) {
       writers = new ArrayList<ColumnWriter>();
       groupToLeafWriter.put(group, writers);
     }
     return writers;
   }
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -62,6 +62,7 @@
 
   private MessageType requestedSchema;
   private MessageType fileSchema;
+  private MessageColumnIO columnIO;
```
isnotinvain: why does this need to get held on to? I don't see it being used

isnotinvain:
overall looks good, some minor comments

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
tsdeng: This is actually another problem I found during this PR although not related.
Previouslly, columnIO was not a field, and it gets re-parsed/generated for each row group. This is completely unnecessary.

I am changing it to be generated in the initialize method, and it is reused for each row group.

This change might bring back some performance when the schema is really large.

I can separate this in another PR, what do you think? @isnotinvain

tsdeng:
ping @isnotinvain @julienledem

isnotinvain:
I made two suggestions above, I thought you mentioned that you addressed them but I don't see it in the PR -- do you need to push?

isnotinvain:
Also, now that the columnIO is being cached / re-used, do you know if it is stateful and needs any kind of resetting between row groups? I'm not sure our test coverage for cross-row-group issues is great

tsdeng:
Just pushed, it seems the columnIO is a function of schema that does not contain any state. The RecordConsumer is the one that contains state. I tested it with some real data that spread many row groups and it worked fine...@julienledem could you let me know if I am wrong?

isnotinvain:
+1

PR #248, opened 2015-07-22
PARQUET-342: Updates to be Java 6 compatible

nezihyigitbasi:


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInputOutputFormatWithPadding.java
```diff
@@ -190,6 +189,23 @@ public void testBasicBehaviorWithPadding() throws Exception {
     ParquetFileWriter.BLOCK_FS_SCHEMES.remove("file");
   }
 
+  private List<String> readAllLines(File file) throws IOException {
```
isnotinvain: maybe we can put this in a shared file so it's not duplicated, maybe in org.apache.parquet.Files, next to org.apache.parquet.Strings?

nezihyigitbasi: @isnotinvain Yep makes sense. Updated.

nezihyigitbasi:
@isnotinvain let me know if you have any other comments.

parquet-common/src/main/java/org/apache/parquet/Files.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+//TODO: Use java.nio.file.Files when Parquet is updated to Java 7
+public final class Files {
+  private Files() { }
+
+  public static List<String> readAllLines(File file) throws IOException {
```
isnotinvain: this should accept a charset as an argument and thread it through

isnotinvain:
Sorry, just one last thing but +1 other than that

nezihyigitbasi:
@isnotinvain yep makes sense, updated.

isnotinvain:
+1 thanks!

PR #249, opened 2015-07-24
PARQUET-343 Caching nulls on group node to improve write performance on wide schema sparse data

tsdeng:
For really wide schema with sparse data, If a group node is empty, it could have a huge number of leaves underneath it. Calling writeMull for each leaf every time when it's ancestor group node is null is in-effcient and is bad for data locality in the memory especially when the number of leaves is huge.

Instead, null can be cached on the group node. Flushing is only triggered when a group node becomes non-null from null. This way, all the cached null values will be flushed to the leaf nodes in a tight loop and improves write performance.

We tested this approach combined with PARQUET-341 on a really large schema and gave us ~2X improvement on write performance

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
@@ -155,8 +155,8 @@ public class MessageColumnIO extends GroupColumnIO {
       @Override
       public String toString() {
         return "VistedIndex{" +
-                "vistedIndexes=" + vistedIndexes +
-                '}';
+            "vistedIndexes=" + vistedIndexes +
+            '}';
       }
 
       public void reset(int fieldsCount) {
@@ -176,9 +176,12 @@ public class MessageColumnIO extends GroupColumnIO {
     private final FieldsMarker[] fieldsWritten;
     private final int[] r;
     private final ColumnWriter[] columnWriter;
-    /** maintain a map of a group and all the leaf nodes underneath it. It's used to optimize writing null for a group node
-     * all the leaves can be called directly without traversing the sub tree of the group node */
-    private Map<GroupColumnIO, List<ColumnWriter>>  groupToLeafWriter = new HashMap<GroupColumnIO, List<ColumnWriter>>();
+    /**
+     * maintain a map of a group and all the leaf nodes underneath it. It's used to optimize writing null for a group node
+     * all the leaves can be called directly without traversing the sub tree of the group node
+     */
+    private Map<GroupColumnIO, List<ColumnWriter>> groupToLeafWriter = new HashMap<GroupColumnIO, List<ColumnWriter>>();
+
     class NullValue {
       int r;
       int d;
@@ -188,12 +191,13 @@ public class MessageColumnIO extends GroupColumnIO {
         this.d = d;
       }
     }
+
     private Map<GroupColumnIO, IntArrayList> groupPreviousNullCount = new HashMap<GroupColumnIO, IntArrayList>();
     private final ColumnWriteStore columns;
     private boolean emptyField = true;
 
     private void buildGroupToLeafWriterMap(PrimitiveColumnIO primitive, ColumnWriter writer) {
-      GroupColumnIO  parent = primitive.getParent();
+      GroupColumnIO parent = primitive.getParent();
       do {
         getLeafWriters(parent).add(writer);
         parent = parent.getParent();
@@ -238,7 +242,7 @@ public class MessageColumnIO extends GroupColumnIO {
 
     private void log(Object m) {
       String indent = "";
-      for (int i = 0; i<currentLevel; ++i) {
+      for (int i = 0; i < currentLevel; ++i) {
         indent += "  ";
       }
       logger.debug(indent + m);
@@ -249,7 +253,7 @@ public class MessageColumnIO extends GroupColumnIO {
       if (DEBUG) log("< MESSAGE START >");
       currentColumnIO = MessageColumnIO.this;
       r[0] = 0;
-      int numberOfFieldsToVisit = ((GroupColumnIO)currentColumnIO).getChildrenCount();
+      int numberOfFieldsToVisit = ((GroupColumnIO) currentColumnIO).getChildrenCount();
       fieldsWritten[0].reset(numberOfFieldsToVisit);
       if (DEBUG) printState();
     }
@@ -266,7 +270,7 @@ public class MessageColumnIO extends GroupColumnIO {
     public void startField(String field, int index) {
       try {
         if (DEBUG) log("startField(" + field + ", " + index + ")");
-        currentColumnIO = ((GroupColumnIO)currentColumnIO).getChild(index);
+        currentColumnIO = ((GroupColumnIO) currentColumnIO).getChild(index);
         emptyField = true;
         if (DEBUG) printState();
       } catch (RuntimeException e) {
@@ -287,11 +291,11 @@ public class MessageColumnIO extends GroupColumnIO {
     }
 
     private void writeNullForMissingFieldsAtCurrentLevel() {
-      int currentFieldsCount = ((GroupColumnIO)currentColumnIO).getChildrenCount();
+      int currentFieldsCount = ((GroupColumnIO) currentColumnIO).getChildrenCount();
       for (int i = 0; i < currentFieldsCount; i++) {
         if (!fieldsWritten[currentLevel].isWritten(i)) {
           try {
-            ColumnIO undefinedField = ((GroupColumnIO)currentColumnIO).getChild(i);
+            ColumnIO undefinedField = ((GroupColumnIO) currentColumnIO).getChild(i);
             int d = currentColumnIO.getDefinitionLevel();
             if (DEBUG)
               log(Arrays.toString(undefinedField.getFieldPath()) + ".writeNull(" + r[currentLevel] + "," + d + ")");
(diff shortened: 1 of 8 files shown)


tsdeng:
This PR is on top of https://github.com/apache/parquet-mr/pull/247
Will review/merge this after https://github.com/apache/parquet-mr/pull/247 is merged

tsdeng:
This one is ready, @isnotinvain  could you take a look?

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -46,14 +46,13 @@
 import org.apache.parquet.io.api.RecordMaterializer;
 import org.apache.parquet.schema.MessageType;
 
+import it.unimi.dsi.fastutil.ints.IntArrayList;
 import static org.apache.parquet.Preconditions.checkNotNull;
 
 /**
  * Message level of the IO structure
  *
- *
```
isnotinvain: in general, try to not make whitespace changes in parts of the code that you didn't change, it makes the diff a little easier to read (just for future PRs)

isnotinvain:
Could you add some longer comments the explain:
1) Why do we cache / batch nulls
2) How it works

whenever we put in performance improvements like this that add complexity to the code I think it's worth it to document the why + how so when we (or someone else) come back to it later we are totally lost (as we have been with other corners of the codebase)

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -294,17 +301,33 @@ private void writeNullForMissingFieldsAtCurrentLevel() {
...
+        // only cache the repetition level, the definition level should always be the definition level of the parent node
+        cacheNullForGroup(groupColumnIO, r);
       }
     }
 
-    private void writeNullToLeaves(GroupColumnIO group, int r, int d) {
-      for(ColumnWriter leafWriter: groupToLeafWriter.get(group)) {
-        leafWriter.writeNull(r,d);
+    private void cacheNullForGroup(GroupColumnIO group, int r) {
+      if (!groupNullCache.containsKey(group)) {
```
isnotinvain: this is in the hot path, so we should make this as efficient as possible, so:
```
IntArrayList key = groupNullCache.get(group)
if (groupNullCache == null) {
  key = new IntArrayList()
  groupNullCache.put(group, key);
}
key.add(r);
```

This does only 1 map lookup when the key is already present, and 1 lookup, 1 put when it isn't (which will only happen once anyway)

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -294,17 +301,33 @@ private void writeNullForMissingFieldsAtCurrentLevel() {
...
+    }
+
+    private void writeNullToLeaves(GroupColumnIO group) {
+      IntArrayList nullCache = groupNullCache.get(group);
+      if (nullCache == null || nullCache.isEmpty())
+        return;
+
+      int parentDefinitionLevel = group.getParent().getDefinitionLevel();
+      for (ColumnWriter leafWriter : groupToLeafWriter.get(group)) {
+        for (int repetitionLevel : groupNullCache.get(group)) {
```
isnotinvain: here you call `groupNullCache.get(group)` once per leafWriter but group is constant, and you've already stored it in nullCache, so this can be:

```
int parentDefinitionLevel = group.getParent().getDefinitionLevel();
for (ColumnWriter leafWriter : groupToLeafWriter.get(group)) {
  for (int repetitionLevel : nullCache) {
    leafWriter.writeNull(repetitionLevel, parentDefinitionLevel);
  }
}
```

isnotinvain: the `for x : y` syntax works for:
a) Arrays
b) `java.lang.Iterable`s

IntArrayList *is* an `Iterable<Integer>`, but that's sort of a convenience that will bypass all the care that went into avoiding boxing / unboxing in IntArrayList to start with.

Looks like what you want is the `intIterator` method which will return you a IntIterator: http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/IntIterator.html

so something like this:

```
for (IntIterator iter = nullCache.intIterator(); iter.hasNext();) {
  int repetitionLevel = iter.nextInt();
  leafWriter.writeNull(repetitionLevel, parentDefinitionLevel);
}
```

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -315,28 +338,50 @@ private void setRepetitionLevel() {
     @Override
     public void startGroup() {
       if (DEBUG) log("startGroup()");
+      GroupColumnIO group = (GroupColumnIO) currentColumnIO;
+      // current group is not null, need to flush all the nulls that were cached before
+      if (hasNullCache(group))
```
isnotinvain: use curly braces even for 1 line if statements

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -315,28 +338,50 @@ private void setRepetitionLevel() {
...
       if (DEBUG) printState();
     }
 
+    private boolean hasNullCache(GroupColumnIO group) {
+      IntArrayList nulls = groupNullCache.get(group);
+      return nulls != null && !nulls.isEmpty();
+    }
+
+
+    private void flushCachedNulls(GroupColumnIO group) {
```
isnotinvain: you could just put the if(hasNullCache(group)) part in this method and make startGroup() just always call flushCachedNulls()

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -399,6 +444,12 @@ public void addDouble(double value) {
       if (DEBUG) printState();
     }
 
+
+    //should flush null for all groups
+    @Override
+    public void flush() {
+      flushCachedNulls(MessageColumnIO.this);
```
isnotinvain: is this called at the end of the row group (or page?) to make sure any dangling nulls are properly flushed? 
Do we have a unit test to cover the scenario where the values are:
```
(some non-nulls)....(some nulls)...(some non-nulls)...(some nulls)(end of stream)
----------------------------------------------------------------------^ verify that these trailing nulls are written
```

parquet-column/src/main/java/org/apache/parquet/io/api/RecordConsumer.java
```diff
@@ -125,4 +125,11 @@
    */
   abstract public void addDouble(double value);
 
+  /**
+   * NoOps by default
+   * Subclass class can implement its own flushing logic
+   */
+  public void flush() {
```
isnotinvain: should this just be called close(), and implement Closeable? Is there a reason for this to not be a Closeable? Or is this called multiple times (close() is for one-time closing)

isnotinvain: nevermind this seems to match flush() in ColumnWriteStore

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -154,6 +157,7 @@ private void checkBlockSizeReached() throws IOException {
 
   private void flushRowGroupToStore()
       throws IOException {
+    recordConsumer.flush();
```
isnotinvain: so this flushes a row group, what happens cross-page though? when a page gets compressed and stored in a big byte buffer, that page should contain the trailing nulls right?

isnotinvain:
oh, I also meant to ask, is there any value in cacheing the primitives? I guess only if we add a new method to leafWriters like `writeNull(int numberOfNulls)`

isnotinvain:
also, does this cache nulls across multiple records or, just for a single record?

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
tsdeng: fixed

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
tsdeng: fixed

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
tsdeng: yes, see the diff in parquet-column/src/test/java/org/apache/parquet/io/TestColumnIO.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
tsdeng: page flushing is handled by the ColumnWriter, basically when you ask ColumnWriter to write sth, it will cut the pages automatically

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
isnotinvain: discussed and not needed

isnotinvain:
+1, awesome work!

IgorBerman:
Hi @tsdeng and @isnotinvain 
I'm a bit late for the party, but
have you considered using fastutils object-2-object map for nulls cache?
for sparse schemas this cache's putVal becomes kind of hot point, so I'm wondering if it can be further optimised

PR #250, opened 2015-07-24
[PARQUET-99] Page size check properties

danielcweeks:
Added property to toggle page size check estimation and initial row size checking.


danielcweeks:
@rdblue could you do a quick review of this?  I think you were most recently in this path for the padding feature.  I've actually had confirmation from another company that this fixed their OOM issues, so we're not the only ones affected.

rdblue:
@danielcweeks, I'll have a look at it today.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -228,13 +235,16 @@ public ColumnWriteStore newColumnWriteStore(
...
+          initialRowCountForSizeCheck,
+          estimateNextSizeCheck,
+          writerVersion);
     case PARQUET_2_0:
       return new ColumnWriteStoreV2(
           schema,
           pageStore,
           pageSize,
-          new ParquetProperties(dictionaryPageSizeThreshold, writerVersion, enableDictionary));
+          new ParquetProperties(dictionaryPageSizeThreshold, writerVersion, enableDictionary, initialRowCountForSizeCheck, estimateNextSizeCheck));
```
rdblue: Why are there only changes to ColumnWriteStoreV1 or ColumnWriterV1? Does V2 not need to be changed?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV1.java
```diff
@@ -109,9 +114,11 @@ private void accountForValueWritten() {
           + dataColumn.getBufferedSize();
       if (memSize > pageSizeThreshold) {
         // we will write the current page and check again the size at the predicted middle of next page
-        valueCountForNextSizeCheck = valueCount / 2;
+        if(estimateNextSizeCheck) {
```
rdblue: The behavior seems odd to me. If `estimateNextSizeCheck` is false, then this will calculate the in-memory size for every value that gets written. At a minimum, it seems like that should be done every `initialRowCountForSizeCheck` values, not every value. That initial size must be small enough to not overshoot the target, so I think this strategy is a reasonable trade-off.

I also want to make sure we definitely need this option because it looks like something that isn't particularly useful. Why is the initial row count check insufficient? Is the variance of encoded size too large?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -74,11 +77,15 @@ public static WriterVersion fromString(String name) {
   private final int dictionaryPageSizeThreshold;
   private final WriterVersion writerVersion;
   private final boolean enableDictionary;
+  private final int initialRowCountForSizeCheck;
+  private final boolean estimateNextSizeCheck;
 
-  public ParquetProperties(int dictPageSize, WriterVersion writerVersion, boolean enableDict) {
+  public ParquetProperties(int dictPageSize, WriterVersion writerVersion, boolean enableDict, int initialRowCountForSizeCheck, boolean estimateNextSizeCheck) {
```
rdblue: Just to be sure, this isn't part of the public API, right?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV1.java
danielcweeks: I believe that if `estimateNextSizeCheck` is false, it will always use `valueCountForNextSizeCheck` for the next check which is set to `initialRowCountForSizeCheck`, so I believe it does exactly what you want it to do (i.e. always check at the initial interval without doing an estimate).

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
danielcweeks: I'll take a look at the v2 path.  It's changed from the v1 path and I'm not sure if the logic is the same.

geovations:
We were seeing this issue with very wide columns of text.  While writing the parquet to disk, the writer would run out of memory.  Once we applied this patch to v1.8.1, the memory issues were a thing of the past.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV1.java
danielcweeks: Forget what I said.  Now I see what you're saying.

phoenixhadoop:
When would you fix this bug?Thank you.

danielcweeks:
@phoenixhadoop I'm working on rebasing this to master now.  Hopefully it will make the next release.

danielcweeks:
@rdblue I've updated the pull request to include both v1 and v2 writers.  This new approach preserves the behavior of both readers by default and exposes the ability for you to set values for min and max row counts for page size checking (v1 only supports min as initial when estimate is disabled).

rdblue:
Thanks! I'll take a look.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
```diff
@@ -40,15 +43,23 @@
...
+    this(pageWriteStore, pageSizeThreshold, dictionaryPageSizeThreshold, enableDictionary, DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK, DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK, writerVersion, allocator);
+  }
+
+  public ColumnWriteStoreV1(PageWriteStore pageWriteStore, int pageSizeThreshold, int dictionaryPageSizeThreshold, boolean enableDictionary, int maxRowCountForPageSizeCheck, boolean estimateNextSizeCheck, WriterVersion writerVersion, ByteBufferAllocator allocator) {
     super();
     this.pageWriteStore = pageWriteStore;
     this.pageSizeThreshold = pageSizeThreshold;
     this.dictionaryPageSizeThreshold = dictionaryPageSizeThreshold;
     this.enableDictionary = enableDictionary;
+    this.initialRowCountForPageSizeCheck = maxRowCountForPageSizeCheck;
```
rdblue: I think this should be `minRowCountForPageSizeCheck`. It looks like the name is wrong, but the value passed in by `ParquetProperties` is the min.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV2.java
```diff
@@ -147,6 +148,11 @@ public void endRecord() {
   }
 
   private void sizeCheck() {
+    if(!parquetProperties.isEstimateNextSizeCheck()) {
```
rdblue: This block only sets the next row count for a size check and leaves out checking whether the pages should be written.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
danielcweeks: Good catch.  Actually, it should be "initial" for the sig.  The issue is that v1 writer doesn't enforce a min or max and I wanted to maintain the same behavior.  So v1 will use "min" as "initial" and v2 will support "min/max".

I'll fix the name.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -355,6 +371,12 @@ private static int getMaxPaddingSize(Configuration conf) {
     if (INFO) LOG.info("Writer version is: " + writerVersion);
     int maxPaddingSize = getMaxPaddingSize(conf);
     if (INFO) LOG.info("Maximum row group padding size is " + maxPaddingSize + " bytes");
+    boolean estimateNextSizeCheck = getEstimatePageSizeCheck(conf);
+    if(INFO) LOG.info("Page size checking is: " + (estimateNextSizeCheck ? "estimated" : "constant"));
```
rdblue: Nit: the rest of the codebase puts a space between if and the left paren.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV2.java
danielcweeks: You're right.  I haven't really taken a look at v2 until now and missed that sizeCheck() is also doing the flush.

I'll revisit this.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
danielcweeks: Fixed

rdblue:
I really like how this uses `ParquetProperties` to carry the new settings. I think that's a step in the right direction. But, it looks like we haven't quite moved over to ParquetProperties in some important ways and that makes this need to add the 3 new settings all over the place. This is a bigger issue than just this fix, so I went ahead and did some cleanup and posted a work-in-progress pull request: https://github.com/apache/parquet-mr/pull/297

It's based on a previous version that doesn't have this morning's fixes. What do you think?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV2.java
```diff
@@ -158,20 +159,25 @@ private void sizeCheck() {
...
       if (rowsToFillPage < minRecordToWait) {
         minRecordToWait = rowsToFillPage;
       }
     }
     if (minRecordToWait == Long.MAX_VALUE) {
-      minRecordToWait = MINIMUM_RECORD_COUNT_FOR_CHECK;
+      minRecordToWait = parquetProperties.getMinRowCountForPageSizeCheck();
+    }
+
+    if(parquetProperties.isEstimateNextSizeCheck()) {
```
rdblue: This looks good to me now.

Nit: the comments aren't needed and reference values no longer used. Maybe just change the "will check again halfway" by adding " if it is between min and max"

julienledem:
overall this looks good to me.

rdblue:
I merged this in #297. Thanks for fixing this, Dan!

PR #251, opened 2015-07-28
PARQUET-344: Limit the number of rows per block and per split

QuentinFra:
As said in the JIRA ticket, we use Parquet to store raw metrics data and then query this data with Hadoop-Pig. The issue is that sometimes we end up with small Parquet files (~80mo) that contain more than 300 000 000 rows, usually because of a constant metric which results in a very good compression. Too good. As a result we have a very few number of maps that process up to 10x more rows than the other maps and we lose the benefits of the parallelization.

I introduced two changes here:
* Add a parameter to limit the number of rows in a block (parquet.writer.max.row.count)
* Add a parameter to limit the number of rows in a split (parquet.max.split.record.count) I'd probably need to add some unit tests for this.

I still need to add unit tests for this but I already wanted your feedback. Also the parquet.max.split.record.count parameter uses the split planning using file footer which is marked as deprecated...

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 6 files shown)


QuentinFra:
Please see https://issues.apache.org/jira/browse/PARQUET-344 for more details

PR #252, opened 2015-07-30
PARQUET-346: Minor fixes for PARQUET-350, PARQUET-348, PARQUET-346, PARQUET-345

isnotinvain:
PARQUET-346:
ThriftSchemaConverter throws for unknown struct or union type
This is triggered when passing a StructType that comes from old file metadata

PARQUET-350:
ThriftRecordConverter throws NPE for unrecognized enum values
This is just some better error reporting.

PARQUET-348:
shouldIgnoreStatistics too noisy
This is just a case of way over logging something, to the point that it make the logs unreadable

PARQUET-345
ThriftMetaData toString() should not try to load class reflectively
This is a case where the error reporting itself crashes, which results in the real error message getting lost

### DIFF
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TField;
 import org.apache.thrift.protocol.TList;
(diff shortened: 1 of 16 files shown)


parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
```diff
@@ -445,7 +447,16 @@ public FieldEnumConverter(List<TProtocol> events, ThriftField field) {
 
     @Override
     public void addBinary(final Binary value) {
-      final int id = enumLookup.get(value);
+      final Integer id = enumLookup.get(value);
+
+      if (id == null) {
```
isnotinvain: FYI, as far as I can tell, this implementation is faster than:

```
try {
  final int id = enumLookup.get(value);
} catch (NullPointerException e) {
  throw new ParquetDecodingException(...)
}
```

I thought maybe the above would be better because there's sort of a hint to the branch predictor that the 'normal' case is for this not to throw so predict that... but i did a small benchmark and the if statement seems to be faster

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -57,7 +57,13 @@ public MessageType convert(Class<? extends TBase<?, ?>> thriftClass) {
   }
 
   public MessageType convert(StructType struct) {
-    MessageType messageType = ThriftSchemaConvertVisitor.convert(struct, fieldProjectionFilter);
+    MessageType messageType = ThriftSchemaConvertVisitor.convert(struct, fieldProjectionFilter, true);
+    fieldProjectionFilter.assertNoUnmatchedPatterns();
+    return messageType;
+  }
+
+  public MessageType convertWithoutProjection(StructType struct) {
```
tsdeng: Maybe add a comment to state the struct being passed in should be file schema stored in json, and the isUnion method will nor be called...
Basically just document why it's separated from the normal convert code path in case we forget it in the future

tsdeng:
Minor commet, LGTM

isnotinvain:
@tsdeng updated with comments, and cleaned up usage of ThriftSchemaConverter

tsdeng:
+1

isnotinvain:
Thanks!

PR #254, opened 2015-08-06
fix repeated type and repeated primitives

nguyenvanthan:
Hello,

I made this patch to be able to read protobuf message converted to parquet (with parquet-protobuf) and then consume by pig (with parquet-pig).

If when use ParquetLoader without the patch, it fail on repeated primitives and read only last value on repeated group. I try the loader with and without additional schema.

One error that i have is that a repeated primitive is wrapped into a Pig Bag. This does not work because pig expect a bag to contains tuple, not directly a type.

This patch is not perfect because it wrap repeated field into a tuple and into a bag prefix by "bag_" (declared in a constant) but it allow to use PigLoader without an additional pig schema.

Tested on pig 0.14 and parquet-mr 1.8.1.

I don't know if it's the right way, but i'm open for discussion :-)
I have create an 'issue' here: [Jira PARQUET-354](https://issues.apache.org/jira/browse/PARQUET-354)

Complex schema example (note repeated in root level and in a nested message):

```
message CustomComplexSchema {
  optional binary log_id;
  optional binary user_id (UTF8);
  repeated binary language (UTF8);
  optional boolean male;
  optional group friends {
    repeated group girls {
      optional int32 age;
      optional double weight;
    }
    repeated binary tags (UTF8);
    optional int32 number;
  }
}

```


julienledem:
Hello, Thanks for the contribution. Sorry for the late review.
I would prefer deciding to do the wrapping/unwrapping based on the Parquet schema itself. 
basically you could replace the if startsWith("bag_") by a check that the field is repeated and the type is primitive since the parquet schema is always available.

julienledem:
@nguyenvanthan does this sound good to you?

nguyenvanthan:
@julienledem At which line would you advice to change the check ?

nguyenvanthan:
@julienledem It will be hard to only check type. This patch wrapped the repeated primitive into a bag and we need to name the bag to access it more easily. The final user (like me), will be able to print the converted schema with a simple command ```DESCRIBE``` in his pig script.

PR #255, opened 2015-08-07
PARQUET-355: Add Statistics Test for Parquet Columns

sircodesalotOfTheRound:
In response to PARQUET-251 created an integration test that generates random values and compares the statistics against the values read from a parquet file.

There are two tools classes `DataGenerationContext` and `RandomValueGenerators` which are located in the same package as the unit test. I'm sure there is a better place to put these, but I leave that to your discretion.

Thanks
Reuben

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
@@ -1,0 +1,328 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.statistics;
+
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.PrimitiveType;
+
+import java.math.BigInteger;
+import java.util.*;
+
+public class RandomValueGenerators {
+  private static final String ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+
+  public static abstract class RandomValueGenerator<T extends Comparable<T>> {
+    protected final Random random;
+    protected final PrimitiveType.PrimitiveTypeName forType;
+
+    protected RandomValueGenerator(long seed, PrimitiveType.PrimitiveTypeName forType) {
+      this.random = new Random(seed);
+      this.forType = forType;
+    }
+
+    public boolean shouldGenerateNull() {
+      return (random.nextInt(10) == 0);
+    }
+
+    public int randomInt() { return randomInt(Integer.MAX_VALUE - 1); }
+    public int randomInt(int maximum) {
+      // Maximum may be a random number (which may be negative).
+      return random.nextInt(Math.abs(maximum) + 1);
+    }
+
+    public long randomLong() { return random.nextLong(); }
+    public long randomLong(long maximum) { return randomLong() % maximum; }
+
+    public float randomFloat() { return random.nextFloat(); }
+    public float randomFloat(float maximum) { return random.nextFloat() % maximum; }
+
+    public double randomDouble() { return random.nextDouble(); }
+    public double randomDouble(double maximum) { return random.nextDouble() % maximum; }
+
+    public BigInteger randomInt96(Random random) {
+      return new BigInteger(95, random);
+    }
+
+    public BigInteger randomInt96(BigInteger maximum, Random random) {
+      BigInteger result;
+      while ((result = randomInt96(random)).compareTo(maximum) > 0);
+      return result;
+    }
+
+    public char randomLetter() {
+      return ALPHABET.charAt(randomInt() % ALPHABET.length());
+    }
+
+    public String randomString(int maxLength) {
+      return randomFixedLengthString(randomInt(maxLength));
+    }
+
+    public String randomFixedLengthString(int length) {
+      StringBuilder builder = new StringBuilder();
+      for (int index = 0; index < length; index++) {
+        builder.append(randomLetter());
+      }
+
+      return builder.toString();
+    }
+
+    public int randomInt(int low, int high) {
+      int range = (high - low);
+      return low + randomInt(range);
+    }
+
+    protected abstract T nextValue();
+  }
+
+  public static abstract class RandomBinaryIterable<T extends Comparable<T>> extends RandomValueGenerator<T> {
+    protected final int bufferLength;
+    protected final byte[] buffer;
+
+    public RandomBinaryIterable(long seed, PrimitiveType.PrimitiveTypeName forType, int bufferLength) {
+      super(seed, forType);
+
+      this.bufferLength = bufferLength;
+      this.buffer = new byte[bufferLength];
+    }
+
+    public abstract Binary nextBinaryValue();
+
+    public Binary asReusedBinary(byte[] data) {
+      Arrays.fill(buffer, (byte)0);
+      System.arraycopy(data, 0, buffer, 0, Math.min(data.length, bufferLength));
+      return Binary.fromReusedByteArray(data);
+    }
+  }
+
+  public static class RandomIntGenerator extends RandomValueGenerator<Integer> {
+    private final RandomRange<Integer> randomRange = new RandomRange<Integer>(randomInt(), randomInt());
+    private final int minimum = randomRange.minimum();
+    private final int maximum = randomRange.maximum();
+    private final int range = (maximum - minimum);
+
+    public RandomIntGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.INT32);
+    }
+
+    @Override
+    protected Integer nextValue() {
+      if (!shouldGenerateNull()) {
+        return (minimum + randomInt(range));
+      } else {
+        return null;
+      }
+    }
+  }
+
+  public static class RandomLongGenerator extends RandomValueGenerator<Long> {
+    private final RandomRange<Long> randomRange = new RandomRange<Long>(randomLong(), randomLong());
+    private final long minimum = randomRange.minimum();
+    private final long maximum = randomRange.maximum();
+    private final long range = (maximum - minimum);
+
+    public RandomLongGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.INT64);
+    }
+
+    @Override
+    protected Long nextValue() {
+      if (!shouldGenerateNull()) {
+        return (minimum + randomLong(range));
+      } else {
+        return null;
+      }
+    }
+  }
+
+  public static class RandomInt96Generator extends RandomBinaryIterable<BigInteger> {
+    private final RandomRange<BigInteger> randomRange = new RandomRange<BigInteger>(randomInt96(random), randomInt96(random));
+    private final BigInteger minimum = randomRange.minimum();
+    private final BigInteger maximum = randomRange.maximum();
+    private final BigInteger range = maximum.subtract(minimum);
+
+    private static final int INT_96_LENGTH = 12;
+
+    public RandomInt96Generator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.INT96, INT_96_LENGTH);
+    }
+
+    @Override
+    protected BigInteger nextValue() {
+      if (!shouldGenerateNull()) {
+        return (minimum.add(randomInt96(range, super.random)));
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public Binary nextBinaryValue() {
+      BigInteger value = nextValue();
+      if (value != null) {
+        return asReusedBinary(value.toByteArray());
+      } else {
+        return null;
+      }
+    }
+  }
+
+
+  public static class RandomFloatGenerator extends RandomValueGenerator<Float> {
+    private final RandomRange<Float> randomRange = new RandomRange<Float>(randomFloat(), randomFloat());
+    private final float minimum = randomRange.minimum();
+    private final float maximum = randomRange.maximum();
+    private final float range = (maximum - minimum);
+
+    public RandomFloatGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.FLOAT);
+    }
+
+    @Override
+    protected Float nextValue() {
+      if (!shouldGenerateNull()) {
+        return (minimum + randomFloat(range));
+      } else {
+        return null;
+      }
+    }
+  }
+
+  public static class RandomDoubleGenerator extends RandomValueGenerator<Double> {
+    private final RandomRange<Double> randomRange = new RandomRange<Double>(randomDouble(), randomDouble());
+    private final double minimum = randomRange.minimum();
+    private final double maximum = randomRange.maximum();
+    private final double range = (maximum - minimum);
+
+    public RandomDoubleGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.DOUBLE);
+    }
+
+    @Override
+    protected Double nextValue() {
+      if (!shouldGenerateNull()) {
+        return (minimum + randomDouble(range));
+      } else {
+        return null;
+      }
+    }
+  }
+
+  public static class RandomStringGenerator extends RandomBinaryIterable<String> {
+    private static final int MAX_STRING_LENGTH = 16;
+    public RandomStringGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.BINARY, MAX_STRING_LENGTH);
+    }
+
+    @Override
+    protected String nextValue() {
+      int stringLength = randomInt(15) + 1;
+
+      if (!shouldGenerateNull()) {
+        return randomString(stringLength);
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public Binary nextBinaryValue() {
+      String value = nextValue();
+      if (value != null) {
+        return asReusedBinary(value.getBytes());
+      } else {
+        return null;
+      }
+    }
+  }
+
+  public static class RandomBinaryGenerator extends RandomBinaryIterable<Binary> {
+    private static final int MAX_STRING_LENGTH = 16;
+    public RandomBinaryGenerator(long seed) {
+      super(seed, PrimitiveType.PrimitiveTypeName.BINARY, MAX_STRING_LENGTH);
+    }
+
+    @Override
+    protected Binary nextValue() {
+      if (!shouldGenerateNull()) {
+        for (int index = 0; index < bufferLength; index++) {
+          buffer[index] = (byte)randomInt();
+        }
+
+        return Binary.fromReusedByteArray(buffer);
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public Binary nextBinaryValue() {
+      Binary value = nextValue();
+      if (value != null) {
+        return value;
+      } else {
+        return null;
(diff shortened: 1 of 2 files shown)


sircodesalotOfTheRound:
A brief outline of the Code:

(1) There is one [test function](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R202) that is generates random data using a random block / page size.

(2) The [`DataGenerationContext`](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-8535a47ba33d3dbd78cdad93ba2dbcfcR33) and the nested [`WriteContext`](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-8535a47ba33d3dbd78cdad93ba2dbcfcR44) callback classes use IOC to generate the [file writer](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-8535a47ba33d3dbd78cdad93ba2dbcfcR74), [write the data](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R141) and then finally [test](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R158) the code. 

(3) The tests each use [`RandomValueGenerators`](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R116) which [generates a random list of values](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-bebe531fc94cb73a9f8c83969f2ff5f2R109), these values are then [written to the file](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R145).

(4) On test, [each page is read from the file](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R175). We use a [`PagingValidator`](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R51) class which reads the statistics from each of the pages. 

(5) The value-generator originally used to write the values is then queried for what it believes ought to the the [correct statistics for the entries on this page](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R65). These values are then compared against [what is actually read from the file](https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76#diff-983c41945e25ccffad2e4b8bf766dfb5R66).

If all of the statistics match the anticipated values, then the test passes.

rdblue:
@sircodesalotOfTheRound, thanks for working on this! I like the overall approach of using randomly-generated data, writing an entire file, and validating the pages individually. I had a few comments about how you're doing those tasks, but overall it is a great start. Thanks!

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,387 @@
...
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
+
+public class RandomValueGenerators {
+  private static final String ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+  private static final Random RANDOM = new Random(System.currentTimeMillis());
```
spena: I think we should use a constant value as seed so that we can have reproducible tests with the same random values. If we use a dynamic value, then a bug might appear randomly.

rdblue: I agree, I just put my comments on the commit rather than the pull request by accident. More comments from me here: https://github.com/sircodesalotOfTheRound/parquet-mr/commit/e05447e536dc74708214870e149907e04d5abb76

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,419 @@
...
+      ParquetFileReader reader = new ParquetFileReader(configuration,
+        metadata.getFileMetaData(),
+        super.fsPath,
+        metadata.getBlocks(),
+        metadata.getFileMetaData().getSchema().getColumns());
+
+      PagingValidator<Integer> i32validator = new PagingValidator<Integer>(super.schema, "i32") {
+        @Override
+        public void updateStats(Statistics<Integer> statistics, ValuesReader valuesReader) {
+          statistics.updateStats(valuesReader.readInteger());
```
rdblue: I don't think the validation should call `updateStats`. Otherwise, you're delegating to the implementation you're trying to test. Instead, the test should be that the min value is less-than or equal-to each value in the page. Similarly, you should count nulls and validate that the number matches up after iterating through the page values. That way we are checking the meaning of the stats values independent of the implementation.

I think I mentioned this before, but it was probably overlooked in a sea of comments.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,419 @@
...
+    private final RandomValueGenerators.RandomBinaryGenerator binaryGenerator;
+    private final RandomValueGenerators.RandomFixedBinaryGenerator fixedBinaryGenerator;
+
+    public DataContext(long seed, String path, int blockSize, int pageSize, boolean enableDictionary, ParquetProperties.WriterVersion version) throws IOException {
+      super(path, buildSchema(seed), blockSize, pageSize, enableDictionary, true, version);
+
+      this.random = new Random(seed);
+      this.recordCount = random.nextInt(100000);
+
+      this.randomIntPair = new RandomValueGenerators.RandomPair<Integer>(random.nextInt(), random.nextInt());
```
rdblue: This should probably be named randomIntValueRange?

rdblue: It would also be good to have a few columns that aren't constrained.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  public static class RandomBinaryGenerator extends RandomValueGenerator<String> {
+    public RandomBinaryGenerator(long seed, int total) {
+      super(seed, PrimitiveType.PrimitiveTypeName.BINARY, total);
+    }
+
+    @Override
+    protected List<String> generateValues(int total, String minimum, String maxiumum) {
+      List<String> values = new ArrayList<String>();
```
rdblue: If you seed the random generator with the same values (or add a reset method) then you shouldn't need to keep the values around for the test. That would decrease the chances of hitting an OOM.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+    public static abstract class ValueIterator<TValueType> implements Iterator<TValueType> {
+      private int count;
+
+      public ValueIterator(int count) {
+        this.count = count;
+      }
+
+      @Override
+      public void remove() {
+        throw new NotImplementedException();
```
rdblue: This should be UnsupportedOperationException. NotImplementedException comes from sun.reflect

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      }
+
+      @Override
+      public boolean hasNext() {
+        return (count --> 0);
```
rdblue: This breaks the contract of Iterator. `hasNext` must not advance the iterator, only check if there is another value. So this should be `count != 0`. In addition, I highly discourage using `--` or `++`, especially when the return value is used in an expression.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+import org.apache.parquet.schema.PrimitiveType;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
+
+import java.math.BigInteger;
+import java.util.*;
+
+public class RandomValueGenerators {
+  private static final String ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+
+  public static abstract class RandomValueGenerator<TValueType extends Comparable<TValueType>> implements Iterable<TValueType> {
```
rdblue: Type parameter names should be all caps by convention.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+    public long randomLong(long maximum) { return randomLong() % maximum; }
+
+    public float randomFloat() { return random.nextFloat(); }
+    public float randomFloat(float maximum) { return random.nextFloat() % maximum; }
+
+    public double randomDouble() { return random.nextDouble(); }
+    public double randomDouble(double maximum) { return random.nextDouble() % maximum; }
+
+    public BigInteger randomInt96(Random random) {
+      return new BigInteger((8 * 12), random);
```
rdblue: Why not use 96?

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,515 @@
...
+
+  public static class DataContext extends DataGenerationContext.WriteContext {
+    private static final int MAX_TOTAL_ROWS = 5000000;
+    private static final int SIZEOF_BYTE = 8;
+    private static final int SIZEOF_INT96 = (SIZEOF_BYTE * 12);
+
+    private final Random random;
+    private final int recordCount;
+
+    private final RandomValueGenerators.RandomRange<Integer> randomIntPair;
```
rdblue: Nit: the variable name should be range, not pair.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      @Override
+      public boolean hasNext() {
+        return (count --> 0);
+      }
+    }
+
+    protected final Random random;
+    protected final PrimitiveType.PrimitiveTypeName forType;
+    protected final int bufferLength;
+    protected final int total;
```
rdblue: Total doesn't seem to be accessed anywhere.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      super(seed, forType, total, minimum, maximum, int96Length);
+
+    }
+
+    public abstract byte[] generateValueAsByteArray(T minimum, T maximum);
+
+    byte[] reusedData = new byte[super.bufferLength];
+    private Binary asReusedBinary(byte[] data) {
+      Arrays.fill(reusedData, (byte)0);
+      System.arraycopy(data, 0, reusedData, 0, Math.min(data.length, reusedData.length));
```
rdblue: This should copy up to the length of data, not the reused array, right? Similarly, the reused array is only valid through the length of data, so you should use the factory method where you pass the length.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      return Binary.fromReusedByteArray(reusedData);
+    }
+
+    public Binary asBinary(byte[] string) {
```
rdblue: Is this used? I think this test should reuse byte arrays where possible.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+
+  public static class RandomFixedBinaryGenerator extends RandomBinaryIterable<String> {
+    public RandomFixedBinaryGenerator(long seed, int total, int length) {
+      super(seed, PrimitiveType.PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY, total, length);
+    }
+
+    @Override
+    protected String generateValue(String minimum, String maxiumum) {
+      if (!shouldGenerateNull()) {
+        return randomFixedLengthString(bufferLength);
```
rdblue: Why is bufferLength a field of the superclass and not of this class. This is the only place where it is used, besides the Binary wrapper method, which doesn't need to know the length.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public byte[] generateValueAsByteArray(String minimum, String maximum) {
+      String value = generateValue(minimum, maximum);
```
rdblue: We should add a case that uses random bytes rather than strings. Strings are too safe.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+        return value.getBytes();
+      } else {
+        return null;
+      }
+    }
+  }
+
+
+
+  public static class RandomRange<T extends Comparable<T>> {
```
rdblue: All this does is validate that min is less than max. Do we need a class for this?

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
```diff
@@ -0,0 +1,395 @@
...
+      this.random = new Random(seed);
+      this.forType = forType;
+      this.bufferLength = 0;
+      this.total = total;
+
+      this.minimum = minimum;
+      this.maximum = maximum;
+    }
+
+    public RandomValueGenerator(long seed, PrimitiveType.PrimitiveTypeName forType, int total, TValueType minimum, TValueType maximum, int bufferLength) {
```
rdblue: I don't think total and bufferLength are needed. If you generate the min and max values from the seed, then you don't need to pass those either. Then you could have a single constructor that takes the seed and a type.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,515 @@
...
+      context.write(writer);
+      writer.close();
+
+      context.test();
+      file.delete();
+    }
+  }
+
+  public static abstract class PagingValidator<T extends Comparable<T>> {
+    private static class ComputedStats<T extends Comparable<T>> {
```
rdblue: Sorry if I wasn't clear earlier: I don't think this should calculate stats at all.

Instead, the validation should get the min/max/numNulls and do this:
```java
T min = minFromStats();
T max = maxFromStats();
long numNullsFromStats = numNullsFromStats();
for (T value : readValue(reader)) {
  if (value != null) {
    assertTrue("min should be <= all values", compare(min, value) <= 0);
    assertTrue("min should be >= all values", compare(max, value) >= 0);
  } else {
    numNulls += 1;
  }
  assertEquals("number of nulls should match", numNullsFromStats, numNulls);
}
```

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,515 @@
...
+        public Statistics<T> visit(DataPageV2 dataPageV2) {
+          return (Statistics<T>)dataPageV2.getStatistics();
+        }
+      });
+    }
+
+    private ComputedStats<T> getStatisticsFromPage(final DataPage page) {
+      return page.accept(new DataPage.Visitor<ComputedStats<T>>() {
+        @Override
+        public ComputedStats<T> visit(DataPageV1 dataPageV1) {
```
rdblue: Why is this computing stats instead of getting the stats object written with the page?

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
```diff
@@ -0,0 +1,515 @@
...
+    public void test() throws IOException {
+      Configuration configuration = new Configuration();
+      ParquetMetadata metadata = ParquetFileReader.readFooter(configuration, super.fsPath, ParquetMetadataConverter.NO_FILTER);
+      ParquetFileReader reader = new ParquetFileReader(configuration,
+        metadata.getFileMetaData(),
+        super.fsPath,
+        metadata.getBlocks(),
+        metadata.getFileMetaData().getSchema().getColumns());
+
+      PagingValidator<Integer> i32validator = new PagingValidator<Integer>(super.schema, "i32") {
```
rdblue: Why is this passing super.schema? That seems suspect to me.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound: Oversight. Will fix.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound: Ah sorry. C# convention.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound: Ah ok. C# Convention, thanks.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound:
```
    @Override
    public Iterator<T> iterator() {
      return new ValueIterator<T>(total) {
        @Override
        protected T nextValue() {
          return generateValue(minimum, maximum);
        }
      };
    }
```

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound: Question about this. How are variable length strings represented with re-usable arrays? I get the case for fixed-binary, and int96, but I don't understand how you can reuse an array with variable length strings.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestStatistics.java
sircodesalotOfTheRound: It does both (gets the stats from the page and compares it with the stats computed). However, I see your note above and can fix this.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValueGenerators.java
sircodesalotOfTheRound: Actually, I see what you mean. Since there is no maximum, there's no need for `total` either. I'll remove this.

rdblue:
@sircodesalotOfTheRound, can you update the title of this issue to "PARQUET-355: Add..." so that we can merge it? Thanks!

sircodesalotOfTheRound:
Okay, updated the title. Thanks!

PR #257, opened 2015-08-13
PARQUET-131: Vectorized reader initial implementation

nezihyigitbasi:
This PR implements an initial version of a vectorized reader and adds various benchmarks (`VectorizedReadBenchmarks`) and unit tests (`ParquetVectorReaderTestSuite`) for that. 

Two new public methods are added to the `ParquetReader` and `ParquetRecordReader` classes:
- `RowBatch nextBatch(RowBatch previous, Class<T> clazz)`: used for reading complex types/arbitrary objects and calls the converters to materialize the records.
- `RowBatch nextBatch(RowBatch previous)`: used for reading primitive types and does not call the converters at all.

@rdblue and @julienledem can you please review?

**Ack**: Thank you @zhenxiao and @dongche for your contributions.


parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -229,4 +229,15 @@ public void writeValue(int field, int index, RecordConsumer recordConsumer) {
...
+  public boolean isNull() {
+    boolean isNull = true;
+    for (List<Object> list : data) {
+      if (list.size() > 0) {
+        isNull = false;
+        break;
+      }
+    }
+    return isNull;
+  }
```
julienledem: should this be called isEmpty() ?
```
public boolean isEmpty() {
  for (List<Object> list : data) {
     if (list.size() > 0) {
       return false;
     }
  return true;
}

File: parquet-column/src/main/java/org/apache/parquet/example/data/simple/convert/GroupRecordConverter.java [hunk 2]
```diff
@@ -46,7 +47,8 @@ public void end() {
 
   @Override
   public Group getCurrentRecord() {
-    return root.getCurrentRecord();
+    SimpleGroup simpleGroup = (SimpleGroup) root.getCurrentRecord();
+    return simpleGroup.isNull() ? null : simpleGroup;
```
julienledem: why do we want to skip a record where all values are null?
These are still valid records.

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+
+public abstract class ColumnVector
+{
+  public static final int DEFAULT_VECTOR_LENGTH = 1024;
+  protected Class valueType;
+  public final boolean [] isNull;
+  private int numValues;
+
+  public ColumnVector() {
+    this.isNull = new boolean[DEFAULT_VECTOR_LENGTH];
```
julienledem: it's not DEFAULT if you can't change it

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * @return the number of values in this vector
+   */
+  public int size() {
+    return numValues;
+  }
+
+  void setNumberOfValues(int numValues)
+  {
```
julienledem: indentation

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+        return new FloatColumnVector();
+      case INT32:
+        return new IntColumnVector();
+      case INT64:
+        return new LongColumnVector();
+      case BINARY:
+        return new ByteColumnVector(1);
+      case INT96:
+        //TODO does this always hold
+        return new ByteColumnVector(12);
```
julienledem: it sounds like it should right?

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+      case DOUBLE:
+        return new DoubleColumnVector();
+      case FLOAT:
+        return new FloatColumnVector();
+      case INT32:
+        return new IntColumnVector();
+      case INT64:
+        return new LongColumnVector();
+      case BINARY:
+        return new ByteColumnVector(1);
```
julienledem: why 1? It is variable size in this case. This parameter seems misleading.

parquet-column/src/main/java/org/apache/parquet/io/RecordReader.java
```diff
@@ -40,4 +42,30 @@
...
+  /**
+   * Reads a vector of rows into the given vectors.
+   * This method is for reading Parquet primitive values and
+   * does not call the converters
+   * @param vectors the vectors to fill
+   * @param columnSchemas schemas of the columns in the column vectors
+   * @param current number of records read so far
+   * @param total total number of records in the row group
+   */
+  public void readVectors(ColumnVector[] vectors, MessageType[] columnSchemas, long current, long total) {
```
julienledem: please create a class for { ColumnVector; MessageType } in cases like this. where you have pairs of things instead of co-indexed arrays.

But anyway, this should probably be in a different class. How about creating a VectorizedRecordReader ? It is already supposed to know the types of the columns.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,13 +434,105 @@ public T read() {
...
   @Override
-  public boolean shouldSkipCurrentRecord() {
-    return shouldSkipCurrentRecord;
+  public void readVectors(ColumnVector[] vectors, MessageType[] columnSchemas, long current, long total) {
+      for (int i = 0 ; i < columnSchemas.length; i++) {
+        MessageType schema  = columnSchemas[i];
+        State state = getState(schema);
+        readVector(state.column, vectors[i], current, total);
+      }
   }
```
julienledem: We should have a VectorizedRecordReaderImpl that is preinitialized with the columns to read (columnSchemas).
It would also maintain the state.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,13 +434,105 @@ public T read() {
...
+      if (shouldSkipCurrentRecord) {
+        if (DEBUG) {
+          LOG.debug("skipping record");
+        }
+        vector.isNull[index] = true;
+      }
+      vector.values[index] = record;
+    }
+
+    ((ColumnVector) vector).setNumberOfValues(index);
```
julienledem: why do we need to cast here?

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,13 +434,105 @@ public T read() {
...
+    ColumnDescriptor column = reader.getDescriptor();
+    int maxDefinitionLevel = column.getMaxDefinitionLevel();
+
+    int index = 0;
+    int fixedLenByteArrayPosition = 0;
+    for ( ; index < ColumnVector.DEFAULT_VECTOR_LENGTH; index++, current++) {
+
+      if (current >= total) {
+        vector.setNumberOfValues(index);
+        return;
```
julienledem: just add as an end condition to the for loop?
```index < ColumnVector.DEFAULT_VECTOR_LENGTH || index >= total```

File: parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java [hunk 11]
```diff
@@ -424,13 +434,105 @@ public T read() {
...
+            intColumnVector.values[index] = reader.getInteger();
+            break;
+          case INT64:
+            LongColumnVector longColumnVector = (LongColumnVector) vector;
+            longColumnVector.values[index] = reader.getLong();
+            break;
+          case INT96:
+          case FIXED_LEN_BYTE_ARRAY:
+            ByteColumnVector fixedLenVector = (ByteColumnVector) vector;
+            byte[] fixedLenBinary = reader.getBinary().getBytes();
```
julienledem: getBytesUnsafe: we're copying them anyway.

julienledem: adding a writeTo(byte[] a,  int offset) method to Binary would be nicer

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,13 +434,105 @@ public T read() {
...
+    int fixedLenByteArrayPosition = 0;
+    for ( ; index < ColumnVector.DEFAULT_VECTOR_LENGTH; index++, current++) {
+
+      if (current >= total) {
+        vector.setNumberOfValues(index);
+        return;
+      }
+
+      if (reader.getCurrentDefinitionLevel() == maxDefinitionLevel) {
+        switch (column.getType()) {
```
julienledem: this could be columnVector.readFrom(reader);
each columnVector would have the right version.

parquet-column/src/main/java/org/apache/parquet/io/vector/VectorizedReader.java
```diff
@@ -0,0 +1,45 @@
...
+public interface VectorizedReader<T> extends Closeable {
+
+  /**
+   * Reads the next batch of rows. This method is used for reading primitive types
+   * and does not call the converters at all.
+   * @param previous a row batch object to be reused by the reader if possible
+   * @return the row batch that was read
+   * @throws java.io.IOException
+   */
+  RowBatch nextBatch(RowBatch previous) throws IOException;
```
julienledem: The record reader is stateful, why not this one?

parquet-column/src/main/java/org/apache/parquet/io/vector/VectorizedReader.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Reads the next batch of rows. This method is used for reading complex types
+   * or arbitrary objects and calls the converters eventually to materialize the record.
+   * @param previous a row batch object to be reused by the reader if possible
+   * @param clazz the type of the record that will be filled into the column vectors of previous
+   * @return the row batch that was read
+   * @throws java.io.IOException
+   */
+  //TODO what if different columns of previous have different complex types
+  RowBatch nextBatch(RowBatch previous, Class<T> clazz) throws IOException;
```
julienledem: it knows the class at construction right?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -257,4 +259,64 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+        current += vector.size();
+
+        //TODO filtering records while batch reading
+        //TODO how to handle this case?
+//        if (currentValue == null) {
+//          // only happens with FilteredRecordReader at end of block
+//          current = totalCountLoadedSoFar;
+//          if (DEBUG) LOG.debug("filtered record reader reached end of block");
+//          continue;
+//        }
```
julienledem: maybe we need a separate VectorizedRecordReader that does not support filter and returns  ObjectColumnVector<T> instead of Ts

julienledem: it sounds like to implement filter we would:
 - read the batch
 - evaluate the filter on the batch
 - filter down the batch. (or wrap columns into filtered ones that have the list of visible indexes)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -257,4 +259,64 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+
+        recordFound = true;
+      } catch (RuntimeException e) {
+        throw new ParquetDecodingException(format("Can not read value at %d in block %d in file %s", current, currentBlock, file), e);
+      }
+    }
+    return true;
```
julienledem: should it return the record count?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -257,4 +259,64 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
     return Collections.unmodifiableMap(setMultiMap);
   }
 
+  //TODO maintain unmaterializableRecordCounter properly
+  public boolean nextBatch(ObjectColumnVector<T> vector) throws IOException {
```
julienledem: do we need this one? It seems that it is the same as using the regular record reader in a tight loop.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -51,6 +61,8 @@
   private final Configuration conf;
   private final Iterator<Footer> footersIterator;
   private final Filter filter;
+  private final GlobalMetaData globalMetaData;
```
julienledem: not used ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -114,6 +126,10 @@ private ParquetReader(Configuration conf,
     List<FileStatus> statuses = Arrays.asList(fs.listStatus(file, HiddenFileFilter.INSTANCE));
     List<Footer> footers = ParquetFileReader.readAllFootersInParallelUsingSummaryFiles(conf, statuses, false);
     this.footersIterator = footers.iterator();
+    globalMetaData = ParquetFileWriter.getGlobalMetaData(footers);
+    MessageType schema = globalMetaData.getSchema();
+    Map<String, Set<String>> extraMetadata = globalMetaData.getKeyValueMetaData();
+    readContext = readSupport.init(new InitContext(conf, extraMetadata, schema));
```
julienledem: this happens in reader.initialize in the record case. Let's make this consistent. It should happen in the same class in all cases.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -133,6 +149,97 @@ public T read() throws IOException {
...
+      rowBatch = new RowBatch();
+    }
+
+    ObjectColumnVector<T> columnVector;
+    if (rowBatch.getColumns() == null) {
+      columnVector = ColumnVector.ofType(clazz);
+      rowBatch.setColumns(new ColumnVector[] { columnVector });
+    } else {
+      columnVector = (ObjectColumnVector<T>) rowBatch.getColumns()[0];
+    }
```
julienledem: this logic should probably be in the RowBatch constructor.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -133,6 +149,97 @@ public T read() throws IOException {
...
+  }
+
+  /**
+   * Reads the next batch of rows. This method is used for reading primitive types
+   * and does not call the converters at all.
+   * @param previous a row batch object to be reused by the reader if possible
+   * @return the row batch that was read
+   * @throws java.io.IOException
+   */
+   public RowBatch nextBatch(RowBatch previous) throws IOException {
```
julienledem: we use "nextBatch" in different contexts to do different things. Let's add a verb in there.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -198,6 +206,10 @@ private void initializeInternalReader(ParquetInputSplit split, Configuration con
     }
 
     MessageType fileSchema = footer.getFileMetaData().getSchema();
+    Map<String, String> fileMetaData = footer.getFileMetaData().getKeyValueMetaData();
+    readContext = readSupport.init(new InitContext(configuration, InternalParquetRecordReader
+        .toSetMultiMap(fileMetaData), fileSchema));
```
julienledem: this should be moved to a utility class

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -227,6 +239,65 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+   * Reads the next batch of rows. This method is used for reading complex types
+   * or arbitrary objects and calls the converters eventually to materialize the record.
+   * @param previous a row batch object to be reused by the reader if possible
+   * @param clazz the class of the record type that will be filled into the column vector
+   * @return the row batch that was read
+   * @throws java.io.IOException
+   */
+  public RowBatch nextBatch(RowBatch previous, Class<T> clazz) throws IOException {
+    throw new UnsupportedOperationException("Reading a batch of rows of complex types is not supported");
+  }
```
julienledem: let's remove this if we don't use it. I don't see how the lower layer would know about T anyway.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/vector/TestParquetVectorReaderGZIP.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+
+import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_1_0;
+import static org.apache.parquet.hadoop.metadata.CompressionCodecName.GZIP;
+
+@RunWith(Parameterized.class)
+public class TestParquetVectorReaderGZIP extends TestParquetVectorReader
+{
```
julienledem: indentation

julienledem:
let's generate the tests files in the test rather than checking in binary files:
parquet-hadoop/src/test/resources/map_string_string.parquet
etc

julienledem:
I think we should separate the vectorized and record reader a little more.
The RecordReader seems to have too many responsibilities right now

julienledem:
Thanks for the contribution, see my comments above.

davies:
@nezihyigitbasi I'm also very interested in this one, thanks for the hard work! Can you share some numbers from the benchmark? And how much work left in order to use the vectorvised interface in ParquetInputFormat (for getting the performance improvements out of box)?

parquet-column/src/main/java/org/apache/parquet/example/data/simple/convert/GroupRecordConverter.java
nezihyigitbasi: Good point. This seems like some leftover code so I removed the relevant commit, which also removes the `isNull()` method from `SimpleGroup` (related to your comment above).

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
nezihyigitbasi: I named it as DEFAULT in case we make it configurable in the future. I can change the name if you really want it.

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
nezihyigitbasi: fixed

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
nezihyigitbasi: just wanted to make sure as I don't trust names.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
nezihyigitbasi: Because `ObjectColumnVector` (type of the `vector` variable) is in a different package than `RecordReaderImplementation` and `ColumnVector` and compilation fails with
```
parquet-mr/parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java:[476,11] cannot find symbol
  symbol:   method setNumberOfValues(int)
  location: variable vector of type org.apache.parquet.io.vector.ObjectColumnVector<T>
[INFO] 1 error
```

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
nezihyigitbasi: updated the loop termination condition to get rid of this check.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
nezihyigitbasi: Added a `readFrom` method, but we still need the switch/case to handle INT96 and FIXED_LEN_BYTE_ARRAY in a special way. The switch statement is simplified anyway.

parquet-column/src/main/java/org/apache/parquet/io/vector/VectorizedReader.java
nezihyigitbasi: The class implementing this interface (`VectorizedParquetRecordReader`) wraps a `ParquetRecordReader`, which is stateful.

parquet-column/src/main/java/org/apache/parquet/io/vector/VectorizedReader.java
nezihyigitbasi: You mean from the type parameter `T`? Then, no, since that type info is erased after compile time, we cannot know the type at constructor unless we move the `Class<T> clazz` parameter to the constructor. Can we?

Having said that I am also not sure whether this is the right signature to use to read a vector of objects, maybe we should name this as `nextObjectBatch` or sth. Any suggestions?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
nezihyigitbasi: you mean the remaining record count? And how will that be useful? If you mean the number of records read, then it is already in `vector.size()`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
nezihyigitbasi: Yep this is a wrapper around the regular record reader and `ParquetReader::readVector(ObjectColumnVector<T> vector)` uses this method (as it only has access to the `InternalParquetRecordReader`, not to the underlying `RecordReader`. Do you have a suggestion?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
nezihyigitbasi: yep, localized it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
nezihyigitbasi: This is done in the constructor as `readContext` is final and can only be assigned in the constructor. Do we want to make this non-final to make things consistently initialized in `initialize()`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
nezihyigitbasi: fixed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
nezihyigitbasi: Not sure whether there is an existing utility class that is suitable to contain this logic, so instead of creating a new utility class specific to this logic how about we add a static utility method to `ReadSupport`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
nezihyigitbasi: fixed.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/vector/TestParquetVectorReaderGZIP.java
nezihyigitbasi: fixed.

nezihyigitbasi:
@davies in my microbenchmarks I saw ~2x improvement for primitive types (since for primitives the current implementation bypasses the converters) while the performance was roughly the same for complex types. As for the readiness of the implementation, we are using it in production at Netflix (with Presto) for some time. And regarding the readiness of this PR, I am still working on addressing the comments.

nezihyigitbasi:
@julien Thanks for the comments, I have made a first pass over them. Regarding reducing the responsibility of `RecordReader`, I moved the vector-related interface to a new interface (`VectorizedRecordReader`, see e3feae1) and created a new `VectorizedRecordReaderImplementation` class that implements it, but I still keep the implementation in `RecordReaderImplementation` as I didn't want `VectorizedRecordReaderImplementation` to replicate the existing state management logic. 

If we really want to get rid of any vector logic in `RecordReaderImplementation`, one option is to replicate the class hierarchy of the non-vector reads for the vector reads (have a Vectorized* for every interface/class such as record reader, record reader impl, parquet reader, etc.). But, this option may perform worse for primitive types if we simply wrap `RecordReaderImplementation::read()` in the vectorized readers (currently for primitives we don't simply wrap the `read()` method, instead we read the primitives directly using the column readers). Do you have any suggestions?

davies:
@nezihyigitbasi That's great, thanks!

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/VectorizedReadBenchmarks.java
```diff
@@ -0,0 +1,266 @@
...
+      Group group;
+      while ((group = defaultReader.read()) != null) {
+        blackhole.consume(group.getBinary("binary_field", 0));
+        blackhole.consume(group.getInteger("int32_field", 0));
+        blackhole.consume(group.getLong("int64_field", 0));
+        blackhole.consume(group.getBoolean("boolean_field", 0));
+        blackhole.consume(group.getFloat("float_field", 0));
+        blackhole.consume(group.getDouble("double_field", 0));
+        blackhole.consume(group.getBinary("flba_field", 0));
+        blackhole.consume(group.getInt96("int96_field", 0));
```
liancheng: All the `getXxx(fieldName, ordinal)` methods require a hash map lookup, which can be unfair for the original `ColumnReader` API in this benchmark.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -227,6 +239,45 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+
+    RowBatch rowBatch = previous;
+    if (rowBatch == null) {
+      rowBatch = new RowBatch(new ColumnVector[nColumns]);
+    }
+
+    columnVectors = rowBatch.getColumns();
+    MessageType[] columnSchemas = new MessageType[nColumns];
+    for (int i = 0; i < nColumns; i++) {
+      ColumnVector columnVector = columnVectors[i];
```
spena: If RowBatch 'previous' is passed as non-null and with less number of columns than 'requestedSchema', then I think this line will throw an ArrayIndexOutOfBoundException. Being this a public method, we should validate that both 'previous' and 'requestedSchema' have the same number of columns and throw a better error message otherwise.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
```diff
@@ -133,6 +148,83 @@ public T read() throws IOException {
...
+     for (int i = 0; i < nColumns; i++) {
+       ColumnVector columnVector = columnVectors[i];
+       columnSchemas[i] = new MessageType(requestedSchema.getFieldName(i), requestedSchema.getType(i));
+
+       if (columnVector == null) {
+         columnVector = ColumnVector.from(columns.get(i));
+       }
+
+       rowBatch.getColumns()[i] = columnVector;
+     }
```
spena: This loop is used to initialize column vectors (once) + column schemas that will be passed to readVectors() everytime we call nextBatch(). nextBatch() might be called many times, so this could add performance overhead. Can this loop be a one-time initialization code? Maybe creating a 'public static RowBatch RowBatch.createFromSchema(MessageType schema)' will allow us to do a one-time initialization? Any thoughts?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
nezihyigitbasi: makes sense.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
nezihyigitbasi: I am not sure we should do it (at least in this PR) for two reasons. 1) We don't really know whether this is an overhead that we should worry about before measuring it. 2) It will complicate the logic here.

If you really think we should optimize this maybe we should open a jira for it and fix it once this gets merged.

nezihyigitbasi:
@julienledem I have addressed most of your comments and fixed the unit tests. Can you take another look?

parquet-column/src/main/java/org/apache/parquet/io/vector/VectorizedReader.java
rdblue: I agree, we should distinguish between the batch operations that return column vectors and batch operations that return complex objects. I wrote down `nextRecordBatch` as a suggestion.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
rdblue: I don't think this is needed and should not be duplicated between this class and the `InternalParquetRecordReader` class. It is used here to create new MessageType objects that are passed into `VectorizedRecordReaderImplementation#readVectors` but I think we need to change that API, probably using `ColumnDescriptor[]` instead of `MessageType`.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -18,15 +18,8 @@
  */
 package org.apache.parquet.io;
 
-import java.util.ArrayList;
```
rdblue: Minor: reordering imports make it difficult to move patches around and backport things.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
rdblue: The problem is that setNumberOfValues is package-private. I actually don't see why it even works to cast like this. If we intend to use this outside of the package, I think it should be public. The only public API part should be the ColumnVector interface, while the abstract base class (that carries this method) should be considered private API. We'll have to separate those two if they are currently just one abstract class.

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,15 +428,68 @@ public T read() {
...
+  private State getState(MessageType schema) {
+    for (State state : states) {
+        if (schema.containsField(state.fieldPath[0])) {
+          return state;
+        }
+    }
+    throw new IllegalArgumentException("Couldn't find state for schema " + schema);
+  }
+
+  protected void readVectors(ColumnVector[] vectors, MessageType[] columnSchemas, long current, long total) {
```
rdblue: `MessageType` is used for root schemas. I'm not sure why one would ever have an array of them.

rdblue:
Thanks for all the work on this @nezihyigitbasi!

I took a close look at this yesterday and I think there are a couple of things we should change. The `VectorizedReader` and supporting classes, like `RowBatch`, mix two modes of operation that I think we should keep separate: accessing the underlying columns in vectors and deserializing and accessing converted records in batches. I don't see a good reason to mix the two together and it complicates the APIs: what happens when the vectorized and non-vectorized calls are mixed? What about when batch object method calls are mixed with column-access calls?

Separating the those two would be a good thing for the API. For example, it is awkward to use `RowBatch` to hold a batch of objects because the type information is lost. Similarly, I think the right level for accessing vectors of column data is from the `ColumnReadStore` rather than through the high-level readers. What about adding a `ColumnReadStore#nextVectorBatch()` method or similar?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/VectorizedParquetRecordReader.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    parquetRecordReader.close();
+  }
+
+  @Override
+  public RowBatch nextBatch(RowBatch previous, Class clazz) throws IOException {
+    throw new UnsupportedOperationException("Reading a batch of rows of complex types is not supported");
```
rdblue: This is one thing that would be fixed if we separated batch reads from vector-based column access.

parquet-column/src/main/java/org/apache/parquet/io/VectorizedRecordReaderImplementation.java
```diff
@@ -0,0 +1,55 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.io;
+
+import org.apache.parquet.io.vector.ObjectColumnVector;
+import org.apache.parquet.schema.MessageType;
+
+class VectorizedRecordReaderImplementation<T> extends VectorizedRecordReader<T> {
```
rdblue: Several classes use this pattern where the original class implements new methods, but they are made available in a Vectorized version. Why not make the original implement `VectorizedReader`?

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.parquet.io.vector.ByteColumnVector;
+import org.apache.parquet.io.vector.DoubleColumnVector;
+import org.apache.parquet.io.vector.FloatColumnVector;
+import org.apache.parquet.io.vector.IntColumnVector;
+import org.apache.parquet.io.vector.LongColumnVector;
+import org.apache.parquet.io.vector.ObjectColumnVector;
+
+public abstract class ColumnVector
+{
+  public static final int MAX_VECTOR_LENGTH = 1024;
```
rdblue: Why hard code this and not make it a setting? Does it never need to change?

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,97 @@
...
+  protected Class valueType;
```
rdblue: This is only used by subclasses to set the valueType in their constructors. I think this should be passed into the `ColumnVector` constructor (`super(boolean.class)`) and set as a final variable.

parquet-column/src/main/java/org/apache/parquet/io/vector/ByteColumnVector.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.parquet.io.vector;
+
+import org.apache.parquet.column.ColumnReader;
+import org.apache.parquet.io.ColumnVector;
+
+public class ByteColumnVector extends ColumnVector
+{
+  public final byte[][] values;
```
rdblue: Why not use `Binary[]`? Using `getBytesUnsafe` will copy the underlying bytes in many cases; for example values from a dictionary or other underlying binary stream are wrapped in `Binary` using the array slice method, which always copies in this situation.

parquet-column/src/main/java/org/apache/parquet/io/vector/ObjectColumnVector.java
```diff
@@ -0,0 +1,41 @@
...
+  public final T[] values;
+
+  public ObjectColumnVector(Class<T> clazz) {
+    this.valueType = clazz;
+    values = (T[]) newInstance(clazz, MAX_VECTOR_LENGTH);
+  }
+
+  @Override
+  public void doReadFrom(ColumnReader reader, int index) {
+  }
```
rdblue: I was a little confused by this method being empty at first, but I think that this class is always used for complex (converted) objects. In that case, it's not really a `ColumnVector`. At a minimum, I think this should throw an exception because the vector is used as a `ColumnVector`, but I think the better option is to replace this with a `RecordBatch<T>` object that doesn't inherit from `ColumnVector` at all.

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.parquet.column.ColumnReader;
+import org.apache.parquet.io.vector.BooleanColumnVector;
+import org.apache.parquet.io.vector.ByteColumnVector;
+import org.apache.parquet.io.vector.DoubleColumnVector;
+import org.apache.parquet.io.vector.FloatColumnVector;
+import org.apache.parquet.io.vector.IntColumnVector;
+import org.apache.parquet.io.vector.LongColumnVector;
+import org.apache.parquet.io.vector.ObjectColumnVector;
+
+public abstract class ColumnVector
```
rdblue: I'd like to see this separated into a public interface and an abstract base class. That way, the `setNumberOfValues` and `readFrom` methods could be hidden and internal.

parquet-column/src/main/java/org/apache/parquet/io/ColumnVector.java
```diff
@@ -0,0 +1,97 @@
...
+    return new ObjectColumnVector<T>(clazz);
+  }
+
+  /**
+   * Reads a single value from the given reader to the specified index into the vector
+   * @param reader
+   * @param index
+   */
+  public void readFrom(ColumnReader reader, int index) {
+    if (index >= MAX_VECTOR_LENGTH) {
```
rdblue: This is checked elsewhere. I think the only reason for the check here is that it is a public API method, but if this were separated into a public interface and a private base implementation we could get rid of this check (since we expect this to be in a tight loop).

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -424,15 +428,68 @@ public T read() {
...
+      for (int i = 0 ; i < columnSchemas.length; i++) {
+        MessageType schema  = columnSchemas[i];
+        State state = getState(schema);
+        readVector(state.column, vectors[i], current, total);
+      }
+  }
+
+  protected void readVector(ObjectColumnVector<T> vector, long current, long total) {
```
rdblue: I think this API is pretty difficult to use. It looks like the purpose is to allow the caller to control this loop, but this passes information about the underlying reader back in from the outside.

I think a cleaner approach would be to delegate to the `ColumnReadStore` to return vectors in batches that align with record boundaries.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -256,4 +260,64 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+  public boolean nextBatch(ColumnVector[] vectors, MessageType[] columns) throws IOException {
+    boolean recordFound = false;
+
+    while (!recordFound) {
+      // no more records left
+      if (current >= total) { return false; }
+
+      try {
+        checkRead();
+        ((VectorizedRecordReader) recordReader).readVectors(vectors, columns, current, totalCountLoadedSoFar);
```
rdblue: Doesn't this assume that there are no repeated values? I don't see this checking repetition level anywhere.

rdblue:
How does this handle nesting? It looks like in `RecordReaderImplementation`, the definition level is checked so you get a vector of nulls and values in the right places, but the repetition level info seems to be ignored. I think that means this would only work with flat columns.

I think we could easily add repetition info for the vector data. Reassembly would be possible if the total number of records in a vector was counted by the number of root records. That would make vectors more complicated: we would have to either cut off a vector at fewer than 1024 root records or use more tables as needed to fit all of the data. Still, I think we should account for nested records.

PR #258, opened 2015-08-16
PARQUET-359: Removes existing _common_metadata when fails to write summary files

liancheng:
When appending Parquet files to a directory containing existing Parquet data together with their summary files, new data may contain different but compatible schema from existing data. Ideally, summary files should be updated and contain a merged version of schema. However, `ParquetOutputCommitter` may fail to write summary files if new files and old files contain conflicting user-defined metadata. In this case, we should remove existing `_metadata` as well as `_common_metadata`.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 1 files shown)


julienledem:
+1

julienledem:
there's a typo in the description. Could you fix it? PARQUE*T*-359

liancheng:
@julienledem Oops, thanks. Fixed.

julienledem:
thanks @liancheng. Could you rebase your branch as well?

liancheng:
@julienledem Actually I just realized that #277 has already addressed this issue. So I'm closing this one. Thanks for the review!

julienledem:
thanks!

PR #259, opened 2015-08-18
PARQUET-360: Handle all map key types with cat tool's json dump

nezihyigitbasi:
When dumping a parquet map with `parquet-cat --json` it throws a class cast exception as it doesn't properly handle all map key types.

```
java.lang.ClassCastException: [B cannot be cast to java.lang.String
	at org.apache.parquet.tools.read.SimpleMapRecord.toJsonObject(SimpleMapRecord.java:34)
	at org.apache.parquet.tools.read.SimpleRecord.toJsonValue(SimpleRecord.java:119)
	at org.apache.parquet.tools.read.SimpleRecord.toJsonObject(SimpleRecord.java:112)
	at org.apache.parquet.tools.read.SimpleRecord.prettyPrintJson(SimpleRecord.java:106)
	at org.apache.parquet.tools.command.CatCommand.execute(CatCommand.java:76)
	at org.apache.parquet.tools.Main.main(Main.java:222)
[B cannot be cast to java.lang.String
```

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleMapRecord.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleMapRecord.java
@@ -45,10 +45,13 @@ public class SimpleMapRecord extends SimpleRecord {
     return result;
   }
 
-  private String keyToString(Object kvValue) {
+  String keyToString(Object kvValue) {
+    if (kvValue == null) {
+      return "null";
+    }
+
     Class<?> type = kvValue.getClass();
     if (type.isArray()) {
-
       if (type.getComponentType() == boolean.class) {
         return Arrays.toString((boolean[]) kvValue);
       }
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleMapRecord.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleMapRecord.java
@@ -20,6 +20,7 @@ package org.apache.parquet.tools.read;
 
 import com.google.common.collect.Maps;
 
+import java.util.Arrays;
 import java.util.Map;
 
 public class SimpleMapRecord extends SimpleRecord {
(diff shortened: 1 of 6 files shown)


parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleMapRecord.java
```diff
@@ -30,14 +31,52 @@ protected Object toJsonObject() {
...
+
+  private String keyToString(Object kvValue) {
+    Class<?> type = kvValue.getClass();
+    if (type.isArray()) {
+
+      if (type.getComponentType() == boolean.class) {
+        return Arrays.toString((boolean[]) kvValue);
+      }
+      else if (type.getComponentType() == byte.class) {
+        return Arrays.toString((byte[]) kvValue);
```
rdblue: I think this should use the BinaryNode that is used elsewhere to convert a Binary or byte[] to JSON. That will base64-encode the data so it is still binary rather than creating an array node.

nezihyigitbasi:
@rdblue updated, thanks.

rdblue:
Sorry, I should have mentioned this earlier: can you also add a test or two that exercises the new keyToString function?

nezihyigitbasi:
@rdblue yeah totally forgot to add them. updated.

parquet-tools/src/test/java/org/apache/parquet/tools/read/TestSimpleMapRecord.java
```diff
@@ -0,0 +1,38 @@
+package org.apache.parquet.tools.read;
```
rdblue: Looks like you forgot to add a license header to this file and that's causing Travis CI to fail.

nezihyigitbasi: added

rdblue:
Just tried to merge, but the format is wrong. Could you remove the square brackets from this PR title? Thanks!

nezihyigitbasi:
@rdblue updated the title.

julienledem:
looks good to me. 
@wesleypeck did you have comments?

PR #261, opened 2015-08-19
PARQUET-361: Add semver prerelease logic.

rdblue:
This also adds more versions where PARQUET-251 is fixed.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -41,6 +41,8 @@ public class CorruptStatistics {
   // the bug involved writing invalid binary statistics, so stats written prior to this
   // fix must be ignored / assumed invalid
   private static final SemanticVersion PARQUET_251_FIXED_VERSION = new SemanticVersion(1, 8, 0);
+  private static final SemanticVersion CDH_5_PARQUET_251_FIXED_START = new SemanticVersion(1, 5, 0, null, "cdh5.5.0", null);
+  private static final SemanticVersion CDH_5_PARQUET_251_FIXED_END = new SemanticVersion(1, 5, 0);
 
   /**
    * Decides if the statistics from a file created by createdBy (the created_by field from parquet format)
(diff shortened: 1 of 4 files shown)


parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -41,6 +41,8 @@
   // the bug involved writing invalid binary statistics, so stats written prior to this
   // fix must be ignored / assumed invalid
   private static final SemanticVersion PARQUET_251_FIXED_VERSION = new SemanticVersion(1, 8, 0);
+  private static final SemanticVersion PARQUET_251_FIXED_START = new SemanticVersion(1, 5, 0, null, "cdh5.5.0", null);
```
isnotinvain: maybe rename these to `CDH_5_PARQUET_251_FIXED_START` and `CDH_5_PARQUET_251_FIXED_END`

parquet-common/src/main/java/org/apache/parquet/SemanticVersion.java
```diff
@@ -141,7 +186,91 @@ public int hashCode() {
...
+    public int compareTo(NumberOrString that) {
+      // Numeric identifiers always have lower precedence than non-numeric identifiers.
+      int cmp = compareBooleans(that.isNumeric, this.isNumeric);
+      if (cmp != 0) {
+        return cmp;
+      }
+
+      if (isNumeric) {
+        // identifiers consisting of only digits are compared numerically
+        return this.number - that.number;
```
isnotinvain: can you use Integer.compare() here, or just inline it as written here:
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/lang/Integer.java#L1021

the x - y trick has issues w/ overflow

parquet-common/src/main/java/org/apache/parquet/SemanticVersion.java
```diff
@@ -141,7 +186,91 @@ public int hashCode() {
...
+      // A larger set of pre-release fields has a higher precedence than a
+      // smaller set, if all of the preceding identifiers are equal
+      int size = Math.min(this.identifiers.size(), that.identifiers.size());
+      for (int i = 0; i < size; i += 1) {
+        int cmp = identifiers.get(i).compareTo(that.identifiers.get(i));
+        if (cmp != 0) {
+          return cmp;
+        }
+      }
+      return this.identifiers.size() - that.identifiers.size();
```
isnotinvain: same here for compare()

isnotinvain:
overall looks good, some minor comments

isnotinvain:
+1

rdblue:
Thanks for the review, Alex!

PR #262, opened 2015-08-20
PARQUET-362 - Fix parquet buffered writer being oversensitive to union schema changes

laurentgo:
Parquet does prevent records with unknown union fields to be written as it would
create a TProtocol violation. But it also prevents records with unions having one their field
itself having an unknown field (which is acceptable if it is a struct).


isnotinvain:
+1, Thanks for fixing, LGTM

Lets try this out internally and make sure it fixes the issue we are seeing.

laurentgo:
Tested on a dataset which had 90% of records flagged as invalid, and now down to 0 with the patch

tsdeng:
＋1

PR #264, opened 2015-08-23
PARQUET-364: Fixes support for Avro/Thrift list of primitive lists

liancheng:
Please refer to [PARQUET-364] [1] for more details.

[1]: https://issues.apache.org/jira/browse/PARQUET-364

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -255,7 +255,7 @@ class AvroIndexedRecordConverter<T extends IndexedRecord> extends GroupConverter
           SpecificData.get().getClass(avroSchema);
       if (fixedClass != null) {
         try {
-          this.fixedClassCtor = 
+          this.fixedClassCtor =
               fixedClass.getConstructor(new Class[] { byte[].class });
         } catch (Exception e) {
           throw new RuntimeException(e);
@@ -313,7 +313,7 @@ class AvroIndexedRecordConverter<T extends IndexedRecord> extends GroupConverter
       Type repeatedType = type.getType(0);
       // always determine whether the repeated type is the element type by
       // matching it against the element schema.
-      if (isElementType(repeatedType, elementSchema)) {
+      if (isElementType(repeatedType, elementSchema, type.getName())) {
         // the element type is the repeated type (and required)
         converter = newConverter(elementSchema, repeatedType, model, new ParentValueContainer() {
           @Override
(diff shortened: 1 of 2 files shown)


parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
```diff
@@ -653,8 +654,8 @@ public void testAvroCompatRequiredGroupInListWithSchema() throws Exception {
     Path test = writeDirect(
         "message AvroCompatRequiredGroupInListWithSchema {" +
             "  optional group locations (LIST) {" +
-            "    repeated group array {" +
-            "      optional group element {" +
+            "    repeated group bag {" +
+            "      optional group array_element {" +
```
liancheng: According the following `LIST` backwards-compatibility rule

> If the repeated field is a group with one field and is named either `array` or uses the `LIST`-annotated group's name with `_tuple` appended then the repeated type is the element type and elements are required.

the repeated `array` group in the original schema should be interpreted as element type since it has only a single field.  However, the testing data below actually treat it as the repeated wrapper group, and thus fails after `AvroIndexedRecordConverter.isElementType()` is fixed.


According to other code within this test case, this schema was designed to test parquet-hive compatibility.  By running the following SQL statement in Hive 0.13.1

```sql
CREATE TABLE parquet_test
STORED AS PARQUET
AS SELECT ARRAY(
  NAMED_STRUCT("latitude", 1.0, "longitude", 2.0),
  NAMED_STRUCT("latitude", 1.0, "longitude", 2.0)
) AS locations;
```

we may get a similar schema (printed with `parquet-schema`):

```
message hive_schema {
  optional group locations (LIST) {
    repeated group bag {
      optional group array_element {
        optional double latitude;
        optional double longitude;
      }
    }
  }
}
```

That's why I updated this test case.

liancheng: @rdblue Just noticed the `testHiveCompatOptionalGroupInList` test case, which is very similar to this one.  I'm a little bit confused by the purpose of this test case now, especially the `RequiredGroupInListWithSchema` part of this test case name: there's no required group appeared in the testing schema.

julienledem: @rdblue @tomwhite comments?

rdblue: I need to take a thorough look at this, sorry it's taking so long!

rdblue: @liancheng, this is testing something slightly different. The backward compatibility rules are strictly implemented in AvroSchemaConverter and are applied when trying to decide what the data within a Parquet schema represents. On the other hand, when an Avro schema is present in the file we know exactly what was represented.

So this is testing that we can override the backward-compatibility rules using an Avro schema when the converters are initialized. In this case, although the schema appears to be a 2-level list, we're requesting that it be interpreted as a 3-level list. I think that this test is correct.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -356,12 +356,16 @@ public void end() {
...
      */
-    static boolean isElementType(Type repeatedType, Schema elementSchema) {
+    static boolean isElementType(Type repeatedType, Schema elementSchema, String parentName) {
       if (repeatedType.isPrimitive() ||
           repeatedType.asGroupType().getFieldCount() > 1) {
         // The repeated type must be the element type because it is an invalid
         // synthetic wrapper (must be a group with one field).
         return true;
+      } else if (repeatedType.asGroupType().getFieldCount() == 1 &&
+          (repeatedType.getName().equals("array") ||
```
rdblue: Rather than checking for "array" here, I think the right thing to do is to check whether the inner type is repeated. That's another case where the structure *must* be a 2-level list because only optional and repeated are allowed there in the 3-level structure.

That check avoids using names, which can be overridden by the requested Avro schema. It will work for both lists of lists and lists of maps.

liancheng: Did you mean "only optional and *required* are allowed there in the 3-level structure"? This sounds good to me and is more general.

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
liancheng: Thanks for the explanation, this makes more sense to me now. But why is there the "RequiredGroup" part in the test case name? I don't see any required groups in either the Parquet schema (written in string) or the Avro schema.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
rdblue: Yes, that's exactly what I meant.

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
rdblue: That's probably a cut and paste issue. I'll fix it in the other PR.

liancheng:
I'm closing this since #272 supersedes this one.

PR #265, opened 2015-08-26
Update Summary.java

waterlx:
PARQUET-365


parquet-pig/src/main/java/org/apache/parquet/pig/summary/Summary.java
```diff
@@ -266,6 +266,11 @@ public void setInputSchema(Schema input) {
       throw new RuntimeException("Usage: B = FOREACH (GROUP A ALL) GENERATE Summary(A); Can not get schema from "+input, e);
     }
   }
+  
+  @Override
+  public Schema getInputSchema() {
+    return this.inputSchema;
+  }
```
julienledem: The schema returned is not consistent with the schema set.
How about adding a call to super.setInputSchema(input) in setInputSchema? 
```
   @Override
   public void setInputSchema(Schema input) {		   
     super.setInputSchema(input);
     try {
```

julienledem:
Is this fixed by https://github.com/apache/parquet-mr/pull/292 ?

PR #268, opened 2015-09-03
PARQUET-368: Add dictionary support to parquet-thrift

isnotinvain:


parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
```diff
@@ -393,20 +533,42 @@ public long readI64() throws TException {
...
-      this.events = events;
+    @Override
+    protected ParquetProtocol decodeFromDictionary(Dictionary d, int i) {
+      final Binary decoded = d.decodeToBinary(i);
+
+      return new ParquetProtocol("readString() binary") {
+        // since we don't know whether this field is a binary or a string,
+        // we lazily convert the first time readString() or readBinary is called.
+        // we can't eagerly convert because a binary blob might not be a valid string
+        // TODO: ensure that this code doesn't need to be threadsafe
```
rdblue: I think this TODO is something that we should track down before committing this since it is about thread-safety.

Luckily, I don't think it matters. Assuming that two threads execute this at the same time, both will convert the string to UTF8 and then set the asString variable (or similar for binary). It doesn't really matter if there is a race condition because one of the converted values will win in the end, right?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
```diff
@@ -445,6 +600,26 @@ public FieldEnumConverter(List<TProtocol> events, ThriftField field) {
       }
     }
 
+     @Override
+     protected ParquetProtocol decodeFromDictionary(Dictionary d, int i) {
+       Binary value = d.decodeToBinary(i);
+       final Integer id = enumLookup.get(value);
```
rdblue: I think some of these could be refactored to reduce duplication. What about a `event` method that creates the ParquetProtocol object? That would reduce the size of this code quite a bit.

parquet-thrift/src/test/java/org/apache/parquet/thrift/TestDictionaryEncoding.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Test
+  public void testDictionaryEncoding() throws Exception {
+    Path p = new Path(new File(tempDir.getRoot(), "out").getAbsolutePath());
+    ThriftParquetWriter<OneOfEach> writer = new ThriftParquetWriter<OneOfEach>(p,
+        OneOfEach.class,
+        CompressionCodecName.UNCOMPRESSED,
+        ParquetWriter.DEFAULT_BLOCK_SIZE,
+        ParquetWriter.DEFAULT_PAGE_SIZE,
+        true, // Dictionary encoding enabled = true
```
rdblue: Should we verify that dictionary encoding was actually used by checking the file's metadata?

parquet-thrift/src/test/java/org/apache/parquet/thrift/TestDictionaryEncoding.java
```diff
@@ -0,0 +1,95 @@
...
+      OneOfEach ooe = new OneOfEach();
+      ooe.setIm_true(i % 2 == 0);
+      ooe.setIm_false(i % 2 != 0);
+      ooe.setA_bite((byte) (i % 10));
+      ooe.setInteger16((short) (i % 100));
+      ooe.setInteger32(i % 28);
+      ooe.setInteger64(i);
+      ooe.setDouble_precision(((double) i) / 2.0);
+      ooe.setSome_characters("a string " + i % 10);
+      ooe.setZomg_unicode("no thanks"); // test dictionary size = 1
```
rdblue: What about testing with [actual unicode](http://unicodesnowmanforyou.com/)?

parquet-thrift/src/test/java/org/apache/parquet/thrift/TestDictionaryEncoding.java
```diff
@@ -0,0 +1,95 @@
...
+      ooe.setWhat_who(false);
+      ooe.setBase64(("zomg bytes " + (i % 10)).getBytes("UTF-8"));
```
rdblue: Why is this called base64 but is actually binary? Isn't that misleading?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
isnotinvain: Sorry, I don't understand. Which part could be de-duplicated?

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
isnotinvain: I think you're right.
Parquet in hadoop doesn't need to be thread safe, but idk about when its used in spark and similar things.

parquet-thrift/src/test/java/org/apache/parquet/thrift/TestDictionaryEncoding.java
isnotinvain: sure, this schema is sort of arbitrary, the field names don't need to mean anything. But I can add some unicode in there.

parquet-thrift/src/test/java/org/apache/parquet/thrift/TestDictionaryEncoding.java
isnotinvain: yeah same as above, this OneOfEach struct is just a thrift struct we have for testing, I think it comes from ElephantBird. Could use any other struct instead.

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
rdblue: This method and addBinary below. Actually, lots of methods are duplicating the inline class definition, which ends up creating two identical .class files that get optimized separately. Code reuse there would probably be a good thing.

isnotinvain: Oh, you mean duplicating the anonymous classes? We could make a named class for each read method if that's what you mean? Like ReadI32ParquetProtocol?

rdblue: There's quite a bit of code duplication as well, which could be cleaned up. One of the problems with the code duplication is that it creates multiple anonymous classes when one would be sufficient. It isn't necessary to add them as named classes because cutting down on the code duplication should fix it.

isnotinvain: I agree that these all follow the same pattern, and it looks like code duplication. However, they all extend a base class, and the only thing that IntConverter / LongConverter / etc. override is the part that has to be different for each.

IntConverter needs to call `dictionary.decodeToInt`, LongConverter needs to call `dictionary.decodeToInt`. IntConverter needs to return a protocol with an overidden `readI32` method, LongConverter needs to override `readI64`, etc.

The only thing I can see that could be de-duplicated is in:

```
static final class IntConverter extends PrimitiveConverterBase {

    IntConverter(List<TProtocol> events) {
      super(events);
    }

    @Override
    protected ParquetProtocol decodeFromDictionary(Dictionary d, int i) {
      final int value = d.decodeToInt(i);
      return new ParquetProtocol("readI32()") {
        @Override
        public int readI32() throws TException {
          return value;
        }
      };
    }

    @Override
    public void addInt(final int value) {
      events.add(new ParquetProtocol("readI32()") {
        @Override
        public int readI32() throws TException {
          return value;
        }
      });
    }
  }
```

we could instead do

```
static final class IntConverter extends PrimitiveConverterBase {

    IntConverter(List<TProtocol> events) {
      super(events);
    }
    
    private static ParquetProtocol event(final int value) {
      return new ParquetProtocol("readI32()") {
        @Override
        public int readI32() throws TException {
          return value;
        }
      };
    }    
    @Override
    protected ParquetProtocol decodeFromDictionary(Dictionary d, int i) {
      final int value = d.decodeToInt(i);
      return event(value);
    }

    @Override
    public void addInt(final int value) {
      events.add(event(value));
    }
  }
```

Is that what you mean?

Thanks,
Alex

rdblue: Exactly. That cuts down on the number of anonymous classes.

PR #270, opened 2015-09-12
PARQUET-374: Add api to read dictionary from each column chunk for predicate pushdown

zhenxiao:


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.parquet.column.Encoding;
+import org.apache.parquet.column.page.DictionaryPage;
+import org.apache.parquet.format.DictionaryPageHeader;
+import org.apache.parquet.format.PageHeader;
+import org.apache.parquet.format.PageType;
+import org.apache.parquet.format.Util;
+
+public class DictionaryStream
+    extends ByteArrayInputStream
+{
```
julienledem: please make sure indentation is consistent with existing code

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,75 @@
...
+    public DictionaryStream(byte[] data, int offset)
+    {
+        super(data);
+        this.pos = offset;
+    }
+
+    protected PageHeader readPageHeader()
+        throws IOException
+    {
+        return Util.readPageHeader(this);
```
julienledem: Please avoid circular dependencies.
this method is either in Util or here but not both.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.apache.parquet.bytes.BytesInput;
+import org.apache.parquet.column.Encoding;
+import org.apache.parquet.column.page.DictionaryPage;
+import org.apache.parquet.format.DictionaryPageHeader;
+import org.apache.parquet.format.PageHeader;
+import org.apache.parquet.format.PageType;
+import org.apache.parquet.format.Util;
+
+public class DictionaryStream
```
julienledem: please add javadoc. It seems that the methods in this class need to be called in a given order for this to work. Please document behavior.
It would be great if this contract was enforced by the class' API

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,75 @@
...
+                                        dicHeader.getNum_values(),
+                                        Encoding.valueOf(dicHeader.getEncoding().name()));
+        }
+        return null;
+    }
+
+    public int getPosition()
+    {
+        return this.pos;
+    }
```
julienledem: should this be exposed?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -434,6 +440,63 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
...
+      }
+    }
+    return dictionaries;
+  }
+
+  private static boolean isOnlyDictionaryEncodingPages(Set<Encoding> encodings) {
+    // more than 1 encodings for values
+    if (encodings.size() > PARQUET_DATA_TRIPLE) {
+      return false;
+    }
```
julienledem: please verify your heuristics. There could be some other cases.
PAGE_V2 formalizes this contract.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -434,6 +440,63 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
...
+          break;
+        }
+      }
+    }
+    return dictionaries;
+  }
+
+  private static boolean isOnlyDictionaryEncodingPages(Set<Encoding> encodings) {
+    // more than 1 encodings for values
+    if (encodings.size() > PARQUET_DATA_TRIPLE) {
```
julienledem: Using a constant here does not improve readability in my opinion.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -434,6 +440,63 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
...
+                                                            compressedPage.getDictionarySize(),
+                                                            compressedPage.getEncoding());
+          Dictionary dictionary = dictionaryPage.getEncoding().initDictionary(columnDescriptor, dictionaryPage);
+          dictionaries.put(columnDescriptor, dictionary);
+          break;
+        }
+      }
+    }
+    return dictionaries;
+  }
```
julienledem: Let's not make this a static method. We are duplicating things like the CODEC initialization in here.
I would expect to have something similar to the existing APIs (readNextRowGroup() but just for dictionaries):
```
DictionaryPageReadStore dicts = new ParquetFileReader(conf, fileMetadata, path, blocks, columns).readNextRowGroupDictionaries();
```
The decoding should not happen in this class.
Possibly you can add a method in MessageColumnIO to return a dictionary reader since it is the class responsible to create RecordReaders (see MessageColumnIO.getRecordReader and ColumnReadStoreImpl)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDictionaryReader.java
```diff
@@ -0,0 +1,193 @@
...
+import static java.lang.Boolean.TRUE;
+import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_1_0;
+import static org.apache.parquet.hadoop.api.ReadSupport.PARQUET_READ_SCHEMA;
+import static org.apache.parquet.hadoop.metadata.CompressionCodecName.GZIP;
+import static org.apache.parquet.schema.MessageTypeParser.parseMessageType;
+import static org.junit.Assert.*;
+import org.junit.BeforeClass;
+
+public class TestDictionaryReader
+{
```
julienledem: indentation.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDictionaryReader.java
```diff
@@ -0,0 +1,193 @@
...
+  @Test
+  public void testFloatRead() throws Exception {
+    MessageType requestedSchema = parseMessageType("message test { required float float_field;}");
+    ParquetMetadata metadata = ParquetFileReader.readFooter(conf, file);
+    for (BlockMetaData block : metadata.getBlocks()) {
+      Map<ColumnDescriptor, Dictionary> result = ParquetFileReader.readDictionary(block, file, conf, requestedSchema);
+      for (Map.Entry<ColumnDescriptor, Dictionary> entry : result.entrySet()) {
+        Dictionary dictionary = entry.getValue();
+        for (int i = 0; i < ENGLISH_CHARACTER_NUMBER; i++) {
+          assertEquals((float) intValues[i % ENGLISH_CHARACTER_NUMBER] * 2.0, dictionary.decodeToFloat(i), 1e-15);
```
julienledem: ```i % invalues.length``` ?

julienledem:
Thanks for the contribution.
Comments are mostly about reorganizing code in the places that make more sense.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -517,7 +523,67 @@ public PageReadStore readNextRowGroup() throws IOException {
...
+    }
+    return dictionaryPageReadStore;
+  }
+
+  // are all the value pages dictionary encoded
+  private boolean isOnlyDictionaryEncodingPages(Set<Encoding> encodings) {
+    // if values have more than one encodings, definitely not dictionary encoding only
+    if (encodings.size() > 3) {
+      return false;
+    }
```
zhenxiao: @julienledem I tested against both PAGE_V1 and PAGE_V2, this heuristics seems working OK, an ideal solution is to use PageEncodingStats, this is a heuristics workaround. I put todo here as a next step

zhenxiao:
thank you, @julienledem get comments addressed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -517,7 +523,67 @@ public PageReadStore readNextRowGroup() throws IOException {
...
+    for (ColumnChunkMetaData columnChunkMetaData : block.getColumns()) {
+      for (ColumnDescriptor columnDescriptor : requestedSchema.getColumns()) {
+        if (columnChunkMetaData.getPath().equals(ColumnPath.get(columnDescriptor.getPath())) &&
+            isOnlyDictionaryEncodingPages(columnChunkMetaData.getEncodings())) {
+          long startingPosition = columnChunkMetaData.getStartingPos();
+          this.f.seek(startingPosition);
+
+          int totalSize = (int) columnChunkMetaData.getTotalSize();
+          byte[] buffer = new byte[totalSize];
+          this.f.readFully(buffer);
```
julienledem: Possibly you could check if you can read only the dictionary page instead of the entire column chunk.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class DictionaryStream extends ByteArrayInputStream {
+
+  public DictionaryStream(byte[] data, int offset) {
+    super(data);
+    this.pos = offset;
+  }
+
+  public DictionaryPage readDictionaryPage() throws IOException {
+    PageHeader pageHeader = Util.readPageHeader(this);
```
julienledem: readPageHeader should just be a member of this class.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.parquet.column.page.DictionaryPage;
+import org.apache.parquet.format.DictionaryPageHeader;
+import org.apache.parquet.format.PageHeader;
+import org.apache.parquet.format.PageType;
+import org.apache.parquet.format.Util;
+
+/**
+ * Data stream to only read dictionary page
+ */
+public class DictionaryStream extends ByteArrayInputStream {
```
julienledem: this class reads a Dictionary page from a stream, it does not need to extend a stream.
Users of this class will likely not use the Stream methods.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryStream.java
```diff
@@ -0,0 +1,61 @@
...
+    int compressedPageSize = pageHeader.getCompressed_page_size();
+    if (pageHeader.type == PageType.DICTIONARY_PAGE) {
+      DictionaryPageHeader dicHeader = pageHeader.getDictionary_page_header();
+      return new DictionaryPage(getBytesInput(compressedPageSize),
+                                uncompressedPageSize,
+                                dicHeader.getNum_values(),
+                                Encoding.valueOf(dicHeader.getEncoding().name()));
+    }
+    return null;
+  }
```
julienledem: let's not duplicate that code.
You can refactor some of the code here:
https://github.com/apache/parquet-mr/blob/c3819688c48480ec75a9563c71f18ea755e34620/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java#L129
So that this logic is reused.
Also make the naming consistent

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkDictionaryPageReadStore.java
```diff
@@ -0,0 +1,89 @@
...
+            compressedDictionaryPage.getEncoding());
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  private final Map<ColumnDescriptor, ColumnChunkDictionaryPageReader> readers = new HashMap<ColumnDescriptor, ColumnChunkDictionaryPageReader>();
+
+  public ColumnChunkDictionaryPageReadStore() {
```
julienledem: this should be package visibility

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkDictionaryPageReadStore.java
```diff
@@ -0,0 +1,89 @@
...
+      throw new IllegalArgumentException(column + " is not in the store: " + readers.keySet());
+    }
+    return readers.get(column);
+  }
+
+  void addColumn(ColumnDescriptor column, ColumnChunkDictionaryPageReader reader) {
+    if (readers.put(column, reader) != null) {
+      throw new RuntimeException(column + " was added twice");
+    }
+  }
```
julienledem: this could be done in the constructor so that we can make this class immutable

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDictionaryReader.java
```diff
@@ -0,0 +1,206 @@
...
+  @Test
+  public void testIntRead() throws Exception {
+    MessageType requestedSchema = parseMessageType("message test { required int32 int32_field;}");
+    ParquetMetadata metadata = ParquetFileReader.readFooter(conf, file);
+    ParquetFileReader fileReader = new ParquetFileReader(conf, metadata.getFileMetaData(), file, metadata.getBlocks(), requestedSchema.getColumns());
+    DictionaryPageReadStore dictionaryReadStore = fileReader.readNextRowGroupDictionaries(requestedSchema);
+    ColumnDescriptor column = requestedSchema.getColumns().get(0);
+    DictionaryPageReader dictionaryPageReader = dictionaryReadStore.getDictionaryPageReader(column);
+    DictionaryPage dictionaryPage = dictionaryPageReader.readDictionaryPage();
+    Dictionary dictionary = dictionaryPage.getEncoding().initDictionary(column, dictionaryPage);
```
julienledem: You should wrap this logic in a class that returns Dictionary objects.
Typically user code does not want to interact with Pages but the final decoded data.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkDictionaryPageReadStore.java
zhenxiao: @julienledem I might not get a good understanding, need to add columnDescriptor and DictionaryPageReader into this class one by one, do you mean I pass a map in the constructor?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
zhenxiao: @julienledem seems like could not decide the compressed page size until reading the page header, so at this time(before reading page header), do not know the actual size to read. I will leave todo here

zhenxiao:
thank you @julienledem get comments addressed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -517,7 +524,74 @@ public PageReadStore readNextRowGroup() throws IOException {
...
+  /**
+   * Reads the requested predicate columns dictionaries from the row group at the current file position.
+   * Not reading all the requested columns, only read predicate columns.
+   * eg. select columnA, columnB, columnC from table where columnC > 10
+   *     only reads columnC's dictionary
+   * @param requestedSchema the predicate columns to read
+   * @throws IOException if an error occurs while reading
+   * @return the DictionaryPageReadStore which can provide DictionaryPageReaders for each column.
+   */
+  public DictionaryPageReadStore readNextRowGroupDictionaries(MessageType requestedSchema) throws IOException {
```
danielcweeks: @zhenxiao Isn't this name a little misleading?  I believe this is better as `getCurrentRowGroupDictionaries(...)` because it just returns the current row group and doesn't advance.

zhenxiao: @danielcweeks my bad, forgot to advance the row group, just get it fixed. Following the readNextRowGroup, call this readNextRowGroupDictionaries

zhenxiao: @danielcweeks you are correct. read dictionaries should not advance the row group, only after reading the whole row group does. I get the method name updated

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -517,7 +524,74 @@ public PageReadStore readNextRowGroup() throws IOException {
...
+  /**
+   * Reads the requested predicate columns dictionaries from the row group at the current file position.
+   * Not reading all the requested columns, only read predicate columns.
+   * eg. select columnA, columnB, columnC from table where columnC > 10
+   *     only reads columnC's dictionary
+   * @param requestedSchema the predicate columns to read
+   * @throws IOException if an error occurs while reading
+   * @return the DictionaryPageReadStore which can provide DictionaryPageReaders for each column.
+   */
+  public DictionaryPageReadStore getCurrentRowGroupDictionaries(MessageType requestedSchema) throws IOException {
```
danielcweeks: @julienledem Unfortunately, I don't think this API works very well for the uses cases we have right now.  The issue @zhenxiao  is seeing is that he wants to identify the row groups not filtered prior to processing them.  This interface requires that you do both in lock step.  The filter api has the same problem in that it basically needs to do up front row group filtering without loading any column data.  We would have to be able to advance the dictionary separately from the row group, which would make this interface rather clunky. 

Overall, I think the static interface was closer to what we currently need (though I can see how this approach can be used to filter while processing).  Maybe we just need to separate out the reading of the dictionary as a utility that gets used by this.

zhenxiao: @julienledem @danielcweeks as @danielcweeks explained, the use case is,
read footer, for each row group read blockMetaData and Dictionary for each predicate column chunks, decide whether to read the row group or not based on stats and dictionary, keep the list of to-be-scanned row groups, and then start scanning the necessary row groups.
In this case advance dictionary is separate from advancing row group data.
I am thinking about the following options:
1. advance blockId in both nextRowGroup() and nextRowGroupDictionaries(), while, they should not be called interleaving or together, only one of them could be called in the application
2. have a user level advanceBlock() which will advance blockId, then user could decide when to call nextRowGroup(), when to advanceBlock(), and when to read dictionaries, this could be a little bit too flexible, and existing applications calling nextRowGroup() needs to be updated
3. have a static method in ParquetFileReader, pass BlockMetaData, configuration, path as input parameters
Agree with @danielcweeks option#3 is better. what do you think?

PR #271, opened 2015-09-14
PARQUET-376: Tolerate square brackets in PR titles

isnotinvain:
This allows for PRs like:

`[PARQUET-XXXX] description`

to be parsed, as we often get this format and we usually have to ask the submitter to change the title for us.

### DIFF
--- a/dev/merge_parquet_pr.py
+++ b/dev/merge_parquet_pr.py
@@ -141,7 +141,9 @@ def merge_pr(pr_num, target_ref):
 
     merge_message_flags = []
 
-    merge_message_flags += ["-m", title]
+    title_without_brackets = remove_brackets_from_pr_title(title)
+
+    merge_message_flags += ["-m", title_without_brackets]
     if body != None:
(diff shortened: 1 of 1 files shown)


rdblue:
Can we also remove the brackets when building the final commit message?

isnotinvain:
@rdblue You mean remove them from the PR title when we put the title in the commit message?

isnotinvain:
anywhere the jira's ID is used it should have the [] already removed, but probably not when the PR title gets pushed into the commit message.

tsdeng:
+1

rdblue:
Yeah, I think we shouldn't allow this to leak into the commit message. Shouldn't be too difficult to mimic the same logic when building the commit message, right?

rdblue:
Nevermind, I see you already did. +1

isnotinvain:
looks like i merged this prematurely, it didn't quite work as expected, I've reverted it for now.

PR #272, opened 2015-09-22
PARQUET-364: Fix compatibility for Avro lists of lists.

rdblue:
This fixes lists of lists that have been written with Avro's 2-level
representation. The conversion setup logic missed the case where the
inner field is repeated and cannot be the element in a 3-level list.

This also fixes the schema conversion for cases where an unknown
writer used a 2-level list of lists.

This is based on @liancheng's #264 but fixes the problem in a slightly different way.

### DIFF
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
@@ -567,9 +567,9 @@ public class TestArrayCompatibility {
   }
 
   @Test
-  public void testAvroCompatRequiredGroupInList() throws Exception {
+  public void testAvroCompatOptionalGroupInList() throws Exception {
     Path test = writeDirect(
-        "message AvroCompatRequiredGroupInList {" +
+        "message AvroCompatOptionalGroupInList {" +
             "  optional group locations (LIST) {" +
             "    repeated group array {" +
             "      optional group element {" +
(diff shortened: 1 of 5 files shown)


rdblue:
The test failures are due to PARQUET-373, a flaky memory manager test.

liancheng:
This looks good to me except for the question about the test case name which I mentioned in [this comment] [1].

[1]: https://github.com/apache/parquet-mr/pull/264#discussion_r40155153

rdblue:
I fixed the test case names for Avro compat. Thanks for looking at this @liancheng!

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
```diff
@@ -714,7 +714,7 @@ public void write(RecordConsumer rc) {
         field("latitude", primitive(Schema.Type.DOUBLE)),
         field("longitude", primitive(Schema.Type.DOUBLE)));
 
-    Schema newSchema = record("HiveCompatOptionalGroupInList",
+    Schema newSchema = record("AvroCompatRequiredGroupInListWithSchema",
```
liancheng: This is still saying "Required" :)

liancheng:
+1

Thanks for fixing this (in a better way)!

rdblue:
@tomwhite could you take a quick look at this as well? Thanks!

rdblue:
@tomwhite, the test failures have been fixed and tests are passing now. Could you take a look at this when you have time?

julienledem:
This looks good to me. 
@rdblue should we merge?

rdblue:
Will do. Thanks for taking a look!

PR #273, opened 2015-09-23
PARQUET-378: Add thoroughly parquet test encodings

spena:
A new test case TestTypeEncodings is added that test v1 and v2 encodings for all supported column types. This test case spans many pages and row groups, and reads each page individually from first-to-last and from last-to-first.


spena:
Deleted pull request because it was created incorrectly.

PR #269, opened 2015-09-11
PARQUET-373: Fix flaky MemoryManager tests.

rdblue:

PR #274, opened 2015-09-23
PARQUET-378: Add thoroughly parquet test encodings

spena:
A new test case TestTypeEncodings is added that test v1 and v2 encodings for all
supported column types. This test case spans many pages and row groups, and reads
each page individually from first-to-last and from last-to-first.

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/encodings/FileEncodingsIT.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/encodings/FileEncodingsIT.java
@@ -1,0 +1,490 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.encodings;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.ParquetProperties.WriterVersion;
+import org.apache.parquet.column.impl.ColumnReaderImpl;
+import org.apache.parquet.column.page.DataPage;
+import org.apache.parquet.column.page.DictionaryPage;
+import org.apache.parquet.column.page.PageReadStore;
+import org.apache.parquet.column.page.PageReader;
+import org.apache.parquet.example.data.Group;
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/encodings/TestTypeEncodings.java
```diff
@@ -0,0 +1,417 @@
...
+    int96Generator = new RandomValues.Int96Generator(random.nextLong());
+    floatGenerator = new RandomValues.FloatGenerator(random.nextLong());
+    doubleGenerator = new RandomValues.DoubleGenerator(random.nextLong());
+    binaryGenerator = new RandomValues.BinaryGenerator(random.nextLong());
+    fixedBinaryGenerator = new RandomValues.FixedGenerator(random.nextLong(), FIXED_LENGTH);
+  }
+
+  @Test
+  public void testTypeEncodings() throws Exception {
+    int pageSize = 64; // 64B
```
rdblue: I think we should use a bigger value for this. I bumped it up to 16k and the tests work just fine. 8k or 16k is a setting more like what we would see in real data, and you can still fit several pages in each row group.

parquet-hadoop/src/test/java/org/apache/parquet/encodings/TestTypeEncodings.java
```diff
@@ -0,0 +1,417 @@
...
+      schema = Types.buildMessage().required(type).length(FIXED_LENGTH).named("field").named("test");
+    } else {
+      schema = Types.buildMessage().required(type).named("field").named("test");
+    }
+
+    SimpleGroupFactory message = new SimpleGroupFactory(schema);
+    GroupWriteSupport.setSchema(schema, configuration);
+
+    ParquetWriter<Group> writer = new ParquetWriter<Group>(file, new GroupWriteSupport(),
+        CompressionCodecName.UNCOMPRESSED, rowGroupSize, pageSize, 0, false, false, version, configuration);
```
rdblue: Could you add a comment that calls out the fact that we are avoiding dictionary encoding?

rdblue:
I tested this locally after removing the fix for PARQUET-246 and it caught the error. I'm +1 once the minor comments are addressed. Thanks @spena!

spena:
@rdblue I committed another patch addressing your feedback. I had to force the push because of problems I had.

In this patch, I added the following:
- testTypesEncodingsWithDictionary
- Increment page size from 64B to 16K
- Reduced row group size from 256K to 128K so that booleans can span more pages
- Increment record count to 2 million to span more pages
- Test every value of WriterVersion.values() in case a new one is added in the future

I run the test without the PARQUET-246 fix, and it catch the bug as well.

rdblue:
Thanks @spena! The tests failed, but both were due to #269. I'll commit this soon.

spena:
@rdblue How can I re-run the tests to see if this patch passes?

rdblue:
Rebase on the current master and force-push. Thanks @spena! I'll review it after.

spena:
@rdblue These tests added ~10 min more time to unit-testing. Would you like to lower the time spent on unit-testing, or is that ok?

rdblue:
Ouch. Any way we can get that down a bit? Adding 10 minutes to the build is pretty ugly. In that case I'd say we should add these as integration tests using the failsafe plugin.

spena:
@rdblue Now the tests run with 'mvn verify' as integration tests. Although, I checked travis-ci, and all tests run in ~17min, Last time with this encodings test were 21min. Maybe there are other tests adding more time, but this one is good for now I think.

julienledem:
This looks good to me.
@rdblue did you want to merge?

rdblue:
Merged. Thanks, Sergio!

PR #275, opened 2015-09-23
PARQUET-372: Do not write stats larger than 4k.

rdblue:
This updates the stats conversion to check whether the min and max
values for page stats are larger than 4k. If so, no statistics for a
page are written.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
@@ -191,6 +191,14 @@ public abstract class Statistics<T extends Comparable<T>> {
   abstract public byte[] getMinBytes();
 
   /**
+   * Abstract method to return whether the min and max values fit in the given
+   * size.
+   * @param size a size in bytes
+   * @return true iff the min and max values are less than size bytes
+   */
+  abstract public boolean isSmallerThan(long size);
+
+  /**
    * toString() to display min, max, num_nulls in a string
    */
   abstract public String toString();
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -285,6 +285,10 @@ public class ParquetMetadataConverter {
   public static Statistics toParquetStatistics(
       org.apache.parquet.column.statistics.Statistics statistics) {
     Statistics stats = new Statistics();
+    // Don't write stats larger than the max size rather than truncating. The
+    // rationale is that some engines may use the minimum value in the page as
+    // the true minimum for aggregations and there is no way to mark that a
+    // value has been truncated and is a lower bound and not in the page.
     if (!statistics.isEmpty() && statistics.isSmallerThan(MAX_STATS_SIZE)) {
       stats.setNull_count(statistics.getNumNulls());
       if (statistics.hasNonNullValue()) {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -77,6 +77,7 @@ public class ParquetMetadataConverter {
 
   public static final MetadataFilter NO_FILTER = new NoFilter();
   public static final MetadataFilter SKIP_ROW_GROUPS = new SkipMetadataFilter();
+  public static final long MAX_STATS_SIZE = 4096; // limit stats to 4k
 
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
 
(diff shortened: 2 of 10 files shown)


rdblue:
@spena, @julienledem, @isnotinvain what do you guys think of this change? We're hitting cases where the min/max stats are about the size of a page, so we end up with pages 3x larger than they should be.

I think the best solution is to not write stats over some limit, which I've set at 4k here. Scanning would be quick when this is the case because if values are that large there wouldn't be many of them in a page. Also, since the large values must be min and/or max, we don't expect many cases where just one huge value prevents stats from being written.

I had suggested on the ticket that we might want to modify the value, but that gets messy and wouldn't allow us to satisfy queries from stats alone (select max(col) from table).

rdblue:
Test failures are again due to flaky memory manager tests, fixed in #263.

rdblue:
I looked at stats merging and I think this is safe because statistics are aggregated before they are written to the file. A column chunk will accumulate the min and max values for all pages accurately, but when individual pages are written, the stats may be discarded instead of written to the file.

spena:
@rdblue The patch looks good.

I just have a question about the limit for the statistics. Why not limiting the size to the current page size used on the row group (or pagesize / 4 or something)? If a user ends setting a big page size, then they won't have the advantage of using the statistics to check whether the row group should be read or not.

rdblue:
Thanks for taking a look, @spena. There are a couple of motivations for the 4k static limit. First, I'd like to keep a hard limit on the page size so they don't get unexpectedly large again. The other reason is just to keep the code simple and not need to pass the current page size through unless we know it's going to be beneficial. We can always make it more fancy later.

rdblue:
@julienledem, I think this (or something similar) should go into 1.9.0. Could you have a look at it?

rdblue:
I've rebased this and added a comment to explain why stats are ignored rather than truncated.

@julienledem could you take a look when you have time? Thanks!

proflin:
hi @rdblue 
Although I wonder if we could just stop updating the stats once we encounter large-sized max+min (to save the max/min compassion for upcoming writes), I'm +1 for this. I've looked through the code changes as well as the added unit tests -- LGTM.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -236,7 +237,11 @@ public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
   public static Statistics toParquetStatistics(
       org.apache.parquet.column.statistics.Statistics statistics) {
     Statistics stats = new Statistics();
-    if (!statistics.isEmpty()) {
+    // Don't write stats larger than the max size rather than truncating. The
+    // rationale is that some engines may use the minimum value in the page as
+    // the true minimum for aggregations and there is no way to mark that a
+    // value has been truncated and is a lower bound and not in the page.
+    if (!statistics.isEmpty() && statistics.isSmallerThan(MAX_STATS_SIZE)) {
```
danielcweeks: Is it possible to have null count without min/max.  This is a nit, but engines could use null count even without min/max present.

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -191,6 +191,14 @@ public void mergeStatistics(Statistics stats) {
   abstract public byte[] getMinBytes();
 
   /**
+   * Abstract method to return whether the min and max values fit in the given
+   * size.
+   * @param size a size in bytes
+   * @return true iff the min and max values are less than size bytes
+   */
+  abstract public boolean isSmallerThan(long size);
```
danielcweeks: size() might be a more intuitive approach and would remove the burden of evaluating this check in the subclasses.

rdblue: I just looked into this. There's nothing preventing us from doing this in the format, but the implementation assumes that there will be non-null min and max if numNulls != numRows. I think it makes sense to add this as a follow-up instead of making this patch larger.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: I just looked into this. There's nothing preventing us from doing this in the format, but the implementation assumes that there will be non-null min and max if numNulls != numRows. I think it makes sense to add this as a follow-up instead of making this patch larger.

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
rdblue: It's been a while, but I think my logic was this:
* `getMinBytes` and `getMaxBytes` allocate byte arrays and generally too much work for this check. We could cache the bytes, but that requires more work.
* `size` isn't determined by `Statistics` because the metadata converter is responsible for actually converting stats into bytes. This only assumes that the entire min and max bytes will be written.

So we delegate to a stats object to avoid work, but can't delegate to stats because it doesn't do serialization. I could change it to `getSizeHint`, but I don't think that's much better and I'm inclined to leave it as is.

danielcweeks:
+1 I agree with all of the follow up to my questions.

PR #276, opened 2015-09-23
PARQUET-380: Fix build when using thrift 0.9.0.

rdblue:
This adds the correct version of libthrift as a dependency for
parquet-cascading and parquet-scrooge, which overrides the 0.7.0
dependency that elephantbird uses.

### DIFF
--- a/parquet-cascading/pom.xml
+++ b/parquet-cascading/pom.xml
@@ -51,9 +51,15 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-        <groupId>org.apache.parquet</groupId>
-        <artifactId>parquet-thrift</artifactId>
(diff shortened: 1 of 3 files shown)


rdblue:
Looks like other people are hitting this as PARQUET-391. I've rebased this on master so we can get it in. Could you take a look @julienledem?

parquet-cascading/pom.xml
```diff
@@ -39,6 +39,10 @@
     </repository>
   </repositories>
 
+  <properties>
+    <slf4j.version>1.5.8</slf4j.version>
+  </properties>
```
julienledem: this should be in the root pom.

julienledem:
one small comment otherwise. LGTM
+1

PR #277, opened 2015-09-24
PARQUET-381: Add feature to merge metadata (summary) files, and control which files are generated

isnotinvain:
1) Add helper to merge 2 summary files, useful for merging 2 directories of data into 1
2) Add more control over whether _common_metadata, _metadata, or both is written

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -480,7 +480,8 @@ public class ParquetFileWriter {
   }
 
   /**
-   *
+   * Given a list of metadata files, merge them into a single ParquetMetadata
+   * Requires that the schemas be compatible, and the extraMetadata be exactly equal.
    */
   public static ParquetMetadata mergeMetadataFiles(List<Path> files,  Configuration conf) throws IOException {
     Preconditions.checkArgument(!files.isEmpty(), "Cannot merge an empty list of metadata");
@@ -499,6 +500,12 @@ public class ParquetFileWriter {
     return new ParquetMetadata(globalMetaData.merge(), blocks);
   }
 
+  /**
+   * Given a list of metadata files, merge them into a single metadata file.
+   * Requires that the schemas be compatible, and the extraMetaData be exactly equal.
+   * This is useful when merging 2 directories of parquet files into a single directory, as long
+   * as both directories were written with compatible schemas and equal extraMetaData.
+   */
   public static void writeMergedCommonMetadataFile(List<Path> files, Path outputPath, Configuration conf) throws IOException {
     ParquetMetadata merged = mergeMetadataFiles(files, conf);
     writeMetadataFile(outputPath, merged, outputPath.getFileSystem(conf));
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -507,7 +507,7 @@ public class ParquetFileWriter {
    * This is useful when merging 2 directories of parquet files into a single directory, as long
    * as both directories were written with compatible schemas and equal extraMetaData.
    */
-  public static void writeMergedCommonMetadataFile(List<Path> files, Path outputPath, Configuration conf) throws IOException {
+  public static void writeMergedMetadataFile(List<Path> files, Path outputPath, Configuration conf) throws IOException {
     ParquetMetadata merged = mergeMetadataFiles(files, conf);
     writeMetadataFile(outputPath, merged, outputPath.getFileSystem(conf));
   }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -522,12 +522,7 @@ public class ParquetFileWriter {
    */
   @Deprecated
   public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers) throws IOException {
-    FileSystem fs = outputPath.getFileSystem(configuration);
-    outputPath = outputPath.makeQualified(fs);
-    ParquetMetadata metadataFooter = mergeFooters(outputPath, footers);
-    writeMetadataFile(outputPath, metadataFooter, fs, PARQUET_METADATA_FILE);
-    metadataFooter.getBlocks().clear();
-    writeMetadataFile(outputPath, metadataFooter, fs, PARQUET_COMMON_METADATA_FILE);
+    writeMetadataFile(configuration, outputPath, footers, JobSummaryLevel.ALL);
   }
 
   /**
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
 import org.apache.parquet.Log;
+import org.apache.parquet.Preconditions;
 import org.apache.parquet.Version;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.BytesUtils;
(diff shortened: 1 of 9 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -474,30 +476,83 @@ private static void serializeFooter(ParquetMetadata footer, FSDataOutputStream o
...
+    return new ParquetMetadata(globalMetaData.merge(), blocks);
+  }
+
+  /**
+   * Given a list of metadata files, merge them into a single metadata file.
+   * Requires that the schemas be compatible, and the extraMetaData be exactly equal.
+   * This is useful when merging 2 directories of parquet files into a single directory, as long
+   * as both directories were written with compatible schemas and equal extraMetaData.
+   */
+  public static void writeMergedCommonMetadataFile(List<Path> files, Path outputPath, Configuration conf) throws IOException {
```
laurentgo: should the name be writeMergedMetadataFile since there's nothing specific to common metadata?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
   }
 
   public static void writeMetaDataFile(Configuration configuration, Path outputPath) {
-    if (configuration.getBoolean(ParquetOutputFormat.ENABLE_JOB_SUMMARY, true)) {
+    JobSummaryLevel level = ParquetOutputFormat.getJobSummaryLevel(configuration);
+
+    if (level != JobSummaryLevel.NONE) {
```
laurentgo: To avoid having too many nested blocks, maybe invert the condition and exit early?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
...
         // If there are no footers, _metadata file cannot be written since there is no way to determine schema!
         // Onus of writing any summary files lies with the caller in this case.
         if (footers.isEmpty()) {
           return;
         }
+
         try {
-          ParquetFileWriter.writeMetadataFile(configuration, outputPath, footers);
+          ParquetFileWriter.writeMetadataFile(configuration, outputPath, footers, level);
         } catch (Exception e) {
```
laurentgo: I didn't noticed before but the exception is never thrown back to the caller, which means the parquet committer never fails when attempting to write the metadata.

From my experience, this has caused OOM for our application but the Application Master didn't immediately exit, and was killed later. That's a behavior I liked to see changed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
...
-          LOG.warn("could not write summary file for " + outputPath, e);
+          LOG.warn("could not write summary file(s) for " + outputPath, e);
+
           final Path metadataPath = new Path(outputPath, ParquetFileWriter.PARQUET_METADATA_FILE);
           if (fileSystem.exists(metadataPath)) {
             fileSystem.delete(metadataPath, true);
```
laurentgo: if the committer were to fail, I guess cleaning after would not be necessary?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
...
           }
+
+          final Path commonMetadataPath = new Path(outputPath, ParquetFileWriter.PARQUET_COMMON_METADATA_FILE);
+          if (fileSystem.exists(commonMetadataPath)) {
+            fileSystem.delete(commonMetadataPath, true);
```
laurentgo: if previous fs operations failed, this would never be called either

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
...
+          }
         }
       } catch (Exception e) {
```
laurentgo: same here, it would be better to let the exception propagate instead of doing a catch-all

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -474,30 +476,83 @@ private static void serializeFooter(ParquetMetadata footer, FSDataOutputStream o
...
+  @Deprecated
   public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers) throws IOException {
+    writeMetadataFile(configuration, outputPath, footers, JobSummaryLevel.ALL);
+  }
+
+  /**
+   * writes _common_metadata file, and optionally a _metadata file depending on the {@link JobSummaryLevel} provided
+   */
+  public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers, JobSummaryLevel level) throws IOException {
+    Preconditions.checkArgument(level == JobSummaryLevel.ALL || level == JobSummaryLevel.COMMON_ONLY,
```
laurentgo: since JobSummaryLevel is an enum, I guess it's really to guard against your own code than a user mistake.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
```diff
@@ -49,29 +50,48 @@ public void commitJob(JobContext jobContext) throws IOException {
...
+    JobSummaryLevel level = ParquetOutputFormat.getJobSummaryLevel(configuration);
+
+    if (level != JobSummaryLevel.NONE) {
       try {
         final FileSystem fileSystem = outputPath.getFileSystem(configuration);
         FileStatus outputStatus = fileSystem.getFileStatus(outputPath);
-        List<Footer> footers = ParquetFileReader.readAllFootersInParallel(configuration, outputStatus);
+        List<Footer> footers;
+
+        if (level == JobSummaryLevel.ALL) {
```
laurentgo: you can use a switch, which works well with enum

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
laurentgo: Looks also it fixes PARQUET-359 (see #258)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: thanks, forgot to rename this one

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
isnotinvain: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
isnotinvain: Sounds like a good idea to me, but I assumed this was intentional (don't fail the entire job because the metadata couldn't be written?)

This catches Exception, it wouldn't catch an OOM which is an Error right? 

We can make failures here fatal, but I think we'd have to discuss with the community to see if anyone cares about that behavior.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
isnotinvain: Not sure, do files written by the committer need to be deleted, or are they being written to a temp dir anyway?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
isnotinvain: haha nice.

Yeah so we need to wrap the above deletion in a try / catch (though some of those method return boolean instead of throwing, let me check)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
isnotinvain: it's to guard against adding new enum values to JobSummaryLevel, but not updating the places where it's used. Similar to having a default clause in all your switch statements. (it's too bad there aren't exhaustive switch statements).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java
isnotinvain: done

isnotinvain:
I filed PARQUET-383 for better error handling in the output committer

tsdeng:
+1

PR #278, opened 2015-09-24
PARQUET-382: Add methods to append encoded data to files.

rdblue:
This allows appending encoded data blocks to open ParquetFileWriters,
which makes it possible to merge multiple Parquet files without
re-encoding all of the records.

This works by finding the column chunk for each column in the file
schema and then streaming the encoded data from one file to the other.
New starting offsets are tracked and the column chunk metadata in the
footer is updated with the new starting positions.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -24,6 +24,7 @@ import static org.apache.parquet.hadoop.ParquetWriter.DEFAULT_BLOCK_SIZE;
 import static org.apache.parquet.hadoop.ParquetWriter.MAX_PADDING_SIZE_DEFAULT;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -34,12 +35,14 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
 import org.apache.parquet.Log;
 import org.apache.parquet.Preconditions;
+import org.apache.parquet.Strings;
 import org.apache.parquet.Version;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.BytesUtils;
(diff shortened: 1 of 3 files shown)


rdblue:
Test failure due to flaky Hadoop MemoryMangager test. See #269.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -100,21 +104,29 @@ private static boolean supportsBlockSize(FileSystem fs) {
...
+  // row group data
+  private BlockMetaData currentBlock; // appended to by endColumn
+
+  // row group data set at the start of a row group
+  private long currentRecordCount; // set in startBlock
+
+  // column chunk data accumulated as pages are written
+  private Set<Encoding> currentEncodings;
   private long uncompressedLength;
   private long compressedLength;
```
rdblue: Need to make sure these are set properly by methods that append blocks.

rdblue: Fixed and added tests.

rdblue:
@spena could you review this when you get a chance? Thanks!

spena:
@rdblue What about file schema evolution cases? Can this patch merge 2 or more files with different schema?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -457,6 +468,133 @@ public void endBlock() throws IOException {
...
+      appendRowGroup(file, block, dropColumns);
+    }
+  }
+
+  public void appendRowGroup(FSDataInputStream from, BlockMetaData rowGroup,
+                             boolean dropColumns) throws IOException {
+    startBlock(rowGroup.getRowCount());
+
+    Map<String, ColumnChunkMetaData> columnsToCopy =
+        new HashMap<String, ColumnChunkMetaData>();
```
spena: Does 'columnToCopy' need to be in order? I've seen ordered issues in Hive with HashMap on JDK 8. We use LinkedHashMap instead to avoid that. Not sure if we support JDK 8 in Parquet, but it is something to consider if it is critical the order.

rdblue: No, this is just preparing a way to look up the source columns that will be copied. The order is determined by the output file's schema and columns are added in that order to `columnsInOrder`. Columns need to match the order of the output file's schema, so they may be reordered, but the copy section handles that and copies them in contiguous chunks.

rdblue:
@spena, the schema for the output file is set when opening the file writer, so the desired output schema is known. If that schema doesn't match the incoming data file's schema (or block's schema) then there are a few cases:

* An [output column is missing from the data results in an exception](https://github.com/apache/parquet-mr/pull/278/files#diff-bb67219da9758a49d71f5d3d91b58a35R282)
* An [input column not used results in an exception if `dropColumns` is false](https://github.com/apache/parquet-mr/pull/278/files#diff-bb67219da9758a49d71f5d3d91b58a35R256)
* An [input column not used is ignored if `dropColumns` is true](https://github.com/apache/parquet-mr/pull/278/files#diff-bb67219da9758a49d71f5d3d91b58a35R217)

julienledem:
@spena you're a committer now ;)

spena:
@rdblue Overall, the patch looks pretty good. I give a +1 to this.
Just one quick question. How are these methods going to be used? Will you add a follow-up jira to call this new implementation?

rdblue:
Thanks, Sergio.

I built a couple tools in a Kite branch so we could use this to merge Impala files for performance testing (https://github.com/rdblue/kite/commit/54ba1f2a3977d0a8913bf2d8beb3b4b7aae72f9e). The typical use is this: https://github.com/rdblue/kite/blob/54ba1f2a3977d0a8913bf2d8beb3b4b7aae72f9e/kite-tools-parent/kite-tools/src/main/java/org/kitesdk/cli/commands/MergeParquetCommand.java#L126

PR #280, opened 2015-10-08
PARQUET-220: Unnecessary warning in ParquetRecordReader.initialize

sircodesalotOfTheRound:
Rather than querying the COUNTER_METHOD up front, the counter method is resolved per object. This allows us to use the
'getCounter' method on any TaskAttemptContext with the correct signature (ignoring versions where TaskAttemptContext does
not have an appropriate method/signature - preserving current behavior).

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
@@ -22,6 +22,8 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.mapreduce.Counter;
@@ -61,8 +63,9 @@ public class ContextUtil {
   private static final Field WRAPPED_CONTEXT_FIELD;
 
   private static final Method GET_CONFIGURATION_METHOD;
-  private static final Method GET_COUNTER_METHOD;
   private static final Method INCREMENT_COUNTER_METHOD;
+
+  private static final Map<Class, Method> COUNTER_METHODS_BY_CLASS = new HashMap<Class, Method>();
 
   static {
     boolean v21 = true;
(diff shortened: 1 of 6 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
```diff
@@ -251,17 +251,36 @@ public static Configuration getConfiguration(JobContext context) {
     }
   }
 
-  public static Counter getCounter(TaskInputOutputContext context,
-                                   String groupName, String counterName) {
-    return (Counter) invoke(GET_COUNTER_METHOD, context, groupName, counterName);
+  public static Counter getCounter(TaskAttemptContext context, String groupName, String counterName) {
+    Method counterMethod = findCounterMethod(context);
+    return (Counter) invoke(counterMethod, context, groupName, counterName);
```
rdblue: Is the cast still necessary now that invoke has a type parameter and casts?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
```diff
@@ -251,17 +251,36 @@ public static Configuration getConfiguration(JobContext context) {
     }
   }
 
-  public static Counter getCounter(TaskInputOutputContext context,
-                                   String groupName, String counterName) {
-    return (Counter) invoke(GET_COUNTER_METHOD, context, groupName, counterName);
```
rdblue: What happened to GET_COUNTER_METHOD? Why isn't that used in findCounterMethod?

rdblue: I think this was intended to be a one-time startup cost. GET_COUNTER_METHOD was a static field that was the method. Now that this code finds them dynamically based on class it should be fine to remove GET_COUNTER_METHOD. Since this is done once per input split, I doubt the overhead is going to matter much.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
sircodesalotOfTheRound: Oversight, I can fix that.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
sircodesalotOfTheRound: I'll split the difference and cache the counter method the first time it's used. Thanks for the catch.

sircodesalotOfTheRound:
Travis build is failing with:

````
Class org.apache.parquet.hadoop.util.counters.mapreduce.MapReduceCounterLoader
 Added Method <init>, desc (Lorg/apache/hadoop/mapreduce/TaskAttemptContext;)V, access public
 Removed Method <init>, sig (Lorg/apache/hadoop/mapreduce/TaskInputOutputContext<****>;)V, desc (Lorg/apache/hadoop/mapreduce/TaskInputOutputContext;)V, access public
[WARNING] Rule 0: org.semver.enforcer.RequireSemanticVersioningConformance failed with message:
Current codebase is incompatible with version <1.7.0>. Version should be at least <2.0.0>.
````

What does this mean?

**Comment by rdblue** [comment]
@sircodesalotOfTheRound, that means this is failing the API compatibility check. This is not binary compatible with the 1.7.0 release. I think it is safe to say that org.apach.parquet.hadoop.util.* is probably not considered public API so we just need to add an exclusion for it in the root POM. Look for the semver plugin configuration.

File: parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java [hunk 3]
```diff
@@ -251,17 +260,41 @@ public static Configuration getConfiguration(JobContext context) {
...
   }
 
   /**
    * Invokes a method and rethrows any exception as runtime exceptions.
    */
-  private static Object invoke(Method method, Object obj, Object... args) {
+  private static <T> T invoke(Method method, Object obj, Object... args) {
     try {
-      return method.invoke(obj, args);
+      return (T)method.invoke(obj, args);
```
julienledem: That would create a warning. The only safe way to do that is casting from the caller.

julienledem:
This change looks good to me but travis build is still failing.

sircodesalotOfTheRound:
Fix'd.

julienledem:
LGTM
@rdblue are we happy with org/apache/parquet/hadoop/util/** excluded from semver?

rdblue:
I think @sircodesalotOfTheRound was going to address the two comments remaining above before merging.

I'm fine excluding org/apache/parquet/hadoop/util/**

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ContextUtil.java
sircodesalotOfTheRound: I replaced `GET_COUNTER_METHOD` with a cache (`COUNTER_METHODS_BY_CLASS`).

julienledem:
@rdblue good to go?

sahilsareen:
Merge this please!

julienledem:
@rdblue Merge?

dongjoon-hyun:
Any update? :)

julienledem:
@rdblue If that's ok, I'm going to merge this.

rdblue:
Yeah, sounds good.

julienledem:
@dongjoon-hyun @sircodesalotOfTheRound @sahilsareen merged

dongjoon-hyun:
Thank you so much! :)

PR #281, opened 2015-10-16
PARQUET-367: "parquet-cat -j" doesn't show all records.

sircodesalotOfTheRound:
Added JsonRecordFormatter which formats SimpleRecords into an structure that can be used with ObjectMapper to create a valid json structure. Unit test included.

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
@@ -24,12 +24,16 @@ import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.OptionBuilder;
 import org.apache.commons.cli.Options;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 
+import org.apache.parquet.hadoop.ParquetFileReader;
 import org.apache.parquet.hadoop.ParquetReader;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 import org.apache.parquet.tools.Main;
 import org.apache.parquet.tools.read.SimpleReadSupport;
 import org.apache.parquet.tools.read.SimpleRecord;
+import org.apache.parquet.tools.json.JsonRecordFormatter;
 
 public class CatCommand extends ArgsOnlyCommand {
   public static final String[] USAGE = new String[] {
@@ -71,9 +75,12 @@ public class CatCommand extends ArgsOnlyCommand {
     try {
       PrintWriter writer = new PrintWriter(Main.out, true);
       reader = ParquetReader.builder(new SimpleReadSupport(), new Path(input)).build();
+      ParquetMetadata metadata = ParquetFileReader.readFooter(new Configuration(), new Path(input));
+      JsonRecordFormatter.JsonGroupFormatter formatter = JsonRecordFormatter.fromSchema(metadata.getFileMetaData().getSchema());
+
       for (SimpleRecord value = reader.read(); value != null; value = reader.read()) {
         if (options.hasOption('j')) {
-          value.prettyPrintJson(writer);
+          writer.write(formatter.formatRecord(value));
         } else {
           value.prettyPrint(writer);
         }
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/json/JsonRecordFormatter.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/json/JsonRecordFormatter.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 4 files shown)


parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
```diff
@@ -71,14 +75,19 @@ public void execute(CommandLine options) throws Exception {
...
         if (options.hasOption('j')) {
-          value.prettyPrintJson(writer);
+          writer.write(formatter.formatRecord(value));
         } else {
           value.prettyPrint(writer);
         }
         writer.println();
       }
+    } catch (Throwable ex) {
+      ex.printStackTrace();
```
julienledem: If we want to do this, we should put it on the side of the caller of execute(...).

parquet-tools/src/main/java/org/apache/parquet/tools/util/JsonRecordFormatter.java
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.tools.util;
```
julienledem: package org.apache.parquet.tools.json; ?
I'd avoid "util" when there's a better name

parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
```diff
@@ -71,14 +75,19 @@ public void execute(CommandLine options) throws Exception {
     try {
       PrintWriter writer = new PrintWriter(Main.out, true);
       reader = ParquetReader.builder(new SimpleReadSupport(), new Path(input)).build();
+      ParquetMetadata metadata = ParquetFileReader.readFooter(new Configuration(), new Path(input));
+      JsonRecordFormatter.JsonGroupFormatter formatter = JsonRecordFormatter.fromSchema(metadata.getFileMetaData().getSchema());
+
       for (SimpleRecord value = reader.read(); value != null; value = reader.read()) {
         if (options.hasOption('j')) {
-          value.prettyPrintJson(writer);
+          writer.write(formatter.formatRecord(value));
```
julienledem: Could you explain how this was missing records? Was it a bug in the prettyPrintJson method?

julienledem:
Thanks for the contribution!

parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
sircodesalotOfTheRound: The primary issue is that `SimpleObject` doesn't understand the concept of an array. It simulates an array by adding multiple items to the same key. The problem with that approach is that the Json representation then depends on the amount of data written. A key with only one value would appear as a singleton-object, while items with multiple entries appended to the same key would appear as an array. This would result in an inconsistent json encoding.

parquet-tools/src/main/java/org/apache/parquet/tools/util/JsonRecordFormatter.java
sircodesalotOfTheRound: Oh, I just picked a place, but let me know if there's a better place to put this.

parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
sircodesalotOfTheRound: I'm not sure why I added this actually... might have been a remnant from debugging. Will remove in final patch.

parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
sircodesalotOfTheRound: Reading you comment again I think I misunderstood your question. Basically, the issue is with `prettyPrintJson`:

```
protected Object toJsonObject() {
    Map<String, Object> result = Maps.newLinkedHashMap();
    for (NameValue value : values) {
      // This doesn't work because there might be  multiple values with the same key - OVERWRITE.
      // Could change this to a Map<String, Collection> but then we have the same problem in reverse
      // that is, most entries will have only one entry while other items may be arrays. We could
      // just print single-item collections as non-arrays, and multiple item collections as arrays
      // but that's very dirty. Doing this sort of formatting here breaks the 'Single Responsibility Principle' (TM) anyway. 
      result.put(value.getName(), toJsonValue(value.getValue()));
    }

    return result;
  }

  protected static Object toJsonValue(Object val) {
    if (SimpleRecord.class.isAssignableFrom(val.getClass())) {
      return ((SimpleRecord) val).toJsonObject();
    } else if (byte[].class == val.getClass()) {
      return new BinaryNode((byte[]) val);
    } else {
      return val;
    }
  }
``` 

It's the same basic issue I outlined in my previous comment (but I was addressing the design choice rather than why it was failing in the first place - which was your actual question I think). Please see the comment above.

Thanks

rdblue:
@julienledem, did you want to commit this?

PR #282, opened 2015-10-19
PARQUET-353: Recycle compressors in parquet write path

nitin2goyal:


liancheng:
How can we retest this PR? The last failure should have been fixed by #269.

nitin2goyal:
Commited a dummy checkin such that tests re-run automatically

liancheng:
+1

@nitin2goyal Thanks for the fix :)

julienledem:
This code has changed. you may want to rebase.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -171,6 +171,8 @@ private final STATE error() throws IOException {
 
   private STATE state = STATE.NOT_STARTED;
 
+  private final CodecFactory codecFactory;
```
julienledem: It seems that this is not the right place to put this as this class does not use the codecFactory.
it just creates it and makes it available. I'd rather have private members not accessible to the outside to respect encapsulation and layering of the code.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -341,7 +341,6 @@ private static int getMaxPaddingSize(Configuration conf) {
         throws IOException, InterruptedException {
     final WriteSupport<T> writeSupport = getWriteSupport(conf);
 
-    CodecFactory codecFactory = new CodecFactory(conf);
```
julienledem: I don't get this change. the creation has been moved to ParquetFileWriter w = new ParquetFileWriter(... just a few lines bellow. Could you explain why we do this?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -267,8 +267,7 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
         conf, schema, file, mode, blockSize, maxPaddingSize);
     fileWriter.start();
 
-    CodecFactory codecFactory = new CodecFactory(conf);
-    CodecFactory.BytesCompressor compressor =	codecFactory.getCompressor(compressionCodecName, 0);
+    CodecFactory.BytesCompressor compressor =	fileWriter.getCodecFactory().getCompressor(compressionCodecName, 0);
```
julienledem: possibly what you really want is pass the codec factory as a parameter to this constructor.

julienledem:
Thanks for looking into this. A lot of the codec factory code has changed recently.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
rdblue: +1

rdblue:
It looks like the purpose of this is to tie the lifecycle of the CodecFactory to ParquetFileWriter. That seems like a good idea (resources are freed when the file is closed), but ends up exposing internals (as Julien noted) because we don't delegate compression to ParquetFileWriter.

I think it would be cleaner to attach the lifecycle of the codec factory to the record writer for MR, which has a close method for clean up. The CodecFactory would be instantiated in the writer's constructor instead of passing in a compressor. This should also call `release` in the `ParquetWriter#close` method to avoid the same leak when not writing via MR.

rdblue:
To get this into the pending release, I've implemented the changes I recommended in #295.

PR #283, opened 2015-10-19
PARQUET-349: VersionParser does not handle versions missing 'build' section

sircodesalotOfTheRound:
This change reworks the regular expression in VersionParser.java to allow for missing 'version' and 'build' sections.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
+++ b/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
@@ -31,7 +31,7 @@ import static org.apache.parquet.Preconditions.checkArgument;
  */
 public class VersionParser {
   // example: parquet-mr version 1.8.0rc2-SNAPSHOT (build ddb469afac70404ea63b72ed2f07a911a8592ff7)
-  public static final String FORMAT = "(.+) version ((.*) )?\\(build ?(.*)\\)";
+  public static final String FORMAT = "(.*?)\\s+version\\s*(?:([^(]*?)\\s*(?:\\(\\s*build\\s*([^)]*?)\\s*\\))?)?";
   public static final Pattern PATTERN = Pattern.compile(FORMAT);
 
   public static class ParsedVersion {
@@ -113,8 +113,8 @@ public class VersionParser {
     }
 
     String application = matcher.group(1);
-    String semver = matcher.group(3);
-    String appBuildHash = matcher.group(4);
+    String semver = matcher.group(2);
+    String appBuildHash = matcher.group(3);
 
     if (Strings.isNullOrEmpty(application)) {
       throw new VersionParseException("application cannot be null or empty");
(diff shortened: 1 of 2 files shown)


parquet-common/src/main/java/org/apache/parquet/VersionParser.java
```diff
@@ -31,7 +31,7 @@
  */
 public class VersionParser {
   // example: parquet-mr version 1.8.0rc2-SNAPSHOT (build ddb469afac70404ea63b72ed2f07a911a8592ff7)
-  public static final String FORMAT = "(.+) version ((.*) )?\\(build ?(.*)\\)";
+  public static final String FORMAT = "(.+) version( |$)?(.*?)( |$)?(\\(build ?(.*)\\))?";
```
rdblue: This re still seems a bit brittle to me. What about matching white-space with \s*?

rdblue:
I think if you rebase and force-push, this will start passing tests because the flaky memory manager test has been fixed.

parquet-common/src/main/java/org/apache/parquet/VersionParser.java
sircodesalotOfTheRound: Truth be told, I think using recursive descent would probably a much more maintainable solution to this, but I wanted to make the least amount of changes. What do you think?

sircodesalotOfTheRound:
That version breaks all of the tests : /

julienledem:
+1 
@rdblue did you want to merge?

rdblue:
@sircodesalotOfTheRound, I just tried the regex I posted and was able to get it to work. The second and third groups needed to be non-greedy to avoid matching the spaces after the version, and the match groups needed to be updated. Otherwise, it works. The new one is `"(.*?)\\s+version\\s*(?:([^(]*?)\\s*(?:\\(\\s*build\\s*([^)]*?)\\s*\\))?)?"`

I'd like to use the one I posted if you think it is okay. As you said, the other looks pretty gnarly and I'm a little concerned that there may be some problematic cases for it. I also think it is better to use more liberal matching for things like the writer name (parquet-mr) with `.*?` instead of limiting the characters. Otherwise it's pretty similar to what you have already.

Here are my changes:

```diff
diff --git a/parquet-common/src/main/java/org/apache/parquet/VersionParser.java b/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
index 8f2d41c..fc4ff6c 100644
--- a/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
+++ b/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
@@ -31,7 +31,7 @@ import static org.apache.parquet.Preconditions.checkArgument;
  */
 public class VersionParser {
   // example: parquet-mr version 1.8.0rc2-SNAPSHOT (build ddb469afac70404ea63b72ed2f07a911a8592ff7)
-  public static final String FORMAT = "([A-Za-z0-9\\-]+)\\s+version(\\s+|$)?(.*?)(\\s+|$)?(\\(\\s*build\\s*(.*)\\))?";
+  public static final String FORMAT = "(.*?)\\s+version\\s*(?:([^(]*?)\\s*(?:\\(\\s*build\\s*([^)]*?)\\s*\\))?)?";
   public static final Pattern PATTERN = Pattern.compile(FORMAT);
 
   public static class ParsedVersion {
@@ -113,8 +113,8 @@ public class VersionParser {
     }
 
     String application = matcher.group(1);
-    String semver = matcher.group(3);
-    String appBuildHash = matcher.group(6);
+    String semver = matcher.group(2);
+    String appBuildHash = matcher.group(3);
 
     if (Strings.isNullOrEmpty(application)) {
       throw new VersionParseException("application cannot be null or empty");
```

sircodesalotOfTheRound:
That's fine by me. +1

rdblue:
@sircodesalotOfTheRound, could you apply that patch and update the PR? Then I'll merge it.

julienledem:
@rdblue does this look good to you?

rdblue:
Yeah, I didn't see that it was updated. Thanks!

phoenixhadoop:
The patch is very good ,thank you very much!

PR #285, opened 2015-10-24
try to use class loader from Hadoop Configuration

smartkiwi:
(thread local clas loader) if system class loader fails to find the class.

Apache Spark uploads jars to after Java executor process starts, so they are not available in system class loader.

This PR solves the problem described in this post - allows to use User Defined Predicate classes and objects.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
@@ -27,6 +27,7 @@ import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.input.ClassLoaderObjectInputStream;
 import org.apache.hadoop.conf.Configuration;
 
 import org.apache.parquet.Closeables;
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
```diff
@@ -97,7 +98,12 @@ public static void writeObjectToConfAsBase64(String key, Object obj, Configurati
       bais = new ByteArrayInputStream(bytes);
       gis = new GZIPInputStream(bais);
       ois = new ObjectInputStream(gis);
-      return (T) ois.readObject();
+      try {
+        return (T) ois.readObject();
+      } catch (ClassNotFoundException e) {
+        ClassLoaderObjectInputStream clois = new ClassLoaderObjectInputStream(conf.getClassLoader(), gis);
```
rdblue: I don't think the Configuration's ClassLoader is appropriate here either. The problem is that framework classes are loaded in some cases by a ClassLoader that is a parent to the ClassLoader used for application or user classes. Anything that is shipped in the user Jar could get ignored in that case. I think the right thing to do for now is to use the Thread context ClassLoader.

Also, this method should use whatever ClassLoader is passed in rather than trying the SerializationUtil's loader and then trying another one. All ClassLoader delegation should happen in the ClassLoader, not the calling class.

I recommend adding a ClassLoader option to this method and using that with the ClassLoaderObjectInputStream, and then adding a replacement method that defaults the ClassLoader to the Thread context's ClassLoader. Does that sound like it will work for your application?

julienledem:
@smartkiwi what did you think of @rdblue's comment?

PR #287, opened 2015-11-10
PARQUET-391: Add thrift dependency for generated sources in parquet-cascading, parquet-scrooge modules

saucam:
When compiling with -Pthrift9  , parquet build fails in parquet-cascading module since we have not specified the thrift dependency ; and compilation breaks for the generated sources from thrift 0.9


saucam:
closed in favor of https://github.com/apache/parquet-mr/pull/276

PR #289, opened 2015-11-21
PARQUET-352: Add object model property to file footers.

rdblue:
WriteSupport now has a getName getter method that is added to the footer
if it returns a non-null string as writer.model.name. This is intended
to help identify files written by object models incorrectly.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/WriteSupport.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/WriteSupport.java
@@ -121,6 +121,18 @@ abstract public class WriteSupport<T> {
   public abstract void write(T record);
 
   /**
+   * Called to get a name to identify the WriteSupport object model.
+   * If not null, this is added to the file footer metadata.
+   * <p>
+   * Defining this method will be required in a future API version.
+   *
+   * @return a String name for file metadata.
+   */
+  public String getName() {
+    return null;
+  }
+
+  /**
    * called once in the end after the last record was written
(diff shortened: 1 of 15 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/WriteSupport.java
```diff
@@ -121,6 +121,18 @@ public FinalizedWriteContext(Map<String, String> extraMetaData) {
...
+   * Called to get a name to identify the WriteSupport object model.
+   * If not null, this is added to the file footer metadata.
+   * <p>
+   * Defining this method will be required in a future API version.
+   *
+   * @return a String name for file metadata.
+   */
+  public String getName() {
+    return null;
+  }
```
julienledem: we could also add a constructor parameter. (and keep the original default constructor for compatibility)

julienledem: and mark the old one deprecated

julienledem:
I made a comment about passing this to the constructor instead, but this looks good to me either way.
+1

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/WriteSupport.java
rdblue: My reasoning is that using the constructor requires updates in a lot more places and it would be easier to overlook cases that need it.

rdblue:
Thanks for reviewing, Julien!

PR #290, opened 2015-11-25
PARQUET-305: Update logging to SLF4J.

rdblue:
This removes the Log implementation based on java.util.logging and
replaces it with SLF4J. The compiler removal of debug log messages still
works because Log.DEBUG and similar final constants are unchanged.

This commit adds slf4j-simple as the test logger implementation.
Configuration for slf4j-simple is in the root pom. Two modules can't use
slf4j-simple, parquet-pig and parquet-thrift, and use slf4j-log4j12
instead because pig depends on log4j and tests die without it.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/Log.java
+++ b/parquet-common/src/main/java/org/apache/parquet/Log.java
@@ -34,7 +34,6 @@ import java.util.logging.Level;
  * @author Julien Le Dem
  *
  */
-@Deprecated
 public class Log {
 
   /**
@@ -53,7 +52,6 @@ public class Log {
    * @return the corresponding logger
    * @deprecated will be removed in 2.0.0; use org.slf4j.LoggerFactory instead.
    */
-  @Deprecated
   public static Log getLog(Class<?> c) {
     return new Log(c);
   }
--- a/parquet-common/src/main/java/org/apache/parquet/Log.java
+++ b/parquet-common/src/main/java/org/apache/parquet/Log.java
@@ -18,16 +18,9 @@
  */
 package org.apache.parquet;
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
-import java.text.MessageFormat;
-import java.util.Date;
-import java.util.logging.Formatter;
-import java.util.logging.Handler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import java.util.logging.Level;
-import java.util.logging.LogRecord;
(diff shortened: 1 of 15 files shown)


rdblue:
@julienledem, as long as I was changing the logging code to fix PARQUET-305, I went ahead and switched it over to SLF4J. Could you have a look?

I marked the `Log` API deprecated. Do you think we should take the time right now to move off of it completely, or do it over the next release?

julienledem:
Thanks for updating the dependency and use slf4j under the hood.
I don't see the need to migrate all the logs to the slf4j api.
I would not deprecate the Log class and just keep it as is as a seen wrapper to slf4j.

rdblue:
The problem with not moving to the slf4j API is that it forces the caller to do a lot of string manipulation to construct a log message regardless of whether it is actually used. The varargs style API defers constructing the log message.

I also don't see much value in having a wrapper around the logging implementation, since that is what slf4j is supposed to do. That API already allows the downstream application to switch out the logger implementation used by Parquet.

My reasoning is: why maintain our own API when it results in not being able to use features of the underlying one that is already general-purpose?

parquet-common/src/main/java/org/apache/parquet/Log.java
```diff
@@ -41,6 +34,7 @@
  * @author Julien Le Dem
  *
  */
+@Deprecated
```
julienledem: I'd rather not make this deprecated in this PR since it will make every class have warnings.
Otherwise, this looks good to me.

julienledem:
+1

parquet-common/src/main/java/org/apache/parquet/Log.java
rdblue: No problem. I'll open a follow-up issue for further discussion. We can deprecate and remove the use of this class for that issue instead of dragging out this one.

julienledem: SGTM. Thanks

proflin:
hi @rdblue, I wonder if there is any following-up discussion/JIRA/PR concerning replacing the use of Log.java with slf4j? Maybe I can do the replacing, should we reach consensus.

julienledem:
I don't recall there's a jira already. If not please make one so that we can have the discussion.

rdblue:
I opened an issue to move over to SLF4J already. It is: https://issues.apache.org/jira/browse/PARQUET-401

PR #292, opened 2015-11-26
PARQUET-334: UT test failure with Pig 0.15

tfriedr:
I made a few updates to the original patch PARQUET-334-1.patch proposed by Daniel. As the inputschema is maintained in EvalFunc, any reference to the private class variable inputSchema should be changed to getInputSchema(), and inputSchema can be removed because it will be null always.


julienledem:
+1

julienledem:
@tfriedr can you update the description to ```PARQUET-334:...```? (with a dash, not an underscore)

tfriedr:
Thanks, Julien. Updated the title.

julienledem:
Thank you @tfriedr !

PR #295, opened 2015-12-01
PARQUET-353: Release compression resources.

rdblue:
This updates the use of CodecFactory in the output format and writer
classes so that its lifecycle is tied to ParquetWriter and
ParquetRecordWriter. When those classes are closed, the resources held
by the CodecFactory associated with the instance are released.

This is an alternative to and closes #282.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
@@ -21,6 +21,7 @@ package org.apache.parquet.hadoop;
 import java.io.IOException;
 import java.util.Map;
 
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 
@@ -28,6 +29,7 @@ import org.apache.parquet.column.ParquetProperties;
 import org.apache.parquet.column.ParquetProperties.WriterVersion;
 import org.apache.parquet.hadoop.CodecFactory.BytesCompressor;
 import org.apache.parquet.hadoop.api.WriteSupport;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.schema.MessageType;
 
 import static org.apache.parquet.Preconditions.checkNotNull;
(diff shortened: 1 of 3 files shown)


julienledem:
LGTM

julienledem:
@jaltekruse did you have comments on this one?

rdblue:
I'll rebase this and merge once #250 is in with the ParquetProperties refactor to avoid blocking the larger change.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -300,6 +301,8 @@ public void close() throws IOException {
     } catch (InterruptedException e) {
       throw new IOException(e);
     }
+    // release after the writer closes in case it is used for a last flush
+    codecFactory.release();
```
jaltekruse: should this be in a finally block to release in the case where an exception occurs?

rdblue: Yep.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
```diff
@@ -112,6 +153,10 @@ public ParquetRecordWriter(
   @Override
   public void close(TaskAttemptContext context) throws IOException, InterruptedException {
     internalWriter.close();
+    // release after the writer closes in case it is used for a last flush
+    if (codecFactory != null) {
```
jaltekruse: similar comment as below, if close above can throw we should put it in a try and call release from a finally block. This method can still throw the checked exceptions, no real need to re-wrap them.

rdblue: Yeah, I wrapped this in a try/finally just to be safe as well. I'm just rebasing on the PARQUET-99 fix and I'll post a new copy.

rdblue:
Okay, rebased on PARQUET-99 and ready for another look. Thanks @julienledem and @jaltekruse!

rdblue:
@jaltekruse, can you take a quick look at this patch? I've addressed the problem you identified and I'd like to get it in. Thanks for your help!

jaltekruse:
+1

rdblue:
Thanks!

PR #296, opened 2015-12-01
Add predicate pushdown using filter2 api

danielcweeks:


rdblue:
@isnotinvain, you might want to have a look at this one.

I'll take a look next week as well. Is this something you'd like to get into 1.9.0?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -70,9 +83,12 @@
  * @author Julien Le Dem
  *
  */
-public class ParquetLoader extends LoadFunc implements LoadMetadata, LoadPushDown {
+public class ParquetLoader extends LoadFunc implements LoadMetadata, LoadPushDown, LoadPredicatePushdown {
   private static final Log LOG = Log.getLog(ParquetLoader.class);
 
+  public static final String ENABLE_PREDICATE_FILTER_PUSHDOWN = "parquet.pig.predicate.pushdown.enable";
+  private static final boolean DEFAULT_PREDICATE_PUSHDOWN_ENABLED = false;
```
rdblue: What's the rationale behind not using predicate pushdown by default? Not breaking existing jobs until it is well tested?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -169,6 +185,7 @@ private void setInput(String location, Job job) throws IOException {
     getConfiguration(job).set(PARQUET_PIG_SCHEMA, pigSchemaToString(schema));
     getConfiguration(job).set(PARQUET_PIG_REQUIRED_FIELDS, serializeRequiredFieldList(requiredFieldList));
     getConfiguration(job).set(PARQUET_COLUMN_INDEX_ACCESS, Boolean.toString(columnIndexAccess));
+    SerializationUtil.writeObjectToConfAsBase64(ParquetInputFormat.FILTER_PREDICATE, getFromUDFContext(ParquetInputFormat.FILTER_PREDICATE), getConfiguration(job));
```
rdblue: Should this use [`ParquetInputFormat.setFilterPredicate`](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java#L187) instead?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
danielcweeks: Yes, but if other who are more familiar with pig pushdown or filter api think it's worth turning on by default, I'm OK with that as well.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
danielcweeks: I stole this line from somewhere else, but this looks like a better approach.

I'll update.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+      }
+
+    } catch (FrontendException e) {
+      throw new RuntimeException("Error processing pushdown for column:" + c , e);
+    }
+
+    return null;
+  }
+
+  private static class ColumnWrapper<T extends Comparable<T>> extends Operators.Column<T> implements Operators.SupportsLtGt, Serializable {
```
rdblue: I don't think this class is needed. The classes returned by `binaryColumn`, `intColumn`, etc. already implement `Operators.SupportsLtGt` and `Serializable`.

Also, this class acts as a Column and wraps a Column. The getter methods already return the values from wrapped because they are passed into the Column constructor. The value of this class seems to be that it can return the original wrapped column, but it behaves like the wrapped column in all cases so I'd just eliminate it.

rdblue: Actually, I see what you're doing here: creating a type that is definitely an Operators.Column and Operators.SupportsLtGt. I think we can handle this using type parameters, though. I'll look into it.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+      return wrapped.getColumnPath();
+    }
+  }
+
+  private Comparable getValue(Expression expr) {
+    switch(expr.getOpType()) {
+      case TERM_COL:
+        //return ((Column) expr).getName();
+      case TERM_CONST:
+        Comparable value = (Comparable) ((Const) expr).getValue();
```
rdblue: As long as the expression is cast to a Const, why not check that it is a Const and use that in this method signature?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+    storeInUDFContext(ParquetInputFormat.FILTER_PREDICATE, pred);
+  }
+
+  private FilterPredicate buildFilter(Expression e) {
+    if (e instanceof BinaryExpression) {
+      Expression lhs = ((BinaryExpression) e).getLhs();
+      Expression rhs = ((BinaryExpression) e).getRhs();
+
+      switch (e.getOpType()) {
+        case OP_EQ: return eq(getColumn((Column) lhs), getValue(rhs));
```
rdblue: What if the column is on the right and the value on the left? It seems like "4 = i" should work as well. Does Pig normalize?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+        case OP_GT: return gt(getColumn((Column) lhs), getValue(rhs));
+        case OP_GE: return gtEq(getColumn((Column) lhs), getValue(rhs));
+        case OP_LT: return lt(getColumn((Column) lhs), getValue(rhs));
+        case OP_LE: return ltEq(getColumn((Column) lhs), getValue(rhs));
+        case OP_AND: return and(buildFilter(lhs), buildFilter(rhs));
```
rdblue: If any sub-expression can't be built then what happens to the filter? Should this return null if the result of rhs or lhs is null?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+      case TERM_COL:
+        //return ((Column) expr).getName();
+      case TERM_CONST:
+        Comparable value = (Comparable) ((Const) expr).getValue();
+
+        if(value instanceof String) {
+          value = Binary.fromString((String)value);
+        }
+        return value;
+      default:
```
rdblue: I think this should reject anything that isn't a Const and cause the calling invocation of buildFilter to return null.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +397,133 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+  }
+
+  private ColumnWrapper getColumn(Column c) {
+    String column = c.getName();
+
+    try {
+      FieldSchema f = schema.getField(column);
+
+      switch (f.type) {
+        case DataType.BOOLEAN: return new ColumnWrapper<Binary>(binaryColumn(f.alias));
```
rdblue: Boolean should use `booleanColumn`, but that is not SupportsLtGt so it will need to be done separately. This is a bug that is hidden by using ColumnWrapper.

rdblue:
@danielcweeks, I found a few things in the filter generation code that looks odd to me so I updated it in my local copy. You might want to have a look at https://github.com/rdblue/parquet-mr/commit/a39fdffcf171b0806c71cb9d0d78b05dac609d38, which fixes most of the issues I pointed out. It doesn't support BOOLEAN yet, but that wouldn't be hard to add.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
danielcweeks: I'm not entirely clear on this one.  I believe that we can't restrict to Const because you can reference another column:  c1 == c2.  I don't know if the filter api actually supports this type of comparison.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
danielcweeks: Yes, I believe Pig normalizes the expression so the column is on the left and value is on the right.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
danielcweeks: I'm not entirely clear, but I think we should change it to throw an exception.  If we trust pig to abide by only sending op types that we allow, than it shouldn't really matter, but it's probably best to throw something just in case.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
rdblue: If that's the case, then I don't think this method should be called at all. Parquet doesn't support pushdown comparison between columns, so I think the right thing to do is to check if it is a `Const` and return null from `buildFilter` instead of calling this and causing a `RuntimeException`.

rdblue: Looks like you're right about throwing exceptions. I thought that the return value of buildFilter signaled that pushdown failed for some reason, but the setPushdownPredicate method has a void return. Throwing an exception looks like the right way to go.

danielcweeks: Also, we wouldn't want to return null because that will be then be used as the value (e.g. c1 == c2 would turn into c1 == null, which is a valid filter, but would produce the wrong result)

rdblue: Yeah, I meant return null from buildFilter and not even calling this method if the expression isn't a Const

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
isnotinvain: Hey, sorry for the slow replies. If you want a colum that suports lt and gt, then just use a type bound like `T extends Column & SupportsLtGt` right?

Or is the issue that boolean column doesn't extend SupportsLtGt and you'd like it to? we could just update boolean column to extend SupportsLtGt. I probably didn't do that originally because it seems more like a mistake to try to compare booleans that way, but if it's something pig allows then I guess we can allow it too.

rdblue: Yeah, the issue is getting the type system to recognize it because you can't use `T extends Column & SupportsLtGt` in some places. You can declare a method with a type parameter, but T is distinct from IntColumn because Java infers the concrete type of T from the invocation context so casting IntColumn to T requires a cast. You can, however, pass IntColumn to such a method.

The easiest way is to use the ColumnWrapper, but I have a branch that gets around it: https://github.com/rdblue/parquet-mr/commit/a39fdffcf171b0806c71cb9d0d78b05dac609d38#diff-19a7f147ae735f66dcb5f580ea724754R521

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +401,132 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+        default: throw new RuntimeException("Unsupported operation type: " + e.getOpType());
+      }
+    } else {
+      throw new RuntimeException("Unsupported expression type: " + e);
+    }
+  }
+
+  private ColumnWrapper getColumn(Column c) {
+    String column = c.getName();
+
```
danielcweeks: This was a mistake on my part and should have been new ColumnWrapper<Boolean>(booleanColumn(f.alias));

In a sense this introduces a "bug", but I'm not sure you can even represent that in pig latin, so it's more to improve readability and let pig sort out the conflict.  Conceptually, I don't like it, but it makes it pretty clean.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +401,132 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+    try {
+      FieldSchema f = schema.getField(column);
+
+      switch (f.type) {
+        case DataType.BOOLEAN: return new ColumnWrapper<Boolean>(booleanColumn(f.alias));
```
danielcweeks: @rdblue (your comment here got stomped with the updates)

This was a mistake on my part and should have been new ColumnWrapper<Boolean>(booleanColumn(f.alias));

In a sense this introduces a "bug", but I'm not sure you can even represent that in pig latin, so it's more to improve readability and let pig sort out the conflict. Conceptually, I don't like it, but it makes it pretty clean.

danielcweeks:
@rdblue I'll take a look to see how boolean would fit into your branch and we can take a look at the two approaches.

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
julienledem: I think it makes sense to have it on by default once we're confident.
Possibly once it has run in production and works well?

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +401,149 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+      case OP_GT: return gt(col, value);
+      case OP_GE: return gtEq(col, value);
+      case OP_LT: return lt(col, value);
+      case OP_LE: return ltEq(col, value);
+    }
+    return null;
+  }
+
+  private <C extends Comparable<C>> C getValue(Expression expr, Class<C> type) {
+    Comparable value = (Comparable) ((Const) expr).getValue();
```
julienledem: ```type.cast(((Const) expr).getValue())```?

julienledem:
This looks good to me.

rdblue:
I think the only thing that still needed to be done was to throw exceptions rather than returning null when a filter can't be constructed. There was discussion about this above, but it was on an outdated diff.

I made those changes and pushed them to https://github.com/rdblue/parquet-mr/tree/PARQUET-397-pig-predicate-pushdown for review.

I also implemented not, !=, in, and between operations because it seems likely that those will be used. In and between build expressions based on comparison operators, or, and and.

PR #297, opened 2015-12-02
PARQUET-99: Add page size check properties

rdblue:
This adds properties to set the min and max number of records that are passed between page checks, as well as a property that controls whether the next check will be based on records already seen or set to the minimum number of records between checks.

* `parquet.page.size.row.check.min` - minimum number of records between page size checks
* `parquet.page.size.row.check.max` - maximum number of records between page size checks
* `parquet.page.size.check.estimate` - whether to estimate the number of records before the next check, or to always use the minimum number of records.

This also updates the internal API to use ParquetProperties to carry encoding settings (used in parquet-column) to reduce the number of parameters passed through internal APIs. It also adds a builder for ParquetProperties to avoid needing to reference defaults in other modules.

This closes #250


danielcweeks:
I like this. ParquetProperties didn't really come into use until later which I think is why v1 is so convoluted.

This is a good refactor and will help to expose more configurability as we've been hardcoding many of these properties.

rdblue:
Thanks, Dan! Sorry to take so long getting back to this. I'll rebase and clean it up next week and we can get both PARQUET-99 (#250) and this one in for the 1.9.0 release.

rdblue:
Okay, I've cleaned up my changes on top of #250 and tests are now passing (at least locally).

I think the next step is to either cherry-pick e072bb9 into #250 or just update and merge this one. Doesn't matter to me which one. @danielcweeks, could you review my commit on top of yours?

@julienledem, you may want to have a look at the API changes as well.

Thanks!

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -75,53 +87,83 @@ public static WriterVersion fromString(String name) {
...
-  public ValuesWriter getColumnDescriptorValuesWriter(int maxLevel, int initialSizePerCol, int pageSize) {
+  public ValuesWriter newRepetitionLevelWriter(ColumnDescriptor path) {
+    return newColumnDescriptorValuesWriter(path.getMaxRepetitionLevel());
+  }
+
+  public ValuesWriter newDefinitionLevelWriter(ColumnDescriptor path) {
+    return newColumnDescriptorValuesWriter(path.getMaxDefinitionLevel());
+  }
+
+  public ValuesWriter newColumnDescriptorValuesWriter(int maxLevel) {
```
julienledem: package protected?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
```diff
@@ -37,20 +38,35 @@
...
   private final Map<ColumnDescriptor, ColumnWriterV1> columns = new TreeMap<ColumnDescriptor, ColumnWriterV1>();
   private final PageWriteStore pageWriteStore;
-  private final int pageSizeThreshold;
-  private final int dictionaryPageSizeThreshold;
-  private final boolean enableDictionary;
-  private final WriterVersion writerVersion;
-  private final ByteBufferAllocator allocator;
+  private final ParquetProperties props;
 
   public ColumnWriteStoreV1(PageWriteStore pageWriteStore, int pageSizeThreshold, int dictionaryPageSizeThreshold, boolean enableDictionary, WriterVersion writerVersion, ByteBufferAllocator allocator) {
```
julienledem: Deprecated ?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
```diff
@@ -37,20 +38,35 @@
...
+        .withPageSize(pageSizeThreshold)
+        .withDictionaryPageSize(dictionaryPageSizeThreshold)
+        .withDictionaryEncoding(enableDictionary)
+        .estimateRowCountForPageSizeCheck(true)
+        .withWriterVersion(writerVersion)
+        .withAllocator(allocator)
+        .build());
+  }
+
+  public ColumnWriteStoreV1(PageWriteStore pageWriteStore, int pageSizeThreshold, int dictionaryPageSizeThreshold, boolean enableDictionary, int initialRowCountForPageSizeCheck, boolean estimateNextSizeCheck, WriterVersion writerVersion, ByteBufferAllocator allocator) {
```
julienledem: Do we need this one?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -341,22 +362,32 @@ private static int getMaxPaddingSize(Configuration conf) {
...
+    if (INFO) LOG.info("Dictionary is " + (props.isEnableDictionary() ? "on" : "off"));
     if (INFO) LOG.info("Validation is " + (validating ? "on" : "off"));
-    WriterVersion writerVersion = getWriterVersion(conf);
-    if (INFO) LOG.info("Writer version is: " + writerVersion);
-    int maxPaddingSize = getMaxPaddingSize(conf);
+    if (INFO) LOG.info("Writer version is: " + props.getWriterVersion());
     if (INFO) LOG.info("Maximum row group padding size is " + maxPaddingSize + " bytes");
+    if (INFO) LOG.info("Page size checking is: " + (props.estimateNextSizeCheck() ? "estimated" : "constant"));
+    if (INFO) LOG.info("Min row count for page size check is: " + props.getMinRowCountForPageSizeCheck());
+    if (INFO) LOG.info("Min row count for page size check is: " + props.getMaxRowCountForPageSizeCheck());
```
julienledem: add this to ParquetProperties? toDisplayString?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestUtils.java
```diff
@@ -53,6 +54,10 @@ public static void assertThrows(
       Assert.fail("No exception was thrown (" + message + "), expected: " +
           expected.getName());
     } catch (Exception actual) {
+      if (!actual.equals(expected)) {
+        System.err.println("Caught an unexpected exception:");
+        actual.printStackTrace(System.err);
+      }
       Assert.assertEquals(message, expected, actual.getClass());
```
julienledem: possibly addSuppressed to the exception thrown by assertEquals

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -234,26 +280,139 @@ public ByteBufferAllocator getAllocator() {
...
+
+    public Builder withMaxRowCountForPageSizeCheck(int max) {
+      Preconditions.checkArgument(max > 0,
+          "Invalid row count for page size check (negative): %s", max);
+      this.maxRowCountForPageSizeCheck = max;
+      return this;
+    }
+
+    // Do not attempt to predict next size check.  Prevents issues with rows that vary significantly in size.
+    public Builder estimateRowCountForPageSizeCheck(boolean estimateNextSizeCheck) {
```
julienledem: with?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -55,6 +57,16 @@
  */
 public class ParquetProperties {
```
julienledem: The name is a little outdated now that you can pass an allocator.
ParquetConfiguration?

julienledem:
@rdblue This is a much needed cleanup and refactoring. Thanks for doing this! I made some comment above. Otherwise this looks good to me.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
rdblue: `withEstimate` implies that you're setting an estimate because it makes estimate a noun. Using estimate as the verb makes the boolean parameter make sense: `estimateRowCountForPageSizeCheck(false)`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
rdblue: I thought about it, but the problem is that this is only printed once and not all of the settings are in parquet properties so it doesn't entirely clean it up. I think we should leave as-is for now and refactor later if we end up needing to reuse it.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
rdblue: Fixed

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
rdblue: Removed since this is internal

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreV1.java
rdblue: Removed.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestUtils.java
rdblue: Good idea

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
rdblue: Since this class defines `ParquetProperties.WriterVersion` that is exposed in the public API, I think it would be a breaking change to rename it. This class is otherwise not part of the public API (and I've tried to keep it that way here) so we should fix this when we separate the API into well-defined packages.

rdblue:
Updated for Julien's comments. Thanks for reviewing!

julienledem:
+1

PR #300, opened 2015-12-11
PARQUET-212: Implement LIST read compatibility rules in Thrift

rdblue:
This implements the read-side compatibility rules for 2-level and 3-level lists in Thrift.

Thrift doesn't allow null elements inside lists, but 3-level lists may have optional elements. This PR adds a property, parquet.thrift.ignore-null-elements, that allows thrift to read lists with optional elements by ignoring nulls. This is off by default, but is provided as an opt-in for compatibility with data written by Hive.

Thrift's schema conversion does not change because a Thrift class (or Scrooge etc.) must be set in a file's metadata or provided when constructing a reader.

This replaces and closes #144.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -32,8 +32,10 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.reflect.ReflectData;
@@ -753,13 +755,14 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
       // synthetic wrapper. Must be a group with one optional or required field
       return true;
     } else if (elementSchema != null &&
-        elementSchema.getType() == Schema.Type.RECORD &&
-        elementSchema.getFields().size() == 1 &&
-        elementSchema.getFields().get(0).name().equals(
-            repeatedType.asGroupType().getFieldName(0))) {
+        elementSchema.getType() == Schema.Type.RECORD) {
+      Set<String> fieldNames = new HashSet<String>();
+      for (Schema.Field field : elementSchema.getFields()) {
+        fieldNames.add(field.name());
+      }
       // The repeated type must be the element type because it matches the
       // structure of the Avro element's schema.
-      return true;
+      return fieldNames.contains(repeatedType.asGroupType().getFieldName(0));
     }
     return false;
   }
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
@@ -20,6 +20,7 @@ package org.apache.parquet.thrift;
 
 import com.twitter.elephantbird.thrift.TStructDescriptor;
 import com.twitter.elephantbird.thrift.TStructDescriptor.Field;
+import org.apache.parquet.schema.Type;
 import org.apache.thrift.TBase;
 import org.apache.thrift.TEnum;
 import org.apache.thrift.TUnion;
@@ -36,6 +37,8 @@ import org.apache.parquet.thrift.struct.ThriftTypeID;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+
+import static org.apache.parquet.schema.Type.Repetition.REPEATED;
 
 /**
  * Given a thrift class, this class converts it to parquet schema,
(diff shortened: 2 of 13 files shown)


rdblue:
@isnotinvain you may want to have a look at this since you guys do the most Thrift work. This adds support in Thrift to read data written by the other object models correctly.

isnotinvain:
I'll try to take a look thanks

rdblue:
Thanks, Alex!

@liancheng may have some comments as well since he is familiar with the compatibility rules.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -753,13 +755,14 @@ static boolean isElementType(Type repeatedType, Schema elementSchema) {
...
-            repeatedType.asGroupType().getFieldName(0))) {
+        elementSchema.getType() == Schema.Type.RECORD) {
+      Set<String> fieldNames = new HashSet<String>();
+      for (Schema.Field field : elementSchema.getFields()) {
+        fieldNames.add(field.name());
+      }
       // The repeated type must be the element type because it matches the
       // structure of the Avro element's schema.
-      return true;
+      return fieldNames.contains(repeatedType.asGroupType().getFieldName(0));
```
liancheng: I don't quite get the purpose of this change. Is this mainly for more robust support for schema evolution? Take the following Avro schema and Parquet schema as an example:

```
// Avro
record Struct {
  int f0;
  int f1;
}

// Parquet
message StructArray {
  required struct_array (LIST) {
    repeated group array {
      required int32 f0;
    }
  }
}
```

According to the condition of this `if` branch, we should interpret `struct_array` as a `LIST` of Avro `Struct`, with all `f1` field set as `null`. However, `f1` defined in `Struct` isn't optional and can't be null.

Did I miss something here?

liancheng: My last comment is a little bit ambiguous. In the example I made, the Avro record `Struct` is the `elementSchema`, while the Parquet repeated group `array` is the `repeatedType` in the code.

liancheng: I guess what is missing here is that we should check all the missing fields are indeed optional, unless this constraint has already been guaranteed somewhere else (then let's document it in the comment).

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
```diff
@@ -98,6 +103,39 @@ private static StructType toStructType(TStructDescriptor struct) {
...
+      // synthetic wrapper. Must be a group with one optional or required field
+      return true;
+    } else if (thriftElement != null && thriftElement.getType() instanceof StructType) {
+      Set<String> fieldNames = new HashSet<String>();
+      for (ThriftField field : ((StructType) thriftElement.getType()).getChildren()) {
+        fieldNames.add(field.getName());
+      }
+      // If the repeated type is a subset of the structure of the ThriftField,
+      // then it must be the element type.
+      return fieldNames.contains(repeatedType.asGroupType().getFieldName(0));
```
liancheng: Same comments as the ones I made in `AvroRecordConverter.java`.

liancheng:
Got some questions about `isListElementType` updates.

Also, it would be nice to add parquet-thrift test cases that cover PARQUET-364, e.g., something like [this][1].

[1]: https://github.com/liancheng/parquet-compat/blob/with-parquet-mr-1.7.0/src/test/scala/com/databricks/parquet/schema/ListCompatibilitySuite.scala#L62-L111

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
rdblue: This is for the case where only one field of a nested record is projected. Avro has a projection schema and an expected schema so they can differ. Previously, the logic to match the requested schema (elementSchema) with the file's structure checked whether the name of the single field matched the name of the single field of the element schema. But that left out the case where the two do match (as in your example) but the requested schema has an optional field that should be defaulted.

Whether the other field is optional doesn't matter at this point in the code. Avro fills in default values later and throws an exception if something is missing and has no default at that point.

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConverter.java
rdblue: Same as above. The field projection (set by a filter expression) is independent of the Thrift type used to carry data. Nulls are added later and handled separately.

rdblue:
@liancheng, you're right that I still need to add more test cases, thanks for pointing that out.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
liancheng: Oh I see, this totally makes sense. Thanks for the explanation!

rdblue:
@liancheng, I added all of the test cases that were added in PARQUET-364 to this commit. I think that addresses all of your concerns.

@isnotinvain, do you still want to have a look at this?

liancheng:
+1

rdblue:
Thanks @liancheng!

PR #301, opened 2015-12-14
PARQUET-404: Replace git@github.com.apache for HTTPS URL on dev/README.md to avoid permission issues

spena:
…lic key on committer github account is needed

@julienledem @rdblue @liancheng Could you help me review this trivial patch for dev/README.md?

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -26,7 +26,7 @@ Merging a pull request requires being a committer on the project.
 * How to merge a Pull request:
 have an apache and apache-github remote setup
 ```
-git remote add apache-github git@github.com:apache/parquet-mr.git
+git remote add apache-github https://github.com/apache/parquet-mr.git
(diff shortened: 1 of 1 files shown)


rdblue:
Is that true? I thought that you could use the HTTPS URL without problems if you were doing read-only operations.

spena:
git@github.com:apache/parquet-mr.git did not work for me when running the script because I did not have the public key in my github account, and it was only fetching code when it failed. Should we replace git@ to the https url instead?

rdblue:
Yeah, we should try it with the HTTPS URL. It isn't surprising that the SSH URL fails because you can't log in at all. The HTTPS one should work for read-only and I think it is better to recommend that one.

spena:
@rdblue I did the change to HTTPS. The permission issue does not happen even if I don't have my public key on my github.

rdblue:
+1

Thanks Sergio!

liancheng:
+1

PR #302, opened 2015-12-18
PARQUET-410: Fix hanging subprocess call in merge script.

rdblue:
This removes the option that redirects stderr to stdout because it
causes git push to hang.


julienledem:
+1

PR #307, opened 2016-01-10
PARQUET-421: Fix mismatch of javadoc names and method parameters in m...

proflin:
…odule encoding, column, and hadoop

Codes change now and then, but some corresponding doc comments are left out.

This PR fixes only the doc comments that should have been changed. It should be OK, since none codes are touched.

@rdblue could you take a look please? Cheers.


rdblue:
Looks good to me, @proflin. Thanks for fixing docs issues!

Could you add a Parquet JIRA issue for this and add it to this PR's summary? We only merge PRs that start like "PARQUET-NNN: Fix mismatch of javadoc names and method parameters".

proflin:
JIRA ticket added & this PR's title updated.

Thanks for taking time reviewing this! :-)

PR #308, opened 2016-01-12
PARQUET-422: Fix a potential bug in MessageTypeParser where we ignore…

proflin:
… and overwrite the initial value of a method parameter

In org.apache.parquet.schema.MessageTypeParser, for addGroupType() and addPrimitiveType(), the initial value of this parameter t is ignored, and t is overwritten here.

This often indicates a mistaken belief that the write to the parameter will be conveyed back to the caller.

This is a bug found by FindBugs™.


proflin:
@julienledem @rdblue could  you take a look at this please? Cheers.

PR #309, opened 2016-01-12
PARQUET-423 Make writing Avro to Parquet less noisy

nielsbasjes:
When writing Avro records into Parquet files I see a log of logging about the columns that have been written. In the scenario I'm working on (streaming Apache Flink, many files, large schema) this becomes too much logging.

I added an additional option when creating a ParquetWriter instance (like the AvroParquetWriter) where you can control this logging by adding a ".withColumnInfoLogging(false)" to the builder.

I would like to write a unit test for this. To do this I can use some directions on how to capture the logging output in a unit test.

```
        ParquetWriter<Record> parquetWriter = AvroParquetWriter
                .<Record>builder(outputFile)
                .withSchema(Record.getClassSchema())
                .withCompressionCodec(CompressionCodecName.GZIP)
                .withDictionaryEncoding(true)
                .withWriterVersion(ParquetProperties.WriterVersion.PARQUET_2_0)
                .withWriteMode(ParquetFileWriter.Mode.OVERWRITE)
                .withColumnInfoLogging(false)
                .build();
```


nielsbasjes:
Note that I deliberately made the default behavior the same as the current version.

rdblue:
@nielsbasjes, thanks for looking into this to fix the annoyance!

I'm wondering if an API-level solution is the right choice, though. Do you think that the logs you're suppressing here would be better as DEBUG logs instead of INFO? That would delegate to the logging framework rather than exposing this option, which I think people would generally prefer to have off by default.

@julienledem, any comment on whether this could be DEBUG instead of INFO?

nielsbasjes:
From my standpoint I see a few viable scenarios.
* Simply switch this message to DEBUG instead of INFO. This makes this logging only available to people who are using a custom built version of Parquet.
* Add a flag to enable/disable this specific logging (what I did here). This makes this logging optional and settable by the application developer.
* Make the logging level 'settable'. I do not see why the logging level was hardcoded in the way it is now.  Perhaps this allows the compiler to optimize the logging statements out completely?  The only relevant mention I could find on this is this post https://groups.google.com/forum/#!topic/parquet-dev/Eb0btS78M5g

proflin:
hi @nielsbasjes Thanks for contributing this. 

CI seems to be failing due to:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:testCompile (default-testCompile) on project parquet-hadoop: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/apache/parquet-mr/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java:[105,43] constructor ColumnChunkPageWriteStore in class org.apache.parquet.hadoop.ColumnChunkPageWriteStore cannot be applied to given types;
[ERROR] required: org.apache.parquet.hadoop.CodecFactory.BytesCompressor,org.apache.parquet.schema.MessageType,org.apache.parquet.bytes.ByteBufferAllocator,boolean
[ERROR] found: org.apache.parquet.hadoop.CodecFactory.BytesCompressor,org.apache.parquet.schema.MessageType,org.apache.parquet.bytes.HeapByteBufferAllocator
[ERROR] reason: actual and formal argument lists differ in length
[ERROR] /home/travis/build/apache/parquet-mr/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java:[164,39] constructor ColumnChunkPageWriteStore in class org.apache.parquet.hadoop.ColumnChunkPageWriteStore cannot be applied to given types;
[ERROR] required: org.apache.parquet.hadoop.CodecFactory.BytesCompressor,org.apache.parquet.schema.MessageType,org.apache.parquet.bytes.ByteBufferAllocator,boolean
[ERROR] found: org.apache.parquet.hadoop.CodecFactory.BytesCompressor,org.apache.parquet.schema.MessageType,org.apache.parquet.bytes.HeapByteBufferAllocator
[ERROR] reason: actual and formal argument lists differ in length

Details are here: https://travis-ci.org/apache/parquet-mr/jobs/101882142#L2900

Could you fix that accordingly, please? Cheers.

nielsbasjes:
@proflin Yes, i'll fix that. But I think that right now it may be better to wait and see if @rdblue and @julienledem think this is the right solution direction. If they agree it is then I'll fix everything.
Perhaps you can point me to a way this can be unit tested? I.e. How can I check the messages that are sent to the Logging framework in a unit test?

rdblue:
@nielsbasjes, we've recently changed to using SLF4J as the logging back-end, which should be released in 1.9.0. That means that as long as we haven't optimized the logging statements out of the compiled version, you should be able to control seeing that info. The statements get optimized out when the if statement is essentially `if (false) logger.info(...)`. But we can just get rid of that if statement and rely on the logging framework.

nielsbasjes:
For the 'complex' logging statements (like the one I ran into): If we replace the 
```if (false) logger.info(...)``` 
with 
```if (logger.isInfoEnabled()) logger.info(...)``` 
I expect the changes to be much simpler than my first attempt and that the performance impact is limited.
For the 'simple' statements I think we can rely on the built in optimizations in combination with the templating feature which is part of slf4j.
What do you think?

**Comment by nielsbasjes** [comment]
I found that in parquet-common there is a org.apache.parquet.Log class that explicitly states:
```
 * Simple wrapper around java.util.logging
 * Adds compile time log level.
 * The compiler removes completely if statements that reference to a false constant
 *
 * <code>
 *   if (DEBUG) LOG.debug("removed by the compiler if DEBUG is a false constant")
 * </code>
```	

To me this "Adds compile time log level" is an unwanted feature.
@julienledem Why was this 'compile time' idea implemented? Performance?

nielsbasjes:
Note: I still have to fully test this

rdblue:
@nielsbasjes, just checking in. If you have time to test this, feel free to ping me for a review.

julienledem:
@nielsbasjes do you want to follow up on this PR?

nielsbasjes:
Yes.
I'm diving into this one as we speak.
I'll keep you posted on my progress.

nielsbasjes:
I found that TestCircularReferences shows a StackOverflow for which I created a separate ticket at AVRO.
https://issues.apache.org/jira/browse/AVRO-1923

nielsbasjes:
Ok, that git action went totally wrong.
I'm creating a new pull request.

nielsbasjes:
New PR is #369

PR #313, opened 2016-01-17
PARQUET-431: Make ParquetOutputFormat.memoryManager volatile

proflin:
Currently ParquetOutputFormat.getRecordWriter() contains an unsynchronized lazy initialization of the non-volatile static field *memoryManager*.

Because the compiler or processor may reorder instructions, threads are not guaranteed to see a completely initialized object, when ParquetOutputFormat.getRecordWriter() is called by multiple threads.

This PR makes *memoryManager* volatile to correct the problem.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
@@ -394,9 +394,12 @@ public class ParquetOutputFormat<T> extends FileOutputFormat<Void, T> {
         MemoryManager.DEFAULT_MEMORY_POOL_RATIO);
     long minAllocation = conf.getLong(ParquetOutputFormat.MIN_MEMORY_ALLOCATION,
         MemoryManager.DEFAULT_MIN_MEMORY_ALLOCATION);
-    if (memoryManager == null) {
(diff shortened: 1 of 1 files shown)


proflin:
This is fairly a small change, @rdblue @liancheng would you please take a look at this?
Cheers.

liancheng:
+1

julienledem:
should we not have a synchronized block around the if (memoryManager == null) ... ?

proflin:
@julienledem Thanks for reviewing and pointing that out. I think we should have the synchronized block.

The latest commit adds this synchronized block for both _memoryManager creation snippet_ and _getMemoryManager() method_.

Would you take another look please? Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -445,9 +448,9 @@ public OutputCommitter getOutputCommitter(TaskAttemptContext context)
   /**
    * This memory manager is for all the real writers (InternalParquetRecordWriter) in one task.
    */
-  private static MemoryManager memoryManager;
+  private static volatile MemoryManager memoryManager;
```
julienledem: volatile is not needed anymore

julienledem:
+1

PR #315, opened 2016-01-27
PARQUET-385 PARQUET-379: Fixes strict schema merging

liancheng:
This PR fixes strict mode schema merging. To merge two `PrimitiveType` `t1` and `t2`, they must satisfy the following conditions:

1. `t1` and `t2` have the same primitive type name
1. `t1` and `t2` either
   - don't have original type, or
   - have the same original type
1. If `t1` and `t2` are both `FIXED_LEN_BYTE_ARRAY`, they should have the same length

Also, merged schema now preserves original name if there's any.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
@@ -525,14 +525,9 @@ public final class PrimitiveType extends Type {
     }
 
     if (strict) {
-      // Can't merge primitive fields of different type names
-      if (!primitive.equals(toMerge.asPrimitiveType().getPrimitiveTypeName())) {
-        reportSchemaMergeError(toMerge);
-      }
-
-      // Can't merge primitive fields of different original types
-      if (getOriginalType() == null && toMerge.getOriginalType() != null ||
-          getOriginalType() != null && !getOriginalType().equals(toMerge.getOriginalType())) {
+      // Can't merge primitive fields of different type names or different original types
+      if (!primitive.equals(toMerge.asPrimitiveType().getPrimitiveTypeName()) ||
+          getOriginalType() != toMerge.getOriginalType()) {
         reportSchemaMergeError(toMerge);
       }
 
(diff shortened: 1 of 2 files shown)


liancheng:
Not quite sure whether we should also deprecate the `strict` argument of `MessageType.union`. I tend to do so since Parquet itself never uses non-strict schema merging internally. Also, non-strict schema merging doesn't make sense and may lead to wrong query result as discussed in [PARQUET-379][1].

[1]: https://issues.apache.org/jira/browse/PARQUET-379?focusedCommentId=14933630&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14933630

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -514,16 +514,41 @@ protected Type union(Type toMerge) {
...
+    if (strict) {
+      // Can't merge primitive fields of different type names
+      if (!primitive.equals(toMerge.asPrimitiveType().getPrimitiveTypeName())) {
+        reportSchemaMergeError(toMerge);
+      }
+
+      // Can't merge primitive fields of different original types
+      if (getOriginalType() == null && toMerge.getOriginalType() != null ||
+          getOriginalType() != null && !getOriginalType().equals(toMerge.getOriginalType())) {
+        reportSchemaMergeError(toMerge);
```
proflin: Since originalType is an Enum, seems we can just do like:
```
if (getOriginalType() != toMerge.getOriginalType())
  reportSchemaMergeError(toMerge);
```
which covers:
1. if both are null, the condition evals to false
2. if one is null and the other is not null, the condition evals to true
3. if both are not null, then the condition evals to false if they take different values, and evals to true otherwise.

liancheng: Thanks! Being in the Scala world for so long a time, I really need to re-learn Java again :)

proflin:
+1, LGTM

julienledem:
+1

julienledem:
@liancheng can you remove the square brackets in your header? That trips up the merge script.

liancheng:
@julienledem Done.

PR #316, opened 2016-01-30
PARQUET-484: Warn when Decimal is stored as INT64 while could be stored as INT32

lw-lin:
Below is documented in [LogicalTypes.md](https://github.com/Parquet/parquet-format/blob/master/LogicalTypes.md#decimal):

> int32: for 1 <= precision <= 9
> int64: for 1 <= precision <= 18; precision < 10 will produce a warning

This PR implements the `precision < 10 will produce a warning` part.

@rdblue @liancheng would mind taking a look at this when you have time? It's a fairly small addition; cheers.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/Types.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/Types.java
@@ -25,6 +25,7 @@ import java.util.List;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
 import org.apache.parquet.schema.Type.ID;
+import org.slf4j.Logger;
 
 /**
  * This class provides fluent builders that produce Parquet schema Types.
@@ -305,6 +306,7 @@ public class Types {
   public abstract static class
       BasePrimitiveBuilder<P, THIS extends BasePrimitiveBuilder<P, THIS>>
       extends Builder<THIS, P> {
+    private final Logger log = org.slf4j.LoggerFactory.getLogger(BasePrimitiveBuilder.class);
     private static final long MAX_PRECISION_INT32 = maxPrecision(4);
     private static final long MAX_PRECISION_INT64 = maxPrecision(8);
     private final PrimitiveTypeName primitiveType;
@@ -406,6 +408,11 @@ public class Types {
                   meta.getPrecision() <= MAX_PRECISION_INT64,
                   "INT64 cannot store " + meta.getPrecision() + " digits " +
                   "(max " + MAX_PRECISION_INT64 + ")");
+              if (meta.getPrecision() <= MAX_PRECISION_INT32) {
+                String msg = String.format("Consider use INT32 instead of INT64 for storing %d digits; see %s for details.",
+                    precision, "https://github.com/Parquet/parquet-format/blob/master/LogicalTypes.md");
+                log.warn(msg);
+              }
             } else if (primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
               Preconditions.checkState(
                   meta.getPrecision() <= maxPrecision(length),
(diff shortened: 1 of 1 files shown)


liancheng:
Had once noticed this issue but decided not to have this warning because this warning is mostly for Parquet data model developers. With this patch, when an end user uses a Parquet data model that doesn't always use the optimal primitive type for decimal, he/she may always see this warning, but can do nothing about it without updating the data model itself. This makes this warning quite disturbing and not helpful. Just my two cents.

proflin:
@liancheng Thank your for the explanation, which is also reasonable I believe.
Is there some consensus of warning nor not reached so far in this community (I'm not aware of any)? I think it'd be good to make some changes either to the code or to the documentation, but this is not an urgent one after all.

rdblue:
I think this warning is valuable. In the case of Spark or Hive, it is for data model developers, but it would definitely help fix a problem introduced there by flagging that the underlying data isn't using an efficient representation. In the case of object models like Avro, the user has control over the underlying type and would benefit from knowing if they chose too wide of a type.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -305,6 +306,7 @@ public P named(String name) {
   public abstract static class
       BasePrimitiveBuilder<P, THIS extends BasePrimitiveBuilder<P, THIS>>
       extends Builder<THIS, P> {
+    private final Logger log = org.slf4j.LoggerFactory.getLogger(BasePrimitiveBuilder.class);
```
rdblue: This should be `static final` and I would normally add an import for `LoggerFactory`.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -406,6 +408,11 @@ protected PrimitiveType build(String name) {
                   meta.getPrecision() <= MAX_PRECISION_INT64,
                   "INT64 cannot store " + meta.getPrecision() + " digits " +
                   "(max " + MAX_PRECISION_INT64 + ")");
+              if (meta.getPrecision() <= MAX_PRECISION_INT32) {
+                String msg = String.format("Consider use INT32 instead of INT64 for storing %d digits; see %s for details.",
```
rdblue: SLF4J can format the message instead of using `String.format`. Minor, but it would help with consistency.

Also, I think the error message should be more direct: `Log.warn("Decimal with {} digits is stored in an INT64, but fits in an INT32. See {}.", precision, url)`

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -406,6 +408,11 @@ protected PrimitiveType build(String name) {
...
+                    precision, "https://github.com/Parquet/parquet-format/blob/master/LogicalTypes.md");
```
rdblue: This URL should reference the Apache project, https://github.com/apache/parquet-format/blob/master/LogicalTypes.md and should probably be a static final constant.

proflin: Thank you for the review & comments! Will address them shortly.

proflin:
Updated according to @rdblue 's comments.

proflin:
One thing worthy noticing is, we'll warn when `precision <= 9` rather than when `precision <= 10`. This has been brought up in [parquet-format PR #26](https://github.com/apache/parquet-format/pull/26).

rdblue:
@lw-lin I think the logic here should match what was decided for the format (that PR has been merged). Is it the same in this PR?

lw-lin:
Yes, it is the same here.

What was decided in the merged PR:
```scala
precision < 10 will produce a warning
```

What's implemented in this PR:
```scala
// MAX_PRECISION_INT32 is actually 9
if (meta.getPrecision() <= MAX_PRECISION_INT32) {
   warn...
}
```

@rdblue thanks for the revisit! :-)

lw-lin:
@rdblue @liancheng thank you all for the review & merging :-)

PR #319, opened 2016-02-02
PARQUET-401: Deprecate Log and move to SLF4J Logger

lw-lin:
The current Log class is intended to allow swapping out logger back-ends, but SLF4J already does this. It also doesn't expose as nice of an API as SLF4J, which can handle formatting to avoid the cost of building log messages that won't be used.

We should deprecate the org.apache.parquet.Log class and move to using SLF4J directly, instead of wrapping SLF4J (PARQUET-305).


liancheng:
@proflin The last build failure was probably caused by Travis network issue. You may push a minor commit to trigger Travis again.

@rdblue Is there any way to trigger Travis without pushing a commit?

julienledem:
@liancheng I think the owner of the PR can retrigger the build from the travis ui (need to be logged in)

julienledem:
This seems fine.
can we have a perf test run to make sure this has no impact on perf?
I would think not but i'd rather be sure since the ```if (DEBUG)``` was optimized out at compile time.

proflin:
To make things simple and easy to review, this PR simply:
1. replaces the orig _if (Log.DEBUG)_ condition with _if(LOGGER.isDebugEnabled())_
2. replaces the orig _Log.debug(msg)_ with _LOGGER.debug(msg)_
(same for _WARN, INFO, ERROR_)

I've written some replacing rules and used them replacing most of things automatically; and I've manually checked that the replacing worked well for each file. So things should work well.

However, this PR does NOT move to use the parametered form of LOGGER.debug(), i.e., does NOT replace _Log.debug("msg is " + msg)_ with _LOGGER.debug("msg is {}", msg)_, since I found it's error-prone to do so with so many files at a time. We'd better do that in a separate PR if it's necessary.

proflin:
@julienledem 
Regarding perf, I digged a little and found [PerfTest](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/test/java/org/apache/parquet/io/PerfTest.java), seems it serves well for our purpose of testing slf4j perf impacts. So I've run it for several times, results are:

(1) if we use Log.java, writing 1,000,000 recs took 1,455 ms
(2) if we use slf4j-simple, writing 1,000,000 recs took 1,675 ms, which is 15% more than (1)
(3) if we use slf4j-log4j, writing 1,000,000 recs took 1,885 ms, which is 30% more than (1)

JProfiler shows that it's _Logger.isDebugEnabled()_ and _isLevelEnabled()_ who introduce the extra latency; and the classes in the blue area below, i.e. **_MessageColumnIO_**, **_ColumnWriterV1_**, and **_ColumnWriterV2_** , are victims of _Logger.isDebugEnabled() and isLevelEnabled()_.
jprofiler 2

So I cleaned up _Logger.isDebugEnabled()_ and _isLevelEnabled()_ a little manually in **_MessageColumnIO_**, **_ColumnWriterV1_**, and **_ColumnWriterV2_**(please see [https://github.com/proflin/parquet-mr/commit/005bc5cfe90a4c95b15a4ce50d13a99175be4056]), and then reran the perf test; results are:

(2+) if we use slf4j-simple, writing 1,000,000 recs took 1,437 ms, which is roughly the same as (1)
(3+) if we use slf4j-log4j, writing 1,000,000 recs took 1,444 ms, which is roughly the same as (1)

So moving to slf4j doesn't seem to bring noticeable perf impact, right?

p.s. detaied _**PerfTest**_ performance results are here: https://docs.google.com/spreadsheets/d/1FLwD71WFmkEfqDWyo2pe1vfkk7BI6bJKMOnAzGCl5XI/edit#gid=0

liancheng:
High level changes and the perf test data LGTM, haven't looked into all the details though.

rdblue:
@proflin, thanks for doing this and providing the perf numbers. It looks like using the static final constants allows the JVM to optimize out the logging at runtime. It's not something we want to do everywhere, but it makes the switch to SLF4J viable. Nice work!

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
```diff
@@ -523,7 +524,9 @@ private void readRepetitionAndDefinitionLevels() {
   private void checkRead() {
     if (isPageFullyConsumed()) {
       if (isFullyConsumed()) {
-        if (DEBUG) LOG.debug("end reached");
+        if (LOGGER.isDebugEnabled()) {
```
rdblue: The call to debug will internally check the log level, right? I thought that it is only necessary to add a conditional on `isDebugEnabled` if the operation is expensive or if you're avoiding logging in a tight loop like you do in the writers. For writing a static debug string, is this needed?

parquet-encoding/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -261,10 +270,16 @@ private SequenceBytesIn(List<BytesInput> inputs) {
     @Override
     public void writeAllTo(OutputStream out) throws IOException {
       for (BytesInput input : inputs) {
-        if (DEBUG) LOG.debug("write " + input.size() + " bytes to out");
-        if (DEBUG && input instanceof SequenceBytesIn) LOG.debug("{");
+        if (LOGGER.isDebugEnabled()) {
+          LOGGER.debug("write " + input.size() + " bytes to out");
+        }
+        if (LOGGER.isDebugEnabled() && input instanceof SequenceBytesIn) {
```
rdblue: Can this be nested in the last if statement instead of calling `isDebugEnabled` twice?

parquet-encoding/src/test/java/org/apache/parquet/column/values/bitpacking/TestByteBitPacking.java
```diff
@@ -59,16 +71,24 @@ private void packUnpack(BytePacker packer, int[] values, int[] unpacked) {
...
     return values;
   }
 
   @Test
   public void testPackUnPackAgainstHandWritten() throws IOException {
-    LOG.debug("");
-    LOG.debug("testPackUnPackAgainstHandWritten");
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("");
+    }
```
rdblue: Merge here, too.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -134,8 +135,12 @@ private void checkRead() throws IOException {
       long timeSpentReading = System.currentTimeMillis() - t0;
       totalTimeSpentReadingBytes += timeSpentReading;
       BenchmarkCounter.incrementTime(timeSpentReading);
-      if (Log.INFO) LOG.info("block read in memory in " + timeSpentReading + " ms. row count = " + pages.getRowCount());
-      if (Log.DEBUG) LOG.debug("initializing Record assembly with requested schema " + requestedSchema);
+      if (LOGGER.isInfoEnabled()) {
+        LOGGER.info("block read in memory in " + timeSpentReading + " ms. row count = " + pages.getRowCount());
+      }
+      if (LOGGER.isDebugEnabled()) {
```
rdblue: This can also be nested. No need to check debug if info isn't enabled.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -374,16 +379,36 @@ private static int getMaxPaddingSize(Configuration conf) {
...
-    if (INFO) LOG.info("Parquet page size to " + props.getPageSizeThreshold());
-    if (INFO) LOG.info("Parquet dictionary page size to " + props.getDictionaryPageSizeThreshold());
-    if (INFO) LOG.info("Dictionary is " + (props.isEnableDictionary() ? "on" : "off"));
-    if (INFO) LOG.info("Validation is " + (validating ? "on" : "off"));
-    if (INFO) LOG.info("Writer version is: " + props.getWriterVersion());
-    if (INFO) LOG.info("Maximum row group padding size is " + maxPaddingSize + " bytes");
-    if (INFO) LOG.info("Page size checking is: " + (props.estimateNextSizeCheck() ? "estimated" : "constant"));
-    if (INFO) LOG.info("Min row count for page size check is: " + props.getMinRowCountForPageSizeCheck());
-    if (INFO) LOG.info("Min row count for page size check is: " + props.getMaxRowCountForPageSizeCheck());
+    if (LOGGER.isInfoEnabled()) {
```
rdblue: All of these should be merged.

parquet-common/src/test/java/org/apache/parquet/TestLog.java
```diff
@@ -21,11 +21,16 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class TestLog {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(TestLog.class);
+
   @Test
   public void test() {
```
rdblue: I don't think this class is needed any more. It just validates that we don't release code with Log.DEBUG set to true. That no longer matters.

rdblue:
This looks almost ready to go. There are a lot of minor problems that appear caused by automatically changing the `if (DEBUG)` lines to `if (LOGGER.isDebugEnabled())`. Many can be merged together, some aren't required, etc. Could you go through and fix those cases? Once that's done I think this is good to go.

proflin:
hi @julienledem @rdblue @liancheng, I think you might want to know [how to trigger a Travis-CI rebuild without pushing a commit](http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit).

So I'll have to close&reopen a PR lol

proflin:
Sure I'll go through and fix those and other cases; will update the PR once it's done.

rdblue:
Thanks for the link @proflin. I think the main problem is that we don't have write access to the Apache mirror project. The idea is to make sure commits come from the Apache infrastructure and are just mirrored to github, but the result is that we can't restart travis CI builds or even close and reopen PRs.

julienledem:
@proflin as the owner of the pr you should be able to retrigger the build right?

proflin:
@julienledem I'm afraid the owner of the PR can't see the [↻ Restart Build] button, even after he/she logs in onto the WebUI. I've verified that close & reopen the PR does re-trigger the CI, though.

proflin:
@rdblue I've spent the last two or three days investigating why CI fails on this PR so far, and what I found is that, after [[PARQUET-305: Update logging to SLF4J][PR#290]](https://github.com/apache/parquet-mr/pull/290), the effective `slf4j` log level in the CI **_hadoop2_** enviroment is actually `DEBUG` rather than `INFO`. This is because our dependency `hadoop-client-2.3.0` introduces `log4j` so `slf4j` is then bound to `log4j` with default log level `DEBUG`.

Should we fix it? 

Good news is, the unintentional `DEBUG` level did expose two bugs(parquet-528,-529) which caused this CI to fail. We hadn't caught this in a non-`DEBUG` environment before :-)

julienledem:
Yes we should fix. It sounds like we should not pull any slf4j binding in the compile or run scope.
We should pull it in the test scope, probably in info through the appropriate log conf file in src/test/resources.
Thoughts?

rdblue:
@proflin, I agree with Julien that we should fix it and that there should be no SLF4J binding leaked through Parquet's dependencies. I think the fix is to exclude the slf4j-log4j binding dependency. Thanks for working on this!

lw-lin:
hi @rdblue, finally this is ready for a another (hopefully the last) round of review.

Changes since the first round:
- many `if (LOGGER.isXXXEnabled())` have been merged together;
- some unnecessary `if (LOGGER.isXXXEnabled())` have been removed;
- the static final constants optimization has been applied to some more classes when necessary.

Also, things that are intentionally left out(let's fix them in following-up PRs):
- moving to use the parametered form of `LOGGER.xxx()`, i.e., replacing `Log.debug("msg is " + msg + ", number is " + number)` with `LOGGER.debug("msg is {}, number is {}", msg, number)`;
- fixing the slf4j log level in the CI hadoop2 enviroment, i.e. slf4J binding leaked through Parquet's dependencies issue.

So, could you take another look at this? Thanks! :-)

lw-lin:
two weeks' ping to some committer @rdblue :-)

rdblue:
@lw-lin, thanks for working on this.

Looks like my previous comment about fixing the `if (LOG.isDebugEnabled())` statements was misinterpreted. Sorry about that, I probably wasn't very clear.

You've replaced a lot of those calls with a static final constant at the top of the file. That's a bad thing because it means we can't turn on debug logging once the code is loaded. We should only do that when it matters for performance, which you said above was in `MessageColumnIO`, `ColumnWriterV1`, and `ColumnWriterV2`. The rest of the constants should be removed, and there shouldn't be unused constants like `ERROR_ENABLED`.

The rest of the calls should not use `if (LOG.isDebugEnabled())` at all. Just call `LOG.debug(...)` and let the framework decide. The benefit of checking the log level is to avoid expensive operations to prepare a log message that is discarded at the current log level. Using the SLF4J formatting calls mostly avoids that.

The only time we should use `LOG.isDebugEnabled()` is when there's an expensive call that can't be handled inside the logger. An example is debugging the total memory used: we would inspect all of the write framework before calling into the logger, so guarding that with a level check is a good idea. But when you can simply pass arguments into the logger and the work is done there by calling `toString` on what you pass in, there's no need for the check.

I think getting the debug logging right is going to require going through the code and making sure each call makes sense, rather than transforming certain patterns with an IDE.

parquet-column/src/main/java/org/apache/parquet/io/RecordConsumerLoggingWrapper.java
```diff
@@ -67,16 +74,20 @@ private String indent() {
...
      * {@inheritDoc}
      */
     @Override
     public void startGroup() {
-      if (DEBUG) ++indent;
-      if (DEBUG) log("<!-- start group -->");
+      if (DEBUG_ENABLED) {
+        ++indent;
+      }
+      if (DEBUG_ENABLED) {
```
rdblue: @lw-lin, it looks like there are still some if statements that can be combined. Please make sure you check through the code before the next round of review for these.

parquet-column/src/main/java/org/apache/parquet/io/RecordConsumerLoggingWrapper.java
```diff
@@ -67,16 +74,20 @@ private String indent() {
     }
 
     private void log(Object value) {
-      logger.debug(indent() + value);
+      LOGGER.debug(indent() + value);
```
rdblue: This should be a formatting call: `LOGGER.debug("{}{}", indent(), value)`

parquet-column/src/main/java/org/apache/parquet/io/RecordReaderImplementation.java
```diff
@@ -376,7 +377,7 @@ private RecordConsumer validator(RecordConsumer recordConsumer, boolean validati
   }
 
   private RecordConsumer wrap(RecordConsumer recordConsumer) {
-    if (Log.DEBUG) {
+    if (LOGGER.isDebugEnabled()) {
```
rdblue: This call to `isDebugEnabled` is fine.

lw-lin:
@rdblue @julienledem sorry for the late response -- oh I somehow missed @rdblue 's kind comments.

I'll update this within this week. Thanks @rdblue @julienledem !

julienledem:
thank you @lw-lin !

lw-lin:
This has been updated a lot -- @julienledem @rdblue would you take a look at your convenience?

lw-lin:
Plus, five rounds of `PerfTest` show that, comparing to the `Log.java` approach, the `slf4j` approach:
- on average takes 0.87% longer in writing 1,000,000 records
- on average takes 1.31% longer in reading 1,000,000 records

For detailed results please refer to https://docs.google.com/spreadsheets/d/1FLwD71WFmkEfqDWyo2pe1vfkk7BI6bJKMOnAzGCl5XI/edit#gid=1865972057)

parquet-common/src/test/java/org/apache/parquet/TestLog.java
```diff
@@ -1,31 +0,0 @@
...
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-package org.apache.parquet;
-
-import org.junit.Assert;
-import org.junit.Test;
-
-public class TestLog {
```
lw-lin: this class is not necessary any more

julienledem:
Thank you @lw-lin for this work. Sorry this particular PR didn't go through.
Those patches that touch a lot of files are painful to maintain.
PR #369 implemented a similar changes and now has been merged.
Thank your for your contribution @lw-lin !

lw-lin:
Thank you @julienledem . I like that PR too. Thank you also @rdblue @liancheng for the efforts you'd put into this!

PR #320, opened 2016-02-04
PARQUET-509: Fix args passed to string format calls

nezihyigitbasi:
This PR fixes the args passed to the `String.format()` call.


jaltekruse:
Thanks for fixing this Nezih, will be more careful in the future.
I'm not a committer, but +1

julienledem:
+1

PR #321, opened 2016-02-05
PARQUET-511: Integer overflow when counting values in column.

goreckm:
This commit fixes an issue when the number of entries in a column page is larger than the size of an integer. No exception is thrown directly, but the def level is set incorrectly, leading to a null value being returned during read.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
@@ -353,7 +353,7 @@ public class ColumnReaderImpl implements ColumnReader {
     }
     this.totalValueCount = pageReader.getTotalValueCount();
     if (totalValueCount <= 0) {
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderImpl.java
```diff
@@ -352,7 +352,7 @@ public ColumnReaderImpl(ColumnDescriptor path, PageReader pageReader, PrimitiveC
       this.dictionary = null;
     }
     this.totalValueCount = pageReader.getTotalValueCount();
-    if (totalValueCount == 0) {
+    if (totalValueCount <= 0) {
       throw new ParquetDecodingException("totalValueCount == 0");
```
mickaellcr: I would also change the error message here to provide the value of totalValueCount (and change the equality)

julienledem:
looks good to me

PR #325, opened 2016-02-14
PARQUET-528: Fix flush() for RecordConsumer and implementations

proflin:
`flush()` was added in `RecordConsumer` and `MessageColumnIO` to help implementing nulls caching.

However, other `RecordConsumer` implementations should also implements `flush()` properly. For instance, `RecordConsumerLoggingWrapper` and `ValidatingRecordConsumer` should call `delegate.flush()` in their `flush()` methods, otherwise data might be mistakenly truncated.

This PR:
- makes `flush()` abstract in `RecordConsumer`
- implements `flush()` properly for all `RecordConsumer` subclasses, specifically:
 - `RecordConsumerLoggingWrapper`
 - `ValidatingRecordConsumer`
 - `ConverterConsumer `
 - `ExpectationValidatingRecordConsumer `


proflin:
@julienledem @rdblue @liancheng would you mind taking a look at this when you have time? This has been blocking [[Parquet-401: Deprecate Log and move to SLF4J Logger][PR#319]](https://github.com/apache/parquet-mr/pull/319). Thanks!

parquet-column/src/main/java/org/apache/parquet/io/api/RecordConsumer.java
```diff
@@ -126,10 +126,8 @@
   abstract public void addDouble(double value);
 
   /**
-   * NoOps by default
-   * Subclass class can implement its own flushing logic
+   * subclass class should implement its own flushing logic
    */
-  public void flush() {
-  }
+  abstract public void flush();
```
julienledem: I think for backward compatibility this needs to not be abstract.

proflin: @julienledem 
Sure. I've issued another commit reverting it back to be non-_abstract_. Is there any need we make it _abstract_ in a major release, like 2.0?

proflin:
hi @julienledem, sorry for keeping pinging on this, but would you take another look when you have time?
Thanks!

rdblue:
+1, looks fine to me.

@julienledem may want to comment, but otherwise I'll merge this in a day or so.

julienledem:
+1

liancheng:
Thanks, merged to master.

PR #326, opened 2016-02-14
PARQUET-529: Avoid evoking job.toString() in ParquetLoader

proflin:
When ran under hadoop2 environment and log level setting to `DEBUG`, ParquetLoader would evoke `job.toString()` in several methods, which might cause the whole application to stop due to :

```
java.lang.IllegalStateException: Job in state DEFINE instead of RUNNING

	at org.apache.hadoop.mapreduce.Job.ensureState(Job.java:283)
	at org.apache.hadoop.mapreduce.Job.toString(Job.java:452)
	at java.lang.String.valueOf(String.java:2847)
	at java.lang.StringBuilder.append(StringBuilder.java:128)
	at org.apache.parquet.pig.ParquetLoader.getSchema(ParquetLoader.java:260)
	at org.apache.parquet.pig.TestParquetLoader.testSchema(TestParquetLoader.java:54)
    ...
```

The reason is that in the hadoop 2.x branch, `org.apache.hadoop.mapreduce.Job.toString()` has added an `ensureState(JobState.RUNNING)` check; see [map-reduce: Job.java#452](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.hadoop/hadoop-mapreduce-client-core/2.3.0/org/apache/hadoop/mapreduce/Job.java#452). In contrast, the hadoop 1.x branch does not contain such checks, so `ParquetLoader` works well.

This PR simply avoids evoking `job.toString()` in `ParquetLoader`.


proflin:
@julienledem @rdblue @liancheng @danielcweeks @aniket486 would you mind taking a look at this when you have time? This has been blocking [[Parquet-401: Deprecate Log and move to SLF4J Logger][PR#319]](https://github.com/apache/parquet-mr/pull/319). Thanks!

parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -137,7 +137,7 @@ public ParquetLoader(Schema requestedSchema, boolean columnIndexAccess) {
 
   @Override
   public void setLocation(String location, Job job) throws IOException {
-    if (DEBUG) LOG.debug("LoadFunc.setLocation(" + location + ", " + job + ")");
```
julienledem: job.getId or something?

rdblue:
+1, thanks @proflin!

PR #330, opened 2016-02-25
PARQUET-384: Add dictionary filtering.

rdblue:
This builds on #286 from @danielcweeks and cleans up some of the interfaces. It introduces `DictionaryPageReadStore` to expose dictionary pages to the filters and cleans up some internal calls by passing `ParquetFileReader`.

When committed, this closes #286.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
@@ -1,0 +1,281 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.filter2.dictionarylevel;
+
+import org.apache.parquet.Log;
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.Dictionary;
+import org.apache.parquet.column.page.DictionaryPage;
+import org.apache.parquet.column.page.DictionaryPageReadStore;
+import org.apache.parquet.filter2.predicate.FilterPredicate;
+import org.apache.parquet.filter2.predicate.Operators.*;
+import org.apache.parquet.filter2.predicate.UserDefinedPredicate;
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
+
+import java.io.IOException;
+import java.util.*;
+
+import static org.apache.parquet.Preconditions.checkArgument;
+import static org.apache.parquet.Preconditions.checkNotNull;
+
+
+/**
+ * Applies filters based on the contents of column dictionaries.
+ */
+public class DictionaryFilter implements FilterPredicate.Visitor<Boolean> {
+
+  private static final Log LOG = Log.getLog(DictionaryFilter.class);
+
+  public static boolean canDrop(FilterPredicate pred, List<ColumnChunkMetaData> columns, DictionaryPageReadStore dictionaries) {
+    checkNotNull(pred, "pred");
+    checkNotNull(columns, "columns");
+    return pred.accept(new DictionaryFilter(columns, dictionaries));
+  }
+
+  private final Map<ColumnPath, ColumnChunkMetaData> columns = new HashMap<ColumnPath, ColumnChunkMetaData>();
+  private final DictionaryPageReadStore dictionaries;
+
+  private DictionaryFilter(List<ColumnChunkMetaData> columnsList, DictionaryPageReadStore dictionaries) {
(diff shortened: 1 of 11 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -0,0 +1,356 @@
...
+    filterColumn.getColumnPath();
+
+    try {
+      Set<T> dictSet = expandDictionary(meta);
+      if (dictSet == null) {
+        return BLOCK_MIGHT_MATCH;
+      }
+
+      for(T entry : dictSet) {
+        if(value.compareTo(entry) > 0) {
```
danielcweeks: Shouldn't this be `>=`?

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -0,0 +1,356 @@
...
+    filterColumn.getColumnPath();
+
+    try {
+      Set<T> dictSet = expandDictionary(meta);
+      if (dictSet == null) {
+        return BLOCK_MIGHT_MATCH;
+      }
+
+      for(T entry : dictSet) {
+        if(value.compareTo(entry) >= 0) {
```
danielcweeks: Same here `>`?

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -0,0 +1,356 @@
...
+    filterColumn.getColumnPath();
+
+    try {
+      Set<T> dictSet = expandDictionary(meta);
+      if (dictSet == null) {
+        return BLOCK_MIGHT_MATCH;
+      }
+
+      for(T entry : dictSet) {
+        if(value.compareTo(entry) < 0) {
```
danielcweeks: `<=`

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -0,0 +1,356 @@
...
+    filterColumn.getColumnPath();
+
+    try {
+      Set<T> dictSet = expandDictionary(meta);
+      if (dictSet == null) {
+        return BLOCK_MIGHT_MATCH;
+      }
+
+      for(T entry : dictSet) {
+        if(value.compareTo(entry) <= 0) {
```
danielcweeks: `<`

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
rdblue: No, I think this is correct. The logic changes to > because the order is reversed, not because the logic is negated.

If `V` is the bound and we find any value, `x`, in the dictionary such that `x < V`, then there may be a matching row. But we call `V.compareTo(x)` and the order is reversed. We could call `x.compareTo(V)`, but it seems like calling the method on `V` is more likely to result in something the JVM can optimize.

The tests also validate that this behavior is correct by testing the boundary conditions with the smallest and largest values.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -163,46 +156,26 @@ public float getProgress() throws IOException, InterruptedException {
...
-    reader = new ParquetFileReader(configuration, parquetFileMetadata, file, blocks, columns);
-    for (BlockMetaData block : blocks) {
-      total += block.getRowCount();
-    }
+    this.total = reader.getRecordCount();
     this.unmaterializableRecordCounter = new UnmaterializableRecordCounter(configuration, total);
     LOG.info("RecordReader initialized will read a total of " + total + " records.");
   }
 
-  private boolean contains(GroupType group, String[] path, int index) {
```
rdblue: This method wasn't used.

rdblue:
I just added caching and a metadata check to the dictionary reader so that dictionary pages are only read once even if used in multiple predicates (which each call `readDictionary`).

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
danielcweeks: Yeah, you're right.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -547,19 +675,113 @@ public PageReadStore readNextRowGroup() throws IOException {
...
       for (Chunk chunk : chunks) {
-        columnChunkPageReadStore.addColumn(chunk.descriptor.col, chunk.readAllPages());
+        currentRowGroup.addColumn(chunk.descriptor.col, chunk.readAllPages());
       }
     }
+
+    // avoid re-reading bytes the dictionary reader is used after this call
+    if (nextDictionaryReader != null) {
+      nextDictionaryReader.setRowGroup(currentRowGroup);
+    }
```
danielcweeks: `advanceToNextBlock` resets nextDictionaryReader.  Does that mean the dictionary reader is not available after the row group read?

rdblue: `advanceToNextBlock` resets the next reader because the next block becomes the current block. This happens inside skip or read. The expected use is to get the dictionary reader for the next block, determine whether to skip or read it, do that operation, and then get the reader for the next block. Skipping or reading should cause the file reader to prepare to return the dictionary reader for the next, which is what is happening here.

rdblue:
#332 is a follow-up to this and adds EncodingStats so to avoid the brittle logic in `hasNonDictionaryEncodedPages`.

danielcweeks:
+1  The ParquetFileReader has some general issues with how it should be used, but I think this works really well for the dictionary filter and cleaning up some of the block filtering.

PR #331, opened 2016-02-25
PARQUET-397: Implement Pig predicate pushdown

rdblue:
This is based on #296 from @danielcweeks and implements a few remaining review items.

Closes #296.


parquet-pig/src/main/java/org/apache/parquet/pig/ParquetLoader.java
```diff
@@ -380,4 +405,163 @@ private Schema getSchemaFromRequiredFieldList(Schema schema, List<RequiredField>
...
+          return and(buildFilter(lhs), buildFilter(rhs));
+        case OP_OR:
+          return or(buildFilter(lhs), buildFilter(rhs));
+        case OP_BETWEEN:
+          BetweenExpression between = (BetweenExpression) rhs;
+          return and(
+              buildFilter(OpType.OP_GE, (Column) lhs, (Const) between.getLower()),
+              buildFilter(OpType.OP_LE, (Column) lhs, (Const) between.getUpper()));
+        case OP_IN:
+          FilterPredicate current = null;
```
danielcweeks: `BETWEEN` and `IN` aren't currently supported in Pig and there's still some question on the bounds (inclusive vs. exclusive) so we might want to comment these out for now and put a TODO.

danielcweeks:
Minor comment on `BETWEEN`.  It looks similar to the ORC implementation for ppd, so I'm ambivalent with respect to removing it.

+1

PR #332, opened 2016-02-26
PARQUET-548: Add EncodingStats.

rdblue:
This adds `EncodingStats`, which tracks the number of pages for each encoding, separated into dictionary and data pages. It also adds convenience functions that are useful for dictionary filtering, like `hasDictionaryEncodedPages` and `hasNonDictionaryEncodedPages`.

`EncodingStats` have a unit test in parquet-column and an integration test in parquet-hadoop that writes a file and verifies the stats are present and correct when it is read.

This includes commits from #330 because it updates the dictionary filter. I'll rebase and remove them once it is merged.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/EncodingStats.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/EncodingStats.java
@@ -29,6 +29,12 @@ import java.util.Set;
 import static org.apache.parquet.column.Encoding.PLAIN_DICTIONARY;
 import static org.apache.parquet.column.Encoding.RLE_DICTIONARY;
 
+/**
+ * EncodingStats track dictionary and data page encodings for a single column within a row group.
+ * These are used when filtering row groups. For example, to filter a row group based on a column's
+ * dictionary, all of the data pages in that column must be dictionary-encoded. This class provides
+ * convenience methods for those checks, like {@link #hasNonDictionaryEncodedPages()}.
+ */
 public class EncodingStats {
   final Map<Encoding, Integer> dictStats;
   final Map<Encoding, Integer> dataStats;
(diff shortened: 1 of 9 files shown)


rdblue:
@danielcweeks and @julienledem, this would be a good one to get into 1.9.0 if we can.

parquet-column/src/main/java/org/apache/parquet/column/EncodingStats.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static org.apache.parquet.column.Encoding.PLAIN_DICTIONARY;
+import static org.apache.parquet.column.Encoding.RLE_DICTIONARY;
+
+public class EncodingStats {
```
julienledem: please add a short description in javadoc

parquet-column/src/main/java/org/apache/parquet/column/EncodingStats.java
```diff
@@ -0,0 +1,161 @@
...
+  final Map<Encoding, Integer> dictStats;
+  final Map<Encoding, Integer> dataStats;
```
julienledem: private?

parquet-column/src/main/java/org/apache/parquet/column/EncodingStats.java
```diff
@@ -0,0 +1,161 @@
...
+      return addDictEncoding(encoding, 1);
+    }
+
+    public Builder addDictEncoding(Encoding encoding, int numPages) {
+      Integer pages = dictStats.get(encoding);
+      if (pages != null) {
+        dictStats.put(encoding, numPages + pages);
+      } else {
+        dictStats.put(encoding, numPages);
+      }
```
julienledem: How about:
```dictStats.put(encoding, numPages ==  null ? 0 : numPages);```

julienledem:
+1 LGTM

PR #334, opened 2016-03-08
PARQUET-560: Synchronize writes to the finishCalled variable

nezihyigitbasi:
Reads of the `finishCalled` variable are properly synchronized, but writes are not -- so there's some sort of inconsistent synch. going on here. This PR fixes that.

/cc @rdblue can you please take a look?

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/SnappyCompressor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/SnappyCompressor.java
@@ -113,7 +113,7 @@ public class SnappyCompressor implements Compressor {
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


rdblue:
@nezihyigitbasi, looks like the compatibility check is catching this change. It's not a binary incompatible change according to the [Oracle docs](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.20), so it should be safe to add a temporary exclusion for this in the root pom. Could you add that and then I'll merge? Thanks!

julienledem:
I guess another option is to make it a synchronized(this) block inside the method

nezihyigitbasi:
thanks @rdblue, yeah this is a safe change so I excluded it.

PR #336, opened 2016-03-10
PARQUET-584 show proper command usage when there's no arguments

coughman:


parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
```diff
@@ -150,11 +150,12 @@ public static void die(String message, boolean usage, String name, Command comma
     }
 
     if (usage) {
-      if (name != null && command != null) {
-        showUsage(name, command);
-      } else {
-        showUsage(name, command);
-      }
+    	if (name == null && command == null) {
+    		showUsage();
```
rdblue: @coughman, I'd like to merge this. Could you remove the formatting changes?

rdblue:
One small fix is needed. Also, could you open an issue in JIRA and add that to the title? We can't merge unless the title starts with "PARQUET-N: "

coughman:
created PARQUET-584

PR #337, opened 2016-03-23
PARQUET-569: Separate metadata filtering for ranges and offsets.

rdblue:
Range filtering should use the row group midpoint and offset filtering
should use the start offset.


julienledem:
do we need to add tests for this? I presume the previous behavior was wrong?

rdblue:
Yes, the previous behavior was wrong. I'll add a test.

rdblue:
@julienledem, I added a test so this is good to go. We've also verified that it fixes the bug in Hive.

julienledem:
+1

PR #338, opened 2016-03-24
PARQUET-571: Fix potential leak in ParquetFileReader.close()

nezihyigitbasi:
If an exception occurs when closing the input stream `f`, the codecs
will not be released. This may cause native memory leaks for some codecs. \cc @rdblue


danielcweeks:
+1

PR #340, opened 2016-04-15
PARQUET-581: Fix two instances of the conflation of the min and max row

mallman:
count for page size check in ParquetOutputFormat.java


rdblue:
+1 (assuming tests pass, I'll merge)

PR #341, opened 2016-04-18
PARQUET-585: Slowly ramp up sizes of int[]s in IntList to keep sizes small when data sets are small

piyushnarang:
One of the follow up items from PR - https://github.com/apache/parquet-mr/pull/339 was to slowly ramp up the size of the int[] created in IntList to ensure we don't allocate 64K arrays right off the bat. This PR updates the code to start with a 4K array then keeps doubling till 64K (and stays at 64K after that).

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
@@ -31,8 +31,8 @@ import java.util.List;
  */
 public class IntList {
 
-  private static final int MAX_SLAB_SIZE = 64 * 1024;
-  private static final int INITIAL_SLAB_SIZE = 4 * 1024;
+  public static final int MAX_SLAB_SIZE = 64 * 1024;
+  public static final int INITIAL_SLAB_SIZE = 4 * 1024;
 
   //Double slab size till we reach the max slab size. At that point we just add slabs of size
   //MAX_SLAB_SIZE. This ensures we don't allocate very large slabs from the start if we don't have
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
@@ -58,7 +58,7 @@ public class IntList {
     }
 
     /**
-     * @return wether there is a next value
+     * @return whether there is a next value
      */
     public boolean hasNext() {
       return current < count;
@@ -80,11 +80,11 @@ public class IntList {
   private int currentSlabPos;
 
   /**
-   * construct an empty list
+   * Construct an empty list
+   * Lazy initialize currentSlab only when needed to save on memory in cases where items might
+   * not be added
    */
-  public IntList() {
-    initSlab();
-  }
+  public IntList() {}
 
   private void initSlab() {
     currentSlab = new int[SLAB_SIZE];
@@ -95,10 +95,13 @@ public class IntList {
    * @param i value to append to the end of the list
    */
   public void add(int i) {
-    if (currentSlabPos == currentSlab.length) {
+    if (currentSlab == null) {
+      initSlab();
+    } else if (currentSlabPos == currentSlab.length) {
       slabs.add(currentSlab);
       initSlab();
     }
+
     currentSlab[currentSlabPos] = i;
     ++ currentSlabPos;
   }
@@ -108,6 +111,10 @@ public class IntList {
    * @return an IntIterator on the content
    */
   public IntIterator iterator() {
+    if (currentSlab == null) {
+      initSlab();
+    }
+
     int[][] itSlabs = slabs.toArray(new int[slabs.size() + 1][]);
     itSlabs[slabs.size()] = currentSlab;
     return new IntIterator(itSlabs, SLAB_SIZE * slabs.size() + currentSlabPos);
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
@@ -76,15 +76,11 @@ public class IntList {
   }
 
   private List<int[]> slabs = new ArrayList<int[]>();
+
+  // Lazy initialize currentSlab only when needed to save on memory in cases where items might
+  // not be added
   private int[] currentSlab;
   private int currentSlabPos;
-
-  /**
-   * Construct an empty list
-   * Lazy initialize currentSlab only when needed to save on memory in cases where items might
-   * not be added
-   */
-  public IntList() {}
 
   private void initSlab() {
     currentSlab = new int[SLAB_SIZE];
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
@@ -31,7 +31,13 @@ import java.util.List;
  */
 public class IntList {
 
-  private static final int SLAB_SIZE = 64 * 1024;
+  private static final int MAX_SLAB_SIZE = 64 * 1024;
+  private static final int INITIAL_SLAB_SIZE = 4 * 1024;
+
+  //Double slab size till we reach the max slab size. At that point we just add slabs of size
+  //MAX_SLAB_SIZE. This ensures we don't allocate very large slabs from the start if we don't have
+  //too much data.
+  private int currentSlabSize = INITIAL_SLAB_SIZE / 2;
 
   /**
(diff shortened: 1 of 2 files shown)


piyushnarang:
@rdblue / @isnotinvain - please take a look.

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/IntListTest.java
```diff
@@ -0,0 +1,81 @@
...
+import junit.framework.Assert;
+
+public class IntListTest {
+
+  /**
+   * Test IntList of fairly small size (< 4K), this tests a single slab being created
+   */
+  @Test
+  public void testSmallList() {
+    int testSize = 100;
```
isnotinvain: maybe make these values relative to MAX_SLAB_SIZE so that if it's ever changed these tests remain valid

isnotinvain:
just one comment but overall LGTM, thanks!

If you grep for 'slab' in the codebase you will probably find other, similar custom lists. It might make sense at some point to unify these since they each have their own little quirks like this.

(there's at least one more where the max is set to some fraction of the page size)

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/IntListTest.java
piyushnarang: sounds good, done

isnotinvain:
+1

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -31,7 +31,13 @@
  */
 public class IntList {
 
-  private static final int SLAB_SIZE = 64 * 1024;
+  public static final int MAX_SLAB_SIZE = 64 * 1024;
+  public static final int INITIAL_SLAB_SIZE = 4 * 1024;
+
+  //Double slab size till we reach the max slab size. At that point we just add slabs of size
```
rdblue: Nit: comments should have a space after "//"

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -31,7 +31,13 @@
...
+  //MAX_SLAB_SIZE. This ensures we don't allocate very large slabs from the start if we don't have
+  //too much data.
+  private int currentSlabSize = INITIAL_SLAB_SIZE / 2;
```
rdblue: Why divide by 2? Shouldn't the initial value be `INITIAL_SLAB_SIZE`?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -68,11 +77,19 @@ public boolean hasNext() {
...
-      ++ current;
+      final int result = slabs[currentRow][currentCol];
+      incrementPosition();
       return result;
     }
 
+    private void incrementPosition() {
+      current++;
+      currentCol++;
+      if(currentCol >= slabs[currentRow].length) {
```
rdblue: Nit: there should be a space between `if` and `(`

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/IntListTest.java
```diff
@@ -0,0 +1,81 @@
...
+   * Test IntList of a fairly large size (> MAX_SLAB_SIZE) so that we have multiple slabs
+   * created of varying sizes
+   */
+  @Test
+  public void testListGreaterThanMaxSlabSize() {
+    int testSize = IntList.MAX_SLAB_SIZE * 4 + 100;
+    doTestIntList(testSize);
+  }
+
+  private void doTestIntList(int testSize) {
```
rdblue: I think this should also make assertions about the allocated slabs. These tests would have passed before this commit, wouldn't they?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
piyushnarang: done

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
piyushnarang: we double it when we start with the initSlab. I could also initialize it to a value like -1 and in initSlab check if its == -1 then set to INITIAL_SLAB_SIZE. Thought this was cleaner

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
piyushnarang: fixed

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/IntListTest.java
piyushnarang: yes, they would've passed before this commit too. I can add assertions on the currentSlabSize.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
rdblue: Why not allocate the slab with the current size and then double it for the next? It just means allocating before calculating the next slab's size.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -31,7 +31,13 @@
  */
 public class IntList {
 
-  private static final int SLAB_SIZE = 64 * 1024;
+  public static final int MAX_SLAB_SIZE = 64 * 1024;
```
rdblue: Can these be package private instead of public?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
piyushnarang: Yeah so if I did:
```
private void initSlab() {
    currentSlab = new int[currentSlabSize];
    currentSlabPos = 0;
    updateCurrentSlabSize();
  }
```

That will mean that currentSlabSize at the end of initSlab points to the next slabSize that we want to use (not the current one in use, which I don't think is ideal). I could break up the initSlab to do just the array allocation in one method and the update slab size in a different method. Something like this:
```
private void initSlab() {
    currentSlab = new int[currentSlabSize];
    currentSlabPos = 0;
  }
  // Double slab size up to the MAX_SLAB_SIZE limit
  private void updateCurrentSlabSize() {
    if (currentSlabSize < MAX_SLAB_SIZE) {
      currentSlabSize *= 2;
      if (currentSlabSize > MAX_SLAB_SIZE) {
        currentSlabSize = MAX_SLAB_SIZE;
      }
    }
  }
```
Downstream where we are calling initSlab I can also make a call to updateCurrentSlabSize. 
e.g. 
```
if (slabIsFull) {
 updateCurrentSlabSize()
 initSlab()
}
```
Does this sound reasonable? Or do you have something else in mind?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
piyushnarang: yeah sure

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
rdblue: That works for me. You could also rename `currentSlabSize` to `nextSlabSize` so it's more clear. Checks should all be in terms of some slab's `int[]#length` anyway.

piyushnarang: Yeah so with the initSlab being broken up to not do the update as well, the currentSlabSize is actually the currentSlabSize so that name can stay. I'm renaming initSlab to allocateSlab though.

piyushnarang:
@rdblue - Thanks for taking a look. Pushed a commit that fixes your comments.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
rdblue: +1 Thanks for fixing this, it will really help maintenance work later.

piyushnarang:
@rdblue - ping. Were you able to take a look at the rework? Thanks

rdblue:
+1

Thanks @piyushnarang! I committed this.

piyushnarang:
Thanks for taking a look & merging @rdblue :-)

PR #344, opened 2016-05-11
PARQUET-521: Add Brotli compression codec.

rdblue:
This adds a Brotli codec that shares code with the Snappy codec. Snappy
is "non-blocking", meaning that it always accepts more data and buffers
it without blocking. The first reads is blocking, while the compression
is done off-heap. This strategy doesn't appear to have a noticable
performance impact, but does get better compression for Brotli than
streaming buffers.

The non-blocking part of Parquet's Snappy codec has been refactored so
it can be used for both Brotli and Snappy, which both use direct byte
buffers and compress outside the JVM.

This currently depends on a snapshot release of parquet-format with
[PARQUET-609](https://issues.apache.org/jira/browse/PARQUET-609).

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/BrotliCompressor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/BrotliCompressor.java
@@ -1,0 +1,98 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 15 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/BrotliCompressor.java
```diff
@@ -0,0 +1,98 @@
...
+    this.parameter = new Brotli.Parameter();
+    parameter.setMode(isText(primitive) ? Brotli.Mode.TEXT : Brotli.Mode.GENERIC);
+    parameter.setQuality(quality);
+  }
+
+  @Override
+  protected int getMaxCompressedLength(int numInputBytes) {
+    // this is based on https://github.com/google/brotli/issues/274
+    // that page is not very clear, so this is much more conservative
+    return numInputBytes * 2;
```
piyushnarang: seems like one of the linked issues pointed to this in the Brotli src: https://github.com/google/brotli/search?utf8=%E2%9C%93&q=max_out_size
```
const size_t max_out_size = 2 * bytes + 500;
```

rdblue: There's also a Brotli method that gives the max compressed length that we should get into the JNI bindings. This value has been working fine for all of my tests, but I'll update the value on the next pass.

rdblue:
This is replaced by #430. Brotli is provided by https://github.com/rdblue/brotli-codec.

PR #345, opened 2016-05-20
PARQUET-544: Add closed flag to allow for closeable contract adherence

markreddy:
The closeable interface states: 
> Closes this stream and releases any system resources associated with it. If the stream is already closed then invoking this method has no effect.

As InternalParquetRecordWriter implements this interface we should adhere to this contract.


julienledem:
LGTM

PR #346, opened 2016-05-26
PARQUET-400: Fix for ByteBuffer incomplete read issue

piyushnarang:
Spinning up a new PR as I don't have write permissions to update @jaltekruse 's existing [PR](https://github.com/apache/parquet-mr/pull/306). Fixed a few of the comments that were outstanding on that PR:
- Use `readFully` for the fallback and check if byteBuffer.hasArray() is true.
- Updated `getBuf` to read all the remaining bytes into the byte buffer. While testing this, I noticed that on our Hadoop (2.x) cluster we end up returning fewer bytes than byteBuffer.remaining(), so I've added a loop to ensure we get all the remaining bytes. Also seems to line up with the javadoc for [FSDataInputStream.read(byteBuffer)](https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/fs/FSDataInputStream.html#read(java.nio.ByteBuffer)). 
- Updated the fallback logic. Create a CompatUtil per parquet file. So if we fail to use the byteBuffer based API once, we end up trying again for the next file. To handle scenarios where users are on Hadoop 1.x and this checking on every file is unnecessary, I've added an option using which they can turn it off. (Open to nicer ways of doing this..)


piyushnarang:
@rdblue / @danielcweeks / @jaltekruse - please take a look

piyushnarang:
Ping @rdblue, can you take a look please?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -102,6 +102,10 @@
 
   public static String PARQUET_READ_PARALLELISM = "parquet.metadata.read.parallelism";
 
+  // configure if we want to use Hadoop's V2 read(bytebuffer). If true, we try to read using the
+  // new Hadoop read(ByteBuffer) api. Else, we skip.
```
isnotinvain: can you put some more comments here about what this v2 read is / how it works / why you would want this on or off?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -102,6 +102,10 @@
...
+  public static String PARQUET_HADOOP_BYTEBUFFER_READ = "parquet.read.useByteBuffer";
```
isnotinvain: maybe `parquet.read.use.byte.buffer`? I'm not sure what the convention here has been in the past but i think it's been just lots of `.`s

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -950,7 +958,7 @@ protected PageHeader readPageHeader() throws IOException {
         // to allow reading older files (using dictionary) we need this.
         // usually 13 to 19 bytes are missing
         // if the last page is smaller than this, the page header itself is truncated in the buffer.
-        this.byteBuf.rewind(); // resetting the buffer to the position before we got the error
+        this.byteBuf.position(initialPos); // resetting the buffer to the position before we got the error
```
isnotinvain: was this a bug?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
```diff
@@ -66,49 +82,73 @@ private V21FileAPI() throws ReflectiveOperationException {
...
+    int initPos = readBuf.position();
+    int remaining = readBuf.remaining();
+    f.readFully(readBuf.array(), readBuf.arrayOffset(), readBuf.remaining());
+    readBuf.position(initPos + remaining);
+    return remaining;
+  }
+
+  private int readWithNewArray(FSDataInputStream f, ByteBuffer readBuf) throws IOException {
+    int remaining = readBuf.remaining();
+    byte[] buf = new byte[readBuf.remaining()];
```
isnotinvain: `byte[] buf = new byte[remaining];` ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
```diff
@@ -66,49 +82,73 @@ private V21FileAPI() throws ReflectiveOperationException {
...
     }
-    return res;
+
+    return remaining;
   }
+
+  private int readWithExistingArray(FSDataInputStream f, ByteBuffer readBuf) throws IOException {
+    int initPos = readBuf.position();
+    int remaining = readBuf.remaining();
+    f.readFully(readBuf.array(), readBuf.arrayOffset(), readBuf.remaining());
```
isnotinvain: `f.readFully(readBuf.array(), readBuf.arrayOffset(), remaining)` ?

isnotinvain:
Invoking the read method via reflection could itself be slow right? I don't have an intuition on how slow that can be, but for such a low level thing as this it seems surprising that we'd do this.

instead of having this Compatibility helper class that has v1/v2 switches all over the place, can we just make an interface (or abstract class) for these operations w/ 2 implementations? Then we just pick an implementation at startup and don't need any if v1 / if v2 logic. Also IIRC you get some decent performance optimizations if you only ever class load a single implementation of a particular interface / abstract class, so it should be better than calling `invoke` all the time.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
```diff
@@ -66,49 +82,73 @@ private V21FileAPI() throws ReflectiveOperationException {
...
-              "using the Hadoop 2 ByteBuffer based read method.", e.getCause());
-        }
-      } catch (IllegalAccessException e) {
-        // This method is public because it is defined in an interface,
-        // there should be no problems accessing it
-        throw new ShouldNeverHappenException(e);
+  private int readWithByteBuffer(FSDataInputStream f, ByteBuffer readBuf) throws IOException {
+    int remaining = readBuf.remaining();
+    try {
+      while (readBuf.hasRemaining()) {
```
isnotinvain: is there no readFully() method we can call that handles this for us? that seems surprising.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: added some comments. Let me know if I can clarify more.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: done. So.many.periods. :-)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: yeah seems like it. Rewind sets pos to 0. Think this was fixed by @jaltekruse in one of the commits I pulled in to complete this PR.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
piyushnarang: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
piyushnarang: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
piyushnarang: Yeah not that I'm aware of https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/fs/FSDataInputStream.html#read(java.nio.ByteBuffer)
`Reads up to buf.remaining() bytes into buf. Callers should use buf.limit(..) to control the size of the desired read.`

piyushnarang:
@isnotinvain - I like your idea of skipping using reflection. Will look into creating a parquet-hadoop2 module in the project that depends on hadoop 2.x so that we have the `readFully(ByteBuffer)` available. If I understand correctly though, we'll still need reflection at startup (to figure out which concrete implementation of the interface we want to load up).

piyushnarang:
@isnotinvain - updated the implementation to skip reflection on the individual read calls. Just using reflection at the start to figure out which of the two interfaces to use. Tested this out with v1 reads & v2 reads with a hadoop job - seems to work ok.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityReader.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface CompatibilityReader {
+
+  /**
+   * This method attempts to read into the provided readBuffer, readBuffer.remaining() bytes.
+   * @return Number of bytes read - should be readBuf.remaining()
+   * @throws EOFException if readBuf.remaining() is greater than the number of bytes available to
+   * read on the FSDataInputStream f.
+   */
+  int readBuf(FSDataInputStream f, ByteBuffer readBuf) throws IOException;
```
isnotinvain: should this be called readFully?

parquet-hadoop2/src/main/java/org/apache/parquet/hadoop/util/CompatibilityReaderV2.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Uses the Hadoop V2 read(ByteBuffer) APIs to read data.
+ */
+public class CompatibilityReaderV2 implements CompatibilityReader {
+
+  @Override
+  public int readBuf(FSDataInputStream f, ByteBuffer readBuf) throws IOException {
+    int remaining = readBuf.remaining();
+    while (readBuf.hasRemaining()) {
+      int readCount = f.read(readBuf);
```
isnotinvain: so why can't we use readFully here? There's a bug in the implementation of readFully()? can we add a comment and a pointer to a jira for that bug?

isnotinvain:
Do we need the configuration property for whether to use v1 or v2? It seems like we can auto-detect this, so any reason for the added config?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityReader.java
piyushnarang: yeah can rename it to readFully

parquet-hadoop2/src/main/java/org/apache/parquet/hadoop/util/CompatibilityReaderV2.java
piyushnarang: There isn't a version of readFully that supports reading to a byteBuffer - https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/fs/FSDataInputStream.html#read(java.nio.ByteBuffer

piyushnarang:
The rationale for the config property was to avoid performing the reflection based check for every file attempted to be read by Parquet in some scenarios. If you know you're running Parquet in a v1 based setup you could specify that you don't want the byteBuffer based read and directly end up using the V1 APIs. We could do the check once and store the result as a static member variable but that means you're constrained to that value for the entire JVM runtime. I believe @rdblue brought that up as a concern on the previous version of this PR - https://github.com/apache/parquet-mr/pull/306. 
Open to other suggestions as well..

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: Is there a test for this? We should make sure we have one.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1039,8 +1051,11 @@ public void addChunk(ChunkDescriptor descriptor) {
     public List<Chunk> readAll(FSDataInputStream f) throws IOException {
       List<Chunk> result = new ArrayList<Chunk>(chunks.size());
       f.seek(offset);
+
+      //Allocate the bytebuffer based on whether the FS can support it.
```
rdblue: Nit: There should be a space before "Allocate".

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -102,6 +103,13 @@
 
   public static String PARQUET_READ_PARALLELISM = "parquet.metadata.read.parallelism";
 
+  // configure if we want to use Hadoop's V2 read(bytebuffer) API.
+  // If true, we try to read using the new Hadoop read(ByteBuffer) api. This reads data into the provided
+  // byteBuffer and allows us to potentially take advantage zero-copy read path in Hadoop.
+  // Else, we skip and either read into the byteBuffer's array (if its present) or allocate one and copy
+  // into it.
+  public static String PARQUET_HADOOP_BYTEBUFFER_READ = "parquet.read.use.byte.buffer";
```
rdblue: Why would we want to turn this property off? Does your performance eval use the byte buffer read path?

parquet-hadoop2/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-column</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>${hadoop2.version}</version>
```
rdblue: Rather than adding a hadoop2 module, what about updating Parquet's dependency to hadoop 2.x? I think it's about time. That way we could avoid adding another module and needing to include it at runtime. The class wouldn't load in Hadoop 1 so it would have the same effect.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
```diff
@@ -18,97 +18,46 @@
...
+  private static boolean isHadoop2x() {
+    boolean v2 = true;
     try {
-      return method.invoke(instance, args);
-    } catch (IllegalAccessException e) {
-      throw new IllegalArgumentException(errorMsg, e);
-    } catch (InvocationTargetException e) {
-      throw new IllegalArgumentException(errorMsg, e);
+      Class.forName("org.apache.hadoop.io.compress.DirectDecompressor");
+    } catch (ClassNotFoundException cnfe) {
```
rdblue: This needs to catch `NoClassDefFoundError` as well, which is thrown when `DirectDecompressor` exists but a dependency of that class does not. That could happen in strange classpaths.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
```diff
@@ -18,97 +18,46 @@
...
+  // If it is, we try to instantiate the V2 CompatibilityReader.
+  private static boolean isHadoop2x() {
+    boolean v2 = true;
     try {
-      return method.invoke(instance, args);
-    } catch (IllegalAccessException e) {
-      throw new IllegalArgumentException(errorMsg, e);
-    } catch (InvocationTargetException e) {
-      throw new IllegalArgumentException(errorMsg, e);
+      Class.forName("org.apache.hadoop.io.compress.DirectDecompressor");
```
rdblue: Why use `DirectDecompressor` and not the class we're actually interested in, `org.apache.hadoop.fs.ByteBufferReadable`?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestCompatibilityReaderV1.java
```diff
@@ -0,0 +1,108 @@
...
+
+    int readCount = reader.readFully(fsDataInputStream, byteBuffer);
+    Assert.assertEquals("Mismatching no of chars read", 10, readCount);
+    Assert.assertFalse("Byte buffer not full", byteBuffer.hasRemaining());
+  }
+
+  @Test
+  public void testReadBufWithSmallerBuffer() throws Exception {
+    CompatibilityReader reader = CompatibilityUtil.getHadoopReader(false);
+    ByteBuffer byteBuffer = ByteBuffer.allocate(5);
```
rdblue: This should also test a large buffer with different values for `position` and `limit`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: Yeah so @isnotinvain and I had some discussion about this property as well (see his comments above). So his suggestion was to get rid of this property all together. We can either:
1) Cache in a static var whether we computed Hadoop V1 / V2 classes and reuse that. Saves us the reflection call on subsequent files. 
2) Perform the Hadoop V1 / V2 check every time we start reading a file. If there is a chance that the Hadoop V1 / V2 situation changes in the lifetime of the JVM, this will help us pick up the appropriate approach. Not sure what the use-case is for this though. I added the knob here to save us the reflection call when you know for sure you're on Hadoop V1. 

Personally I like option 1. We pay the cost once and it helps us get rid of one knob to tune. Thoughts?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
piyushnarang: Sure can switch to ByteBufferReadable. Didn't really revisit it while pulling in Jason's original implementation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: I'm currently testing something out on this. Basically, we should get the class for ByteBufferReadable, then use it like this:

```
if (bbReadableClass != null && newV2ReaderCtor != null && bbReadableClass.isInstance(stream)) {
  return newV2ReaderCtor.newInstance(stream);
} else {
  return new V1Reader(stream);
}
```

parquet-hadoop2/pom.xml
piyushnarang: Yeah I don't mind updating to Hadoop 2.x, would be nice to skip the additional module. I'll try this out and see if there's a lot of stuff that breaks. If I understand correctly, this should still work in Hadoop V1? (just that the `ByteBufferReadable` will not be loaded?)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompatibilityUtil.java
piyushnarang: thanks for pointing out. Fixed.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestCompatibilityReaderV1.java
piyushnarang: sounds good.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
piyushnarang: I'm not sure this code is even needed any more. The WorkaroundChunk class was added 3 years back and the comment from then says "it deals with a now fixed bug where compressedLength was missing a few bytes". While we do seem to exercise it (as we always use it for the last chunk), I'm not sure how often the bug which this catch was meant to solve gets hit. 
If we need it, I can try and create a unit test for the WorkaroundChunk class. Don't see a straightforward way to test it as part of the FileReader as it is too many levels deep in the class.

piyushnarang:
Closing in favor of: https://github.com/apache/parquet-mr/pull/349

PR #348, opened 2016-06-28
PARQUET-645: Fix null handling in DictionaryFilter.

rdblue:
This fixes how null is handled by `DictionaryFilter` for equals predicates. Null is never in the dictionary and is encoded by the definition level, so the `DictionaryFilter` would never find the value in the dictionary and would incorrectly filter row groups whenever the filter was `col == null`.


piyushnarang:
👍

julienledem:
LGTM

PR #349, opened 2016-06-29
PARQUET-400: Replace CompatibilityUtil with SeekableInputStream.

rdblue:
This fixes PARQUET-400 by replacing `CompatibilityUtil` with `SeekableInputStream` that's implemented for hadoop-1 and hadoop-2. The benefit of this approach is that `SeekableInputStream` can be used for non-Hadoop file systems in the future.

This also changes the default Hadoop version to Hadoop-2. The library is still compatible with Hadoop 1.x, but this makes building Hadoop-2 classes, like `H2SeekableInputStream`, much easier and removes the need for multiple hadoop versions during compilation.


rdblue:
@piyushnarang, this is the update to your PR that I think may be a bit cleaner.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  // Visisble for testing
+  static void readFullyHeapBuffer(FSDataInputStream f, ByteBuffer buf) throws IOException {
+    f.readFully(buf.array(), buf.arrayOffset() + buf.position(), buf.remaining());
+    buf.position(buf.limit());
+  }
+
+  // Visisble for testing
+  static int readDirectBuffer(FSDataInputStream f, ByteBuffer buf, byte[] temp) throws IOException {
```
rdblue: To cut down on allocation, this uses an 8k buffer that's tied to the instance. We can change this and maybe even make it configurable to find out where a good trade-off is.

piyushnarang:
Think for the most part this is similar to what we had in the other PR. Guess I'm not entirely clear on how this enables non Hadoop file systems in the future - you're still depending on `FSDataInputStream` and such in the SeekableInputStream. You could in the other implementation just derive [CompatibilityReader](https://github.com/apache/parquet-mr/pull/346/files#diff-8c94893b24edb4a7971eb0cbd167754aR33) right? 

Taking a step back, does it make sense to break out the ParquetFileReader code to create a ParquetFileReader interface / abstract that has a Hadoop based implementation & other implementations? Feels like we're trying to retro fit this support in the ParquetFileReader cause it currently has the logic inline. If we pulled it out it might make things easier.

Think we should try and eliminate the reflection overhead (or benchmark it to confirm its low) if possible. With this implementation we call the reflection based ctor of the V2 reader everytime we create a ParquetFileReader.

rdblue:
@piyushnarang, it's definitely similar, I just made a few changes to your PR.

The `SeekableInputStream` is a start for separating Parquet's internals from Hadoop classes, like `FSDataInputStream`. That's what other projects, like Avro, use to pass Hadoop or non-Hadoop input streams to the internals and this has been something we've been meaning to do for a while. Now seems like a good time. You're right that as long as it imports Hadoop classes, it doesn't help much so I've separated the helper method from the actual class. Maybe we should move `SeekableInputStream` to parquet-common as well?

I think we do eventually want to make a reader API that is independent of Hadoop, but I don't think we need to do it in this PR. I'm just trying to make this get us further along toward the goal of not needing the Hadoop API to use Parquet.

On reflection overhead: this has the same overhead that the previous implementations had, instantiating a class to handle Hadoop 2 streams. `newV2Reader` was looking up the class and calling `newInstance`. And, I don't think this is going to make much of a difference because it is once per stream.

piyushnarang:
Yeah I personally prefer keeping such changes small and iterative. Would be nice if we just tackled fixing the immediate concern (reads are broken) first and then followed up with a change to make Parquet more friendly to non-Hadoop setups. That said, if you're keen on adding that as part of this fix let's go ahead. I can close my prior PR in favor of this. 

Makes sense to move `SeekableInputStream` to parquet-common. Wouldn't make sense to take a dependency on parquet-hadoop if it isn't needed. 

Let me know when you're happy with the implementation, I can take a more detailed look.

rdblue:
@piyushnarang, I'd prefer to go with this approach since it solves the same problem and sets us up for separating the APIs later. I've added a big file of tests based on your MockInputStream so I think this is ready to go.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -0,0 +1,40 @@
...
+ *  under the License.
+ */
+
+package org.apache.parquet.io;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public abstract class SeekableInputStream extends InputStream {
```
piyushnarang: Lets add some javadoc on this abstract class and its purpose etc.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -102,6 +102,13 @@
 
   public static String PARQUET_READ_PARALLELISM = "parquet.metadata.read.parallelism";
 
+  // configure if we want to use Hadoop's V2 read(bytebuffer) API.
+  // If true, we try to read using the new Hadoop read(ByteBuffer) api. This reads data into the provided
+  // byteBuffer and allows us to potentially take advantage zero-copy read path in Hadoop.
+  // Else, we skip and either read into the byteBuffer's array (if its present) or allocate one and copy
+  // into it.
+  public static String PARQUET_HADOOP_BYTEBUFFER_READ = "parquet.read.use.byte.buffer";
```
piyushnarang: think we can drop this now

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
piyushnarang: Wondering if SeekableInputStream is the right name? Maybe ParquetInputStream?

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public abstract long getPos() throws IOException;
+
+  public abstract void seek(long newPos) throws IOException;
+
+  public abstract void readFully(byte[] bytes) throws IOException;
```
piyushnarang: lets document the contract of this method a bit - I did that here: https://github.com/apache/parquet-mr/pull/346/files#diff-8c94893b24edb4a7971eb0cbd167754aR41. We can either reuse that or tweak it..

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
```diff
@@ -0,0 +1,150 @@
...
+
+package org.apache.parquet.hadoop.util;
+
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.parquet.io.SeekableInputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+public class H1SeekableInputStream extends SeekableInputStream {
```
piyushnarang: rename to HadoopV1*InputStream? 
maybe also add some javadoc?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
```diff
@@ -0,0 +1,89 @@
...
+
+import org.apache.hadoop.fs.ByteBufferReadable;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.parquet.Preconditions;
+import org.apache.parquet.io.SeekableInputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+public class H2SeekableInputStream extends SeekableInputStream {
```
piyushnarang: rename to HadoopV2*Stream?
Also javadoc.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.parquet.io.ParquetDecodingException;
+import org.apache.parquet.io.SeekableInputStream;
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.nio.ByteBuffer;
+
+public class HadoopStreams {
+
+  private static final Log LOG = Log.getLog(SeekableInputStream.class);
```
piyushnarang: typo HadoopStreams?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public static SeekableInputStream wrap(FSDataInputStream stream) {
+    if (byteBufferReadableClass != null && h2SeekableConstructor != null &&
+        byteBufferReadableClass.isInstance(stream) &&
+        supportsByteBufferReads(stream)) {
+      try {
+        return h2SeekableConstructor.newInstance(stream);
+      } catch (InstantiationException e) {
+        LOG.warn("Could not instantiate H2SeekableInputStream, falling back to byte array reads", e);
```
piyushnarang: typo - should be HadoopStreams here and below

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+    } catch (NoClassDefFoundError e) {
+      return null;
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Class<SeekableInputStream> getH2SeekableClass() {
+    try {
+      return (Class<SeekableInputStream>) Class.forName(
+          "org.apache.parquet.hadoop.util.H2SeekableInputStream");
```
piyushnarang: pull out as a constant

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Constructor<SeekableInputStream> h2SeekableConstructor;
+
+  static {
+    byteBufferReadableClass = getReadableClass();
+    h2SeekableConstructor = getH2SeekableConstructor();
+  }
+
+  public static SeekableInputStream wrap(FSDataInputStream stream) {
+    if (byteBufferReadableClass != null && h2SeekableConstructor != null &&
```
piyushnarang: lets wrap all these conditionals into a method to improve readability. Something like:
`private boolean useHadoop2Reads() { ... }` ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+      } catch (InvocationTargetException e) {
+        throw new ParquetDecodingException(
+            "Could not instantiate H2SeekableInputStream", e.getTargetException());
+      }
+    } else {
+      return new H1SeekableInputStream(stream);
+    }
+  }
+
+  private static final ByteBuffer ZERO_LEN_BYTE_BUFFER = ByteBuffer.wrap(new byte[0]);
```
piyushnarang: lets move this static to the top of the class?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+  private static boolean supportsByteBufferReads(FSDataInputStream stream) {
+    // FSDataInputStream implements ByteBufferReadable, but may throw
+    // UnsupportedOperationException if it isn't actually supported. This tests
+    // whether the method throws by trying to read a 0-length buffer, which has
+    // no effect on the stream when it is supported.
+    try {
+      stream.read(ZERO_LEN_BYTE_BUFFER);
+      return true;
+    } catch (UnsupportedOperationException e) {
+      return false;
```
piyushnarang: to help debugging should we log the false / null scenarios in this class? we basically end up falling back in them to the HadoopV1 stream. We could log as debug as it might be spammy

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+      return Class.forName("org.apache.hadoop.fs.ByteBufferReadable");
+    } catch (ClassNotFoundException e) {
+      return null;
+    } catch (NoClassDefFoundError e) {
+      return null;
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Class<SeekableInputStream> getH2SeekableClass() {
```
piyushnarang: rename to getHadoopV2SeekableClass?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+      return (Class<SeekableInputStream>) Class.forName(
+          "org.apache.parquet.hadoop.util.H2SeekableInputStream");
+    } catch (ClassNotFoundException e) {
+      return null;
+    } catch (NoClassDefFoundError e) {
+      return null;
+    }
+  }
+
+  private static Constructor<SeekableInputStream> getH2SeekableConstructor() {
```
piyushnarang: getHadoopV2SeekableConstructor?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestByteBufferReads.java
```diff
@@ -0,0 +1,826 @@
...
+import org.apache.parquet.hadoop.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+import java.io.ByteArrayInputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.concurrent.Callable;
+
+public class TestByteBufferReads {
```
piyushnarang: this seems to be only testing the H1 streams right? should we:
1) rename this class to TestHadoopV1*Streams? 
2) Add a bunch of tests for H2 as well

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
julienledem: This api is not really Parquet specific. It allows to seek and read blocks of the file.
Being in a parquet package is enough "parquet" in the name I think.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
julienledem: agreed. We should document this class.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public abstract void readFully(byte[] bytes, int start, int len) throws IOException;
+
+  public abstract int read(ByteBuffer buf) throws IOException;
+
+  public abstract void readFully(ByteBuffer buf) throws IOException;
+
+}
```
julienledem: this is a great addition, this should please people who want more decoupling from hadoop as well.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -519,7 +526,7 @@ public ParquetFileReader(
     this.conf = configuration;
     this.fileMetaData = fileMetaData;
     FileSystem fs = filePath.getFileSystem(configuration);
-    this.f = fs.open(filePath);
+    this.f = HadoopStreams.wrap(fs.open(filePath));
```
julienledem: eventually we could replace the "FileStatus file" parameter with something that provides a SeekableInputStream.
Possibly we just need length and Name to SeekableInputStream and call it ScannableFile ?
It is something that has a length (so that we can locate the footer) and where we and load predefined chunks from, in one go.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
julienledem: +1

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
julienledem: same comment

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.parquet.Log;
+import org.apache.parquet.io.ParquetDecodingException;
+import org.apache.parquet.io.SeekableInputStream;
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.nio.ByteBuffer;
+
+public class HadoopStreams {
```
julienledem: javadoc

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Log LOG = Log.getLog(SeekableInputStream.class);
+
+  private static final Class<?> byteBufferReadableClass;
+
+  private static final Constructor<SeekableInputStream> h2SeekableConstructor;
+
+  static {
+    byteBufferReadableClass = getReadableClass();
+    h2SeekableConstructor = getH2SeekableConstructor();
```
julienledem: those initializers can be just on the same line as the declaration. Right?

julienledem:
I made some comments. Thanks @piyushnarang and @rdblue for taking care of this

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
rdblue: Will add docs. I don't think we need to rename, but that's because I don't think this class should be public. No need to make it public if we recommend using `HadoopStreams.wrap`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: I don't think we need to log this because it's determined by the implementation at release time.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
piyushnarang: making it private sounds good. Renaming it seems nicer from a readability perspective.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: I'm not sure we need both the stream and the file to be one object, but I'm not sure what the right design for it is. I think it's pretty safe to say we'll want a seekable stream and we can add a ScannableFile that inherits from that stream later when we design a Hadoop-free API.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: We don't need to expose the class, so I don't think we need to expose its name anywhere. That pretty much removes the need to separate this out as a constant.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Fixed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
rdblue: Added javadoc and made the class package-private. No need to expose these.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
rdblue: Added javadoc for the whole class.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: Removed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Removed in favor of a better check. FSDataInputStream decides whether to throw the exception based on whether the underlying stream is a ByteBufferReadable, so instead of running this method I'm just checking that stream directly now.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Added.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Fixed.

rdblue:
@piyushnarang, @julienledem, I've addressed the review items.

I tested the hadoop 2 readFully method for ByteBuffer using the same tests I wrote for the hadoop 1 implementation, but the test won't compile in hadoop-1 so we would have to create another module for hadoop-2 tests and exclude it from the hadoop 1 test run. I don't think testing the readFully method is worth the trouble because it is so simple and unlikely to change from its current working and tested state. The tests are available in 4f273e4 if you'd like to have a look, but they were removed in the next commit to get tests passing in Jenkins.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Wraps a {@link FSDataInputStream} in a {@link SeekableInputStream}
+   * implementation for Parquet readers.
+   *
+   * @param stream a Hadoop FSDataInputStream
+   * @return a SeekableInputStream
+   */
+  public static SeekableInputStream wrap(FSDataInputStream stream) {
+    if (byteBufferReadableClass != null && h2SeekableConstructor != null &&
+        byteBufferReadableClass.isInstance(stream.getWrappedStream())) {
```
piyushnarang: I'm wondering if this check is a bit fragile? It will work if the stream's immediate inner stream is an instance of ByteBufferReadable that has a concrete implementations of `read(ByteBuffer buf)`. If that stream too ends up delegating to its inner stream then we might have a problem right?

piyushnarang:
@rdblue - given that we expect the bulk of people to use the hadoop 2 version of the code (correct me if I'm wrong), we should have the tests running. Even if we have to create a new module for the tests, I think it is probably worth it as this functionality is pretty important to ensuring Parquet reads work. Will give future developers something to use in case they need to change the hadoop 2 read code. Might have misunderstood so let me know if you had something else in mind..

rdblue:
I don't think it's worth including the tests. If we do end up modifying the implementation, which I think is unlikely, then we can take the test cases from the hadoop-1 tests then. That's what I did to verify the current implementation. There's a good chance that we will be able to drop support for Hadoop 1 before we need to change anything here, and the method is both simple and reasonably tested by higher-level checks.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Yes, if the wrapped class implemented ByteBufferReadable in order to delegate to an implementation that may not be, then it would be a problem and that would be caught by the previous version but not this method. But, I think it's fine to assume that when a stream implements ByteBufferReadable that it actually does. That's what ByteBufferReadable is intended to signal and it's weird that we have to implement a special case for FSDataInputStream at all.

Also, the previous implementation calls a method that isn't present in hadoop-1 and to do that relies on when methods are linked. I think the trade-off is worth this cleaner way of deciding whether to use the ByteBuffer interface, but I'm happy to change it if others agree with you that we should handle more levels of wrapper classes.

piyushnarang:
I might be missing something here. Why not keep the v2 tests as the default? We switched the hadoop version in the pom to 2.x right?

rdblue:
Right now our test setup relies on compiling the main classes once and the test classes for each test environment. So while our main classes can rely on hadoop-2 classes as long as they are backward-compatible, we can't have tests that rely on hadoop-2 classes. Otherwise, we have to add a separate module that we exclude when running hadoop-1 tests. Since this is the only case where we need it and the amount of code to test is tiny, I don't think it's worth the trouble.

The bulk of the code that needs tested here is for Hadoop 1 because we're implementing methods that we rely on being provided in Hadoop 2. Perhaps we could be more clever here and find an easy way to test the code without needing the ByteBufferReadable interface. Still, I'm not inclined to spend a ton of time on it since I think what's already here is reliable.

piyushnarang:
Ok, the setup makes more sense now. These changes look good to me. In general I prefer to err on the side of more tests as it can save us some pain later on, but I think we can go with your judgement on this.

rdblue:
@julienledem, could you take another look? I think this is about ready. Once it's in we can commit @piyushnarang's encoding fixes and start a release candidate.

piyushnarang:
@rdblue - My encoding fixes were merged a few weeks back - https://github.com/apache/parquet-mr/pull/347

isnotinvain:
One thing I don't follow from the discussion -- as is do our tests cover the codepaths that will be used in hadoop2 environments?

isnotinvain:
It looks like the H2 SeekableInputStream is never tested, is that right? I think we should definitely figure out how to make sure it is similarly tested.

rdblue:
@isnotinvain, @piyushnarang, I've added back the hadoop 2 tests for the readFully implementation and fixed them to compile in hadoop-1. It required adding a wrapper instead of passing FSDataInputStream, but with only one implementation (the interface is package-private) it should be fine at runtime.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private class H2Reader implements Reader {
+    @Override
+    public int read(ByteBuffer buf) throws IOException {
+      return stream.read(buf);
+    }
+  }
+
+  public static void readFully(Reader stream, ByteBuffer buf) throws IOException {
```
piyushnarang: rename from stream to reader

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestByteBufferReads.java
```diff
@@ -0,0 +1,1133 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.concurrent.Callable;
+
+import static org.apache.parquet.hadoop.util.MockInputStream.TEST_ARRAY;
+
+public class TestByteBufferReads {
```
piyushnarang: minor readability suggestion - would be nice if this class is broken up into a class with H1 tests and one with H2 tests. Might be nice if we name them something like TestH1SeekableInputStream / TestH2SeekableInputStream as well.

piyushnarang:
Thanks Ryan. Couple of minor comments but think it looks good to me.

rdblue:
I fixed the nits that @piyushnarang pointed out. Anything else? @julienledem or @isnotinvain?

isnotinvain:
👍 Thanks @piyushnarang and @rdblue for tackling this!

+1

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * {@code SeekableInputStream} is an interface with the methods needed by
+ * Parquet to read data from a file or Hadoop data stream.
+ */
+public abstract class SeekableInputStream extends InputStream {
```
julienledem: Let's add getLength() as well.
That's the only extra information we need to read the footer. It's missing in FSDataInputStream and that would simplify some code where we have to pass the FileStatus object along.
it is always available since we get the stream with FileStatus.open().

rdblue: A seekable stream is a fairly normal construct, while a stream that knows its own length is irregular. I think we are probably better off having a slightly higher-level concept of a stream provider that knows the length of streams it opens. That would basically encapsulate FileStatus and FileSystem so you can pass a single object that can open parallel streams for a single Parquet file.

How about doing this as a follow up? This issue is a blocker for 1.9.0 so I'd like to get it in. We can discuss the right way to pass around the length but also work toward getting 1.9.0 out. I'll open an issue for this.

rdblue: Opened [PARQUET-674](https://issues.apache.org/jira/browse/PARQUET-674).

julienledem: fair enough

julienledem:
+1

isnotinvain:
Looks like https://github.com/apache/parquet-mr/commit/76a2ac814caa194c46be1cf7a3f5dc129546b1c1 conflicts with this PR

rdblue:
@isnotinvain, I rebased and fixed the conflicts, which were small. I'll commit this later today.

julienledem:
sounds good to me

isnotinvain:
Great, thanks!

piyushnarang:
Thanks, glad this is now out :-)

PR #351, opened 2016-07-10
PARQUET-660: Ignore extension fields in protobuf messages.

jkukul:
Currently, converting protobuf messages with extension can result in an uninformative error or a data corruption. A more detailed explanation in the corresponding [jira](https://issues.apache.org/jira/browse/PARQUET-660).

This patch simply ignores extension fields in protobuf messages. 

In the longer run, I'd like to add a proper support for Protobuf extensions. This might take a little longer though, so I've decided to improve the current situation with this patch.

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
@@ -156,7 +156,6 @@ public class ProtoWriteSupport<T extends MessageOrBuilder> extends WriteSupport<
       List<Descriptors.FieldDescriptor> fields = descriptor.getFields();
       fieldWriters = (FieldWriter[]) Array.newInstance(FieldWriter.class, fields.size());
 
-      int i = 0;
       for (Descriptors.FieldDescriptor fieldDescriptor: fields) {
         String name = fieldDescriptor.getName();
         Type type = schema.getType(name);
@@ -169,8 +168,7 @@ public class ProtoWriteSupport<T extends MessageOrBuilder> extends WriteSupport<
         writer.setFieldName(name);
         writer.setIndex(schema.getFieldIndex(name));
 
-        fieldWriters[i] = writer;
-        i++;
+        fieldWriters[fieldDescriptor.getIndex()] = writer;
       }
     }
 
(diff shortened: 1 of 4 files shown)


parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -218,6 +218,13 @@ private void writeAllFields(MessageOrBuilder pb) {
 
       for (Map.Entry<Descriptors.FieldDescriptor, Object> entry : changedPbFields.entrySet()) {
         Descriptors.FieldDescriptor fieldDescriptor = entry.getKey();
+
+        if(fieldDescriptor.isExtension()) {
+          // Field index of an extension field might overlap with a base field, thus
+          // they have to be ignored.
+          continue;
+        }
+
```
jkukul: Another option would be throw an Exception here and explicitly do not support messages.

I could also log a warning here, somehow only once though, to avoid the noise.

julienledem:
@lukasnalezenec what do you think?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
julienledem: I like the approach of throwing an exception. It is easy to add support for a special case later. It is harder to change behavior as it breaks existing applications.
For example in Avro we explicitly throw in the case of recursive schemas.

lukasnalezenec:
Exception is good idea but I am not sure if it is acceptable change of behaviour.
We should at least add some logging there.
How about making it configurable ?

jkukul:
@lukasnalezenec I adjusted the PR to throw an exception in 27580ab32f161e9e733727117a33272fa05c515e.  I also think this is a better solution, after some consideration. 

The PR doesn't introduce any change in behaviour. An informative exception is thrown, instead of a confusing one which would follow otherwise.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -169,8 +168,7 @@ void writeField(Object value) {
         writer.setFieldName(name);
         writer.setIndex(schema.getFieldIndex(name));
 
-        fieldWriters[i] = writer;
-        i++;
+        fieldWriters[fieldDescriptor.getIndex()] = writer;
```
julienledem: is this the same change?

jkukul: Yes, fields in the list over which we're iterating are ordered by index. Here's how the list is built:
https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/Descriptors.java#L826

julienledem:
this looks good to me.
@lukasnalezenec ?

schrepfler:
Is there any movement on this item and get support of serialising protobuffer extensions.

PR #352, opened 2016-07-11
PARQUET-651: Improve Avro's isElementType check.

rdblue:
The Avro implementation needs to check whether the read schema that is
passed by the user (or automatically converted from the file schema)
expects an extra 1-field layer to be returned, which matches the
previous behavior of Avro when reading a 3-level list. Before this
commit, the check was done by testing the structure of the expected list
element type against the repeated group's schema. If they matched, then
Avro assumed that the user expected an extra layer. However, for records
that happened to match (1-field records with a field named "element")
the check could be wrong and would cause exceptions later.

This commit updates the check to convert the file's element schema to
Avro and compare the compatibility of that schema with what was passed
by the user. This checks the entire tree from the element down and gets
the answer right based on the element and its children, not just the
field names on the element.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -42,9 +42,11 @@ import org.apache.avro.AvroTypeException;
 import org.apache.avro.Conversion;
 import org.apache.avro.LogicalType;
 import org.apache.avro.Schema;
+import org.apache.avro.SchemaCompatibility;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.reflect.AvroIgnore;
 import org.apache.avro.reflect.AvroName;
+import org.apache.avro.reflect.AvroSchema;
 import org.apache.avro.reflect.ReflectData;
 import org.apache.avro.reflect.Stringable;
 import org.apache.avro.specific.SpecificData;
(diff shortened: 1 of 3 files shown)


rdblue:
@liancheng, could you take a look at this? It's based on your solution to SPARK-16344. Thanks!

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -827,6 +832,14 @@ public void end() {
...
   }
 
+  // Converter used to test whether a requested schema is a 2-level schema.
+  // This is used to convert the file's type assuming that the file uses
+  // 2-level lists and the result is checked to see if it matches the requested
+  // element type. This should always convert assuming 2-level lists because
+  // 2-level and 3-level can't be mixed.
+  private static final AvroSchemaConverter CONVERTER =
+      new AvroSchemaConverter(true);
+
```
liancheng: Instead of always set `AvroScehmaConverter.assumeRepeatedIsListElement` to true, shall we pass the properly constructed `AvroScehmaConverter` from `AvroReadSupport` to `AvroRecordMaterializer`, and then here?

liancheng: Oh I see your point, there's no need to do so.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -54,7 +54,7 @@
 
   public static final String ADD_LIST_ELEMENT_RECORDS =
       "parquet.avro.add-list-element-records";
-  private static final boolean ADD_LIST_ELEMENT_RECORDS_DEFAULT = true;
+  private static final boolean ADD_LIST_ELEMENT_RECORDS_DEFAULT = false;
```
liancheng: Guess this change caused the build failure.

rdblue: Oops! I was making sure the compatibility tests passed when this is flipped, but didn't mean to change the default.

liancheng:
+1!

PR #353, opened 2016-07-12
PARQUET-654: Add option to disable record-level filtering.

rdblue:
This can be used by frameworks that use codegen for filtering to avoid
running filters within Parquet.


julienledem:
what framework in particular?
LGTM

rdblue:
Spark. There was a small discussion on the dev list about it. I think they at least want to be able to test with Parquet filtering off.

PR #354, opened 2016-07-13
PARQUET-389: Support predicate push down on missing columns.

rdblue:
Predicate push-down will complain when predicates reference columns that aren't in a file's schema. This makes it difficult to implement predicate push-down in engines where schemas evolve because each task needs to process the predicates and prune references to columns not in that task's file. This PR implements predicate evaluation for missing columns, where the values are all null. This allows engines to pass predicates as they are written.

A future commit should rewrite the predicates to avoid the extra work currently done in record-level filtering, but that isn't included here because it is an optimization.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
@@ -167,6 +167,11 @@ public class SchemaCompatibilityValidator implements FilterPredicate.Visitor<Voi
     }
 
     ColumnDescriptor descriptor = getColumnDescriptor(path);
+    if (descriptor == null) {
+      // the column is missing from the schema. evaluation uses calls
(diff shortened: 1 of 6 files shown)


rdblue:
@danielcweeks, @liancheng, this is to avoid the predicate push-down problems in Spark. Can you review? Thanks!

liancheng:
+1

danielcweeks:
+1 as well.  Thanks!

liancheng:
Is it possible to have a 1.8.2 release that includes this fix? I just checked and it seems that there isn't a dedicated branch for 1.8.x?

viirya:
I am curious, why this patch doesn't do the same thing to the `visit` method for `UserDefinedPredicate` in `StatisticsFilter`?

Is it intentional?

viirya:
@liancheng @rdblue @danielcweeks I submitted #389 to extend this kind of fixing to `UserDefinedPredicate`. Don't know if it is appropriate? Please remind me if it is not. Thank you.

PR #357, opened 2016-07-31
PARQUET-669: allow reading footers from provided file listing and streams

robert3005:
The use case is that I want to reuse existing listing of files and avoid doing it again when opening streams. This is in case where filesystem.open is expensive but you have other means of obtaining input stream for a file.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -434,22 +434,30 @@ public class ParquetFileReader implements Closeable {
     FileSystem fileSystem = file.getPath().getFileSystem(configuration);
     FSDataInputStream in = fileSystem.open(file.getPath());
     try {
-      return readFooter(file, in, filter);
+      return readFooter(file.getLen(), file.getPath().toString(), in, filter);
     } finally {
       in.close();
     }
   }
 
-  private static final ParquetMetadata readFooter(FileStatus file, FSDataInputStream f, MetadataFilter filter) throws IOException {
-    long l = file.getLen();
+  /**
+   * Reads the meta data block in the footer of the file using provided input stream
+   * @param fileLen length of the file
+   * @param filePath file location
+   * @param f input stream for the file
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -440,7 +470,15 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
...
-  private static final ParquetMetadata readFooter(FileStatus file, FSDataInputStream f, MetadataFilter filter) throws IOException {
+  /**
+   * Reads the meta data block in the footer of the file using provided input stream
+   * @param f input stream for the file
+   * @param file the parquet File
+   * @param filter the filter to apply to row groups
+   * @return the metadata blocks in the footer
+   * @throws IOException if an error occurs while reading the file
+   */
+  public static final ParquetMetadata readFooter(FileStatus file, FSDataInputStream f, MetadataFilter filter) throws IOException {
```
julienledem: We only need need the file size here not the full FileStatus object.
(since we're making this public)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -237,6 +237,36 @@ private static MetadataFilter filter(boolean skipRowGroups) {
...
   /**
+   * read all the footers of the files provided with input streams
+   * (not using summary files)
+   * @param configuration the conf to access the File System
+   * @param partFiles the files to read with equivalent input streams
+   * @param skipRowGroups to skip the rowGroup info
+   * @return the footers
+   * @throws IOException
+   */
+  public static List<Footer> readAllFootersInParallel(final Configuration configuration, Map<FileStatus, FSDataInputStream> partFiles, final boolean skipRowGroups) throws IOException {
```
julienledem: this does not need to be a map (collection of file length and streams?). Also this method does not seem necessary in Parquet. one may do their own threading if necessary when calling readFooter(stream) defined bellow.

julienledem:
LGTM

PR #358, opened 2016-07-31
PARQUET-668 - Provide option to disable auto crop feature in dump

djhworld:
https://issues.apache.org/jira/browse/PARQUET-668

1. Added option `--disable-crop`
2. Updated `README.md` to reflect changes

### DIFF
--- a/parquet-tools/README.md
+++ b/parquet-tools/README.md
@@ -61,51 +61,13 @@ java jar ./parquet-tools-<VERSION>.jar <command> my_parquet_file.lzo.parquet
 
 ## Commands Usage
 
-To run it on hadoop, you should use "hadoop jar" instead of "java jar"
+To see usage instructions for all commands: 
 
-```sh
-usage: java -jar ./parquet-tools-<VERSION>.jar cat [option...] <input>
-where option is one of:
-       --debug     Enable debug output
-    -h,--help      Show this help string
-       --no-color  Disable color output even if supported
-where <input> is the parquet file to print to stdout
-
-usage: java -jar ./parquet-tools-<VERSION>.jar head [option...] <input>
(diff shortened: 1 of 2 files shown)


parquet-tools/README.md
```diff
@@ -99,9 +99,10 @@ where option is one of:
     -c,--column <arg>  Dump only the given column, can be specified more than
                        once
     -d,--disable-data  Do not dump column data
-       --debug         Disable color output even if supported
+       --debug         Enable debug output
```
julienledem: thanks for fixing. I see the same error above. (line 85, 92)
either add -n in those paragraphs as well or rework a little to avoid repetition. 
(common options paragraph or something)

julienledem:
just one comment regarding README. otherwise. LGTM
Thanks!

parquet-tools/README.md
djhworld: This text in the README is just the output of running `java -jar parquet-tools.jar -h`, and that help string is generated in `Main.java` (it merges the common option flags like `--debug` into every other command, see https://github.com/apache/parquet-mr/blob/master/parquet-tools/src/main/java/org/apache/parquet/tools/Main.java#L94) - so this help text is auto generated 

I've updated the others items in the README with the correct output.

djhworld: Although do we even need the whole help string in the README? This could happen again if the options change. 

Might be better to just direct the user to run `-h` to get usage details

julienledem: good point. Yes, showing the command line to get help would be better.

djhworld:
Updated README as discussed, only the `--help` flag is displayed now.

julienledem:
LGTM

PR #359, opened 2016-07-31
PARQUET-670: allow for different metadata output path than paths in file footers

robert3005:
Since you might have a virtual filesystem and write and read paths are different

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -681,19 +681,28 @@ public class ParquetFileWriter {
    */
   @Deprecated
   public static void writeMetadataFile(Configuration configuration, Path outputPath, List<Footer> footers) throws IOException {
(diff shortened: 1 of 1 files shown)


julienledem:
Thank you for your contribution.
can you add a test for this?
I'm surprised it works without modifying this:
https://github.com/apache/parquet-mr/blob/60b6d5aa3feeb31c03baa9f06e2de32172724d78/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L727

andreweduffy:
(I work with rob)

Which part explicitly should be unit tested here? There shouldn't require a change to mergeFooters because you're passing in the rootpath for the footers as the second argument.

julienledem:
@andreweduffy I see, that makes sense. 
you can add a test around there: https://github.com/apache/parquet-mr/blob/60b6d5aa3feeb31c03baa9f06e2de32172724d78/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java#L554

just writing a metadata file passing the different path and checking that it looks good (possibly that it just did not fail and the file is in the right location).

PR #361, opened 2016-08-22
Quoted identifiers in column names.

ptkool:


julienledem:
Hi @ptkool, thanks for your contribution.
Can you open a Parquet jira for this? (https://issues.apache.org/jira/browse/PARQUET)
then prefix the title of your PR with "PARQUET-XXX: "
Thank you!

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
```diff
@@ -426,7 +426,7 @@ public void testAll() throws Exception {
     assertEquals(emptyMap, nextRecord.get("myemptymap"));
     assertEquals(genericFixed, nextRecord.get("myfixed"));
   }
-
+
```
rdblue: Nit: Please remove the whitespace-only changes. It's harder to pick commits between branches and do maintenance with these changes.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -18,16 +18,19 @@
  */
 package org.apache.parquet.schema;
 
-import java.util.Arrays;
```
rdblue: Nit: Please don't move imports around. (Again, harder to pick commits, rebase, etc.)

parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
```diff
@@ -173,4 +173,16 @@ public void testIDs() throws Exception {
     assertEquals(schema, schema2);
     assertEquals(schema.toString(), schema2.toString());
   }
+
+  @Test
+  public void testQuotedIdentifiers() throws Exception {
+    MessageType schema = new MessageType("test",
+        new PrimitiveType(REQUIRED, BINARY, "foo bar"),
+        new PrimitiveType(REQUIRED, BINARY, "foo()`bar"),
```
rdblue: Do we need to support ` in names? Is it supported elsewhere, like in Hive?

It doesn't seem worth the added complexity in the tokenizer to me.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -142,6 +142,82 @@ public void test() throws Exception {
...
   }
 
+  @Test
+  public void testQuotedColumns() throws Exception {
+    Configuration conf = new Configuration();
+    Path root = new Path("target/tests/TestParquetWriter/");
+    enforceEmptyDir(conf, root);
+    MessageType schema = parseMessageType(
+        "message test { "
+        + "required binary `binary field`; "
```
rdblue: Can you change some of these to test cases other than space? I'd like to see `.` tested, along with other common patterns.

rdblue:
Thanks for posting a patch @ptkool!

Is this preventing Parquet from being used for some cases? I want to make sure the trade-off is worth including this. Right now, we don't have to worry about column name compatibility across object models. Avro, for example, will generate Java and C code for schemas so it only allows column names that are alphanumeric (plus '_'). This would break compatibility with those Avro fields and probably with Thrift and some Protobuf as well. Is that worth fixing strange Hive columns?

I'm wondering if we shouldn't solve this problem by using field IDs or string aliases for column names that are alphanumeric to be consistent across object models.

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
ptkool: :+1:

parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
ptkool: Yes, it is supported in Hive.

https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL

Have a look at the `CREATE TABLE` documentation.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -140,8 +147,14 @@ public Type(String name, Repetition repetition, OriginalType originalType) {
    */
   Type(String name, Repetition repetition, OriginalType originalType, ID id) {
     super();
-    this.name = checkNotNull(name, "name");
+
+    name = checkNotNull(name, "name");
+    if (name.charAt(0) == '`') {
+      name = name.substring(1, name.length() - 1).replace("``", "`");
+    }
```
julienledem: we should check that the last character is also a quote.
We should factor out the logic with MessageTypeParser.Tokenizer.getName()

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -114,6 +116,11 @@ public boolean isMoreRestrictiveThan(Repetition other) {
   private final ID id;
 
   /**
+   * Pattern for matching column names that need to wrapped in backquotes.
+   */
+  private static final Pattern pattern = Pattern.compile("[^a-zA-Z\\d_-]");
```
julienledem: @rdblue : does that satisfy your requirement?
Ideally this should describe only what was supported for parquet before that change.
We don't want to quote strings that worked before without quoting.
Should this be anything that is in MessageTypeParser.Tokenizer's constructor initialization of the delimiter list? " ,;{}()\n\t="
What are the other character that wouldn't work? I think "." since it is used as a column path delimiter.

parquet-column/src/test/java/org/apache/parquet/io/TestColumnIO.java
```diff
@@ -457,6 +457,22 @@ public void testOptionalRequiredInteraction() {
...
+            + "  required group foo {\n"
+            + "    required int64 `bar?@`;\n"
+            + "  }\n"
+            + "}\n");
+
+    GroupFactory gf = new SimpleGroupFactory(schema);
+    Group g1 = gf.newGroup();
+    g1.addGroup("foo").append("bar?@", 2l);
+
+    testSchema(schema, Arrays.asList(g1));
```
julienledem: but did that fail before without quotes?

parquet-column/src/test/java/org/apache/parquet/parser/TestParquetParser.java
```diff
@@ -310,4 +310,26 @@ public void testEmbeddedAnnotations() {
...
+        .required(INT32).as(INT_8).named("foo bar")
+        .required(INT32).as(INT_16).named("foo.bar")
+        .required(INT32).as(INT_32).named("foo`bar")
+        .required(INT64).as(INT_64).named("foo@#$bar")
+        .named("IntMessage");
+
+    assertEquals(expected, parsed);
+    MessageType reparsed = MessageTypeParser.parseMessageType(parsed.toString());
+    assertEquals(expected, reparsed);
+  }
```
julienledem: we should add checks that column names that worked before do not get quoted when generating the schema now.
Otherwise older versions of parquet won't be able to read those files.

parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/ColumnPath.java
```diff
@@ -19,15 +19,21 @@
...
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 import org.apache.parquet.Strings;
 
 import static org.apache.parquet.Preconditions.checkNotNull;
 
 public final class ColumnPath implements Iterable<String>, Serializable {
 
+  private static Pattern pattern = Pattern.compile("[^a-zA-Z\\d_-]");
```
julienledem: why do we need to define this pattern again?
We should define it once.

parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/ColumnPath.java
```diff
@@ -41,7 +47,14 @@ protected ColumnPath toCanonical(ColumnPath value) {
 
   public static ColumnPath fromDotString(String path) {
     checkNotNull(path, "path");
-    return get(path.split("\\."));
+    String[] parts = path.split("\\.(?=([^`]*`[^`]*`)*[^`]*$)");
```
julienledem: I think we should centralize the quote parsing logic.
This is the 3rd place where we unquote/unescape quoted names.

parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/ColumnPath.java
```diff
@@ -41,7 +47,14 @@ protected ColumnPath toCanonical(ColumnPath value) {
...
+    for (int i = 0; i < parts.length; ++i) {
+      String part = parts[i];
+      if (part.startsWith("`") && part.endsWith("`")) {
+        parts[i] = part.substring(1, part.length() - 1);
```
julienledem: unescaping "``" ?

parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/ColumnPath.java
```diff
@@ -54,6 +67,14 @@ private ColumnPath(String[] path) {
     this.p = path;
   }
 
+  /**
+   * @return the name of the type, possibly wrapped in backquotes.
+   */
+  private String getQuotedName(String name) {
+    Matcher matcher = pattern.matcher(name);
+    return matcher.find() ? "`" + name.replace("`", "``") + "`" : name;
+  }
```
julienledem: duplicated with schema

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
julienledem: and also tests that checks some names with special characters that worked before don't get quoted.
For backward compatibility as older versions of Parquet won't understand quotes.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
rdblue: What requirement are you referring to?

My main concern is that if we allow special characters in column names we will allow creating Parquet files that are incompatible with most object models. Thrift and protobuf can't generate a method that contains "`" and Avro's spec disallows special characters specifically for compatibility across languages and uses (code gen, generic, etc.).

I don't think it's a good idea to add this support. The alternative is for the user to choose a column name in Hive that isn't going to break others systems, and I think that's a completely reasonable requirement.

julienledem:
@ptkool did you want to update this?
I still think this would be a useful addition.
@rdblue I think it is compatible as long as we don't quote names that did not need quoting before.

rdblue:
@julienledem, the problem is that this allows names in Parquet files that will break some object models.

julienledem:
@rdblue But typically object model integration (Avro, Thrift, Protobuf) is used to import data into Parquet or read data imported in the same model. Since people have existing schemas, it makes sense to me to allow escaping special characters rather than forcing renaming individual fields which is painful and brittle. Possibly we could add a flag on write to enable this. That would force users to acknowledge that they may not be able to read this with Avro (for example). But even then something could be done at that layer to convert to an accepted field name.
There are probably a few places that rely on names not containing "." for example for path representation that would need to be fixed similarly.

ptkool:
@julienledem Yes, I will update this.

rdblue:
I'm still -1 on this because it is allowing names that we know will break object models. I wouldn't say that Thrift, Protobuf, and Avro are only used to import data that will be read by other models after that. Avro originally only read files where there was an Avro schema, but we quickly had requests to be able to read non-Avro files (those written by Impala) with Avro.

julienledem:
@rdblue I agree with your statement that we should not restrict reading a parquet file to only the model that wrote it however I don't think that this PR forces that.
Each model has different rules about what is a valid name and what is not. By restricting the names in a parquet schema we force mangling them while writing as well as when reading them back. Often schemas already exist with the name as is. 

To read back from a model that has different naming restrictions I think we should deal with that on read by having a standard rule per model on how to transform a name into one that is valid in that model on read. All we need is a way to map that name on read. 

There are already valid Parquet field that are considered keywords in thrift for example.

ptkool:
@julienledem Any more thoughts on this?

MatthiasEgli:
This is currently blocking us from using Parquet as we are having arbitrary field names in our (so far json-backed) schema. While it is possible to escape/translate them to conform to Parquet limitations, it would be way preferable if Parquet could use arbitrary field names.

fpompermaier:
Any news about this PR? It would be very useful IMHO

airshad:
Any update on this issue? I am facing this issue in one of my project with parquet format and its pissing me off. I don't want to alter the dataset by replacing space with `_` OR removing it altogether.

cwensel:
I'm a +1 on this, but I guess it's a few years too late.

I think parity with how SQL is used in the wild should be more of a priority.

It's really unfortunate this stalled.

julienledem:
If someone wants to push this to the finish line that would be nice. There's still a little bit of consensus building needed. 
What would the final solution look like?

PR #366, opened 2016-09-08
PARQUET-623: Fix DeltaByteArrayReader#skip.

rdblue:
Previously, this passed the skip to the underlying readers, but would
not update previous and would corrupt values or cause exceptions.


julienledem:
+1 LGTM

PR #367, opened 2016-09-09
PARQUET-686: Do not return min/max for the wrong order.

rdblue:
Min and max are currently calculated using the default Java ordering
that uses signed comparison for all values. This is not correct for
binary types like strings and decimals or for unsigned numeric types.
This commit prevents statistics accumulated using the signed ordering
from being returned by ParquetMetadataConverter when the type should use
the unsigned ordering.

Because many binary strings are not affected by using the wrong
ordering, this adds a property, parquet.strings.use-signed-order to
allow overriding this change.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -36,6 +36,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.CorruptStatistics;
 import org.apache.parquet.Log;
 import org.apache.parquet.format.PageEncodingStats;
@@ -80,6 +81,16 @@ public class ParquetMetadataConverter {
   public static final long MAX_STATS_SIZE = 4096; // limit stats to 4k
 
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
+
+  private final boolean useSignedStringMinMax;
+
+  public ParquetMetadataConverter() {
+    this.useSignedStringMinMax = false;
+  }
+
+  public ParquetMetadataConverter(Configuration conf) {
+    this.useSignedStringMinMax = conf.getBoolean("parquet.strings.use-signed-order", false);
+  }
 
   // NOTE: this cache is for memory savings, not cpu savings, and is used to de-duplicate
   // sets of encodings. It is important that all collections inserted to this cache be
@@ -308,20 +319,117 @@ public class ParquetMetadataConverter {
     return fromParquetStatistics(null, statistics, type);
   }
 
+  /**
+   * @deprecated Use {@link #fromParquetStatistics(String, Statistics, PrimitiveType)} instead.
+   */
+  @Deprecated
   public static org.apache.parquet.column.statistics.Statistics fromParquetStatistics
       (String createdBy, Statistics statistics, PrimitiveTypeName type) {
+    return fromParquetStatisticsInternal(createdBy, statistics, type, defaultSortOrder(type));
+  }
+
+  // Visible for testing
+  static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal
+      (String createdBy, Statistics statistics, PrimitiveTypeName type, SortOrder expectedOrder) {
     // create stats object based on the column type
     org.apache.parquet.column.statistics.Statistics stats = org.apache.parquet.column.statistics.Statistics.getStatsBasedOnType(type);
     // If there was no statistics written to the footer, create an empty Statistics object and return
 
     // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
-    if (statistics != null && !CorruptStatistics.shouldIgnoreStatistics(createdBy, type)) {
+    if (statistics != null && !CorruptStatistics.shouldIgnoreStatistics(createdBy, type) &&
+        SortOrder.SIGNED == expectedOrder) {
       if (statistics.isSetMax() && statistics.isSetMin()) {
         stats.setMinMaxFromBytes(statistics.min.array(), statistics.max.array());
       }
       stats.setNumNulls(statistics.null_count);
     }
     return stats;
+  }
+
+  public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
+      String createdBy, Statistics statistics, PrimitiveType type) {
+    SortOrder expectedOrder = isSignedOrderOkay(type) ? SortOrder.SIGNED : sortOrder(type);
+    return fromParquetStatisticsInternal(
+        createdBy, statistics, type.getPrimitiveTypeName(), expectedOrder);
+  }
+
+  enum SortOrder {
+    SIGNED,
+    UNSIGNED,
+    UNKNOWN
+  }
+
+  private boolean isSignedOrderOkay(PrimitiveType type) {
+    if (!useSignedStringMinMax) {
+      return false;
+    }
+
+    // even if the override is set, only return stats for string-ish types
+    if (type.getPrimitiveTypeName() != PrimitiveTypeName.BINARY) {
+      return false;
+    }
+    if (type.getOriginalType() == null) {
+      return true; // plain binary is okay
+    }
+    switch (type.getOriginalType()) {
+      case UTF8:
+      case ENUM:
+      case BSON:
+      case JSON:
+        return true;
+      default: // includes decimal
+        return false;
+    }
+  }
+
+  private static SortOrder defaultSortOrder(PrimitiveTypeName primitive) {
+    switch (primitive) {
+      case BOOLEAN:
+      case INT32:
+      case INT64:
+      case FLOAT:
+      case DOUBLE:
+      case BINARY: // without a logical type, signed is okay
+      case FIXED_LEN_BYTE_ARRAY:
+        return SortOrder.SIGNED;
+      case INT96: // only used for timestamp, which uses unsigned values
+        return SortOrder.UNSIGNED;
+    }
+    return SortOrder.UNKNOWN;
+  }
+
+  private static SortOrder sortOrder(PrimitiveType primitive) {
+    OriginalType annotation = primitive.getOriginalType();
+    if (annotation != null) {
+      switch (annotation) {
+        case INT_8:
+        case INT_16:
+        case INT_32:
+        case INT_64:
+        case DATE:
+        case TIME_MICROS:
+        case TIME_MILLIS:
+        case TIMESTAMP_MICROS:
+        case TIMESTAMP_MILLIS:
+          return SortOrder.SIGNED;
+        case UINT_8:
+        case UINT_16:
+        case UINT_32:
+        case UINT_64:
+        case DECIMAL:
+        case ENUM:
+        case UTF8:
+        case BSON:
+        case JSON:
+          return SortOrder.UNSIGNED;
+        case LIST:
+        case MAP:
+        case MAP_KEY_VALUE:
+        case INTERVAL:
+          return SortOrder.UNKNOWN;
+      }
+    }
+    return defaultSortOrder(primitive.getPrimitiveTypeName());
   }
 
   public PrimitiveTypeName getPrimitive(Type type) {
@@ -687,7 +795,7 @@ public class ParquetMetadataConverter {
               fromParquetStatistics(
                   parquetMetadata.getCreated_by(),
                   metaData.statistics,
-                  messageType.getType(path.toArray()).asPrimitiveType().getPrimitiveTypeName()),
+                  messageType.getType(path.toArray()).asPrimitiveType()),
               metaData.data_page_offset,
               metaData.dictionary_page_offset,
               metaData.num_values,
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -85,11 +85,15 @@ public class ParquetMetadataConverter {
   private final boolean useSignedStringMinMax;
 
   public ParquetMetadataConverter() {
-    this.useSignedStringMinMax = false;
+    this(false);
   }
 
   public ParquetMetadataConverter(Configuration conf) {
-    this.useSignedStringMinMax = conf.getBoolean("parquet.strings.use-signed-order", false);
+    this(conf.getBoolean("parquet.strings.signed-min-max.enabled", false));
+  }
+
+  private ParquetMetadataConverter(boolean useSignedStringMinMax) {
+    this.useSignedStringMinMax = useSignedStringMinMax;
   }
 
   // NOTE: this cache is for memory savings, not cpu savings, and is used to de-duplicate
@@ -330,14 +334,17 @@ public class ParquetMetadataConverter {
 
   // Visible for testing
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal
-      (String createdBy, Statistics statistics, PrimitiveTypeName type, SortOrder expectedOrder) {
+      (String createdBy, Statistics statistics, PrimitiveTypeName type, SortOrder typeSortOrder) {
     // create stats object based on the column type
     org.apache.parquet.column.statistics.Statistics stats = org.apache.parquet.column.statistics.Statistics.getStatsBasedOnType(type);
     // If there was no statistics written to the footer, create an empty Statistics object and return
 
     // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
+    // The sort order is checked to avoid returning min/max stats that are not
+    // valid with the type's sort order. Currently, all stats are aggregated
+    // using a signed ordering, which isn't valid for strings or unsigned ints.
     if (statistics != null && !CorruptStatistics.shouldIgnoreStatistics(createdBy, type) &&
-        SortOrder.SIGNED == expectedOrder) {
+        SortOrder.SIGNED == typeSortOrder) {
       if (statistics.isSetMax() && statistics.isSetMin()) {
         stats.setMinMaxFromBytes(statistics.min.array(), statistics.max.array());
       }
@@ -348,40 +355,56 @@ public class ParquetMetadataConverter {
 
   public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
       String createdBy, Statistics statistics, PrimitiveType type) {
-    SortOrder expectedOrder = isSignedOrderOkay(type) ? SortOrder.SIGNED : sortOrder(type);
+    SortOrder expectedOrder = overrideSortOrderToSigned(type) ?
+        SortOrder.SIGNED : sortOrder(type);
     return fromParquetStatisticsInternal(
         createdBy, statistics, type.getPrimitiveTypeName(), expectedOrder);
   }
 
+  /**
+   * Sort order for page and column statistics. Types are associated with sort
+   * orders (e.g., UTF8 columns should use UNSIGNED) and column stats are
+   * aggregated using a sort order. As of parquet-format version 2.3.1, the
+   * order used to aggregate stats is always SIGNED and is not stored in the
+   * Parquet file. These stats are discarded for types that need unsigned.
+   *
+   * See PARQUET-686.
+   */
   enum SortOrder {
     SIGNED,
     UNSIGNED,
     UNKNOWN
   }
 
-  private boolean isSignedOrderOkay(PrimitiveType type) {
-    if (!useSignedStringMinMax) {
-      return false;
-    }
+  private static final Set<OriginalType> STRING_TYPES = Collections
+      .unmodifiableSet(new HashSet<>(Arrays.asList(
+          OriginalType.UTF8, OriginalType.ENUM, OriginalType.JSON
+      )));
 
+  /**
+   * Returns whether to use signed order min and max with a type. It is safe to
+   * use signed min and max when the type is a string type and contains only
+   * ASCII characters (where the sign bit was 0). This checks whether the type
+   * is a string type and uses {@code useSignedStringMinMax} to determine if
+   * only ASCII characters were written.
+   *
+   * @param type a primitive type with a logical type annotation
+   * @return true if signed order min/max can be used with this type
+   */
+  private boolean overrideSortOrderToSigned(PrimitiveType type) {
     // even if the override is set, only return stats for string-ish types
-    if (type.getPrimitiveTypeName() != PrimitiveTypeName.BINARY) {
-      return false;
-    }
-    if (type.getOriginalType() == null) {
-      return true; // plain binary is okay
-    }
-    switch (type.getOriginalType()) {
-      case UTF8:
-      case ENUM:
-      case BSON:
-      case JSON:
-        return true;
-      default: // includes decimal
-        return false;
-    }
+    // a null type annotation is considered string-ish because some writers
+    // failed to use the UTF8 annotation.
+    OriginalType annotation = type.getOriginalType();
+    return useSignedStringMinMax &&
+        PrimitiveTypeName.BINARY == type.getPrimitiveTypeName() &&
+        (annotation == null || STRING_TYPES.contains(annotation));
   }
 
+  /**
+   * @param primitive a primitive physical type
+   * @return the default sort order used when the logical type is not known
+   */
   private static SortOrder defaultSortOrder(PrimitiveTypeName primitive) {
     switch (primitive) {
       case BOOLEAN:
@@ -389,15 +412,19 @@ public class ParquetMetadataConverter {
       case INT64:
       case FLOAT:
       case DOUBLE:
-      case BINARY: // without a logical type, signed is okay
-      case FIXED_LEN_BYTE_ARRAY:
         return SortOrder.SIGNED;
+      case BINARY:
+      case FIXED_LEN_BYTE_ARRAY:
       case INT96: // only used for timestamp, which uses unsigned values
         return SortOrder.UNSIGNED;
     }
     return SortOrder.UNKNOWN;
   }
 
+  /**
+   * @param primitive a primitive type with a logical type annotation
+   * @return the "correct" sort order of the type that applications assume
+   */
   private static SortOrder sortOrder(PrimitiveType primitive) {
     OriginalType annotation = primitive.getOriginalType();
     if (annotation != null) {
@@ -416,12 +443,12 @@ public class ParquetMetadataConverter {
         case UINT_16:
         case UINT_32:
         case UINT_64:
-        case DECIMAL:
         case ENUM:
         case UTF8:
         case BSON:
         case JSON:
           return SortOrder.UNSIGNED;
+        case DECIMAL:
         case LIST:
         case MAP:
         case MAP_KEY_VALUE:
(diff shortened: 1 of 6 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -450,6 +450,19 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
...
+  /**
+   * Reads the meta data block in the footer of the file using provided input stream
+   * @param fileLen length of the file
+   * @param filePath file location
+   * @param f input stream for the file
+   * @param filter the filter to apply to row groups
+   * @return the metadata blocks in the footer
+   * @throws IOException if an error occurs while reading the file
+   */
+  private static final ParquetMetadata readFooter(ParquetMetadataConverter converter, long fileLen, String filePath, SeekableInputStream f, MetadataFilter filter) throws IOException {
```
andreweduffy: the new `converter` param seems to not be used?

rdblue: It is, on line 493. The name is the same as the previous version's static converter so that line didn't change at all.

andreweduffy:
What's the upgrade path for the future, i.e. if you want to use unsigned statistics for Binary columns? This doesn't seem to address that, just disables returning statistics for those cases.

rdblue:
This doesn't address fixing the problem and writing correct stats because that will rely on deciding how to store the order in parquet-format and how we want to expose different orderings to users. The purpose of this patch isn't to decide those, it is to fix the correctness issue by default for non-ascii characters and to provide a way to opt in to use the current behavior.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -81,6 +82,16 @@
 
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
 
+  private final boolean useSignedStringMinMax;
+
+  public ParquetMetadataConverter() {
+    this.useSignedStringMinMax = false;
```
julienledem: let's not have a default here.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -81,6 +82,16 @@
...
+  }
+
+  public ParquetMetadataConverter(Configuration conf) {
+    this.useSignedStringMinMax = conf.getBoolean("parquet.strings.use-signed-order", false);
```
julienledem: let's have ParquetMetadataConverter(boolean useSignedStringMinMax) instead and pull up reading the configuration with the other config parameters

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -308,14 +319,25 @@ public static Statistics toParquetStatistics(
...
+  static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal
+      (String createdBy, Statistics statistics, PrimitiveTypeName type, SortOrder expectedOrder) {
     // create stats object based on the column type
     org.apache.parquet.column.statistics.Statistics stats = org.apache.parquet.column.statistics.Statistics.getStatsBasedOnType(type);
     // If there was no statistics written to the footer, create an empty Statistics object and return
 
     // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
-    if (statistics != null && !CorruptStatistics.shouldIgnoreStatistics(createdBy, type)) {
+    if (statistics != null && !CorruptStatistics.shouldIgnoreStatistics(createdBy, type) &&
+        SortOrder.SIGNED == expectedOrder) {
```
julienledem: shouldn't we move the SortOrder check to CorruptStatistics.shouldIgnoreStatistics ?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +346,92 @@ public static Statistics toParquetStatistics(
     return stats;
   }
 
+  public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
+      String createdBy, Statistics statistics, PrimitiveType type) {
+    SortOrder expectedOrder = isSignedOrderOkay(type) ? SortOrder.SIGNED : sortOrder(type);
```
julienledem: I'd move the checks to shouldIgnoreStatistics.
It seems there are more steps than necessary to decide whether the sort order is OK.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -450,6 +450,19 @@ public static final ParquetMetadata readFooter(Configuration configuration, File
    * @throws IOException if an error occurs while reading the file
    */
   public static final ParquetMetadata readFooter(long fileLen, String filePath, SeekableInputStream f, MetadataFilter filter) throws IOException {
+    return readFooter(new ParquetMetadataConverter(), fileLen, filePath, f, filter);
```
julienledem: Since this method has not been released yet, you can change its signature.
How about introducing a ParquetReadConfig class that would have the field useSignedStringMinMax (or a better name)?
That way we don't add an unnecessary dependency on hadoop.Configuration.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -771,7 +787,7 @@ DictionaryPage readDictionary(ColumnChunkMetaData meta) throws IOException {
         compressedPage.getEncoding());
   }
```
julienledem: we should open a JIRA to factor out those constructor and clean up the deprecated ones.
lots of duplication here.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -452,6 +452,7 @@ public void testWriteReadStatistics() throws Exception {
 
     Path path = new Path(testFile.toURI());
     Configuration configuration = new Configuration();
+    configuration.set("parquet.strings.use-signed-order", "true");
```
julienledem: please add a ParquetInputFormat static method to add this setting.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -81,6 +82,16 @@
 
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
 
+  private final boolean useSignedStringMinMax;
```
julienledem: how about: useSignedMinMaxForUnsignedValues

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +346,92 @@ public static Statistics toParquetStatistics(
...
+      case ENUM:
+      case BSON:
+      case JSON:
+        return true;
+      default: // includes decimal
+        return false;
+    }
+  }
+
+  private static SortOrder defaultSortOrder(PrimitiveTypeName primitive) {
```
julienledem: please add javadoc: "used when the type has no OriginalType annotation"

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +346,92 @@ public static Statistics toParquetStatistics(
...
+        case TIME_MICROS:
+        case TIME_MILLIS:
+        case TIMESTAMP_MICROS:
+        case TIMESTAMP_MILLIS:
+          return SortOrder.SIGNED;
+        case UINT_8:
+        case UINT_16:
+        case UINT_32:
+        case UINT_64:
+        case DECIMAL:
```
julienledem: Decimal is unsigned?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +346,92 @@ public static Statistics toParquetStatistics(
...
+    switch (type.getOriginalType()) {
+      case UTF8:
+      case ENUM:
+      case BSON:
+      case JSON:
+        return true;
+      default: // includes decimal
+        return false;
+    }
+  }
```
julienledem: Why not simply:
```
private boolean isMinMaxUsableFor(PrimitiveType type) {
 SortOrder expected = sortOrder(type);
 return (expected == SIGNED) || (expected == UNSIGNED && useSignedMinMaxForUnsigned);
}
```

the case for getOriginalType is already treated in sortOrder(). Why adding another layer here?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +346,92 @@ public static Statistics toParquetStatistics(
...
+      case BINARY: // without a logical type, signed is okay
+      case FIXED_LEN_BYTE_ARRAY:
+        return SortOrder.SIGNED;
+      case INT96: // only used for timestamp, which uses unsigned values
+        return SortOrder.UNSIGNED;
+    }
+    return SortOrder.UNKNOWN;
+  }
+
+  private static SortOrder sortOrder(PrimitiveType primitive) {
```
julienledem: s/sortOrder/expectedSortOrder ?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
julienledem: I'd call this setting ```parquet.read.use-signed-order-for-unsigned-types.enabled```
see: https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: This is a short-lived change and the order used to produce the min and max should be passed along with the stats in the next release. I don't think it makes sense to pass more type info, the actual order, and whether to override the order into the `shouldIgnoreStatistics` method (which is public) only to change it next release, so I'd like to keep this here to minimize the changes.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: See my comment below.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
rdblue: Done.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: I changed the property to `parquet.strings.signed-min-max.enabled` and moved it to `ParquetInputFormat`. The different name is because it only affects strings.

I initially changed the parameter to a boolean as you suggested, but when we move to a better stats representation and pass the order used to produce min/max along with stats, we won't need this any more. It will be a useless part of the public API, but we may have another situation where we need a Configuration in the future, so I think it makes more sense to go with this version.

rdblue:
@julienledem, I made changes and addressed your comments. Could you have another look? Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
julienledem: It is already prevented from being used by isSignedOrderOkay() returning false when the OriginalType is DECIMAL.

Your answer sounds like "it depends". Then we should make it UNKNOWN for now.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
julienledem: ok

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
julienledem: I don't see why we restrict it further since the flag is already about lifting the restriction but that's fine with me if you feel we need this.

couple more comments:
- The name of the method could be clearer. What it actually does is controlling if we're going to override the result of sortOrder().

- The way the logic is organized here is a little hard to read.
If you define the accepted String types:
```
private static final Set<OriginalType> acceptedStringTypes = Collections.unmodifiableSet(new HashSet(Arrays.asList(
  UTF8, ENUM, JSON
)));
```
then it boils down to a boolean expression that I find easier to understand:
```
boolean overrideSortOrderToUnsigned(PrimitiveType type) {
  return useSignedStringMinMax 
    && (type.getPrimitiveTypeName() == BINARY) 
    && (type.getOriginalType() == null || acceptedStringTypes.contains(type.getOriginalType());
}
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
julienledem: We should not pass hadoop configurations around. see my comments on #368.

Regarding the config I was just suggesting using it in this particular method and that particular setting. Since it has not been released yet.
Once this method is released changing its signature will require deprecating it.
But adding fields to the ParquetReadConfig object will be backwards compatible.

the parameterless ParquetMetadataConverter() should be deprecated

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -81,6 +83,20 @@
 
   private static final Log LOG = Log.getLog(ParquetMetadataConverter.class);
 
+  private final boolean useSignedStringMinMax;
+
+  public ParquetMetadataConverter() {
```
julienledem: If for backward compatibility we should mark this as deprecated.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
julienledem: please add this as a javadoc comment here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -230,6 +233,22 @@ public static Filter getFilter(Configuration conf) {
...
+  public static void enableSignedStringMinMax(Configuration conf) {
+    conf.setBoolean(SIGNED_ORDER_ENABLED, true);
+  }
+
+  /**
+   * Returns whether using signed string min and max is enabled.
+   */
+  public static boolean isSignedStringMinMaxEnabled(Configuration conf) {
+    return conf.getBoolean(SIGNED_ORDER_ENABLED, SIGNED_ORDER_ENABLED_DEFAULT);
+  }
```
julienledem: Hey, Sorry,
I know I am the one who asked for those methods here but as @isnotinvain mentioned on https://github.com/apache/parquet-mr/pull/362 we don't want to pollute the api with too many settings that are of limited use.
How about we removing these methods from there and use the string in the test like you had before?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Good point.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
rdblue: Agreed.

rdblue:
@julienledem, I updated this PR for your comments other than the `ParquetMetadataConverter` constructor. I don't think it's a good idea to create a constructor that takes the `useSignedStringMinMax` parameter because it is going to no longer be needed by the metadata converter when we start passing order along with stats. We need to pass it now, which means we can either pass the `Configuration` or something else that doesn't depend on Hadoop.

I think we should pass the `Configuration`. It is beyond the scope of this PR to introduce a `ParquetReadConfig` object to replace `Configuration`. When we do that, I want to make sure we design it well rather than doing it quickly to get this PR done. We already pass a Hadoop conf in a lot of places so I don't think it is worth avoiding here when we will have to remove it elsewhere. This class is still in the parquet-hadoop package anyway, so we're a long way from being able to remove the Hadoop conf.

julienledem:
+1 LGTM
@isnotinvain: It would be great if you could review this.

rdblue:
I agree, it would be great if @isnotinvain could have a look.

In the interest of getting a release out, I'd like to commit this in the next day or two unless there are other comments. Thanks for reviewing, @julienledem and @andreweduffy!

isnotinvain:
I will take a look. I was out of town for a bit, what did we end up deciding in the discussion in parquet-format?

isnotinvain:
Basically we won't use the old min/max for <, <=, >, >= checks? But we'll still use it for == and !=?

rdblue:
@isnotinvain, this commit will prevent signed min/max from being passed up from file metadata unless you have a property set to do so for strings. This only fixes correctness. In the next release, I think the right thing to do is to pass up the stats along with the ordering that produced them, so that we can implement == and != like you're talking about.

isnotinvain:
OK, thanks for clarifying. The property that controls whether to pass these up for strings -- is it only for strings, or for all binaries? Is there a reason to treat strings and binaries separately?

rdblue:
The property is for binary without a logical type (Impala doesn't write strings annotated with UTF8 :disappointed:), and for UTF8, JSON, and ENUMs. That excludes DECIMAL, for which the min/max are always invalid, and unsigned ints that have the same problem.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
isnotinvain: Given that this controls all binaries except enum/unsigned int, can we rename this something like useSignedBinaryMinMax ?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
isnotinvain: +1

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -81,6 +82,20 @@
...
+    this(false);
+  }
+
+  public ParquetMetadataConverter(Configuration conf) {
+    this(conf.getBoolean("parquet.strings.signed-min-max.enabled", false));
```
isnotinvain: can this key live in a public static constant? and again here, maybe use 'binary' instead of 'string' ?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +350,103 @@ public static Statistics toParquetStatistics(
...
+   * Returns whether to use signed order min and max with a type. It is safe to
+   * use signed min and max when the type is a string type and contains only
+   * ASCII characters (where the sign bit was 0). This checks whether the type
+   * is a string type and uses {@code useSignedStringMinMax} to determine if
+   * only ASCII characters were written.
+   *
+   * @param type a primitive type with a logical type annotation
+   * @return true if signed order min/max can be used with this type
+   */
+  private boolean overrideSortOrderToSigned(PrimitiveType type) {
```
isnotinvain: this is adding a lot of complication / corner cases. Do we really need to support using signed ordering for ascii only strings? It's especially confusing that we allow this for un-annotated binary columns (we just assume they are string-ish, which doesn't seem valid).

It seems like this could be simplified by simply returning the found min/max and the ordering with which they were written and letting the filter layer decide what to do there? I think you mentioned that maybe that's the plan going forward, if it is, can we mark all these extra methods around dealing with overrides as temporary / to be deleted soon?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +350,103 @@ public static Statistics toParquetStatistics(
...
+    return useSignedStringMinMax &&
+        PrimitiveTypeName.BINARY == type.getPrimitiveTypeName() &&
+        (annotation == null || STRING_TYPES.contains(annotation));
+  }
+
+  /**
+   * @param primitive a primitive physical type
+   * @return the default sort order used when the logical type is not known
+   */
+  private static SortOrder defaultSortOrder(PrimitiveTypeName primitive) {
```
isnotinvain: can you add comments explaining what is meant by default sort order? why are non-banary types involved in this?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: This isn't for binaries, it's for all string representations.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Why should this be deprecated?

It's perfectly valid to create a ParquetMetadataConverter without any configuration. It could easily be that when we restructure this into parquet-common, it's the other constructor that goes away, so I don't think we should consider it deprecated.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: We removed the public constant to avoid leaking this into the public API and not being able to remove it. Multiple packages reference it, so it would need to be public. I think it's fine to use a String for now.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Impala writes Strings as unannotated binary, so we should consider it string-ish if `useSignedStringMinMax` is set.

While passing stats and the sort order should be the approach we take, I think it will require some design and thought to make sure we don't pass invalid stats to applications. If we simply added the sort order to the current stats objects, engines like Presto would keep using the bad stats without checking it. This commit will prevent applications from getting stats and then we can take the time to fix it correctly.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Default sort order for a primitive type is the order we think is appropriate when the logical type isn't known. This is only used for backward-compatibility with the existing static methods to convert stats.

rdblue:
For some reason, not all of my replies are shown above and you have to go to the code tab. Something to do with the new review interface.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -308,14 +323,25 @@ public static Statistics toParquetStatistics(
...
+  }
+
+  // Visible for testing
+  static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal
+      (String createdBy, Statistics statistics, PrimitiveTypeName type, SortOrder typeSortOrder) {
     // create stats object based on the column type
     org.apache.parquet.column.statistics.Statistics stats = org.apache.parquet.column.statistics.Statistics.getStatsBasedOnType(type);
     // If there was no statistics written to the footer, create an empty Statistics object and return
 
     // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
```
piyushnarang: might also want to add a line on the sort order bit.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +350,103 @@ public static Statistics toParquetStatistics(
...
 
+  public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
+      String createdBy, Statistics statistics, PrimitiveType type) {
+    SortOrder expectedOrder = overrideSortOrderToSigned(type) ?
+        SortOrder.SIGNED : sortOrder(type);
+    return fromParquetStatisticsInternal(
+        createdBy, statistics, type.getPrimitiveTypeName(), expectedOrder);
+  }
+
+  enum SortOrder {
```
piyushnarang: add a couple of lines of javadoc to indicate what this struct is for? Also what is the convention we follow for types like enums / nested classes - should they be declared at the top of the class? Or anywhere goes? (persnally prefer the top)

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -324,6 +350,103 @@ public static Statistics toParquetStatistics(
...
+    SIGNED,
+    UNSIGNED,
+    UNKNOWN
+  }
+
+  private static final Set<OriginalType> STRING_TYPES = Collections
```
piyushnarang: move to top?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: I think the field name is pretty clear. What information would you like to see added in a comment?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: The problem is that there is no good place to add this and we don't want to add it to the public API. The long-term fix will very likely move this option outside of the converter (returning the min/max and the sort order used during aggregation), and it isn't something we want to expose in ParquetInputFormat or ParquetReader, or else people will turn it on without understanding what it does. I agree it isn't ideal, but I think this is fine until we have a permanent solution to the problem.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: Done.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: This is internal and will be removed (or moved into common and made public) when we add the long-term fix. That's why I'm keeping all of these classes together. That way the chance of a conflict is minimized.

piyushnarang:
@rdblue thanks for the updates. 👍

julienledem:
+1

PR #368, opened 2016-09-12
PARQUET-674: Add InputFile abstraction for openable files.

rdblue:


rdblue:
@julienledem and @robert3005, we might want to get this in 1.9.0 (though this is a low priority). It adds `ParquetDataSource`, which is what we discussed adding in discussion on PARQUET-400 to encapsulate file size, location, and providing new `SeekableInputStream` instances. I think this would be a bit cleaner for the `readFooters` method added in #357. Currently this only works with Hadoop, but you could easily add an implementation for other file systems.

robert3005:
:+1: That looks like a good abstraction in general. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -53,6 +54,8 @@
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
+import org.apache.commons.math3.analysis.function.Add;
```
piyushnarang: remove import?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.hadoop;
 
+import static org.apache.hadoop.yarn.webapp.hamlet.HamletSpec.InputType.file;
```
piyushnarang: is this the right import?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: Ugh, I hate how IntelliJ now adds static imports automatically. It constantly adds these when I pause typing a name. Thanks for catching it!

parquet-common/src/main/java/org/apache/parquet/io/ParquetDataSource.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * {@code ParquetDataSource} is an interface with the methods needed by Parquet
+ * to read data files using {@link SeekableInputStream} instances.
+ */
+public interface ParquetDataSource {
+
+  /**
+   * Returns the file location.
+   */
+  String getLocation();
```
piyushnarang: should we keep this as Path?

piyushnarang:
👍

parquet-common/src/main/java/org/apache/parquet/io/ParquetDataSource.java
rdblue: No, this is an abstraction that isn't tied to Hadoop or another FS library. A string location should be portable across implementations.

piyushnarang: Yeah I was thinking more on the lines of a java Path / URI.

rdblue:
@julienledem, could you take a look at this? It would be better to do this for 1.9.0 than to do it later because it would prevent exposing a public method. Thanks!

parquet-common/src/main/java/org/apache/parquet/io/ParquetDataSource.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.apache.parquet.io;
+
+import java.io.IOException;
+
+/**
+ * {@code ParquetDataSource} is an interface with the methods needed by Parquet
+ * to read data files using {@link SeekableInputStream} instances.
+ */
+public interface ParquetDataSource {
```
julienledem: It's a SeekableInputStream provider with a length.
maybe call it InputFile ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopDataSource.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.hadoop.conf.Configurable;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.io.SeekableInputStream;
+import org.apache.parquet.io.ParquetDataSource;
+import java.io.IOException;
+
+public class HadoopDataSource implements ParquetDataSource, Configurable {
```
julienledem: Do we need to make it Configurable?
The conf is passed in the constructor and does not need to be settable or exposed.
even better once initialized, the conf is not used anymore. I would remove the conf field as well.

rdblue:
@julienledem, thanks for your comments. I implemented your suggestions so I think this is about ready when tests are passing.

julienledem:
+1

PR #370, opened 2016-09-23
PARQUET-726: Increase max difference of testMemoryManagerUpperLimit to 10%

nielsbasjes:


nielsbasjes:
For this change I simply chose to not change the way the test was already done and simply add a little extra info if it fails. I can remove the extra info if you want.

nielsbasjes:
I tried the method you mentioned and it doesn't improve the readability of the error message:
`Failed tests:   testMemoryManagerUpperLimit(org.apache.parquet.hadoop.TestMemoryManager): Pool size should be within 5.0% of the expected value expected:<4.53745044E8> but was:<4.79644871E8>
`

julienledem:
better error messages is great. 
Hopefully this will fix the test flakiness.

julienledem:
+1

PR #371, opened 2016-09-23
PARQUET-727: Ensure correct version of thrift is used

nielsbasjes:
This will make the build fail if the wrong version of thrift is used before building the generated sources fails.

### DIFF
--- a/parquet-thrift/pom.xml
+++ b/parquet-thrift/pom.xml
@@ -142,34 +142,6 @@
       </plugin>
       <!-- thrift -->
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>exec-maven-plugin</artifactId>
-        <version>1.2.1</version>
-        <executions>
-          <execution>
-            <id>check-thrift-version</id>
-            <phase>generate-sources</phase>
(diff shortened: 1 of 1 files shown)


nielsbasjes:
I essentially copied this code fragment from HBase where I implemented the same check a while ago.

I agree that creating a dependency on a specific OS is not a good idea, yet looking at the systems developers actually use I see that "everyone" uses either Linux or MacOS X as their platform.
People who really want to work on Windows can always start a VM/Docker with Linux to build it.
I recently heard that Windows 10 now comes with Ubuntu as a 'ready to run' environment, so even for those developers this shouldn't be a problem at all. 

So my personal opinion is that assuming a *nix build environment for software developers is fine.

julienledem:
this looks good to me.
Possibly make it easy to turn off for windows people?
See example here:
https://github.com/apache/arrow/blob/7fb4d24a35269db99fa112c0512d4a32c372dd74/java/format/pom.xml#L77

nielsbasjes:
Ok, how about I put this in a profile that activates if not Windows?

julienledem:
+1

PR #372, opened 2016-10-04
PARQUET-685 - Deprecated ParquetInputSplit constructor passes paramet…

gszadovszky:
The problem was not discovered because the test was bugous. Updated both sides.


piyushnarang:
👍

julienledem:
+1

julienledem:
@rdblue opinion on this? Is this used by Hive?

rdblue:
+1

I don't think this is used by Hive, and even if it were it's still a bug that should be fixed.

PR #373, opened 2016-10-06
PARQUET-740: Introduce editorconfig

nielsbasjes:
I examined some of the code and as far as I can tell this specification matches what has been done in this project so far. Please verify.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,43 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,43 @@
...
+indent_size = 2
+
+[*.yaml]
+indent_size = 2
+
+[*.xml]
+indent_size = 2
+
+# Tab indentation (no size specified)
+[Makefile]
```
julienledem: Makefile for?

julienledem:
+1

.editorconfig
nielsbasjes: Because I copied this file from an other project I wrote.
Remove it? Or leave it as it is now?

julienledem: that's fine

PR #376, opened 2016-10-11
PARQUET-743: Fix DictionaryFilter when compressed dictionaries are reused.

rdblue:
BytesInput is not supposed to be held and reused, but decompressed
dictionary pages do this. Reusing the dictionary will cause a failure,
so the cleanest option is to keep the bytes around once the underlying
stream has been read.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryPageReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryPageReader.java
@@ -19,6 +19,7 @@
 package org.apache.parquet.hadoop;
 
 import org.apache.parquet.Strings;
+import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.EncodingStats;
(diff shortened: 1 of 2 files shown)


parquet-encoding/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -247,7 +248,10 @@ public void writeAllTo(OutputStream out) throws IOException {
 
     public byte[] toByteArray() throws IOException {
       if (DEBUG) LOG.debug("read all "+ byteCount + " bytes");
-      byte[] buf = new byte[byteCount];
+      if (buf != null) {
+        return buf;
+      }
+      this.buf = new byte[byteCount];
```
julienledem: I'd rather we treat BytesInput as not reusable.
See writeAllTo() above line 243.
This is one of the remaining copies that we should get rid of.
This change forces to copy the content on heap.

rdblue: I removed this change and implemented it another way. Instead of caching this copy, I make a reusable dictionary to cache instead. I also updated the tests to exercise the path that causes this error (using compression).

julienledem:
+1

PR #378, opened 2016-10-14
PARQUET-750: Parquet-tools cat/head parity, small fixes

sadikovi:
This PR updates error message in `Main.java` to show exception class as it is easier to understand the problem. Also adds json option to `head` and fixes problem when `cat` and `head` (with json option enabled) could not read directory with parquet files.

Tested manually on Spark output.


sadikovi:
Looks like `HeadCommand` is essentially `CatCommand` with limit, no processing differences - is it worth factoring out code from those commands?

sadikovi:
@rdblue @julienledem could you review this PR, please? Thanks.

julienledem:
LGTM.
It would be great if we had tests for this.

sadikovi:
@julienledem thanks! will add tests.

sadikovi:
@julienledem could you review PR, please? I added tests for head/cat commands, but I also had to add sample Parquet files to use in tests. Is that acceptable to keep table as a resource?

julienledem:
@sadikovi: I'd rather we generate the parquet files in the tests rather than check them in.
In general I'd prefer no binary files checked in for tests.
Here's an example: https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestReadWriteEncodingStats.java

sadikovi:
@julienledem I think it is better to reopen PR, so no binaries are checked in.

sadikovi:
@julienledem I updated PR, can you have a look?

parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
```diff
@@ -162,7 +162,8 @@ public static void die(String message, boolean usage, String name, Command comma
   }
 
   public static void die(Throwable th, boolean usage, String name, Command command) {
-    die(th.getMessage(), usage, name, command);
+    String message = th.getClass().getSimpleName() + ": " + th.getMessage();
```
julienledem: th.toString() ?

julienledem:
also check out: https://github.com/rdblue/parquet-cli

parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
sadikovi: I added name + message to make clear it is a failure in cases when exception's `toString()` is overwritten to return just message.

sadikovi:
@julienledem thank you for the link! Looks like I did not do proper research before submitting patch. But it seems like `parquet-cli` does not support yet reading from directory of Parquet files, which was the main driver for this patch, though I might be misreading code in repository.

Do you want me to close this PR and link to `parquet-cli` in README?

rdblue:
@sadikovi, I'm planning to contribute parquet-cli to the Apache project. I just need to get the licensing straightened out and make the build maven-based. No need to link to the external project, but you may want to use it in the mean time. Feel free to contribute patches and those will make it in.

sadikovi:
@rdblue thanks! I am closing this PR then. Thanks guys for review!

julienledem:
@sadikovi @rdblue we can merge this PR to parquet-tools independently of the new cli.

sadikovi:
It's fine, I will open similar patch against parquet-cli, if necessary:)

PR #379, opened 2016-10-14
PARQUET-751: Add setRequestedSchema to ParquetFileReader.

rdblue:
This fixes a bug introduced by dictionary filters, which reused an
existing file reader to avoid opening multiple input streams. Before
that commit, a new file reader was opened and passed the projection
columns from the read context. The fix is to set the requested schema on
the file reader instead of creating a new instance.

This also adds a test to ensure that column projection works to catch
bugs like this in the future.


rdblue:
@danielcweeks, could you review this? Thanks!

danielcweeks:
LGTM, but there are tests failing and it looks like the one you're adding.

Running org.apache.parquet.hadoop.TestInputFormatColumnProjection

PR #380, opened 2016-10-17
PARQUET-753: Fixed GroupType.union() to handle original type

adeneche:
also fixed GroupType.equals() to compare the original type and 2 unit tests that weren't setting the original type properly on the expected results

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
@@ -1,4 +1,4 @@
-/*
+/* 
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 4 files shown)


julienledem:
LGTM. Please open a Parquet JIRA

adeneche:
opened PARQUET-753

liancheng:
+1

julienledem:
@rdblue is this one ok to merge or do you prefer to wait after the release?

rdblue:
I think it's fine

parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -1,14 +1,14 @@
-/*
```
adeneche: my IDE is probably messing with those, I will fix this

PR #381, opened 2016-10-26
PARQUET-755: create parquet-arrow module with schema converter

julienledem:


wesm:
Looks pretty good to me. We haven't dealt with all nested types in parquet-cpp yet, so this will be helpful when we get to that.

wesm:
+1 to merge, if no one else has feedback

PR #382, opened 2016-11-05
PARQUET-765 - Upgrade Avro to 1.8.1

gszadovszky:
Escaping the field name `date` to make the test compatible to Avro 1.8.1 as _date_ is a keyword in avdl from 1.8.1.

Tested by setting the maven parameter `avro.version` to `1.8.1` and initiated a full build by `mvn clean install`.

### DIFF
--- a/parquet-avro/src/test/resources/car.avdl
+++ b/parquet-avro/src/test/resources/car.avdl
@@ -21,7 +21,7 @@
 protocol Cars {
 
     record Service {
-        long date;
+        long `date`;
         string mechanic;
     }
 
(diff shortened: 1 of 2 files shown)


julienledem:
LGTM.
How do we decide what version of avro we want in the pom?
should we set the version to 1.8.1?

gszadovszky:
I don't know if this issue was stopped us using 1.8.1 or just forgot to upgrade to the latest one. 
As this is a minor release upgrade I would don't see any reason why we cannot set it to 1.8.1.
Do you agree on increasing the avro version in this commit?

julienledem:
sounds good to me

gszadovszky:
Changed the avro version to 1.8.1. Modified the title of the JIRA.

PR #383, opened 2016-11-07
PARQUET-768: Add Uwe L. Korn to KEYS

xhochy:


xhochy:
Under the assumption I can propose releases as a committer (that's how I interpret the Apache release process). Correct me if I'm wrong ;)

julienledem:
That's correct.
You can also update the file there: https://dist.apache.org/repos/dist/release/parquet/KEYS
+1

PR #384, opened 2016-11-13
PARQUET-777: Add Parquet CLI.

rdblue:
This adds a new parquet-cli module with an improved command-line tool. The parquet-cli/README.md file has instructions for building and testing locally.

### DIFF
--- a/parquet-cli/README.md
+++ b/parquet-cli/README.md
@@ -48,10 +48,10 @@ To run from the target directory instead of using the `hadoop` command, first co
 mvn dependency:copy-dependencies
 ```
 
-Then, run the command-line using the `target/dependencies` for the classpath:
+Then, run the command-line and add `target/dependencies/*` to the classpath:
 
 ```
-java -cp 'target/parquet-cli-1.9.1-runtime.jar:target/dependency/*' org.apache.parquet.cli.Main
+java -cp 'target/*:target/dependency/*' org.apache.parquet.cli.Main
 ```
 
 
--- a/parquet-cli/README.md
+++ b/parquet-cli/README.md
@@ -1,0 +1,107 @@
+<!--
+  - Licensed to the Apache Software Foundation (ASF) under one
+  - or more contributor license agreements.  See the NOTICE file
+  - distributed with this work for additional information
+  - regarding copyright ownership.  The ASF licenses this file
+  - to you under the Apache License, Version 2.0 (the
+  - "License"); you may not use this file except in compliance
+  - with the License.  You may obtain a copy of the License at
+  -
+  -   http://www.apache.org/licenses/LICENSE-2.0
+  -
+  - Unless required by applicable law or agreed to in writing,
+  - software distributed under the License is distributed on an
+  - "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  - KIND, either express or implied.  See the License for the
+  - specific language governing permissions and limitations
+  - under the License.
+  -->
+
+## Building
+
+You can build this project using maven:
+
+```
+mvn clean install -DskipTests
+```
+
+
+## Running
+
+The build produces a shaded Jar that can be run using the `hadoop` command:
+
+```
+hadoop jar parquet-cli-1.9.1-runtime.jar org.apache.parquet.cli.Main
+```
+
+For a shorter command-line invocation, add an alias to your shell like this:
+
+```
+alias parquet="hadoop jar /path/to/parquet-cli-1.9.1-runtime.jar org.apache.parquet.cli.Main --dollar-zero parquet"
+```
+
+### Running without Hadoop
+
+To run from the target directory instead of using the `hadoop` command, first copy the dependencies to a folder:
+
+```
+mvn dependency:copy-dependencies
+```
+
+Then, run the command-line using the `target/dependencies` for the classpath:
+
+```
+java -cp 'target/parquet-cli-1.9.1-runtime.jar:target/dependency/*' org.apache.parquet.cli.Main
+```
+
+
+### Help
+
+The `parquet` tool includes help for the included commands:
+
+```
+parquet help
+```
+```
+Usage: parquet [options] [command] [command options]
+
+  Options:
+
+    -v, --verbose, --debug
+        Print extra debugging information
+
+  Commands:
+
+    help
+        Retrieves details on the functions of other commands
+    meta
+        Print a Parquet file's metadata
+    pages
+        Print page summaries for a Parquet file
+    dictionary
+        Print dictionaries for a Parquet column
+    check-stats
+        Check Parquet files for corrupt page and column stats (PARQUET-251)
+    schema
+        Print the Avro schema for a file
+    csv-schema
+        Build a schema from a CSV data sample
+    convert-csv
+        Create a file from CSV data
+    convert
+        Create a Parquet file from a data file
+    to-avro
+        Create an Avro file from a data file
+    cat
+        Print the first N records from a file
+    head
+        Print the first N records from a file
+
+  Examples:
+
+    # print information for create
+    parquet help create
+
+  See 'parquet help <command>' for more information on a specific command.
+```
+
--- a/parquet-cli/pom.xml
+++ b/parquet-cli/pom.xml
@@ -1,0 +1,153 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.apache.parquet</groupId>
+    <artifactId>parquet</artifactId>
+    <relativePath>../pom.xml</relativePath>
+    <version>1.9.1-SNAPSHOT</version>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>parquet-cli</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Apache Parquet Command-line</name>
+  <url>https://parquet.apache.org</url>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-avro</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>${avro.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>net.sf.opencsv</groupId>
+      <artifactId>opencsv</artifactId>
+      <version>${opencsv.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>${jackson2.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.beust</groupId>
+      <artifactId>jcommander</artifactId>
+      <version>${jcommander.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>${commons-codec.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-client</artifactId>
+      <version>${hadoop.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <!-- This module disables semver checks because it is not a public API.
+      <plugin>
+        <artifactId>maven-enforcer-plugin</artifactId>
+      </plugin>
+      -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+							<shadedArtifactAttached>true</shadedArtifactAttached>
+							<shadedClassifierName>runtime</shadedClassifierName>
+              <minimizeJar>false</minimizeJar>
+              <filters>
+                <filter>
+                  <artifact>org.xerial.snappy:*</artifact>
+                  <excludes>
+                    <exclude>**/LICENSE</exclude>
+                  </excludes>
+                </filter>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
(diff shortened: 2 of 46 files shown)


rdblue:
There are still Cloudera copyright headers, which @tomwhite is planning to change to the Apache header on Cloudera's behalf. Thanks, Tom!

parquet-cli/README.md
```diff
@@ -0,0 +1,107 @@
...
+To run from the target directory instead of using the `hadoop` command, first copy the dependencies to a folder:
+
+```
+mvn dependency:copy-dependencies
+```
+
+Then, run the command-line using the `target/dependencies` for the classpath:
+
+```
+java -cp 'target/parquet-cli-1.9.1-runtime.jar:target/dependency/*' org.apache.parquet.cli.Main
```
rdblue: Fixed. Thanks for taking a look!

sadikovi:
@rdblue is `parquet-tools` going to be removed, and will `parquet-cli` be available on maven?

rdblue:
@sadikovi, the plan is for parquet-cli to be available through maven in the next release. I'm not sure about removing parquet-tools. I don't use them and I think that this is going to be easier to maintain, but I'm not going to insist they be removed.

julienledem:
Let's not remove parquet-tools right now since the commands are not the same. If people shift to the new tool then we can deprecate and then remove it.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/CheckParquet251Command.java
```diff
@@ -0,0 +1,351 @@
...
+import org.slf4j.Logger;
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
+
+@Parameters(commandDescription = "Check Parquet files for corrupt page and column stats (PARQUET-251)")
+public class CheckParquet251Command extends BaseCommand {
```
piyushnarang: how about naming this check corrupt stats instead of check parquet 251?

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,330 @@
...
+    } else if (bytes > MB) {
+      return String.format("%.03f MB", bytes / MB);
+    } else if (bytes > KB) {
+      return String.format("%.03f kB", bytes / KB);
+    } else {
+      return String.format("%.02f B", bytes);
+    }
+  }
+
+  public static String humanReadable(long bytes) {
```
sadikovi: Should method be like snippet below instead of copying logic (there is also a possible typo around `%.02f B` in above case and `%d B` below)?
```java
public static String humanReadable(long bytes) {
  float fbytes = (float) bytes;
  return humanReadable(fbytes);
}
```

parquet-cli/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.apache.parquet</groupId>
+    <artifactId>parquet</artifactId>
+    <relativePath>../pom.xml</relativePath>
+    <version>1.9.1-SNAPSHOT</version>
```
rdblue: This matches the other module POM files and is updated by Maven's release plugin. I don't have much of an opinion either way on best practice, but I think that this should match the others for now. Feel free to open a follow-up issue to change over to `${project.version}` if you think we should.

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
rdblue: The bytes argument in the one above is a float because it is typically used as an average, showing the average number of bytes for values in a row group or page. That's why the float version uses `%.02f` and not `%d`. This function is called when the number of bytes is exact, so it uses `%d`.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/CheckParquet251Command.java
rdblue: This only checks for PARQUET-251, where min and max aren't the actual min and max values for some String columns. It doesn't check, for example, that the correct sort order was used or that the min and max for numeric columns are valid. Because it is limited to checking for PARQUET-251, I'd leave the name as it is.

parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
```diff
@@ -0,0 +1,457 @@
...
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.charset.Charset;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+public abstract class BaseCommand implements Command, Configurable {
```
rdblue: The FS operations are convenience methods to help commands. That's basically the purpose of this base class and they require access to the Configuration set on the command instance. I think it makes sense to keep those here, so they're easily accessed by commands.

I agree that some of the static classes can be refactored and placed in a util package.

rdblue:
@piyushnarang, thanks for taking the time to review. I'll fix a few things that you pointed out and I've responded to some comments to clarify.

> I think it would be great if we could break it up to make it iterative.

I don't think I have the time to break this into a series of commits. Perhaps it would be helpful to review a section at a time along the lines of the commits you suggested? It probably isn't very valuable to break it across commits in the long run, unless there are a huge number of changes that are needed.

> Seems like we're missing thrift / protobuf support right?

This shouldn't hold up what's already done. It would be great if someone added the ability to read thrift and protobuf, but that can be done later.

> If possible it would be nice to add some unit tests for each of the commands.

It is possible to test, and a good idea in the long run. I wanted to get this out to users sooner than I would be able to find time to thoroughly test. Is it okay to focus on areas that must be tested to in order to get this committed, and add tests as it evolves later?

parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
rdblue: Removed the static embedded classes.

isnotinvain:
I want to echo what @piyushnarang has commented here, if not for this PR but for future ones.

This PR is really large. The larger a PR is, the less likely it is to be thoroughly reviewed. I think we really need to aim for smaller / more incremental PRs than this in general. I don't know if it's worth the time to split this one up now that it's been done, but I think we should avoid setting a precedent of PRs this large being common, especially for all new code (as opposed to large refactors that sometimes have to be large).

Additionally, we need to also make sure that new code is accompanied by matching new tests. Wanting to get something into the hands of users early w/o the matching tests is also not something I think we should set a precedent for either. I don't think that means 100% test coverage, but there should be test coverage for the most important parts.

Was this a project developed separately, and now it's being imported into the parquet codebase? That sort of situation is difficult, as it may have been built incrementally and reviewed thoroughly, and it's odd to break it up and import it piece by piece after the fact, so I do understand that.

rdblue:
@isnotinvain, this was developed as a separate project. I originally thought it would be better as a separate Parquet repository since it isn't really a part of Parquet MR (but does depend on it). I think it's more important to get this project released than to add thorough unit tests at this point because we've been using it internally and know that it works for the most part. It also provides some great help for people experimenting with Parquet that you can't get elsewhere, like CSV, JSON, and Avro file conversions.

I completely agree that we should aim for smaller PRs with test coverage. I think that's more the norm, like #390 that is a series of commits and has thorough tests for the ByteBufferInputStream classes.

parquet-cli/src/main/java/org/apache/parquet/cli/Help.java
```diff
@@ -0,0 +1,147 @@
...
+        List<String> examples = ((Command) commander.getObjects().get(0)).getExamples();
+        if (examples != null) {
+          console.info("\n  Examples:");
+          for (String example : examples) {
+            if (example.startsWith("#")) {
+              // comment
+              console.info("\n    {}", example);
+            } else {
+              console.info("    {} {} {}",
+                  new Object[] {programName, cmd, example});
```
julienledem: console doesn't have varargs methods?

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,335 @@
...
+      case UNCOMPRESSED:
+        return "_";
+      case SNAPPY:
+        return "S";
+      case GZIP:
+        return "G";
+      case LZO:
+        return "L";
+      default:
+        return "?";
```
julienledem: possibly you could add this as a method in the codec enum

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,335 @@
...
+        return isDict ? "_" : "R";
+      case RLE_DICTIONARY:
+        return "R";
+      case DELTA_BINARY_PACKED:
+      case DELTA_LENGTH_BYTE_ARRAY:
+      case DELTA_BYTE_ARRAY:
+        return "D";
+      default:
+        return "?";
+    }
```
julienledem: same comment

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,335 @@
...
+
+  public static String encodingStatsAsString(EncodingStats encodingStats) {
+    StringBuilder sb = new StringBuilder();
+    if (encodingStats.hasDictionaryPages()) {
+      for (Encoding encoding: encodingStats.getDictionaryEncodings()) {
+        sb.append(encodingAsString(encoding, true));
+      }
+      sb.append(" ");
+    } else {
+      sb.append("  ");
```
julienledem: mv sb.append out of the if?

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,335 @@
...
+        sb.append("R");
+      }
+      if (encodings.contains(PLAIN)) {
+        sb.append("_");
+      }
+      if (encodings.contains(DELTA_BYTE_ARRAY) ||
+          encodings.contains(DELTA_BINARY_PACKED) ||
+          encodings.contains(DELTA_LENGTH_BYTE_ARRAY)) {
+        sb.append("D");
+      }
```
julienledem: factor out shortString(encodings) ?

parquet-cli/src/main/java/org/apache/parquet/cli/Util.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+    return sb.toString();
+  }
+
+  private static final Splitter DOT = Splitter.on('.');
```
julienledem: we should centralize path splitting

parquet-cli/src/main/java/org/apache/parquet/cli/commands/CheckParquet251Command.java
julienledem: SGTM

parquet-cli/src/main/java/org/apache/parquet/cli/commands/CheckParquet251Command.java
```diff
@@ -0,0 +1,351 @@
...
+        throw new BadStatsException("Number of nulls doesn't match.");
+      }
+
+      console.debug(String.format(
+          "Validated stats min=%s max=%s nulls=%d for page=%s col=%s",
+          String.valueOf(stats.genericGetMin()),
+          String.valueOf(stats.genericGetMax()), stats.getNumNulls(), page,
+          Arrays.toString(desc.getPath())));
+    }
+  }
```
julienledem: nice feature

parquet-cli/src/main/java/org/apache/parquet/cli/util/Schemas.java
```diff
@@ -0,0 +1,498 @@
...
+import org.codehaus.jackson.node.NullNode;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+public class Schemas {
```
julienledem: should this be in parquet-avro?

parquet-cli/src/main/java/org/apache/parquet/cli/util/RuntimeIOException.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.parquet.cli.util;
+
+import java.io.IOException;
+
+/**
+ * RuntimeException wrapper for IOExceptions
+ */
+public class RuntimeIOException extends RuntimeException {
```
julienledem: should extend ParquetRuntimeException?

parquet-cli/src/main/java/org/apache/parquet/cli/util/RecordException.java
```diff
@@ -0,0 +1,53 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.cli.util;
+
+/**
+ * Exception to signal that a record could not be read or written.
+ */
+public class RecordException extends RuntimeException {
```
julienledem: extends ParquetRuntimeException?

parquet-cli/src/main/java/org/apache/parquet/cli/util/RecordException.java
```diff
@@ -0,0 +1,53 @@
...
+   *          A String message for the exception.
+   */
+  public static void check(boolean isValid, String message, Object... args) {
+    if (!isValid) {
+      String[] argStrings = new String[args.length];
+      for (int i = 0; i < args.length; i += 1) {
+        argStrings[i] = String.valueOf(args[i]);
+      }
+      throw new RecordException(
+          String.format(String.valueOf(message), (Object[]) argStrings));
```
julienledem: you could also declare argString as Object[] (instead of casting)

parquet-cli/src/main/java/org/apache/parquet/cli/util/Expressions.java
```diff
@@ -0,0 +1,391 @@
...
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.util.Utf8;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+
+public class Expressions {
```
julienledem: javadoc?

parquet-cli/src/main/java/org/apache/parquet/cli/util/Expressions.java
```diff
@@ -0,0 +1,391 @@
...
+  private static final Pattern NUMERIC_RE = Pattern.compile("^\\d+$");
+
+  public static Object select(Schema schema, Object datum, String path) {
```
julienledem: javadoc? It would make it easier to follow what this does

parquet-cli/src/main/java/org/apache/parquet/cli/util/Expressions.java
```diff
@@ -0,0 +1,391 @@
...
+        GenericRecord record = (GenericRecord) datum;
+        Schema.Field field = schema.getField(token.value);
+        Preconditions.checkArgument(field != null,
+            "No such field '%s' in schema: %s", token.value, schema);
+        return select(field.schema(), record.get(token.value), token.children);
+
+      case MAP:
+        Preconditions.checkArgument(datum instanceof Map,
+            "Not a map: %s", datum);
+        Map<Object, Object> map = (Map<Object, Object>) datum;
```
julienledem: you can also use `Map<?,?>` and cast on get to avoid warnings

parquet-cli/src/main/java/org/apache/parquet/cli/util/Codecs.java
```diff
@@ -0,0 +1,50 @@
...
+public class Codecs {
+  public static CompressionCodecName parquetCodec(String codec) {
+    try {
+      return CompressionCodecName.valueOf(codec.toUpperCase(Locale.ENGLISH));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Unknown compression codec: " + codec);
+    }
+  }
+
+  public static CodecFactory avroCodec(String codec) {
```
julienledem: in parquet-avro?

parquet-cli/src/main/java/org/apache/parquet/cli/json/AvroJsonReader.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public void remove() {
+    throw new UnsupportedOperationException("Remove is not implemented.");
+  }
+
+  @Override
+  public Iterator<E> iterator() {
+    return this;
```
julienledem: the contract of iterable would be to get a new one everytime

julienledem:
@isnotinvain @piyushnarang In general I agree that we should avoid large patches and get smaller ones. However this particular patch doesn't touch any of the existing parquet code and by itself has a lot of independent parts that are easy to review independently in one patch (various commands, some utility packages, etc) and I don't think that it would help much to have *this particular patch* split into smaller ones since it would basically be split by files. It would be more work on Ryan's side with very little benefit on the reviewer end. So I think we can make an exception here and review as is. I did a first pass review above.

parquet-cli/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+              </filters>
+              <artifactSet>
+                <includes>
+                  <include>*</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+									<!-- relocate Avro in the runtime jar to avoid conflicts with
+                       on-cluster Avro versions.
```
rdblue: Guava shouldn't be necessary because Hadoop uses 11.0.2 and this specifies 11.0.

julienledem:
we should merge this.

xhochy:
+1, I'm already using this regularly!

rdblue:
Great! I'll rebase and commit it shortly. I think I was working on tests when I last had a chance... so I'll make sure that's reasonably complete.

PR #385, opened 2016-11-23
PARQUET-781: ParquetOuptputFormat should support custom OutputCommitter

mikkokupsu:


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -436,11 +438,20 @@ private static int getMaxPaddingSize(Configuration conf) {
...
+        OUTPUT_COMMITTER_CLASS, ParquetOutputCommitter.class.getName());
+      try {
+        committer = (OutputCommitter) Class.forName(outputCommitterClassName)
+          .getDeclaredConstructor(Path.class, TaskAttemptContext.class)
+          .newInstance(output, context);
+      } catch (final ClassNotFoundException | NoSuchMethodException | InstantiationException |
+          IllegalAccessException | InvocationTargetException ex) {
+        throw new BadConfigurationException("could not instantiate output committer class: "
+          + outputCommitterClassName, ex);
+      }
```
julienledem: this can be achieved by extending the ParquetOutputFormat. In what context do you need this?

PR #387, opened 2016-11-30
PARQUET-137 Add support for Pig datetimes

osayankin:


parquet-pig/src/main/java/org/apache/parquet/pig/TupleWriteSupport.java
```diff
@@ -174,6 +180,8 @@ private void writeValue(Type type, FieldSchema pigType, Tuple t, int i) {
             bytes = ((DataByteArray)t.get(i)).get();
           } else if (pigType.type == DataType.CHARARRAY) {
             bytes = ((String)t.get(i)).getBytes("UTF-8");
+          } else if (pigType.type == DataType.DATETIME) {
+            bytes = convertDateTimeToInt96((DateTime) t.get(i));
```
xhochy: As we want to move away from INT96 to TIMESTAMP_MILLIS and TIMESTAMP_MICROS, shouldn't they be rather used here?

parquet-pig/src/main/java/org/apache/parquet/pig/PigSchemaConverter.java
```diff
@@ -368,7 +368,6 @@ private Type convertWithName(FieldSchema fieldSchema, String name) {
       case DataType.DOUBLE:
         return primitive(name, PrimitiveTypeName.DOUBLE);
       case DataType.DATETIME:
-        throw new UnsupportedOperationException();
```
julienledem: intended to be just binary?

parquet-pig/src/main/java/org/apache/parquet/pig/TupleWriteSupport.java
julienledem: seconded

julienledem: Is there a specific use case here? Like working with Hive or Impala?

parquet-pig/src/main/java/org/apache/parquet/pig/PigSchemaConverter.java
osayankin: Why not?

parquet-pig/src/test/java/org/apache/parquet/pig/TestParquetStorer.java
```diff
@@ -83,6 +86,54 @@ public void testStorer() throws ExecException, Exception {
   }
 
   @Test
+  public void testStorerWithDateTime() throws Exception {
+    String out = "target/out";
```
julienledem: "target/testStorerWithDateTime"

parquet-pig/src/test/java/org/apache/parquet/pig/TestParquetStorer.java
```diff
@@ -83,6 +86,54 @@ public void testStorer() throws ExecException, Exception {
...
+    pigServer.registerQuery("Store A into '"+out+"' using "+ParquetStorer.class.getName()+"();");
+    if (pigServer.executeBatch().get(0).getStatus() != JOB_STATUS.COMPLETED) {
+      throw new RuntimeException("Job failed", pigServer.executeBatch().get(0).getException());
+    }
+
+    pigServer.registerQuery("B = LOAD '"+out+"' USING "+ParquetLoader.class.getName()+"();");
+    pigServer.registerQuery("Store B into 'out' using mock.Storage();");
+    if (pigServer.executeBatch().get(0).getStatus() != JOB_STATUS.COMPLETED) {
+      throw new RuntimeException("Job failed", pigServer.executeBatch().get(0).getException());
+    }
```
julienledem: factor out method for this repeated logic?

parquet-pig/src/main/java/org/apache/parquet/pig/TupleWriteSupport.java
julienledem: We want to deprecate int96 in favor of other timestamp types here:
https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#datetime-types
Millis would work right?

parquet-pig/src/main/java/org/apache/parquet/pig/PigSchemaConverter.java
julienledem: if we use timestamp it would be a long with an original type of TIMESTAMP.

PR #389, opened 2016-12-06
PARQUET-791: Add missing column support for UserDefinedPredicate

viirya:
This extends the fixing #354 to UserDefinedPredicate.


rdblue:
Is it valid to call a UserDefinedPredicate with null? If so, then we could run it once with null, cache the return value, and use that instead of BLOCK_MIGHT_MATCH every time.

viirya:
@rdblue Thanks for reviewing. In this case (`getColumnChunk` returns null), I think the overhead is very tiny. I am not sure if caching can be actually benefit.

rdblue:
I'm not really talking about the overhead of running the function. Just that the default assumption is that we don't know anything about whether the predicate will accept a block of nulls. I think there are a lot of cases where null will be filtered, so this makes the conservative but wrong decision. If we can actually call the predicate with null, we can use it to eliminate whole row groups and files.

viirya:
@rdblue Thanks. I see. That makes sense to me. We can run `UserDefinedPredicate.keep` with null value to know if the predicate will keep or drop null values, and use the result to eliminate whole row groups.

viirya:
@rdblue I've updated this like above. Please take a look. Thank you.

PR #394, opened 2016-12-15
PARQUET-801: Allow UserDefinedPredicates in DictionaryFilter

pwoody:


parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -363,14 +363,54 @@ public Boolean visit(Not not) {
         "This predicate contains a not! Did you forget to run this predicate through LogicalInverseRewriter? " + not);
   }
 
+  private <T extends Comparable<T>, U extends UserDefinedPredicate<T>> Boolean visit(UserDefined<T, U> ud, boolean inverted) {
+    Column<T> filterColumn = ud.getColumn();
+    ColumnChunkMetaData meta = getColumnChunk(filterColumn.getColumnPath());
+
+    // Cannot make assumptions around UDPs and acceptance of null values
+    if (meta == null || hasNonDictionaryPages(meta)) {
```
rdblue: When `meta == null`, the column is missing and all values in that column are null. In @viirya's recent patch, 71cff7c5940b7101ff098601850d46b7a4698180, the user-defined predicate is called with null to see whether all rows can be filtered. I think this patch can work the same way.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -363,14 +363,54 @@ public Boolean visit(Not not) {
...
+    }
+
+    try {
+      Set<T> dictSet = expandDictionary(meta);
+      if (dictSet == null) {
+        return BLOCK_MIGHT_MATCH;
+      }
+
+      boolean allMatch = true;
+      for(T entry : dictSet) {
```
rdblue: Nit: please insert a space between control flow keywords and parens.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -363,14 +363,54 @@ public Boolean visit(Not not) {
...
+        if (ud.getUserDefinedPredicate().keep(entry)) {
+          if (!inverted) {
+            return BLOCK_MIGHT_MATCH;
+          }
+        } else {
+          allMatch = false;
```
rdblue: I think this can have similar short-circuit logic. If the entry does not match but the predicate is inverted, then it does match and so does the entire block.

That would allow you to simplify the logic below, which cannot get to the third (else) case without hitting the short-circuit condition above.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
```diff
@@ -331,6 +338,43 @@ public void testOr() throws Exception {
         canDrop(or(x, y), ccmd, dictionaries));
   }
 
+
+  @Test
+  public void testUdp() throws Exception {
+    InInt32UDP udp = new InInt32UDP(ImmutableSet.of(42));
```
rdblue: Nit: it would be more clear if udp and udp1 had more descriptive names.

pwoody:
@rdblue appreciate the quick review and the comments! I've made the changes.

Somewhat related: has parquet-mr had something like checkstyle in the past or would the project be open to it?

rdblue:
We'd certainly be open to adding checkstyle to the build.

rdblue:
This looks good to me. I'll merge once tests pass.

PR #395, opened 2016-12-19
PARQUET-772: Fix locale-specific test failures.

zivanfi:
The statistics tests were failing in locales with a decimal mark other than "."


julienledem:
@xhochy you should feel free to merge :)

PR #396, opened 2016-12-19
PARQUET-806: Parquet-tools silently suppresses error messages

zivanfi:
The "error message" that used to be

org/apache/hadoop/conf/Configuration

now becomes:

NoClassDefFoundError: org/apache/hadoop/conf/Configuration

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
@@ -162,7 +162,7 @@ public class Main {
   }
 
   public static void die(Throwable th, boolean usage, String name, Command command) {
-    die(th.getMessage(), usage, name, command);
+    die(th.toString(), usage, name, command);
   }
 
   public static void main(String[] args) {


parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
```diff
@@ -226,7 +226,12 @@ public void close() throws IOException {}
       die("Invalid arguments: " + ex.getMessage(), true, name, command);
     } catch (Throwable th) {
       if (debug) th.printStackTrace(Main.err);
-      die(th, false, name, command);
```
sadikovi: I think it is better to just modify `die()` to print `Throwable` as exception class + message instead of doing it specifically in `Main`.

zivanfi: Fixed, PTAL.

parquet-tools/src/main/java/org/apache/parquet/tools/Main.java
```diff
@@ -162,7 +162,12 @@ public static void die(String message, boolean usage, String name, Command comma
   }
 
   public static void die(Throwable th, boolean usage, String name, Command command) {
-    die(th.getMessage(), usage, name, command);
+    String errorMsg = th.getClass().getSimpleName();
+    String errorDetails = th.getMessage();
+    if (errorDetails != null && !errorDetails.isEmpty()) {
+      errorMsg += ": " + errorDetails;
+    }
```
julienledem: I think th.toString() produces the same result.

zivanfi: Ah, thanks for the tip!

PR #398, opened 2017-01-03
PARQUET-822: Upgrade java dependencies

gszadovszky:
2 minor code/config modification related to the version upgrades:
- TestMemoryManager.java: I guess, it was caused by the junit upgrade however, it is not clear why it was working before. The issue was that the second run of `createWriter(1).close(null)` failed with `IOException` about that the file already exists.
- pom.xml (added exclusion for fastutil): The shaded dependency upgrade in `parquet-column`  caused failure of API version compatibility check.

`mvn clean install` worked fine. Any idea about additional testing is welcomed.

### DIFF
--- a/parquet-benchmarks/pom.xml
+++ b/parquet-benchmarks/pom.xml
@@ -32,7 +32,7 @@
   <url>https://parquet.apache.org</url>
 
   <properties>
-    <jmh.version>1.3.4</jmh.version>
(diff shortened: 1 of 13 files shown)


julienledem:
LGTM. @rdblue ?

rdblue:
+1

It would be nice to move some of the versions embedded in submodules to properties, but I don't think that needs to block this commit.

PR #399, opened 2017-01-04
PARQUET-825: Static analyzer findings (NPEs, resource leaks)

gszadovszky:
Some trivial code fixes based on findings on static code analyzer tools (Sonar, Fortify)
@piyushnarang: Sorry, renaming the branch caused the closing of the original PR...

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/command/Registry.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/command/Registry.java
@@ -39,7 +39,7 @@ public final class Registry {
     for (Map.Entry<String,Class<? extends Command>> entry : registry.entrySet()) {
       try {
         results.put(entry.getKey(), entry.getValue().newInstance());
-      } catch (Throwable th) {
+      } catch (Exception ex) {
       }
     }
 
@@ -54,7 +54,7 @@ public final class Registry {
 
     try {
       return clazz.newInstance();
-    } catch (Throwable th) {
+    } catch (Exception ex) {
       return null;
     }
   }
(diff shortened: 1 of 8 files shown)


piyushnarang:
👍  thanks looks good to me. Might be a nice idea for us to include these static analyzers as part of the default build.

gszadovszky:
Thanks for the review, @piyushnarang.

I agree it would be nice but the tools where these findings came from are commercial or partly commercial so, I am not sure we would be able to do it. 

However, there are open source tools (e.g. findbugs, checkstyle, pmd) which might worth trying out. (Avro is already using checkstyle and findbugs during the maven build.)

parquet-tools/src/main/java/org/apache/parquet/tools/command/Registry.java
```diff
@@ -39,7 +39,7 @@
     for (Map.Entry<String,Class<? extends Command>> entry : registry.entrySet()) {
       try {
         results.put(entry.getKey(), entry.getValue().newInstance());
-      } catch (Throwable th) {
+      } catch (Exception ex) {
```
julienledem: I don't get why we had this catch in the first place

parquet-tools/src/main/java/org/apache/parquet/tools/command/Registry.java
```diff
@@ -54,7 +54,7 @@ public static Command getCommandByName(String name) {
 
     try {
       return clazz.newInstance();
-    } catch (Throwable th) {
+    } catch (Exception ex) {
```
julienledem: same

gszadovszky:
Both invoke Class.newInstance() which can cause several exceptions. It is not nice that we swallow all of these but removing the catch would be a backward incompatible change. 
Maybe, some logging should be added. What do you think, @julienledem?

julienledem:
@gszadovszky We could just wrap them in a RuntimeException and throw. But that would be a separate PR.

julienledem:
+1

PR #401, opened 2017-02-01
PARQUET-852: Slowly ramp up sizes of byte[] in ByteBasedBitPackingEncoder

JohnPJenkins:
https://issues.apache.org/jira/browse/PARQUET-852

### DIFF
--- a/parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBasedBitPackingEncoder.java
+++ b/parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBasedBitPackingEncoder.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBasedBitPackingEncoder.java
```diff
@@ -75,6 +79,10 @@ public void writeInt(int value) throws IOException {
       pack();
       if (packedPosition == slabSize) {
         slabs.add(BytesInput.from(packed));
+        totalFullSlabSize += slabSize;
+        if (slabSize < bitWidth * MAX_SLAB_SIZE_MULT) {
+          slabSize *= 2;
+        }
```
julienledem: This looks fine
@rdblue @isnotinvain @piyushnarang does it look good to you?
@JohnPJenkins did you check the performance impact of this?

JohnPJenkins: I did not, only the memory impact. I'll give parquet-benchmarks a try.

julienledem:
@JohnPJenkins One thing you could add in the test is verifying that the slab count in the end is a reasonable number.

@piyushnarang @isnotinvain @rdblue If that looks good to you, we should go ahead and merge

julienledem:
LGTM.
@piyushnarang @rdblue ?

piyushnarang:
👍  thanks for adding the updated test :-)

PR #403, opened 2017-02-01
PARQUET-856: Do not pass knownClass if it is abstract/interface/Object.

aeroevan:
Fixes a very small cornercase in parquet-avro if reflection of abstract generic fields are used.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -138,6 +138,11 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
       };
 
       Class<?> fieldClass = fields.get(avroField.name());
+      if ((null != fieldClass) && (Modifier.isAbstract(fieldClass.getModifiers()) ||
+          Modifier.isInterface(fieldClass.getModifiers()) ||
+          fieldClass.equals(Object.class)) ) {
+        fieldClass = null;
+      }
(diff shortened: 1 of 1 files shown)


parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -138,6 +138,11 @@ public void add(Object value) {
       };
 
       Class<?> fieldClass = fields.get(avroField.name());
+      if ((null != fieldClass) && (Modifier.isAbstract(fieldClass.getModifiers()) ||
+          Modifier.isInterface(fieldClass.getModifiers()) ||
+          fieldClass.equals(Object.class)) ) {
+        fieldClass = null;
+      }
```
julienledem: looks ok to me.
@rdblue is the Avro expert

aeroevan: We were chatting on slack about the full reasoning.

julienledem: @aeroevan did you get to a conclusion?

julienledem: @aeroevan @rdblue what's the conclusion?

rdblue:
@aeroevan, can you post a simple test that reproduces the problem? I'd like to make sure that this has a test so it doesn't become a regression. That will also help me with the review, since I want to make sure that this wouldn't break valid cases. Thanks!

PR #404, opened 2017-02-17
Parquet-884: Add support for Decimal datatype to Parquet-Pig record reader

EllenKletscher:
Adds conversion support to Pig for Decimal datatype. Based on the scala code in the spark project that provides a similar function for their sql library.

### DIFF
--- a/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
+++ b/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
@@ -1,0 +1,42 @@
+package org.apache.parquet.pig.convert;
+
+import java.nio.ByteBuffer;
+import java.math.BigInteger;
+import java.math.BigDecimal;
+import static java.lang.Math.pow;
+
+import org.apache.parquet.io.api.Binary;
+
+/*
+ * Conversion between Parquet Decimal Type to Java BigDecimal in Pig
+ * Code Based on the Apache Spark ParquetRowConverter.java
+ * 
+ *
+ */
+
+public class DecimalUtils {
+
+  public static BigDecimal binaryToDecimal(Binary value, int precision, int scale) {
+    if (precision <= 18) {
+      ByteBuffer buffer = value.toByteBuffer();
+      byte[] bytes = buffer.array();
+      int start = buffer.arrayOffset() + buffer.position();
+      int end = buffer.arrayOffset() + buffer.limit();
+      long unscaled = 0L;
+      int i = start;
+      while ( i < end ) {
+        unscaled = ( unscaled << 8 | bytes[i] & 0xff );
+        i++;
+      }
+      int bits = 8*(end - start);
+      long unscaledNew = (unscaled << (64 - bits)) >> (64 - bits);
+      if (unscaledNew <= -pow(10,18) || unscaledNew >= pow(10,18)) {
+        return new BigDecimal(unscaledNew);
+      } else {
+        return BigDecimal.valueOf(unscaledNew / pow(10,scale));
+      }
+    } else {
+      return new BigDecimal(new BigInteger(value.getBytes()), scale);
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


parquet-pig/src/main/java/org/apache/parquet/pig/convert/TupleConverter.java
```diff
@@ -530,6 +535,32 @@ public void addBinary(Binary value) {
...
+
+    @Override
+    final public void addBinary(Binary value) {
+      if (primitiveType != null) {
+        int precision = primitiveType.asPrimitiveType().getDecimalMetadata().getPrecision();
+        int scale = primitiveType.asPrimitiveType().getDecimalMetadata().getScale();
+        BigDecimal finaldecimal = DecimalUtils.binaryToDecimal(value, precision, scale);
+        parent.add(finaldecimal);
+      } else {
+        parent.add("Broken DecimalType");
```
EllenKletscher: I think I may just remove the check for primitiveType null... I was looking at the code and realized that there's already a check for it before newConverter gets called...
https://github.com/EllenKletscher/parquet-mr/blob/master/parquet-pig/src/main/java/org/apache/parquet/pig/convert/TupleConverter.java#L82

parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
```diff
@@ -0,0 +1,61 @@
...
+ * Conversion between Parquet Decimal Type to Java BigDecimal in Pig
+ * Code Based on the Apache Spark ParquetRowConverter.java
+ * 
+ *
+ */
+
+public class DecimalUtils {
+
+  public static BigDecimal binaryToDecimal(Binary value, int precision, int scale) {
+    if (precision <= 18) {
```
EllenKletscher: added, thanks!

parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.parquet.io.api.Binary;
+
+/*
+ * Conversion between Parquet Decimal Type to Java BigDecimal in Pig
+ * Code Based on the Apache Spark ParquetRowConverter.scala
+ * 
+ *
+ */
+
+public class DecimalUtils {
```
julienledem: this should go either in a shared parquet module (parquet-common, parquet-encoding, parquet-column) so that it can be reused accross modules.
In particular there is support for decimal in parquet-avro and we should verify that they are compatible.
It sounds like this has been tested with spark?
I just want to make sure we are compatible.

EllenKletscher:
Parquet-Avro converter comes from the avro conversions library: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L80-L86
There’s a simple converter in the tools lib for Parquet: https://github.com/apache/parquet-mr/blob/master/parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleRecordConverter.java#L174-L176
Hive has its own set of converters as well: https://github.com/apache/parquet-mr/blob/master/parquet-hive/parquet-hive-storage-handler/src/main/java/org/apache/hadoop/hive/ql/io/parquet/writable/BigDecimalWritable.java#L79-L93
The converter for Spark is essentially the same as what I have added: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetRowConverter.scala#L381-L391
In functional testing on the same dataset, I’ve compared the results from the spark, hive and this pig converter and seen the same results (with some fuzz depending on how well the decimals are typed when they are written).

rdblue:
If you want example conversion code for fixed and binary, support for the same format has been released in Avro: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/Conversions.java

Why is there a special case for precision <= 19?

julienledem:
@EllenKletscher Thank you for the extensive answer! 💯  
Could you move DecimalUtils.java to parquet-column (or somewhere you deem appropriate, but the Binary class is in parquet-column) and add those links in the doc that would be very useful.
This looks good to me.

EllenKletscher:
@julienledem So in the spirit of git-fu, can we approve this pull request as is, and then open up a new issue to refactor all of the decimal calls in parquet (pig, hive and tools) into the same lib?  I don't want to exceed scope for the associated issue.

PR #408, opened 2017-03-31
PARQUET-990 More detailed error messages in footer parsing

ash211:
Include invalid values in exception messages when reading footer for two situations:

- too-short files (include file length)
- files with corrupted footer lengths (include calculated footer start index)


ash211:
Thanks @julienledem !  Filed as https://issues.apache.org/jira/browse/PARQUET-990 and updated this PR's title

PR #409, opened 2017-04-25
PARQUET-964 Fix ParquetDecodingException: totalValueCount '0' <= 0

costimuraru:
https://issues.apache.org/jira/browse/PARQUET-964

Problem is reproduced for:
```
message TestProtobuf.ListOfList {
  optional binary top_field (UTF8);
  required group first_array (LIST) {
    repeated group array {
      optional int32 inner_field;
      required group second_array (LIST) {
        repeated int32 array;
      }
    }
  }
}
```

When `inner_field` is not populated, the `totalValueCount` is 0 and fails with exception, instead of skipping the column and parsing the `second_array` (which is populated). This PR fixes the issue.


costimuraru:
Found the actual problem. See PARQUET-964

PR #412, opened 2017-05-01
PARQUET-969: Update parquet-tools to convert Decimal datatype to BigD…

dsfcode:
Update parquet-tools so that decimal datatypes in parquet files are converted to their actual number representation when cat'ing to stdout. Currently they are output in binary format.


julienledem:
thank you @dsfcode !

dsfcode:
No problem!

PR #413, opened 2017-05-25
PARQUET-1005: Fix DumpCommand parsing to allow column projection

gerashegalov:
DumpCommand option for -c is specified as hasArgs() for unlimited
number of arguments following -c. The very description of the option
shows the real intent of using hasArg() such that multiple columns
can be specified as '-c c1 -c c2 ...'. Otherwise, the input path
is parsed as an argument for -c instead of the command itself.

### DIFF
--- a/parquet-protobuf/pom.xml
+++ b/parquet-protobuf/pom.xml
@@ -32,6 +32,8 @@
   <properties>
     <elephant-bird.version>4.4</elephant-bird.version>
     <protobuf.version>3.2.0</protobuf.version>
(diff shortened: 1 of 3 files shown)


julienledem:
Hi @gerashegalov 
Can you open a jira for this and add it to the description? "PARQUET-XXX: Fix DumpCommand parsing to allow column projection"
That's how we manage the changelog.
Thank you!

gerashegalov:
@julienledem Thank you, done!

julienledem:
Thank you!
+1

PR #414, opened 2017-06-06
PARQUET-1020: Support for Protobuf DynamicMessages by allowing user to set descriptor

a-buck:
Hello,

This is an initial implementation for this feature. I haven't added tests yet and perhaps it could be tidied up a bit but wanted to make sure that I am on the right track before spending too much time on it.
https://issues.apache.org/jira/browse/PARQUET-1020

Please let me know what you think.

Many thanks,

Alex

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoParquetOutputFormat.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoParquetOutputFormat.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


julienledem:
CCing folks who showed interest in parquet-proto:
@qinghui-xu @matt-martin @lukasnalezenec @costimuraru @kgalieva: comments?

andredasilvapinto:
I've made a few changes on top of @a-buck changes so it doesn't depend on static state but uses the hadoop config to pass the Protobuf descriptors (as base64 strings).

The code is here:
https://github.com/andredasilvapinto/parquet-mr/commit/9669b634bd001858ff4261609c07156cfa4f35c1

lukasnalezenec:
Hi, good idea.
We should not use static state and i would prefer to extract this code to some strategy (design pattern) class. Everybody could inject own strategy implementation using hadoop configuration.

alexcardell:
This PR seems to have stagnated but this is exactly what we're looking for, if I fork and fix those conflicts can we reignite the discussion?

PR #416, opened 2017-06-09
PARQUET-1026: allow unsigned binary stats when min == max

julienledem:
When min equals max this is a special case where unsigned stats would actually be the same as signed stats since there is only one value.
This is useful when the data is partitioned by that column and there's only one value in the file.
Drill for example takes advantage of this.

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
@@ -1,4 +1,4 @@
-/*
+/* 
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -39,32 +59,26 @@
...
-import static org.apache.parquet.schema.Type.Repetition.*;
-import static org.apache.parquet.hadoop.TestUtils.enforceEmptyDir;
-
-import org.apache.parquet.example.data.Group;
-import org.apache.parquet.example.data.simple.SimpleGroup;
-
-import org.apache.parquet.hadoop.example.GroupWriteSupport;
+import org.junit.Assume;
+import org.junit.Rule;
+import org.junit.Test;
```
rdblue: Could you revert the import ordering changes? Those make it hard to port changes across versions and people are still maintaining the pre-1.7.0 package names.

julienledem: will do

PR #421, opened 2017-08-03
PARQUET-1077: Use long key ids in KEYS file

lekv:
Created like so:

gpg --import < KEYS 2>&1 | grep key | sed -e 's/.*"\(.*\)".*/\1/' | \
while read k; do gpg --list-sigs --keyid-format long $k; gpg --export \
--armor $k; done > newkeys

### DIFF
--- a/KEYS
+++ b/KEYS
@@ -2,19 +2,17 @@ This file contains the PGP keys of various developers.
 
 Users: pgp < KEYS
   gpg --import KEYS
-Developers: 
+Developers:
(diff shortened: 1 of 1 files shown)


lekv:
@xhochy @rdblue - Does one of you have time for a quick look? The file hasn't changed since I pushed this, so it will still apply cleanly. Thanks! :)

rdblue:
Sorry, I thought I merged this.

+1

PR #425, opened 2017-09-05
PARQUET-1342:Add bloom filter utility class

cjjnjust:
This is an initial patch just include bloom filter itself.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -48,6 +48,9 @@ public class ParquetProperties {
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
 
+  // TODO: need to discuss a maximum value
+  public static final int DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES = 16 * 1024 * 1024;
+
   public static final ValuesWriterFactory DEFAULT_VALUES_WRITER_FACTORY = new DefaultValuesWriterFactory();
 
   private static final int MIN_SLAB_SIZE = 64;
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -48,7 +48,7 @@ public class ParquetProperties {
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
 
-  // TODO: need to discuss a maximum value
+  // TODO: need to discuss a maximum value when doing write side task
   public static final int DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES = 16 * 1024 * 1024;
 
   public static final ValuesWriterFactory DEFAULT_VALUES_WRITER_FACTORY = new DefaultValuesWriterFactory();
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -1,4 +1,4 @@
-/*
+/* 
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ * 
  *   http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -47,9 +47,6 @@ public class ParquetProperties {
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
-
-  // TODO: need to discuss a maximum value when doing write side task
-  public static final int DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES = 16 * 1024 * 1024;
 
   public static final ValuesWriterFactory DEFAULT_VALUES_WRITER_FACTORY = new DefaultValuesWriterFactory();
 
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
@@ -54,19 +54,13 @@ public class Bloom {
     BLOCK,
   }
 
-  /**
-   * Default false positive probability value use to calculate optimal number of bits
-   * used by bloom filter.
-   */
-  public final double DEFAULT_FPP = 0.01;
-
   // Bloom filter data header, including number of bytes, hash strategy and algorithm.
   public static final int HEADER_SIZE = 12;
 
   // Bytes in a tiny bloom filter block.
   public static final int BYTES_PER_FILTER_BLOCK = 32;
 
-  // Default seed for hash function
+  // Default seed for hash function, it comes from Murmur3 from Hive.
   public static final int DEFAULT_SEED = 104729;
 
   // Hash strategy used in this bloom filter.
@@ -222,7 +216,7 @@ public class Bloom {
    * the hash value of its plain encoding result.
    * @param hash the hash result of element.
    */
-  private void addElement(long hash) {
+  public void insert(long hash) {
     int bucketIndex = (int)(hash >> 32) & (bitset.length / BYTES_PER_FILTER_BLOCK - 1);
     int key = (int)hash;
 
@@ -241,7 +235,7 @@ public class Bloom {
    * @param hash the hash value of element plain encoding result.
    * @return false if element is must not in set, true if element probably in set.
    */
-  private boolean contains(long hash) {
+  public boolean find(long hash) {
     int bucketIndex = (int)(hash >> 32) & (bitset.length / BYTES_PER_FILTER_BLOCK - 1);
     int key = (int)hash;
 
@@ -348,22 +342,5 @@ public class Bloom {
    */
   public long hash(Binary value) {
       return hashFunction.hashBytes(value.toByteBuffer()).asLong();
-  }
-
-  /**
-   * Insert element to set represented by bloom bitset.
-   * @param hash the hash of value to insert into bloom filter..
-   */
-  public void insert(long hash) {
-      addElement(hash);
-  }
-
-  /**
-   * Determine whether an element exist in set or not.
-   * @param hash the element to contain.
-   * @return false if value is definitely not in set, and true means PROBABLY in set.
-   */
-  public boolean find(long hash) {
-    return contains(hash);
   }
 }
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
@@ -1,0 +1,369 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bloom;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.IntBuffer;
+
+import com.google.common.hash.Hashing;
+import com.google.common.hash.HashFunction;
+import org.apache.parquet.Preconditions;
+import org.apache.parquet.bytes.*;
+import org.apache.parquet.column.ParquetProperties;
+import org.apache.parquet.io.api.Binary;
+
+/**
+ * Bloom Filter is a compact structure to indicate whether an item is not in set or probably
+ * in set. Bloom class is underlying class of Bloom Filter which stores a bit set represents
+ * elements set, hash strategy and bloom filter algorithm.
+ *
+ * Bloom Filter algorithm is implemented using block Bloom filters from Putze et al.'s "Cache-,
+ * Hash- and Space-Efficient Bloom Filters". The basic idea is to hash the item to a tiny Bloom
+ * Filter which size fit a single cache line or smaller. This implementation sets 8 bits in
+ * each tiny Bloom Filter. Tiny bloom filter are 32 bytes to take advantage of 32-bytes SIMD
+ * instruction.
+ */
+
+public class Bloom {
+  // Hash strategy available for bloom filter.
+  public enum HashStrategy {
+    MURMUR3_X64_128,
+  }
+
+  // Bloom filter algorithm.
+  public enum Algorithm {
+    BLOCK,
+  }
+
+  /**
+   * Default false positive probability value use to calculate optimal number of bits
+   * used by bloom filter.
+   */
+  public final double DEFAULT_FPP = 0.01;
+
+  // Bloom filter data header, including number of bytes, hash strategy and algorithm.
+  public static final int HEADER_SIZE = 12;
+
+  // Bytes in a tiny bloom filter block.
+  public static final int BYTES_PER_FILTER_BLOCK = 32;
+
+  // Default seed for hash function
+  public static final int DEFAULT_SEED = 104729;
+
+  // Hash strategy used in this bloom filter.
+  public final HashStrategy hashStrategy;
+
+  // Algorithm applied of this bloom filter.
+  public final Algorithm algorithm;
+
+  // The underlying byte array for bloom filter bitset.
+  private byte[] bitset;
+
+  // A integer array buffer of underlying bitset help setting bits.
+  private IntBuffer intBuffer;
+
+  // Hash function use to compute hash for column value.
+  private HashFunction hashFunction;
+
+  // The block based algorithm needs 8 odd SALT values to calculate eight index
+  // of bit to set, one bit in 32-bit word.
+  private static final int SALT[] = {0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d,
+    0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
+
+  /**
+   * Constructor of bloom filter, if numBytes is zero, bloom filter bitset
+   * will be created lazily and the number of bytes will be calculated through
+   * distinct values in cache. It use murmur3_x64_128 as its default hash function
+   * and block based algorithm as default algorithm.
+   * @param numBytes The number of bytes for bloom filter bitset, set to zero can
+   *                 let it calculate number automatically by using default DEFAULT_FPP.
+   */
+  public Bloom(int numBytes) {
+    this(numBytes, HashStrategy.MURMUR3_X64_128, Algorithm.BLOCK);
+  }
+
+  /**
+   * Constructor of bloom filter, if numBytes is zero, bloom filter bitset
+   * will be created lazily and the number of bytes will be calculated through
+   * distinct values in cache.
+   * @param numBytes The number of bytes for bloom filter bitset, set to zero can
+   *                 let it calculate number automatically by using default DEFAULT_FPP.
+   * @param hashStrategy The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  private Bloom(int numBytes, HashStrategy hashStrategy, Algorithm algorithm) {
+    initBitset(numBytes);
+
+    switch (hashStrategy) {
+      case MURMUR3_X64_128:
+        this.hashStrategy = hashStrategy;
+        hashFunction = Hashing.murmur3_128(DEFAULT_SEED);
+        break;
+      default:
+        throw new RuntimeException("Not supported hash strategy");
+    }
+
+    this.algorithm = algorithm;
+  }
+
+
+  /**
+   * Construct the bloom filter with given bit set, it is used when reconstruct
+   * bloom filter from parquet file.It use murmur3_x64_128 as its default hash
+   * function and block based algorithm as default algorithm.
+   * @param bitset The given bitset to construct bloom filter.
+   */
+  public Bloom(byte[] bitset) {
+    this(bitset, HashStrategy.MURMUR3_X64_128, Algorithm.BLOCK);
+  }
+
+  /**
+   * Construct the bloom filter with given bit set, it is used
+   * when reconstruct bloom filter from parquet file.
+   * @param bitset The given bitset to construct bloom filter.
+   * @param hashStrategy The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  private Bloom(byte[] bitset, HashStrategy hashStrategy, Algorithm algorithm) {
+    this.bitset = bitset;
+    this.intBuffer = ByteBuffer.wrap(bitset).asIntBuffer();
+
+    switch (hashStrategy) {
+      case MURMUR3_X64_128:
+        this.hashStrategy = hashStrategy;
+        hashFunction = Hashing.murmur3_128(DEFAULT_SEED);
+        break;
+      default:
+        throw new RuntimeException("Not supported hash strategy");
+    }
+    this.algorithm = algorithm;
+  }
+
+  /**
+   * Create a new bitset for bloom filter, at least 256 bits will be create.
+   * @param numBytes number of bytes for bit set.
+   */
+  private void initBitset(int numBytes) {
+    if (numBytes < BYTES_PER_FILTER_BLOCK) {
+      numBytes = BYTES_PER_FILTER_BLOCK;
+    }
+
+    // Get next power of 2 if it is not power of 2.
+    if ((numBytes & (numBytes - 1)) != 0) {
+      numBytes = Integer.highestOneBit(numBytes) << 1;
+    }
+
+    if (numBytes > ParquetProperties.DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES || numBytes < 0) {
+      numBytes = ParquetProperties.DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES;
+    }
+
+    this.bitset = new byte[numBytes];
+    this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.BIG_ENDIAN).asIntBuffer();
+  }
+
+  /**
+   * Write bloom filter to output stream. A bloom filter structure should include
+   * bitset length, hash strategy, algorithm, and bitset.
+   * @param out output stream to write
+   */
+  public void writeTo(OutputStream out) throws IOException {
+    // Write number of bytes of bitset.
+    out.write(BytesUtils.intToBytes(bitset.length));
+
+    // Write hash strategy
+    out.write(BytesUtils.intToBytes(this.hashStrategy.ordinal()));
+
+    // Write algorithm
+    out.write(BytesUtils.intToBytes(this.algorithm.ordinal()));
+
+    // Write bitset
+    out.write(bitset);
+  }
+
+  private int[] setMask(int key) {
+    int mask[] = new int[8];
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = key * SALT[i];
+    }
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = mask[i] >>> 27;
+    }
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = 0x1 << mask[i];
+    }
+
+    return mask;
+  }
+
+  /**
+   * Add an element to bloom filter, the element content is represented by
+   * the hash value of its plain encoding result.
+   * @param hash the hash result of element.
+   */
+  private void addElement(long hash) {
+    int bucketIndex = (int)(hash >> 32) & (bitset.length / BYTES_PER_FILTER_BLOCK - 1);
+    int key = (int)hash;
+
+    // Calculate mask for bucket.
+    int mask[] = setMask(key);
+
+    for (int i = 0; i < 8; i++) {
+      int value = intBuffer.get(bucketIndex * (BYTES_PER_FILTER_BLOCK / 4) + i);
+      value |= mask[i];
+      intBuffer.put(bucketIndex * (BYTES_PER_FILTER_BLOCK / 4) + i, value);
+    }
+  }
+
+  /**
+   * Determine where an element is in set or not.
+   * @param hash the hash value of element plain encoding result.
+   * @return false if element is must not in set, true if element probably in set.
+   */
+  private boolean contains(long hash) {
+    int bucketIndex = (int)(hash >> 32) & (bitset.length / BYTES_PER_FILTER_BLOCK - 1);
+    int key = (int)hash;
+
+    // Calculate mask for bucket.
+    int mask[] = setMask(key);
+
+    for (int i = 0; i < 8; i++) {
+      if (0 == (intBuffer.get(bucketIndex * (BYTES_PER_FILTER_BLOCK / 4) + i) & mask[i])) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  /**
+   * Calculate optimal size according to the number of distinct values and false positive probability.
+   * @param n: The number of distinct values.
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    Preconditions.checkArgument((p > 0.0 && p < 1.0),
+      "FPP should be less than 1.0 and great than 0.0");
+
+    final double M = -8 * n / Math.log(1 - Math.pow(p, 1.0 / 8));
+    final double MAX = ParquetProperties.DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES << 3;
+    int numBits = (int)M;
+
+    // Handle overflow.
+    if (M > MAX || M < 0) {
+      numBits = (int)MAX;
+    }
+
+    // Get next power of 2 if bits is not power of 2.
+    if ((numBits & (numBits - 1)) != 0) {
+      numBits = Integer.highestOneBit(numBits) << 1;
+    }
+
+    // Minimum
+    if (numBits < (BYTES_PER_FILTER_BLOCK << 3)) {
+      numBits = BYTES_PER_FILTER_BLOCK << 3;
+    }
+
+    return numBits;
+  }
+
+  /**
+   * used to decide if we want to work to the next page
+   * @return Bytes buffered of bloom filter.
+   */
+  public long getBufferedSize() {
+    return bitset.length;
+  }
+
+  /**
+   * Compute hash for int value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(int value) {
+    ByteBuffer plain = ByteBuffer.allocate(Integer.SIZE/Byte.SIZE);
+    plain.order(ByteOrder.LITTLE_ENDIAN).putInt(value);
+    return hashFunction.hashBytes(plain.array()).asLong();
+  }
+
+  /**
+   * Compute hash for long value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(long value) {
+    ByteBuffer plain = ByteBuffer.allocate(Long.SIZE/Byte.SIZE);
+    plain.order(ByteOrder.LITTLE_ENDIAN).putLong(value);
+    return hashFunction.hashBytes(plain.array()).asLong();
+  }
+
+  /**
+   * Compute hash for double value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(double value) {
+    ByteBuffer plain = ByteBuffer.allocate(Double.SIZE/Byte.SIZE);
+    plain.order(ByteOrder.LITTLE_ENDIAN).putDouble(value);
+    return hashFunction.hashBytes(plain.array()).asLong();
+  }
+
+  /**
+   * Compute hash for float value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(float value) {
+    ByteBuffer plain = ByteBuffer.allocate(Float.SIZE/Byte.SIZE);
+    plain.order(ByteOrder.LITTLE_ENDIAN).putFloat(value);
+    return hashFunction.hashBytes(plain.array()).asLong();
+  }
+
+  /**
+   * Compute hash for Binary value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(Binary value) {
+      return hashFunction.hashBytes(value.toByteBuffer()).asLong();
+  }
+
+  /**
+   * Insert element to set represented by bloom bitset.
+   * @param hash the hash of value to insert into bloom filter..
+   */
+  public void insert(long hash) {
+      addElement(hash);
+  }
+
+  /**
+   * Determine whether an element exist in set or not.
+   * @param hash the element to contain.
+   * @return false if value is definitely not in set, and true means PROBABLY in set.
+   */
+  public boolean find(long hash) {
+    return contains(hash);
+  }
+}
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
@@ -145,7 +145,7 @@ public class Bloom {
    */
   private Bloom(byte[] bitset, HashStrategy hashStrategy, Algorithm algorithm) {
     this.bitset = bitset;
-    this.intBuffer = ByteBuffer.wrap(bitset).asIntBuffer();
+    this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.LITTLE_ENDIAN).asIntBuffer();
 
     switch (hashStrategy) {
       case MURMUR3_X64_128:
@@ -177,7 +177,7 @@ public class Bloom {
     }
 
     this.bitset = new byte[numBytes];
-    this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.BIG_ENDIAN).asIntBuffer();
+    this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.LITTLE_ENDIAN).asIntBuffer();
   }
 
   /**
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
@@ -66,12 +66,14 @@ public class TestBloom {
 
   @Test
   public void testBinaryBloom() throws IOException {
+    int totalCount = 100000;
+    double fpp = 0.01;
     final long SEED = 104729;
-    Bloom binaryBloom = new Bloom(Bloom.optimalNumOfBits(100000, 0.01));
 
+    Bloom binaryBloom = new Bloom(Bloom.optimalNumOfBits(totalCount, fpp));
     List<String> strings = new ArrayList<>();
     RandomStr randomStr = new RandomStr(new Random(SEED));
-    for(int i = 0; i < 100000; i++) {
+    for(int i = 0; i < totalCount; i++) {
       String str = randomStr.get(10);
       strings.add(str);
       binaryBloom.insert(binaryBloom.hash(Binary.fromString(str)));
@@ -98,7 +100,7 @@ public class TestBloom {
 
     // exist can be true at probability 0.01.
     int exist = 0;
-    for (int i = 0; i < 100000; i++) {
+    for (int i = 0; i < totalCount; i++) {
       String str = randomStr.get(8);
       if (binaryBloom.find(binaryBloom.hash(Binary.fromString(str)))) {
         exist ++;
@@ -106,6 +108,6 @@ public class TestBloom {
     }
 
     // exist should be probably less than 1000 according default FPP 0.01.
-    assertTrue(exist < 1200);
+    assertTrue(exist < totalCount*fpp);
   }
 }
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.column.values.bloom;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+import org.apache.parquet.column.values.RandomStr;
+import org.apache.parquet.io.api.Binary;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestBloom {
+  @Test
+  public void testIntBloom () throws IOException {
+    Bloom bloom = new Bloom(279);
+    assertEquals("bloom filter size should be adjust to 512 bytes if input bytes is 279 bytes",
+      bloom.getBufferedSize(), 512);
+
+    for(int i = 0; i<10; i++) {
+      bloom.insert(bloom.hash(i));
+    }
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream((int)bloom.getBufferedSize() + bloom.HEADER_SIZE);
+    bloom.writeTo(baos);
+
+    ByteBuffer bloomBuffer = ByteBuffer.wrap(baos.toByteArray());
+
+    int length = Integer.reverseBytes(bloomBuffer.getInt());
+    int hash = Integer.reverseBytes(bloomBuffer.getInt());
+    int algorithm = Integer.reverseBytes(bloomBuffer.getInt());
+
+    byte[] bitset = new byte[length];
+    bloomBuffer.get(bitset);
+
+    bloom = new Bloom(bitset);
+
+    for(int i = 0; i < 10; i++) {
+      assertTrue(bloom.find(bloom.hash(i)));
+    }
+  }
+
+  @Test
+  public void testBinaryBloom() throws IOException {
+    final long SEED = 104729;
+    Bloom binaryBloom = new Bloom(Bloom.optimalNumOfBits(100000, 0.01));
+
+    List<String> strings = new ArrayList<>();
+    RandomStr randomStr = new RandomStr(new Random(SEED));
+    for(int i = 0; i < 100000; i++) {
+      String str = randomStr.get(10);
+      strings.add(str);
+      binaryBloom.insert(binaryBloom.hash(Binary.fromString(str)));
+    }
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream(
+      (int)binaryBloom.getBufferedSize() + binaryBloom.HEADER_SIZE);
+    binaryBloom.writeTo(baos);
+
+    ByteBuffer bloomBuffer = ByteBuffer.wrap(baos.toByteArray());
+
+    int length = Integer.reverseBytes(bloomBuffer.getInt());
+    int hash = Integer.reverseBytes(bloomBuffer.getInt());
+    int algorithm = Integer.reverseBytes(bloomBuffer.getInt());
+
+    byte[] bitset = new byte[length];
+    bloomBuffer.get(bitset);
+
+    binaryBloom = new Bloom(bitset);
+
+    for(int i = 0; i < strings.size(); i++) {
+      assertTrue(binaryBloom.find(binaryBloom.hash(Binary.fromString(strings.get(i)))));
+    }
+
+    // exist can be true at probability 0.01.
+    int exist = 0;
+    for (int i = 0; i < 100000; i++) {
+      String str = randomStr.get(8);
+      if (binaryBloom.find(binaryBloom.hash(Binary.fromString(str)))) {
+        exist ++;
+      }
+    }
+
+    // exist should be probably less than 1000 according default FPP 0.01.
+    assertTrue(exist < 1200);
+  }
+}
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/util/Expressions.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/util/Expressions.java
@@ -19,7 +19,7 @@
 
 package org.apache.parquet.cli.util;
 
-import com.google.common.base.Objects;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import org.apache.avro.Schema;
@@ -385,7 +385,7 @@ public class Expressions {
 
     @Override
     public String toString() {
-      return Objects.toStringHelper(this)
+      return MoreObjects.toStringHelper(this)
           .add("type", type)
           .add("value", value)
           .add("children", children)
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/BloomFilter.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/bloom/BloomFilter.java
@@ -28,23 +28,22 @@ import com.google.common.hash.Hashing;
 import com.google.common.hash.HashFunction;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.*;
-import org.apache.parquet.column.ParquetProperties;
 import org.apache.parquet.io.api.Binary;
 
 /**
- * Bloom Filter is a compact structure to indicate whether an item is not in set or probably
- * in set. Bloom class is underlying class of Bloom Filter which stores a bit set represents
- * elements set, hash strategy and bloom filter algorithm.
+ * A Bloom filter is a compact structure to indicate whether an item is not in a set or probably
+ * in a set. BloomFilter class stores a bit set represents a elements set, a hash strategy and a
+ * Bloom filter algorithm.
  *
- * Bloom Filter algorithm is implemented using block Bloom filters from Putze et al.'s "Cache-,
- * Hash- and Space-Efficient Bloom Filters". The basic idea is to hash the item to a tiny Bloom
- * Filter which size fit a single cache line or smaller. This implementation sets 8 bits in
- * each tiny Bloom Filter. Tiny bloom filter are 32 bytes to take advantage of 32-bytes SIMD
+ * This Bloom filter is implemented using block-based Bloom filter algorithm from Putze et al.'s
+ * "Cache-, Hash- and Space-Efficient Bloom filters". The basic idea is to hash the item to a tiny
+ * Bloom filter which size fit a single cache line or smaller. This implementation sets 8 bits in
+ * each tiny Bloom filter. Each tiny Bloom filter is 32 bytes to take advantage of 32-byte SIMD
  * instruction.
  */
 
-public class Bloom {
-  // Hash strategy available for bloom filter.
+public class BloomFilter {
+  // Bloom filter Hash strategy .
   public enum HashStrategy {
     MURMUR3_X64_128,
   }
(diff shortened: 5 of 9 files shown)


cjjnjust:
CI check failed on thrift build. No idea why..

winningsix:
Building is failed. If you change the thrift file in Parquet-format, you have to wait for the commitment for that part at first to make CI happy.

cjjnjust:
@winningsix In this patch I don't change any thrift file but just add Bloom and Murmur3 class. Strange.

winningsix:
@cjjnjust you can use the command locally to do a build.

cjjnjust:
@winningsix 
From the log we can see the error is happen when building thrift-0.7.0, and detail is related to PHP support. 

/home/travis/build/apache/parquet-mr/thrift-0.7.0/lib/php/src/ext/thrift_protocol/php_thrift_protocol.cpp:95:8: error: ‘function_entry’ does not name a type

I search this and found a JIRA https://issues.apache.org/jira/browse/THRIFT-1602.  I recall I met this error when building thrift also.

cjjnjust:
@winningsix 
I fixed build error by adding --without-php when configure thrift.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+    }
+  }
+
+  /**
+   * Determine where an element is in set or not.
+   * @param hash the hash value of element plain encoding result.
+   * @return false if element is must not in set, true if element probably in set.
+   */
+  public boolean bloomFind(long hash) {
+    int bucketIndex = (int)(hash >> 32) & (numBytes / BYTES_PER_BUCKET - 1);
```
jbapple-cloudera: This returns incorrect results when elements is empty.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  /**
+   * Calculate optimal size according to the number of distinct values and false positive probability.
+   * @param n: The number of distinct values.
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    int bits = (int)(-n * Math.log(p) / (Math.log(2) * Math.log(2)));
```
jbapple-cloudera: Please add sanity checking for p

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+
+    bits --;
```
jbapple-cloudera: Please explain what you are doing here

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+
+    return bits;
+  }
+
+
+  /**
+   * Element is represented by hash in bloom filter. The hash function takes plain encoding
+   * of element as input.
+   */
+  public Encoding getEncoding() {
```
jbapple-cloudera: unused

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+    return Encoding.PLAIN;
+  }
+
+  /**
+   * @return Bytes buffered of bloom filter.
+   */
+  public long getBufferedSize() {
+    return elements.size() * 8 + numBytes;
```
jbapple-cloudera: 1. What do you mean by buffered bytes?

2. Also, why multiply by 8?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  /**
+   * @return Bytes buffered of bloom filter.
+   */
+  public long getAllocatedSize() {
```
jbapple-cloudera: What does this mean?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+    switch(type) {
+      case INT32:
+        return new IntBloom(size, hash, algorithm);
+      case INT64:
+        return new LongBloom(size, hash, algorithm);
+      case FLOAT:
+        return new FloatBloom(size, hash, algorithm);
+      case DOUBLE:
+        return new DoubleBloom(size, hash, algorithm);
+      case BINARY:
```
jbapple-cloudera: You don't need lines 331 or 333, right?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+   * Determine whether an element exist in set or not.
+   * @param value the element to find.
+   * @return false if value is definitely not in set, and true means PROBABLY in set.
+   */
+  public boolean find (T value) {
+    return bloomFind(hash(value));
+  }
+
+  public static class BinaryBloom extends Bloom<Binary> {
+    private CapacityByteArrayOutputStream arrayout = new CapacityByteArrayOutputStream(1024, 64 * 1024, new HeapByteBufferAllocator());
```
jbapple-cloudera: What are the constants? Give them courtesy names, perhaps.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+   */
+  public void bloomInsert(long hash) {
+    int bucketIndex = (int)(hash >> 32) & (numBytes / BYTES_PER_BUCKET - 1);
+    int key = (int)hash;
+
+    setMask(key);
+
+    for (int i = 0; i < 8; i++) {
+      int bitsetIndex = bucketIndex * BYTES_PER_BUCKET + i * 4;
+      bitset[bitsetIndex] |= (byte)(mask[i] >>> 24);
```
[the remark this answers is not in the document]
cjjnjust: Java use big endian so we need to handle the byte order inside integer. Otherwise we need to allocate temporary variable to store reverted order bytes.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: 1. Buffered bytes is the total number of bytes we allocate in heap for bloom filter, please refer to similar definition in [dictionaryValueWriter](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java#L149).

2. Elements are stored in Long which is 8 bytes.

jbapple-cloudera: 1. Please add a comment that explains it as well as the method on https://github.com/apache/parquet-mr/blob/70f28810a5547219e18ffc3465f519c454fee6e5/parquet-column/src/main/java/org/apache/parquet/column/values/ValuesWriter.java

2. A HashSet<Long> will use a good bit more than 8 heap bytes per value.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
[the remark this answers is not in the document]
jbapple-cloudera: https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#reverseBytes-int-

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+    this.numBytes = input.array().length;
+    this.bloomFilterHash = hash;
+    this.bloomFilterAlgorithm = algorithm;
+  }
+
+  /**
+   * Create a new bitset for bloom filter, at least 256 bits will be create.
+   * @param numBytes number of bytes for bit set.
+   */
+  public void initBitset(int numBytes) {
```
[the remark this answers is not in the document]
cjjnjust: It will force to bucket size alignment.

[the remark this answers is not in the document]
jbapple-cloudera: Sorry - I meant power of 2. This must force the number of bytes allocated to be a power of 2, which I do not believe it does.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+  public static final int BLOOM_HEADER_SIZE = 12;
+
+  // Bytes in a bucket.
+  public static final int BYTES_PER_BUCKET = 32;
+
+  // Minimum bloom filter data size.
+  public static final int MINIMUM_BLOOM_SIZE = 256;
+
+  // Hash strategy used in this bloom filter.
+  public HASH bloomFilterHash = HASH.MURMUR3_X64_128;
```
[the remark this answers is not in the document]
cjjnjust: I planed to set it final, but I think it would be better to set as variable in case someone may have another implementation or option.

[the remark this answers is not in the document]
jbapple-cloudera: But it shouldn't change once the constructor runs. You can give it the "final" modifier and remove the initializer.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+   * will be created lazily and the number of bytes will be calculated through
+   * distinct values in cache.
+   * @param numBytes The number of bytes for bloom filter bitset, set to zero can
+   *                 let it calculate number automatically by using default FPP.
+   * @param hash The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(int numBytes, HASH hash, ALGORITHM algorithm) {
+    if (numBytes != 0) {
+      initBitset(numBytes);
```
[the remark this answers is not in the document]
cjjnjust: what do you mean "elements = null "? clear the item or set reference to null ?
This should be already empty since it is a constructor.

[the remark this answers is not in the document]
jbapple-cloudera: Set it to null. Better: don't give it an initializer outside of a constructor, since we won't know until a constructor runs if it should ever be created.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+   */
+  public final double FPP = 0.01;
+
+  // Bloom filter data header, including number of bytes, hash strategy and algorithm.
+  public static final int BLOOM_HEADER_SIZE = 12;
+
+  // Bytes in a bucket.
+  public static final int BYTES_PER_BUCKET = 32;
+
+  // Minimum bloom filter data size in byte.
```
jbapple-cloudera: Why should the minimum size be 256 bytes, rather than 32 bytes?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  // Algorithm applied of this bloom filter.
+  public ALGORITHM bloomFilterAlgorithm = ALGORITHM.BLOCK;
+
+  // A hash set to cache the distinct value from column.
+  private HashSet<Long> elements = new HashSet<Long>();
+
+  // The underlying byte array for bloom filter bitset.
+  private byte[] bitset;
+
+  // The size of bitset in byte.
```
jbapple-cloudera: That's already stored in bitset.length, right? Why duplicate it?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  private int numBytes;
+
+  // List of byte input to construct the bloom filter.
```
jbapple-cloudera: Please add comments to the file: Under what circumstances is it used? What are the data structure invariants? Can both it and bitset be non-null?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+   * Calculate optimal size according to the number of distinct values and false positive probability.
+   * @param n: The number of distinct values.
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    Preconditions.checkArgument((p > 0.0 && p < 1.0),
+      "FPP should be less than 1.0 and great than 0.0");
+
+    int bits = (int)(-n * Math.log(p) / (Math.log(2) * Math.log(2)));
```
jbapple-cloudera: I do not believe this math is correct when k, the number of hash functions, is fixed at 8.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+
+    return true;
+  }
+
+  /**
+   * Bloom filter bitset can be created lazily, flush() will set bits for
+   * all elements in cache. If bitset was already created and set, it do nothing.
+   */
+  public void flush() {
+    if (!elements.isEmpty() && bitset == null) {
```
jbapple-cloudera: Why does this silently do nothing if !elements.isEmpty() but bitset != null? Should it throw an exception?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  /**
+   * Determine whether an element exist in set or not.
+   * @param value the element to find.
+   * @return false if value is definitely not in set, and true means PROBABLY in set.
+   */
+  public boolean find (T value) {
+    return bloomFind(hash(value));
+  }
+
+  public static class BinaryBloom extends Bloom<Binary> {
```
jbapple-cloudera: Why nest this class inside Bloom<T>?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+    final int initSlabSize = 1024;
+    final int maxSlabSize = 64*1024;
+    private CapacityByteArrayOutputStream arrayout = new CapacityByteArrayOutputStream(initSlabSize,
+      maxSlabSize, new HeapByteBufferAllocator());
+    private LittleEndianDataOutputStream out = new LittleEndianDataOutputStream(arrayout);
```
jbapple-cloudera: This is not thread-safe.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+    this.bloomFilterAlgorithm = algorithm;
+  }
+
+  /**
+   * Construct the bloom filter with given bit set.
+   * @param bitset The given bitset to construct bloom filter.
+   * @param hash The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(byte[] bitset, HASH hash, ALGORITHM algorithm) {
```
jbapple-cloudera: Can you audit which methods are used? I don't see any calls to this.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  public static final int MINIMUM_BLOOM_SIZE = 256;
+
+  // Hash strategy used in this bloom filter.
+  public HASH bloomFilterHash = HASH.MURMUR3_X64_128;
+
+  // Algorithm applied of this bloom filter.
+  public ALGORITHM bloomFilterAlgorithm = ALGORITHM.BLOCK;
+
+  // A hash set to cache the distinct value from column.
+  private HashSet<Long> elements = new HashSet<Long>();
```
jbapple-cloudera: The data structures of "buffer for a Bloom filter" and "Bloom Filter" are entangled. Why not make a buffer structure with a HashSet<Long> that exists outside of Bloom<T>?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+ * elements set, hash strategy and bloom filter algorithm.
+ *
+ * Bloom Filter algorithm is implemented using block Bloom filters from Putze et al.'s "Cache-,
+ * Hash- and Space-Efficient Bloom Filters". The basic idea is to hash the item to a tiny Bloom
+ * Filter which size fit a single cache line or smaller. This implementation sets 8 bits in
+ * each tiny Bloom Filter. Tiny bloom filter are 32 bytes to take advantage of 32-bytes SIMD
+ * instruction.
+ */
+
+public abstract class Bloom<T extends Comparable<T>> {
```
jbapple-cloudera: Using inheritance and an abstract method to implement hash functions seems unnecessarily complex. Can you think of another way to do this, perhaps by making a BloomFilter that knew nothing about any type T, but just dealt with the resulting 64-bit hash values?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+   * Default false positive probability value use to calculate optimal number of bits
+   * used by bloom filter.
+   */
+  public final double FPP = 0.01;
+
+  // Bloom filter data header, including number of bytes, hash strategy and algorithm.
+  public static final int BLOOM_HEADER_SIZE = 12;
+
+  // Bytes in a bucket.
+  public static final int BYTES_PER_BUCKET = 32;
```
jbapple-cloudera: Why do you need both BYTES_PER_BUCKET and MINIMUM_BLOOM_SIZE? Isn't the MINIMUM_BLOOM_SIZE one bucket?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  }
+
+  /**
+   * Default false positive probability value use to calculate optimal number of bits
+   * used by bloom filter.
+   */
+  public final double FPP = 0.01;
+
+  // Bloom filter data header, including number of bytes, hash strategy and algorithm.
+  public static final int BLOOM_HEADER_SIZE = 12;
```
jbapple-cloudera: It is an anti-pattern to name constants with a class Foo "FOO_CONSTANT_NAME" - you're repeating the "Foo". The same is true of method names.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+
+    ByteBuffer bytes = ByteBuffer.allocate(numBytes);
+    this.bitset = bytes.array();
+    this.numBytes = numBytes;
+  }
+
+  /**
+   * @return Bytes in buffered in this bloom filter.
+   */
+  public BytesInput getBytes() {
```
jbapple-cloudera: I don't see any callers.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+      0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
+
+    int mask[] = new int[8];
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = key * SALT[i];
+    }
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = mask[i] >> 27;
```
jbapple-cloudera: This looks to me like a correctness issue: it should be unsigned right shift, yes?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: What if this overflows - can bits end up negative, or just truncated and too low? Please consider that and add some checks.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    Preconditions.checkArgument((p > 0.0 && p < 1.0),
+      "FPP should be less than 1.0 and great than 0.0");
+
+    int bits = (int)(-n * Math.log(p) / (Math.log(2) * Math.log(2)));
+
+    // Get next power of 2 if bits is not power of 2.
```
jbapple-cloudera: Use https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#highestOneBit-int-

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+  }
+
+  /**
+   * @return Bytes buffered of bloom filter.
+   */
+  public long getAllocatedSize() {
+    return getBufferedSize();
+  }
+
+  public String memUsageString(String prefix) {
```
jbapple-cloudera: I don't see any callers

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,485 @@
...
+    return String.format(
+      "%s BloomDataWriter{\n" + "%s}\n",
+      prefix, String.valueOf(bitset.length)
+    );
+  }
+
+  public static Bloom getBloomOnType(PrimitiveType.PrimitiveTypeName type,
```
jbapple-cloudera: I don't see any callers.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Yes, It can be replaced by using different version hash function which take different types.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: This can be used to construct bloom filter when reading parquet file.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: What do you mean of "A HashSet will use a good bit more than 8 heap bytes per value."?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: That makes sense, but why isn't it then used somewhere else in this patch? Does this patch not exercise the code path of reading a previously-serialized BF? It probably should.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Both Statistic and Dictionary values writer use inheritance to handle the various type T.  It seems not consistent if I use overloaded hash function. What do you think?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: I'm not sure how to say it another way. A HashSet<Long> containing N Longs will use substantially more than N*8 bytes of memory. This is because of object overhead, empty space, and other overheads associated with hash tables occupy heap space.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: Can you point me to that code?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Yes, it only contains the Bloom class, Murmur3 class and a unit test for bloom filter functionality.

Both read and write code patch is not in this patch.  I can add some read/write code path unit tests to cover this.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: The original bloomfilter value writer has two path, one is writing values to a hash set as cache, and finalize bloom filter when finalize the column chunk. Another is setting bits immediately when writing values which does not need a flush since the bitset is ready.

jbapple-cloudera: I don't know what the "original bloom filter value writer" is. This is the first one I have seen.

Please add a check here for this case, or, preferably, move the buffering BF construction out of the BF class.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: I saw the formula from impala is "bits = -k * n / log(1- pow(p, 1.0/k))". This should also apply to here right?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Statistic : https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java

Dictionary: https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: understood!

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Agreed, I just remove the elements.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: I do not think that wikipedia page supports your statement when k is fixed at 8.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: I those are justified by their use of overriding methods that take primitives (non-objects) as values, since primitives do not work with generics.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -282,22 +250,15 @@ public static int optimalNumOfBits(long n, double p) {
...
-    bits --;
-    bits |= bits >> 1;
-    bits |= bits >> 2;
-    bits |= bits >> 4;
-    bits |= bits >> 8;
-    bits |= bits >> 16;
-    bits++;
-
-    return bits;
+    return Integer.highestOneBit(bits) << 1;
```
jbapple-cloudera: Please check if bits is 0, negative, or a power of 2 first.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -348,31 +309,22 @@ public static Bloom getBloomOnType(PrimitiveType.PrimitiveTypeName type,
...
   }
 
   public static class BinaryBloom extends Bloom<Binary> {
-    final int initSlabSize = 1024;
-    final int maxSlabSize = 64*1024;
-    private CapacityByteArrayOutputStream arrayout = new CapacityByteArrayOutputStream(initSlabSize,
-      maxSlabSize, new HeapByteBufferAllocator());
-    private LittleEndianDataOutputStream out = new LittleEndianDataOutputStream(arrayout);
-
+    final int maxSlabSize = 64 * 1024;
```
jbapple-cloudera: final compile-time constants are often given names "LIKE_THIS"

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -382,12 +334,10 @@ public BinaryBloom(int size, HASH hash, ALGORITHM algorithm) {
     @Override
     public long hash(Binary value) {
       try {
-        out.writeInt(value.length());
-        value.writeTo(out);
-        out.flush();
-        byte[] encoded = BytesInput.from(arrayout).toByteArray();
-        arrayout.reset();
-        switch (bloomFilterHash) {
+        PlainValuesWriter plainValuesWriter = new PlainValuesWriter(value.length() + 4, maxSlabSize, new HeapByteBufferAllocator());
```
jbapple-cloudera: This is complex enough to deserve a test.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -250,7 +256,12 @@ public static int optimalNumOfBits(long n, double p) {
     int bits = (int)(-n * Math.log(p) / (Math.log(2) * Math.log(2)));
 
     // Get next power of 2 if bits is not power of 2.
-    return Integer.highestOneBit(bits) << 1;
+    bits = Integer.highestOneBit(bits) << 1;
+
+    if (bits < 0) {
+      bits = ParquetProperties.DEFAULT_DICTIONARY_PAGE_SIZE * 8;
```
jbapple-cloudera: Can you explain in the comments why you pick this number of bits?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,423 @@
...
+   * Calculate optimal size according to the number of distinct values and false positive probability.
+   * @param n: The number of distinct values.
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    Preconditions.checkArgument((p > 0.0 && p < 1.0),
+      "FPP should be less than 1.0 and great than 0.0");
+
+    final double M = -8 * n / Math.log(1 - Math.pow(p, 1.0 / 8));
```
cjjnjust: Given 0<p<1 => 0 < pow(p, 0.125) < 1 => log (1- pow(p, 0.125)) < 0 => M>0 or (M<0 overflow). 
 
So M<0 when result overflow according to formula, it should be set to predefined maximum number of bytes ParquetProperties.DEFAULT_MAXIMUM_BLOOM_FILTER_SIZE << 3.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  /**
+   * Construct the bloom filter with given bit set.
+   * @param bitset The given bitset to construct bloom filter.
+   * @param hash The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(byte[] bitset, HASH hash, ALGORITHM algorithm) {
+    this.bitset = bitset;
```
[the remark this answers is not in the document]
cjjnjust: elements member variable has default null values if no value assigned.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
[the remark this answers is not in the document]
cjjnjust: I checked ByteBuffer APIs again, we can skip reverseBytes from Integer since ByteBuffer wrap bytes in BIG_ENDIAN order. But the bytes streaming OR is faster than integer reverse and OR in unit test.  testBinaryBloom takes average 200ms  vs 168ms.

Following is int OR: 
int origin = ByteBuffer.wrap(bitset, bitsetIndex, 4).asIntBuffer().get();
origin |= mask[i];
ByteBuffer.wrap(bitset, bitsetIndex, 4).asIntBuffer().put(origin);

This may because of the function call overhead.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,422 @@
...
+  // Bloom filter algorithm.
+  public enum ALGORITHM {
+    BLOCK,
+  }
+
+  /**
+   * Default false positive probability value use to calculate optimal number of bits
+   * used by bloom filter.
+   */
+  public final double FPP = 0.05;
```
jbapple-cloudera: Missed review comment: 0.01 is closer to the optimal range for this FPP.

Also, new comment: Maybe call it DEFAULT_FPP, since the fpp can be different?

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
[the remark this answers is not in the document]
jbapple-cloudera: What if it is an int array rather than a byte buffer? That might more naturally match the semantics of the tiny split Bloom filter.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
```diff
@@ -0,0 +1,333 @@
...
+package org.apache.parquet.column.values.bloom;
+
+/**
+ * Murmur3 is successor to Murmur2 fast non-crytographic hash algorithms.
+ *
+ * Murmur3 32 and 128 bit variants.
+ * 32-bit Java port of https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash3.cpp#94
+ * 128-bit Java port of https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash3.cpp#255
+ *
+ * This is a public domain code with no copyrights.
```
jbapple-cloudera: Why not use the code directly and use JNI to call it?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
```diff
@@ -0,0 +1,83 @@
...
+      assertTrue(binaryBloom.find(binaryBloom.hash(Binary.fromString(strings.get(i)))));
+    }
+
+    // exist can be true in a very low probability.
+    boolean exist = binaryBloom.find(binaryBloom.hash(Binary.fromString("not exist")));
+    assertFalse(exist);
+  }
+
+  @Test
+  public void testMurmur3() throws IOException {
```
jbapple-cloudera: I would like to see some more tests on some different lengths.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
```diff
@@ -0,0 +1,83 @@
...
+      strings.add(str);
+      binaryBloom.insert(binaryBloom.hash(Binary.fromString(str)));
+    }
+
+    for(int i = 0; i<strings.size(); i++) {
+      assertTrue(binaryBloom.find(binaryBloom.hash(Binary.fromString(strings.get(i)))));
+    }
+
+    // exist can be true in a very low probability.
+    boolean exist = binaryBloom.find(binaryBloom.hash(Binary.fromString("not exist")));
```
jbapple-cloudera: I think this should include many more calls to find on different input - enough to be confident of the false positive rate.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.parquet.column.values.RandomStr;
+import org.apache.parquet.io.api.Binary;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestBloom {
```
jbapple-cloudera: I would like to see the C++ version available and integration tested before committing either the Java or C++ version

wesm: Can anyone assist with an implementation for parquet-cpp? If anyone needs help with "where to put the code" either I, @xhochy, or @majetideepak can assist

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
cjjnjust: it will introduce a c++ dependency, not sure whether current parquet-mr has any C++ dependency yet. Also I didn't see any formal release of Murmur3.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
cjjnjust: Given 0.01 FPP,  10 false positive tests can get a 0.99 ^ 10 about 0.9 probability pass.  I will update to 10 false positive tests here.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: this is used when serialize bloom filter to file.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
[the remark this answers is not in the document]
cjjnjust: Yes, it looks more semantics for the 32 bytes block. 
But it needs to convert endianness when serializing and deserializing which may need extra buffer.

parquet-column/src/main/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: when elements is empty, it return false since no bit is set in bitset. Why it returns wrong result?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
cjjnjust: What integration test you mean?  Interact with cpp version?  Such as write to file using java version and read with cpp version?

I'd like to try to implement the cpp version.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+  public static final int BYTES_PER_BUCKET = 32;
+
+  // Hash strategy used in this bloom filter.
+  public final HASH hash;
+
+  // Algorithm applied of this bloom filter.
+  public final ALGORITHM algorithm;
+
+  // The underlying byte array for bloom filter bitset.
+  private byte[] bitset;
```
jbapple-cloudera: Why not just store int[] bitset?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+   * @param numBytes number of bytes for bit set.
+   */
+  private void initBitset(int numBytes) {
+    if (numBytes < BYTES_PER_BUCKET) {
+      numBytes = BYTES_PER_BUCKET;
+    }
+
+    // Get next power of 2 if it is not power of 2.
+    if ((numBytes & (numBytes - 1)) != 0) {
+      numBytes = Integer.highestOneBit(numBytes) << 1;
```
jbapple-cloudera: This could be negative if a bit gets shifted into the highest-order position, yes?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+    this.intBuffer = ByteBuffer.wrap(bitset).asIntBuffer();
+  }
+
+  /**
+   * Write bloom filter to output stream. A bloom filter structure should include
+   * bitset length, hash strategy, algorithm, and bitset.
+   * @param out output stream to write
+   */
+  public void writeTo(OutputStream out) throws IOException {
+    Preconditions.checkArgument(bitset.length > 0, "Bloom filter bitset length should be larger than 0");
```
jbapple-cloudera: Is this possible? Java promises never to return a negative for the length of an array, right?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+      0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
+
+    int mask[] = new int[8];
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = key * SALT[i];
+    }
+
+    for (int i = 0; i < 8; ++i) {
+      mask[i] = mask[i] >>> 27;
```
jbapple-cloudera: For switching between little-endian and big-endian, I think you can subtract mask[i] from 31 after this operation.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
jbapple-cloudera: What methodology have you used to check that your Java translation computes the same values as the C++ version?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
jbapple-cloudera: yes, that is what I mean.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
```diff
@@ -0,0 +1,100 @@
...
+      String str = randomStr.get(10);
+      strings.add(str);
+      binaryBloom.insert(binaryBloom.hash(Binary.fromString(str)));
+    }
+
+    for(int i = 0; i<strings.size(); i++) {
+      assertTrue(binaryBloom.find(binaryBloom.hash(Binary.fromString(strings.get(i)))));
+    }
+
+    // exist can be true in a very low probability.
```
jbapple-cloudera: I do not think this is sufficient to test whether the FPP is around 0.01, as specified, or worse.

I would suggest trying 100,000 more random values and checking to see how many are present. You would expect around 1000, right?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
```diff
@@ -0,0 +1,333 @@
...
+
+  private static final int DEFAULT_SEED = 104729;
+
+  /**
+   * Murmur3 32-bit variant.
+   *
+   * @param data - input byte array
+   * @return - hashcode
+   */
+  public static int hash32(byte[] data) {
```
jbapple-cloudera: Why make this public if the Bloom filter implementation does not use this. Same question for other methods.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
```diff
@@ -0,0 +1,100 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestBloom {
+  @Test
+  public void testIntBloom () throws IOException {
+    Bloom bloom = new Bloom(279, Bloom.HASH.MURMUR3_X64_128, Bloom.ALGORITHM.BLOCK);
```
jbapple-cloudera: Please add a comment explaining the rationale for the number 279

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  /**
+   * Construct the bloom filter with given bit set, it is used
+   * when reconstruct bloom filter from parquet file.
+   * @param bitset The given bitset to construct bloom filter.
+   * @param hash The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(byte[] bitset, HASH hash, ALGORITHM algorithm) {
```
jbapple-cloudera: I have stated my opinion in the past on this review that it is my opinion that all public methods of a class should be tested.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,412 @@
...
+   */
+  public void flush() {
+    if (elements != null && bitset == null) {
+      initBitset(optimalNumOfBits(elements.size(), DEFAULT_FPP) / 8);
+
+      for (long hash : elements) {
+        addElement(hash);
+      }
+
+      elements.clear();
```
jbapple-cloudera: elements = null allows the GC to clean it up, I suspect.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -50,6 +50,7 @@
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
+  public static final int DEFAULT_MAXIMUM_BLOOM_FILTER_SIZE = 16 * 1024 * 1024;
```
jbapple-cloudera: "SIZE" is ambiguous; prefer "BYTES"

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Since when serializing and deserializing you will need to additional considering about the byte order and converting byte array to int array which should have some overhead.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/TestBloom.java
cjjnjust: This just a value I want to check what bitset length will be next power of 2.  I added a test for that.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
cjjnjust: I checked it manually.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: yes, I updated overflow handle code following.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: No need to switch endianness now.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: I just added write/read from outputstream in test.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: This doubles the memory requirement, and I don't think it is worth it -- doubly so since you can change the endianness with the subtraction bit below.

cjjnjust: The intBuffer just use bitset as backup array.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -50,6 +50,7 @@
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
+  public static final int DEFAULT_MAXIMUM_BLOOM_FILTER_BYTES = 16 * 1024 * 1024;
```
rdblue: The default bloom filter max is 1/8th of the default row group size?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+ * Bloom Filter algorithm is implemented using block Bloom filters from Putze et al.'s "Cache-,
+ * Hash- and Space-Efficient Bloom Filters". The basic idea is to hash the item to a tiny Bloom
+ * Filter which size fit a single cache line or smaller. This implementation sets 8 bits in
+ * each tiny Bloom Filter. Tiny bloom filter are 32 bytes to take advantage of 32-bytes SIMD
+ * instruction.
+ */
+
+public class Bloom {
+  // Hash strategy available for bloom filter.
+  public enum HASH {
```
rdblue: Nit: Should be `HashAlgorithm`.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+    MURMUR3_X64_128,
+  }
+
+  // Bloom filter algorithm.
+  public enum ALGORITHM {
```
rdblue: Nit: Should be `Algorithm` (not all caps). Is there a better name for this? It isn't really an algorithm. It is a variant of the data structure.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+   * Default false positive probability value use to calculate optimal number of bits
+   * used by bloom filter.
+   */
+  public final double DEFAULT_FPP = 0.01;
+
+  // Bloom filter data header, including number of bytes, hash strategy and algorithm.
+  public static final int HEADER_SIZE = 12;
+
+  // Bytes in a bucket.
+  public static final int BYTES_PER_BUCKET = 32;
```
rdblue: Can we use a better name than "bucket"? In the paper, each filter is called a block. That's an overused term in Hadoop, so maybe we should call it something more specific, like `FilterBlock`.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * Constructor of bloom filter, if numBytes is zero, bloom filter bitset
+   * will be created lazily and the number of bytes will be calculated through
+   * distinct values in cache.
+   * @param numBytes The number of bytes for bloom filter bitset, set to zero can
+   *                 let it calculate number automatically by using default DEFAULT_FPP.
+   * @param hash The hash strategy bloom filter apply.
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(int numBytes, HASH hash, ALGORITHM algorithm) {
```
rdblue: I doubt we are going to use this class for other bloom filter variants, so I don't see the need to pass the algorithm in. Same thing with the hash function. Since there is only one, let's avoid passing it explicitly.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+  public final ALGORITHM algorithm;
+
+  // The underlying byte array for bloom filter bitset.
+  private byte[] bitset;
+
+  // A integer array buffer of underlying bitset help setting bits.
+  private IntBuffer intBuffer;
+
+  // A cache to column distinct value (hash)
+  private Set<Long> elements;
```
rdblue: Nit: if this stores the hash of elements, then calling it `elements` is confusing.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+   * @param algorithm The algorithm of bloom filter.
+   */
+  public Bloom(ByteBuffer input, HASH hash, ALGORITHM algorithm) {
+    if (input.isDirect()){
+      bitset = new byte[input.remaining()];
+      input.get(bitset);
+    } else {
+      this.bitset = input.array();
+    }
+    this.intBuffer = ByteBuffer.wrap(bitset).asIntBuffer();
```
rdblue: Should this be explicit about the endianness used?

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+    out.write(BytesUtils.intToBytes(this.algorithm.ordinal()));
+
+    // Write bitset
+    out.write(bitset);
+  }
+
+  private int[] setMask(int key) {
+    // The block based algorithm needs 8 odd SALT values to calculate eight index
+    // of bit to set, one bit in 32-bit word.
+    final int SALT[] = {0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d,
```
rdblue: This should be private static final and at the class level, not here.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+      intBuffer.put(bucketIndex * (BYTES_PER_BUCKET / 4) + i, value);
+    }
+  }
+
+  /**
+   * Determine where an element is in set or not.
+   * @param hash the hash value of element plain encoding result.
+   * @return false if element is must not in set, true if element probably in set.
+   */
+  private boolean isContain(long hash) {
```
rdblue: Nit: Should be called `contains`

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+
+  /**
+   * Calculate optimal size according to the number of distinct values and false positive probability.
+   * @param n: The number of distinct values.
+   * @param p: The false positive probability.
+   * @return optimal number of bits of given n and p.
+   */
+  public static int optimalNumOfBits(long n, double p) {
+    Preconditions.checkArgument((p > 0.0 && p < 1.0),
+      "DEFAULT_FPP should be less than 1.0 and great than 0.0");
```
rdblue: This isn't `DEFAULT_FPP`, it is whatever FPP was passed in as `p`

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+
+  /**
+   * Compute hash for Binary value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(Binary value) {
+    try {
+      ByteBuffer plain = ByteBuffer.allocate(Integer.SIZE/Byte.SIZE);
+      plain.order(ByteOrder.LITTLE_ENDIAN).putInt(value.length());
```
rdblue: Why does the hash include the length of the string? That seems unnecessary.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * Compute hash for int value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(int value) {
+    ByteBuffer plain = ByteBuffer.allocate(Integer.SIZE/Byte.SIZE);
+    plain.order(ByteOrder.LITTLE_ENDIAN).putInt(value);
+    switch (hash) {
+      case MURMUR3_X64_128: return Murmur3.hash64(plain.array());
```
rdblue: This should hash a `ByteBuffer` instead of `byte[]`. That will require less copying for types like `Binary`.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(Binary value) {
+    try {
+      ByteBuffer plain = ByteBuffer.allocate(Integer.SIZE/Byte.SIZE);
+      plain.order(ByteOrder.LITTLE_ENDIAN).putInt(value.length());
+      ByteArrayOutputStream baos = new ByteArrayOutputStream(value.length() + 4);
+      baos.write(plain.array(), 0, 4);
+      value.writeTo(baos);
```
rdblue: This shouldn't copy. It should retrieve a byte buffer for the binary using `toByteBuffer`.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+    } catch (IOException e) {
+      throw new ParquetEncodingException("could not insert Binary value to bloom ", e);
+    }
+  }
+
+  /**
+   * Insert element to set represented by bloom bitset.
+   * @param value the value to insert into bloom filter..
+   */
+  public void insert(long value) {
```
rdblue: This requires the hash of a value, not a value.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  /**
+   * Determine whether an element exist in set or not.
+   * @param hash the element to contain.
+   * @return false if value is definitely not in set, and true means PROBABLY in set.
+   */
+  public boolean find(long hash) {
+    // Elements are in cache, flush them firstly.
+    if (elements != null && !elements.isEmpty()) {
```
rdblue: I don't think we need a class that handles reads mixed with writes. Reads and writes will be separated, so I'm a little concerned that this over-complicates the implementation.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Murmur3.java
rdblue: There are implementations of Murmur3 available elsewhere. Google guava has one, for example, that handles ints and longs directly. I don't think we should include this.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
jbapple-cloudera: Some hash functions (I don't know if Murmur3 is one of them) have trivial collisions when length is not included in the hashing

rdblue: You mean something like "abc" and "abc\x00\x00" hash to the same value?

I haven't seen any warnings about this behavior in murmur3 implementations and a quick google search doesn't turn up much. I'd prefer not to include the length because it makes the hash harder to compute, but we should either make sure we're okay with the trivial collisions or find out whether this is true.

I'm leaning toward just accepting possible collisions, depending on the definition of trivial. We expect low cardinality columns to be dictionary-encoded. High cardinality columns would be more like UUIDs or other sufficiently random values that wouldn't hit this case very much.

rdblue: This doesn't affect murmur3. The length of the buffer is XORed into the result: https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp#L141

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
cjjnjust: This need to discuss.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: guava support ByteBuffer argument from version 23. I just update pom.xml also .

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: We defined to compute plain encoding of value. So it depends on how plain encoding of binary value it is. From the PlainValueWriter.java, it store the length of Binary.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: I updated to use toByteBuffer , not sure whether this be consistent with plain encoding definition of parquet.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Looks like guava 23.0 changes quite some APIs which cause several build issues in parquet-cli.  I reverted back to use guava 20.0 to use hash64(byte[]) API firstly.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
```diff
@@ -0,0 +1,413 @@
...
+    return hashFunction.hashBytes(plain.array()).asLong();
+  }
+
+  /**
+   * Compute hash for Binary value by using its plain encoding result.
+   * @param value the value to hash
+   * @return hash result
+   */
+  public long hash(Binary value) {
+      return hashFunction.hashBytes(value.toByteBuffer().array()).asLong();
```
cjjnjust: I think this is still a open since we agreed to use plain encoding input. In parquet thrift definition,  BYTE_ARRAY is encoded with length, and FIXED_LEN_BYTE_ARRAY is encoded without length.  In PlainValuesWriter.java, it stored length for Binary.

Do we need to still use plain encoding input for Binary?  @rdblue @jbapple-cloudera

rdblue: We don't need to include the length.

Also, this shouldn't use `.array()` without checking whether the buffer is direct or on heap, and without supplying an offset because the ByteBuffer may point to the middle of its backing array. Here's what I'm using in another project:

```java
    // MURMUR3 is the configured hash function from guava; this uses the 32-bit variant
    public int hash(ByteBuffer value) {
      if (value.hasArray()) {
        return MURMUR3.hashBytes(value.array(),
            value.arrayOffset() + value.position(),
            value.arrayOffset() + value.remaining()).asInt();
      } else {
        byte[] copy = new byte[value.remaining()];
        value.get(copy);
        return MURMUR3.hashBytes(copy).asInt();
      }
    }
```

There should also be test cases for the following:

* Direct byte buffer
* Direct byte buffer with non-zero `position`
* Heap byte buffer
* Heap byte buffer with non-zero `arrayOffset`
* Heap byte buffer with non-zero `position`

cjjnjust: The ByteBuffer is from Binary API which already consider things you mentioned.

The same as (getBytes/getUnsafeBytes[no copy]).  Looks like I can use getUnsafeBytes to avoid copy, since it doesn't change any value.

rdblue: This should use the ByteBuffer and add tests for the cases that I listed. ByteBuffers will require fewer copies. Otherwise, this will need to be rewritten before actually using it to produce bloom filters.

cjjnjust: I understand ByteBuffer should avoid copies, but Murmur3 from guava 20.0 only support byte[] parameter.  We can change to use ByteBuffer when update to guava23.0.

The difference from your reference code is here we use Binary as input parameter, and we should trust the APIs from Binary class, shouldn't we?  If it accept a ByteBuffer parameter, we should add tests you list accordingly.

In addition, in you reference code, why you add arrayOffset to remaining? remaining should be (limit -position) and irrelevant with arraryOffset, otherwise the length changes, am I wrong?

rdblue: Trusting an API for correctness and using the most efficient method from that API are two different things. If you use `toByteArray`, your code will often copy data it doesn't need to, which is why it will need to be replaced before actually using this. If Guava doesn't support ByteBuffer, is there another implementation that does? Can we update Guava?

cjjnjust: Yes, we can update Guava to 23.0, it needs some minor changes to fix build issues.  Please have a look the latest commit.  

I also update maven-shade-plugin version since it show ArrayIndexOutOfBoundsException when creating shaded jar for "com/google/common/cache/LocalCache$EntrySet.class".

cjjnjust:
Hi @wesm 
@winningsix  will work on parquet-cpp part.

BenoitHanotte:
I am quite interested in this topic, is there any plan to move forward with this?

As we already depend on guava 18.0 in the parquet-hadoop module, why don't we simply use the BloomFilter class it provides? (https://google.github.io/guava/releases/18.0/api/docs/com/google/common/hash/BloomFilter.html)
It is very efficient and well tested. If it fit our needs I would suggest not bothering re-implementing the filters as it is not a trivial task.

cjjnjust:
Hi @winningsix 
any update on this?

cjjnjust:
Hi developers

Sorry for no update long time for bloom filter topic due to transition, now I'm back to move this forward start from this rebase.

No obviously change in this commit, I just remove unrelated changes.

@rdblue @jbapple-cloudera Could you please kindly to have a look?
@wesm As you point out, I had a PR in parquet-cpp also, could you please also have a look?

Many thanks

cjjnjust:
Hi @rdblue 
I have updated code according your comments before and recently I rebased code also. I see there is still changes requests from this PR page and I checked them one by one, I think almost all comments have related update. So could you please kindly take a look again? 

Thanks in advanced.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Done.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Changed to private and provide another constructor.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Removed cache.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Yes, updated to use little endian.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Done.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Changed to name "find".

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Done.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Done.

parquet-column/src/test/java/org/apache/parquet/column/values/bloom/Bloom.java
cjjnjust: Removed this cache logic to keep simple.

cjjnjust:
ping @rdblue

cjjnjust:
Hi @rdblue, do you get time to have a look?

jbapple-cloudera:
Can this be closed in favor of https://github.com/apache/parquet-mr/pull/425?

cjjnjust:
Close this due to duplicated one : https://github.com/apache/parquet-mr/pull/521

PR #428, opened 2017-10-17
PARQUET-1141: Fix field ID handling

rdblue:
There are two places where field IDs are dropped:
* Map and list type builders were not passing IDs when building
* ParquetMetadataConverter was not writing field IDs or reading the ID for root schemas


danielcweeks:
LGTM

+1

PR #430, opened 2017-10-17
PARQUET-1143: Update to Parquet format 2.4.0.

rdblue:
This adds new compression codecs that are required by format 2.4.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,4 +32,3 @@ env:
 
 install: mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
 script: mvn test -P $HADOOP_PROFILE
-dist: precise
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
 
 env:
   - HADOOP_PROFILE=hadoop-1 TEST_CODECS=uncompressed
-  - HADOOP_PROFILE=default TEST_CODECS=gzip,snappy
+  - HADOOP_PROFILE=default TEST_CODECS=gzip,snappy,brotli
 
 install: mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
 script: mvn test -P $HADOOP_PROFILE
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ before_install:
   - cd ..
 
 env:
-  - HADOOP_PROFILE=hadoop-1 TEST_CODECS=uncompressed
-  - HADOOP_PROFILE=default TEST_CODECS=gzip,snappy,brotli
+  - HADOOP_PROFILE=default TEST_CODECS=uncompressed,brotli
+  - HADOOP_PROFILE=default TEST_CODECS=gzip,snappy
 
 install: mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
 script: mvn test -P $HADOOP_PROFILE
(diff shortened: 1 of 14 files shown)


parquet-benchmarks/pom.xml
```diff
@@ -73,6 +79,11 @@
 
   <build>
     <plugins>
+      <!-- This module disables semver checks because it is not a public API.
```
zivanfi: The purpose of adding this plugin and commenting out at the same time is not clear to me. Could you please elaborate?

parquet-hadoop/pom.xml
```diff
@@ -79,6 +85,11 @@
       <artifactId>commons-pool</artifactId>
       <version>1.6</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.rdblue</groupId>
```
zivanfi: com.github.rdblue seems like a temporary groupId to me, do you want to add this dependency as it is to parquet-format?

parquet-benchmarks/pom.xml
rdblue: The plugin is configured for all projects to avoid copying, but this one is internal and shouldn't be checked.

parquet-hadoop/pom.xml
rdblue: This is how jitpack.io creates coordinates.

rdblue:
@julienledem, can you take a look? I removed hadoop-1 from the tests and made brotli-codec an optional dependency.

.travis.yml
```diff
@@ -22,12 +22,12 @@ before_install:
...
   - cd thrift-0.7.0
   - chmod +x ./configure
-  - ./configure --disable-gen-erl --disable-gen-hs --without-php --without-nodejs --without-ruby --without-haskell --without-erlang
+  - ./configure --disable-gen-erl --disable-gen-hs --without-ruby --without-haskell --without-erlang --without-php
   - sudo make install
   - cd ..
 
 env:
-  - HADOOP_PROFILE=hadoop-1 TEST_CODECS=uncompressed
+  - HADOOP_PROFILE=default TEST_CODECS=uncompressed,brotli
```
zivanfi: While I don't have any objections to removing the hadoop-1 profile, please note that the hadoop-1 travis builds don't fail any more, so it's not strictly necessary.

rdblue: As long as we are in agreement that we don't need it any more, then I'm for removing it from testing.

rdblue:
Thanks for reviewing, @zivanfi and @danielcweeks! I'll merge this soon.

scottcarey:
This is great!   I would love to test out writing some parquet files using zstd compression.

It appears I can not do so without a parquet release however, containing this work.

Am I mistaken? Is there a way to manually supply parquet-format 2.4 and combine it with released versions of parquet-avro/mr/etc and spark and output zstd files?

If not, what is the rough ETA on a 1.9.1 or 1.10.0 release of parquet that would unlock zstd compression?

rdblue:
I'd like to get 1.10.0 out in the next week or two.

scottcarey:
Anything that I can help with to get 1.10.0 out?  I'll be happy to test out any RCs on my use case.  I'd rather spend time helping with 1.10.0 than testing a custom built version, but I may be forced to build and test a custom version if an official release with zstd available takes too long.

scottcarey:
FWIW, I tested out the current master code, overriding the version in my spark projects.  I could not output zstandard parquet files because spark-sql's `ParquetOptions` class intercepts the config strings and maps them to a `CompressionCodecName` in parquet-hadoop, rather than just delegating the name lookup to parquet-hadoop, so it does not understand the string 'zstd'.

This coupling means that using this from spark will require a new version of spark-sql.  Honestly, the code here should be responsible for converting from a simple name to the codec, not spark.  Then one could upgrade only the parquet version and gain access to new compression codecs without recompiling/releasing spark.

rdblue:
My goal is to get an RC out tomorrow.

There are a few changes needed for Spark to work. The vectorized readers need an update in addition to the property changes for zstd and brotli. I'll open a PR once the release is out.

scottcarey:
Yeah, I looked a little further into what is needed on the Spark side too.   Part way in modifying the vectorized readers to use method signatures that use ByteBufferInputStream rather than (byte[], offset), I hit a spot where they called back into code here that did not take a ByteBufferInputStream.

It looks like changes on both sides are needed.

I think that whole area of code would work better if coded with a DataInput interface instead.  You can wrap a ByteBufferInputStream in an DataInputStream, and get free (and decently efficient but not amazing) tools for reading littleEndian ints, etc.  DataInputStream will be quite a bit faster than calling read() 4 times in a row and constructing the int by hand, though its technique of maintaining a small buffer for reading primitives can be emulated.

rdblue:
@scottcarey, you don't need to update Spark, I have a branch with it updated that we're already running in production.

PR #433, opened 2017-11-02
PARQUET-1115: Warn users when misusing parquet-tools merge

nandorKollar:


parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -57,6 +60,13 @@ public MergeCommand() {
   }
 
   @Override
+  public String getCommandDescription() {
+    return "Merges multiple Parquet files into one. " +
+      "The command doesn't merge row groups, just places one after the other. " +
+      "When used to merge many small files, the resulting file will still contain small row groups.";
```
zivanfi: I would add ", which usually leads to bad query performance."

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -65,14 +75,28 @@ public void execute(CommandLine options) throws Exception {
...
+          input.getFileSystem(conf).getFileStatus(input).getLen());
+        tooSmallFilesMerged = true;
+      }
+
       writer.appendFile(conf, input);
     }
+
+    if (tooSmallFilesMerged) {
+      out.println("Warning: you merged too small files. " +
+        "Despite the size of the merged file is bigger, it STILL contains small row groups, thus you don't have the advantage of big row groups!");
```
zivanfi: I would use stronger wording after the last comma, like "which usually leads to bad query performance." (Additionally, I'm not a native speaker either, but I believe you either need a "that" after "Despite" or use some other word, for example "Although".)

nandorKollar:
@zivanfi thanks for you review, changed the wording accordingly.

nandorKollar:
@zivanfi removed the whitespace-only changes

NickCrews:
It might be nice if we actually suggested an alternative instead of just saying "don't do this."

You can see my solution at https://gist.github.com/NickCrews/7a47ef4083160011e8e533531d73428c.

PR #440, opened 2017-12-08
PARQUET-1174: Concurrent scan micro benchmarks

takeshi-yoshimura:
parquet-benchmarks only contain read and write benchmarks with a single thread.
I add concurrent Parquet file scans like typical data-parallel computing.

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BenchmarkConstants.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BenchmarkConstants.java
@@ -39,4 +39,6 @@ public class BenchmarkConstants {
   public static final int PAGE_SIZE_8M = 8 * 1024 * 1024;
 
   public static final int DICT_PAGE_SIZE = 512;
+
+  public static final int SIXTEEN_THREADS = 16;
 }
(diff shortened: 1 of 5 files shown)


parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BenchmarkConstants.java
```diff
@@ -39,4 +39,6 @@
   public static final int PAGE_SIZE_8M = 8 * 1024 * 1024;
 
   public static final int DICT_PAGE_SIZE = 512;
+
+  public static final int SIXTEEN_THREADS = 16;
```
zivanfi: If we were to change this value, we would have to rename the constant as well. I would suggest using a more generic name instead, like THREAD_COUNT.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ReadBenchmarks.java
```diff
@@ -103,4 +111,61 @@ public void read1MRowsDefaultBlockAndPageSizeGZIP(Blackhole blackhole)
...
+    }
+
+    for (Thread t: threads) {
+      t.join();
+    }
+  }
+
+  @Benchmark
+  @Threads(1)
+  public void readUncompressed16Thread(Blackhole blackhole)
```
zivanfi: Could you remove the specific thread count (16) from the name of these methods? You could use Multi instead (e.g., readUncompressedMultiThread).

takeshi-yoshimura:
I renamed them. thank you!

PR #444, opened 2018-01-05
PARQUET-1185: TestBinary#testBinary unit test fails after PARQUET-1141

nandorKollar:


rdblue:
@nandorKollar, thanks for fixing this! The tests were passing on the PR I merged for PARQUET-1141, I'm not sure what happened there.

Can you write a quick summary of what this changes and why the test was failing? I'm not sure what this commit does other than fix the tests. We generally like to have that in the PR description that will be the final commit log, if not also in the PR's commit log for reviewers. Thank you!

nandorKollar:
@rdblue thanks for taking a look at the PR, I added some description to the change in my second commit (and also deleted an unused import). Could you please have a look at it?

parquet-thrift/src/test/java/org/apache/parquet/hadoop/thrift/TestBinary.java
```diff
@@ -81,9 +80,13 @@ public void testBinary() throws IOException {
   private void assertSchema(ParquetMetadata parquetMetadata) {
     List<Type> fields = parquetMetadata.getFileMetaData().getSchema().getFields();
     assertEquals(2, fields.size());
-    assertEquals(Types.required(PrimitiveType.PrimitiveTypeName.BINARY).named("s"), fields.get(0));
```
rdblue: Why not add the ID to this type instead of verifying each field individually?

zivanfi: I recommended Nandor to check the fields separately after I asked him what he would like to assert. Since he wanted to check name, primary type and logical type, but the equals method checks the name, primary type and ID (so the logical type is not checked, but the ID is), I thought it would be cleaner to assert on the actual fields we care about.

If you prefer, we can also do an equals check on the object (with an ID added to the reference value) and an additional separate check for the logical type field.

rdblue: I don't see a reason not to use normal type equality for this check.

nandorKollar: Done, added IDs for the expected type.

gszadovszky:
@rdblue, just realized that the `hashCode()` of `Type` takes `originalType` into account while `equals(Type)` does not check it. It breaks the specification of `hashCode` but it is not clear which one should be the expected behavior. Moreover, fixing this would be a breaking change.
What do you think?

zivanfi:
Nice catch, Gabor!

> Moreover, fixing this would be a breaking change.

On the other hand, the current behaviour is non-deterministic. If two instances are equal according to equals() but have different logical types, then both of the following cases are possible:

- The two objects may accidentally have the same hash, which is consistent with the equals() method but is pure coincidence and has a very low probablility.
- The two objects may have different hashes and end up in different buckets of a hash map, leading to a situation where we can't find a value in the hash map in spite that it's equals() would return true.

Since one can not rely on the resulting behaviour, I don't consider fixing this a true breaking change, since it only changes an already non-deterministic behaviour.

zivanfi:
I created PARQUET-1191 to track the hash-equals inconsistency.

PR #446, opened 2018-01-05
PARQUET-1183 AvroParquetWriter needs OutputFile based Builder

wernerdaehn:
https://issues.apache.org/jira/browse/PARQUET-1183

AvroParquetWriter needs OutputFile based Builder


rdblue:
Looks like this contains unrelated changes to schemas. Could you clean it up to just include the OutputFile builder changes?

PR #447, opened 2018-01-11
PARQUET-869 Configurable min/max record counts for block size check

pradeepg26:
Make min/max record counts for block size check are no longer hard coded inside InternalParquetRecordWriter.


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -102,6 +103,8 @@ private void initStore() {
     MessageColumnIO columnIO = new ColumnIOFactory(validating).getColumnIO(schema);
     this.recordConsumer = columnIO.getRecordWriter(columnStore);
     writeSupport.prepareForWrite(recordConsumer);
+    System.out.println(String.format("Created ParquetWriter with [%d, %d] for block size checks. Estimation(%s). BlockSize(%d)",
```
rdblue: Please use SLF4J for logging. You can see examples in the rest of the library.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -142,13 +145,23 @@ private void checkBlockSizeReached() throws IOException {
         LOG.info("mem size {} > {}: flushing {} records to disk.", memSize, nextRowGroupSize, recordCount);
         flushRowGroupToStore();
         initStore();
-        recordCountForNextMemCheck = min(max(MINIMUM_RECORD_COUNT_FOR_CHECK, recordCount / 2), MAXIMUM_RECORD_COUNT_FOR_CHECK);
+        if (estimateNextBlockSizeCheck) {
```
rdblue: Do you need this setting, or is it included to match the page size checking? I don't think we would ever want to set it to true because it is important to be under the max page size.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -142,9 +142,12 @@
   public static final String MEMORY_POOL_RATIO    = "parquet.memory.pool.ratio";
   public static final String MIN_MEMORY_ALLOCATION = "parquet.memory.min.chunk.size";
   public static final String MAX_PADDING_BYTES    = "parquet.writer.max-padding";
-  public static final String MIN_ROW_COUNT_FOR_PAGE_SIZE_CHECK = "parquet.page.size.row.check.min";
-  public static final String MAX_ROW_COUNT_FOR_PAGE_SIZE_CHECK = "parquet.page.size.row.check.max";
-  public static final String ESTIMATE_PAGE_SIZE_CHECK = "parquet.page.size.check.estimate";
+  public static final String MIN_ROW_COUNT_FOR_PAGE_SIZE_CHECK = "parquet.page.size_row_check_min";
```
rdblue: The page properties should not change, and the row group properties should match the naming convention for pages.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -48,8 +48,11 @@
   public static final boolean DEFAULT_IS_DICTIONARY_ENABLED = true;
   public static final WriterVersion DEFAULT_WRITER_VERSION = WriterVersion.PARQUET_1_0;
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
-  public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
-  public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
+  public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_BLOCK_SIZE_CHECK = true;
+  public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_PAGE_SIZE_CHECK = 100;
+  public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_PAGE_SIZE_CHECK = 10000;
+  public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_BLOCK_SIZE_CHECK = 100;
```
rdblue: Instead of block, please use "row group". Block is an overused term that we are avoiding in the public API, except for those places where it already appears.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -42,14 +42,11 @@
...
 
-  private static final int MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
-  private static final int MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
-
   private final ParquetFileWriter parquetFileWriter;
   private final WriteSupport<T> writeSupport;
   private final MessageType schema;
   private final Map<String, String> extraMetaData;
-  private final long rowGroupSize;
+  private final boolean estimateNextBlockSizeCheck;
```
rdblue: Was rowGroupSize not used? Why was it removed?

rdblue:
Thanks for doing this, @pradeepg26! We've occasionally had use cases where it would have been nice. My two main concerns are the unnecessary boolean controlling whether or not to estimate when to do the next check and the naming. Naming should use "row group" instead of "block".

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
pradeepg26: Oops... missed removing this statement.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
pradeepg26: I wrote this code over a year ago... I don't remember, but I'll check. I think it was an unused variable.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
pradeepg26: Will do.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
pradeepg26: Will do.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
pradeepg26: Oops... it was changed for internal reasons... forgot to change it back.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
pradeepg26: Working on the comments @rdblue... do you think instead of a min/max record count for row group size checks, we should check every n rows?

pradeepg26: I've taken a closer look at this code. The only change with the `estimateNextBlockSizeCheck` is that it uses a static `min` number of records before performing the next size check if disabled.

If `estimateNextBlockSizeCheck` is enabled, then existing behavior is preserved where we try to estimate when to perform the next size check by performing the check at the halfway point of the predicted number of records (unless it's larger than the max).

So, I added the `estimateNextBlockSizeCheck` to keep it consistent with page size checking.

rdblue: Consistency here is fine. I'm not sure we will use it, but I can certainly think of valid use cases for it.

pwais:
bump? note disclosure on https://eng.uber.com/petastorm/

rdblue:
@pwais, this was replaced by #470 that includes updates for problems in this PR. Unfortunately, other committers decided they did not want to change internal APIs so it was not committed.

pwais:
Thank you!

PR #448, opened 2018-01-15
PARQUET-1170: Logical-type-based toString for proper representeation in tools/logs

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Override
+    String stringifyNotNull(Binary value) {
+      ByteBuffer buffer = value.toByteBuffer();
+      StringBuilder builder = new StringBuilder(2 + buffer.remaining() * 2);
+      builder.append(BINARY_HEXA_PREFIX);
+      for (int i = buffer.position(), n = buffer.limit(); i < n; ++i) {
+        byte b = buffer.get(i);
+        builder.append(digits[(b >>> 4) & 0x0F]);
```
zivanfi: Have you considered using String.format("%02X", b)?

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
```diff
@@ -253,7 +257,7 @@ public void testFloatMinMax() {
     assertEquals(statsFromBytesMinMax.getMin(), Float.MIN_VALUE, 1e-10);
 
     // Test print formatting
-    assertEquals(stats.toString(), String.format("min: %.5f, max: %.5f, num_nulls: %d", 0.00010, 553.59998, 0));
+    assertEquals("min: 1.0E-4, max: 553.6, num_nulls: 0", stats.toString());
```
zivanfi: These asserts were using String.format for the expected output intentionally. Earlier expected values used to be hard-coded as string literals (e.g, "0.5"), but since the textual representation depends on the locale, tests used to fail on locales that use comma as a decimal separator (e.g., "0,5").

Float.toString() and similar calls that you use for stringifying are locale-aware, therefore you should also use String.format for the expected value. Alternatively, you may change the stringifying logic to work in a locale-agnostic way.

Since this occurs many times below, I suggest you double-check your tests by running them using LANG=hu_HU or LC_NUMERIC=hu_HU LC_MEASUREMENT=hu_HU.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
gszadovszky: The performance overhead for using `String.format` for each byte elements would be significant. A don't think the exchanging of the existing two lines (one for the upper and one for the lower bits) to one would worth it.

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
zivanfi: You're right. Thanks for adding unit tests anyway.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -0,0 +1,358 @@
...
+
+  static final PrimitiveStringifier INTERVAL_STRINGIFIER = new BinaryStringifierBase("INTERVAL_STRINGIFIER") {
+    @Override
+    String stringifyNotNull(Binary value) {
+      if (value.length() != 12) {
+        return BINARY_INVALID;
+      }
+      ByteBuffer buffer = value.toByteBuffer();
+      int pos = buffer.position();
+      String months = UNSIGNED_STRINGIFIER.stringify(buffer.getInt(pos));
```
zivanfi: I don't know whether intentionally or by mistake, but [parquet-format](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#interval) specifies intervals to be little-endian. @rdblue, @julienledem, do you know whether intervals really should be little-endian?

PR #449, opened 2018-01-15
org.apache.commons.codec.binary.Base64     clash  

lhyundeadsoul:
The class "org.apache.commons.codec.binary.Base64"  in parquet-tools.jar  version 1.9.0 is same at package and class name to which is in commons-codec.jar(all version)  ,but its implement is much older. 
That makes a jar-hell when both  parquet-tools.jar and commons-codec.jar are used . 
e.g. 
```
Exception in thread "main" java.lang.NoSuchMethodError: org.apache.commons.codec.binary.Base64.encodeBase64String([B)Ljava/lang/String;
```


Fokko:
Commons codec isn't used anymore, closing this for now. Thanks for the PR.

PR #436, opened 2017-11-22
PARQUET-1065: Deprecate type-defined sort ordering for INT96 type.

zivanfi:

PR #454, opened 2018-01-25
PARQUET-1065: Deprecate type-defined sort ordering for INT96 type.

zivanfi:


zivanfi:
Already approved in #436, but created a new PR before merging for technical reasons.

PR #457, opened 2018-02-14
PARQUET-1215: Add getFooter to ParquetWriter.

rdblue:
This adds getFooter to ParquetWriter, which will return the file footer that was written after the file is closed.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -121,6 +121,9 @@ public class ParquetFileWriter {
   private long currentChunkFirstDataPage;         // set in startColumn (out.pos())
   private long currentChunkDictionaryPageOffset;  // set in writeDictionaryPage
 
+  // set when end is called
+  private ParquetMetadata footer = null;
+
   /**
    * Captures the order in which methods should be called
    *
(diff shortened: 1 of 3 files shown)


rdblue:
@zivanfi, can you look at this?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -670,7 +673,7 @@ private static void copy(SeekableInputStream from, PositionOutputStream to,
   public void end(Map<String, String> extraMetaData) throws IOException {
     state = state.end();
     LOG.debug("{}: end", out.getPos());
-    ParquetMetadata footer = new ParquetMetadata(new FileMetaData(schema, extraMetaData, Version.FULL_VERSION), blocks);
+    this.footer = new ParquetMetadata(new FileMetaData(schema, extraMetaData, Version.FULL_VERSION), blocks);
```
zivanfi: (nit, optional) "this." is superfluous, but doesn't hurt either.

rdblue: Using `this` signals that this sets an instance field, not a local. I think it helps readability when not in an IDE.

zivanfi: Yes, I agree, but the next line contains pure {{footer}} and it initally mislead me to believe that that is a different {{footer}}.

rdblue: I think the rationale for this style is that when you see an access, you go looking for where the variable was assigned. That assignment should be clearly a local or instance field. Not using `this` for access is done because it is superfluous like you say.

zivanfi: I see, thanks for the explanation. SGTM.

PR #458, opened 2018-02-19
PARQUET-1217: Incorrect handling of missing values in Statistics

gszadovszky:


parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -166,12 +174,17 @@ private boolean hasNulls(ColumnChunkMetaData column) {
       return isAllNulls(meta);
     }
 
-    if (hasNulls(meta)) {
+    if (stats.isNumNullsSet() && hasNulls(meta)) {
```
zivanfi: (optional) You might consider adding the "`stats.isNumNullsSet()`" clause to the body of `hasNulls()` itself, but I don't really have a strong opinion on this.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -401,17 +402,20 @@ private static boolean isMinMaxStatsSupported(PrimitiveType type) {
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal
       (String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
     // create stats object based on the column type
-    org.apache.parquet.column.statistics.Statistics stats = org.apache.parquet.column.statistics.Statistics.createStats(type);
+    StatisticsBuilder builder = org.apache.parquet.column.statistics.Statistics.getBuilder(type);
```
zivanfi: (optional, nit): I think `stats` would be more descriptive regarding the purpose of this variable, even if it is a builder. Alternatively, `statsBuilder`.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/statisticslevel/TestStatisticsFilter.java
```diff
@@ -87,13 +89,16 @@ private static ColumnChunkMetaData getDoubleColumnMeta(DoubleStatistics stats, l
 
   private static final IntStatistics intStats = new IntStatistics();
   private static final IntStatistics nullIntStats = new IntStatistics();
+  private static final org.apache.parquet.column.statistics.Statistics<?> missingMinMaxIntStats = org.apache.parquet.column.statistics.Statistics
```
zivanfi: Not only man and max are missing from this one, but `numNulls` as well. Would `missingIntStats` be a better name? Alternatively, set `numNulls`.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/statisticslevel/TestStatisticsFilter.java
```diff
@@ -105,6 +110,9 @@ private static ColumnChunkMetaData getDoubleColumnMeta(DoubleStatistics stats, l
       getIntColumnMeta(nullIntStats, 177L), // column of all nulls
       getDoubleColumnMeta(doubleStats, 177L));
 
+  private static final List<ColumnChunkMetaData> missingMinMaxColumnMetas = Arrays.asList(
+      getIntColumnMeta(missingMinMaxIntStats, 177L),        // missing min/max, no null values => stats is empty
```
zivanfi: Please clarify: Is it really the lack of null values or the lack of `numNulls`?

parquet-hadoop/src/test/java/org/apache/parquet/filter2/statisticslevel/TestStatisticsFilter.java
```diff
@@ -297,6 +329,26 @@ public boolean keep(Integer value) {
     }
   }
 
+  public static class AllPositiveUdp extends UserDefinedPredicate<Double> {
+    @Override
+    public boolean keep(Double value) {
+      if (value == null) {
```
zivanfi: Should/could this be an assert instead?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -145,24 +145,24 @@ public void testWriteRead() throws Exception {
...
+    w.writeDataPage(2, 4, BytesInput.from(BYTES1), EMPTY_STATS, BIT_PACKED, BIT_PACKED, PLAIN);
+    w.writeDataPage(3, 4, BytesInput.from(BYTES1), EMPTY_STATS, BIT_PACKED, BIT_PACKED, PLAIN);
     w.endColumn();
     long c1Ends = w.getPos();
     w.startColumn(C2, 6, CODEC);
     long c2Starts = w.getPos();
-    w.writeDataPage(2, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
-    w.writeDataPage(3, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
-    w.writeDataPage(1, 4, BytesInput.from(BYTES2), STATS2, BIT_PACKED, BIT_PACKED, PLAIN);
+    w.writeDataPage(2, 4, BytesInput.from(BYTES2), EMPTY_STATS, BIT_PACKED, BIT_PACKED, PLAIN);
```
zivanfi: I wonder what purpose `STATS1` and `STATS2` might have served in the original code.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
gszadovszky: It would not work correctly because `hasNulls` is called negated as well.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/statisticslevel/TestStatisticsFilter.java
gszadovszky: It is more an UDP implementation for testing purposes than the test itself. In addition, we have to return a boolean even if an assertion is added.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
gszadovszky: I have no idea, why we had two Statistics objects for the same purpose. I think, it is more readable this way.

rdblue:
@gszadovszky, can you update the description with a summary of the problem and how this fixes it? It's easier to review a patch if I know the general idea behind it.

zivanfi:
@rdblue, until Gabor updates the description, you can read about the error in the JIRA: https://issues.apache.org/jira/browse/PARQUET-1217

rdblue:
Have any engines actually written stats with `num_nulls` set, `min` and `max` not set, and `num_nulls != num_values`? That's the case that this is fixing, right? I don't think Parquet MR will write any stats if min and max are not written.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
rdblue: +1

I don't remember either.

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -35,6 +35,44 @@
  */
 public abstract class Statistics<T extends Comparable<T>> {
 
+  /**
+   * Builder class to build Statistics objects. Used to read the statistics from the Parquet file.
+   */
+  public static class StatisticsBuilder {
```
rdblue: Why add this to the public API?

Because this is nested, it should be `Statistics.Builder`, not `Statistics.StatisticsBuilder`.

gszadovszky: It needs to be public because `ParquetMetadataConverter` uses it. The concept is that you are only able to initialize `num_nulls` to `-1` (meaning it is not set in the file) by using the builder (read path). If you are creating a new `Statistics` object in any other way (write path) `num_nulls` will be initialized to `0`.
I'll rename it to Builder.

gszadovszky:
Impala started to write the new `min_value`/`max_value` with `num_nulls` set. In the same CDH release we use a parquet-mr release where the handling of the new `min_value`/`max_value` is not backported therefore, we are seeing only unset `min` and `max` values with set `num_nulls`.
In addition, there was an Impala release where no `num_nulls` were written which is correct from the parquet-format point-of-view but in parquet-mr we were handled unset `num_nulls` the same way as the `num_nulls` would be `0`.

PR #459, opened 2018-02-21
PARQUET-952: Avro union with single type fails with 'is not a group'

nandorKollar:


michaelandrepearce:
Any update on this getting merged and in a release? We have his the same issue.

PR #461, opened 2018-03-13
PARQUET-1246: Ignore float/double statistics in case of NaN

gszadovszky:
Because of the ambigous sorting order of float/double the following changes made at the reading path of the related statistics:
- Ignoring statistics in case of it contains a NaN value.
- Using -0.0 as min value and +0.0 as max value independently from which 0.0 value was saved in the statistics.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
@@ -73,6 +73,70 @@ public abstract class Statistics<T extends Comparable<T>> {
     }
   }
 
+  // Builder for FLOAT type to handle special cases of min/max values like NaN, -0.0, and 0.0
+  private static class FloatBuilder extends Builder {
+    public FloatBuilder(PrimitiveType type) {
+      super(type);
+      assert type.getPrimitiveTypeName() == PrimitiveTypeName.FLOAT;
+    }
+
+    @Override
+    public Statistics<?> build() {
+      FloatStatistics stats = (FloatStatistics) super.build();
+      if (stats.hasNonNullValue()) {
+        Float min = stats.genericGetMin();
+        Float max = stats.genericGetMax();
+        // Drop min/max values in case of NaN as the sorting order of values is undefined for this case
+        if (min.isNaN() || max.isNaN()) {
+          stats.setMinMax(0.0f, 0.0f);
+          ((Statistics<?>) stats).hasNonNullValue = false;
+        } else {
+          // Updating min to -0.0 and max to +0.0 to ensure that no 0.0 values would be skipped
+          if (min == 0.0f) {
+            stats.setMinMax(-0.0f, max);
+            min = -0.0f;
+          }
+          if (max == -0.0f) {
+            stats.setMinMax(min, 0.0f);
+          }
+        }
+      }
+      return stats;
+    }
+  }
+
+  // Builder for DOUBLE type to handle special cases of min/max values like NaN, -0.0, and 0.0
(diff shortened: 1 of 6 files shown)


parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -73,6 +73,70 @@ public Builder withNumNulls(long numNulls) {
...
+
+    @Override
+    public Statistics<?> build() {
+      FloatStatistics stats = (FloatStatistics) super.build();
+      if (stats.hasNonNullValue()) {
+        Float min = stats.genericGetMin();
+        Float max = stats.genericGetMax();
+        // Drop min/max values in case of NaN as the sorting order of values is undefined for this case
+        if (min.isNaN() || max.isNaN()) {
+          stats.setMinMax(0.0f, 0.0f);
```
zivanfi: This seems unnecessary. On the other hand, this is the default value when it's unset, so I'm fine with leaving this.

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -73,6 +73,70 @@ public Builder withNumNulls(long numNulls) {
...
+          ((Statistics<?>) stats).hasNonNullValue = false;
+        } else {
+          // Updating min to -0.0 and max to +0.0 to ensure that no 0.0 values would be skipped
+          if (min == 0.0f) {
```
zivanfi: +0 and -0 are equal according to ==. Please use Float.compareTo() instead.

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -73,6 +73,70 @@ public Builder withNumNulls(long numNulls) {
...
+            stats.setMinMax(-0.0f, max);
+            min = -0.0f;
```
zivanfi: (nit) In my opinion,
```
min = -0.0f;
stats.setMinMax(min, max);
```
would be cleaner. Similar for the max case (although max is not used any more in this function).

parquet-column/src/main/java/org/apache/parquet/column/statistics/Statistics.java
```diff
@@ -73,6 +73,70 @@ public Builder withNumNulls(long numNulls) {
...
+
+  // Builder for DOUBLE type to handle special cases of min/max values like NaN, -0.0, and 0.0
+  private static class DoubleBuilder extends Builder {
+    public DoubleBuilder(PrimitiveType type) {
+      super(type);
+      assert type.getPrimitiveTypeName() == PrimitiveTypeName.DOUBLE;
+    }
+
+    @Override
+    public Statistics<?> build() {
```
zivanfi: Same comments as for the float case.

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
```diff
@@ -637,4 +648,106 @@ private void testMergingStringStats() {
...
+    stats = builder.withNumNulls(42).withMin(minBytes).withMax(maxBytes).build();
+    assertFalse(stats.isEmpty());
+    assertTrue(stats.isNumNullsSet());
+    assertTrue(stats.hasNonNullValue());
+    assertEquals(42, stats.getNumNulls());
+    assertEquals(min, stats.genericGetMin());
+    assertEquals(max, stats.genericGetMax());
+  }
+
+  @Test
```
zivanfi: Please add test case for -0 and +0.

PR #463, opened 2018-03-29
PARQUET-1253: Support for new logical type representation

nandorKollar:
This PR implements the new logical type representation in parquet-mr which is already available in parquet-format. Reviews are welcome!

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
@@ -88,7 +88,6 @@ public class GroupType extends Type {
    * @param fields the contained fields
    * @param id the id of the field
    */
-  @Deprecated
   GroupType(Repetition repetition, String name, OriginalType originalType, List<Type> fields, ID id) {
     super(name, repetition, originalType, id);
     this.fields = fields;
--- a/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -47,7 +47,7 @@ public class GroupType extends Type {
    * @param fields the contained fields
    */
   public GroupType(Repetition repetition, String name, List<Type> fields) {
-    this(repetition, name, null, fields, null);
+    this(repetition, name, (LogicalTypeAnnotation) null, fields, null);
   }
 
   /**
@@ -88,8 +88,18 @@ public class GroupType extends Type {
    * @param fields the contained fields
    * @param id the id of the field
    */
+  @Deprecated
   GroupType(Repetition repetition, String name, OriginalType originalType, List<Type> fields, ID id) {
     super(name, repetition, originalType, id);
+    this.fields = fields;
+    this.indexByName = new HashMap<String, Integer>();
+    for (int i = 0; i < fields.size(); i++) {
+      indexByName.put(fields.get(i).getName(), i);
+    }
+  }
+
+  GroupType(Repetition repetition, String name, LogicalTypeAnnotation logicalTypeAnnotation, List<Type> fields, ID id) {
+    super(name, repetition, logicalTypeAnnotation, id);
     this.fields = fields;
     this.indexByName = new HashMap<String, Integer>();
     for (int i = 0; i < fields.size(); i++) {
--- a/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
@@ -280,6 +280,14 @@ public interface LogicalTypeAnnotation {
       this.precision = precision;
     }
 
+    public int getPrecision() {
+      return precision;
+    }
+
+    public int getScale() {
+      return scale;
+    }
+
     @Override
     public LogicalType toLogicalType() {
       return LogicalType.DECIMAL(new DecimalType(scale, precision));
--- a/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.schema;
 
+import org.apache.parquet.Preconditions;
 import org.apache.parquet.format.BsonType;
 import org.apache.parquet.format.ConvertedType;
 import org.apache.parquet.format.DateType;
@@ -61,6 +62,13 @@ public interface LogicalTypeAnnotation {
   OriginalType toOriginalType();
 
   /**
+   * Visits this logical type with the given visitor
+   *
+   * @param logicalTypeAnnotationVisitor the visitor to visit this type
+   */
+  void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor);
+
+  /**
    * Helper method to convert the old representation of logical types (OriginalType) to new logical type.
    */
   static LogicalTypeAnnotation fromOriginalType(OriginalType originalType, DecimalMetadata decimalMetadata) {
@@ -91,21 +99,21 @@ public interface LogicalTypeAnnotation {
       case TIME_MICROS:
         return TimeLogicalTypeAnnotation.create(true, LogicalTypeAnnotation.TimeUnit.MICROS);
       case UINT_8:
-        return IntLogicalTypeAnnotation.create((byte) 8, false);
+        return IntLogicalTypeAnnotation.create(8, false);
       case UINT_16:
-        return IntLogicalTypeAnnotation.create((byte) 16, false);
+        return IntLogicalTypeAnnotation.create(16, false);
       case UINT_32:
-        return IntLogicalTypeAnnotation.create((byte) 32, false);
+        return IntLogicalTypeAnnotation.create(32, false);
       case UINT_64:
-        return IntLogicalTypeAnnotation.create((byte) 64, false);
+        return IntLogicalTypeAnnotation.create(64, false);
       case INT_8:
-        return IntLogicalTypeAnnotation.create((byte) 8, true);
+        return IntLogicalTypeAnnotation.create(8, true);
       case INT_16:
-        return IntLogicalTypeAnnotation.create((byte) 16, true);
+        return IntLogicalTypeAnnotation.create(16, true);
       case INT_32:
-        return IntLogicalTypeAnnotation.create((byte) 32, true);
+        return IntLogicalTypeAnnotation.create(32, true);
       case INT_64:
-        return IntLogicalTypeAnnotation.create((byte) 64, true);
+        return IntLogicalTypeAnnotation.create(64, true);
       case ENUM:
         return EnumLogicalTypeAnnotation.create();
       case JSON:
@@ -142,6 +150,11 @@ public interface LogicalTypeAnnotation {
     @Override
     public OriginalType toOriginalType() {
       return OriginalType.UTF8;
+    }
+
+    @Override
+    public void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor) {
+      logicalTypeAnnotationVisitor.visit(this);
     }
 
     @Override
@@ -182,6 +195,11 @@ public interface LogicalTypeAnnotation {
     }
 
     @Override
+    public void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor) {
+      logicalTypeAnnotationVisitor.visit(this);
+    }
+
+    @Override
     public boolean equals(Object obj) {
       return obj instanceof MapLogicalTypeAnnotation;
     }
@@ -219,6 +237,11 @@ public interface LogicalTypeAnnotation {
     }
 
     @Override
+    public void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor) {
+      logicalTypeAnnotationVisitor.visit(this);
+    }
+
+    @Override
     public boolean equals(Object obj) {
       return obj instanceof ListLogicalTypeAnnotation;
     }
@@ -253,6 +276,11 @@ public interface LogicalTypeAnnotation {
     @Override
     public OriginalType toOriginalType() {
       return OriginalType.ENUM;
+    }
+
+    @Override
+    public void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor) {
+      logicalTypeAnnotationVisitor.visit(this);
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -252,7 +252,12 @@ protected final THIS repetition(Type.Repetition repetition) {
      * @return this builder for method chaining
      */
     public THIS as(OriginalType type) {
-      this.originalType = type;
+      this.logicalTypeAnnotation = LogicalTypeAnnotation.fromOriginalType(type);
+      return self();
+    }
+
+    public THIS as(LogicalTypeAnnotation type) {
```
gszadovszky: This method breaks the fluent API of the builder as you need to use an outside factory method of the final type to create a LogicalTypeAnnotation. If it is practically feasible, I would suggest to refactor the LogicalTypeAnnotation API to fit more in the fluent API of Types.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -586,108 +595,105 @@ Type getType(PrimitiveTypeName type) {
...
+        return LogicalTypeAnnotation.TimestampLogicalTypeAnnotation.create(timestamp.isAdjustedToUTC, convertTimeUnit(timestamp.unit));
       default:
-        throw new RuntimeException("Unknown original type " + type);
-     }
-   }
+        throw new RuntimeException("Unknown logical type " + type);
+    }
+  }
+
+  LogicalTypeAnnotation.TimeUnit convertTimeUnit(TimeUnit unit) {
```
gszadovszky: As far as I can see this method can be private. Please, use visibility as tight as possible and comment if something is visible for testing purposes only.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java
```diff
@@ -41,6 +40,10 @@
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  static {
+    objectMapper.configure(SerializationConfig.Feature.FAIL_ON_EMPTY_BEANS, false);
```
gszadovszky: Could you please explain why it is necessary?

nandorKollar: Without this configuration change on the objectMapper three tests in parquet-cascading3 fail with the following error:

  testReadPattern(org.apache.parquet.cascading.TestParquetTupleScheme): [namecp] could not build flow from assembly: [java.io.IOException: Could not read footer: java.lang.RuntimeException: org.codehaus.jackson.map.JsonMappingException: No serializer found for class org.apache.parquet.schema.LogicalTypeAnnotation$StringLogicalTypeAnnotation and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationConfig.Feature.FAIL_ON_EMPTY_BEANS) ) (through reference chain: org.apache.parquet.hadoop.metadata.ParquetMetadata["fileMetaData"]->org.apache.parquet.hadoop.metadata.FileMetaData["schema"]->org.apache.parquet.schema.MessageType["fields"]->java.util.ArrayList[0]->org.apache.parquet.schema.PrimitiveType["logicalTypeAnnotation"])]

The reason is: the logical types are no longer represented as enum, but a classed, and Jackson can't serialize empty classes because FAIL_ON_EMPTY_BEANS feature is enabled by default.

gszadovszky: I see. Thanks.
Could you please add some comments about that so others will also understand?

nandorKollar: Sure, no problem, I will.

parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -88,6 +88,7 @@ public GroupType(Repetition repetition, String name, OriginalType originalType,
    * @param fields the contained fields
    * @param id the id of the field
    */
+  @Deprecated
```
gszadovszky: I think, it is enough to deprecate public API. If a method is not public we can freely modify/remove it.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -401,15 +400,18 @@ public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
...
+    this(repetition, primitive, length, name, (LogicalTypeAnnotation) null, null, null);
   }
 
   /**
    * @param repetition OPTIONAL, REPEATED, REQUIRED
    * @param primitive STRING, INT64, ...
    * @param name the name of the type
    * @param originalType (optional) the original type to help with cross schema convertion (LIST, MAP, ...)
+   *
+   * @deprecated use {@link #PrimitiveType(Repetition, PrimitiveTypeName, String, LogicalTypeAnnotation)} instead
```
gszadovszky: The usual pattern in parquet for deprecating is to mention that the related method will be removed in 2.0.0. In case of the suggestion would be to use one of the overloaded methods then it is fine to not mention.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -436,13 +438,20 @@ public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
    * @param originalType (optional) the original type (MAP, DECIMAL, UTF8, ...)
    * @param decimalMeta (optional) metadata about the decimal type
    * @param id the id of the field
+   *
+   * @deprecated use {@link #PrimitiveType(Repetition, PrimitiveTypeName, int, String, LogicalTypeAnnotation, ID)} instead
```
gszadovszky: See above.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -436,13 +438,20 @@ public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
...
    */
+  @Deprecated
   public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
                        int length, String name, OriginalType originalType,
                        DecimalMetadata decimalMeta, ID id) {
     this(repetition, primitive, length, name, originalType, decimalMeta, id, null);
   }
 
+  /**
+   * @deprecated use {@link #PrimitiveType(Repetition, PrimitiveTypeName, int, String, LogicalTypeAnnotation, ID, ColumnOrder)} instead
```
gszadovszky: Not public, no need for deprecation.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -436,13 +438,20 @@ public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
...
+   */
+  @Deprecated
```
gszadovszky: See above.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -459,6 +468,37 @@ public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
     this.columnOrder = requireValidColumnOrder(columnOrder);
   }
 
+  public PrimitiveType(Repetition repetition, PrimitiveTypeName primitive,
```
gszadovszky: I think, in long term we do not want to expose the construction of types from the API. We would like the clients to use the builder instead. Therefore, I would suggest not adding public constructors if your code does not need it.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -146,11 +146,18 @@ public Type(String name, Repetition repetition, OriginalType originalType) {
    * @param repetition OPTIONAL, REPEATED, REQUIRED
    * @param originalType (optional) the original type to help with cross schema conversion (LIST, MAP, ...)
    * @param id (optional) the id of the fields.
+   *
+   * @deprecated use {@link #Type(String, Repetition, LogicalTypeAnnotation, ID)} instead
    */
+  @Deprecated
```
gszadovszky: Not public, no deprecating is required.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -146,11 +146,18 @@ public Type(String name, Repetition repetition, OriginalType originalType) {
...
   Type(String name, Repetition repetition, OriginalType originalType, ID id) {
     this(name, repetition, originalType, null, id);
   }
 
+  /**
+   * @deprecated use {@link #Type(String, Repetition, LogicalTypeAnnotation, ID)} instead
+   */
+  @Deprecated
```
gszadovszky: see above

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
nandorKollar: Ah sorry, missed this one. Sure, I can make it package private

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -77,63 +77,116 @@ static LogicalTypeAnnotation fromOriginalType(OriginalType originalType, Decimal
...
+
+  static JsonLogicalTypeAnnotation jsonType() {
+    return JsonLogicalTypeAnnotation.INSTANCE;
+  }
+
+  static BsonLogicalTypeAnnotation bsonType() {
+    return BsonLogicalTypeAnnotation.INSTANCE;
+  }
+
+  static IntervalLogicalTypeAnnotation intervalType() {
```
gszadovszky: Do we want the user to be able to create an Interval or a MapKeyValue type? If they are only exist for backward compatibility reasons the factory method should not be public so users won't be able to create them.

nandorKollar: I think MapKeyValue is here only for backward compatibility, but Interval should be supported in the future. Since these factory methods are defined on an interface, I can't easily change the visibility to private/package protected. Should I change it to an abstract class?

gszadovszky: Interval is not supported in LogicalType currently, so I would not expose it yet. It is always possible to make a restricted access public but in the other way it would break backward compatibility.
Abstract class sounds reasonable to me.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -36,42 +36,152 @@
 import org.apache.parquet.format.TimeType;
 import org.apache.parquet.format.TimestampType;
 
+import java.util.List;
 import java.util.Objects;
 
-public interface LogicalTypeAnnotation {
+public abstract class LogicalTypeAnnotation {
+  public enum LogicalTypes {
```
gszadovszky: This enum is used only for parsing/printing and we don't want the users to really use them. So, I would suggest using a name that suggests its use e.g. `LogicalTypeParseHelper`?
Also, it would be nice if we could annotate/comment that this one is not part of the public API.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -36,42 +36,152 @@
...
+    protected abstract LogicalTypeAnnotation fromString(List<String> params);
+  }
+
   /**
    * Convert this parquet-mr logical type to parquet-format LogicalType.
    *
    * @return the parquet-format LogicalType representation of this logical type implementation
    */
-  LogicalType toLogicalType();
+  public abstract LogicalType toLogicalType();
```
gszadovszky: We usually try to keep the usage of the thrift generated objects (from `parquet-format`) in `ParquetMetadataConverter`. Do you think it is possible to refactor the related code parts to there (not necessarily to the same class but to the same module)?

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -36,42 +36,152 @@
...
 
   /**
    * Visits this logical type with the given visitor
    *
    * @param logicalTypeAnnotationVisitor the visitor to visit this type
    */
-  void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor);
+  public abstract void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor);
+
+  public abstract LogicalTypes getType();
```
gszadovszky: It seems this one is not used outside of this class/package. I would suggest not being public.

parquet-column/src/test/java/org/apache/parquet/parser/TestParquetParser.java
```diff
@@ -293,6 +298,36 @@ public void testIntAnnotations() {
...
+      "  required int64 i64 (INT(64,true));" +
+      "  required int32 u8 (INT(8,false));" +
+      "  required int32 u16 (INT(16,false));" +
+      "  required int32 u32 (INT(32,false));" +
+      "  required int64 u64 (INT(64,false));" +
+      "}\n";
+
+    MessageType parsed = MessageTypeParser.parseMessageType(message);
+    MessageType expected = Types.buildMessage()
+      .required(INT32).as(LogicalTypeAnnotation.intType(8, true)).named("i8")
```
gszadovszky: nit: I think, the static import of the method intType would help reading the code.

nandorKollar:
@gszadovszky it looks like the version of maven-shade-plugin and enforcer-rule dependency in maven-enforcer-plugin Maven plugins used in Parquet don't like Java 8 lambda expressions, that's why Travis build failed. Since Parquet is already upgraded to Java 8, I think we should upgrade these plugins. I'll create a separate Jira for this issue.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
nandorKollar: accept is intended for public use, for the visitors

gszadovszky: Sorry, I meant `getType()`.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -19,28 +19,13 @@
...
-import org.apache.parquet.format.TimeType;
-import org.apache.parquet.format.TimestampType;
 
 import java.util.List;
 import java.util.Objects;
 
 public abstract class LogicalTypeAnnotation {
-  public enum LogicalTypes {
+  // This is a private enum intended only for internal use for parsing the schema
+  public enum LogicalTypeToken {
```
gszadovszky: As far as I can see this enum is used only from the schema package. I would suggest using package private access so the comment is not needed and the clients cannot misuse it.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -164,7 +135,7 @@ protected LogicalTypeAnnotation fromString(List<String> params) {
    */
   public abstract void accept(LogicalTypeAnnotationVisitor logicalTypeAnnotationVisitor);
 
-  public abstract LogicalTypes getType();
+  protected abstract LogicalTypeToken getType();
```
gszadovszky: nit: package private would be fine here.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
nandorKollar: Good idea, thanks Gabor!

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
gszadovszky: Good point. I would say if something is not necessary to be public then restrict its access before committing. The later the harder to remove from the public API.

wesm:
What will happen if you try reading a Parquet file with the new logical types with an old reader (or another implementation, like Apache Impala)?

PR #469, opened 2018-04-29
PARQUET-1285: [Java] SchemaConverter should not convert from TimeUnit.SECOND and TimeUnit.NANOSECOND of Arrow

masayuki038:
Arrow's 'Time' definition is below:
```
{ "name" : "time", "unit" : "SECOND|MILLISECOND|MICROSECOND|NANOSECOND", "bitWidth": /* integer: 32 or 64 */ }
```
http://arrow.apache.org/docs/metadata.html

But Parquet only supports 'TIME_MILLIS' and 'TIME_MICROS'.
https://github.com/Apache/parquet-format/blob/master/LogicalTypes.md

Therefore SchemaConverter should not convert from TimeUnit.SECOND and TimeUnit.NANOSECOND of Arrow to Parquet.

### DIFF
--- a/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
+++ b/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
@@ -19,20 +19,7 @@
 package org.apache.parquet.arrow.schema;
 
 import static java.util.Arrays.asList;
-import static org.apache.parquet.schema.OriginalType.DATE;
-import static org.apache.parquet.schema.OriginalType.DECIMAL;
-import static org.apache.parquet.schema.OriginalType.INTERVAL;
-import static org.apache.parquet.schema.OriginalType.INT_16;
-import static org.apache.parquet.schema.OriginalType.INT_32;
-import static org.apache.parquet.schema.OriginalType.INT_64;
-import static org.apache.parquet.schema.OriginalType.INT_8;
-import static org.apache.parquet.schema.OriginalType.TIMESTAMP_MILLIS;
-import static org.apache.parquet.schema.OriginalType.TIME_MILLIS;
-import static org.apache.parquet.schema.OriginalType.UINT_16;
-import static org.apache.parquet.schema.OriginalType.UINT_32;
-import static org.apache.parquet.schema.OriginalType.UINT_64;
-import static org.apache.parquet.schema.OriginalType.UINT_8;
-import static org.apache.parquet.schema.OriginalType.UTF8;
+import static org.apache.parquet.schema.OriginalType.*;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BOOLEAN;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.DOUBLE;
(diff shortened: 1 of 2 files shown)


parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -245,7 +246,14 @@ public TypeMapping visit(Date type) {
 
       @Override
       public TypeMapping visit(Time type) {
-        return primitive(INT32, TIME_MILLIS);
+        int bitWidth = type.getBitWidth();
+        org.apache.arrow.vector.types.TimeUnit timeUnit = type.getUnit();
```
xhochy: This should be imported at the top to have a much shorter code line.

masayuki038: @xhochy Thanks for your review. I will fix it.

parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -407,11 +416,11 @@ public TypeMapping convertINT32(PrimitiveTypeName primitiveTypeName) throws Runt
...
-            return field(new ArrowType.Timestamp(org.apache.arrow.vector.types.TimeUnit.MICROSECOND, "UTC"));
+            return field(new ArrowType.Timestamp(TimeUnit.MICROSECOND, "UTC"));
           case TIMESTAMP_MILLIS:
-            return field(new ArrowType.Timestamp(org.apache.arrow.vector.types.TimeUnit.MILLISECOND, "UTC"));
+            return field(new ArrowType.Timestamp(TimeUnit.MILLISECOND, "UTC"));
           case TIME_MILLIS:
-            return field(new ArrowType.Time(org.apache.arrow.vector.types.TimeUnit.MILLISECOND, 32));
+            return field(new ArrowType.Time(TimeUnit.MILLISECOND, 32));
           default:
           case TIME_MICROS:
```
rdblue: Shouldn't `TIME_MICROS` be supported as well?

rdblue:
One conversion is missing, but otherwise it looks good to me. @xhochy, is this the same logic that parquet-cpp uses?

I'm fine with you merging PRs like this on the Java side. Just be sure to ping someone if you're not sure about something.

parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
masayuki038: @rdblue Thanks. That's true. And this have to add some tests about converting from parquet to arrow. I will fix.

parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -245,7 +247,14 @@ public TypeMapping visit(Date type) {
 
       @Override
       public TypeMapping visit(Time type) {
```
nandorKollar: Does Timestamp Arrow type have the same problem? The visit method always maps to `primitive(INT64, TIMESTAMP_MILLIS)` regardless off Arrow type time unit.

masayuki038: @rdblue @xhochy Colud you please check so far? I will fix Timestamp issue as an another pull request.

PR #473, opened 2018-05-01
Adding constructors to ProtoParquetWriter with writeSpecsCompliant flag

chawlakunal:
This PR adds constructors to ProtoParquetWriter with flags for specs compliant writer (based on #411). Now a specs compliant ProtoParquetWriter can be created as
```
new ProtoParquetWriter(file, protoMessageClass, true);
```

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoParquetWriter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoParquetWriter.java
@@ -20,6 +20,7 @@ package org.apache.parquet.proto;
 
 import com.google.protobuf.Message;
 import com.google.protobuf.MessageOrBuilder;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.hadoop.ParquetWriter;
 import org.apache.parquet.hadoop.api.WriteSupport;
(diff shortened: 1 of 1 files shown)


parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoParquetWriter.java
```diff
@@ -81,4 +82,63 @@ public ProtoParquetWriter(Path file, Class<? extends Message> protoMessage) thro
...
+   */
+  public ProtoParquetWriter(Path file, Class<? extends Message> protoMessage,
+                            CompressionCodecName compressionCodecName, int blockSize,
+                            int pageSize, boolean writeSpecsCompliant) throws IOException {
+    this(file, protoMessage, compressionCodecName, blockSize, pageSize, DEFAULT_IS_DICTIONARY_ENABLED, 
+            DEFAULT_IS_VALIDATING_ENABLED, writeSpecsCompliant);
+  }
+
+  /**
+   * Create a new {@link ProtoParquetWriter}. The default block size is 50 MB.The default
```
BenoitHanotte: I believe the default block size is 128MB by default (see https://github.com/apache/parquet-mr/blob/0d55abd05b0e5027c18e60d1ac3b22998dd00951/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L41

BenoitHanotte:
Thamks @chawlakunal for the commit, it looks good to me. I believe there is a slight mistake an a comment (see my review above). 
You may want to open a Jira ticket for this commit and add the ticket name in the message ;)

chawlakunal:
@BenoitHanotte Happy to contribute (in a very small way though :P) 

I have created a JIRA for this but I am unable to assign it to myself.

zivanfi:
@chawlakunal, I added you the list of contributors so that you can assign JIRA-s to yourself. I assigned this one to you to make sure it works.

lukasnalezenec:
Hi, 
Thank you for your work.
I looked at it briefly, I will send you detailed opinion after weekend.
I am not sure if we need so many new constructors.
We should publish the Configuration parameter from superclass.

chawlakunal:
Hi @lukasnalezenec
Waiting to hear from you.

costimuraru:
+1 on finding a way to pass the `writeSpecsCompliant` flag directly to the `ProtoParquetWriter` and simplify things.

Right now we have to do something like:
```java
ProtoWriteSupport.setWriteSpecsCompliant(conf, true);

ParquetWriter<MyMessage> writer 
   = new ParquetWriter<>(file, new ProtoWriteSupport<MyMessage>(cls), CompressionCodecName.GZIP, 256 * 1024 * 1024, 1 * 1024 * 1024,  1 * 1024 * 1024, true, false, ParquetWriter.DEFAULT_WRITER_VERSION, conf);
```

chawlakunal:
I am still waiting for comments from @lukasnalezenec

lukasnalezenec:
I think that problem is that there was no ProtoParquetWriter with conf parameter.
I would recommend adding the conf parameter and not writeSpecsCompliant parameter. 
AFAIK there is no elegant solution how we can accept writeSpecsCompliant parameter, set its value to configuration and pass it super constructor.
(see getConfigWithWriteSpecsCompliant())


I would add those two new constructors:

ProtoParquetWriter(Path file, Class<? extends Message> protoMessage, Configuration conf);

ProtoParquetWriter(Path file, Class<? extends Message> protoMessage, ParquetFileWriter.Mode mode, CompressionCodecName compressionCodecName, int blockSize, int pageSize, int dictionaryPageSize, boolean enableDictionary, boolean validating, WriterVersion writerVersion, Configuration conf

We might add also extends some other existing constructors with conf parameter.

If you disagree with me, I'm open to another solution.

costimuraru:
@lukasnalezenec From my perspective, it looks great!

BenoitHanotte:
Hello @costimuraru @chawlakunal 
As we might again be adding a new flag if https://github.com/apache/parquet-mr/pull/410 passes, I think we may want to follow @lukasnalezenec suggestion and try to have constructors that accept the conf object, which would allow to avoid duplicating all the existing constructors for each new flag.

costimuraru:
@chawlakunal any luck?

georgehdd:
@chawlakunal Any news on this?

PR #476, opened 2018-05-11
PARQUET-1296: Travis kills build after 10 minutes, because "no output…

nandorKollar:
… was received"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,5 @@ env:
   - HADOOP_PROFILE=default TEST_CODECS=uncompressed,brotli
   - HADOOP_PROFILE=default TEST_CODECS=gzip,snappy
 
-install: travis_wait mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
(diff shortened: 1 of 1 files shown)


xhochy:
> Why do we log to a file to print out its content when the execution fails? Travis captures the output, so I don't think we need to store the logs as a file.

Having an enormous log is quite annoying when looking for a problem in another part of the job. Basically it makes the Travis UI unusable and you need to resort to downloading the log and use grep on your machine. Only printing parts of the log helped me in other projects to more quickly find the problem and give appropriate feedback on PRs than. If you do a lot of review, this is quite a significant time saver.

gszadovszky:
Thanks, @xhochy for the explanation. I'm OK with the actual solution, then. (After it is working properly...)

zivanfi:
How about adding `sudo apt-get install pv` the `before_install` then using `| pv -fbi 60 > mvn_install.log` instead of ` > mvn_install.log`? This will display the size of the output every 60 seconds like so:

     190 B
     530 B
     770 B
    1.10KiB
    1.43KiB
    1.78KiB

A few extra lines would end up in the travis log, but nothing too distractive and it would be enough to prevent timeouts due to no output.

xhochy:
@zivanfi This is nearly the same thing as `travis_wait` does. The benefit of `travis_wait` is that is already installed.

zivanfi:
@xhochy Yes, but according to the latest Travis build getting `travis_wait` to work with our command is non-trivial.

zivanfi:
@xhochy, are you okay with this approach?

PR #479, opened 2018-05-17
PARQUET-1212: Column indexes: Show indexes in tools

gszadovszky:


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
```diff
@@ -0,0 +1,165 @@
...
+  @Parameter(names = { "-c", "--column" }, description = "Shows the column/offset indexes for the given column only")
+  List<String> ColumnPaths;
+
+  @Parameter(names = { "-b",
+      "--block" }, description = "Shows the column/offset indexes for the given block (row-group) only; "
+          + "blocks are referenced by their indexes from 0")
+  List<String> blockIndexes;
+
+  @Parameter(names = { "-i", "--column-index" }, description = "Shows the column indexes; "
+      + "the default behavior is if both -i and -o would be set")
```
zivanfi: (subjective, optional) I would suggest "active by default unless -o is used"

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
```diff
@@ -0,0 +1,165 @@
...
+  boolean showColumnIndex;
+
+  @Parameter(names = { "-o", "--offset-index" }, description = "Shows the offset indexes; "
+      + "the default behavior is if both -i and -o would be set")
```
zivanfi: (subjective, optional) I would suggest "active by default unless -i is used"

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
```diff
@@ -0,0 +1,165 @@
...
+
+  @Parameter(names = { "-c", "--column" }, description = "Shows the column/offset indexes for the given column only")
+  List<String> ColumnPaths;
+
+  @Parameter(names = { "-b",
+      "--block" }, description = "Shows the column/offset indexes for the given block (row-group) only; "
+          + "blocks are referenced by their indexes from 0")
+  List<String> blockIndexes;
+
+  @Parameter(names = { "-i", "--column-index" }, description = "Shows the column indexes; "
```
zivanfi: (subjective, optional) I find -i very counter-intuitive. Could this be -c and --column something else? For example -f/--field? Alternatively, -k/--column? Or -p/--column-path?

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
```diff
@@ -0,0 +1,165 @@
...
+              console.info(offsetIndex.toString());
+            }
+          }
+        }
+      }
+    }
+    return 0;
+  }
+
+  private String getBlockIndex(int i) {
```
zivanfi: Could you please add a short comment about what happens here and why?

If I understand correclty:
* without blockIndexes, we address into blocks directly, but
* blockIndexes introduces and additional layer of indirection if exists

Although I'm still unsure about why.

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnIndexCommand.java
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.hadoop.util.HadoopInputFile;
+import org.apache.parquet.io.InputFile;
+import org.apache.parquet.tools.Main;
+
+/**
+ * parquet-tools command to print column and offset indexes.
+ */
+public class ColumnIndexCommand extends ArgsOnlyCommand {
```
zivanfi: Same comments as for the CLI command.

parquet-tools/src/main/java/org/apache/parquet/tools/command/Registry.java
```diff
@@ -34,6 +34,7 @@
     registry.put("merge", MergeCommand.class);
     registry.put("rowcount", RowCountCommand.class);
     registry.put("size", SizeCommand.class);
+    registry.put("columnindex", ColumnIndexCommand.class);
```
zivanfi: I think the command name should be the same in the CLI and in parquet-tools.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
gszadovszky: `--column -c` is used in other commands. I think, it is better to keep this option as is to be common between the different commands.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowColumnIndexCommand.java
zivanfi: Thanks to your comment, finally I undestood what is going on but I find it very confusing.
* If the user wants to see all blocks, the index of elements returned by getBlocks() corresponds to their actual index in the file.
* If, on the other hand, the user wants to view specific blocks, getBlocks() returns elements in the order they were requested and the actual index can only be obtained via a reverse lookup.

I think it would be a lot cleaner the following way:
* The getBlocks() method should return an index to block map instead of a list of blocks.
* This way when iterating though the results of getBlocks(), we won't have to do a reverse lookup for the actual index, since it is a part of the container returned.

PR #480, opened 2018-05-17
PARQUET-1213: Column indexes: Limit index size

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -195,13 +195,19 @@ int compareMinValues(PrimitiveComparator<Binary> comparator, int index1, int ind
     int compareMaxValues(PrimitiveComparator<Binary> comparator, int index1, int index2) {
       return 0;
     }
+
+    @Override
+    int sizeOf(Object minMax) {
```
zivanfi: Could this be abstract instead? Additionally, in every other sizeOf() implementation, de parameter is called value, not minMax.

gszadovszky: This is not the base class but a special implementation for no-op. This builder is used to not to gather any data and build null therefore the return value of sizeof does not matter.

PR #481, opened 2018-05-17
PARQUET-1214: Column indexes: Truncate min/max values

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/BinaryTruncator.java
```diff
@@ -0,0 +1,207 @@
...
+    private byte[] incrementUtf8(byte[] array) {
+      if (array == null) {
+        return null;
+      }
+      ByteBuffer buffer = ByteBuffer.wrap(array);
+      for (int i = array.length - 1; i >= 0; --i) {
+        byte prev = array[i];
+        byte inc = prev;
+        while (++inc != 0) { // Until overflow: 0xFF -> 0x00
+          array[i] = inc;
```
zivanfi: This one seems to increment bytes up to FF, but then not write the overflow back to the array. I.e., if I understand correctly, the array may go through the following changes:

42 FB
42 FC
42 FD
42 FE
42 FF
43 FF
...

I think the last one should be 43 00

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -0,0 +1,207 @@
...
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Types;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tests for {@link BinaryTruncator}
+ */
+public class TestBinaryTruncator {
```
zivanfi: I think it would be worth to have some test cases with specific expected values as well instead of just checking that the actual result satisfies the contract.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -143,6 +143,7 @@
   public static final String MIN_ROW_COUNT_FOR_PAGE_SIZE_CHECK = "parquet.page.size.row.check.min";
   public static final String MAX_ROW_COUNT_FOR_PAGE_SIZE_CHECK = "parquet.page.size.row.check.max";
   public static final String ESTIMATE_PAGE_SIZE_CHECK = "parquet.page.size.check.estimate";
+  public static final String CULOMN_INDEX_TRUNCATE_LENGTH = "parquet.columnindex.truncate.length";
```
zivanfi: Typo: CULOMN should be COLUMN.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -0,0 +1,269 @@
...
+        truncator.truncateMax(Binary.fromString(
+            UTF8_1BYTE_MAX_CHAR
+                + UTF8_2BYTES_MAX_CHAR
+                + UTF8_3BYTES_MAX_CHAR
+                + UTF8_4BYTES_MAX_CHAR),
+            5));
+
+    // Truncate invalid UTF-8 values -> truncate without validity check
+    assertEquals(binary(0xFF, 0xFE, 0xFD), truncator.truncateMin(binary(0xFF, 0xFE, 0xFD, 0xFC, 0xFB, 0xFA), 3));
+    assertEquals(binary(0xFF, 0xFE, 0xFE), truncator.truncateMax(binary(0xFF, 0xFE, 0xFD, 0xFC, 0xFB, 0xFA), 3));
```
zivanfi: Could you also add a test where there is a carry over some bytes? For example, truncateMax(binary(0xFF, 0xFE, 0xFD, 0xFF, 0xFF, 0xFF), 5) should become binary(0xFF, 0xFE, 0xFE) or binary(0xFF, 0xFE, 0xFE, 0x00, 0x00).

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -0,0 +1,269 @@
...
+                + UTF8_2BYTES_MAX_CHAR
+                + UTF8_3BYTES_MAX_CHAR
+                + UTF8_4BYTES_MAX_CHAR),
+        truncator.truncateMax(Binary.fromString(
+            UTF8_1BYTE_MAX_CHAR
+                + UTF8_2BYTES_MAX_CHAR
+                + UTF8_3BYTES_MAX_CHAR
+                + UTF8_4BYTES_MAX_CHAR),
+            5));
+
```
zivanfi: Could you also add a test where there is a carry over some bytes? For example, "abc" followed by a few max chars and truncated to 4 or more bytes should become "abd" potentially followed by a few \0-s.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -0,0 +1,269 @@
...
+  public void testStringTruncate() {
+    BinaryTruncator truncator = BinaryTruncator.getTruncator(Types.required(BINARY).as(UTF8).named("test_utf8"));
+
+    // Truncate 1 byte characters
+    assertEquals(Binary.fromString("abc"), truncator.truncateMin(Binary.fromString("abcdef"), 3));
+    assertEquals(Binary.fromString("abd"), truncator.truncateMax(Binary.fromString("abcdef"), 3));
+
+    // Truncate 1-2 bytes characters
+    assertEquals(Binary.fromString("árvízt"), truncator.truncateMin(Binary.fromString("árvíztűrő"), 8));
+    assertEquals(Binary.fromString("árvízu"), truncator.truncateMax(Binary.fromString("árvíztűrő"), 8));
```
zivanfi: Could you also add a test case where the truncation cutoff happens to be inside a single unicode codepoint? For example, byte position 9 of "árvíztűrő" ends up inside the bytes representing "ű" (c5 b1), so you could add a test case that truncates that to a length of 9 bytes.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
gszadovszky: Just realized after modifying the code that in case of UTF-8 it is not easy to put `\0` to the end for carry on. I'll put back the original algorithm that writes back the original bytes in case the incrementation fails so the last maximum UTF-8 characters will be unchanged.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
gszadovszky: That was the original idea just miscalculated the length...

PR #482, opened 2018-05-17
PARQUET-1303 correct ClassCastException for Avro @Stringable fields

zebehringer:


zebehringer:
TestReflectLogicalType.testWriteUUIDMissingLogicalType no longer throws ClassCastException after this change. In one sense the test is no longer valid, but in a strict sense it is since the UUID field in that test is not annotated with @Stringable, but there doesn't seem to be any way to detect that in AvroWriteSupport, so it now passing is a side effect of the change.

parquet-avro/src/test/java/org/apache/parquet/avro/TestReflectLogicalTypes.java
```diff
@@ -426,33 +426,6 @@ public void testWriteNullableUUID() throws IOException {
         read(REFLECT, nullableUuidStringSchema, test));
   }
 
-  @Test(expected = ClassCastException.class)
```
nandorKollar: I think this tests an important aspect: what happens when the logical type conversion is not present, and the field is not converted to string. Before, this case threw an exception, but with your change, it will convert the UUID to string (since the schema tells that that field is a string).

In my opinion we should not drop this test case, but instead remove the expected and add a relevant assert, unless there's any test case which verifies this behavior.

zebehringer: I have restored the test, removed the expected exception and added a couple asserts, similar to testWriteUUID, which uses REFLECT for the write() instead of uuidSchema. Let me know if that covers it.

nandorKollar: @Test should still stay there, just the expected exception is not needed. :)

Fokko:
@nandorKollar Still something that we want?

PR #486, opened 2018-05-24
*FOR DISCUSSION* Flag to control snappy decompression on/off heap

agrove-rms:
This is in response to https://issues.apache.org/jira/browse/PARQUET-118

We are running into issues with containers being killed due to excessive off heap memory usage that we cannot control.

For us, adding this flag is a good solution even if it doesn't provide the best performance. Stability > performance.

I'm open to working on a more complete solution to control how much off-heap is used too, as a longer term fix.


agrove-rms:
I now see that this doesn't work without changes to the underlying xenial-snappy library so am closing this while I dig deeper into this.

PR #493, opened 2018-06-08
PARQUET-1321: LogicalTypeAnnotation.LogicalTypeAnnotationVisitor#visit methods should have a return value

nandorKollar:
LogicalTypeAnnotation.LogicalTypeAnnotationVisitor#visit methods return Optionals,
the default implementations returns Optional#empty(). Clients should override these methods,
and return the relevant value.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
@@ -23,6 +23,7 @@ import org.apache.parquet.Preconditions;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 import static java.util.Optional.empty;
 
@@ -837,6 +838,11 @@ public abstract class LogicalTypeAnnotation {
    * The default implementation for each logical type specific visitor method is empty.
    * <p>
    * Example usage: logicalTypeAnnotation.accept(new LogicalTypeAnnotationVisitor() { ... });
+   *
+   * Every visit method returns {@link Optional#empty()} by default.
+   * It means that for the given logical type no specific action is needed.
+   * Client code can use {@link Optional#orElse(Object)} to return a default value for unhandled types,
+   * or {@link Optional#orElseThrow(Supplier)} to throw exception if omitting a type is not allowed.
    */
   public interface LogicalTypeAnnotationVisitor<T> {
     default Optional<T> visit(StringLogicalTypeAnnotation logicalTypeAnnotation) {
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -834,45 +838,63 @@ public int hashCode() {
...
    * Example usage: logicalTypeAnnotation.accept(new LogicalTypeAnnotationVisitor() { ... });
+   *
+   * Every visit method returns {@link Optional#empty()} by default.
+   * It means that for the given logical type no specific action is needed.
+   * Client code can use {@link Optional#orElse(Object)} to return a default value for unhandled types,
+   * or {@link Optional#orElseThrow(Supplier)} to throw exception if omitting a type is not allowed.
    */
-  public interface LogicalTypeAnnotationVisitor {
-    default void visit(StringLogicalTypeAnnotation logicalTypeAnnotation) {
+  public interface LogicalTypeAnnotationVisitor<T> {
```
gszadovszky: Thanks for adding the comments. I accept the concept of using Optionals here.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -241,18 +242,13 @@ private void visitChildren(final List<SchemaElement> result,
   }
 
   LogicalType convertToLogicalType(LogicalTypeAnnotation logicalTypeAnnotation) {
-    LogicalTypeConverterVisitor logicalTypeConverterVisitor = new LogicalTypeConverterVisitor();
-    logicalTypeAnnotation.accept(logicalTypeConverterVisitor);
-    return logicalTypeConverterVisitor.logicalType;
+    return logicalTypeAnnotation.accept(new LogicalTypeConverterVisitor()).get();
```
gszadovszky: I would suggest using a constant instead of creating a new instance every time this method is invoked.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -241,18 +242,13 @@ private void visitChildren(final List<SchemaElement> result,
...
   }
 
   ConvertedType convertToConvertedType(LogicalTypeAnnotation logicalTypeAnnotation) {
-    LogicalTypeConverterVisitor logicalTypeConverterVisitor = new LogicalTypeConverterVisitor();
-    logicalTypeAnnotation.accept(logicalTypeConverterVisitor);
-    return logicalTypeConverterVisitor.convertedType;
+    return logicalTypeAnnotation.accept(new ConvertedTypeConverterVisitor()).get();
```
gszadovszky: See above.

PR #494, opened 2018-06-11
PARQUET-1322: Statistics is not available for DECIMAL types

vrozov:


zivanfi:
Based on the discussion in the JIRA, this PR seems to be obsolete. Can we close it?

PR #497, opened 2018-06-24
PARQUET-1336: PrimitiveComparator should implements Serializable

wangyum:
`PrimitiveComparator` should implements `Serializable`. Otherwise, the following `UserDefinedPredicate` will throw `NotSerializableException`:
```scala
new UserDefinedPredicate[Binary] with Serializable {
  private val strToBinary = Binary.fromReusedByteArray(v.getBytes)
  private val size = strToBinary.length
  val comparator = PrimitiveComparator.UNSIGNED_LEXICOGRAPHICAL_BINARY_COMPARATOR

  override def canDrop(statistics: Statistics[Binary]): Boolean = {
    val max = statistics.getMax
    val min = statistics.getMin
    comparator.compare(max.slice(0, math.min(size, max.length)), strToBinary) < 0 ||
      comparator.compare(min.slice(0, math.min(size, min.length)), strToBinary) > 0
  }

  override def inverseCanDrop(statistics: Statistics[Binary]): Boolean = false

  override def keep(value: Binary): Boolean =
    UTF8String.fromBytes(value.getBytes).startsWith(UTF8String.fromString(v))
}
```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
@@ -28,8 +28,9 @@ import java.util.Comparator;
  * {@link Comparator} implementation that also supports the comparison of the related primitive type to avoid the
  * performance penalty of boxing/unboxing. The {@code compare} methods for the not supported primitive types throw
  * {@link UnsupportedOperationException}.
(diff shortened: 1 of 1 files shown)


wangyum:
Thanks @gszadovszky May be this is the reason: https://github.com/apache/parquet-mr/blob/0d55abd05b0e5027c18e60d1ac3b22998dd00951/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java#L227

PR #498, opened 2018-06-25
PARQUET-1338: Fix PrimitiveType.equals throw NPE

wangyum:
Fix `PrimitiveType.equals` throw NPE.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
@@ -611,7 +611,7 @@ public final class PrimitiveType extends Type {
    */
   @Override
   protected boolean equals(Type other) {
-    if (!other.isPrimitive()) {
+    if (other == null || !other.isPrimitive()) {
       return false;
     }
     PrimitiveType otherPrimitive = other.asPrimitiveType();
--- a/parquet-column/src/main/java/org/apache/parquet/schema/Type.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/Type.java
@@ -283,7 +283,7 @@ abstract public class Type {
 
   @Override
   public boolean equals(Object other) {
-    if (other == null || !(other instanceof Type)) {
+    if (!(other instanceof Type) || other == null) {
       return false;
     }
     return equals((Type)other);
--- a/parquet-column/src/main/java/org/apache/parquet/schema/Type.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/Type.java
@@ -283,7 +283,7 @@ abstract public class Type {
 
   @Override
   public boolean equals(Object other) {
-    if (!(other instanceof Type) || other == null) {
+    if (other == null || !(other instanceof Type)) {
       return false;
     }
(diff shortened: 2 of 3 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -611,7 +611,7 @@ protected boolean typeEquals(Type other) {
    */
   @Override
   protected boolean equals(Type other) {
-    if (!other.isPrimitive()) {
+    if (other == null || !other.isPrimitive()) {
```
nandorKollar: How come the other is null? Adding a null check sounds reasonable for me, though I wonder in what situation does one call equals with a null parameter.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -283,7 +283,7 @@ protected boolean equals(Type other) {
 
   @Override
   public boolean equals(Object other) {
-    if (!(other instanceof Type) || other == null) {
```
nandorKollar: Why is this required? The test you wrote fails when I don't apply the change on PrimitiveType, but doesn't fail if I don't apply this.

nandorKollar: Actually I think null check is not needed here at all: https://docs.oracle.com/javase/specs/jls/se10/html/jls-15.html#jls-15.20.2

wangyum:
It seems that my usage is wrong.

rdblue:
@wangyum, the PrimitiveType's equals method still seems wrong. I think the Type implementation that uses instanceof is fine (because null instanceof Anything is always false) but PrimitiveType is definitely wrong.

wangyum:
Thanks @rdblue. I reopened this.

nandorKollar:
It seems to me that this equals method is called from the public equals method of Type, where the null check is already present. Since null check is already done before (the 'real' equals), what's the advantage of doing it here again?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
rdblue: It's part of the method contract: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)

This basically eliminates the need to add two checks when you want to check equality and the other object may be null. There are lots of cases where equals is used and the code doesn't need to handle null checks in addition.

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -31,6 +31,7 @@
 import static org.apache.parquet.schema.Type.Repetition.*;
 
 public class TestTypeBuilders {
+
```
rdblue: Can you remove this? It is a non-functional change.

rdblue:
@nandorKollar, any `equals` implementation should fulfill the documented contract, regardless of how it is used at the moment. Code changes over time and you don't want NPEs thrown because you didn't foresee how something would be used.

PR #499, opened 2018-06-28
PARQUET-1341: Fix null count stats in unsigned-sort columns.

rdblue:
When a column is all nulls, it has no min and max values, causing it to be handled with the conversion from min and max (not the newer min_value and max_value). In that path, stats are suppressed if the column's expected sort order is not a signed order because the old code only used signed ordering.

This PR always adds the null count, regardless of the handling for min and max or min_value and max_value.


rdblue:
@gszadovszky, can you have a look at this? I think it is a (minor) bug from the recent stats work.

nandorKollar:
There's a failing unit test, which seems to be related to this change:

`Failed tests:   testIgnoreStatsWithSignedSortOrder(org.apache.parquet.format.converter.TestParquetMetadataConverter): Stats should be empty: num_nulls: 3, min/max not defined
`

rdblue:
@nandorKollar, the test failure is because null count is no longer ignored when min and max are. I need to update the test before merging this.

PR #500, opened 2018-06-29
PARQUET-1344: Type builders don't honor new logical types

nandorKollar:
Call propert constructor when builder is caller with new logical type,
call the deprecated OriginalType version otherwise.

### DIFF
--- a/parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
+++ b/parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
@@ -26,6 +26,9 @@ import org.junit.Test;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
 import org.apache.parquet.schema.Type.Repetition;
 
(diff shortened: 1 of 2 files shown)


parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -1423,6 +1423,32 @@ public void testDecimalLogicalTypeWithDeprecatedPrecision() {
     Assert.assertEquals(expected, actual);
   }
 
+  @Test
+  public void testTimestampLogicalTypeWithUTCParameter() {
+    PrimitiveType utcMillisExpected = new PrimitiveType(REQUIRED, INT64, "aTimestamp",
+      LogicalTypeAnnotation.timestampType(true, LogicalTypeAnnotation.TimeUnit.MILLIS));
```
gszadovszky: I think, it would be more readable by using static imports. e.g. `timestampType(true, MILLIS)`

nandorKollar:
@gszadovszky updated the PR with your recommended changes.

PR #504, opened 2018-07-23
PARQUET-1353: Fix random data generator.

zivanfi:
The random data generator used for tests used to repeat the same value
over and over again.

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValues.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/statistics/RandomValues.java
@@ -84,19 +84,18 @@ public class RandomValues {
 
   static abstract class RandomBinaryBase<T extends Comparable<T>> extends RandomValueGenerator<T> {
     protected final int bufferLength;
-    protected final byte[] buffer;
 
     public RandomBinaryBase(long seed, int bufferLength) {
       super(seed);
 
       this.bufferLength = bufferLength;
-      this.buffer = new byte[bufferLength];
     }
 
     public abstract Binary nextBinaryValue();
 
     public Binary asReusedBinary(byte[] data) {
       int length = Math.min(data.length, bufferLength);
+      byte[] buffer = new byte[length];
       System.arraycopy(data, 0, buffer, 0, length);
       return Binary.fromReusedByteArray(data, 0, length);
     }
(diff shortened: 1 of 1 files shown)


rdblue:
Isn't this the intended behavior? The calls specifically mark the returned Binary objects as reused. Parquet should always do the right thing with reused arrays.

What is the impact of this? I'm concerned that committing this would stop testing the paths that are intended to support reused arrays.

gszadovszky:
@rdblue,
The purpose of the different types of `Binary` was never really clear to me. What is clear, though that the `byte[]` of a `Binary` is never copied at initialization whichever factory method is used (`fromReusedByteArray(byte[], int, int)`, `fromReusedByteArray(byte[])`, `fromConstantByteArray(byte[], int, int)`, or `fromConstantByteArray(byte[])`).
As the data generator is using the same `byte[]` to fill the values and to create `Binary` objects they will always contain the same values.

zivanfi:
@rdblue I don't think this is intended behavior. Certainly not in the way how it is used in `parquet-hadoop/src/test/java/org/apache/parquet/encodings/FileEncodingsIT.java`. Every time it calls binaryGenerator.nextBinaryValue(), the freshly generated random value overwrites every previous value as well, since a reference is held to the internal buffer:

|n-th call|returned value|accumulated list expected|accumulated list actual|
|---|---|---|---|
|1|6C|6C|6C|
|2|8F|6C 8F|8F 8F|
|3|52|6C 8F 52|52 52 52|
|4|B8|6C 8F 52 B8|B8 B8 B8 B8|

(Actual values are multiple bytes long, but I used a single-byte random value above for simplicity.)

We either have to make a new buffer in the random generator every time that a new value is requested or make a copy of the return value where that function is invoked. Since a bunch of identical values can not be considered random, there is no reason to use the funcion without decoupling the random value from the previous ones. This sugggest that we should change the logic inside the random number value generator and not outside at the place of call.

That being said, we don't see how a reused binary is supposed to enhance memory usage or runtime in general, since you have to make a copy of the returned value anyway. You probably know more about reused and constant binaries, could you please describe the use case scenarios to us or point us to some relevant documentation?

Thanks!

rdblue:
Sorry I didn't reply to this earlier. I think that this probably breaks what it was intended to test.

You can pass Parquet the same buffer with different values and Parquet should make a defensive copy of the buffer if it is clear that the buffer will be reused. By changing to a new buffer for every value, we are no longer testing the case where Parquet has to make that defensive copy.

We need to test this path for cases where input buffers are reused. For example, when you read an Avro file with a fixed field, the same GenericFixed instance is used and its buffer is passed as a reused buffer when writing to Parquet.

If this is misused for testing in `FileEncodingsIT`, then we should fix that class. The random data generator works like a common real-world case, so it shouldn't change.

zivanfi:
Ryan, could you point me to the code location where Parquet makes such defensive copies? I would like to check how removing that defensive copying affects test results. Thanks!

rdblue:
Defensive copies are part of the behavior of `Binary`. If the binary is marked as reused, then `getBytes` will make a copy. That's a more reliable way of making sure the copy happens if needed: The producer of the binary marks whether or not it will be reused and the consumer of a binary uses `getBytes` if it will be kept around, and `getBytesUnsafe` if it will be consumed `immedaitely`.

PR #505, opened 2018-07-24
PARQUET-1355: Improvement Binary write performance

wangyum:
Details can be found here: [PARQUET-1355](https://issues.apache.org/jira/browse/PARQUET-1355).
The write performance will be increased from `50983 ms` to `45423 ms`, close to `44432 ms`.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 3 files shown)


rdblue:
I agree that we want to avoid exposing the internal buffer. If we did that, we would lose information about whether or not it is reused and we would also break the abstraction. Is there a way to get the performance gain without exposing the underlying byte array? I'm surprised that using a ByteBuffer, which should be backed by the same bytes, is so much slower.

wangyum:
@rednaxelafx Do you have a good idea?

scottcarey:
ByteBuffer has always been significantly slower, in my experience, if you are not reading/writing in large chunks (e.g. any reading of singular values like ints/longs versus copying out or in large byte ranges).

The JVM can optimize loops over byte[] much more easily, the ByteBuffer abstraction gets in the way -- firstly by having at least two implementations in memory it introduces virtual dispatch that is usually optimized away but not always, and its harder for the JVM to see through the interface to elide redundant bounds-checks and similar.

A ByteBuffer can match a byte array if you use certain sun.misc.Unsafe access methods, maybe the Java 9+ VarHandle stuff can also match it, but I have not tried.

gatorsmile:
@rdblue Does this mean a performance regression is introduced in the write path of binary data after we upgrade Parquet from 1.10 to 1.8 in Spark?

PR #506, opened 2018-07-30
PARQUET-1280: [parquet-protobuf] Use maven protoc plugin

nandorKollar:
Use Maven plugin to generate test sources for Protobuf. Unfortunately I couldn't find any "official" plugin for this, not sure if the one I picked is the best choice.

### DIFF
--- a/parquet-protobuf/pom.xml
+++ b/parquet-protobuf/pom.xml
@@ -32,7 +32,6 @@
   <properties>
     <elephant-bird.version>4.4</elephant-bird.version>
     <protobuf.version>3.5.1</protobuf.version>
-    <!-- allow using protoc from an alternative path -->
   </properties>
 
 
--- a/parquet-protobuf/pom.xml
+++ b/parquet-protobuf/pom.xml
@@ -32,8 +32,8 @@
   <properties>
     <elephant-bird.version>4.4</elephant-bird.version>
     <protobuf.version>3.5.1</protobuf.version>
+    <protoc.version>3.5.1</protoc.version>
     <!-- allow using protoc from an alternative path -->
-    <protoc.path>protoc</protoc.path>
   </properties>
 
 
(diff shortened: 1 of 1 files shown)


parquet-protobuf/pom.xml
```diff
@@ -148,28 +148,24 @@
...
+              <protocVersion>${protoc.version}</protocVersion>
+              <addProtoSources>all</addProtoSources>
+              <outputDirectory>${project.build.directory}/generated-test-sources/java</outputDirectory>
+              <inputDirectories>
+                <include>src/test/resources</include>
+              </inputDirectories>
+            </configuration>
           </execution>
         </executions>
       </plugin>
```
qinghui-xu: Looks good to me

qinghui-xu:
It might be interesting to merge it soon.

qinghui-xu:
Let's merge it? @nandorKollar

nandorKollar:
@qinghui-xu sorry, though I wish I could, but I can't merge this PR, since I'm not a committer on Parquet. In addition, AFAIK an other committers approval is needed to get this merged.

parquet-protobuf/pom.xml
```diff
@@ -32,8 +32,8 @@
   <properties>
     <elephant-bird.version>4.4</elephant-bird.version>
     <protobuf.version>3.5.1</protobuf.version>
+    <protoc.version>3.5.1</protoc.version>
```
gszadovszky: Why this new property is required next to the existing `protobuf.version` with the same value?

nandorKollar: Good question. Can't think of any scenario where having two version (one for protobuf dependency and one for protoc plugin) would be beneficial.

gszadovszky: So, it seems some cleanup might be required.
I think, it is good to use the same version number for the `protobuf-java` version and the protoc binary version. Let's use the one we already have.
We also have the `protoc.path` property which is not required anymore.

gszadovszky:
@nandorKollar, this way we do not need the protobuf in our travis environment. Could you please remove the related parts

qinghui-xu:
Thanks for merging this PR ;-)

PR #507, opened 2018-07-31
PARQUET-1364: Invalid row indexes for pages starting with nulls

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -84,6 +84,10 @@ private void definitionLevel(int definitionLevel) {
 
   private void repetitionLevel(int repetitionLevel) {
     repetitionLevelColumn.writeInteger(repetitionLevel);
+    assert pageRowCount == 0 ? repetitionLevel == 0 : true : "Every page shall start on record boundaries";
```
zhaochengzhch, line 87: What is the logic of adding verification here? I have encountered a situation where the valuecount is 0 but the replicationlevel is not 0. Is this situation itself normal? Why do you need to add this check after columnindex

gszadovszky: @zhaochengzhch, I think the message describes it. We require to end/start pages at record boundaries so the repetition level shall be `0` when the page row count is `0` (which means we are starting a page). If the repetition level is not `0` at this point it breaks the mentioned requirement which is needed for column indexes.

PR #508, opened 2018-08-02
PARQUET-1367 upgrade a few libs for security reasons

mattdarwin:
There are a number of libraries which need updating.  Among other reasons, there are [several security issues filed in CVE for Hadoop](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=hadoop) and [guava](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-10237)


parquet-hadoop/pom.xml
```diff
@@ -76,7 +76,7 @@
     <dependency>
       <groupId>org.xerial.snappy</groupId>
       <artifactId>snappy-java</artifactId>
-      <version>1.1.2.6</version>
```
nandorKollar: could you please link the CVEs for snappy-java?

nandorKollar:
Looks like there are compilation failures, probably because of the Guava upgrade. Could you please have a look at it?

parquet-hadoop/pom.xml
mattdarwin: There is no CVE that I found for snappy-java.  But here at Oracle the security team won't approve use of software more than 18 months or 5 versions old.  snappy-java 1.1.26 is 26 months and 9 releases old.  For that reason I included it in the PR.  (Same applies for fastutil and slf4j.)

mattdarwin:
Thanks, I've now pushed in the fix.

nandorKollar (approved):
LGTM

mattdarwin:
I attempted to upgrade elephant bird (since 4.4 is also quite old), but it appears that from version 4.5 onwards they removed some packages, including `com.twitter.elephantbird.thrift.test`, so I had to revert it.

Fokko:
Just rebased. It looks like a lot of the stuff is already on the master branch.

Fokko:
Closing this since it is outdated

PR #510, opened 2018-08-03
PARQUET-1368: ParquetFileReader should close its input stream for the failure in constructor

HyukjinKwon:
This PR proposes to close the stream open in ParquetFileReader's constructor when it throws an error when it is used (`readFooter`), which causes a resource leak. Otherwise, looks there's no way to close it outside.

For more details, please see the JIRA ticket https://issues.apache.org/jira/browse/PARQUET-1368.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -686,7 +686,14 @@ public class ParquetFileReader implements Closeable {
     this.file = file;
     this.f = file.newStream();
     this.options = options;
-    this.footer = readFooter(file, options, f, converter);
+    try {
+      this.footer = readFooter(file, options, f, converter);
+    } catch (Exception e) {
+      // In case that reading footer throws an exception in the constructor, the new stream
+      // should be closed. Otherwise, there's no way to close this outside.
+      f.close();
+      throw e;
+    }
     this.fileMetaData = footer.getFileMetaData();
     this.blocks = filterRowGroups(footer.getBlocks());
     for (ColumnDescriptor col : footer.getFileMetaData().getSchema().getColumns()) {


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -686,7 +686,14 @@ public ParquetFileReader(InputFile file, ParquetReadOptions options) throws IOEx
     this.file = file;
     this.f = file.newStream();
     this.options = options;
-    this.footer = readFooter(file, options, f, converter);
+    try {
+      this.footer = readFooter(file, options, f, converter);
+    } catch (Exception e) {
```
nandorKollar: Why catch the exception and rethrowing it instead of closing the stream in finally?

HyukjinKwon: In that case, this will always be closed.

nandorKollar: Ah, ok, sorry. Next time I should read the comment in the PR before asking silly questions.

By the way, isn't this a problem in the other constructors in this class, and shouldn't we extend the scope of this try-catch until the end of the constructor? I guess if filterRowGroups throws an exception for some reason, then it could result in unclosed streams too no?

HyukjinKwon: Ah, it's fine. It took me a while before I replay since that's a common pattern. 

I was thinking about that too. Can the same thing happen in other constructors? I thought other constructors are fine since other constructors look not directly using the stream and deprecated.

HyukjinKwon:
Let me also cc @rdblue as well who's also active in Spark side.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
nandorKollar: I'm not sure if any exception could happen in other constructors, but in my opinion in this case it would make sense to wrap the entire constructor body in try-catch blocks like you did. @gszadovszky what do you think?

rdblue:
+1

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: I don't think any other code paths need this change, unless I missed something.

I'm not following the logic of expanding the try-catch blocks. I'm going to go ahead and merge this. If we want to follow up with more fixes, lets do that in other issues. Thanks, everyone!

zivanfi: I agree with Nandor. Creating a Closable object in a constructor means that if anything throws an exception (a RuntimeException possibly) it the rest of the constructor, the Closable won't be closed.

rdblue: Makes sense. Do you want to follow this up with a patch to expand the try/catch?

HyukjinKwon:
Thanks guys!

dongjoon-hyun:
Thank you, @HyukjinKwon !

PR #511, opened 2018-08-06
PARQUET-1371: Time/Timestamp UTC normalization parameter doesn't work

nandorKollar:


nandorKollar:
@zivanfi added a new test case for incompatible converted/logical types. Could you please review?

PR #512, opened 2018-08-14
PARQUET-1381: Add merge blocks command to parquet-tools

kgalieva:
Current implementation of merge command in parquet-tools doesn't merge row groups, just places one after the other. 
Add another command to be able to merge small blocks into larger ones up to specified size in bytes limit.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
@@ -262,4 +262,9 @@ class ColumnChunkPageWriteStore implements PageWriteStore {
     }
   }
 
+  void flushToFileWriter(ColumnDescriptor path, ParquetFileWriter writer) throws IOException {
+    ColumnChunkPageWriter pageWriter = writers.get(path);
+    pageWriter.writeToFileWriter(writer);
+  }
+
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1194,7 +1194,7 @@ public class ParquetFileReader implements Closeable {
    * @return the ByteBuffer blocks
    * @throws IOException if there is an error while reading from the stream
    */
-  public List<ByteBuffer> readBlocks(SeekableInputStream f, long offset, int length) throws IOException {
+  List<ByteBuffer> readBlocks(SeekableInputStream f, long offset, int length) throws IOException {
     f.seek(offset);
 
     int fullAllocations = length / options.getMaxAllocationSize();
@@ -1218,7 +1218,7 @@ public class ParquetFileReader implements Closeable {
     return buffers;
   }
 
-  public Optional<PageReader> readColumnInBlock(int blockIndex, ColumnDescriptor columnDescriptor) throws IOException {
+  Optional<PageReader> readColumnInBlock(int blockIndex, ColumnDescriptor columnDescriptor) {
     BlockMetaData block = blocks.get(blockIndex);
     if (block.getRowCount() == 0) {
       throw new RuntimeException("Illegal row group of 0 rows");
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -40,6 +40,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -56,6 +57,7 @@ import org.apache.parquet.ParquetReadOptions;
 import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.page.DictionaryPageReadStore;
+import org.apache.parquet.column.page.PageReader;
 import org.apache.parquet.compression.CompressionCodecFactory.BytesInputDecompressor;
 import org.apache.parquet.filter2.compat.FilterCompat;
 import org.apache.parquet.filter2.compat.RowGroupFilter;
@@ -1160,27 +1162,8 @@ public class ParquetFileReader implements Closeable {
      * @throws IOException if there is an error while reading from the stream
      */
     public List<Chunk> readAll(SeekableInputStream f) throws IOException {
-      List<Chunk> result = new ArrayList<Chunk>(chunks.size());
-      f.seek(offset);
-
-      int fullAllocations = length / options.getMaxAllocationSize();
-      int lastAllocationSize = length % options.getMaxAllocationSize();
-
-      int numAllocations = fullAllocations + (lastAllocationSize > 0 ? 1 : 0);
-      List<ByteBuffer> buffers = new ArrayList<>(numAllocations);
-
-      for (int i = 0; i < fullAllocations; i += 1) {
-        buffers.add(options.getAllocator().allocate(options.getMaxAllocationSize()));
-      }
-
-      if (lastAllocationSize > 0) {
-        buffers.add(options.getAllocator().allocate(lastAllocationSize));
-      }
-
-      for (ByteBuffer buffer : buffers) {
-        f.readFully(buffer);
-        buffer.flip();
-      }
+      List<Chunk> result = new ArrayList<>(chunks.size());
+      List<ByteBuffer> buffers = readBlocks(f, offset, length);
 
       // report in a counter the data we just scanned
       BenchmarkCounter.incrementBytesRead(length);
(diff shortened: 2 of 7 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1206,4 +1208,63 @@ public long endPos() {
...
   }
 
+
+  public Optional<PageReader> readColumnInBlock(int blockIndex, ColumnDescriptor columnDescriptor) throws IOException {
+    BlockMetaData block = blocks.get(blockIndex);
+    if (block.getRowCount() == 0) {
+      throw new RuntimeException("Illegal row group of 0 rows");
+    }
+    Optional<ColumnChunkMetaData> mc = findColumnByPath(block, columnDescriptor.getPath());
+    if (mc.isPresent()) {
```
alexeyzavyalov: Maybe `return mc.map(column -> {
 ChunkDescriptor chunk = new ChunkDescriptor(columnDescriptor, column, column.getStartingPos(), (int) column.getTotalSize());
return readChunk(f, chunk)
});` ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1206,4 +1208,63 @@ public long endPos() {
...
+    }
+
+    ByteBufferInputStream stream = ByteBufferInputStream.wrap(buffers);
+    Chunk chunk = new WorkaroundChunk(descriptor, stream.sliceBuffers(descriptor.size), f);
+    return chunk.readAllPages();
+  }
+
+  private Optional<ColumnChunkMetaData> findColumnByPath(BlockMetaData block, String[] path) {
+    for (ColumnChunkMetaData column : block.getColumns()) {
+      if (Arrays.equals(column.getPath().toArray(), (path))) {
```
alexeyzavyalov: I think you could remove unnecessary  brackets around `path`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1206,4 +1208,63 @@ public long endPos() {
...
+  private ColumnChunkPageReader readChunk(SeekableInputStream f, ChunkDescriptor descriptor) throws IOException {
+    f.seek(descriptor.fileOffset);
+
+    int fullAllocations = descriptor.size / options.getMaxAllocationSize();
+    int lastAllocationSize = descriptor.size % options.getMaxAllocationSize();
+
+    int numAllocations = fullAllocations + (lastAllocationSize > 0 ? 1 : 0);
+    List<ByteBuffer> buffers = new ArrayList<>(numAllocations);
+
+    for (int i = 0; i < fullAllocations; i += 1) {
```
alexeyzavyalov: i++

kgalieva:
@alexeyzavyalov Thank you! Implemented all suggested changes.

zivanfi:
Could you please reference the JIRA number in the PR title? The standard format is "PARQUET-###: Description". If there is no JIRA yet, could you please create one? Thanks!

kgalieva:
@zivanfi and @nandorKollar thank you for quick feedback and good suggestions! Agree with you, changed the code and added JIRA task with description.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -63,18 +102,32 @@ public MergeCommand() {
...
+      "which usually leads to bad query performance. " +
+      "To have adjacent small blocks merged together use -b option. " +
+      "Blocks will be grouped into larger one until the upper bound is reached. " +
+      "Default block upper bound 128 MB and default compression SNAPPY can be customized using -l and -c options";
   }
 
   @Override
   public void execute(CommandLine options) throws Exception {
+    boolean mergeBlocks = options.hasOption('b');
+    int maxBlockSize = options.hasOption('l')? Integer.parseInt(options.getOptionValue('l')) : 128;
```
zivanfi: I may be mistaken, but I think this will result in a desired row group size of 128 bytes, not 128 megabytes.

I would suggest using the ParquetWriter.DEFAULT_BLOCK_SIZE constant instead of a hard-coded value.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ParquetTripletUtils.java
```diff
@@ -0,0 +1,63 @@
...
+ * under the License.
+ */
+package org.apache.parquet.hadoop.util;
+
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.ColumnReader;
+import org.apache.parquet.column.ColumnWriter;
+import org.apache.parquet.schema.PrimitiveType;
+
+public class ParquetTripletUtils {
```
nandorKollar: Since apart from ParquetFileWriter it seems that no other code uses this utils, I'd recommend moving the only method into a private method in ParquetFileWriter.

kgalieva:
@zivanfi , @nandorKollar thank you! Implemented changes and added tests.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -262,4 +262,9 @@ public void flushToFileWriter(ParquetFileWriter writer) throws IOException {
     }
   }
 
+  public void flushToFileWriter(ColumnDescriptor path, ParquetFileWriter writer) throws IOException {
```
nandorKollar: If I'm not mistaken, the visibility of this method can be restricted to package private

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -519,6 +531,106 @@ public void appendFile(InputFile file) throws IOException {
...
+  private String[] getExisingParentPath(ColumnDescriptor path, MessageType inputFileSchema) {
+    List<String> parentPath = Arrays.asList(path.getPath());
+    while (parentPath.size() > 0 && !inputFileSchema.containsPath(parentPath.toArray(new String[parentPath.size()]))) {
+      parentPath = parentPath.subList(0, parentPath.size() - 1);
+    }
+    return parentPath.toArray(new String[parentPath.size()]);
+  }
+
+  private List<ParquetFileReader> getReaders(List<InputFile> inputFiles) throws IOException {
+    List<ParquetFileReader> readers = new ArrayList<>();
```
nandorKollar: nit: the ArrayList size is know in advance, it would be nice to add as constructor parameter to avoid resize

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/BlocksCombiner.java
```diff
@@ -0,0 +1,101 @@
...
+      blocks.add(new SmallBlocksUnion(smallBlocks, largeBlockRecords));
+    }
+    return unmodifiableList(blocks);
+  }
+
+  public static void closeReaders(List<ParquetFileReader> readers) {
+    readers.forEach(r -> {
+      try {
+        r.close();
+      } catch (IOException e) {
```
nandorKollar: It's not very nice to swallow an exception. Either throw a runtime exception, or at least log that something happened while closing the file.

kgalieva:
Thank you @nandorKollar! I've made the suggested changes.

nandorKollar:
Thanks, this sounds a very useful feature, I'd like to have a final look at the PR.

One notice, it seems you force push each time, could you please simply commit the changes instead? If there are multiple commits for a PR, those will be merged into one when the PR is merged, and if you commit instead of force push, it is easier to see the changes you for the response for the reviews.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -49,12 +58,43 @@
...
+  private static final Options OPTIONS;
+  static {
+    OPTIONS = new Options();
+
+    Option block = Option.builder("b")
+      .longOpt("block")
+      .desc("Merge adjacent blocks into one up to upper bound size limit default to 128 MB")
+      .build();
+
+    Option limit = Option.builder("l")
```
gszadovszky: Thanks a lot for working on this.

Instead of grabbing some write properties like `block size` and `compression` and adding command line options for them I would suggest allowing to set any of the [parquet properties](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java#L132) in a future proof way.
Some ideas:
- Have a command line parameter that argument is a key-value pair; the parameter can be used multiple times
- Have a command line parameter that argument is a list of key-value pairs
- Have a command line parameter that argument is a file that contains the key-value pairs

Multiple solutions might also make sense (e.g. set the key-value pairs from command line as well as from file).
The help shall reference the some docs or the source code for the up-to-date list of available options. It shall also list some of the most important options like `parquet.block.size`, `parquet.compression`, `parquet.page.size` or `parquet.writer.max-padding`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1206,4 +1189,72 @@ public long endPos() {
 
   }
 
+  /**
+   * @param f file to read the blocks from
+   * @return the ByteBuffer blocks
+   * @throws IOException if there is an error while reading from the stream
+   */
+  public List<ByteBuffer> readBlocks(SeekableInputStream f, long offset, int length) throws IOException {
```
nandorKollar: The visibility could be changed to package private.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1206,4 +1189,72 @@ public long endPos() {
...
+    }
+
+    for (ByteBuffer buffer : buffers) {
+      f.readFully(buffer);
+      buffer.flip();
+    }
+    return buffers;
+  }
+
+  public Optional<PageReader> readColumnInBlock(int blockIndex, ColumnDescriptor columnDescriptor) throws IOException {
```
nandorKollar: The visibility could be restricted to package protected. nit: this method doesn't throw IOException.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/BlocksCombiner.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+    if (!smallBlocks.isEmpty()) {
+      blocks.add(new SmallBlocksUnion(smallBlocks, largeBlockRecords));
+    }
+    return unmodifiableList(blocks);
+  }
+
+  public static void closeReaders(List<ParquetFileReader> readers) {
+    readers.forEach(r -> {
```
nandorKollar: With this logic, if there's an exception when closing a reader, then the rest remains unclosed. Though in a tool like this this is not a big deal, if one uses this method for other purposes, it could lead resource leakage.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -519,6 +531,106 @@ public void appendFile(InputFile file) throws IOException {
...
+        }
+        if (columnIndex == 0) {
+          this.startBlock(smallBlocks.getRowCount());
+        }
+        columnWriteStoreV1.flush();
+        store.flushToFileWriter(path, this);
+      }
+      this.endBlock();
+    }
+    this.end(new HashMap<>());
```
nandorKollar: nit: Collections.emptyMap() would be reasonable here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -519,6 +531,106 @@ public void appendFile(InputFile file) throws IOException {
...
+
+    BlocksCombiner.closeReaders(readers);
```
nandorKollar: Readers should be closed in a finally block to avoid unclosed readers in case of exception.

I was also wondering if it would be possible to keep only two files open at the same time: one which Parquet writes to, and one which is currently being processed. Since it seems sufficient for me to keep only these files open, and close the current one as soon as the processing of it is finished, and then open the next one, we won't keep files open longer than needed, and in addition we don't keep too many files open for no reason at the same time. I'm not sure if this makes sense, just a thought. If it is too complicated to implement, then I'm fine with the current approach too.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
zivanfi: It seems that we have several options here with no clear best choice. Since this review is already getting quite long, I think we can defer to do these in a follow-up change if the need arises. Are you okay with that Gabor?

gszadovszky: Agreed.

zivanfi:
I merged your commit @kgalieva, thanks for your contribution!

PR #513, opened 2018-08-17
PARQUET-1383: Parquet tools should indicate UTC parameter for time/timestamp types

nandorKollar:
This pull request addresses two topics:
- write logical type in parquet tools meta besides original type
- take to UTC normalized parameter into account when printing time/timestamp values (using stringifiers)

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
@@ -310,7 +310,7 @@ public abstract class PrimitiveStringifier {
   };
 
   static final PrimitiveStringifier TIMESTAMP_MILLIS_UTC_STRINGIFIER = new DateStringifier(
-    "TIMESTAMP_MILLIS_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSSZ") {
+    "TIMESTAMP_MILLIS_UTC_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSSZ") {
     @Override
     Instant getInstant(long value) {
       return Instant.ofEpochMilli(value);
@@ -318,7 +318,7 @@ public abstract class PrimitiveStringifier {
   };
 
   static final PrimitiveStringifier TIMESTAMP_MICROS_UTC_STRINGIFIER = new DateStringifier(
-    "TIMESTAMP_MICROS_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ") {
+    "TIMESTAMP_MICROS_UTC_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ") {
     @Override
     Instant getInstant(long value) {
       return Instant.ofEpochSecond(MICROSECONDS.toSeconds(value), MICROSECONDS.toNanos(value % SECONDS.toMicros(1)));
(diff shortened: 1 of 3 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -331,6 +330,32 @@ private long convert(long duration, TimeUnit from, TimeUnit to, TimeUnit higher)
...
+        long toMillis(long value) {
+          return value;
+        }
+      };
+    }
+    return new DateStringifier(
+      "TIMESTAMP_MICROS_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSS z", timeZone) {
+      @Override
+      public String stringify(long value) {
+        return super.stringify(value) + String.format("%03d", Math.abs(value % 1000));
```
zivanfi: Does this work as intended? It seems to me that it appends the micros after the already stringified millis-timestamp, but that does not end in the millis, but in the time zone instead (yyyy-MM-dd'T'HH:mm:ss.SSS z).

I would expect this to print something like "2018-08-17T07:41:12.345 America/Los_Angeles678".

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -331,6 +330,32 @@ private long convert(long duration, TimeUnit from, TimeUnit to, TimeUnit higher)
     }
   };
 
+  static PrimitiveStringifier createTimestampStringifier(final LogicalTypeAnnotation.TimeUnit timeUnit, final boolean isAdjustedToUTC) {
```
zivanfi: Since others may use this code as a reference for proper timestamp-handling, I think it would better to use the correct types (`Instant` for UTC-normalized and `LocalDateTime` for non-UTC-normalized) instead of just printing the timestamps correctly.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -309,48 +309,89 @@ Instant getInstant(long value) {
...
     }
+  };
 
+  static final PrimitiveStringifier TIMESTAMP_MICROS_UTC_STRINGIFIER = new DateStringifier(
+    "TIMESTAMP_MICROS_UTC_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ") {
     @Override
-    public String stringify(long micros) {
-      return toTimeString(micros, MICROSECONDS);
+    Instant getInstant(long value) {
+      return Instant.ofEpochSecond(MICROSECONDS.toSeconds(value), MICROSECONDS.toNanos(value % SECONDS.toMicros(1)));
```
zivanfi: I think this will lead to incorrect results with negative numbers.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -165,6 +167,7 @@ public void testTimestampMillisStringifier() {
     PrimitiveStringifier stringifier = PrimitiveStringifier.TIMESTAMP_MILLIS_STRINGIFIER;
 
     assertEquals("1970-01-01T00:00:00.000", stringifier.stringify(0l));
+    assertEquals("1970-01-01T00:00:00.000+0000", PrimitiveStringifier.TIMESTAMP_MILLIS_UTC_STRINGIFIER.stringify(0l));
```
zivanfi: Could you make this more consistent? Currently one stringifier is in a local variable, while the other is referenced directly.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
nandorKollar: @zivanfi I extended the test cases to run on both UTC and non-UTC stringifiers, and they pass. Could you please give some hints which values could be problematic?

zivanfi: I thought that negative numbers (representing timestamps before 1970-01-01 00:00:00) would behave incorrectly because of splitting to whole seconds plus nanoseconds part using the % operator. For example, 1.2 would split into 1 + 0.2, which is good, but -1.2 would split into -2 + 0.8 instead of -1 + -0.2. However, based on your test it works correctly, so it seems that I was mistaken (although I still don't see why it works).

PR #517, opened 2018-08-22
PARQUET-1399: Move parquet-mr related code from parquet-format

gszadovszky:


gszadovszky:
I kept the history of the moved files. To be committed with merge so the history will be kept in parquet-mr as well.

parquet-format-structures/pom.xml
```diff
@@ -0,0 +1,184 @@
...
+         </executions>
+       </plugin>
+      <!-- thrift -->
+      <plugin>
+        <groupId>org.apache.thrift.tools</groupId>
+        <artifactId>maven-thrift-plugin</artifactId>
+        <version>0.1.11</version>
+        <configuration>
+          <thriftSourceRoot>${parquet.thrift.path}</thriftSourceRoot>
+          <thriftExecutable>${thrift.executable}</thriftExecutable>
```
rdblue: Should this be `format.thrift.executable`? `thrift.executable` is used by the parquet-thrift module and we want to be able to set it separately.

parquet-format-structures/pom.xml
```diff
@@ -0,0 +1,184 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
+      <version>${thrift.version}</version>
```
rdblue: Similar here, should this be a different version of thrift for the format? Or do we want to use the same one?

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -0,0 +1,243 @@
...
+import org.apache.parquet.format.event.EventBasedThriftReader;
+import org.apache.parquet.format.event.TypedConsumer.I32Consumer;
+import org.apache.parquet.format.event.TypedConsumer.I64Consumer;
+import org.apache.parquet.format.event.TypedConsumer.StringConsumer;
+
+/**
+ * Utility to read/write metadata
+ * We use the TCompactProtocol to serialize metadata
+ *
+ * @author Julien Le Dem
```
rdblue: I recently had to remove all of these in the parquet-mr project in the commit to fix javadoc for Java 8. It would be nice to remove them for this import and make sure that the javadoc compiles.

pom.xml
```diff
@@ -154,6 +155,17 @@
     </dependency>
   </dependencies>
 
+  <dependencyManagement>
+    <dependencies>
+      <!-- Force using the proper version of thrift from the transitive dependencies -->
```
rdblue: Doesn't this force transitive dependencies to use the same version of thrift? I wouldn't say "from the transitive dependencies" since those are the ones that are changed.

parquet-format-structures/pom.xml
gszadovszky: `parquet-format` and `parquet-mr` use the same `0.9.3` thrift.
I separate the configuration for `parquet-thrift` and `format` but keep the same values.

pom.xml
gszadovszky: Actually, this is a solution for the problem of `com.twitter.elephantbird` bringing in an old version of thrift. So, I move this to the submodules where `elephantbird` is used and modify the comment accordingly.

PR #518, opened 2018-08-30
PARQUET-1408: Make parquet-tools display fields with missing values

rushton:
When using parquet-tools on a parquet file with null records the null columns are omitted from the output.

Example:
```
scala> case class Foo(a: Int, b: String)
defined class Foo

scala> org.apache.spark.sql.SparkSession.builder.getOrCreate.createDataset((0 to 1000).map(x => Foo(1,null))).write.parquet("/tmp/foobar/")
```
Pre-patch:
```
☁  parquet-tools [master] ⚡  java -jar target/parquet-tools-1.10.1-SNAPSHOT.jar cat -j /tmp/foobar/part-00000-436a4d37-d82a-4771-8e7e-e4d428464675-c000.snappy.parquet | head -n5
{"a":1}
{"a":1}
{"a":1}
{"a":1}
{"a":1}
```
Post-patch:
```
☁  parquet-tools [master] ⚡  java -jar target/parquet-tools-1.10.1-SNAPSHOT.jar cat -j /tmp/foobar/part-00000-436a4d37-d82a-4771-8e7e-e4d428464675-c000.snappy.parquet | head -n5
{"a":1,"b":null}
{"a":1,"b":null}
{"a":1,"b":null}
{"a":1,"b":null}
{"a":1,"b":null}
 ```

### DIFF
--- a/parquet-tools/src/test/java/org/apache/parquet/tools/read/TestSimpleReadSupport.java
+++ b/parquet-tools/src/test/java/org/apache/parquet/tools/read/TestSimpleReadSupport.java
@@ -115,7 +115,6 @@ public class TestSimpleReadSupport extends DirectWriterTest{
         printer = new PrintWriter(sw);
         r.prettyPrintJson(printer);
         String jsonOutput = sw.toString();
-        System.out.println(jsonOutput);
 
         Assert.assertTrue(jsonOutput.contains("B\":21"));
         Assert.assertTrue(jsonOutput.contains("element\":34"));

### CODE REVIEW

**Comment by zivanfi** [comment]
I'll take another look, thanks.

For future reviews In general I would suggest you to address review comments in a separate commit instead of amending and force-pushing. A separate commit makes it easier to see the changes in the PR and when a committer merges the PR, they will squash the whole PR into a single commit anyway.

It is also recommended to not auto-format whole files, because the resulting whitespace changes add clutter to the PR diff and the git blame history and more importantly can lead to conflicts in the future when cherry-picking, backporting or merging in general. Fortunately in this PR this has only happened to a few comment lines, so it's not a problem here, but for more substantial unrelated whitespace changes we usually ask the author to revert affected lines, which is a lot more work than not changing them in the first place.

File: parquet-tools/src/test/java/org/apache/parquet/tools/read/TestSimpleReadSupport.java [hunk 1]
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.parquet.hadoop.example.GroupWriteSupport;
+import org.apache.parquet.schema.MessageType;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.parquet.hadoop.ParquetWriter;
+import org.apache.parquet.hadoop.ParquetReader;
+import org.apache.parquet.hadoop.example.ExampleParquetWriter;
+import org.apache.parquet.example.data.Group;
+import org.apache.parquet.hadoop.example.GroupReadSupport;
+
+public class TestSimpleReadSupport {
```
zivanfi: This unit test fails in the Travis build: testNullValues(org.apache.parquet.tools.read.TestSimpleReadSupport): optional fixed_len_byte_array(3) bar is not a group

parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java
```diff
@@ -47,7 +47,13 @@
     Option help = OptionBuilder.withLongOpt("json")
                                .withDescription("Show records in JSON format.")
                                .create('j');
+
+    Option showEmpty = OptionBuilder.withLongOpt("show-empty")
+                        .withDescription("Show fields with null values.")
+                        .create('a');
```
zivanfi: The long name of the option and single-letter abbreviation should be related, but "show-empty" does not suggest 'a' at all. Some alternatives you could consider:

- leave "show-empty" but change the short form to 'e' as in _empty_.
- leave 'a' as the short form and rename the long option to something with an 'a', like 'all-values-including-empty' (although it's a bit too long).
- use 'n' for the short form and "show-null-values" or "show-nulls" for the long name.

parquet-tools/src/main/java/org/apache/parquet/tools/read/SimpleReadSupport.java
```diff
@@ -28,8 +28,23 @@
 import org.apache.parquet.schema.MessageType;
 
 public class SimpleReadSupport extends ReadSupport<SimpleRecord> {
+  private final boolean showEmpty;
+
+  public SimpleReadSupport() {
+      super();
```
zivanfi: (nit) super() seems unnecessary here and the other constructor, since it automatically happens anyway.

parquet-tools/src/test/java/org/apache/parquet/tools/read/TestSimpleReadSupport.java
zivanfi: Earlier you mentioned that you would like to add a unit test to check whether nested types work correctly, but the unit test only contains a record consisting of two primitive types. What you should test is not a field that is _set_ to null, but a list or map that _contains_ null values.

zivanfi: See https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists in the specification.

rushton: seems a change in master broke, this. I built against master and get a similar error when running the `cat` tool: ```☁  parquet-tools [master] java -jar target/parquet-tools-1.10.1-SNAPSHOT.jar  cat -debug -j /tmp/foobar/part-00000-1ca7e0f8-30ab-4ce9-860d-a791e2df081e-c000.snappy.parquet  
java.lang.ClassCastException: required int32 element is not a group
```

I will investigate a fix

**Comment by addomafi** [comment]
I just want to notice that this PR works well for cat command, but i guess that it's not covering other commands like head.

Follow the error that i'm facing with head command:
```
hadoop jar parquet-tools/target/parquet-tools-1.10.1-SNAPSHOT.jar head --show-nulls --debug example.parquet
java.lang.NumberFormatException: null
	at java.lang.Long.parseLong(Long.java:552)
	at java.lang.Long.parseLong(Long.java:631)
	at org.apache.parquet.tools.command.HeadCommand.execute(HeadCommand.java:82)
	at org.apache.parquet.tools.Main.main(Main.java:223)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.apache.hadoop.util.RunJar.run(RunJar.java:318)
	at org.apache.hadoop.util.RunJar.main(RunJar.java:232)
java.lang.NumberFormatException: null
```

PR #519, opened 2018-08-31
PARQUET-1388: Nanosecond precision time and timestamp - parquet-mr

nandorKollar:
This PR introduces nanosecond precision time and timestamp types to parquet-mr. It depends on https://github.com/apache/parquet-format/pull/102, should be committed once this format change is released.

Nanosecond precision is introduced only into the new logical type API, the old version (ConvertedType in format and OriginalType in mr) doesn't have a corresponding enum value for nanoseconds. Therefore in the Thrift schema these fields will be null, only new releases can interpret nanosecond precision, older readers can only see the physical type.

In addition to nanosecond precision, I also refactored the modules to use the new logical type API for internal decisions (e.g.: replaced the OriginalType-based switch cases, for example in type builder when checking if the proper annotation is present on physical type). This part (
commit with title: Refactor modules to use the new logical type API) doesn't need new format release, and can be split out from this PR if needed.

### DIFF
--- a/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
+++ b/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
@@ -23,6 +23,7 @@ import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.MICROS;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.MILLIS;
+import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.NANOS;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.dateType;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.decimalType;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.intType;
@@ -247,6 +248,8 @@ public class SchemaConverter {
           return primitive(INT32, timeType(true, MILLIS));
         } else if (bitWidth == 64 && timeUnit == TimeUnit.MICROSECOND) {
           return primitive(INT64, timeType(true, MICROS));
+        } else if (bitWidth == 64 && timeUnit == TimeUnit.NANOSECOND) {
+          return primitive(INT64, timeType(true, NANOS));
         }
         throw new UnsupportedOperationException("Unsupported type " + type);
       }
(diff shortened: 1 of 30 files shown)


gszadovszky:
I would suggest implement the refactor part as a separate JIRA. It is more a follow-up task for PARQUET-1253.
It also can be committed separately (as it does not require any format change).

nandorKollar:
Ok, created a separate PR for the refactor: https://github.com/apache/parquet-mr/pull/520

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -306,6 +307,19 @@ long toMillis(long value) {
     }
   };
 
+  static final PrimitiveStringifier TIMESTAMP_NANOS_STRINGIFIER = new DateStringifier(
+    "TIMESTAMP_NANOS_STRINGIFIER", "yyyy-MM-dd'T'HH:mm:ss.SSS") {
+    @Override
+    public String stringify(long value) {
+      return super.stringify(value) + String.format("%06d", Math.abs(value % 1000_000));
```
zivanfi: Nit: 1_000_000 (also a few lines below)

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -200,6 +202,28 @@ public void testTimestampMicrosStringifier() {
...
+  public void testTimestampNanosStringifier() {
+    PrimitiveStringifier stringifier = PrimitiveStringifier.TIMESTAMP_NANOS_STRINGIFIER;
+
+    assertEquals("1970-01-01T00:00:00.000000000", stringifier.stringify(0l));
+
+    Calendar cal = Calendar.getInstance(UTC);
+    cal.clear();
+    cal.set(2053, Calendar.JULY, 10, 22, 13, 24);
+    cal.set(Calendar.MILLISECOND, 84);
+    long micros = cal.getTimeInMillis() * 1000_000 + 1900;
```
zivanfi: Please rename the variable to nanos and use 1_000_000 instead of 1000_000.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -200,6 +202,28 @@ public void testTimestampMicrosStringifier() {
...
+    assertEquals("2053-07-10T22:13:24.084001900", stringifier.stringify(micros));
+
+    cal.clear();
+    cal.set(1848, Calendar.MARCH, 15, 9, 23, 59);
+    cal.set(Calendar.MILLISECOND, 765);
+    micros = cal.getTimeInMillis() * 1000_000 - 1;
+    assertEquals("1848-03-15T09:23:59.765000001", stringifier.stringify(micros));
```
zivanfi: The timestamp from the calendar is 1848-03-15T09:23:59.765. You subtract 1 nanosec from it, so it should be 1 nanosec earlier than 1848-03-15T09:23:59.765, which is 1848-03-15T09:23:59.765999999 and not 1848-03-15T09:23:59.765000001. Why does this pass?

PR #522, opened 2018-09-14
PARQUET-1417: BINARY_AS_SIGNED_INTEGER_COMPARATOR fails with IOBE for the same arrays with the different length

vvysotskyi:


vvysotskyi:
@gszadovszky, thanks for the review!

PR #523, opened 2018-09-18
PARQUET-1337: Current block alignment logic may lead to several row groups per block

zivanfi:


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -950,7 +950,24 @@ public long getPos() throws IOException {
   }
 
   public long getNextRowGroupSize() throws IOException {
-    return alignment.nextRowGroupSize(out);
+    return getNextRowGroupSize(out.getPos());
+  }
+
+  public long getNextRowGroupSize(long outPos) {
```
gszadovszky: Not sure if these methods need to be public. If yes, please comment at least that these are for internal use only.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -139,21 +145,41 @@ public long getDataSize() {
...
+          memSize, nextRowGroupSize, newMemSize, minSizeForAlignment);
+        memSize = newMemSize;
+        recordSize = memSize / recordCount;
+        // Start a new group if we are inside the padding area.
+        if (memSize > minSizeForAlignment) {
+          startNewRowGroup = true;
+          LOG.debug("mem size {} > {}: flushing {} records to disk.", memSize, minSizeForAlignment, recordCount);
+        }
+      }
+      if (startNewRowGroup) {
```
gszadovszky: nit: don't really like this flagging concept in the same method. I would suggest putting the related code to the place where the flag is true.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1102,22 +1135,61 @@ public void alignForRowGroup(PositionOutputStream out) throws IOException {
...
       if (isPaddingNeeded(remaining)) {
-        return rowGroupSize;
+        // A padding will be added to the end of the current row group,
+        // therefore the next row group will start at a block boundary.
+        remaining = dfsBlockSize;
       }
 
-      return Math.min(remaining, rowGroupSize);
+      if (rowGroupSize < remaining - maxPaddingSize) {
+        return rowGroupSize - maxPaddingSize/2;
```
gszadovszky: I think, the correction with maxPaddingSize/2 is not required here.

parquet-hadoop/src/test/java/org/apache/parquet/BlockAlignmentIT.java
```diff
@@ -0,0 +1,352 @@
...
+    fixedBinaryGenerator = new RandomValues.FixedGenerator(random.nextLong(), FIXED_LENGTH);
+  }
+
+  private void printParameters() {
+    LOG.info("Testing {}/{}/{} encodings with the following settings:", writerVersion, paramTypeName, compression);
+    LOG.info("  recordCount = {}, columnCount = {}, pageSize = {},", recordCount, columnCount, pageSize);
+    LOG.info("  rowGroupSize = {}, dfsBlockSize = {}, maxPaddingSize = {}", rowGroupSize, dfsBlockSize, maxPaddingSize);
+  }
+
+  //@Test
```
gszadovszky: Why is it commented out?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
zivanfi: Good point, made them package-private.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
zivanfi: I restructured the code, but now I have a return inside the middle of the function, personally I find this more confusing.

gszadovszky: I don't have problem with the return. But it is more a personal opinion. I let you choose which one you prefer.

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -63,14 +64,45 @@ public FallbackValuesWriter(I initialWriter, F fallBackWriter) {
...
     }
-    final double utilization = initialWriter.getUtilization();
     final long dictEncodedByteSize = initialWriter.getBufferedSize();
-    final long weightedSize = (long) (utilization * rawDataByteSize + (1 - utilization) * dictEncodedByteSize);
+    final double utilization = initialWriter.getUtilization();
+    long weightedSize;
+    if (utilization < UTILIZATION_THRESHOLD) {
+      weightedSize = dictEncodedByteSize;
+    } else {
+      final double weight_for_raw_size = (utilization - UTILIZATION_THRESHOLD) / (1 - UTILIZATION_THRESHOLD);
```
gszadovszky: nit: In java, only constants should contain `_` characters

rdblue:
@zivanfi, @gszadovszky, what is this fixing?

I don't think that the description -- that there could be multiple row groups per HDFS block -- is a bad thing. Row group size ends up being correlated with memory consumption, so it's reasonable to want to use several row groups in the same HDFS block. Smaller row groups allow fine-grained splitting if you have the parallelism and less memory consumption even if your tasks are a full block.

Also, when I looked through this I saw a bunch of changes to memory management and dictionary pages. How is that related to row group and block alignment?

rdblue:
@zivanfi, @gszadovszky, can you explain how this was fixed?

zivanfi:
@rdblue I started writing a DD about the solution that I plan to send to the dev list on Monday. Although Gabor approved this change, I don't plan to merge it soon as there are even more measures needed to address all problems and I will also have to resolve conflicts with the column indexes feature.

rdblue:
Sounds great, thank you!

zivanfi:
@rdblue, you can now review the [design doc](https://docs.google.com/document/d/1FJAVwzszZGkxZa8FtKtSbgBKm7qkS4cXuNW8hl4YKwU/edit#) for details. Thanks!

PR #530, opened 2018-10-08
PARQUET-1440: Convert int32 or int64 decimal values to BigDecimals

ryangardner:
This pull requests updates the `DecimalConverter` to override the `addInt` and `addLong` methods so it will return a BigDecimal with the proper scale when reading an `int32` or `int64` that is a `DECIMAL` with a specific scale. 

Without this override in place, the scale factor is ignored and the `int32 DECIMAL` values are displayed inaccurately


zivanfi:
Could you please open a JIRA about this issue and reference its number in PR title? Thanks!

ryangardner:
> Could you please open a JIRA about this issue and reference its number in PR title? Thanks!

I did. Let me know if there's anything else I can do.

zivanfi:
Merged, thanks!

PR #531, opened 2018-10-18
PARQUET-1414: Limit page size based on maximum row count

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreBase.java
```diff
@@ -183,6 +186,9 @@ public void close() {
   @Override
   public void endRecord() {
     ++rowCount;
+    if (rowCount >= rowCountForNextRowCountCheck) {
```
rdblue: This is in a tight loop, so I think we should be a bit more careful. I think this check could go inside size check `if` statement if we guarantee that `rowCountForNextSizeCheck <= rowCountForNextRowCountCheck` by ensuring the size check count isn't larger than the row count check. Then we would only need to do one test for every row.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriteStoreBase.java
```diff
@@ -190,13 +190,18 @@ public void endRecord() {
 
   private void sizeCheck() {
     long minRecordToWait = Long.MAX_VALUE;
+    long maxUnwrittenRows = 0;
```
rdblue: The logic using maxUnwrittenRows is confusing to me. What does it mean for a row to be "unwritten"?

I think this should calculate the next size check row count more directly.

zivanfi:
@rdblue Sure, sorry about that. Next time we will give more time to review.

PR #532, opened 2018-10-22
PARQUET-1228: parquet-format code changes for encryption support

ggershinsky:


parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,76 @@
...
+package org.apache.parquet.format;
+
+import java.io.IOException;
+
+public interface BlockCipher{
+
+  public interface Encryptor{
+    /**
+     * Encrypts the plaintext.
+     * Make sure the returned contents starts at offset 0 and fills up the byte array.
```
zivanfi: (nit) This "make sure" sound like the caller has to make sure to fulfill these conditions, when in fact it is the method implementation that does this.

```suggestion
     * The ciphertext starts at offset 0 of the return value and fills up the entire byte array.
```

Similarly in the other methods below.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -216,8 +277,41 @@ private static TProtocol protocol(InputStream from) {
   private static InterningProtocol protocol(TIOStreamTransport t) {
     return new InterningProtocol(new TCompactProtocol(t));
   }
+  
+  private static InputStream decryptStream(InputStream from, BlockCipher.Decryptor decryptor) throws IOException {
```
zivanfi: Could you please use more descriptive variable names instead of i32rd, clen, cbuf, pbuf, tmb, plen and i32out?

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
ggershinsky: SGTM.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: Sure, will do.

ggershinsky:
will be re-opened on an updated branch. all comments to be factored in.

PR #534, opened 2018-11-04
Adding change in the parquet record-reader so that it reads the first value lazily

26rajata:
Original problem:-
when we combine parquet splits using the elephant-bird library and try to read in multiple parquet splits in one mapper, it ignores the last value of split and reads the first value twice of the subsequent split.

For Example:-
If we try to read 4 partitions of parquet files in one map task then, we end up duplicating 3 records and end up missing three records

partition1 || partition 2 || partition 3 || partition 4
abc || xyz || rst || ghf
.... ||.... || .... || ....
..... ||... ||.... || ....
def || pqr || qwe || tyu

So we end up reading def, pqr, qwe twice ( Last record)
and ends up missing xyz , rst, ghf in our map task. (First record).

Root cause:-
1) when we try to combine multiple splits using the elephant-bird library, elephant-bird library creates all the record-readers of multiple-splits at the beginning( reads the first value of all the record-readers at the beginning itself and initialize the container with first value )

2) While reading values from the record-readers one by one, elephant-bird library always updates the container of record-reader with the last value read, when it starts reading the new record reader after initialization we update container's value with the last read value to keep the container's value consistent. And in the process of doing so, we overwrite the already first value at the time of initialization of record-reader.

3) Elephant-bird Code pointer which overrides the first value
https://github.com/twitter/elephant-bird/blob/master/core/src/main/java/com/twitter/elephantbird/mapreduce/input/combine/CompositeRecordReader.java#L134. We override the values via setKeyValue to keep the container's value consistent with the last read value but in this case, we end up overriding the already read value
 at the time of initialization of record-reader

Solution :-
If we read values lazily not at initialization time, then it will fix this problem.
 We can read the parquet record-reader values lazily so that even if we update the record-reader values with the last value read of the previous record-reader, it will read the correct the first value from the file since we do not pre-fetch first value of record-reader at the of initialization.

### DIFF
--- a/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
+++ b/parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
@@ -23,6 +23,7 @@ import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.MICROS;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.MILLIS;
+import static org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit.NANOS;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.dateType;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.decimalType;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.intType;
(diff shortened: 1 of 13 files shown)


ttim:
Is it correct that basically next contract is broken:
```
val inputReader1 = ... from file1
val inputReader2 = ... from file2

val container = inputReader1.createValue
val firstValueFromFile2 = inputReader2.next(..., container)
// at this point container doesn't have actual first value from file2
```

Is it a contract for map reduce readers? If yes could you add one more test to show how this contract is broken. If it's not maybe we should fix elephantbird instead?

Also could you rewrite PR description to highlight original problem, root cause, test and solution in different sections?

26rajata:
@rdblue @zivanfi  can you review?

nandorKollar:
@26rajata could you please open a Jira for this issue and assign it to yourself? Here's the link to it: https://jira.apache.org/jira/projects/PARQUET/summary

tlazaro:
This is solved by https://github.com/apache/parquet-mr/pull/844 and can be closed.

gszadovszky:
Thanks, @tlazaro.

PR #536, opened 2018-11-07
PARQUET-1435: Benchmark filtering column-indexes

gszadovszky:


parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/FilteringBenchmarks.java
```diff
@@ -0,0 +1,430 @@
...
+@Measurement(iterations = 50, batchSize = 1)
+@OutputTimeUnit(MILLISECONDS)
+public class FilteringBenchmarks {
+  private static final int RECORD_COUNT = 2_000_000;
+  private static final Logger LOGGER = LoggerFactory.getLogger(FilteringBenchmarks.class);
+
+  /*
+   * For logging human readable file size
+   */
+  private static class FileSize {
```
zivanfi: You could have used org.apache.commons.io.FileUtils.byteCountToDisplaySize.

gszadovszky: This implementation is not exactly the same as `FileUtils.byteCountToDisplaySize`. This one writes floats with 2 precision. Moreover, the concept behind this one is to create a new object with the file size and create the string representation only when it is required so if the log level is lower then the string calculation would not invoked.
(However, this is not crucial for this benchmark as it is not production code. I can remove this implementation and use `FileUtils.byteCountToDisplaySize` if you think it would make to code simpler.)

zivanfi: Just mentioned as an FYI, no need to modify anything.

PR #550, opened 2018-11-14
PARQUET-138: Allow merging more restrictive field in less restrictive field

ntrinquier:
https://issues.apache.org/jira/browse/PARQUET-138

### DIFF
--- a/parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
+++ b/parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
@@ -20,9 +20,6 @@ package org.apache.parquet.schema;
 
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
 import static org.apache.parquet.schema.OriginalType.LIST;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT32;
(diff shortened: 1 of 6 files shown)


parquet-column/src/test/java/org/apache/parquet/schema/TestMessageType.java
```diff
@@ -90,15 +90,15 @@ public void testMergeSchema() {
         new PrimitiveType(REQUIRED, BINARY, "a"));
 
     try {
-      t3.union(t4);
+      t4.union(t3);
```
rdblue: Why is this dependent on which schema is the left-hand operand?

ntrinquier:
With the change I made, merging something less restrictive into something
more restrictive would fail, while the opposite would now succeed. So
merging is dependent on the “direction” of the merge: you merge X into Y.

If we want merging to always succeed, it is also possible, and in this case
there would not be any “direction”: you merge X and Y.

What do you think the behavior should be?

rdblue:
I see no reason why union would not by symmetric. Otherwise, this is just changing the direction, which isn't very useful.

ntrinquier:
I see no reason either, actually. Updated the PR to change that!

parquet-column/src/test/java/org/apache/parquet/schema/TestType.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.parquet.schema;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestType {
+  @Test
+  public void test() {
```
nandorKollar: nit: can this test have a more descriptive name like for example testLeastRestrictiveRepetition and the class named as TestRepetitionType

ntrinquier: Done!

tpmusielak:
Hey team,

I have came across this issue in my project and I was wondering if you are likely to make this change anytime soon? 

Thanks!

rdblue:
@tpmusielak, we need to make sure we don't think this behavior change will cause problems first. I'd like to see it in the next release.

mccheah:
What's required to get this in for the next release? Do we need more tests to ensure we don't break existing behavior?

ntrinquier:
Can we put this behind a flag to stay backward compatible and users can opt-in to use it?

vinooganesh:
Hey @julienledem @rdblue - friendly ping here and agreed it would be nice to get this into the next release. If this looks good, can we merge?

rdblue:
Thanks for the reminder, @vinooganesh. I see no reason why it would not be safe for union to return the same result when operand position is swapped.

boriskar:
hey guys, 
so in what pyspark version this will work? 
I have a problem to recreate a summary file after writing more restrictive schema into a less restrictive one

ntrinquier:
@boriskar I guess this is more a question for apache/spark. This change is available in parquet 1.11.0 or higher, so spark needs to bump its version to that (it's currently on 1.10.1).

PR #552, opened 2018-11-19
PARQUET-1407: Data loss on duplicate values with AvroParquetWriter/Reader

nandorKollar:
Based on #551: added a test case for the fix.


rdblue:
Thanks, @nandorKollar!

nandorKollar:
Thanks Ryan, though the credit goes to you, I just wrote a test! :)

PR #558, opened 2018-11-26
[WIP] PARQUET-1464: Allowing backwards compatible projections

moulimukherjee:
This PR addresses two scenarios, which currently fails.

1. Adding an optional field to an existing thrift schema and making it part of the projection for reading data written using older thrift. 

2. Projecting a required field using optional.


parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java
```diff
@@ -349,6 +355,28 @@ void checkGroupContains(Type subType) {
...
+  void checkGroupCompatibility(Type subType) {
+    if (subType.isPrimitive()) {
+      throw new InvalidRecordException(subType + " found: expected " + this);
+    }
+    List<Type> fields = subType.asGroupType().getFields();
+    for (Type otherType : fields) {
+      if(this.containsField(otherType.getName())) {
+        Type thisType = this.getType(otherType.getName());
+        thisType.checkCompatibility(otherType);
+      } else if (otherType.getRepetition() == Repetition.REQUIRED) {
```
moulimukherjee: Unlike `checkContains` this throws an exception only when the extra field is `required`. Since, `optional` and `repeated` are 0 or 1 and 0 or more respectively, it's still compatible to have them as extra fields.

parquet-column/src/main/java/org/apache/parquet/schema/Type.java
```diff
@@ -333,6 +333,21 @@ void checkContains(Type subType) {
...
+  /**
+   * Checks compatibility with subType
+   * Allows subtype to have a less strict repetition
+   * For eg: a REQUIRED type is compatible with an OPTIONAL subType
+   *
+   * @param subType
+   */
+  void checkCompatibility(Type subType) {
+    if (!this.name.equals(subType.name)
+      || (this.repetition != subType.repetition && !this.repetition.isMoreRestrictiveThan(subType.repetition))) {
```
moulimukherjee: Similarly, unlike in `checkContains` this checks if the `subType`'s `repetition` is more restrictive or not. If it's less restrictive, it's still compatible. This would allow a required field to be projected using an optional field.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/ReadSupport.java
```diff
@@ -53,7 +53,7 @@ public static MessageType getSchemaForRead(MessageType fileMessageType, String p
   }
 
   public static MessageType getSchemaForRead(MessageType fileMessageType, MessageType projectedMessageType) {
-    fileMessageType.checkContains(projectedMessageType);
+    fileMessageType.checkCompatibility(projectedMessageType);
```
moulimukherjee: Instead of the stricter `checkContains`, using `checkCompatibility` instead.

moulimukherjee:
@isnotinvain Can you take a look?

PR #560, opened 2018-11-29
PARQUET-1441: SchemaParseException: Can't redefine: list in AvroIndexedRecordConverter

nandorKollar:
Parquet Avro reader can't convert such Parquet schemas, where a group field name is reused
in an inner structure. The converter creates Avro record schma in this case,
but in Avro record types should have a unique name, therefore the result is an invalid Avro
schema. This patch fixes this case by adding a namespace for the record if the name was
defined before, this way making the record names unique.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -34,7 +34,9 @@ import org.apache.parquet.schema.Types;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import static java.util.Optional.empty;
@@ -243,17 +245,18 @@ public class AvroSchemaConverter {
   }
 
   public Schema convert(MessageType parquetSchema) {
-    return convertFields(parquetSchema.getName(), parquetSchema.getFields());
+    return convertFields(parquetSchema.getName(), parquetSchema.getFields(), new HashMap<>());
   }
 
   Schema convert(GroupType parquetSchema) {
-    return convertFields(parquetSchema.getName(), parquetSchema.getFields());
+    return convertFields(parquetSchema.getName(), parquetSchema.getFields(), new HashMap<>());
   }
 
-  private Schema convertFields(String name, List<Type> parquetFields) {
+  private Schema convertFields(String name, List<Type> parquetFields, Map<String, Integer> names) {
     List<Schema.Field> fields = new ArrayList<Schema.Field>();
+    Integer nameCount = names.merge(name, 1, (k, v) -> v + 1);
     for (Type parquetType : parquetFields) {
-      Schema fieldSchema = convertField(parquetType);
+      Schema fieldSchema = convertField(parquetType, names);
       if (parquetType.isRepetition(REPEATED)) {
         throw new UnsupportedOperationException("REPEATED not supported outside LIST or MAP. Type: " + parquetType);
       } else if (parquetType.isRepetition(Type.Repetition.OPTIONAL)) {
@@ -264,12 +267,12 @@ public class AvroSchemaConverter {
             parquetType.getName(), fieldSchema, null, (Object) null));
       }
     }
-    Schema schema = Schema.createRecord(name, null, null, false);
+    Schema schema = Schema.createRecord(name, null, nameCount > 1 ? name + nameCount : null, false);
     schema.setFields(fields);
     return schema;
   }
 
-  private Schema convertField(final Type parquetType) {
+  private Schema convertField(final Type parquetType, Map<String, Integer> names) {
     if (parquetType.isPrimitive()) {
       final PrimitiveType asPrimitive = parquetType.asPrimitiveType();
       final PrimitiveTypeName parquetPrimitiveTypeName =
@@ -342,13 +345,13 @@ public class AvroSchemaConverter {
             }
             if (isElementType(repeatedType, parquetGroupType.getName())) {
               // repeated element types are always required
-              return of(Schema.createArray(convertField(repeatedType)));
+              return of(Schema.createArray(convertField(repeatedType, names)));
             } else {
               Type elementType = repeatedType.asGroupType().getType(0);
               if (elementType.isRepetition(Type.Repetition.OPTIONAL)) {
-                return of(Schema.createArray(optional(convertField(elementType))));
+                return of(Schema.createArray(optional(convertField(elementType, names))));
               } else {
-                return of(Schema.createArray(convertField(elementType)));
+                return of(Schema.createArray(convertField(elementType, names)));
               }
             }
           }
@@ -382,9 +385,9 @@ public class AvroSchemaConverter {
             }
             Type valueType = mapKeyValType.getType(1);
             if (valueType.isRepetition(Type.Repetition.OPTIONAL)) {
-              return of(Schema.createMap(optional(convertField(valueType))));
+              return of(Schema.createMap(optional(convertField(valueType, names))));
             } else {
-              return of(Schema.createMap(convertField(valueType)));
+              return of(Schema.createMap(convertField(valueType, names)));
             }
           }
 
@@ -395,7 +398,7 @@ public class AvroSchemaConverter {
         }).orElseThrow(() -> new UnsupportedOperationException("Cannot convert Parquet type " + parquetType));
       } else {
         // if no original type then it's a record
-        return convertFields(parquetGroupType.getName(), parquetGroupType.getFields());
+        return convertFields(parquetGroupType.getName(), parquetGroupType.getFields(), names);
       }
     }
   }
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -254,7 +254,7 @@ public class AvroSchemaConverter {
 
   private Schema convertFields(String name, List<Type> parquetFields, Map<String, Integer> names) {
     List<Schema.Field> fields = new ArrayList<Schema.Field>();
-    Integer nameCount = names.merge(name, 1, (k, v) -> v + 1);
+    Integer nameCount = names.merge(name, 1, (oldValue, value) -> oldValue + 1);
     for (Type parquetType : parquetFields) {
       Schema fieldSchema = convertField(parquetType, names);
       if (parquetType.isRepetition(REPEATED)) {
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/AvroTestUtil.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/AvroTestUtil.java
@@ -37,10 +37,14 @@ import org.junit.rules.TemporaryFolder;
 
 public class AvroTestUtil {
 
-  public static Schema record(String name, Schema.Field... fields) {
-    Schema record = Schema.createRecord(name, null, null, false);
+  public static Schema record(String name, String namespace, Schema.Field... fields) {
+    Schema record = Schema.createRecord(name, null, namespace, false);
     record.setFields(Arrays.asList(fields));
     return record;
+  }
+
+  public static Schema record(String name, Schema.Field... fields) {
+    return record(name, null, fields);
   }
 
   public static Schema.Field field(String name, Schema schema) {
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
@@ -37,6 +37,10 @@ import java.util.Collections;
 
 import static org.apache.avro.Schema.Type.INT;
 import static org.apache.avro.Schema.Type.LONG;
+import static org.apache.parquet.avro.AvroTestUtil.field;
+import static org.apache.parquet.avro.AvroTestUtil.optionalField;
+import static org.apache.parquet.avro.AvroTestUtil.primitive;
+import static org.apache.parquet.avro.AvroTestUtil.record;
 import static org.apache.parquet.schema.OriginalType.DATE;
 import static org.apache.parquet.schema.OriginalType.TIMESTAMP_MICROS;
 import static org.apache.parquet.schema.OriginalType.TIMESTAMP_MILLIS;
@@ -726,6 +730,65 @@ public class TestAvroSchemaConverter {
             }
           });
     }
+  }
+
+  @Test
+  public void testReuseNameInNestedStructure() throws Exception {
+    Schema innerA1 = record("a1", "a12",
+      field("a4", primitive(Schema.Type.FLOAT)));
+
+    Schema outerA1 = record("a1",
+      field("a2", primitive(Schema.Type.FLOAT)),
+      optionalField("a1", innerA1));
+    Schema schema = record("Message",
+      optionalField("a1", outerA1));
+
+    String parquetSchema = "message Message {\n" +
+        "      optional group a1 {\n" +
+        "        required float a2;\n" +
+        "        optional group a1 {\n" +
+        "          required float a4;\n"+
+        "         }\n" +
+        "      }\n" +
+        "}\n";
+
+    testParquetToAvroConversion(schema, parquetSchema);
+    testParquetToAvroConversion(NEW_BEHAVIOR, schema, parquetSchema);
+  }
+
+  @Test
+  public void testReuseNameInNestedStructureAtSameLevel() throws Exception {
+    Schema a2 = record("a2",
+      field("a4", primitive(Schema.Type.FLOAT)));
+    Schema a22 = record("a2", "a22",
+      field("a4", primitive(Schema.Type.FLOAT)),
+      field("a5", primitive(Schema.Type.FLOAT)));
+
+    Schema a1 = record("a1",
+      optionalField("a2", a2));
+    Schema a3 = record("a3",
+      optionalField("a2", a22));
+
+    Schema schema = record("Message",
+      optionalField("a1", a1),
+      optionalField("a3", a3));
+
+    String parquetSchema = "message Message {\n" +
+      "      optional group a1 {\n" +
+      "        optional group a2 {\n" +
+      "          required float a4;\n"+
+      "         }\n" +
+      "      }\n" +
+      "      optional group a3 {\n" +
+      "        optional group a2 {\n" +
+      "          required float a4;\n"+
+      "          required float a5;\n"+
+      "         }\n" +
+      "      }\n" +
+      "}\n";
+
+    testParquetToAvroConversion(schema, parquetSchema);
+    testParquetToAvroConversion(NEW_BEHAVIOR, schema, parquetSchema);
   }
 
   public static Schema optional(Schema original) {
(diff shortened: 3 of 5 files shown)


liorchaga:
Really looking forward for this patch!

So I've tested the patch locally, and it fails on following schema for 2nd parquet nested column:
```
optional group pv_acceptedLanguages (LIST) {
  repeated group list {
    optional binary element (STRING);
  }
}
```

This is the `GroupType` that represents following column in parquet file: 
```
{
   "name":"pv_acceptedLanguages",
   "type":{
      "type":"array",
      "elementType":"string",
      "containsNull":true
   },
   "nullable":true,
   "metadata":{
   }
}
```

The exception I'm getting is:
```
Exception in thread "main" java.lang.ClassCastException: optional binary element (STRING) is not a group
	at org.apache.parquet.schema.Type.asGroupType(Type.java:227)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.newConverter(AvroIndexedRecordConverter.java:168)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.access$200(AvroIndexedRecordConverter.java:48)
	at org.apache.parquet.avro.AvroIndexedRecordConverter$AvroArrayConverter$ElementConverter.<init>(AvroIndexedRecordConverter.java:381)
	at org.apache.parquet.avro.AvroIndexedRecordConverter$AvroArrayConverter.<init>(AvroIndexedRecordConverter.java:344)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.newConverter(AvroIndexedRecordConverter.java:172)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.<init>(AvroIndexedRecordConverter.java:94)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.<init>(AvroIndexedRecordConverter.java:66)
	at org.apache.parquet.avro.AvroCompatRecordMaterializer.<init>(AvroCompatRecordMaterializer.java:34)
	at org.apache.parquet.avro.AvroReadSupport.newCompatMaterializer(AvroReadSupport.java:144)
	at org.apache.parquet.avro.AvroReadSupport.prepareForRead(AvroReadSupport.java:136)
	at org.apache.parquet.hadoop.InternalParquetRecordReader.initialize(InternalParquetRecordReader.java:183)
	at org.apache.parquet.hadoop.ParquetReader.initReader(ParquetReader.java:156)
	at org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:135)
```

The exception is triggered just after successfully processing the following parquet column (first nested column it encountered):
```
{
   "name":"pv_events",
   "type":{
      "type":"array",
      "elementType":{
         "type":"struct",
         "fields":[
            {
               "name":"eventTime",
               "type":"long",
               "nullable":true,
               "metadata":{}
            },
            {
               "name":"name",
               "type":"string",
               "nullable":true,
               "metadata":{}
            },
            {
               "name":"type",
               "type":"string",
               "nullable":true,
               "metadata":{}
            }
         ]
      },
      "containsNull":true
   },
   "nullable":true,
   "metadata":{}
}
```

liorchaga:
@nandorKollar 
Attached is a sample parquet file containing two columns - availableLanguages and events. Both are nested columns.

Used the following code snippet to get avro schema:
```
AvroReadSupport readSupport = new AvroReadSupport();
ParquetReader<GenericData.Record> avroParquetReader =
        AvroParquetReader.builder(readSupport,
             new Path("file:///path/to/file/part-00000-a13d1582-160e-4e18-9be3-9b178641db76-c000.snappy.parquet"))
            .build();
GenericData.Record record = avroParquetReader.read();
record.getSchema();
````

[part-00000-a13d1582-160e-4e18-9be3-9b178641db76-c000.snappy.parquet.zip](https://github.com/apache/parquet-mr/files/2647579/part-00000-a13d1582-160e-4e18-9be3-9b178641db76-c000.snappy.parquet.zip)

**Comment by nandorKollar** [comment]
@liorchaga this particular case I think you should explicitly tell, that the file schema is written using 3-level structure by setting `ADD_LIST_ELEMENT_RECORDS` to `false` in the `Configuration` object, and build the reader with it:
```
    org.apache.hadoop.conf.Configuration conf = new org.apache.hadoop.conf.Configuration(false);
    conf.set(AvroSchemaConverter.ADD_LIST_ELEMENT_RECORDS, "false");
    AvroReadSupport readSupport = new AvroReadSupport();
    ParquetReader<GenericData.Record> avroParquetReader =
      AvroParquetReader.builder(readSupport,
        new Path("file:////Users/nkollar/Downloads/part-00000-a13d1582-160e-4e18-9be3-9b178641db76-c000.snappy.parquet"))
        .withConf(conf)
        .build();
    GenericData.Record record = avroParquetReader.read();
    record.getSchema();

```

**Comment by liorchaga** [comment]
Thanks @nandorKollar . For some reason I though that the patch obviates the need for this flag. 
Well, now it fails on array nested within array: 

> Exception in thread "main" org.apache.avro.SchemaParseException: Can't redefine: element2.element

[part-00000-83d942f8-dec0-4420-aaaa-4e3e33ba340a-c000.snappy.parquet.zip](https://github.com/apache/parquet-mr/files/2649092/part-00000-83d942f8-dec0-4420-aaaa-4e3e33ba340a-c000.snappy.parquet.zip)


parquet schema:

```
{
   "type":"struct",
   "fields":[
      {
         "name":"history",
         "type":{
            "type":"array",
            "elementType":{
               "type":"struct",
               "fields":[
                  {
                     "name":"user",
                     "type":"string",
                     "nullable":true,
                     "metadata":{}
                  },
                  {
                     "name":"historyDepthByUserHistoryType",
                     "type":{
                        "type":"array",
                        "elementType":{
                           "type":"struct",
                           "fields":[
                              {
                                 "name":"historyType",
                                 "type":"string",
                                 "nullable":true,
                                 "metadata":{}
                              },
                              {
                                 "name":"numEvents",
                                 "type":"long",
                                 "nullable":true,
                                 "metadata":{}
                              }
                           ]
                        },
                        "containsNull":true
                     },
                     "nullable":true,
                     "metadata":{}
                  }
               ]
            },
            "containsNull":true
         },
         "nullable":true,
         "metadata":{}
      }
   ]
}
```

**Comment by nandorKollar** [comment]
Strange, I tried, but I didn't get the exception you mentioned.

**Comment by liorchaga** [comment]
Well, just to make sure - this is the code I executed. Now I see it actually failed on the `avroParquetReader.read()`, and not on the schema. 
```
import java.io.IOException;

import org.apache.hadoop.hdfs.HdfsConfiguration;
import org.apache.parquet.avro.AvroParquetReader;
import org.apache.parquet.avro.AvroReadSupport;
import org.apache.parquet.avro.AvroSchemaConverter;
import org.apache.parquet.hadoop.ParquetReader;

public class Tester {
    public static void main(String[] args) throws IOException {
        HdfsConfiguration conf = new HdfsConfiguration();
        conf.set(AvroSchemaConverter.ADD_LIST_ELEMENT_RECORDS, "false");
        AvroReadSupport readSupport = new AvroReadSupport();
        ParquetReader<GenericData.Record> avroParquetReader =
                AvroParquetReader.builder(readSupport,
                        new Path("file:///path/to/file/part-00001-ea373355-e5b6-4372-b7cc-f707e94111fb-c000.snappy.parquet"))
                        .withConf(conf)
                        .build();
        GenericData.Record record = avroParquetReader.read();
        final Schema schema = record.getSchema();
    }
}
```

**Comment by liorchaga** [comment]
Ok, my bad. If failed on another file with wider schema, containing also this field. I'll check myself again (must be out of focus today)

**Comment by liorchaga** [comment]
Ok, so once I have two upper level arrays, and one of them contains nested arrays, it actually generates a valid schema, but fails on schema.toString() with `SchemaParseException: Can't redefine: element2.element`

schema:
```
message spark_schema {
  optional group history (LIST) {
    repeated group list {
      optional group element {
        optional binary user (UTF8);
        optional group historyDepthByUserHistoryType (LIST) {
          repeated group list {
            optional group element {
              optional binary historyType (UTF8);
              optional int64 numEvents;
            }
          }
        }
      }
    }
  }
  optional group campaigns (LIST) {
    repeated group list {
      optional group element {
        optional double calibratedRecsHistory;
        optional int64 campaignId;
        optional int64 eventTime;
        optional int64 recsHistory;
      }
    }
  }
}
```

file attached.
[part-00000-4b659403-e0ec-4f55-b15c-09c25894e4f1-c000.snappy.parquet.zip](https://github.com/apache/parquet-mr/files/2650048/part-00000-4b659403-e0ec-4f55-b15c-09c25894e4f1-c000.snappy.parquet.zip)

Full stack trace:
```
Exception in thread "main" org.apache.avro.SchemaParseException: Can't redefine: element2.element
	at org.apache.avro.Schema$Names.put(Schema.java:1128)
	at org.apache.avro.Schema$NamedSchema.writeNameRef(Schema.java:562)
	at org.apache.avro.Schema$RecordSchema.toJson(Schema.java:690)
	at org.apache.avro.Schema$UnionSchema.toJson(Schema.java:882)
	at org.apache.avro.Schema$ArraySchema.toJson(Schema.java:805)
	at org.apache.avro.Schema$UnionSchema.toJson(Schema.java:882)
	at org.apache.avro.Schema$RecordSchema.fieldsToJson(Schema.java:716)
	at org.apache.avro.Schema$RecordSchema.toJson(Schema.java:701)
	at org.apache.avro.Schema.toString(Schema.java:324)
	at org.apache.avro.Schema.toString(Schema.java:314)
	at java.lang.String.valueOf(String.java:2994)
	at java.io.PrintStream.println(PrintStream.java:821)
	at Tester.main(Tester.java:24)
```

**Comment by liorchaga** [comment]
Also  - when I'm running on full schema, same exception occurs on `avroParquetReader.read()`
I'll try to isolate and provide relevant schema

**Comment by liorchaga** [comment]
Well, turns out that if I just switch the order of columns in the schema, it fails on read. Attached file here:
[part-00000-471d7fc7-dea3-4198-9db3-3a968b1a075d-c000.snappy.parquet.zip](https://github.com/apache/parquet-mr/files/2650231/part-00000-471d7fc7-dea3-4198-9db3-3a968b1a075d-c000.snappy.parquet.zip)

Stack trace:
```
org.apache.avro.SchemaParseException: Can't redefine: element2.element
	at org.apache.avro.Schema$Names.put(Schema.java:1128)
	at org.apache.avro.Schema$NamedSchema.writeNameRef(Schema.java:562)
	at org.apache.avro.Schema$RecordSchema.toJson(Schema.java:690)
	at org.apache.avro.Schema$UnionSchema.toJson(Schema.java:882)
	at org.apache.avro.Schema$ArraySchema.toJson(Schema.java:805)
	at org.apache.avro.Schema$UnionSchema.toJson(Schema.java:882)
	at org.apache.avro.Schema$RecordSchema.fieldsToJson(Schema.java:716)
	at org.apache.avro.Schema$RecordSchema.toJson(Schema.java:701)
	at org.apache.avro.Schema.toString(Schema.java:324)
	at org.apache.avro.Schema.toString(Schema.java:314)
	at org.slf4j.helpers.MessageFormatter.safeObjectAppend(MessageFormatter.java:305)
	at org.slf4j.helpers.MessageFormatter.deeplyAppendParameter(MessageFormatter.java:277)
	at org.slf4j.helpers.MessageFormatter.arrayFormat(MessageFormatter.java:231)
	at org.slf4j.helpers.MessageFormatter.format(MessageFormatter.java:152)
	at org.slf4j.impl.Log4jLoggerAdapter.debug(Log4jLoggerAdapter.java:251)
	at org.apache.avro.SchemaCompatibility$ReaderWriterCompatiblityChecker.getCompatibility(SchemaCompatibility.java:228)
	at org.apache.avro.SchemaCompatibility.checkReaderWriterCompatibility(SchemaCompatibility.java:61)
	at org.apache.parquet.avro.AvroRecordConverter.isElementType(AvroRecordConverter.java:866)
	at org.apache.parquet.avro.AvroIndexedRecordConverter$AvroArrayConverter.<init>(AvroIndexedRecordConverter.java:333)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.newConverter(AvroIndexedRecordConverter.java:172)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.<init>(AvroIndexedRecordConverter.java:94)
	at org.apache.parquet.avro.AvroIndexedRecordConverter.<init>(AvroIndexedRecordConverter.java:66)
	at org.apache.parquet.avro.AvroCompatRecordMaterializer.<init>(AvroCompatRecordMaterializer.java:34)
	at org.apache.parquet.avro.AvroReadSupport.newCompatMaterializer(AvroReadSupport.java:144)
	at org.apache.parquet.avro.AvroReadSupport.prepareForRead(AvroReadSupport.java:136)
	at org.apache.parquet.hadoop.InternalParquetRecordReader.initialize(InternalParquetRecordReader.java:183)
	at org.apache.parquet.hadoop.ParquetReader.initReader(ParquetReader.java:156)
	at org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:135)
	at Tester.main(Tester.java:22)
```

**Comment by nandorKollar** [comment]
@liorchaga thanks for catching this, I think I found the problem: I used Map#merge wrongly. Pushed a new commit, it should fix this case.

**Comment by liorchaga** [comment]
Superb. I'll test it on full schema again soon

**Comment by liorchaga** [comment]
@nandorKollar fix works likes magic. Thanks!
I think it worth adding a unit test to cover it.

**Comment by heuermh** [comment]
Any updates to the state of this pull request?

This effects us downstream of Apache Spark, which is at the moment still going through the RC process for a version 2.4.1 release.  (This issue broke us as of Spark version 2.4.0)

https://issues.apache.org/jira/browse/SPARK-25588
https://github.com/bigdatagenomics/adam/issues/2058
https://github.com/bigdatagenomics/adam/pull/2056

**Comment by nandorKollar** [comment]
@rdblue what do you think, is using Avro namespace to solve this problem the right approach? It looks like name collision in Avro schema (while converting Parquet lists for example, where the group name is the same for embedded list) causes problems like this, and I think Avro namespaces would be a good option to solve issues like this.

**Comment by liorchaga** [comment]
I personally think this fix is superb. Already use it in production with shaded jar including tgis patch, reading and writing very complex parquet schema.
+1 for merging

**Comment by rdblue** [comment]
I think it is okay to use the namespace to solve this problem. The idea is to use the path within the schema as the namespace and the last component as the record name, right? So it would produce list_one.element and list_two.element instead of two records named element?

We fixed this in Iceberg by moving to IDs. You could consider doing something similar when there is a field ID defined on the Parquet type. We use `r12` for ID 12, for example.

**Comment by nandorKollar** [comment]
@rdblue something similar, but the namespace is not the entire path of the record, but simply the record name + a counter. So in your example the namespaces will be element1 and element2, and the two record names are element1.element and element2.element. Namespaces are used only when it is indeed required: when the record name was encountered before, so if there's only one list with element, then no namespace is used. Would you prefer entire path to the record as namespaces?

IDs you mean Parquet field IDs? In that case, IDs would give the namespaces for the Avro records right? That sounds also promising for me, however in that case the client should modify the Parquet schema by adding unique ID fields, since AFAIK IDs are optional.

**Comment by rdblue** [comment]
Sounds reasonable to me. I don't think it matters whether it is a path or a counter.

We used field IDs to name the record itself without using namespaces.

**Comment by heuermh** [comment]
Might it be possible to get this merged and into a bugfix release 1.10.2 soon?  If the Spark 2.4.1 RC process drags out a bit longer, it would be nice to get this in.

**Comment by rdblue** [comment]
I don't think it is necessary to get this into Spark 2.4.1. Spark doesn't use parquet-avro, except in tests.

**Comment by heuermh** [comment]
> I don't think it is necessary to get this into Spark 2.4.1. Spark doesn't use parquet-avro, except in tests.

Sorry if I sound frustrated, I feel like I keep having to re-explain myself.  The versions of Parquet and Avro in Spark are incompatible with each other, and downstream we don't have any way to workaround the conflict.  Per

https://issues.apache.org/jira/browse/SPARK-25588

We have been struggling with Spark's conflicting Parquet and Avro dependencies for many versions.  Our most recent workaround is to pin parquet-avro to version 1.8.1 and exclude all its transitive dependencies.  This workaround worked for 2.3.2, thus I gave the last Spark RC a non-binding +1.

https://github.com/bigdatagenomics/adam/blob/master/pom.xml#L520

That workaround does not work for 2.4.0, as this pinned version 1.8.1 conflicts at runtime with version 1.10.0 brought in by Spark.

```
$ mvn test
...
*** RUN ABORTED ***
  java.lang.NoSuchFieldError: BROTLI
  at org.apache.parquet.hadoop.metadata.CompressionCodecName.<clinit>(CompressionCodecName.java:31)
  at org.bdgenomics.adam.rdd.JavaSaveArgs$.$lessinit$greater$default$4(GenomicRDD.scala:78)
```
 
Removing the pinned version and dependency exclusions, bringing the build dependency version to 1.10.0, results in the error reported here in our unit tests under Spark version 2.4.0.

https://github.com/bigdatagenomics/adam/pull/2056

I believe this pull request will fix the `SchemaParseException` error, and when the Parquet dependency version in Spark is bumped to a Parquet release containing this fix, downstream projects such as ours will no longer be blocked going to Spark 2.4.x.

**Comment by rdblue** [comment]
@heuermh, you should be able to shade parquet-avro to avoid that conflict. Have you tried that?

**Comment by heuermh** [comment]
> you should be able to shade parquet-avro to avoid that conflict. Have you tried that?

I've not yet found any combination of dependency exclusion, version overrides, and dependency shading that works, not for lack of trying.

Spark 2.4.1 rc8 has failed and there might be a chance to get a PR in rc9 or a later release candidate that updates the Parquet dependency to include this fix.  How heavyweight is the release process here for a 1.10.2 release?

**Comment by rdblue** [comment]
@heuermh, have you tried setting the schemas manually instead of letting Parquet convert them?

Unfortunately, I think this needs to wait to be in the 1.11.0 release. 1.10.2 would be a patch release and we want to keep those to just bug fixes. While this is a bug, it can break existing code because it changes the names of structures in the Avro schema.

This could cause a problem when reading with a projection schema. Inside of unions, types are matched by name. Because this changes the full name of types, it could cause existing code to not be able to read unions when projecting with an Avro schema.

**Comment by heuermh** [comment]
> This could cause a problem when reading with a projection schema. Inside of unions, types are matched by name. Because this changes the full name of types, it could cause existing code to not be able to read unions when projecting with an Avro schema.

Ah right, good point.


> have you tried setting the schemas manually instead of letting Parquet convert them?

It is Spark SQL generating the schemas when saving a Dataset or DataFrame as Parquet; I'm not sure there is a hook to specify an Avro-generated schema there.  Will look into it.

In any case, I generated and deployed SNAPSHOT builds all the way up our stack from this branch (Parquet &rarr; Spark &rarr; [bdg-utils](https://github.com/bigdatagenomics/utils) &rarr; [ADAM](https://github.com/bigdatagenomics/adam)) and it seems to work for us.

**Comment by heuermh** [comment]
@zivanfi Thank you for the merge!  Which Parquet version would this fix be released in?

**Comment by zivanfi** [comment]
@heuermh Sorry, I can't tell, unfortunately I don't have too much impact on the release process.

**Comment by elkhand** [comment]
While trying to read with master branch build, no exceptions are thrown, but the values of repeated array types are set to null. I verified the values indeed exist in parquet file via `parquet-tools` and `pyspark`:
```
optional group trip_uuids (LIST) {
      repeated binary array (UTF8);
}
```
Corresponding avro schema:
```
{
              "name": "trip_uuids",
              "type": [
                "null",
                {
                  "type": "array",
                  "items": "string"
                }
              ],
              "default": null
}
```

I have this Hadoop configuration:
```
this.hadoopConf = new Configuration(false);
this.hadoopConf.setBoolean("parquet.avro.add-list-element-records", false);
this.hadoopConf.setBoolean("parquet.avro.write-old-list-structure", false);
```

**Comment by elkhand** [comment]
The issue was the parquet file was written with Parquet version 1.8.1 and avro 1.8.2, while trying to read with the built from master  (1.12) it causes extra issues. Maybe backward compatibility is broken.
and hadoop  conf I used was just default one:
```
this.hadoopConf = new Configuration();
```

PR #561, opened 2018-11-30
PARQUET-1455: [parquet-protobuf] Handle protobuf enum schema evolution and unknown enum value

qinghui-xu:


julienledem:
Benoit Hanotte: if you could give a code review and approve that'd be great

parquet-protobuf/pom.xml
```diff
@@ -159,28 +159,24 @@
       </plugin>
 
       <plugin>
-        <artifactId>maven-antrun-plugin</artifactId>
+        <groupId>com.github.os72</groupId>
+        <artifactId>protoc-jar-maven-plugin</artifactId>
+        <version>3.6.0.1</version>
```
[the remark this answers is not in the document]
qinghui-xu: The protoc (protobuf version) is defined by the plugin configuration `protocVersion`, so any recent version of plugin should be good.

parquet-protobuf/pom.xml
```diff
@@ -159,28 +159,24 @@
       </plugin>
 
       <plugin>
-        <artifactId>maven-antrun-plugin</artifactId>
+        <groupId>com.github.os72</groupId>
```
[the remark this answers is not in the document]
qinghui-xu: Agree. I'm trying to have #506 merged before this PR.

parquet-protobuf/pom.xml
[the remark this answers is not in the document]
BenoitHanotte: thanks, I just wanted to be sure

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoSchemaEvolutionTest.java
```diff
@@ -0,0 +1,68 @@
...
+  @Test
+  public void testEnumSchemaWriteV2ReadV1() throws IOException {
+    TestProto3SchemaV2.MessageSchema dataV2 = TestProto3SchemaV2.MessageSchema.newBuilder()
+      .setOptionalLabelNumberPair(TestProto3SchemaV2.MessageSchema.LabelNumberPair.SECOND)
+      .setOptionalString("string value")
+      .build();
+    Path file = writeMessages(dataV2);
+    List<TestProto3SchemaV1.MessageSchema> messagesV1 = readMessages(file, TestProto3SchemaV1.MessageSchema.class);
+    assertEquals(messagesV1.size(), 1);
+    assertEquals(messagesV1.get(0).getOptionalLabelNumberPairValue(), 2);
```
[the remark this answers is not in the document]
qinghui-xu: I'm not sure we need to test the label `UNKNOWN_ENUM_VALUE_*`, it is generated only when calling protobuf reflection API, and we are interested mainly in being able to get the number back.
It should be a test for protobuf reflection API that tests by passing an unknown number to an enum field, we can get a label `UNKNOWN_ENUM_VALUE_*` by using protobuf reflection API. Less relevant to put it here.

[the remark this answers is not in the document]
BenoitHanotte: OK, got it, thanks!

BenoitHanotte:
@qinghui-xu Can we remove the maven plugin part from this PR? Once it is removed I don't see anything left to address before we can merge :+1:

qinghui-xu:
Rebased on the lastest upstream master (with the maven-protoc-plugin) and force-pushed to the PR.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -130,13 +137,36 @@ public WriteContext init(Configuration configuration) {
...
+  public FinalizedWriteContext finalizeWrite() {
+    Map<String, String> protoMetadata = new HashMap<>();
+    for (Map.Entry<String, Map<String, Integer>> enumNameNumberMapping : protoEnumBookKeeper.entrySet()) {
+      StringBuilder nameNumberPairs = new StringBuilder();
+      if (enumNameNumberMapping.getValue().isEmpty()) {
+        // No enum is ever written to any column of this file, put an empty string as the value in the metadata
+        LOG.info("No enum is written for " + enumNameNumberMapping.getKey());
+      }
+      int idx = 0;
+      for (Map.Entry<String, Integer> nameNumberPair : enumNameNumberMapping.getValue().entrySet()) {
```
nandorKollar: nit: since Parquet already depends on Guava, we can use Guava's [Joiner](https://google.github.io/guava/releases/27.0.1-jre/api/docs/com/google/common/base/Joiner.html) class to stringify the map to spare this for loop, something like this:
```
Joiner.on(METADATA_ENUM_ITEM_SEPARATOR)
        .withKeyValueSeparator(METADATA_ENUM_KEY_VALUE_SEPARATOR)
        .join(enumNameNumberMapping.getValue())
```
I couldn't find anything to simplify the reading path. Just an idea, maybe it is better not to depend on 3rd party libs too much. :)

parquet-protobuf/src/test/java/org/apache/parquet/proto/TestUtils.java
```diff
@@ -145,22 +146,37 @@ private static void checkSameBuilderInstance(MessageOrBuilder[] messages) {
...
+   * @param <T>
+   * @return List of protobuf messages for the given type.
+   */
+  public static <T extends MessageOrBuilder> List<T> readMessages(Path file, Class<T> messageClass) throws IOException {
+    ParquetReader.Builder readerBuilder = ProtoParquetReader.builder(file);
+    if (messageClass != null) {
+      readerBuilder.set(ProtoReadSupport.PB_CLASS, messageClass.getName()).build();
+    }
+    ParquetReader reader = readerBuilder.build();
+    try {
```
nandorKollar: we could use try-with-resource construct here

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoRecordConverter.java
```diff
@@ -45,14 +48,13 @@ public void add(Object a) {
     }
   }
 
-
-  public ProtoRecordConverter(Class<? extends Message> protoclass, MessageType parquetSchema) {
-    super(new SkipParentValueContainer(), protoclass, parquetSchema);
+  public ProtoRecordConverter(Configuration conf, Class<? extends Message> protoclass, MessageType parquetSchema, Map<String, String> extraMetadata) {
```
nandorKollar: These two constructors are public in a public class. Can we mark the old constructors deprecated and keep them? Despite (AFAIK) this class is not supposed to be part of the public API, unfortunately it isn't package private, and clients could use it: a change like this could break their code. Though this scenario is unlikely, unfortunately we faced similar issues before.

parquet-protobuf/src/test/java/org/apache/parquet/proto/TestUtils.java
qinghui-xu: Done.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
qinghui-xu: I would prefer to not use guava if possible

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoRecordConverter.java
qinghui-xu: It sounds fair. I'll keep deprecated constructors for now.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
costimuraru: In order to make the method easier to read, if we're not going with guava (which I do like, though I don't see it being used in `parquet-proto`), we could extract the joining part into a dedicated method. Something like:

```java
  @Override
  public FinalizedWriteContext finalizeWrite() {
    Map<String, String> protoMetadata = new HashMap<>();
    for (Map.Entry<String, Map<String, Integer>> enumNameNumberMapping : protoEnumBookKeeper.entrySet()) {
      
      if (enumNameNumberMapping.getValue().isEmpty()) {
        // No enum is ever written to any column of this file, put an empty string as the value in the metadata
        LOG.info("No enum is written for " + enumNameNumberMapping.getKey());
      }
      
      String nameNumberPairs = joinNameNumberPairs(enumNameNumberMapping.getValue());
      protoMetadata.put(METADATA_ENUM_PREFIX + enumNameNumberMapping.getKey(), nameNumberPairs);
    }
    return new FinalizedWriteContext(protoMetadata);
  }

  public String getNameNumberPairs(Map<String, Integer> nameNumberMap) {
    StringBuilder nameNumberPairs = new StringBuilder();
    int idx = 0;
    for (Map.Entry<String, Integer> nameNumberPair : nameNumberMap.entrySet()) {
      nameNumberPairs.append(nameNumberPair.getKey())
        .append(METADATA_ENUM_KEY_VALUE_SEPARATOR)
        .append(nameNumberPair.getValue());
      idx ++;
      if (idx < nameNumberMap.size()) {
        nameNumberPairs.append(METADATA_ENUM_ITEM_SEPARATOR);
      }
    }
    return nameNumberPairs;
  }
```

What do you think?

browles:
Hey guys, what is the status on this PR? We currently experiencing this issue.

gszadovszky:
Seems like the guys missed to push it after approving. 
@qinghui-xu, are you willing to pick this up again and resolve the conflicts?

qinghui-xu:
Hey, guys.
Will be glad to work on it. Would try to push another patch at the beginning of next week.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
qinghui-xu: Sorry for the late reply. @costimuraru
I refactor out all the logic of building enum metadata in a method. I think this would be more clear.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -22,7 +22,10 @@
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
 import com.twitter.elephantbird.util.Protobufs;
+import org.apache.commons.lang.StringUtils;
```
gszadovszky, line 25: There is no direct dependency to `commons-lang` in `parquet-protobuf` so I would not start using it for such a simple functionality.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -190,25 +211,45 @@ private Converter newScalarConverter(ParentValueContainer pvc, Message.Builder p
...
-    private Map<Binary, Descriptors.EnumValueDescriptor> makeLookupStructure(Descriptors.FieldDescriptor enumFieldType) {
-      Descriptors.EnumDescriptor enumType = enumFieldType.getEnumType();
+    private Map<Binary, Descriptors.EnumValueDescriptor> makeLookupStructure(Descriptors.EnumDescriptor enumType) {
       Map<Binary, Descriptors.EnumValueDescriptor> lookupStructure = new HashMap<Binary, Descriptors.EnumValueDescriptor>();
 
-      List<Descriptors.EnumValueDescriptor> enumValues = enumType.getValues();
+      if (extraMetadata.containsKey(METADATA_ENUM_PREFIX + enumType.getFullName())) {
+        String enumNameNumberPairs = extraMetadata.get(METADATA_ENUM_PREFIX + enumType.getFullName());
+        if (StringUtils.isBlank(enumNameNumberPairs)) {
+          LOG.info("No enum is written for " + enumType.getFullName());
```
gszadovszky, line 236: As far as I understand this one will be logged quite often. We've already had some about logging too much. I would suggest having this on debug level instead.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -222,11 +263,37 @@ public ProtoEnumConverter(ParentValueContainer parent, Descriptors.FieldDescript
...
       if (protoValue == null) {
-        Set<Binary> knownValues = enumLookup.keySet();
-        String msg = "Illegal enum value \"" + binaryValue + "\""
-                + " in protocol buffer \"" + fieldType.getFullName() + "\""
-                + " legal values are: \"" + knownValues + "\"";
-        throw new InvalidRecordException(msg);
+        // in case of unknown enum value, protobuf is creating new EnumValueDescriptor with the unknown number
+        // and name as following "UNKNOWN_ENUM_VALUE_" + parent.getName() + "_" + number
+        // so the idea is to parse the name for data created by parquet-proto before this patch
+        String unknownLabel = new String(binaryValue.getBytes());
```
gszadovszky, line 269: I would suggest using `Binary.toStringUsingUTF8()` instead.

Fokko:
@qinghui-xu The CI is experiencing some connectivity issues, could you rebase against master to retrigger the build?

qinghui-xu:
@Fokko It's already on top of the master branch. Could you retrigger the CI by hand somehow?

gszadovszky:
Trying to re-trigger Travis by close-reopen

gszadovszky (approved):

PR #578, opened 2018-12-14
PARQUET-1478: Can't read spec compliant, 3-level lists via parquet-proto

nandorKollar:
Test to demonstrate the problem.

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
@@ -145,7 +145,7 @@ class ProtoMessageConverter extends GroupConverter {
       public Optional<Converter> visit(LogicalTypeAnnotation.MapLogicalTypeAnnotation mapLogicalType) {
         return of(new MapConverter(parentBuilder, fieldDescriptor, parquetType));
       }
-    }).orElse(newScalarConverter(parent, parentBuilder, fieldDescriptor, parquetType));
+    }).orElseGet(() -> newScalarConverter(parent, parentBuilder, fieldDescriptor, parquetType));
   }
 
   private Converter newScalarConverter(ParentValueContainer pvc, Message.Builder parentBuilder, Descriptors.FieldDescriptor fieldDescriptor, Type parquetType) {
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
@@ -54,13 +54,13 @@ class ProtoMessageConverter extends GroupConverter {
   private final Message.Builder myBuilder;
 
   // used in record converter
-  ProtoMessageConverter(ParentValueContainer pvc, Class<? extends Message> protoClass, GroupType parquetSchema) {
-    this(pvc, Protobufs.getMessageBuilder(protoClass), parquetSchema);
+  ProtoMessageConverter(ParentValueContainer pvc, Class<? extends Message> protoClass, GroupType parquetSchema, boolean specCompliant) {
+    this(pvc, Protobufs.getMessageBuilder(protoClass), parquetSchema, specCompliant);
   }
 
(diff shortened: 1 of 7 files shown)


parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -54,13 +54,13 @@
   private final Message.Builder myBuilder;
 
   // used in record converter
-  ProtoMessageConverter(ParentValueContainer pvc, Class<? extends Message> protoClass, GroupType parquetSchema) {
-    this(pvc, Protobufs.getMessageBuilder(protoClass), parquetSchema);
+  ProtoMessageConverter(ParentValueContainer pvc, Class<? extends Message> protoClass, GroupType parquetSchema, boolean specCompliant) {
+    this(pvc, Protobufs.getMessageBuilder(protoClass), parquetSchema, specCompliant);
   }
```
qinghui-xu: The parquet-protobuf should be able to read parquet spec compliant (3-level) repeated fields transparently, without relying a parameter/configuration.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -80,12 +80,18 @@
       Descriptors.FieldDescriptor protoField = protoDescriptor.findFieldByName(parquetField.getName());
 
       if (protoField == null) {
+        // Skip extra list/key_value layer for spec-compliant, 3-level structures,
+        // when the proto schema doesn't have a field with this name.
+        if (specCompliant &&
+          ("key_value".equals(parquetField.getName()) || "list".equals(parquetField.getName()))) {
+          continue;
+        }
         String description = "Scheme mismatch \n\"" + parquetField + "\"" +
```
qinghui-xu: This looks like a hacky way to fix the problem, it might be hard to maintain for later.

parquet-protobuf/src/test/java/org/apache/parquet/proto/utils/WriteUsingMR.java
```diff
@@ -55,7 +54,7 @@ public WriteUsingMR() {
   }
 
   public WriteUsingMR(Configuration conf) {
-    this.conf = new Configuration();
+    this.conf = conf;
```
qinghui-xu: Nice catch!

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -132,23 +138,24 @@ public void add(Object value) {
...
       }
 
       @Override
       public Optional<Converter> visit(LogicalTypeAnnotation.MapLogicalTypeAnnotation mapLogicalType) {
-        return of(new MapConverter(parentBuilder, fieldDescriptor, parquetType));
+        return of(new MapConverter(parentBuilder, fieldDescriptor, parquetType, specCompliant));
       }
-    }).orElse(newScalarConverter(parent, parentBuilder, fieldDescriptor, parquetType));
+    }).orElse(newScalarConverter(parent, parentBuilder, fieldDescriptor, parquetType, specCompliant));
   }
```
qinghui-xu: In fact, the root cause is we use orElse here, instead of orElseGet.
orElseGet is the lazy way to get the fallback value, while orElse is not. In the case of repeated fields (list and map), it was still trying to call newScalarConverter while using orELse, which is irrelevant (and thus the bug).

nandorKollar: Thanks @qinghui-xu , looks like the fix you suggested works well!

In this case, the good news: this is a bug, which is not yet released, because that orElse was added by me in PARQUET-1410 (my bad). Bad news: it is a regression, would be nice to get this committed before next release.

qinghui-xu: Thanks for the prompt fix.

PR #581, opened 2019-01-02
PARQUET-1485: Fix Snappy direct memory leak

liupc:
### What's the problem?
Snappy compressor/decompressor will not release direct buffer in time, it might cause direct memory oom, especially for those application whose direct memory is limited while heap memory is redundant.

This PR is to fix this issue by calling `cleaner.clean` to release directly memory in SnappyCompressor/SnappyDecompressor, it's implemented through reflection to solve incompatibility issues on java9+.

This PR also includes some minor changes: add initial buffer size to boost the memory allocation and avoid memory wasting in SnappyCompressor/SnappyDecompressor.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
@@ -26,6 +26,11 @@ import java.nio.ByteBuffer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * A Helper class which use reflections to clean up DirectBuffer. It's implemented for
+ * better compatibility with both java8 and java9+, because the Cleaner class is moved to
+ * another place since java9+.
+ */
 public class CleanUtil {
   private static final Logger logger = LoggerFactory.getLogger(CleanUtil.class);
   private static final Field CLEANER_FIELD;
@@ -33,23 +38,27 @@ public class CleanUtil {
 
   static {
     ByteBuffer buf = null;
+    Field cleanerField = null;
+    Method cleanMethod = null;
     try {
       buf = ByteBuffer.allocateDirect(1);
-      CLEANER_FIELD = buf.getClass().getDeclaredField("cleaner");
-      CLEANER_FIELD.setAccessible(true);
-      Object cleaner = CLEANER_FIELD.get(buf);
-      CLEAN_METHOD = cleaner.getClass().getDeclaredMethod("clean");
-    } catch (NoSuchFieldException | NoSuchMethodException | IllegalAccessException e) {
+      cleanerField = buf.getClass().getDeclaredField("cleaner");
+      cleanerField.setAccessible(true);
+      Object cleaner = cleanerField.get(buf);
+      cleanMethod = cleaner.getClass().getDeclaredMethod("clean");
+    } catch (Throwable e) {
       clean(buf);
-      throw new IllegalStateException("No available cleaner found.", e);
+      logger.warn("Initialization failed for cleanerField or cleanMethod", e);
     }
+    CLEANER_FIELD = cleanerField;
+    CLEAN_METHOD = cleanMethod;
   }
 
   public static void clean(ByteBuffer buffer) {
     try {
       Object cleaner = CLEANER_FIELD.get(buffer);
       CLEAN_METHOD.invoke(cleaner);
-    } catch (IllegalAccessException | InvocationTargetException e) {
+    } catch (Throwable e) {
       // Ignore clean failure
       logger.warn("Clean failed for buffer " + buffer.getClass().getSimpleName(), e);
     }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
@@ -46,9 +46,10 @@ public class CleanUtil {
       cleanerField.setAccessible(true);
       Object cleaner = cleanerField.get(buf);
       cleanMethod = cleaner.getClass().getDeclaredMethod("clean");
-    } catch (Throwable e) {
-      clean(buf);
+    } catch (NoSuchFieldException | NoSuchMethodException | IllegalAccessException e) {
       logger.warn("Initialization failed for cleanerField or cleanMethod", e);
+    } finally {
+      clean(buf);
     }
     CLEANER_FIELD = cleanerField;
     CLEAN_METHOD = cleanMethod;
@@ -58,7 +59,7 @@ public class CleanUtil {
     try {
       Object cleaner = CLEANER_FIELD.get(buffer);
       CLEAN_METHOD.invoke(cleaner);
-    } catch (Throwable e) {
+    } catch (IllegalAccessException | InvocationTargetException | NullPointerException e) {
       // Ignore clean failure
       logger.warn("Clean failed for buffer " + buffer.getClass().getSimpleName(), e);
     }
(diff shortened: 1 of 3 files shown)


liupc:
@gszadovszky  Thanks for review, I will update.

liupc:
@gszadovszky I have updated this PR as your suggestion, and passed the tests. can you review again?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CleanUtil {
```
gszadovszky: It would be nice to have some comments about the purpose of this class and why we are using the relfections here to do the clean (compatibility with both java8 and java11...)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,57 @@
...
+    ByteBuffer buf = null;
+    try {
+      buf = ByteBuffer.allocateDirect(1);
+      CLEANER_FIELD = buf.getClass().getDeclaredField("cleaner");
+      CLEANER_FIELD.setAccessible(true);
+      Object cleaner = CLEANER_FIELD.get(buf);
+      CLEAN_METHOD = cleaner.getClass().getDeclaredMethod("clean");
+    } catch (NoSuchFieldException | NoSuchMethodException | IllegalAccessException e) {
+      clean(buf);
+      throw new IllegalStateException("No available cleaner found.", e);
```
gszadovszky: This Cleaner class is not crucial for the parquet-mr library while with this throwing exception it will cause serious failures. I would suggest logging this exception instead of logging it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,57 @@
...
+  static {
+    ByteBuffer buf = null;
+    try {
+      buf = ByteBuffer.allocateDirect(1);
+      CLEANER_FIELD = buf.getClass().getDeclaredField("cleaner");
+      CLEANER_FIELD.setAccessible(true);
+      Object cleaner = CLEANER_FIELD.get(buf);
+      CLEAN_METHOD = cleaner.getClass().getDeclaredMethod("clean");
+    } catch (NoSuchFieldException | NoSuchMethodException | IllegalAccessException e) {
+      clean(buf);
```
gszadovszky: Cannot see the purpose of invoking clean in case of the field or the method was not properly initialized.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,57 @@
...
+      throw new IllegalStateException("No available cleaner found.", e);
+    }
+  }
+
+  public static void clean(ByteBuffer buffer) {
+    try {
+      Object cleaner = CLEANER_FIELD.get(buffer);
+      CLEAN_METHOD.invoke(cleaner);
+    } catch (IllegalAccessException | InvocationTargetException e) {
```
gszadovszky: The cases of the uninitialized (null) field and/or method should also be handled somehow.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: It's ok, I will update.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: Good review, yes, it's not crucial.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: Yes, you are right, here the `clean` may not do real work for `CLEAN_METHOD` and `CLEAN_FIELD` is not available, but I think keep the behavior consistent is necessary.

> buf = ByteBuffer.allocateDirect(1);  // allocated memory, then we should release it when exception happens.

And also it can print some message about the memory leak, though, it's non big deal.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: I will fix it, Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
gszadovszky: If you only want to be consistent then you should clean the buffer in finally instead in a catch block. Currently, it only cleans the buffer if an error occurs but doesn't if everything goes normally.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+    CLEANER_FIELD = cleanerField;
+    CLEAN_METHOD = cleanMethod;
+  }
+
+  public static void clean(ByteBuffer buffer) {
+    try {
+      Object cleaner = CLEANER_FIELD.get(buffer);
+      CLEAN_METHOD.invoke(cleaner);
+    } catch (Throwable e) {
```
gszadovszky: I think, it would be better to handle the case of uninitialized field or method separately from the case of an exception happens during the invocation. If the initialization fails then we already have a logged stack trace about the problem, no need several additional stack traces about NPE. In the other hand, if we have some other unknown issue it would be nice to have some debug info. 
(See above about catching Throwable.)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    ByteBuffer buf = null;
+    Field cleanerField = null;
+    Method cleanMethod = null;
+    try {
+      buf = ByteBuffer.allocateDirect(1);
+      cleanerField = buf.getClass().getDeclaredField("cleaner");
+      cleanerField.setAccessible(true);
+      Object cleaner = cleanerField.get(buf);
+      cleanMethod = cleaner.getClass().getDeclaredMethod("clean");
+    } catch (Throwable e) {
```
gszadovszky: Catching Throwable is not nice. I think, it would be better to catch only the known exceptions and let all the others (some serious JVM errors like OOM) throw out.

liupc: It's ok to use all known exceptions here, but I think it really don't matter as I don't think OOM-like JVM errors can happen here, and I believe there might be many places throwing `Throwable` in this project, if we just need to handle NonFatal exceptions, there should be a NonFatal(Customed exception) exception rather than list all potential exceptions.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: @gerashegalov fixed! Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
liupc: Updated, Thanks!

liupc:
Thanks for merging! @gszadovszky

PR #587, opened 2019-01-09
PARQUET-1328: Add Bloom filter reader and writer

cjjnjust:
the original pull request is base on master. This one is created against bloom-filter branch.

### DIFF
--- a/parquet-column/pom.xml
+++ b/parquet-column/pom.xml
@@ -119,12 +119,17 @@
               <artifactSet>
                 <includes>
                   <include>it.unimi.dsi:fastutil</include>
+                  <include>com.google.guava:guava</include>
                 </includes>
               </artifactSet>
               <relocations>
                 <relocation>
                   <pattern>it.unimi.dsi</pattern>
                   <shadedPattern>org.apache.parquet.it.unimi.dsi</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>com.google.common</pattern>
+                  <shadedPattern>org.apache.parquet.com.google.common</shadedPattern>
                 </relocation>
               </relocations>
             </configuration>
--- a/parquet-column/pom.xml
+++ b/parquet-column/pom.xml
@@ -93,7 +93,6 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
-      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -35,6 +35,9 @@ import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWrite
 import org.apache.parquet.column.values.factory.ValuesWriterFactory;
 import org.apache.parquet.schema.MessageType;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * This class represents all the configurable Parquet properties.
  */
@@ -49,6 +52,7 @@ public class ParquetProperties {
   public static final int DEFAULT_MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
   public static final int DEFAULT_COLUMN_INDEX_TRUNCATE_LENGTH = 64;
   public static final int DEFAULT_PAGE_ROW_COUNT_LIMIT = 20_000;
+  public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
 
   public static final ValuesWriterFactory DEFAULT_VALUES_WRITER_FACTORY = new DefaultValuesWriterFactory();
 
@@ -86,11 +90,16 @@ public class ParquetProperties {
   private final ByteBufferAllocator allocator;
   private final ValuesWriterFactory valuesWriterFactory;
   private final int columnIndexTruncateLength;
+  private final boolean enableBloomFilter;
+
+  // The key-value pair represents the column name and its expected distinct number of values in a row group.
+  private final Map<String, Long> bloomFilterExpectedDistinctNumbers;
   private final int pageRowCountLimit;
 
   private ParquetProperties(WriterVersion writerVersion, int pageSize, int dictPageSize, boolean enableDict, int minRowCountForPageSizeCheck,
                             int maxRowCountForPageSizeCheck, boolean estimateNextSizeCheck, ByteBufferAllocator allocator,
-                            ValuesWriterFactory writerFactory, int columnIndexMinMaxTruncateLength, int pageRowCountLimit) {
+                            ValuesWriterFactory writerFactory, int columnIndexMinMaxTruncateLength, int pageRowCountLimit,
+                            boolean enableBloomFilter, Map<String, Long> bloomFilterExpectedDistinctNumber) {
     this.pageSizeThreshold = pageSize;
     this.initialSlabSize = CapacityByteArrayOutputStream
       .initialSlabSizeHeuristic(MIN_SLAB_SIZE, pageSizeThreshold, 10);
@@ -104,6 +113,8 @@ public class ParquetProperties {
 
     this.valuesWriterFactory = writerFactory;
     this.columnIndexTruncateLength = columnIndexMinMaxTruncateLength;
+    this.enableBloomFilter = enableBloomFilter;
+    this.bloomFilterExpectedDistinctNumbers = bloomFilterExpectedDistinctNumber;
     this.pageRowCountLimit = pageRowCountLimit;
   }
 
@@ -201,6 +212,14 @@ public class ParquetProperties {
     return pageRowCountLimit;
   }
 
+  public boolean isBloomFilterEnabled() {
+    return enableBloomFilter;
+  }
+
+  public Map<String, Long> getBloomFilterColumnExpectedNDVs() {
+    return bloomFilterExpectedDistinctNumbers;
+  }
+
   public static Builder builder() {
     return new Builder();
   }
@@ -220,6 +239,8 @@ public class ParquetProperties {
     private ByteBufferAllocator allocator = new HeapByteBufferAllocator();
     private ValuesWriterFactory valuesWriterFactory = DEFAULT_VALUES_WRITER_FACTORY;
     private int columnIndexTruncateLength = DEFAULT_COLUMN_INDEX_TRUNCATE_LENGTH;
+    private boolean enableBloomFilter = DEFAULT_BLOOM_FILTER_ENABLED;
+    private Map<String, Long> bloomFilterColumnExpectedNDVs = new HashMap<>();
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
 
     private Builder() {
@@ -236,6 +257,8 @@ public class ParquetProperties {
       this.valuesWriterFactory = toCopy.valuesWriterFactory;
       this.allocator = toCopy.allocator;
       this.pageRowCountLimit = toCopy.pageRowCountLimit;
+      this.enableBloomFilter = toCopy.enableBloomFilter;
+      this.bloomFilterColumnExpectedNDVs = toCopy.bloomFilterExpectedDistinctNumbers;
     }
 
     /**
@@ -324,6 +347,27 @@ public class ParquetProperties {
       return this;
     }
 
+    /**
+     * Set to enable Bloom filter.
+     *
+     * @param enableBloomFilter a boolean to indicate whether to enable Bloom filter.
+     * @return this builder for method chaining.
+     */
+    public Builder withBloomFilterEnabled(boolean enableBloomFilter) {
+      this.enableBloomFilter = enableBloomFilter;
+      return this;
+    }
+    /**
+     * Set Bloom filter info for columns.
+     *
+     * @param columnExpectedNDVs the columns expected number of distinct values in a row group
+     * @return this builder for method chaining
+     */
+    public Builder withBloomFilterInfo(HashMap<String, Long> columnExpectedNDVs) {
+      this.bloomFilterColumnExpectedNDVs = columnExpectedNDVs;
+      return this;
+    }
+
     public Builder withPageRowCountLimit(int rowCount) {
       Preconditions.checkArgument(rowCount > 0, "Invalid row count limit for pages: " + rowCount);
       pageRowCountLimit = rowCount;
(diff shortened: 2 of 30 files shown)


cjjnjust:
Hi @rdblue, @majetideepak , do you have time to have a look on this as well?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -0,0 +1,294 @@
...
+  public static final double DEFAULT_FPP = 0.01;
+
+  // Hash strategy used in this Bloom filter.
+  public final HashStrategy hashStrategy;
+
+  // The underlying byte array for Bloom filter bitset.
+  private byte[] bitset;
+
+  // A integer array buffer of underlying bitset to help setting bits.
+  private IntBuffer intBuffer;
```
cjjnjust: Jim's comments: [https://github.com/apache/parquet-mr/pull/521#discussion_r228761288](url)

Just get some time for this. JMeter seems not suit for showing the memory layout for java heap. I wrote a example application and use jmap to dump the heap.   The code snippet is like: 
```java
        byte [] bitset = new byte[1024*1024];
        IntBuffer intBuffer = ByteBuffer.wrap(bitset).asIntBuffer();
        ByteBuffer byteBuffer = ByteBuffer.allocate(1024*1024);
        byte[] bitset2 = new byte[1024*1024];
```

The Eden space in heap dumps after every sentence are: 

Eden Space:                                                                                                                                                                                                 
   capacity = 66060288 (63.0MB)                                                                                                                                                                             
   used     = 3691032 (3.5200424194335938MB)                                                                                                                                                                
   free     = 62369256 (59.479957580566406MB)                                                                                                                                                               
   5.587368919735863% used    

Eden Space:
   capacity = 66060288 (63.0MB)
   used     = 3691032 (3.5200424194335938MB)
   free     = 62369256 (59.479957580566406MB)
   5.587368919735863% used

Eden Space:
   capacity = 66060288 (63.0MB)
   used     = 4739624 (4.520057678222656MB)
   free     = 61320664 (58.479942321777344MB)
   7.17469472733755% used

Eden Space:
   capacity = 66060288 (63.0MB)
   used     = 5788216 (5.520072937011719MB)
   free     = 60272072 (57.47992706298828MB)
   8.762020534939236% used

The Eden space does not increase when we call the wrap API of ByteBuffer.

cjjnjust:
@jbapple-cloudera,  I keep last comment that had not address last time in here. Other comments should all be addressed, it can also be found [https://github.com/apache/parquet-mr/pull/521. ](url)

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -256,6 +276,28 @@ public Builder withDictionaryPageSize(int dictionaryPageSize) {
...
+      return this;
+    }
+
+    /**
+     * Set Bloom filter info for columns.
+     *
+     * @param columnExpectedNDVs the columns expected number of distinct values in a row group
+     * @return this builder for method chaining
+     */
+    public Builder withBloomFilterInfo(HashMap<String, Long> columnExpectedNDVs) {
```
gszadovszky: So, the user is able to specify the NDV for the columns which he would like to write bloom filters for. If this map is `null` or empty then no bloom filters should be written otherwise bloom filters are written only for the columns which are NDV is specified for. 
Why do we need the additional flag of enabling bloom filters?
Is there any other way to build a bloom filter without a specified NDV? Usually, parquet-mr does not ask for such values about the data characteristics but measures them. I don't know how would the clients present this data and therefore if they willing to use this API.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -199,6 +215,8 @@ public static Builder copy(ParquetProperties toCopy) {
     private int pageSize = DEFAULT_PAGE_SIZE;
     private int dictPageSize = DEFAULT_DICTIONARY_PAGE_SIZE;
     private boolean enableDict = DEFAULT_IS_DICTIONARY_ENABLED;
+    private boolean enableBloomFilter = DEFAULT_BLOOM_FILTER_ENABLED;
+    private HashMap<String, Long> bloomFilterColumnExpectedNDV = new HashMap<>();
```
gszadovszky: We usually don't use `HashMap` as a reference type but `Map` so the implementation is hidden. Is it intentional that you `HashMap` as reference type and even return this reference in a public method?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -194,6 +198,10 @@ public long getMemSize() {
 
     public void writeToFileWriter(ParquetFileWriter writer) throws IOException {
       writer.startColumn(path, totalValueCount, compressor.getCodecName());
+      if (bloomFilter != null) {
```
gszadovszky: So, we write the bloom filters into the row-group. It means we have to seek to the row-group and read from them even if the whole row-group could be dropped based on the bloom filter. Because of the similar issue we decided to the put the column indexes after all row-groups (just before the footer). I don't know if it would be feasible/beneficial for bloom filters as well.

.travis.yml
```diff
@@ -29,6 +29,10 @@ before_install:
   - sudo make install
   - cd ..
   - date
+  - git clone https://github.com/apache/parquet-format.git
```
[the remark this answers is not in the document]
cjjnjust: Since the travis update is on master, I can cherry-pick that commit and put this "hack" to branch specific script.

parquet-column/pom.xml
```diff
@@ -93,7 +93,6 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
-      <scope>test</scope>
```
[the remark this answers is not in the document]
cjjnjust: Agreed, I will update PR to use shading guava.

.travis.yml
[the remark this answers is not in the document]
gszadovszky: Please, also keep in mind that this branch is 41 commits behind master currently. 1 of this 41 is the column indexes feature which was quite huge. I would suggest getting these changes to this feature branch as soon as possible or at least check how much work does the merge would require.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
cjjnjust: You are right,  if user doesn't specific any NDV, no bloom filter will be build according to current logic. The additional flag is a redundant flag.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
cjjnjust: According to design, we put bloom filter at beginning of row group.  Not sure how much overhead if we seek at zero and read them. But I can try to put it near to foot to avoid such overhead.

cjjnjust:
Thanks @gszadovszky . These comments are very useful, I will rebase PR to master update it according to your comments ASAP.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
cjjnjust: Actually, I'm not sure what is the overhead if we read some index data from row group. Does seek read extra data?

gszadovszky: If you use HDFS as the storage layer and configures parquet properly (aligning the row-groups to HDFS blocks) then you will have different row-groups in different blocks which are usually placed in different hosts. So, seeking in the file might require remote reads which consuming resources on the remote host and the network bandwidth as well. Usually, reading the footer is also a remote read but you have to read the footer anyway. 
See the [column-index design](https://github.com/apache/parquet-format/blob/master/PageIndex.md#technical-approach) for more details.

I would also like to highlight that I am not suggesting we shall go the same direction as we did for the column-indexes. I would only like to encourage you to discuss this topic with the community and include it in the design. It might also require some benchmarking as well.

gszadovszky:
The build has failed because you are not allowed to use SNAPSHOT dependencies on the master branch. You will not be able to merge this change to master before having the required changes in a parquet-format release and depending on that new release.

cjjnjust:
Thanks very much. Please help to merge master to feature branch. 

So the last commit passed build since it is target feature branch, Now I understood.

gszadovszky:
The branch {{bloom-filter}} is now at the current {{master}}. You may re-target this PR to {{bloom-filter}}.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
ryanrupp: I'm not all that familiar with the internals of bloom filters but was hoping this could work similar to dictionary fallback where Parquet automatically uses a dictionary provided the size of distinct values does not grow beyond the specified max dictionary size e.g. dictionaries would be preferred but would fallback to bloom filters for higher cardinality fields. Then again, it doesn't make sense to probably try to do this for all columns automatically (so opt-in) although it'd be nice if there were some heuristics that could decide that automatically. It also looks like that may not be possible though because the # of distinct values needs to be known ahead of time to size this correctly? If that number is incorrect it looks like it increases the likelihood of false positives and if it's too large it takes up more space than needed?

I looked briefly into how this works for ORC where the bloom filter columns are opt-in specified via column name (without providing unique counts). But, the reason they're able to omit the # of distinct values is because the bloom filters are created at the `stride` level where ORC controls the stride size by number of rows so they always just size it to the stride row size (default of 10k). So, they assume worst case scenario that all values are unique within the stride (so the filter ends up being a larger size than it needs to be maybe but without the false positive probability increasing).

I'm not sure if it will be problematic here to require clients to specify the number of distinct values upfront but I can't really think of a better solution at the moment (short of something painful/inefficient like Parquet doing two passes of the data).

cjjnjust: Thanks @ryanrupp , I know the FPP may not close to expected value since user may not estimate NDV correctly before row group size is determined.  I used to think cache hashed value and build bloom filter when writing column chunk, that way can keep FPP as we expected and can take advantage some vector computation while it may lead to OOM in case of very large NDV.  Another option should be build bloom filter on page level so the behavior can be same as ORC side but that may make it more complicated.

Specify NDV looks like a bit difficult to user, so I 'm thinking to add another property named max size of bloom filter of column, so that users can specify how much space they would like to use as bloom filter index for a column. It looks simple and direct to users. What do you think?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
cjjnjust: Thanks for the nice explanation. As you point out, it would better to execute some benchmark firstly and let data talk. Will show benchmark data later.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
gszadovszky: I agree with @ryanrupp, the current behavior that the user have to specify the NDVs is somewhat against the usual way how _parquet-mr_ works. If it is hard to guess such values the clients may not use this feature and the whole development is useless. 
Also, I think such discussions (about how the library would guess the parameters of a bloom filter or which columns should we build one) would more fit in a design doc.
However, the current development is on the feature branch so, for the sake of incremental development, if it is easier for you to start the implementation and/or execute benchmarks, I can accept the current way for now.

cjjnjust: Thanks @gszadovszky @ryanrupp , I opened a [discussion](https://lists.apache.org/thread.html/f783f4a6e4e98f839285e78d0fa27db05628348e28a6608791fba740@%3Cdev.parquet.apache.org%3E) on dev mail list, welcome to put your opinion.  I will update design doc and implementation if we have some agreement.

.travis.yml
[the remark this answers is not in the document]
gszadovszky: The current PR seems fine only that the travis build was not re-triggered after you re-targeted to `bloom-filter`. If you send an additional commit (even a whitespace or something) it will trigger travis and we'll see if everything is fine.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -210,6 +214,29 @@ public static boolean getEnableDictionary(JobContext jobContext) {
...
+    if (columnNamesConf == null || expectedNDVsConf == null) {
+      return kv;
+    }
+
+    String[] columnNames = columnNamesConf.split(",");
+    String[] expectedNDVs = expectedNDVsConf.split(",");
+
+    if (columnNames.length == expectedNDVs.length) {
+      for (int i = 0; i < columnNames.length; i++) {
+        kv.put(columnNames[i], Long.getLong(expectedNDVs[i]));
```
garawalid: @chenjunjiedada 
I have a small question about `BLOOM_FILTER_EXPECTED_NDV`.  
Why do we get the **system property** after setting `parquet.bloom.filter.expected.ndv`? Wouldn't it be better if we just parse the string with `Long.parseLong()`?

chenjunjiedada: @garawalid,  Usually the compute engine, such as hive, spark, etc., set the properties for parquet so we may not able to get the string directly other than from configuration.

garawalid: @chenjunjiedada Thanks for the clarification.
In fact, when I pass `parquet.bloom.filter.column.names` and  `parquet.bloom.filter.expected.ndv` this way :  
```java
    // Configuration conf = new Configuration();
    conf.set("parquet.bloom.filter.column.names", "content,line");
    conf.set("parquet.bloom.filter.expected.ndv","1000,200");
```

I got `kv` as follows `"line" -> null` and `"content" -> null`. 

You can reproduce this behavior by adding that configuration to `testReadWrite()` in `TestInputOutputFormat.java`. 

Do you think this is normal? In my case, I changed `Long.getLong(expectedNDVs[i])` by `Long.parseLong(expectedNDVs[i])` to build the `kv` HashMap.

chenjunjiedada:
@garawalid , I can get configuration when putting settings in `runMapReduceJob`. Where did you put the settings?

garawalid:
@chenjunjiedada 
Strange! I add the configuration here: https://github.com/apache/parquet-mr/blob/1fc273329cd6d54085de81ecccfe99be181f24be/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/example/TestInputOutputFormat.java#L234-L240

Configuration:
```java
    conf.set("parquet.bloom.filter.column.names", "content,line");
    conf.set("parquet.bloom.filter.expected.ndv","1000,200");
```

garawalid:
@chenjunjiedada 
I noticed that `expectedNDVs[i]` gets the string value as expected (which is 1000 in our case). The problem comes when we cast it to Long (via `Long.getLong()`) which leads to a `null`.

PR #590, opened 2019-01-09
PARQUET-1490: Add branch-specific Travis steps

zivanfi:
The script for the main branch makes sure that POM files in the master
branch do not refer to SNAPSHOT versions.

The possiblity of scripts for feature branches will allow building a
SNAPSHOT version of parquet-format and depending on it in the POM files.


dev/travis-before-install.sh
```diff
@@ -0,0 +1,34 @@
...
+pwd
+wget -nv http://archive.apache.org/dist/thrift/0.9.3/thrift-0.9.3.tar.gz
+tar zxf thrift-0.9.3.tar.gz
+cd thrift-0.9.3
+chmod +x ./configure
+./configure --disable-gen-erl --disable-gen-hs --without-ruby --without-haskell --without-erlang --without-php --without-nodejs
+sudo make install
+cd ..
+branch_specific_script="dev/travis-before-install-${TRAVIS_BRANCH}.sh"
+if [[ -e "$branch_specific_script" ]]
```
rdblue: Should this go before the protobuf and thrift install since it doesn't require those steps?

zivanfi: I also considered that option, but feature branches may need those, depending on what they want to do (for example, build parquet-format). Since in general I expect this check to fail very seldom, I think generally it is not a waste of resources to build protobuf and thrift first.

rdblue: Sounds fine to me, although we can always change it later when we have more checks.

zivanfi: We already know that the bloom-filter branch wants to build its own parquet-format (https://github.com/apache/parquet-mr/pull/587/files). I would like their .travis.yml modification to be moved to the branch-specific script instead, and for that we need to invoke the script after building thrift.

PR #594, opened 2019-01-19
Add an option to convert Parquet Int96 to Arrow Timestamp

yongyanw:
Just as mentioned in original code, Parquet Int96 is probably timestamp, we should provide an option to convert Parquet Int96 to Arrow Timestamp


parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -492,8 +500,11 @@ private String getTimeZone(LogicalTypeAnnotation.TimestampLogicalTypeAnnotation
 
       @Override
       public TypeMapping convertINT96(PrimitiveTypeName primitiveTypeName) throws RuntimeException {
-        // Possibly timestamp
-        return field(new ArrowType.Binary());
+        if (convertInt96ToArrowTimestamp) {
+          return field(new ArrowType.Timestamp(TimeUnit.MILLISECOND, null));
```
xhochy: INT96 is a nanosecond timestamp, so this should be `TimeUnit.NANOSECOND`.

parquet-arrow/src/test/java/org/apache/parquet/arrow/schema/TestSchemaConverter.java
```diff
@@ -419,6 +420,27 @@ public void testParquetInt64TimeMicrosToArrow() {
...
+    Assert.assertEquals(expected, converter.fromParquet(parquet).getArrowSchema());
+  }
+
+  @Test
+  public void testParquetInt96ToArrowTimestamp() {
+    final SchemaConverter converterInt96ToTimestamp = new SchemaConverter(true);
+    MessageType parquet = Types.buildMessage()
+      .addField(Types.optional(INT96).named("a")).named("root");
+    Schema expected = new Schema(asList(
+      field("a", new ArrowType.Timestamp(TimeUnit.MILLISECOND, null))
```
xhochy: As above: `TimeUnit.NANOSECOND`

yongyanw:
Hi @xhochy @nandorKollar @zivanfi,

I did the rebase on master, created PARQUET-1504 and updated the commit message. Could you help me to review my change and merge it if it is good for you? My current work needs this change. Thanks!

parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -86,10 +86,18 @@
...
+
   /**
    * For when we'll need this to be configurable
    */
   public SchemaConverter() {
+    this(false);
+  }
+
+  public SchemaConverter(final boolean convertInt96ToArrowTimestamp) {
+    this.convertInt96ToArrowTimestamp = convertInt96ToArrowTimestamp;
```
qinghui-xu: Will it be possible that in the future we need to convert int96 to some other type in arrow (or for any other type convertion)? If that's something likely to happen, perhaps it would be better to pass the parameter in a configuration object (such as `org.apache.hadoop.conf.Configuration`)

xhochy: We need to pass this later on as an option from the outside. The current routines don't know anything about the file creator. @yongyanw can you add this as a TODO comment and open a JIRA for it? I don't expect that we can solve this in this PR.

yongyanw: PARQUET-1511 was created and TODO comment was added.

PR #596, opened 2019-01-21
PARQUET-1499: Add Java 11 to Travis

Fokko:
https://issues.apache.org/jira/browse/PARQUET-1497


.travis.yml
```diff
@@ -2,6 +2,14 @@ language: java
 before_install:
   - bash dev/travis-before_install.sh
 
+jdk:
+  - openjdk8
+  - openjdk11
+
+matrix:
+  allow_failures:
```
gszadovszky: In my opinion, it does not make sense to allow the failures for the tool we are using to catch failures. I would suggest removing this part and hold this PR until the related issues are fixed.

dev/Dockerfile
```diff
@@ -0,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
Fokko: No, first I wanted to isolate the tests inside the Dockerfile, but Travis also gives us the possibility to run under different JDK's. I'll remove the Dockerfile since there aren't a lot of dependencies for Parquet.

.travis.yml
Fokko: Makes sense @gszadovszky

Fokko: I can rebase often to check when the JDK11 becomes green.

Fokko:
Thrift was using Gradle 4.4.1, but has been updated to 5.1.1 which supports Java11:
https://github.com/apache/thrift/commit/98f379ece17c8acb6a91cb3fd5b16acc3f0d1698#diff-7762a898c27a3d29c23245cc2abd99baR5

Fokko:
I've split some of the stuff in this PR into https://github.com/apache/parquet-mr/pull/656 and https://github.com/apache/parquet-mr/pull/654 to make the PR's nice and manageable. For this PR, we're still dependent on https://github.com/apache/parquet-mr/pull/605. After updating Scala, I expect the CI to pass.

Fokko:
Rebased onto master and it is green now :-)

PR #597, opened 2019-01-21
[PARQUET-1500] Replace Closeables with try-with-resources

Fokko:
https://jira.apache.org/jira/browse/PARQUET-1500

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
```diff
@@ -53,22 +53,14 @@ private SerializationUtil() { }
...
-    try {
-      baos = new ByteArrayOutputStream();
-      gos = new GZIPOutputStream(baos);
-      oos = new ObjectOutputStream(gos);
-      oos.writeObject(obj);
-    } finally {
-      Closeables.close(oos);
-      Closeables.close(gos);
-      Closeables.close(baos);
+    try(ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
ryanrupp: You can consolidate these into a single `try` e.g.:

```
try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
       GZIPOutputStream gos = new GZIPOutputStream(baos);
       ObjectOutputStream oos = new ObjectOutputStream(gos)) {
    oos.writeObject(obj);
    conf.set(key, new String(Base64.encodeBase64(baos.toByteArray()), "UTF-8"));
}
```

Fokko: Sweet suggestion, thanks!

nandorKollar:
@Fokko looks like semantic versioning plugin didn't like the removal of Closeables. How about deprecating it instead of removing so that this addition could get committed to 1.x line?

Fokko:
Thanks for the clarification @nandorKollar. The CI error was a bit cryptic to me, but now it makes sense. I've added a deprecation warning instead of removing the class.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/SerializationUtil.java
```diff
@@ -53,22 +53,13 @@ private SerializationUtil() { }
...
+      try(GZIPOutputStream gos = new GZIPOutputStream(baos);
+            ObjectOutputStream oos = new ObjectOutputStream(gos)) {
+        oos.writeObject(obj);
+      }
+      conf.set(key, new String(Base64.encodeBase64(baos.toByteArray()), StandardCharsets.UTF_8));
```
gszadovszky: This code is not equivalent to the original one: `baos.toByteArray()` was invoked after closing the stream, now it is the other way around.

Fokko: This is true, but with the try-with-resource, we don't have a reference after closing the stream. Please note that the `.close()` on the `ByteArrayOutputStream` doesn't have any effect: https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#close()
The important thing here is that the `ObjectOutputStream` and `GZIPOutputStream` are flushed (and closed).

gszadovszky: You're right, thanks for explaining.

PR #608, opened 2019-01-28
PARQUET-1516: Store Bloom filters near to footer

cjjnjust:
Store the bloom filter data near to footer, so that client does not have to read them if it is not doing selective query.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -559,7 +559,7 @@ public class ParquetFileWriter {
    * Write a Bloom filter
    * @param bloomFilter the bloom filter of column values
    */
-  protected void writeBloomFilter(BloomFilter bloomFilter)  {
+  void writeBloomFilter(BloomFilter bloomFilter)  {
     currentBloomFilters.add(bloomFilter);
   }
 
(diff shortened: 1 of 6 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -558,6 +555,14 @@ private void innerWriteDataPage(
     currentEncodings.add(valuesEncoding);
   }
 
+  /**
+   * Write a Bloom filter
+   * @param bloomFilter the bloom filter of column values
+   */
+  public void writeBloomFilter(BloomFilter bloomFilter)  {
```
gszadovszky: Currently, only `TestParquetFileWriter` uses this method. If it is not really required to be `public` I would suggest having it package protected.

cjjnjust: Thanks, Done.

gszadovszky: I've ment _package protected_ (aka _package private_) which is marked with using no modifiers. 
```suggestion
  void writeBloomFilter(BloomFilter bloomFilter)  {
```

cjjnjust: Oops, understood now.

PR #613, opened 2019-02-11
PARQUET-1228: Format Structures encryption

ggershinsky:


gszadovszky:
To have a passing travis on your feature branch you may use the dev script feature introduced in PARQUET-1490. See https://github.com/apache/parquet-mr/blob/bloom-filter/dev/travis-before_install-bloom-filter.sh as an example.

ggershinsky:
@gszadovszky Thanks, but this wont help, because the format pull request is not merged yet in that feature branch,
https://github.com/apache/parquet-format/pull/124
Would appreciate if you could check if this can be accelerated..

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -226,11 +347,24 @@ private static InterningProtocol protocol(TIOStreamTransport t) {
...
+    if (null == encryptor) { 
+      try {
+        tbase.write(protocol(to));
+        return;
+      } catch (TException e) {
+        throw new IOException("can not write " + tbase, e);
+      }
+    }
+    // Serialize and encrypt the structure
+    TMemoryBuffer thriftMemoryBuffer = new TMemoryBuffer(100);
```
Fokko: This one implements `Closeable`, can we wrap a try-with-resources around this? And why 100 buffer size? Maybe move this to a constant, instead of having a magic variable.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -216,8 +327,18 @@ private static TProtocol protocol(InputStream from) {
   private static InterningProtocol protocol(TIOStreamTransport t) {
     return new InterningProtocol(new TCompactProtocol(t));
   }
+  
 
-  private static <T extends TBase<?,?>> T read(InputStream from, T tbase) throws IOException {
+  private static <T extends TBase<?,?>> T read(InputStream input, T tbase, BlockCipher.Decryptor decryptor, byte[] AAD) throws IOException {
+    InputStream from;
```
Fokko: I like to make variables like these `final`

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -198,8 +301,16 @@ public void consume(RowGroup rowGroup) {
           }
         })));
       }
+      
+      InputStream from;
```
Fokko: I like to make variables like these `final`

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -155,13 +226,33 @@ public void addRowGroup(RowGroup rowGroup) {
...
   }
-
+  
+  public static void readFileMetaData(InputStream from, FileMetaDataConsumer consumer, 
+      BlockCipher.Decryptor decryptor, byte[] AAD) throws IOException {
+    readFileMetaData(from, consumer, false, decryptor, AAD);
+  }
+  
   public static void readFileMetaData(InputStream from, final FileMetaDataConsumer consumer, boolean skipRowGroups) throws IOException {
+    readFileMetaData(from, consumer, skipRowGroups, (BlockCipher.Decryptor) null, (byte[]) null);
```
Fokko: I would omit the casting since it doesn't add any value.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -155,13 +226,33 @@ public void addRowGroup(RowGroup rowGroup) {
...
+
+    @Override
+    public void setFooterSigningKeyMetadata(byte[] footerSigningKeyMetadata) {
+      md.setFooter_signing_key_metadata(footerSigningKeyMetadata);
+    }
   }
 
   public static void readFileMetaData(InputStream from, FileMetaDataConsumer consumer) throws IOException {
-    readFileMetaData(from, consumer, false);
+    readFileMetaData(from, consumer, (BlockCipher.Decryptor) null, (byte[]) null);
```
Fokko: I would omit the casting since it doesn't add any value.

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,70 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.format;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public interface BlockCipher{
```
Fokko: I would split the `Encryptor` and `Decryptor` in separate files and put them in the `BlockCipher` package. WDYT?

ggershinsky:
Hi Fokko, thanks for the review, much appreciated. Finally, I got some respite from the C++ PRs, now getting to the Java ones.
Regarding autoformat - can do, but please suggest a (Linux cli?) tool for that.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: try-with-resources sounds good.
100 is the initial alloc. Usually enough for page headers etc, but will auto-grow for larger ones.
Will make it a constant.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: The value is in compiler detection of buggy calls to wrong methods (wrong signature), causes by typeless nulls.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: as above.

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
ggershinsky: I'm reluctant to do that. Extra package, with two small files; extra file..

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: caused

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -55,36 +59,88 @@
 public class Util {
 
   public static void writeColumnIndex(ColumnIndex columnIndex, OutputStream to) throws IOException {
-    write(columnIndex, to);
+    writeColumnIndex(columnIndex, to, (BlockCipher.Encryptor) null, (byte[]) null);
```
chenjunjiedada: I think you don't need to enforce casting here. It will be done by implicit casting.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -198,8 +301,16 @@ public void consume(RowGroup rowGroup) {
           }
         })));
       }
+      
+      InputStream from;
+      if (null == decryptor) {
+        from = input;
+      }
+      else {
```
chenjunjiedada: nit: please move this line to the end of last line.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
chenjunjiedada: Could you please paste what kind of compiling error you met?  Maybe we can set an option in the compiler?

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -226,11 +347,24 @@ private static InterningProtocol protocol(TIOStreamTransport t) {
...
-      tbase.write(protocol(to));
+      tbase.write(new InterningProtocol(new TCompactProtocol(thriftMemoryBuffer)));
     } catch (TException e) {
       throw new IOException("can not write " + tbase, e);
     }
+    byte[] encryptedBuffer = encryptor.encrypt(thriftMemoryBuffer.getArray(), AAD);
+    to.write(encryptedBuffer);
   }
 }
+
```
nandorKollar: nit: remove empty line here

dev/travis-before_install-encryption.sh
```diff
@@ -0,0 +1,27 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
nandorKollar: I'm afraid that in order for this to work, we'll need https://github.com/apache/parquet-mr/commit/3d8eae78291ace9d9a12839a932c52a785b0b308 too. @ggershinsky should we rebase this branch onto master?

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,70 @@
...
+  
+
+  public interface Encryptor{
+    /**
+     * Encrypts the plaintext.
+     * 
+     * @param plaintext The plaintext starts at offset 0 of the input value, and fills up the entire byte array.
+     * @param AAD Encryption AAD (ignored in case of CTR cipher)
```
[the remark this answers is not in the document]
ggershinsky: sounds good

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,70 @@
...
+     * @return lengthAndCiphertext The first 4 bytes of the returned value are the ciphertext length (little endian int). 
+     * The ciphertext starts at offset 4  and fills up the rest of the returned byte array.
+     * The ciphertext includes the nonce and (in case of GCM cipher) the tag, as detailed in the 
+     * Parquet Modular Encryption specification.
+     * @throws IOException
```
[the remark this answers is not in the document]
ggershinsky: will do

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param AAD Encryption AAD (ignored in case of CTR cipher)
+     * @return plaintext The plaintext starts at offset 0 of the input value, and fills up the entire byte array.
+     * @throws IOException
+     */
+    public byte[] decrypt(byte[] lengthAndCiphertext, byte[] AAD) throws IOException;
+    
+    /**
+     * Convenience decryption method that reads the length and ciphertext from the input stream.
+     * 
+     * @param from
```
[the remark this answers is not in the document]
ggershinsky: sounds good

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Decrypts the ciphertext. 
+     * 
+     * @param lengthAndCiphertext The first 4 bytes of the input value are the ciphertext length (little endian int). 
+     * The ciphertext starts at offset 4  and fills up the rest of the input byte array.
+     * The ciphertext includes the nonce and (in case of GCM cipher) the tag, as detailed in the 
+     * Parquet Modular Encryption specification.
+     * @param AAD Encryption AAD (ignored in case of CTR cipher)
+     * @return plaintext The plaintext starts at offset 0 of the input value, and fills up the entire byte array.
+     * @throws IOException
```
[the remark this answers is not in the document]
ggershinsky: will document

dev/travis-before_install-encryption.sh
ggershinsky: sounds good to me. this won't affect anything. should I close the pr, and open a new one against the parquet-mr/master?

ggershinsky: @nandorKollar can I ask you to rebase this branch? Or let me know if you prefer me to create a pull request for that.

nandorKollar: @ggershinsky done, let me retrigger Travis to see the test results.

nandorKollar: Seems that Travis fails because the default JDK version is now 11, #665 fixes it.

nandorKollar:
I think Travis fails, because the branch specific install script clones parquet-format inside the parquet-mr working folder (see this path /home/travis/build/apache/parquet-mr/parquet-format/pom.xml), and Rat check fails because parquet-format/blob/master/.github/PULL_REQUEST_TEMPLATE.md is not excluded (in format it is - that's why format builds are not failing, but in mr no exclude pattern matches this file).

I recommend modifying the encryption specific before_install script with cloning and building parquet-format in build directory (before clone, move to TRAVIS_BUILD_DIR `cd $TRAVIS_BUILD_DIR`), and at the end going back to the original directory. Something like this:
```
cd $TRAVIS_BUILD_DIR
git clone https://github.com/apache/parquet-format.git
mvn -f $TRAVIS_BUILD_DIR/apache/parquet-format/pom.xm install -DskipTests
cd -
```

nandorKollar:
Hmm okay, my bad, TRAVIS_BUILD_DIR is the cloned repository path as per https://docs.travis-ci.com/user/environment-variables/.
Then maybe this?
```
cd ../..
git clone -b encryption --single-branch https://github.com/apache/parquet-format.git
cd parquet-format
mvn install -DskipTests
cd $TRAVIS_BUILD_DIR
```

nandorKollar:
Ouch, hit log record limit. When building parquet-format, I recommend adding --batch-mode so that download progress report messages are not emitted.

ggershinsky:
yesss. thanks @nandorKollar for helping with this! I'll get working on addressing the comments.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
nandorKollar: I removed the extra casts, and didn't get any compile error. I'm also curious what was the error message you got.

ggershinsky: Yep, it compiles, but the reason is to prevent compiler to cast this to an overloaded method with a different object type in place of the decryptor etc. A sort of precaution.

ggershinsky:
Hi guys, I think I've addressed all comments, but let me know if I've missed anything.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
nandorKollar: I can't imagine any scenario where this could happen. I think if there's an overloaded method with different parameters, and the compile can't figure out which method to call, then it would be a compilation error. In that case casting is the way to give hint to the compiler, that you'd like to call the method with BlockCipher.Decryptor and byte[] parameters, but in this case the method invocation is clear without casting too.

ggershinsky: I see. Ok, looks like we have two options: 1) leave the code as is - the cast helps code readers to understand what is nulled (in this case, that we call methods without encryption); also, it's  harmless 2) remove the casts in this and other pull requests. 
Please let me know your preference, I'm fine with either option.

nandorKollar: I personally prefer more compact code, and looks like @Fokko thinks the same, so if you don't mind, then let's remove those casts, that are not necessary and don't add any value!

parquet-format-structures/src/main/java/org/apache/parquet/format/BlockCipher.java
nandorKollar: I think keeping these interfaces in the same file is acceptable, they're not too big.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
nandorKollar: nit, but not yet addressed, @ggershinsky could you please apply Junjie's recommendation?

nandorKollar:
@ggershinsky I think you missed some of the comments from @Fokko about final variables and formatting.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: Ok, will do that.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
ggershinsky: will do, sorry for missing this.

ggershinsky:
@nandorKollar I have made these variables final as per the recommendation from @Fokko. Have I missed some  variables? I'll check again.
Regarding auto-formatting - I don't know how to do that, that's why I've asked for an advise before. Is there a tool for that? Or a simple "ctrl-i" in Eclipse will do?

ggershinsky:
@nandorKollar  comments are addressed. 
To format the code, I've used the Eclipse formatting tool.

PR #620, opened 2019-02-15
PARQUET-1531: Page row count limit causes empty pages to be written from MessageColumnIO

gszadovszky:


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -305,6 +305,9 @@ long getRowsWrittenSoFar() {
    * Writes the current data to a new page in the page store
    */
   void writePage() {
+    if (valueCount == 0) {
+      throw new ParquetEncodingException("writing empty page");
```
vinooganesh: This leaves it up to the clients to handle right? Should this be transparent to the clients / only dealt with on the parquet side?

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -297,6 +297,11 @@ public void startMessage() {
     @Override
     public void endMessage() {
       writeNullForMissingFieldsAtCurrentLevel();
+
+      // We need to flush the cached null values before ending the record to ensure that everything is sent to the
+      // writer before the current page would be closed
+      flush();
```
rdblue: We need to run the benchmarks to see what effect this has on performance.

@isnotinvain, didn't you add some commits to delay flushing nulls to improve performance? I wonder if this reverts that change?

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
rdblue: @vinooganesh, I think this is intended as a sanity check to ensure we don't write empty pages. The new `flush` call in `MessageColumnIO` should fix the problem.

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
mccheah: Want to echo this - it looks like it effectively cancels out the null buffering from the above procedure `writeNullForMissingFieldsAtCurrentLevel`.

gszadovszky:
Based on the benchmark the flush-per-record leads to a significant 36% performance loss. See https://tinyurl.com/y3lhgx35 for details.
I'll try to come up with a more advanced solution to save performance.

gszadovszky:
Added a more advanced solution. It is not nice but I did not have a better idea. It even has a 12% performance gain (however I am not sure why): https://tinyurl.com/y2qxodyv.
Any comments/ideas are welcomed.

(The failure is caused by the new parquet-tools merge feature [PARQUET-1381](https://issues.apache.org/jira/browse/PARQUET-1381). I has conceptional design problems and as it is not a high priority one I am planning to revert it and reopen the JIRA.)

gszadovszky:
Any thoughts? I would like to push it in soon so I can come up with a new RC.

rdblue:
I'll have a look today or tomorrow.

gszadovszky:
@rdblue, could you please have a look on it so I can start a new RC for 1.11.0 (if you find this one correct)?

mccheah:
Ping @rdblue - would be great to confirm that this approach is how we want to go about fixing this issue.

rdblue:
+1

This looks good to me. Thanks for fixing it, @gszadovszky!

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
cervebar: @rdblue , @gszadovszky ... hi, sorry for writing this way, I think that probably write here direct to code would reach authors (as I'm starting to be despair :D ) .. I hit this error and don't know how to deal with it, I'm using it in MapReduce and do nothing special, provide WritteSupport and that's all...unfortunately I'm not able to reproduce it in any test,all test with my WriteSupport class passes ... here is my [ question on StackOverFlow](https://stackoverflow.com/questions/61776917/parquet-writer-org-apache-parquet-io-parquetencodingexception-writing-empty-pa) ... thanks for any help

imperio-wxm: > @rdblue , @gszadovszky ... hi, sorry for writing this way, I think that probably write here direct to code would reach authors (as I'm starting to be despair :D ) .. I hit this error and don't know how to deal with it, I'm using it in MapReduce and do nothing special, provide WritteSupport and that's all...unfortunately I'm not able to reproduce it in any test,all test with my WriteSupport class passes ... here is my [ question on StackOverFlow](https://stackoverflow.com/questions/61776917/parquet-writer-org-apache-parquet-io-parquetencodingexception-writing-empty-pa) ... thanks for any help

I encountered the same problem, how to solve it?

gszadovszky: You may check my answer at the referenced StackOverFlow thread.

PR #621, opened 2019-02-19
Revert "PARQUET-1381: Add merge blocks command to parquet-tools (#512)"

gszadovszky:
This reverts commit 863a081850e56bbbb38d7b68b478a3bd40779723.

The design of this feature has conceptional problems and also works incorrectly. See [PARQUET-1381](https://issues.apache.org/jira/browse/PARQUET-1381?focusedCommentId=16772080#comment-16772080) for more details.


zivanfi:
@kgalieva Unfortunately, as Gabor described above, the block-merging functionality seems to be incompatible with the internal logic of the API in its current form. I am sorry to see your contribution about to be reverted and would like to emphasize that your efforts have really been appreciated. If a way is found to resolve this conceptional misalignment, we would be happy to include this functionality in the future in an updated form.

PR #624, opened 2019-02-27
[PARQUET-1542] Merge multiple I/O to one time I/O in method readFooter

wangshisan:
Currently in method readFooter of class ParquetFileReader, it will invoke method converter.readParquetMetadata. To decode file metadata,  there may be quite a number of times of I/O processing underlying, which may heavily slow down parquet read performance.

A better way is to read all the footer bytes in a single I/O, since we have already got footer length before, and pass down all the bytes to underlying.

And we have a test in our environment, this change may bring  50 percent improvement for some cases.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -540,6 +540,8 @@ public class ParquetFileReader implements Closeable {
       throw new RuntimeException("corrupted file: the footer index is not within the file: " + footerIndex);
     }
     f.seek(footerIndex);
+    // Read all the footer bytes in one time to avoid multiple read operations,
+    // since it can be pretty time consuming for a single read operation in HDFS.
     ByteBuffer footerBytesBuffer = ByteBuffer.allocate(footerLength);
     f.readFully(footerBytesBuffer);
     LOG.debug("Finished to read all footer bytes.");
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -30,6 +30,7 @@ import static org.apache.parquet.hadoop.ParquetFileWriter.PARQUET_COMMON_METADAT
 import static org.apache.parquet.hadoop.ParquetFileWriter.PARQUET_METADATA_FILE;
 
 import java.io.Closeable;
+import java.io.InputStream;
 import java.io.IOException;
 import java.io.SequenceInputStream;
 import java.nio.ByteBuffer;
@@ -539,7 +540,12 @@ public class ParquetFileReader implements Closeable {
       throw new RuntimeException("corrupted file: the footer index is not within the file: " + footerIndex);
     }
     f.seek(footerIndex);
-    return converter.readParquetMetadata(f, options.getMetadataFilter());
+    ByteBuffer footerBytesBuffer = ByteBuffer.allocate(footerLength);
+    f.readFully(footerBytesBuffer);
+    LOG.debug("Finished to read all footer bytes.");
+    footerBytesBuffer.flip();
+    InputStream footerBytesStream = ByteBufferInputStream.wrap(footerBytesBuffer);
+    return converter.readParquetMetadata(footerBytesStream, options.getMetadataFilter());
   }
 
   /**


nandorKollar:
There are test failures, however they look unrelated, it looks like Twitter maven repo is down, and we pull something from there. Travis should be retriggered once it is up again.

gszadovszky:
I'm afraid I don't understand why your change would boost the reading. As far as I can see thrift reads all the bytes of the footer only that in some cases not the whole thrift structure is created. Thrift does not seem to do any further seeks or skipping bytes in the underlying input stream.
Are you sure this change made your system working faster? Could you write a benchmark that proves this change would increase the performance in your scenario and would not decrease it significantly in any other ones?

wangshisan:
Underlie of method converter.readParquetMetadata, it will read from the InputStream and deserialize to a FileMetaData object. During serialization, it will read all the filed bytes one by one, you can find this logic in method FileMetaData.read(), unfortunately, the InputStream is without any buffer, that means for each field, there will be at least one I/O, this is pretty time consuming especially for HDFS.

gszadovszky:
I see, thanks for describing.
Could you please provide some benchmarks for different scenarios (by simulating remote reads, reading local files etc.). It would be nice to see that this change really improves the reading in specific cases (if not all of them) and does not have a significant penalty for the others.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -539,7 +540,12 @@ private static final ParquetMetadata readFooter(InputFile file, ParquetReadOptio
       throw new RuntimeException("corrupted file: the footer index is not within the file: " + footerIndex);
     }
     f.seek(footerIndex);
-    return converter.readParquetMetadata(f, options.getMetadataFilter());
+    ByteBuffer footerBytesBuffer = ByteBuffer.allocate(footerLength);
```
gszadovszky: Could you add some comments about the purpose of this code part?

chenjunjiedada: I think this is used to read all footer into a byte buffer instead of reading them byte by byte.

wangshisan: Sorry to miss this for so long... Will add a comment later.

wangshisan:
@gszadovszky Sorry, I'm not quite clear, do you mean I need add a benchmark in Parquet-Benchmarks module, or some benchmark result in our Product environment?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
wangshisan: Comment added

gszadovszky:
The best thing would be adding some benchmarks into the parquet-benchmarks module. However, I am not sure how can we properly simulate HDFS. On local disks, it probably won't have any performance benefits because the OS would pre-read the block to memory before giving it to the java reader.
I don't know if HDFS would do similar tricks on remote reads or it would stream the data through the wire. That's why I am interested in some benchmarks. Currently, I am not convinced this change would boost the footer reading.
If you are able to do some benchmarks in your environment with HDFS remote and local reads with and without this fix and provide the results, it would be great. :)

wangshisan:
I made a benchmark on two of our production tables.

- Table a, 66 columns:

Run | Before(ms) | After(ms)
-- | -- | --
Run 1 | 65 | 40
Run 2 | 67 | 43
Run 3 | 59 | 43
Run 4 | 65 | 46
Run 5 | 60 | 39



- Table b, 155 columns.

Run | Before(ms) | After(ms)
-- | -- | --
Run 1 | 85 | 52
Run 2 | 99 | 57
Run 3 | 83 | 56
Run 4 | 91 | 53
Run 5 | 87 | 52

gszadovszky:
Thanks a lot, @wangshisan.
Do you know if these all were remote reads? It would be surprising to me if your change would boost local reads.
Anyway, I'll approve this change.

wangshisan:
They are all remote read.

wangshisan:
@gszadovszky  Thanks for your approval.
@chenjunjiedada Do you have any suggestion?

chenjunjiedada:
LGTM +1

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -539,7 +540,14 @@ private static final ParquetMetadata readFooter(InputFile file, ParquetReadOptio
...
     }
     f.seek(footerIndex);
-    return converter.readParquetMetadata(f, options.getMetadataFilter());
+    // Read all the footer bytes in one time to avoid multiple read operations,
+    // since it can be pretty time consuming for a single read operation in HDFS.
+    ByteBuffer footerBytesBuffer = ByteBuffer.allocate(footerLength);
+    f.readFully(footerBytesBuffer);
+    LOG.debug("Finished to read all footer bytes.");
+    footerBytesBuffer.flip();
+    InputStream footerBytesStream = ByteBufferInputStream.wrap(footerBytesBuffer);
```
Fokko: Shouldn't we close this input stream?

Fokko: PTAL: https://github.com/apache/parquet-mr/pull/685/

gszadovszky: Technically, you're right. But `ByteBufferInputStream.close()` does nothing. So it is not a bug to not close it.

PR #627, opened 2019-03-11
PARQUET-1534:  [parquet-cli] Argument error: Illegal character in opaque part at index 2 on Windows

masayuki038:
Some commands of parquet-cli don't work on Windows.

```
$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main schema C:\Users\masayuki\Downloads\test.parquet                                                               Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.parquet          
                                    
$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main csv-schema C:\Users\masayuki\Downloads\test.csv --record-name Test                                              
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.csv              

$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main convert-csv C:\Users\masayuki\Downlo ads\test.csv -o C:\Users\masayuki\Downloads\test-csv.parquet
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.csv

$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main convert C:\Users\masayuki\Downloads\ userdata1.avro -o C:\Users\masayuki\Downloads\userdata1.parquet
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\userdata1.avro

$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main to-avro C:\Users\masayuki\Downloads\ test.parquet -o C:\Users\masayuki\Downloads\test.avro
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.parquet

$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main cat C:\Users\masayuki\Downloads\test.parquet
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.parquet

$ java -cp target/classes;target/dependency/* org.apache.parquet.cli.Main head -n 10 C:\Users\masayuki\Downloa ds\test.parquet
Argument error: Illegal character in opaque part at index 2: C:\Users\masayuki\Downloads\test.parquet
```

Calling BaseCommand#qualifiedURI with Windows file path,  java.net.URI.create throws IllegalArgumentException and the command execution is aborted.

This PR fixes the issue. With this PR, parquet-cli will work all commands on Windows.

parquet-cli has no tests. I created PR to add simple tests for each command at https://github.com/apache/parquet-mr/pull/625. Could you review that first?

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
@@ -50,6 +50,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
@@ -175,12 +176,13 @@ public abstract class BaseCommand implements Command, Configurable {
    * @throws IOException if there is an error creating a qualified URI
    */
   public URI qualifiedURI(String filename) throws IOException {
-    URI fileURI = URI.create(filename);
-    if (RESOURCE_URI_SCHEME.equals(fileURI.getScheme())) {
-      return fileURI;
-    } else {
-      return qualifiedPath(filename).toUri();
-    }
+    try {
+      URI fileURI = new URI(filename);
+      if (RESOURCE_URI_SCHEME.equals(fileURI.getScheme())) {
+        return fileURI;
+      }
+    } catch (URISyntaxException ignore) {}
+    return qualifiedPath(filename).toUri();
   }
 
   /**
@@ -392,5 +394,4 @@ public abstract class BaseCommand implements Command, Configurable {
           "Could not determine file format of %s.", source));
     }
   }
-
 }
--- a/parquet-cli/src/test/java/org/apache/parquet/cli/BaseCommandTest.java
+++ b/parquet-cli/src/test/java/org/apache/parquet/cli/BaseCommandTest.java
@@ -21,6 +21,7 @@ package org.apache.parquet.cli;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


gszadovszky:
Could you fix the test failures?

masayuki038:
@gszadovszky I fixed the test failures. Could you review this?

parquet-cli/src/test/java/org/apache/parquet/cli/BaseCommandTest.java
```diff
@@ -45,24 +46,32 @@ public void setUp() {
...
   public void qualifiedPathOnLinuxTest() throws IOException {
+    Assume.assumeFalse
+             (System.getProperty("os.name").toLowerCase().startsWith("win"));
     Path path = this.command.qualifiedPath(LINUX_FILE_PATH);
     Assert.assertEquals("test.parquet", path.getName());
   }
 
   @Test
   public void qualifiedPathOnWinTest() throws IOException {
+    Assume.assumeFalse
```
gszadovszky: Is it working on a Mac? Why assumeFalse? I think, it would be cleaner to execute tests on an actual OS even if some are not covered (e.g. Mac).

masayuki038: @gszadovszky Thanks for your response.

I forgot about Mac ...
I don't have it now. Could anyone check this on Mac?

> Why assumeFalse? 

The reason why it uses assumeFalse, `org.apache.hadoop.fs.Path.initialize` throws IllegalArgumentException depending on the condition. 

| Path Style |  Linux Style  |  Windows Style |
| ---- | ---- | ---- |
| on Linux |  OK  |  IllegalArgumentException |
| on Windows |  OK  |  OK  |

For testing on Linux OS, I changed the code to not execute the Windows Path Style tests. 

> I think, it would be cleaner to execute tests on an actual OS even if some are not covered (e.g. Mac).

I agree. Should I eliminate test code that depends on OS?

gszadovszky: I don't have a Mac either but I think it should work similarly as on Linux.

There is no problem with OS dependent tests like this one where the behavior itself depends on the OS. What I didn't like is `assumeFalse` instead of `assumeTrue`. I mean that we should assume on an actual OS and not that we are not using one of them. It may lead to some unexpected situations. I think, it is better to control the execution on the OS we know how it should work.
So, I suggest executing Win related tests with assuming it is a Win OS and executing the others assuming Linux or Mac.

masayuki038: @gszadovszky Thanks for your comment. I fix the test code. Could you review it again?

masayuki038:
Could anyone merge or review this?

parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
```diff
@@ -175,12 +176,13 @@ public Path qualifiedPath(String filename) throws IOException {
...
-      return fileURI;
-    } else {
-      return qualifiedPath(filename).toUri();
-    }
+    try {
+      URI fileURI = new URI(filename);
+      if (RESOURCE_URI_SCHEME.equals(fileURI.getScheme())) {
+        return fileURI;
+      }
+    } catch (URISyntaxException ignore) {}
```
rdblue: This is a behavior change that breaks the stated contract of this method: `@throws IOException if there is an error creating a qualified URI`.

In addition, the tests do not validate this change. What happens to invalid URIs after this change? Are characters automatically escaped by using `Path`? What about a resource URI with an unescaped character?

It is fine to change the method contract, but the new behavior must be as well understood as the previous behavior and it should be tested. The tests in this PR check, for example, that the file name has not changed and don't exercise the error handling code at all.

masayuki038: @rdblue Thanks for your comment.

> This is a behavior change that breaks the stated contract of this method: @throws IOException if there is an error creating a qualified URI.

I don't agree. `URI.create` and `fileURI.getScheme()` don't throw `IOException` and `URISyntaxException` is not a subclass of IOException. 

- https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#create-java.lang.String-
- https://docs.oracle.com/javase/8/docs/api/java/net/URISyntaxException.html

> In addition, the tests do not validate this change. What happens to invalid URIs after this change? 

The difference from `URI.create` and `new URI` is  only whether the exception to throw is the IllegalArgumentException or URISyntaxException. The reason to use `new URI` is that catching with IllegalArgumentException may catch an unintended exception.

https://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/2d4edd21bb38/src/share/classes/java/net/URI.java#l848

PR #628, opened 2019-03-18
PARQUET-1544: Possible over-shading of modules

gszadovszky:


mccheah:
Tested locally, I think this fixes the problem. Thanks!

PR #629, opened 2019-03-21
add meta data when writing union types to parquet

mjhoshea:
Include metadata when writing nullable fields to parquet


mjhoshea:
Hi @gszadovszky, does this seem reasonable?

gszadovszky:
Unfortunately, I'm not familiar with this code part.
But, you should create a JIRA for this issue describing your problem and name this PR after the JIRA. (See any recent pull request/JIRA for example.) Also, it would be nice to have a unit test to validate your change.

PR #631, opened 2019-03-27
Adjust the PagedRecordFilter so it supports multiple groups

hadrienk:
The method UnboundRecordFilter#bind() is called more than once so a new instance of PagedRecordFilter is created several times with the currentPos set to 0 again.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter/PagedRecordFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter/PagedRecordFilter.java
@@ -20,14 +20,16 @@ package org.apache.parquet.filter;
 
 import org.apache.parquet.column.ColumnReader;
 
+import java.util.concurrent.atomic.AtomicLong;
+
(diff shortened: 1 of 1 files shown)


parquet-column/src/main/java/org/apache/parquet/filter/PagedRecordFilter.java
```diff
@@ -52,14 +49,25 @@ private PagedRecordFilter(long startPos, long pageSize) {
     this.endPos   = startPos + pageSize;
   }
 
-  /**
-   * Keeps track of how many times it is called. Only returns matches when the
-   * record number is in the range.
-   */
   @Override
-  public boolean isMatch() {
```
nandorKollar: @hadrienk I'm afraid that the build fails, because semantic versioning doesn't like the removal of this method. Can you please find a way to keep `isMatch`, while also implement the `UnboundRecordFilter` interface too?

PR #636, opened 2019-04-24
PARQUET-1557 Replace deprecated Avro methods

Fokko:
Some fields are deprecated in Avro 1.8.2 and are being removed in Avro 1.9.0, therefore we want to replace these fields.

https://issues.apache.org/jira/browse/PARQUET-1557


Fokko:
Gente ping @zivanfi :-)

zivanfi:
> Gentle ping @zivanfi :-)

The reason for me not replying earlier is that... I haven't been asked to review this PR before. :)

Fokko:
Thanks @zivanfi 

As I see you occasionally merging PR's, so I suspect that you would go down the list once a while :-)

zivanfi:
> As I see you occasionally merging PR's, so I suspect that you would go down the list once a while :-)

Recently not so much, only if I am specifically asked to review a change and have time at hand and expertise to do so.

PR #643, opened 2019-05-20
PARQUET-1229 Parquet-MR encryption

ggershinsky:


parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -480,6 +548,8 @@ private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGrou
       parquetColumns.add(columnChunk);
     }
     RowGroup rowGroup = new RowGroup(parquetColumns, block.getTotalByteSize(), block.getRowCount());
+    rowGroup.setTotal_compressed_size(block.getCompressedSize());
+    rowGroup.setOrdinal(rowGroupOrdinal);
     rowGroups.add(rowGroup);
```
chrysan: why you remove "rowGroup.setFile_offset(block.getStartingPos());" here? It will cause problem when the first column is encrypted but we want to read other un-encrypted columns without decryptProperties.

shangxinli:
I think we should stop merging Parquet Encryption for now before we can have a clear understanding of how this feature is adopted.

ggershinsky:
Too outdated, closing it. A new pull request for this jira will be sent soon.

PR #647, opened 2019-06-11
PARQUET-1580: Page-level CRC checksum verfication for DataPageV1

bbraams:
This PR implements page-level CRC checksums for DataPageV1. It is the implementation follow up to the clarification of the checksums in `parquet-format` ([Jira](https://jira.apache.org/jira/browse/PARQUET-1539), [PR](https://github.com/apache/parquet-format/pull/126)). Key points:
- Checksums are calculated using the CRC32 implementation in `java.util.zip`
- Checksums are implemented for `DataPageV1` and `DictionaryPage`
- Writing out checksums is **enabled** by default (impact is minimal, see Benchmark results below)
- Checksum verification on the read path is **disabled** by default (benchmarks indicate minimal impact, but requires more thorough benchmarking)
- Backwards compatibility with Parquet files that do not have the CRC field set, even if checksum verification is turned on

### Jira
- [x] My PR addresses the following [Parquet Jira](https://jira.apache.org/jira/browse/PARQUET-1580) issue:
  - https://jira.apache.org/jira/browse/PARQUET-1580

### Tests
This PR adds the following tests in a new test suite `TestDataPageV1Checksums`:
- Test writing out checksums without verification on read, verify calculated checksums are correct
- Test not writing out checksums without verification on read, tests feature can be turned off entirely
- Test not writing out checksums with verification on read, tests backward compatibility of read path with files without checksums
- Test writing out checksums with verification on read, tests that the checksum is correctly set and that we can read what we wrote
- Test corruption detection by corrupting bytes in data section of pages, tests that these corruptions go undetected with the feature disabled and are detected with the feature enabled
- Test checksums with compression enabled, tests that the checksum is calculated on the compressed version of the data (as per the specification)
- Test checksums with compression and nested schema, tests that the checksum is calculated on the compressed concatenation of the repetition levels, definition levels and the actual data (as per the specification)
- Test checksums with dictionary encoded data, tests that the dictionary page is also checksummed

### Documentation
The feature is feature flagged and is disabled by default. Both writing out checksums and verifying them on the read path can be turned on individually, via the following two new config flags:
* `parquet.page.write-checksum.enabled` (default: `true`)
* `parquet.page.verify-checksum.enabled` (default: `false`)

### Benchmark results
This PR adds 2 new benchmark suites, benchmarking the penalty of both writing out checksums and verifying them. I've run the suites on the following setup:
- AWS m4.4xlarge instance
- Single shot mode
- 3 warmup iterations
- 5 measured iterations (3 forked JVMs)

**PageChecksumWriteBenchmarks**

```
Benchmark                                  Mode  Cnt   Score   Error  Units
write100KRowsGzipWithChecksums               ss   15   0.500 ? 0.012   s/op
write100KRowsGzipWithoutChecksums            ss   15   0.497 ? 0.013   s/op
write100KRowsSnappyWithChecksums             ss   15   0.235 ? 0.015   s/op
write100KRowsSnappyWithoutChecksums          ss   15   0.239 ? 0.018   s/op
write100KRowsUncompressedWithChecksums       ss   15   0.213 ? 0.012   s/op
write100KRowsUncompressedWithoutChecksums    ss   15   0.210 ? 0.013   s/op
write1MRowsGzipWithChecksums                 ss   15   4.720 ? 0.018   s/op
write1MRowsGzipWithoutChecksums              ss   15   4.718 ? 0.013   s/op
write1MRowsSnappyWithChecksums               ss   15   2.288 ? 0.182   s/op
write1MRowsSnappyWithoutChecksums            ss   15   2.122 ? 0.122   s/op
write1MRowsUncompressedWithChecksums         ss   15   1.945 ? 0.137   s/op
write1MRowsUncompressedWithoutChecksums      ss   15   1.913 ? 0.124   s/op
write10MRowsGzipWithChecksums                ss   15  47.226 ? 0.127   s/op
write10MRowsGzipWithoutChecksums             ss   15  47.430 ? 0.306   s/op
write10MRowsSnappyWithChecksums              ss   15  21.164 ? 0.628   s/op
write10MRowsSnappyWithoutChecksums           ss   15  20.905 ? 0.332   s/op
write10MRowsUncompressedWithChecksums        ss   15  18.806 ? 0.390   s/op
write10MRowsUncompressedWithoutChecksums     ss   15  18.637 ? 0.201   s/op
```

**PageChecksumReadBenchmarks**

```
Benchmark                                    Mode  Cnt  Score   Error  Units
read100KRowsGzipWithVerification               ss   15  0.105 ? 0.005   s/op
read100KRowsGzipWithoutVerification            ss   15  0.104 ? 0.004   s/op
read100KRowsSnappyWithVerification             ss   15  0.080 ? 0.003   s/op
read100KRowsSnappyWithoutVerification          ss   15  0.081 ? 0.006   s/op
read100KRowsUncompressedWithVerification       ss   15  0.074 ? 0.002   s/op
read100KRowsUncompressedWithoutVerification    ss   15  0.071 ? 0.005   s/op
read1MRowsGzipWithVerification                 ss   15  0.956 ? 0.012   s/op
read1MRowsGzipWithoutVerification              ss   15  0.952 ? 0.025   s/op
read1MRowsSnappyWithVerification               ss   15  0.707 ? 0.025   s/op
read1MRowsSnappyWithoutVerification            ss   15  0.699 ? 0.013   s/op
read1MRowsUncompressedWithVerification         ss   15  0.676 ? 0.013   s/op
read1MRowsUncompressedWithoutVerification      ss   15  0.651 ? 0.025   s/o
read10MRowsGzipWithVerification                ss   15  9.717 ? 0.588   s/op
read10MRowsGzipWithoutVerification             ss   15  9.593 ? 0.310   s/op
read10MRowsSnappyWithVerification              ss   15  7.038 ? 0.113   s/op
read10MRowsSnappyWithoutVerification           ss   15  6.854 ? 0.078   s/op
read10MRowsUncompressedWithVerification        ss   15  6.544 ? 0.071   s/op
read10MRowsUncompressedWithoutVerification     ss   15  6.463 ? 0.121   s/op
```

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumReadBenchmarks.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumReadBenchmarks.java
@@ -20,13 +20,29 @@ package org.apache.parquet.benchmarks;
 
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.example.data.Group;
-import org.apache.parquet.hadoop.*;
+import org.apache.parquet.hadoop.ParquetReader;
 import org.apache.parquet.hadoop.example.GroupReadSupport;
-import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.infra.Blackhole;
 
-import static org.apache.parquet.benchmarks.BenchmarkConstants.*;
-import static org.apache.parquet.benchmarks.BenchmarkFiles.*;
+import static org.apache.parquet.benchmarks.BenchmarkConstants.ONE_K;
+import static org.apache.parquet.benchmarks.BenchmarkConstants.ONE_MILLION;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.configuration;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_100K_CHECKSUMS_UNCOMPRESSED;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_100K_CHECKSUMS_GZIP;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_100K_CHECKSUMS_SNAPPY;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_1M_CHECKSUMS_UNCOMPRESSED;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_1M_CHECKSUMS_GZIP;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_1M_CHECKSUMS_SNAPPY;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_10M_CHECKSUMS_UNCOMPRESSED;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_10M_CHECKSUMS_GZIP;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.file_10M_CHECKSUMS_SNAPPY;
 
 import java.io.IOException;
 
@@ -36,7 +52,7 @@ public class PageChecksumReadBenchmarks {
   private PageChecksumDataGenerator pageChecksumDataGenerator = new PageChecksumDataGenerator();
 
   @Setup(Level.Trial)
-  public void setup() throws IOException {
+  public void setup() {
     pageChecksumDataGenerator.generateAll();
   }
 
@@ -47,21 +63,21 @@ public class PageChecksumReadBenchmarks {
 
   private void readFile(Path file, int nRows, boolean verifyChecksums, Blackhole blackhole)
     throws IOException {
-    ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), file)
-      .withConf(configuration)
-      .usePageChecksumVerification(verifyChecksums)
-      .build();
-    for (int i = 0; i < nRows; i++) {
-      Group group = reader.read();
-      blackhole.consume(group.getLong("long_field", 0));
-      blackhole.consume(group.getBinary("binary_field", 0));
-      Group subgroup = group.getGroup("group", 0);
-      blackhole.consume(subgroup.getInteger("int_field", 0));
-      blackhole.consume(subgroup.getInteger("int_field", 1));
-      blackhole.consume(subgroup.getInteger("int_field", 2));
-      blackhole.consume(subgroup.getInteger("int_field", 3));
+    try (ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), file)
+        .withConf(configuration)
+        .usePageChecksumVerification(verifyChecksums)
+        .build()) {
+      for (int i = 0; i < nRows; i++) {
+        Group group = reader.read();
+        blackhole.consume(group.getLong("long_field", 0));
+        blackhole.consume(group.getBinary("binary_field", 0));
+        Group subgroup = group.getGroup("group", 0);
+        blackhole.consume(subgroup.getInteger("int_field", 0));
+        blackhole.consume(subgroup.getInteger("int_field", 1));
+        blackhole.consume(subgroup.getInteger("int_field", 2));
+        blackhole.consume(subgroup.getInteger("int_field", 3));
+      }
     }
-    reader.close();
   }
 
   // 100k rows, uncompressed, GZIP, Snappy
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumReadBenchmarks.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumReadBenchmarks.java
@@ -1,0 +1,163 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.benchmarks;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.example.data.Group;
+import org.apache.parquet.hadoop.*;
+import org.apache.parquet.hadoop.example.GroupReadSupport;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import static org.apache.parquet.benchmarks.BenchmarkConstants.*;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.*;
+
+import java.io.IOException;
+
+@State(Scope.Thread)
+public class PageChecksumReadBenchmarks {
+
+  private PageChecksumDataGenerator pageChecksumDataGenerator = new PageChecksumDataGenerator();
+
+  @Setup(Level.Trial)
+  public void setup() throws IOException {
+    pageChecksumDataGenerator.generateAll();
+  }
+
+  @Setup(Level.Trial)
+  public void cleanup() {
+    pageChecksumDataGenerator.cleanup();
+  }
+
+  private void readFile(Path file, int nRows, boolean verifyChecksums, Blackhole blackhole)
+    throws IOException {
+    ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), file)
+      .withConf(configuration)
+      .usePageChecksumVerification(verifyChecksums)
+      .build();
+    for (int i = 0; i < nRows; i++) {
+      Group group = reader.read();
+      blackhole.consume(group.getLong("long_field", 0));
+      blackhole.consume(group.getBinary("binary_field", 0));
+      Group subgroup = group.getGroup("group", 0);
+      blackhole.consume(subgroup.getInteger("int_field", 0));
+      blackhole.consume(subgroup.getInteger("int_field", 1));
+      blackhole.consume(subgroup.getInteger("int_field", 2));
+      blackhole.consume(subgroup.getInteger("int_field", 3));
+    }
+    reader.close();
+  }
+
+  // 100k rows, uncompressed, GZIP, Snappy
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsUncompressedWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_UNCOMPRESSED, 100 * ONE_K, false, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsUncompressedWithVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_UNCOMPRESSED, 100 * ONE_K, true, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsGzipWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_GZIP, 100 * ONE_K, false, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsGzipWithVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_GZIP, 100 * ONE_K, true, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsSnappyWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_SNAPPY, 100 * ONE_K, false, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read100KRowsSnappyWithVerification(Blackhole blackhole) throws IOException {
+    readFile(file_100K_CHECKSUMS_SNAPPY, 100 * ONE_K, true, blackhole);
+  }
+
+  // 1M rows, uncompressed, GZIP, Snappy
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read1MRowsUncompressedWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_1M_CHECKSUMS_UNCOMPRESSED, ONE_MILLION, false, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read1MRowsUncompressedWithVerification(Blackhole blackhole) throws IOException {
+    readFile(file_1M_CHECKSUMS_UNCOMPRESSED, ONE_MILLION, true, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read1MRowsGzipWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_1M_CHECKSUMS_GZIP, ONE_MILLION, false, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read1MRowsGzipWithVerification(Blackhole blackhole) throws IOException {
+    readFile(file_1M_CHECKSUMS_GZIP, ONE_MILLION, true, blackhole);
+  }
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
+  public void read1MRowsSnappyWithoutVerification(Blackhole blackhole) throws IOException {
+    readFile(file_1M_CHECKSUMS_SNAPPY, ONE_MILLION, false, blackhole);
+  }
+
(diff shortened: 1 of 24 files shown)


bbraams:
@gszadovszky Thanks for quick review and comments! I agree that your proposed approach would indeed simplify the implementation and that there is indeed little use of exposing access to the crc in the DataPage class. There are however a few drawbacks of doing it this way that I can think of:

1. Less robust testing. We lose reference to the CRC up the call stack if we do it under the hood, so for testing we will have to rely on the 'side effects' of the implementation to verify correct behavior (either an exception is thrown on failed verification or not). Maybe we could implement some helper functions just for use in testing to expose the crc. But I'm not sure if we wouldn't end up with an implementation very similar to the current one (where we would have CRC in `Page` instead).
2. No easy way for users to see if their files have CRC set. In the current implementation, in `parquet-tools`, I just hook into the loop over all pages in the `dump` command. Again, I think we could solve this by implementing a boolean flag at the `PageReadStore` level or something set to true whenever we encountered a checksum in `readAllPages`.

Regarding point 1, given that the checksum calculation and verification are really only a handful of loc's, we might get away with less robust testing. Regarding point 2, I think this can be acceptable.

I'm on board for going for the 'under the hood' implementation given my point above. Your thoughts?

gszadovszky:
As the crc in the footer structure is listed in `PageHeader` we should add it to `Page` instead of `DataPageV1` or `DataPage`. What I don't really like here is that we already have a couple of constructors which we should not modify (backward compatibility) but add new ones and deprecate the old ones. However, we can solve this by implementing builders. If we are going to the direction of propagating the `crc` to the "user objects" I would vote on having `crc` in `Page` and creating the proper builders for all the supported page types.

If we prefer the "under the hood" solution then I agree that the testing and extending the tools to deal with the crc would be much harder.

After thinking a bit more I am not that confident that the "under the hood" solution is better than the other one but I would still vote on it because the public API could stay a bit cleaner and most of the users are not interested in the crc value. From testing point of view I think we can still validate all the possible cases (crc verification on/off + valid/invalid/missing crc values) only that we have to implement the tests at a higher level (writing actual parquet files with "hacked" crc values and read them back with different config setup).

It is another issue if we really want to have a tool that lists the crc values or the fact that there are valid crc values saved for the pages. It is a separate topic but I think a tool a tool that could list the original values of the thrift object without translating to them to "user objects" would help a lot to debug some low level issues. Such tool would easily list the crc values of the page headers even if the check is implemented "under the hood". For this current issue I can live without such a tool. :)

bbraams:
Alright, I agree that we should move forward with the 'under the hood' solution. Some points to note:

- Even though we won't be adding the crc to the implementations of the `Page` class, we should still respect the specification and thus also set/verify the crc for `DictionaryPage`'s.
- Previously we would perform the verification in `ColumnChunkPageReader.readPage`, which I believe is called 'on-demand'. We will now move it to `readAllPages`, which would make it less lazy in a sense, since the checksum verification will be performed before the pages are actually delivered? I don't believe this to be a big problem though.

gszadovszky:
The specification (`parquet.thrift`) contains 5 different page types. One was never used in parquet-mr (`IndexPageHeader`), another is not implemented / not in master yet (`BloomFilterPageHeader`). The other 3 is used: pagev1, pagev2 and dictionary. Since this JIRA is about implementing CRC for V1 pages only, I am OK with implementing it for that only. However, it does not seem to be a big deal to implement it for all the 3.

The current idea is to throw an exception (and therefore fail the whole reading process) if the verify flag is on and a CRC check fails. I think, it is better to throw that exception as early as possible. All the pages are processed in `readAllPages` will be required later anyway.

bbraams:
I’ve addressed some of the code review comments and made some changes:
- The writing and verification of the checksums are now done ‘under the hood’, removing the explicit crc from DataPageV1
- Added support for dictionary pages as per the specification
- Added getters/setter for a copy of the crc value to `Page` solely for the purpose of testing

With respect to the last change, I believe this is a reasonable compromise for what we’ve been discussing (having the crc in the public API for robust testing vs. handling crc’s only under the hood and having to rely on pre-generated files with baked crc’s for testing). Note that crc that we store in the `Page` object is merely a copy of the thrift object, and is not actually used to write out the crc or during validation (we use low level thrift values accessed in `readAllPages` for the latter). The presence of the crc in `Page` is not reflected in any (subclass) constructor, and is therefore relatively hidden. Ideally we’d have the setters/getters be package private, but this is unfortunately not possible (`parquet.column` vs `parquet.hadoop`). What we do get is more robust testing (see the `TestDataPageV1Checksums` suite), and the ability to display the crc in `parquet-tools` (I’ve added it to the `dump` command).

Is this an agreeable solution for you?

I’m currently writing some benchmarks for `parquet-benchmarks` to see the performance penalty on both the write and read path. These numbers should give us confidence (or not) in enabling writing checksums by default.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1338,13 +1338,15 @@ public void writeDataPageHeader(
       org.apache.parquet.column.Encoding rlEncoding,
       org.apache.parquet.column.Encoding dlEncoding,
       org.apache.parquet.column.Encoding valuesEncoding,
+      int crc,
```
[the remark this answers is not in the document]
bbraams: Thanks for the catch, thought I'd caught all the extensions of deprecated methods.

parquet-column/src/main/java/org/apache/parquet/column/page/Page.java
```diff
@@ -43,4 +43,25 @@ public int getUncompressedSize() {
...
+  private boolean isSetCrc = false;
+
+  // Visible for testing
+  public void setCrc(int crc) {
+    this.crc = crc;
+    this.isSetCrc = true;
+  }
+
+  // Visible for testing
+  public int getCrc() {
```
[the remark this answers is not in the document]
bbraams: You're right, will do!

parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
```diff
@@ -274,6 +274,7 @@ public Void visit(DataPageV1 pageV1) {
                 } else {
                   out.format(" ST:[none]");
                 }
+                out.format(" CRC:%s", pageV1.isSetCrc() ? pageV1.getCrc() : "-");
```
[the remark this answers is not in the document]
bbraams: That indeed makes more sense, will add the result of verification here instead of just printing out the checksum.

bbraams:
I've addressed your comments and included benchmarks. I've created a new Jira ticket for adding support for DataPageV2 (https://jira.apache.org/jira/browse/PARQUET-1629).

I'm running the benchmarks as we speak and will report the results once I have them!

Fokko:
There are still failing tests:
```
Failed tests:   testAlignmentWithNoPaddingNeeded(org.apache.parquet.hadoop.TestParquetFileWriter): Second row group should start after no padding expected:<109> but was:<139>
  testAlignmentWithPadding(org.apache.parquet.hadoop.TestParquetFileWriter): First row group should end before the block size (120)
```

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+import org.apache.parquet.io.*;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.MessageTypeParser;
+import org.apache.parquet.schema.Types;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT32;
+import static org.junit.Assert.*;
```
Fokko: Please avoid wildcard imports

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+import org.apache.parquet.example.data.Group;
+import org.apache.parquet.example.data.GroupFactory;
+import org.apache.parquet.example.data.simple.SimpleGroupFactory;
+import org.apache.parquet.hadoop.codec.SnappyCompressor;
+import org.apache.parquet.hadoop.example.ExampleParquetWriter;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.hadoop.util.HadoopInputFile;
+import org.apache.parquet.hadoop.util.HadoopOutputFile;
+import org.apache.parquet.io.*;
```
Fokko: Please avoid wildcard imports

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+import java.util.*;
+import java.util.zip.CRC32;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.bytes.BytesInput;
+import org.apache.parquet.bytes.HeapByteBufferAllocator;
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.Encoding;
+import org.apache.parquet.column.page.*;
```
Fokko: Please avoid wildcard imports

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.hadoop;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
```
Fokko: Please avoid wildcard imports

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+
+  private Path writeNestedWithNullsSampleParquetFile(Configuration conf,
+                                                     boolean dictionaryEncoding,
+                                                     CompressionCodecName compression)
+    throws IOException {
+    File file = tempFolder.newFile();
+    file.delete();
+    Path path = new Path(file.toURI());
+
+    ParquetWriter<Group> writer = ExampleParquetWriter.builder(path)
```
Fokko: Can we wrap a try-with-resource around this

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+   */
+  @Test
+  public void testCorruptedPage() throws IOException {
+    Configuration conf = new Configuration();
+    conf.setBoolean(ParquetOutputFormat.PAGE_WRITE_CHECKSUM_ENABLED, true);
+
+    Path path = writeSimpleParquetFile(conf, CompressionCodecName.UNCOMPRESSED);
+
+    InputFile inputFile = HadoopInputFile.fromPath(path, conf);
+    SeekableInputStream inputStream = inputFile.newStream();
```
Fokko: Can we use a try-with-resource here as well?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,547 @@
...
+    inputStream.readFully(fileBytes);
+    inputStream.close();
+
+    // There are 4 pages in total (2 per column), we corrupt the first page of the first column and
+    // the second page of the second column. We do this by altering a byte roughly in the middle of
+    // each page to be corrupted
+    fileBytes[fileLen / 8]++;
+    fileBytes[fileLen / 8 + ((fileLen / 4) * 3)]++;
+
+    OutputFile outputFile = HadoopOutputFile.fromPath(path, conf);
```
Fokko: Can we use a try-with-resource here as well?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -99,8 +99,9 @@ public DataPage readPage() {
         public DataPage visit(DataPageV1 dataPageV1) {
           try {
             BytesInput decompressed = decompressor.decompress(dataPageV1.getBytes(), dataPageV1.getUncompressedSize());
+            DataPageV1 decompressedPage;
```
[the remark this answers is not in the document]
bbraams: Good catch, thanks

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
bbraams: Again good catch, thanks

parquet-column/src/main/java/org/apache/parquet/column/page/Page.java
```diff
@@ -43,4 +45,18 @@ public int getUncompressedSize() {
     return uncompressedSize;
   }
 
+  // Note: the following fields are only used for testing purposes and are NOT used in checksum
```
gszadovszky: nit:
```suggestion
  // Note: the following field is only used for testing purposes and is NOT used in checksum
```

bbraams:
Thanks for the feedback. I've added the benchmark results to the PR description. As expected, the impact on the write path is minimal given the default page size and various compression schemes. I also ran the benchmarks on the read path, which also show minimal impact, however I'm not sure 100% the code path taken is representative of normal use. If we want to enable verification by default in the future we will need some more thorough testing.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumReadBenchmarks.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  @Setup(Level.Trial)
+  public void cleanup() {
+    pageChecksumDataGenerator.cleanup();
+  }
+
+  private void readFile(Path file, int nRows, boolean verifyChecksums, Blackhole blackhole)
+    throws IOException {
+    ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), file)
```
Fokko: Can we wrap a try-with-resource around this?

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumWriteBenchmarks.java
```diff
@@ -0,0 +1,136 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.benchmarks;
+
+import org.openjdk.jmh.annotations.*;
```
Fokko: Can we get rid of this wildcard imports?

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumWriteBenchmarks.java
```diff
@@ -0,0 +1,136 @@
...
+
+import static org.apache.parquet.benchmarks.BenchmarkConstants.*;
+import static org.apache.parquet.benchmarks.BenchmarkFiles.*;
```
Fokko: Can we get rid of these wildcard imports?

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/PageChecksumWriteBenchmarks.java
```diff
@@ -0,0 +1,136 @@
...
+
+import java.io.IOException;
+
+import static org.apache.parquet.hadoop.metadata.CompressionCodecName.*;
```
Fokko: Can we get rid of this wildcard imports?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -105,6 +109,8 @@ private ParquetProperties(WriterVersion writerVersion, int pageSize, int dictPag
     this.valuesWriterFactory = writerFactory;
     this.columnIndexTruncateLength = columnIndexMinMaxTruncateLength;
     this.pageRowCountLimit = pageRowCountLimit;
+
```
Fokko: Please remove this whitespace

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java
```diff
@@ -0,0 +1,563 @@
...
+      colBPage2Bytes[i] = (byte) (i - 100);
+    }
+
+    ParquetFileWriter writer =  new ParquetFileWriter(conf, schemaSimple, path,
+      ParquetWriter.DEFAULT_BLOCK_SIZE, ParquetWriter.MAX_PADDING_SIZE_DEFAULT);
+
+    writer.start();
+    writer.startBlock(numRecordsLargeFile);
+
+    CodecFactory codecFactory = new CodecFactory(conf, 1024 * 1024);
```
Fokko:
```suggestion
    CodecFactory codecFactory = new CodecFactory(conf, PAGE_SIZE);
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -119,14 +127,28 @@ public void writePage(BytesInput bytes,
...
+        crc.reset();
+        crc.update(compressedBytes.toByteArray());
+        parquetMetadataConverter.writeDataPageV1Header(
           (int)uncompressedSize,
           (int)compressedSize,
           valueCount,
           rlEncoding,
           dlEncoding,
           valuesEncoding,
+          (int) crc.getValue(),
```
wangyum, line 140: @gszadovszky Do we need to consider int overflow? For example:
```
crc.getValue() = 4169965210
(int) crc.getValue() = -125002086
(int) (crc.getValue() & 0x7FFFFFFF) = 2022481562
```

bbraams: @wangyum See the verification here https://github.com/apache/parquet-mr/pull/647/files#diff-8da24c84aef62e6e836d073938f7843d289785baaeddf446f3afeae6d4ef4b10R1176-R1182

PR #650, opened 2019-06-16
PARQUET-1599: Fix to-avro to respect the overwrite option

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-1599

### Tests

  - testToAvroCommandOverwriteExistentFile
  - testToAvroCommandOverwriteExistentFileWithoutOverwriteOption

### Commits

### Documentation


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ToAvroCommand.java
```diff
@@ -96,9 +97,8 @@ public int run() throws IOException {
 
     Path outPath = qualifiedPath(outputPath);
     FileSystem outFS = outPath.getFileSystem(getConf());
-    if (overwrite && outFS.exists(outPath)) {
```
nandorKollar: SchemaCommand has the same problem.

Instead of throwing an exception here if the file already exists, how about adding a new method to BaseCommand for no overwrite case, and using that when creating the file in this command (and in SchemaCommand)?

Something like this: a new createWithNoOverwrite, which calls the private create with an extra overwrite=false flag, and it will use fs.create with this parameter value. How does this sound? We can also remove this delete file section in this case, and just call createWithNoOverwrite when overwrite is true, and call create otherwise.

sekikn: Thank you for the review @nandorKollar, that sounds good to me.
I'll update the PR a bit later.

sekikn: @nandorKollar Sorry for my long silence, just addressed your comments :)

sekikn:
Rebased on master and resolve conflicts.

sekikn:
Rebased on master and resolved conflicts.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ToAvroCommand.java
```diff
@@ -92,27 +92,22 @@ public int run() throws IOException {
...
     long count = 0;
 
     DatumWriter<Record> datumWriter = new GenericDatumWriter<>(schema);
     try (DataFileWriter<Record> fileWriter = new DataFileWriter<>(datumWriter)) {
       fileWriter.setCodec(codecFactory);
-      try (DataFileWriter<Record> writer=fileWriter.create(projection, create(outputPath))) {
-        for (Record record : reader) {
-          writer.append(record);
-          count += 1;
+      try (OutputStream os = overwrite ?
```
Fokko: I think we can combine the two `try`'s here

sekikn: Sorry that I missed your comment @Fokko! Just updated the PR.

Fokko:
I've retriggered the failing build, it looked unrelated.

Fokko:
Thanks @sekikn !

PR #654, opened 2019-06-19
PARQUET-1550: CleanUtil does not work in Java 11

Fokko:

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references them in the PR title. For example, "PARQUET-1550: My Parquet PR"
  - https://issues.apache.org/jira/browse/PARQUET-1550

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/pom.xml
+++ b/parquet-hadoop/pom.xml
@@ -86,6 +86,11 @@
       <version>1.6</version>
     </dependency>
     <dependency>
+      <groupId>it.unimi.dsi</groupId>
+      <artifactId>fastutil</artifactId>
+      <version>${fastutil.version}</version>
+    </dependency>
+    <dependency>
       <groupId>com.github.rdblue</groupId>
       <artifactId>brotli-codec</artifactId>
       <version>${brotli-codec.version}</version>
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
@@ -27,8 +27,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /*
-Strongly inspired by:
-https://github.com/apache/tomcat/blob/master/java/org/apache/tomcat/util/buf/ByteBufferUtils.java
+ * A Helper class which use reflections to clean up DirectBuffer. It's implemented for
+ * better compatibility with both java8 and java9+, because the Cleaner class is moved to
+ * another place since java9+.
+ *
+ * Strongly inspired by:
+ * https://github.com/apache/tomcat/blob/master/java/org/apache/tomcat/util/buf/ByteBufferUtils.java
  */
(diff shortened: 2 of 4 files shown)


parquet-hadoop/pom.xml
```diff
@@ -85,6 +85,11 @@
       <artifactId>commons-pool</artifactId>
       <version>1.6</version>
     </dependency>
+    <dependency>
+      <groupId>it.unimi.dsi</groupId>
```
Fokko: The `parquet-hadoop` actually depends on the `fastutil` package, without this package I was unable to test the project in isolation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -26,45 +26,84 @@
...
+        | InvocationTargetException | SecurityException e) {
+        // Ignore
+      }
+    } else if (invokeCleanerMethod != null) {
+      try {
+        invokeCleanerMethod.invoke(unsafe, buf);
+      } catch (IllegalAccessException | IllegalArgumentException
+        | InvocationTargetException | SecurityException e) {
+        // Ignore
+        e.printStackTrace();
```
gszadovszky: I would not print anything to stderr. I would suggest using a logger and use it in both cases (`cleanMethod` or `invokeCleanerMethod`)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
```diff
@@ -26,45 +26,84 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * A Helper class which use reflections to clean up DirectBuffer. It's implemented for
- * better compatibility with both java8 and java9+, because the Cleaner class is moved to
- * another place since java9+.
+/*
```
gszadovszky: I think, it would be better to keep the javadoc comments with the usual content about the class and add the hint about the inspiration.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
Fokko: Good point, I've updated the code to write it to the logger.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/CleanUtil.java
Fokko: I've restored the javadocs.

gszadovszky: Failed to restore the one `*` that makes it javadoc :)

Fokko: Good catch, added the `*` :-)

PR #662, opened 2019-07-11
PARQUET-1624: Use Hadoop options for HadoopInputFiles.

rdblue:
This is a minor fix. When opening a HadoopInputFile, use its configuration for read options.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -590,7 +590,13 @@ public class ParquetFileReader implements Closeable {
    * @throws IOException if there is an error while opening the file
    */
   public static ParquetFileReader open(InputFile file) throws IOException {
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -590,7 +590,13 @@ public static ParquetFileReader open(Configuration conf, Path file, ParquetMetad
    * @throws IOException if there is an error while opening the file
    */
   public static ParquetFileReader open(InputFile file) throws IOException {
-    return new ParquetFileReader(file, ParquetReadOptions.builder().build());
+    ParquetReadOptions options;
```
gszadovszky: Maybe it worth some documentation that in one case (`open(InputFile)`) the hadoop config is used if the `InputFile` is a `HadoopInputFile` while in the other case (`open(InputFile, ParquetReadOptions)`) the hadoop config is omitted and only the specified `ParquetReadOptions` is used.

PR #672, opened 2019-08-29
PARQUET-1644: Clean up some benchmark code and docs.

RyanSkraba:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-1644

### Tests

- [X] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason: *The benchmarking module is used for test purposes*

### Commits

### Documentation


parquet-benchmarks/run.sh
```diff
@@ -21,10 +21,10 @@
 SCRIPT_PATH=$( cd "$(dirname "$0")" ; pwd -P )
 
 echo "Starting WRITE benchmarks"
-java -jar ${SCRIPT_PATH}/target/parquet-benchmarks.jar p*Write* "$@"
+java -jar ${SCRIPT_PATH}/target/parquet-benchmarks.jar org.apache.parquet.benchmarks.WriteBenchmarks "$@"
```
nandorKollar: With this change, `NestedNullWritingBenchmarks` won't be executed in this benchmark suite, but I think that's fine. @gszadovszky do you agree?

gszadovszky: I agree. When this script was created only `ReadBenchmarks` and `WriteBenchmarks` were exist. However, it might make sense to have a more descriptive name (e.g. `runReadWrite.sh`).

RyanSkraba:
Thanks for taking a look!  I'm going to make the run.sh script a bit more descriptive to run "feature sets" so that there's a standard way to run *all* or some benchmarks.  I don't want to go crazy with bash, but I think there's some value to have one entry point to running benchmarks in a standard way.

JMH isn't behaving as I would like for setting up and cleaning resources...  I have a couple more fixes to make and I'll push.

parquet-benchmarks/run.sh
RyanSkraba: I combined the two existing run scripts into one with some very limited functionality to run predefined "suites" by keyword, but also to run all benchmarks.

What do you think?

RyanSkraba:
For info, I made a change around general benchmarks setup/cleanup.

Before, the read files for `ReadBenchmark` were generated and handled differently than the read files for `PageChecksumReadBenchmarks`.  I tried to make them consistent across benchmarks using the following logic:

* When reading, ensure that any necessary file exists during `@Setup` but don't do any cleanup during the benchmark run.
* When writing, ensure that the output file *doesn't* exist during `@Setup`
* All cleanup needs to be done outside of the process running the benchmark.

As far as I can tell, this is the preferred way to use JMH when running "macro-benchmarks".

I didn't observe any actual functional differences after changing the setup strategy -- the total user time to run the `PageChecksumReadBenchmarks` went down, for example, but the measured operation time stayed the same.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ReadBenchmarks.java
```diff
@@ -47,35 +56,44 @@ private void read(Path parquetFile, int nRows, Blackhole blackhole) throws IOExc
...
+
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
   public void read1MRowsDefaultBlockAndPageSizeUncompressed(Blackhole blackhole)
           throws IOException
   {
     read(file_1M, ONE_MILLION, blackhole);
   }
 
-  @Benchmark
+  @Benchmark @BenchmarkMode(Mode.SingleShotTime)
```
gszadovszky: nit: I would put annotations to separate lines

PR #681, opened 2019-09-23
PARQUET-1576 Upgrade to Avro 1.9.1

iemejia:
Avro 1.9.1 fixes some [regression issues](https://github.com/apache/avro/releases/tag/release-1.9.1) from 1.9.0
I am reusing the same JIRA issue than the upgrade to 1.9.0 because this has not been yet released, so probably worth to upgrade the JIRA issue.


gszadovszky:
Usually reusing a jira is a bad idea. PARQUET-1576 headline is about Avro 1.9.0. I think, it would be more clear to have a separate jira about update Avro to 1.9.1.

iemejia:
Ok will do.

PR #683, opened 2019-09-23
PARQUET-1662 Upgrade Jackson to version 2.9.10

iemejia:
Jackson 2.9.10 addresses multiple CVE issues from previous Jackson versions, so we need to upgrade it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,8 +75,8 @@
     <maven-jar-plugin.version>2.4</maven-jar-plugin.version>
     <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
     <jackson.package>com.fasterxml.jackson</jackson.package>
-    <jackson.version>2.9.9</jackson.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -75,8 +75,7 @@
     <maven-jar-plugin.version>2.4</maven-jar-plugin.version>
     <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
     <jackson.package>com.fasterxml.jackson</jackson.package>
-    <jackson.version>2.9.9</jackson.version>
-    <jackson-databind.version>2.9.9.3</jackson-databind.version>
```
gszadovszky: Once, we had to separate the databind version because it was released separately. I would suggest keeping them separated and use the same version for both.

iemejia: ok, done now.

iemejia:
Kept `jackson-databind`.version separated as suggested. PTAL @gszadovszky

PR #696, opened 2019-10-30
PARQUET-1685: Truncate Min/Max for Statistics

shangxinli:

### Jira

PARQUET-1685: Truncate Min/Max for Statistics

### Tests

Added unit test
Write a Spark application test and it passed 

### Commits

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
@@ -103,6 +103,17 @@ public class BinaryStatistics extends Statistics<Binary> {
     return !hasNonNullValue() || ((min.length() + max.length()) < size);
   }
 
+  public boolean isSmallerThanWithTruncation(long size, int truncationLength) {
+    if (!hasNonNullValue()) {
+      return true;
+    }
+
+    int minTruncateLength = Math.min(min.length(), truncationLength);
+    int maxTruncateLength = Math.min(max.length(), truncationLength);
+
+    return minTruncateLength + maxTruncateLength < size;
+  }
+
   /**
    * @param min_value a min binary
    * @param max_value a max binary
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -44,6 +44,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.CorruptStatistics;
 import org.apache.parquet.ParquetReadOptions;
+import org.apache.parquet.column.ParquetProperties;
+import org.apache.parquet.column.statistics.BinaryStatistics;
 import org.apache.parquet.format.BsonType;
 import org.apache.parquet.format.CompressionCodec;
 import org.apache.parquet.format.DateType;
@@ -91,10 +93,12 @@ import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.column.EncodingStats;
 import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.internal.column.columnindex.BinaryTruncator;
 import org.apache.parquet.internal.column.columnindex.ColumnIndexBuilder;
 import org.apache.parquet.internal.column.columnindex.OffsetIndexBuilder;
 import org.apache.parquet.internal.hadoop.metadata.IndexReference;
 import org.apache.parquet.io.ParquetDecodingException;
+import org.apache.parquet.io.api.Binary;
 import org.apache.parquet.schema.ColumnOrder.ColumnOrderName;
 import org.apache.parquet.schema.GroupType;
 import org.apache.parquet.schema.MessageType;
@@ -120,11 +124,15 @@ public class ParquetMetadataConverter {
   private static final Logger LOG = LoggerFactory.getLogger(ParquetMetadataConverter.class);
   private static final LogicalTypeConverterVisitor LOGICAL_TYPE_ANNOTATION_VISITOR = new LogicalTypeConverterVisitor();
   private static final ConvertedTypeConverterVisitor CONVERTED_TYPE_CONVERTER_VISITOR = new ConvertedTypeConverterVisitor();
-
+  private final int statisticsTruncateLength;
   private final boolean useSignedStringMinMax;
 
   public ParquetMetadataConverter() {
     this(false);
+  }
+
+  public ParquetMetadataConverter(int statisticsTruncateLength) {
+    this(false, statisticsTruncateLength);
   }
 
   /**
@@ -141,7 +149,15 @@ public class ParquetMetadataConverter {
   }
 
   private ParquetMetadataConverter(boolean useSignedStringMinMax) {
+    this(useSignedStringMinMax, ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH);
+  }
+
+  private ParquetMetadataConverter(boolean useSignedStringMinMax, int statisticsTruncateLength) {
+    if (statisticsTruncateLength <= 0) {
+      throw new IllegalArgumentException("Truncate length should be greater than 0");
+    }
     this.useSignedStringMinMax = useSignedStringMinMax;
+    this.statisticsTruncateLength = statisticsTruncateLength;
   }
 
   // NOTE: this cache is for memory savings, not cpu savings, and is used to de-duplicate
@@ -464,7 +480,7 @@ public class ParquetMetadataConverter {
           columnMetaData.getFirstDataPageOffset());
       columnChunk.meta_data.dictionary_page_offset = columnMetaData.getDictionaryPageOffset();
       if (!columnMetaData.getStatistics().isEmpty()) {
-        columnChunk.meta_data.setStatistics(toParquetStatistics(columnMetaData.getStatistics()));
+        columnChunk.meta_data.setStatistics(toParquetStatistics(columnMetaData.getStatistics(), this.statisticsTruncateLength));
       }
       if (columnMetaData.getEncodingStats() != null) {
         columnChunk.meta_data.setEncoding_stats(convertEncodingStats(columnMetaData.getEncodingStats()));
@@ -582,18 +598,31 @@ public class ParquetMetadataConverter {
   }
 
   public static Statistics toParquetStatistics(
-      org.apache.parquet.column.statistics.Statistics stats) {
+    org.apache.parquet.column.statistics.Statistics stats) {
+    return toParquetStatistics(stats, ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH);
+  }
+
+  public static Statistics toParquetStatistics(
+      org.apache.parquet.column.statistics.Statistics stats, int truncateLength) {
     Statistics formatStats = new Statistics();
     // Don't write stats larger than the max size rather than truncating. The
     // rationale is that some engines may use the minimum value in the page as
     // the true minimum for aggregations and there is no way to mark that a
     // value has been truncated and is a lower bound and not in the page.
-    if (!stats.isEmpty() && stats.isSmallerThan(MAX_STATS_SIZE)) {
+    if (!stats.isEmpty() && withinLimit(stats, truncateLength)) {
       formatStats.setNull_count(stats.getNumNulls());
       if (stats.hasNonNullValue()) {
-        byte[] min = stats.getMinBytes();
-        byte[] max = stats.getMaxBytes();
+        byte[] min;
+        byte[] max;
 
+        if (stats instanceof BinaryStatistics && truncateLength != Integer.MAX_VALUE) {
+          BinaryTruncator truncator = BinaryTruncator.getTruncator(stats.type());
+          min = tuncateMin(truncator, truncateLength, stats.getMinBytes());
+          max = tuncateMax(truncator, truncateLength, stats.getMaxBytes());
+        } else {
+          min = stats.getMinBytes();
+          max = stats.getMaxBytes();
+        }
         // Fill the former min-max statistics only if the comparison logic is
         // signed so the logic of V1 and V2 stats are the same (which is
         // trivially true for equal min-max values)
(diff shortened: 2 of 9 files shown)


parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -78,8 +87,8 @@ public void mergeStatisticsMinMax(Statistics stats) {
    */
   @Override
   public void setMinMaxFromBytes(byte[] minBytes, byte[] maxBytes) {
-    max = Binary.fromReusedByteArray(maxBytes);
-    min = Binary.fromReusedByteArray(minBytes);
+    max = truncator.truncateMax(Binary.fromReusedByteArray(maxBytes), truncateLength);
```
gszadovszky: This method is invoked when we read the values from a file. It does not make sense truncating the values in the read path.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
```diff
@@ -110,8 +119,94 @@ public boolean isSmallerThan(long size) {
    */
   @Deprecated
   public void updateStats(Binary min_value, Binary max_value) {
-    if (comparator().compare(min, min_value) > 0) { min = min_value.copy(); }
-    if (comparator().compare(max, max_value) < 0) { max = max_value.copy(); }
+    if (comparator().compare(min, min_value) > 0) { min = truncator.truncateMin(min_value.copy(), truncateLength); }
```
gszadovszky: This method is invoked for every values to be written to the file. This is a huge overhead and useless. I would do the truncate just before writing to the file only once for each column chunk.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
shangxinli: Agree! Removed it.

parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java
shangxinli: Agree! Removed it.

shangxinli:
Yes. Moved it to ParquetMetadataConverter for the write path only.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -611,6 +624,18 @@ public static Statistics toParquetStatistics(
     return formatStats;
   }
 
+  public static void setStatisticsTruncateLength(int statisticsTruncateLength) {
```
gszadovszky: This is not correct this way. Writer has a constructor argument so you shall not set here statically. I would suggest creating a new `toParquetStatistics` method with this argument instead.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -274,6 +274,25 @@ public ParquetFileWriter(OutputFile file, MessageType schema, Mode mode,
...
+   * @param schema the schema of the data
+   * @param mode file creation mode
+   * @param rowGroupSize the row group size
+   * @param maxPaddingSize the maximum padding
+   * @param columnIndexTruncateLength the length which the min/max values in column indexes tried to be truncated to
+   * @param statisticsTruncateLength the length which the min/max values in row groups tried to be truncated to
+   * @param pageWriteChecksumEnabled whether to write out page level checksums
+   * @throws IOException if the file can not be created
+   */
+  public ParquetFileWriter(OutputFile file, MessageType schema, Mode mode,
```
gszadovszky: Please, check where the previous constructor is invoked and update those code parts as well.
In addition, if this change is targeted to the next release (1.11.0) and the previous constructor was not released yet, you may simply add the new argument to it. The key is to not break API compatibility between releases.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: Yes, I know that setting it here is a little hacky. The reasons for that are:  1) ParquetMetadataConverter is created as static inside ParquetFileWriter(don't' understand why). ParquetFileWriter has no chance to pass the parameter that it gets from it's contractor.  2) 'toParquetStatistics()' is defined as public static, so the constructor parameter doesn't wor for it. But you suggested rewriting my own toParquetStatistics(). The concern is code duplicate.   

The #2 is less concern that I can refactor the code a bit to reduce the duplicate. Any suggestions for #1? Maybe change metadataConverter to non-static? Concerns?

shangxinli: Just created another iteration by changing the metadataConverter to non-static and create my own  'toParquetStatistics'. Please have a look.

gszadovszky: Sorry for the late response. I didn't want to recommend refactoring the usage of `MetadataConverter` for your change. I've just ment to have the truncateLength as a parameter of `toParquetStatistics`. You may keep the original method header as well so you do not need to rewrite all the calls only the ones are required for this change.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
gszadovszky: This is because or faulty releasing process. 1.11.0 is not release yet. 1.11.0 tag will be moved to the finally released RC.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: Got it. Thanks.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
shangxinli: Cool. Let's just add one parameter instead of creating a new signature.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -580,6 +583,42 @@ private void testBinaryStats(StatsHelper helper) {
     Assert.assertTrue(roundTripStats.isEmpty());
   }
 
+  @Test
+  public void testBinaryStatsWithTruncation() {
+    int defaultTruncLen = ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH;
+    int[] validLengths = {1, 2, 16, 64, defaultTruncLen - 1, defaultTruncLen - 1};
```
gszadovszky: Why do you test `defaultTruncLen -1` twice? I guess, `defaultTrundLen` itself shall also be tested.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -580,6 +583,42 @@ private void testBinaryStats(StatsHelper helper) {
...
+    for (int len : validLengths) {
+      testBinaryStatsWithTruncation(len);
+    }
+
+    int[] invalidLengths = {-1, 0, defaultTruncLen + 1};
```
gszadovszky: `defaultTruncLen + 1` is invalid only because `ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH` is `Integer.MAX_VALUE` so it overflows. Meanwhile, semantically a larger value of the default one shall be supported.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -591,9 +609,17 @@ public static Statistics toParquetStatistics(
     if (!stats.isEmpty() && stats.isSmallerThan(MAX_STATS_SIZE)) {
```
gszadovszky: So, we are not truncating if the size is too large? It should be the other way around: check the size after the truncate.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -580,6 +583,42 @@ private void testBinaryStats(StatsHelper helper) {
...
+    for (int len : invalidLengths) {
+      try {
+        testBinaryStatsWithTruncation(len);
+      } catch (IllegalArgumentException e) {
+        // expected, nothing to do
+      }
+    }
+  }
+
+  private void testBinaryStatsWithTruncation(int truncateLen) {
```
gszadovszky: Please add test data where we would reach the size limit of statistics.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Good catch. That is a typo, I meant to test defaultTruncLen.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Make sense. I will make it as "-1, 0, Integer.MAX_VALUE + 1"

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
gszadovszky: `- 1` is still there. :)

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
gszadovszky: Still not fixed.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -580,6 +589,62 @@ private void testBinaryStats(StatsHelper helper) {
...
+  private void testBinaryStatsWithTruncation(int truncateLen, int minLen, int maxLen) {
+    BinaryStatistics stats = new BinaryStatistics();
+    byte[] min = generateRandomString("a", minLen).getBytes();
+    byte[] max = generateRandomString("b", maxLen).getBytes();
+    stats.updateStats(Binary.fromConstantByteArray(min));
+    stats.updateStats(Binary.fromConstantByteArray(max));
+    ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter(truncateLen);
+    org.apache.parquet.format.Statistics formatStats = metadataConverter.toParquetStatistics(stats);
+
+    if (minLen >= ParquetMetadataConverter.MAX_STATS_SIZE || maxLen >= ParquetMetadataConverter.MAX_STATS_SIZE) {
```
gszadovszky: `MAX_STATS_SIZE` is for the whole statistics so this should be `minLen + maxLen >= MAX_STATS_SIZE`

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -582,18 +595,31 @@ public EncodingStats convertEncodingStats(List<PageEncodingStats> stats) {
...
+    if (!stats.isEmpty() && withinLimit(stats, truncateLength)) {
       formatStats.setNull_count(stats.getNumNulls());
       if (stats.hasNonNullValue()) {
-        byte[] min = stats.getMinBytes();
-        byte[] max = stats.getMaxBytes();
-
+        byte[] min;
+        byte[] max;
+
+        if (stats instanceof BinaryStatistics && truncateLength != Integer.MAX_VALUE) {
```
gszadovszky: `truncateLength != MAX_VALUE` is special that the `truncator` handles anyway. I would not put it here.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Don't know how it is slipped. Fix it now.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Don't know how it is slipped. Fix it now.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: Got it.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Added it.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: My concern here is if we don't put it here, there will be some objects created in getTruncator() and tuncateMin/Max() although it will bail out eventually. The default value is Integer.MAX_VALUE. So it would be the most of cases.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: Good catch.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: fixed

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
shangxinli: added

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -580,6 +589,62 @@ private void testBinaryStats(StatsHelper helper) {
...
+      testBinaryStatsWithTruncation(len, (int) ParquetMetadataConverter.MAX_STATS_SIZE, 190);
+      testBinaryStatsWithTruncation(len, 280, (int) ParquetMetadataConverter.MAX_STATS_SIZE);
+      testBinaryStatsWithTruncation(len, (int) ParquetMetadataConverter.MAX_STATS_SIZE, (int) ParquetMetadataConverter.MAX_STATS_SIZE);
+    }
+
+    int[] invalidLengths = {-1, 0,  Integer.MAX_VALUE + 1};
+    for (int len : invalidLengths) {
+      try {
+        testBinaryStatsWithTruncation(len, 80, 20);
+      } catch (IllegalArgumentException e) {
```
gszadovszky: This test would not fail if no exception is thrown.
```suggestion
        Assert.fail("...");
      } catch (IllegalArgumentException e) {
```

PR #705, opened 2019-12-10
PARQUET-1622: Add implementation for BYTE_STREAM_SPLIT

martinradev:
The patch adds an implementation and tests for the
BYTE_STREAM_SPLIT encoding.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -33,10 +33,12 @@ import org.apache.parquet.column.page.PageWriteStore;
 import org.apache.parquet.column.values.ValuesWriter;
 import org.apache.parquet.column.values.bitpacking.DevNullValuesWriter;
 import org.apache.parquet.column.values.factory.DefaultValuesWriterFactory;
+import org.apache.parquet.column.values.factory.ValuesWriterFactory;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWriter;
-import org.apache.parquet.column.values.factory.ValuesWriterFactory;
 import org.apache.parquet.schema.MessageType;
+
+import static org.apache.parquet.bytes.BytesUtils.getWidthFromMaxInt;
 
 /**
  * This class represents all the configurable Parquet properties.
@@ -46,6 +48,7 @@ public class ParquetProperties {
   public static final int DEFAULT_PAGE_SIZE = 1024 * 1024;
   public static final int DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
   public static final boolean DEFAULT_IS_DICTIONARY_ENABLED = true;
+  public static final boolean DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED = false;
   public static final WriterVersion DEFAULT_WRITER_VERSION = WriterVersion.PARQUET_1_0;
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
@@ -95,11 +98,12 @@ public class ParquetProperties {
   private final int statisticsTruncateLength;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
+  private final boolean enableByteStreamSplit;
 
   private ParquetProperties(WriterVersion writerVersion, int pageSize, int dictPageSize, boolean enableDict, int minRowCountForPageSizeCheck,
                             int maxRowCountForPageSizeCheck, boolean estimateNextSizeCheck, ByteBufferAllocator allocator,
                             ValuesWriterFactory writerFactory, int columnIndexMinMaxTruncateLength, int pageRowCountLimit,
-                            boolean pageWriteChecksumEnabled, int statisticsTruncateLength) {
+                            boolean pageWriteChecksumEnabled, int statisticsTruncateLength, boolean enableByteStreamSplit) {
     this.pageSizeThreshold = pageSize;
     this.initialSlabSize = CapacityByteArrayOutputStream
       .initialSlabSizeHeuristic(MIN_SLAB_SIZE, pageSizeThreshold, 10);
@@ -116,6 +120,7 @@ public class ParquetProperties {
     this.statisticsTruncateLength = statisticsTruncateLength;
     this.pageRowCountLimit = pageRowCountLimit;
     this.pageWriteChecksumEnabled = pageWriteChecksumEnabled;
+    this.enableByteStreamSplit = enableByteStreamSplit;
   }
 
   public ValuesWriter newRepetitionLevelWriter(ColumnDescriptor path) {
@@ -170,6 +175,10 @@ public class ParquetProperties {
 
   public boolean isEnableDictionary() {
     return enableDictionary;
+  }
+
+  public boolean isByteStreamSplitEnabled() {
+      return enableByteStreamSplit;
   }
 
   public ByteBufferAllocator getAllocator() {
@@ -242,6 +251,7 @@ public class ParquetProperties {
     private int statisticsTruncateLength = DEFAULT_STATISTICS_TRUNCATE_LENGTH;
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
     private boolean pageWriteChecksumEnabled = DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED;
+    private boolean enableByteStreamSplit = DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED;
 
     private Builder() {
     }
@@ -258,6 +268,7 @@ public class ParquetProperties {
       this.allocator = toCopy.allocator;
       this.pageRowCountLimit = toCopy.pageRowCountLimit;
       this.pageWriteChecksumEnabled = toCopy.pageWriteChecksumEnabled;
+      this.enableByteStreamSplit = toCopy.enableByteStreamSplit;
     }
 
     /**
@@ -281,6 +292,11 @@ public class ParquetProperties {
      */
     public Builder withDictionaryEncoding(boolean enableDictionary) {
       this.enableDict = enableDictionary;
+      return this;
+    }
+
+    public Builder withByteStreamSplitEncoding(boolean enableByteStreamSplit) {
+      this.enableByteStreamSplit = enableByteStreamSplit;
       return this;
     }
 
@@ -366,7 +382,7 @@ public class ParquetProperties {
         new ParquetProperties(writerVersion, pageSize, dictPageSize,
           enableDict, minRowCountForPageSizeCheck, maxRowCountForPageSizeCheck,
           estimateNextSizeCheck, allocator, valuesWriterFactory, columnIndexTruncateLength,
-          pageRowCountLimit, pageWriteChecksumEnabled, statisticsTruncateLength);
+          pageRowCountLimit, pageWriteChecksumEnabled, statisticsTruncateLength, enableByteStreamSplit);
       // we pass a constructed but uninitialized factory to ParquetProperties above as currently
       // creation of ValuesWriters is invoked from within ParquetProperties. In the future
       // we'd like to decouple that and won't need to pass an object to properties and then pass the
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
@@ -1,0 +1,99 @@
+/* 
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import org.apache.parquet.bytes.ByteBufferInputStream;
+import org.apache.parquet.column.values.ValuesReader;
+import org.apache.parquet.io.ParquetDecodingException;
+
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ByteStreamSplitValuesReader extends ValuesReader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ByteStreamSplitValuesReader.class);
+  private final int numStreams;
+  private final int elementSizeInBytes;
+  private byte[] byteStreamData;
+  private int indexInStream;
+  private int numberOfValues;
+
+  protected ByteStreamSplitValuesReader(int elementSizeInBytes) {
+    this.numStreams = elementSizeInBytes;
+    this.elementSizeInBytes = elementSizeInBytes;
+    this.indexInStream = 0;
+    this.numberOfValues = 0;
+  }
+
+  protected void gatherElementDataFromStreams(byte[] gatheredData)
+          throws IOException, ArrayIndexOutOfBoundsException {
+    if (gatheredData.length != numStreams) {
+      throw new IOException("gatherData buffer is not of the expected size.");
+    }
+    if (indexInStream >= numberOfValues) {
+      throw new ArrayIndexOutOfBoundsException("Byte-stream data was already exhausted.");
+    }
+    for (int i = 0; i < numStreams; ++i) {
+      gatheredData[i] = byteStreamData[i * numberOfValues + indexInStream];
+    }
+    ++indexInStream;
+  }
+
+  @Override
+  public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IOException {
+    LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
+
+    if (valueCount * elementSizeInBytes > stream.available()) {
+      throw new IOException("Stream does not contain enough data for the given number of values.");
+    }
+
+    /* Allocate buffer for all of the byte stream data */
+    final int totalSizeInBytes = valueCount * numStreams;
+    byteStreamData = new byte[totalSizeInBytes];
+
+    /* Eagerly read the data for each stream */
+    final int numRead = stream.read(byteStreamData, 0, totalSizeInBytes);
+    if (numRead != totalSizeInBytes) {
+      String errorMessage = String.format("Failed to read requested number of bytes. Expected: %d. Read %d.",
+              totalSizeInBytes, numRead);
+      throw new IOException(errorMessage);
+    }
+
+    indexInStream = 0;
+    numberOfValues = valueCount;
+  }
+
+  @Override
+  public void skip() {
+    skip(1);
+  }
+
+  @Override
+  public void skip(int n) {
+    if (indexInStream + n > numberOfValues || n < 0) {
+      String errorMessage = String.format(
+              "Cannot skip this many elements. Current index: %d. Skip %d. Total number of elements: %d",
+              indexInStream, n, numberOfValues);
+      throw new ParquetDecodingException(errorMessage);
+    }
+    indexInStream += n;
+  }
+
+}
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
@@ -29,55 +29,56 @@ import org.slf4j.LoggerFactory;
 public abstract class ByteStreamSplitValuesReader extends ValuesReader {
 
   private static final Logger LOG = LoggerFactory.getLogger(ByteStreamSplitValuesReader.class);
-  private final int numStreams;
   private final int elementSizeInBytes;
   private byte[] byteStreamData;
   private int indexInStream;
-  private int numberOfValues;
+  private int valuesCount;
 
   protected ByteStreamSplitValuesReader(int elementSizeInBytes) {
-    this.numStreams = elementSizeInBytes;
     this.elementSizeInBytes = elementSizeInBytes;
     this.indexInStream = 0;
-    this.numberOfValues = 0;
+    this.valuesCount = 0;
   }
 
   protected void gatherElementDataFromStreams(byte[] gatheredData)
-          throws IOException, ArrayIndexOutOfBoundsException {
-    if (gatheredData.length != numStreams) {
-      throw new IOException("gatherData buffer is not of the expected size.");
+          throws ParquetDecodingException {
+    if (gatheredData.length != elementSizeInBytes) {
+      throw new ParquetDecodingException("gatherData buffer is not of the expected size.");
     }
-    if (indexInStream >= numberOfValues) {
-      throw new ArrayIndexOutOfBoundsException("Byte-stream data was already exhausted.");
+    if (indexInStream >= valuesCount) {
+      throw new ParquetDecodingException("Byte-stream data was already exhausted.");
     }
-    for (int i = 0; i < numStreams; ++i) {
-      gatheredData[i] = byteStreamData[i * numberOfValues + indexInStream];
+    for (int i = 0; i < elementSizeInBytes; ++i) {
+      gatheredData[i] = byteStreamData[i * valuesCount + indexInStream];
     }
     ++indexInStream;
   }
 
   @Override
-  public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IOException {
+  public void initFromPage(int valuesCount, ByteBufferInputStream stream) throws ParquetDecodingException, IOException {
     LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
 
-    if (valueCount * elementSizeInBytes > stream.available()) {
-      throw new IOException("Stream does not contain enough data for the given number of values.");
+    if (valuesCount * elementSizeInBytes > stream.available()) {
+      String errorMessage = String.format(
+              "Stream does not contain enough data for the given number of values. Num values: %d. Element size: %d Bytes available: %d",
+              valuesCount, elementSizeInBytes, stream.available());
+      throw new ParquetDecodingException(errorMessage);
     }
 
-    /* Allocate buffer for all of the byte stream data */
-    final int totalSizeInBytes = valueCount * numStreams;
+    // Allocate buffer for all of the byte stream data.
+    final int totalSizeInBytes = valuesCount * elementSizeInBytes;
     byteStreamData = new byte[totalSizeInBytes];
 
-    /* Eagerly read the data for each stream */
+    // Eagerly read the data for each stream.
     final int numRead = stream.read(byteStreamData, 0, totalSizeInBytes);
     if (numRead != totalSizeInBytes) {
       String errorMessage = String.format("Failed to read requested number of bytes. Expected: %d. Read %d.",
               totalSizeInBytes, numRead);
-      throw new IOException(errorMessage);
+      throw new ParquetDecodingException(errorMessage);
     }
 
     indexInStream = 0;
-    numberOfValues = valueCount;
+    this.valuesCount = valuesCount;
   }
 
   @Override
@@ -87,10 +88,10 @@ public abstract class ByteStreamSplitValuesReader extends ValuesReader {
 
   @Override
   public void skip(int n) {
-    if (indexInStream + n > numberOfValues || n < 0) {
+    if (n < 0 || indexInStream + n > valuesCount) {
       String errorMessage = String.format(
               "Cannot skip this many elements. Current index: %d. Skip %d. Total number of elements: %d",
-              indexInStream, n, numberOfValues);
+              indexInStream, n, valuesCount);
       throw new ParquetDecodingException(errorMessage);
     }
     indexInStream += n;
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
@@ -1,0 +1,51 @@
+/* 
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import org.apache.parquet.io.ParquetDecodingException;
+
+import java.io.IOException;
+
+public class ByteStreamSplitValuesReaderForDouble extends ByteStreamSplitValuesReader {
+
+  private byte[] valueByteBuffer;
+
+  public ByteStreamSplitValuesReaderForDouble() {
+    super(Double.BYTES);
+    valueByteBuffer = new byte[Double.BYTES];
+  }
+
+  @Override
+  public double readDouble() {
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      long value = (long)(valueByteBuffer[0] & 0xFF) |
+                   ((long)(valueByteBuffer[1] & 0xFF) << 8) |
+                   ((long)(valueByteBuffer[2] & 0xFF) << 16) |
+                   ((long)(valueByteBuffer[3] & 0xFF) << 24) |
+                   ((long)(valueByteBuffer[4] & 0xFF) << 32) |
+                   ((long)(valueByteBuffer[5] & 0xFF) << 40) |
+                   ((long)(valueByteBuffer[6] & 0xFF) << 48) |
+                   ((long)(valueByteBuffer[7] & 0xFF) << 56);
+      return Double.longBitsToDouble(value);
+    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
+      throw new ParquetDecodingException("Could not read double.", ex);
+    }
+  }
+}
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
@@ -18,13 +18,12 @@
  */
 package org.apache.parquet.column.values.bytestreamsplit;
 
-import org.apache.parquet.io.ParquetDecodingException;
-
-import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 
 public class ByteStreamSplitValuesReaderForDouble extends ByteStreamSplitValuesReader {
 
-  private byte[] valueByteBuffer;
+  private final byte[] valueByteBuffer;
 
   public ByteStreamSplitValuesReaderForDouble() {
     super(Double.BYTES);
@@ -33,19 +32,7 @@ public class ByteStreamSplitValuesReaderForDouble extends ByteStreamSplitValuesR
 
   @Override
   public double readDouble() {
-    try {
-      gatherElementDataFromStreams(valueByteBuffer);
-      long value = (long)(valueByteBuffer[0] & 0xFF) |
-                   ((long)(valueByteBuffer[1] & 0xFF) << 8) |
-                   ((long)(valueByteBuffer[2] & 0xFF) << 16) |
-                   ((long)(valueByteBuffer[3] & 0xFF) << 24) |
-                   ((long)(valueByteBuffer[4] & 0xFF) << 32) |
-                   ((long)(valueByteBuffer[5] & 0xFF) << 40) |
-                   ((long)(valueByteBuffer[6] & 0xFF) << 48) |
-                   ((long)(valueByteBuffer[7] & 0xFF) << 56);
-      return Double.longBitsToDouble(value);
-    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
-      throw new ParquetDecodingException("Could not read double.", ex);
-    }
+    gatherElementDataFromStreams(valueByteBuffer);
+    return ByteBuffer.wrap(valueByteBuffer).order(ByteOrder.LITTLE_ENDIAN).getDouble();
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
@@ -18,8 +18,7 @@
  */
 package org.apache.parquet.column.values.bytestreamsplit;
 
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
+import org.apache.parquet.bytes.BytesUtils;
 
 public class ByteStreamSplitValuesReaderForDouble extends ByteStreamSplitValuesReader {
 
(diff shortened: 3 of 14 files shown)


martinradev:
Note that this patch depends on an unreleased version of parquet-format where the BYTE_STREAM_SPLIT encoding was added.
I decided to jump ahead, start a PR and address whatever issues there are in the implementation in advance.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderTest.java
```diff
@@ -0,0 +1,223 @@
...
+      float f = reader.readFloat();
+      assertEquals(2.25f, f, 0.0f);
+      boolean exceptionWasThrown;
+      try {
+        reader.readFloat();
+        exceptionWasThrown = false;
+      } catch (ParquetDecodingException ex) {
+        exceptionWasThrown = true;
+      }
+      assertEquals(true, exceptionWasThrown);
```
gszadovszky: Suggestion:
If you are expecting an exception, you might specify it as an argument of the Test annotation. This way you do not have to try-catch and do any additional assertions.
If you would like to verify that the exception is thrown from the exact code line you might use try-catch and simply `Assert.fail(msg)` in the try block after the line you are testing.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriterTest.java
```diff
@@ -0,0 +1,167 @@
...
+              capacity, capacity, new DirectByteBufferAllocator());
+    }
+
+    @Test
+    public void TestSingleElement() throws Exception {
+      final double value = 23.6718811111112;
+      ByteStreamSplitValuesWriter.DoubleByteStreamSplitValuesWriter writer = getWriter(1);
+      writer.writeDouble(value);
+
+      // Check that the buffer size is exactly the size of one float in bytes.
```
gszadovszky:
```suggestion
      // Check that the buffer size is exactly the size of one double in bytes.
```

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesEndToEndTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    reader.initFromPage(numElements, ByteBufferInputStream.wrap(input.toByteBuffer()));
+
+    for (float expectedValue : values) {
+      float newValue = reader.readFloat();
+      assertEquals(expectedValue, newValue, 0.0f);
+    }
+
+    writer.reset();
+    writer.close();
```
gszadovszky: I would suggest invoking reset/close in the `finally` block after putting the whole code in a `try` block. This way the writer will be reset and closed even if an exception is occurred during the execution.

pom.xml
```diff
@@ -82,7 +82,7 @@
     <hadoop1.version>1.2.1</hadoop1.version>
     <cascading.version>2.7.1</cascading.version>
     <cascading3.version>3.1.2</cascading3.version>
-    <parquet.format.version>2.7.0</parquet.format.version>
+    <parquet.format.version>2.8.0-SNAPSHOT</parquet.format.version>
```
gszadovszky: Referencing a snapshot dependency is not allowed on master and won't work anyway. After releasing the new parquet-format it will be available under the version number `2.8.0`. Until then, we won't be able to push this change.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -92,11 +93,12 @@ public static WriterVersion fromString(String name) {
   private final int statisticsTruncateLength;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
+  private final boolean enableByteStreamSplit;
 
   private ParquetProperties(WriterVersion writerVersion, int pageSize, int dictPageSize, boolean enableDict, int minRowCountForPageSizeCheck,
                             int maxRowCountForPageSizeCheck, boolean estimateNextSizeCheck, ByteBufferAllocator allocator,
                             ValuesWriterFactory writerFactory, int columnIndexMinMaxTruncateLength, int pageRowCountLimit,
-                            boolean pageWriteChecksumEnabled, int statisticsTruncateLength) {
+                            boolean pageWriteChecksumEnabled, int statisticsTruncateLength, boolean enableByteStreamSplit) {
```
shangxinli: While we keep adding, the parameter list becomes so long. It is not just because of the one you added, so it is OK for this change. But I think we need to refactor it later. I created PARQUET-1764 to address it.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import org.apache.parquet.io.ParquetDecodingException;
+
+import java.io.IOException;
+
+public class ByteStreamSplitValuesReaderForFloat extends ByteStreamSplitValuesReader {
+
+  private static final int FLOAT_SIZE_IN_BYTES = 4;
```
shangxinli: You can use Float.BYTES like this "FLOAT_SIZE_IN_BYTES = Float.BYTES"

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import org.apache.parquet.io.ParquetDecodingException;
+
+import java.io.IOException;
+
+public class ByteStreamSplitValuesReaderForDouble extends ByteStreamSplitValuesReader {
+
+  private static final int DOUBLE_SIZE_IN_BYTES = 8;
```
shangxinli: You can use Double.BYTES like DOUBLE_SIZE_IN_BYTES = Double.BYTES.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+  private byte[] valueByteBuffer;
+
+  public ByteStreamSplitValuesReaderForFloat() {
+    super(ByteStreamSplitValuesReaderForFloat.FLOAT_SIZE_IN_BYTES);
```
shangxinli: Since this is inside the same class, I think just "super(FLOAT_SIZE_IN_BYTES)" should work;

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(ByteStreamSplitValuesReader.class);
+  private final int numStreams;
+  private final int elementSizeInBytes;
+  private byte[] byteStreamData;
+  private int indexInStream;
+  private int numberOfValues;
+
+  protected ByteStreamSplitValuesReader(int elementSizeInBytes) {
+    numStreams = elementSizeInBytes;
```
shangxinli: Please add "this." to numStreams, indexInStream, numberOfValues to be consistent for all the fields.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -0,0 +1,53 @@
...
+      long b3 = valueByteBuffer[2] & 0xFF;
+      long b4 = valueByteBuffer[3] & 0xFF;
+      long b5 = valueByteBuffer[4] & 0xFF;
+      long b6 = valueByteBuffer[5] & 0xFF;
+      long b7 = valueByteBuffer[6] & 0xFF;
+      long b8 = valueByteBuffer[7] & 0xFF;
+      long value = b1 | (b2 << 8) | (b3 << 16) | (b4 << 24) | (b5 << 32) | (b6 << 40) | (b7 << 48) | (b8 << 56);
+      return Double.longBitsToDouble(value);
+    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
+      throw new ParquetDecodingException("Could not read double.");
```
shangxinli: It would be hard to debug the error especially when it throws in production env, when we just show a simple string "Could not read double.". I suggest to include the 'ex' itself and wrap it up with ParquetDecodingException so that the exception can have more information. It is also recommended to print out more information of the data stream itself so that it is easy to troubleshoot the error.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      int b1 = valueByteBuffer[0] & 0xFF;
+      int b2 = valueByteBuffer[1] & 0xFF;
+      int b3 = valueByteBuffer[2] & 0xFF;
+      int b4 = valueByteBuffer[3] & 0xFF;
+      int value = b1 | (b2 << 8) | (b3 << 16) | (b4 << 24);
+      return Float.intBitsToFloat(value);
+    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
+      throw new ParquetDecodingException("Could not read float.");
```
shangxinli: Same comments as above.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+  public ByteStreamSplitValuesReaderForFloat() {
+    super(ByteStreamSplitValuesReaderForFloat.FLOAT_SIZE_IN_BYTES);
+    valueByteBuffer = new byte[FLOAT_SIZE_IN_BYTES];
+  }
+
+  @Override
+  public float readFloat() {
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      int b1 = valueByteBuffer[0] & 0xFF;
```
shangxinli: Why not name it from 0, like b0, b1, b2, b3 so that it matches the index of the array?

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+      int b2 = valueByteBuffer[1] & 0xFF;
+      int b3 = valueByteBuffer[2] & 0xFF;
+      int b4 = valueByteBuffer[3] & 0xFF;
+      int value = b1 | (b2 << 8) | (b3 << 16) | (b4 << 24);
```
shangxinli: You can save 4 intermediate int variables in the stack by writing like followings. Also, please put the most significant bytes from left to right, so that it is easy to understand. 

((valueByteBuffer[4] & 0xFF) << 24) | 
            ((valueByteBuffer[3] & 0xFF) << 16) | 
            ((valueByteBuffer[2] & 0xFF) << 8 ) | 
            ((valueByteBuffer[1] & 0xFF) << 0 );

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import org.apache.parquet.io.ParquetDecodingException;
+
+import java.io.IOException;
+
+public class ByteStreamSplitValuesReaderForFloat extends ByteStreamSplitValuesReader {
+
+  private static final int FLOAT_SIZE_IN_BYTES = 4;
+  private byte[] valueByteBuffer;
```
shangxinli: Usually, the type of the variable shouldn't be in the name. You can just name it as value or something.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -0,0 +1,53 @@
...
+  public ByteStreamSplitValuesReaderForDouble() {
+    super(ByteStreamSplitValuesReaderForDouble.DOUBLE_SIZE_IN_BYTES);
+    valueByteBuffer = new byte[DOUBLE_SIZE_IN_BYTES];
+  }
+
+  @Override
+  public double readDouble() {
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      long b1 = valueByteBuffer[0] & 0xFF;
```
shangxinli: Same comments as ByteStreamSplitValuesReaderForFloat

gszadovszky:
FYI: parquet-format 2.8.0 is released.

martinradev:
I am a bit overwhelmed with other tasks but I will get back to this in the upcoming weekend.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV1ValuesWriterFactory.java
```diff
@@ -100,11 +101,17 @@ private ValuesWriter getInt96ValuesWriter(ColumnDescriptor path) {
   }
 
   private ValuesWriter getDoubleValuesWriter(ColumnDescriptor path) {
+    if (this.parquetProperties.isByteStreamSplitEnabled()) {
```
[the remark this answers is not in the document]
martinradev: I changed it so that ByteStreamSplit is only used if dictionary is not enabled and the bytestreamsplit is enabled.
The BYTE_STREAM_SPLIT encoding only makes sense when the floating-point data has high enough entropy and also it has to combined with a compressor.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesEndToEndTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.apache.parquet.bytes.ByteBufferInputStream;
+import org.apache.parquet.bytes.BytesInput;
+import org.apache.parquet.bytes.DirectByteBufferAllocator;
+import org.junit.Test;
+
+public class ByteStreamSplitValuesEndToEndTest {
+
+  @Test
+  public void TestFloatPipeline() throws Exception {
```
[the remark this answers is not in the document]
martinradev: I think I addressed this in all places.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderTest.java
martinradev: Done

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriterTest.java
martinradev: Done

pom.xml
martinradev: Bumped to 2.8.0 in this patch. Hopefully not an issue that it's not in a separate patch.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesEndToEndTest.java
martinradev: I addressed this in my patch. I was thinking that if an exception is thrown, then this should be still fine since the garbage collector would destroy the writer and the stream would be closed.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: Changed to Float.BYTES everywhere instead of introducing a new name.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
martinradev: Changed to Double.BYTES everywhere instead of introducing a new name.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: Refactored.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
martinradev: Done

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
martinradev: Passed the exception object to ParquetDecodingException

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: Passed the exception object to ParquetDecodingException

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: Refactored to remove the temporary variables.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: I did not examine the bytecode and I am not familiar with Java, but wouldn't the static compiler optimize this and the jit compiler do further optimizations to guarantee that the values are actually in registers.

martinradev: Anyway, changed it as you recommended.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: No longer relevant for the current version of the patch.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
martinradev: Done.

martinradev:
Hey @gszadovszky and @shangxinli , thanks for the reviews. I updated the patch.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV1ValuesWriterFactory.java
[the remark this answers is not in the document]
gszadovszky: Dictionary encoding works the following way. Dictionary encodes the incoming values until all the values are written or if the dictionary gets full (there is a size limit) then the actual page is re-encoded to the fallback encoder and then all the left values are encoded with it. It means that some (or 0) pages are dictionary encoded and some (or 0) are encoded with the fallback encoder.
So, I think BYTE_STREAM_SPLIT should be used as the fallback encoder instead of the plain encoder in case of the related property is true.

parquet-column/src/test/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesEndToEndTest.java
gszadovszky: As per the java spec it is not guaranteed that the method `finalize` is invoked when the object is garbage collected. So, one shall ensure that the resources occupied outside of the JVM (e.g. I/O streams) are freed at some point (before the references are lost).

gszadovszky:
@martinradev, please either rebase your branch or resolve the conflicts here.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV1ValuesWriterFactory.java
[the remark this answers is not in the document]
martinradev: I think it would be better if byte_stream_split is used when dictionary is not enabled and byte_stream_split is requested, not as a fallback to dictionary.
In my evaluation byte_stream_split made sense for cases when the data has high enough entropy: ~15bits and higher. This included almost all samples of scientific data from our gathered data set.
I also did an artificial test to compare the Dictionary encoding with this one as the entropy of the data increases: https://github.com/martinradev/arrow-fp-compression-bench/blob/master/entropy_ratio.jpeg
It is noticeable that the Dictionary encoder falls back to the plain encoder around 17 bits of entropy but that the byte_stream_split encoding is superior for data with less entropy.
Also, the byte_stream_split encoding/decoding is essentially a scatter/gather operation, so it's as fast as memcpy. The dictionary encoding implementation is not as fast as memcpy:
https://github.com/martinradev/arrow-fp-compression-bench/blob/master/entropy_write_speed.jpeg
Note that the latter plot is not my optimized simd implementation.
I do not know how the results would look for the java implementation.

So, I think that the control should be left in the user's hands and I think it makes sense to sanitize building a parquet writer to disallow requesting both dictionary and the byte_stream_split encoding.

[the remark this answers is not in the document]
gszadovszky: I have some problems with this approach:
* parquet-mr tries to figure out the correct encoding. It might not be able to always choose the best but the idea of dictionary+fallback follows this approach. If the dictionary encoding is not good enough (because of the high entropy) it will fallback to another encoding. The fallback might happen even at the first page meaning that no dictionary encoding takes place for that column. It seems that byte_stream_split+compression is always a better choice than plain+compression so why would we use it?
* The fallback happens because of the dictionary page gets full. The size of the dictionary page is configurable so the entropy when the fallback happens is not a fix border and is different by default for float and double.
* Currently, you are not able to set dictionary encoding (or anything else) separately for the different columns. So, following your approach you have to decide if you want byte_stream_split encoding for your fp column or you want dictionary encoding for your string column but you cannot have both.

I would't say I know the best approach here. It might be a good idea to propose a new framework and configuration to allow the user to set the exact encoding/compression etc. for each column but it is a much bigger issue.
If we agree on that byte_stream_split is always better than plain, let's simply drop using plain encoding and use byte_stream_split in its place. If you would like to allow the user to enforce byte_stream_split over dictionary in all cases then you can have a property for that but what if you have more than one fp column with different entropy?

pom.xml
gszadovszky: 2.8.0 was mostly released for byte_stream_split so I think it's fine to upgrade the version in this patch. :)

shangxinli:
@gszadovszky, @martinradev, I will get back to you in a week or so for the comments.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -0,0 +1,51 @@
...
+  public ByteStreamSplitValuesReaderForDouble() {
+    super(Double.BYTES);
+    valueByteBuffer = new byte[Double.BYTES];
+  }
+
+  @Override
+  public double readDouble() {
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      long value = (long)(valueByteBuffer[0] & 0xFF) |
```
shangxinli: I think ByteBuffer.wrap(bytes).order().getDouble() should can do the work. You don't need to reimplement it.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -0,0 +1,47 @@
...
+  public ByteStreamSplitValuesReaderForFloat() {
+    super(Float.BYTES);
+    valueByteBuffer = new byte[Float.BYTES];
+  }
+
+  @Override
+  public float readFloat() {
+    try {
+      gatherElementDataFromStreams(valueByteBuffer);
+      int value = (int)(valueByteBuffer[0] & 0xFF) |
```
shangxinli: Same comments as above.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+  protected final int numStreams;
+  protected final int elementSizeInBytes;
+  private CapacityByteArrayOutputStream[] byteStreams;
+
+  public ByteStreamSplitValuesWriter(int elementSizeInBytes, int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+    this.numStreams = elementSizeInBytes;
+    this.elementSizeInBytes = elementSizeInBytes;
+    this.byteStreams = new CapacityByteArrayOutputStream[elementSizeInBytes];
+
+    // Round-up the capacity hint.
```
shangxinli: I see you put comments in both ways of // or /* */.  Please use just one way, maybe '//'.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(ByteStreamSplitValuesReader.class);
+  private final int numStreams;
+  private final int elementSizeInBytes;
+  private byte[] byteStreamData;
+  private int indexInStream;
+  private int numberOfValues;
+
+  protected ByteStreamSplitValuesReader(int elementSizeInBytes) {
+    this.numStreams = elementSizeInBytes;
```
shangxinli: Does it make sense to use two final variables but having the same value? It adds complexity.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.parquet.io.ParquetEncodingException;
+
+public abstract class ByteStreamSplitValuesWriter extends ValuesWriter {
+
+  protected final int numStreams;
+  protected final int elementSizeInBytes;
+  private CapacityByteArrayOutputStream[] byteStreams;
+
+  public ByteStreamSplitValuesWriter(int elementSizeInBytes, int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+    this.numStreams = elementSizeInBytes;
```
shangxinli: Please throw an exception if elementSizeInBytes == 0 because you are going to use it to divide.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+  @Override
+  public void close() {
+    for (CapacityByteArrayOutputStream stream : byteStreams) {
+      stream.close();
+    }
+  }
+
+  protected void scatterBytes(byte[] bytes) {
+    if (bytes.length != this.numStreams) {
+      throw new ParquetEncodingException("Number of bytes doesn't match the number of streams.");
```
shangxinli: Add the value of bytes.length  and this.numStreams into the excepton

martinradev: Done.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
martinradev: Done, Changed to <= 0

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
martinradev: Changed it to one.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
martinradev: Changed to //

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
martinradev: Done.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
martinradev: Changed it but I didn't modify the tests, which test whether the Encoding or Decoding paths produce correct results, to rely on the ByteBuffer serialization/deserialization.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Override
+  public void skip() {
+    skip(1);
+  }
+
+  @Override
+  public void skip(int n) {
+    if (indexInStream + n > numberOfValues || n < 0) {
```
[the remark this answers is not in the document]
martinradev: I changed it but I'm curious why this is the case.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -0,0 +1,99 @@
...
+    /* Eagerly read the data for each stream */
+    final int numRead = stream.read(byteStreamData, 0, totalSizeInBytes);
+    if (numRead != totalSizeInBytes) {
+      String errorMessage = String.format("Failed to read requested number of bytes. Expected: %d. Read %d.",
+              totalSizeInBytes, numRead);
+      throw new IOException(errorMessage);
+    }
+
+    indexInStream = 0;
+    numberOfValues = valueCount;
```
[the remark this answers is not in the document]
martinradev: Done.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    ++indexInStream;
+  }
+
+  @Override
+  public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IOException {
+    LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
+
+    if (valueCount * elementSizeInBytes > stream.available()) {
+      throw new IOException("Stream does not contain enough data for the given number of values.");
```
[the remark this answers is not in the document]
martinradev: Done

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+      long vAsLong = Double.doubleToLongBits(v);
+      byte[] bytes = new byte[DOUBLE_SIZE_IN_BYTES];
+      for (int i = 0; i < DOUBLE_SIZE_IN_BYTES; ++i) {
+        byte b = (byte) ((vAsLong >> (i * 8)) & 0xFF);
+        bytes[i] = b;
+      }
+      super.scatterBytes(bytes);
+    }
+
+    public String memUsageString(String prefix) {
```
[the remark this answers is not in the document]
martinradev: Added the override annotation. It's an abstract method.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static final int DOUBLE_SIZE_IN_BYTES = 8;
+
+    public DoubleByteStreamSplitValuesWriter(int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+      super(DOUBLE_SIZE_IN_BYTES, initialCapacity, pageSize, allocator);
+    }
+
+    @Override
+    public void writeDouble(double v) {
+      long vAsLong = Double.doubleToLongBits(v);
```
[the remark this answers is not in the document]
martinradev: Done.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    public String memUsageString(String prefix) {
+      return String.format("%s FloatByteStreamSplitWriter %d bytes", prefix, getAllocatedSize());
+    }
+  }
+
+  public static class DoubleByteStreamSplitValuesWriter extends ByteStreamSplitValuesWriter {
+
+    private static final int DOUBLE_SIZE_IN_BYTES = 8;
```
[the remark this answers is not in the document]
martinradev: Removed DOUBLE_SIZE_IN_BYTES,

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static final int FLOAT_SIZE_IN_BYTES = 4;
+
+    public FloatByteStreamSplitValuesWriter(int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+      super(FLOAT_SIZE_IN_BYTES, initialCapacity, pageSize, allocator);
+    }
+
+    @Override
+    public void writeFloat(float v) {
+      int vAsInt = Float.floatToIntBits(v);
```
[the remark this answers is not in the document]
martinradev: Done.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+      byte[] bytes = new byte[FLOAT_SIZE_IN_BYTES];
+      for (int i = 0; i < FLOAT_SIZE_IN_BYTES; ++i) {
+        byte b = (byte) ((vAsInt >> (i * 8)) & 0xFF);
```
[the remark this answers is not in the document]
martinradev: Removed.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -0,0 +1,150 @@
...
+    long totalCapacity = 0;
+    for (CapacityByteArrayOutputStream stream : byteStreams) {
+      totalCapacity += stream.getCapacity();
+    }
+    return totalCapacity;
+  }
+
+  public static class FloatByteStreamSplitValuesWriter extends ByteStreamSplitValuesWriter {
+
+    private static final int FLOAT_SIZE_IN_BYTES = 4;
```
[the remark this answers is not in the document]
martinradev: Removed and substituted occurrences with Float.BYTES

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -33,19 +32,7 @@ public ByteStreamSplitValuesReaderForDouble() {
...
-                   ((long)(valueByteBuffer[4] & 0xFF) << 32) |
-                   ((long)(valueByteBuffer[5] & 0xFF) << 40) |
-                   ((long)(valueByteBuffer[6] & 0xFF) << 48) |
-                   ((long)(valueByteBuffer[7] & 0xFF) << 56);
-      return Double.longBitsToDouble(value);
-    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
-      throw new ParquetDecodingException("Could not read double.", ex);
-    }
+    gatherElementDataFromStreams(valueByteBuffer);
+    return ByteBuffer.wrap(valueByteBuffer).order(ByteOrder.LITTLE_ENDIAN).getDouble();
```
gszadovszky: I don't agree with @shangxinli. The `ByteBuffer` functionality might be readable but here performance is critical. You can speed up this line by having the `ByteBuffer` object created at the class level and set the order so you only invoke `getDouble()` here but even in this case a lower level implementation will perform better.
To reuse code, I would suggest to take a look on [ByteUtils](https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/bytes/BytesUtils.java).

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFloat.java
```diff
@@ -33,15 +32,7 @@ public ByteStreamSplitValuesReaderForFloat() {
...
       gatherElementDataFromStreams(valueByteBuffer);
-      int value = (int)(valueByteBuffer[0] & 0xFF) |
-                  ((int)(valueByteBuffer[1] & 0xFF) << 8) |
-                  ((int)(valueByteBuffer[2] & 0xFF) << 16) |
-                  ((int)(valueByteBuffer[3] & 0xFF) << 24);
-      return Float.intBitsToFloat(value);
-    } catch (IOException | ArrayIndexOutOfBoundsException ex) {
-      throw new ParquetDecodingException("Could not read float.", ex);
-    }
+      return ByteBuffer.wrap(valueByteBuffer).order(ByteOrder.LITTLE_ENDIAN).getFloat();
```
gszadovszky: Same as above.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -102,47 +109,33 @@ public long getAllocatedSize() {
...
     @Override
     public void writeFloat(float v) {
-      int vAsInt = Float.floatToIntBits(v);
-      byte[] bytes = new byte[FLOAT_SIZE_IN_BYTES];
-      for (int i = 0; i < FLOAT_SIZE_IN_BYTES; ++i) {
-        byte b = (byte) ((vAsInt >> (i * 8)) & 0xFF);
-        bytes[i] = b;
-      }
-      super.scatterBytes(bytes);
+      super.scatterBytes(ByteBuffer.allocate(Float.BYTES).order(ByteOrder.LITTLE_ENDIAN).putFloat(v).array());
```
gszadovszky: I would prefer the previous implementation because of performance reasons. You may check [ByteUtils](https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/bytes/BytesUtils.java) to reuse code.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -102,47 +109,33 @@ public long getAllocatedSize() {
...
     @Override
     public void writeDouble(double v) {
-      long vAsLong = Double.doubleToLongBits(v);
-      byte[] bytes = new byte[DOUBLE_SIZE_IN_BYTES];
-      for (int i = 0; i < DOUBLE_SIZE_IN_BYTES; ++i) {
-        byte b = (byte) ((vAsLong >> (i * 8)) & 0xFF);
-        bytes[i] = b;
-      }
-      super.scatterBytes(bytes);
+      super.scatterBytes(ByteBuffer.allocate(Double.BYTES).order(ByteOrder.LITTLE_ENDIAN).putDouble(v).array());
```
gszadovszky: See above.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
martinradev: I see, using ByteUtils is likely better. I am not aware what the jit compiler will generate and I have not measured the performance impact of the proposed change. But using those static functions to serialize/deserialize the values is probably better.
I have my own benchmark I will use in the near future to check whether performance is bad.
Will address this in the evening.

PR #749, opened 2020-01-16
PARQUET-1771: Support configurable for DirectByteBufferAllocator from Hadoop Configuration

LiShuMing:
### Jira

- "Support configurable for DirectByteBufferAllocator from Hadoop Configuration": https://jira.apache.org/jira/browse/PARQUET-1771#

### Tests

- Old unit tests.

### Commits

- PARQUET-1771: Support configurable for DirectByteBufferAllocator from Hadoop Configuration

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
@@ -243,8 +243,8 @@ public class ParquetReadOptions {
       return this;
     }
 
-    public Builder useDirectByteBufferAllocator(boolean val) {
-      if (val) {
+    public Builder useDirectByteBufferAllocator(boolean useDirectByteBuffer) {
+      if (useDirectByteBuffer) {
         this.allocator = new DirectByteBufferAllocator();
       }
       return this;
--- a/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
@@ -20,6 +20,7 @@
 package org.apache.parquet;
 
 import org.apache.parquet.bytes.ByteBufferAllocator;
+import org.apache.parquet.bytes.DirectByteBufferAllocator;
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
```diff
@@ -223,6 +224,13 @@ public Builder withRecordFilter(FilterCompat.Filter rowGroupFilter) {
       return this;
     }
 
+    public Builder useDirectByteBufferAllocator(boolean val) {
```
Fokko:
```suggestion
    public Builder useDirectByteBufferAllocator(boolean allocator) {
```

Fokko:
We already have:
```
    public Builder withAllocator(ByteBufferAllocator allocator) {
      this.allocator = allocator;
      return this;
    }
```
which provides similar functionality. Or am I missing something.

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
LiShuMing: Thanks for replies. 

In `Spark`,   [ParquetFileReader](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala#L272)  passes `options` from `Hadoop Configuration` (using ParquetReadOptions.Builder) rather than `ParquetReadOptions`.

In `ParquetReadOptions.Builder` constructor,  now `Allocator` param cannot be passed to `ParquetReadOptions`.

Fokko: Makes sense, could you still rename `val` to `useDirectByteBuffer`?

LiShuMing: Hi~ I have fixed this and rebased the newest codes.

Can you review this again?

vdiravka:
I wonder, if it solves https://issues.apache.org/jira/browse/PARQUET-1006 too. Could you check it please?

PR #754, opened 2020-02-06
PARQUET-1784: Column-wise configuration

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ColumnProperty.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ColumnProperty.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.column;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.parquet.hadoop.metadata.ColumnPath;
+
+/**
+ * Represents a Parquet property that may have different values for the different columns.
+ */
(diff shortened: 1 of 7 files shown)


parquet-column/src/main/java/org/apache/parquet/column/ColumnProperty.java
```diff
@@ -0,0 +1,137 @@
...
+      return Objects.toString(getDefaultValue());
+    }
+  }
+
+  private static class MultipleColumnProperty<T> extends DefaultColumnProperty<T> {
+    private final Map<ColumnPath, T> values;
+
+    private MultipleColumnProperty(T defaultValue, Map<ColumnPath, T> values) {
+      super(defaultValue);
+      assert !values.isEmpty();
```
chenjunjiedada: Is it better to use a warning here?

gszadovszky: I usually use `assert` to ensure something that if not happens then it is a coding issue. Here, this private constructor is invoked only at one place where the map is not empty. It is just that I've coded this line first and wanted to not forget about the check :)
(`assert` is good for such checks because they do not impact performance in production.)

chenjunjiedada: I see, thanks for the explanation.

chenjunjiedada:
LGTM, it is useful for bloom filter as well.

parquet-column/src/main/java/org/apache/parquet/column/ColumnProperty.java
```diff
@@ -0,0 +1,137 @@
...
+      return getDefaultValue();
+    }
+
+    @Override
+    public String toString() {
+      return Objects.toString(getDefaultValue());
+    }
+  }
+
+  private static class MultipleColumnProperty<T> extends DefaultColumnProperty<T> {
```
shangxinli: why not name it as 'MultipleDefaultColumnProperty'?

gszadovszky: I usually don't like long class names if it is not required and here it is a private class. But if you have a strong opinion to change the name I'm happy to do so. :)

shangxinli: I usually like the name to be more specific than being short. But I don't have a strong opinion as this is a pretty personal preference.

chenjunjiedada:
We still need a committer for reviewing. @Fokko

mixermt:
Hi,
It is not clear how the property should work with lists
Example:

listField:                                                                         OPTIONAL F:1 
.list:                                                                                                REPEATED F:1 
..element:                                                                                            OPTIONAL INT64 R:1 D:3

gszadovszky:
> Hi, It is not clear how the property should work with lists Example:
> 
> listField: OPTIONAL F:1 .list: REPEATED F:1 ..element: OPTIONAL INT64 R:1 D:3

@mixermt, you should reference the primitives by their paths in the schema. In your example it would be `listField.list.element`.

PR #755, opened 2020-02-09
PARQUET-1791: Add 'prune' command to parquet-tools

shangxinli:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/command/PruneColumnsCommand.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/command/PruneColumnsCommand.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.tools.command;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.example.data.simple.Primitive;
+import org.apache.parquet.format.converter.ParquetMetadataConverter;
+import org.apache.parquet.hadoop.ParquetFileReader;
+import org.apache.parquet.hadoop.ParquetFileWriter;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
+import org.apache.parquet.hadoop.metadata.FileMetaData;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.hadoop.util.HadoopInputFile;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.OriginalType;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.tools.Main;
+
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.HashSet;
+
+public class PruneColumnsCommand extends ArgsOnlyCommand {
+
+  public static final String[] USAGE = new String[] {
+    "<input> <output> [<column> ...]",
+
+    "where <input> is the source parquet file",
+    "    <output> is the destination parquet file," +
+    "    [<column> ...] are the columns in the case senstive dot format" +
+      " to be pruned, for example a.b.c"
+  };
+
+  /**
+   * Biggest number of columns we can prune.
+   */
+  private static final int MAX_COL_NUM = 100;
+
+  private Configuration conf;
+
+  public PruneColumnsCommand() {
+    super(3, MAX_COL_NUM + 1);
+
+    conf = new Configuration();
+  }
+
+  @Override
+  public String[] getUsageDescription() {
+    return USAGE;
+  }
+
+  @Override
+  public String getCommandDescription() {
+    return "Prune column(s) in a Parquet file and save it to a new file. " +
+      "The columns left are not changed.";
+  }
+
+  @Override
+  public void execute(CommandLine options) throws Exception {
+    List<String> args = options.getArgList();
+    Path inputFile = new Path(args.get(0));
+    Path outputFile = new Path(args.get(1));
+    List<String> cols = args.subList(2, args.size());
+
+    Set<ColumnPath> prunePaths = convertToColumnPaths(cols);
+
+    ParquetMetadata pmd = ParquetFileReader.readFooter(conf, inputFile, ParquetMetadataConverter.NO_FILTER);
+    FileMetaData metaData = pmd.getFileMetaData();
+    MessageType schema = metaData.getSchema();
+    List<String> paths = new ArrayList<>();
+    getPaths(schema, paths, null);
+
+    for (String col : cols) {
+      if (!paths.contains(col)) {
+        throw new IllegalArgumentException("Input column name " + col + " doesn't show up in the schema of file " + inputFile.getName());
+      }
+    }
+
+    ParquetFileWriter writer = new ParquetFileWriter(conf,
+      pruneColumnsInSchema(schema, prunePaths), outputFile, ParquetFileWriter.Mode.CREATE);
+
+    writer.start();
+    writer.appendFile(HadoopInputFile.fromPath(inputFile, conf));
+    writer.end(metaData.getKeyValueMetaData());
+  }
+
+  // We have to rewrite getPaths because MessageType only get level 0 paths
+  private void getPaths(GroupType schema, List<String> paths, String parent) {
+    List<Type> fields = schema.getFields();
+    String prefix = (parent == null) ? "" : parent + ".";
+    for (Type field : fields) {
+      paths.add(prefix + field.getName());
+      if (field instanceof GroupType) {
+        getPaths(field.asGroupType(), paths, prefix + field.getName());
+      }
+    }
+  }
+
+  private MessageType pruneColumnsInSchema(MessageType schema, Set<ColumnPath> prunePaths) {
+
+    List<Type> fields = schema.getFields();
+    List<String> currentPath = new ArrayList<>();
+    List<Type> prunedFields = pruneColumnsInFields(fields, currentPath, prunePaths);
+    MessageType newSchema = new MessageType(schema.getName(), prunedFields);
+    return newSchema;
+  }
+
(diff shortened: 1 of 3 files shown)


parquet-tools/src/main/java/org/apache/parquet/tools/command/PruneColumnsCommand.java
```diff
@@ -0,0 +1,149 @@
...
+
+  private MessageType pruneColumnsInSchema(MessageType schema, Set<ColumnPath> prunePaths) {
+
+    List<Type> fields = schema.getFields();
+    List<String> currentPath = new ArrayList<>();
+    List<Type> prunedFields = pruneColumnsInFields(fields, currentPath, prunePaths);
+    MessageType newSchema = new MessageType(schema.getName(), prunedFields);
+    return newSchema;
+
+    //return schema;
```
gszadovszky: Please, remove this line.

shangxinli: sure

shangxinli:
@gszadovszky, Thanks for reviewing the code. 
1. I will add parquet-cli later and I created ticket Parquet-1800 for it. But meanwhile, I think it is still beneficial to add to parquet-tools since it is already ready. We can remove parquet-tools as a whole when we want to do it. 

2. Yeah, it is not a trivial task to add column-index, but I created a ticket Parquet-1801 to do that. 

3. I just added several unit tests(base cases, nested columns, negative tests).

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.Type;
+import org.junit.Test;
+
+import java.util.List;
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestPruneColumnsCommand {
```
gszadovszky: Checking in binary parquet files is not really future proof. It is usually a better idea to create the input file in the test and remove it at the end. (There are many examples in the tests to do so.)

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,112 @@
...
+
+  private Configuration conf = new Configuration();
+  private PruneColumnsCommand command = new PruneColumnsCommand();
+  private Random rand = new Random();
+
+  @Test
+  public void testPruneOneColumn() throws Exception {
+    // File test.parquet has 3 columns: "name", "age", "gender"
+    String inputFile = "src/test/java/org/apache/parquet/tools/data/test.parquet";
```
gszadovszky: This path depends on the path where the maven command is executed from. Maybe, that's why it is failing on Travis.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,112 @@
...
+    String outputFile =  rand.nextInt() + ".parquet";
+    String cargs[] = {inputFile, outputFile, "gender"};
+    executeCommandLine(cargs);
+    ParquetMetadata pmd = ParquetFileReader.readFooter(conf, new Path(outputFile), ParquetMetadataConverter.NO_FILTER);
+    MessageType schema = pmd.getFileMetaData().getSchema();
+    List<Type> fields = schema.getFields();
+    assertEquals(fields.size(), 2);
+    assertEquals(fields.get(0).getName(), "name");
+    assertEquals(fields.get(1).getName(), "age");
```
gszadovszky: It would also be great to verify the actual data. Maybe, you may use the `dump` command?

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: Sounds good! I saw there are several places using the binary files, so I just followed. But you are right. We should create the files on the fly. I changed the tests to do that. Please have a look.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: Yes. I just wrote a reader to verify.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: Fixed.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,258 @@
...
+      new PrimitiveType(REQUIRED, INT64, "DocId"),
+      new PrimitiveType(REQUIRED, BINARY, "Name"),
+      new PrimitiveType(REQUIRED, BINARY, "Gender"),
+      new GroupType(OPTIONAL, "Links",
+        new PrimitiveType(REPEATED, INT64, "Backward"),
+        new PrimitiveType(REPEATED, INT64, "Forward")));
+
+    conf.set(GroupWriteSupport.PARQUET_EXAMPLE_SCHEMA, schema.toString());
+
+    Path file = new Path(root, + rand.nextInt(100000) + "input.parquet");
```
gszadovszky: I would prefer using creating the file in a temporary space instead. (See e.g. `java.nio.file.Files.createTempFile`)
Creating a file int the hard-coded `"target/"` relative path is a bit error prone. What happens if the `mvn` command is executed from outside of the project structure?

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,258 @@
...
+        if (!prunePaths.contains("Links.Forward")) {
+          assertEquals(3l, subGroup.getLong("Forward", 0));
+        }
+      }
+    }
+    reader.close();
+  }
+
+  private String createParquetFile() throws IOException {
+    MessageType schema = new MessageType("schema",
```
gszadovszky: nit: Using the builders in `org.apache.parquet.schema.Types` would result a nicer and more readable code.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,258 @@
...
+      new PrimitiveType(REQUIRED, BINARY, "Gender"),
+      new GroupType(OPTIONAL, "Links",
+        new PrimitiveType(REPEATED, INT64, "Backward"),
+        new PrimitiveType(REPEATED, INT64, "Forward")));
+
+    conf.set(GroupWriteSupport.PARQUET_EXAMPLE_SCHEMA, schema.toString());
+
+    Path file = new Path(root, + rand.nextInt(100000) + "input.parquet");
+
+    ParquetWriter<Group> writer = new ParquetWriter<>(file, new GroupWriteSupport(), ParquetWriter.DEFAULT_COMPRESSION_CODEC_NAME,
```
gszadovszky: nit: Using `org.apache.parquet.hadoop.example.ExampleParquetWriter.builder` makes creating the writer easier and more readable.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,258 @@
...
+      g.add("DocId", 1l);
+      g.add("Name", "foo");
+      g.add("Gender", "male");
+      Group links = g.addGroup("Links");
+      links.add(0, 2l);
+      links.add(1, 3l);
+      writer.write(g);
+    }
+
+    writer.close();
```
gszadovszky: I would prefer using a `try-with-resources` construct instead.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
```diff
@@ -0,0 +1,258 @@
...
+    assertEquals(subFields.get(0).getName(), "Backward");
+    assertEquals(subFields.get(1).getName(), "Forward");
+
+    // Verify the data are not changed for the columns not pruned
+    List<String> prunePaths = Arrays.asList("Name", "Gender");
+    validateColumns(inputFile, prunePaths);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testNotExistsColumn() throws Exception {
```
gszadovszky: Let's check the following scenario.
User wants to prune the columns `a`, `b` and `c` from several parquet files. The schema of the files are not the same because they evolved (e.g. column `c` is added after some files were already created or column `a` is used for partitioning so many files do not contain it).
If you think the previous scenario is valid then the tool should not throw an exception in case a column to be pruned is not in the file. Meanwhile, the user might mistyped the column name and it may not notify. So, it is your choice how you should handle this case. :)

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: Fixed

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: I see this way is used in multiple places of the other tests. Since this is a pretty simple schema, it seems straightforward. But if you have a strong opinion, let me know and I will fix it.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: Great idea!

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: sure

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
shangxinli: That is a really good point! I was trying to prevent 'silent failing' when user has a typo for the column names. Now I think we can just print a warning message.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
gszadovszky: It was just nit picking, no strong opinion. (I guess the other places uses this way because they were developed before the schema builder design was created.)

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestPruneColumnsCommand.java
gszadovszky: You still expects the exception to be thrown in the two related tests. That's why Travis fails.

shangxinli: Thanks for pointing out! Just fixed both.

shangxinli:
Thanks @gszadovszky for reviewing the code and valuable comments!

gszadovszky:
Thank you for your contribution, @shangxinli!

PR #756, opened 2020-02-09
PARQUET-1790: Add Api for writing DataPageV2 to ParquetFileWriter class

brimzi:
This PR addresses [PARQUET-1790](https://issues.apache.org/jira/projects/PARQUET/issues/PARQUET-1790).

The ParquetFileWriter class currently does not  have an API for writing a DataPageV2 page. 
A similar API is already defined in ColumnChunkPageWriteStore and inspiration can/should be derived from there in implementing this.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -886,6 +951,15 @@ private static void serializeColumnIndexes(
     }
   }
 
+  private int toIntWithCheck(long size) {
```
gszadovszky: In other places we use `java.util.Math.toIntExact(long)`.

brimzi: Yes you are right, I was thinking of keeping it this way here(like in ColumnChunkPageWriteStore ) so the message can be clearer to the caller of what failed

gszadovszky: It's fine for me.
Then, I would suggest using a similar approach of implementation then they have it in `toIntExact(long)`. The one you used is not universal (think of negative numbers out of int).

PR #753, opened 2020-02-05
PARQUET-1787: Fix parsing expected distinct numbers

garawalid:
Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references them in the PR title. For example, "PARQUET-1234: My Parquet PR"
  - https://issues.apache.org/jira/browse/PARQUET-XXX
  - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #763, opened 2020-02-27
PARQUET-1805: Refactor the configuration for bloom filters

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -464,7 +464,7 @@ public class ParquetProperties {
     public Builder withBloomFilterNDV(String columnPath, long ndv) {
       Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %d", columnPath, ndv);
       this.bloomFilterNDVs.withValue(columnPath, ndv);
-      // Setting and NDV for a column implies writing a bloom filter
+      // Setting an NDV for a column implies writing a bloom filter
       this.bloomFilterEnabled.withValue(columnPath, true);
       return this;
     }
@@ -473,7 +473,7 @@ public class ParquetProperties {
      * Enable or disable the bloom filter for the columns not specified by
      * {@link #withBloomFilterEnabled(String, boolean)}.
      *
-     * @param enabled whether bloom filter shall be enabled
+     * @param enabled whether bloom filter shall be enabled for all columns
      *
      * @return this builder for method chaining
      */
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -455,6 +455,8 @@ public class ParquetProperties {
 
     /**
      * Set Bloom filter NDV (number of distinct values) for the specified column.
+     * If set for a column then the writing of the bloom filter for that column will be automatically enabled (see
+     * {@link #withBloomFilterEnabled(String, boolean)}).
      *
      * @param columnPath the path of the column (dot-string)
      * @param ndv the NDV of the column
@@ -483,7 +485,10 @@ public class ParquetProperties {
     }
 
     /**
-     * Enable or disable the bloom filter for the specified column
+     * Enable or disable the bloom filter for the specified column.
+     * One may either disable bloom filters for all columns by invoking {@link #withBloomFilterEnabled(boolean)} with a
+     * {@code false} value and then enable the bloom filters for the required columns one-by-one by invoking this
+     * method or vice versa.
      *
      * @param columnPath the path of the column (dot-string)
      * @param enabled    whether bloom filter shall be enabled
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -18,13 +18,8 @@
  */
 package org.apache.parquet.column;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
+import java.util.OptionalLong;
 
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.ByteBufferAllocator;
@@ -61,6 +56,7 @@ public class ParquetProperties {
   public static final int DEFAULT_STATISTICS_TRUNCATE_LENGTH = Integer.MAX_VALUE;
   public static final int DEFAULT_PAGE_ROW_COUNT_LIMIT = 20_000;
   public static final int DEFAULT_MAX_BLOOM_FILTER_BYTES = 1024 * 1024;
+  public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
@@ -102,10 +98,10 @@ public class ParquetProperties {
   private final int columnIndexTruncateLength;
   private final int statisticsTruncateLength;
 
-  // The key-value pair represents the column name and its expected distinct number of values in a row group.
-  private final Map<String, Long> bloomFilterExpectedDistinctNumbers;
+  // The expected NDV (number of distinct values) for each columns
+  private final ColumnProperty<Long> bloomFilterNDVs;
   private final int maxBloomFilterBytes;
-  private final List<String> bloomFilterColumns;
+  private final ColumnProperty<Boolean> bloomFilterEnabled;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final boolean enableByteStreamSplit;
@@ -125,8 +121,8 @@ public class ParquetProperties {
     this.valuesWriterFactory = builder.valuesWriterFactory;
     this.columnIndexTruncateLength = builder.columnIndexTruncateLength;
     this.statisticsTruncateLength = builder.statisticsTruncateLength;
-    this.bloomFilterExpectedDistinctNumbers = builder.bloomFilterColumnExpectedNDVs;
-    this.bloomFilterColumns = builder.bloomFilterColumns;
+    this.bloomFilterNDVs = builder.bloomFilterNDVs.build();
+    this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
@@ -257,16 +253,13 @@ public class ParquetProperties {
     return pageWriteChecksumEnabled;
   }
 
-  public Map<String, Long> getBloomFilterColumnExpectedNDVs() {
-    return bloomFilterExpectedDistinctNumbers;
+  public OptionalLong getBloomFilterNDV(ColumnDescriptor column) {
+    Long ndv = bloomFilterNDVs.getValue(column);
+    return ndv == null ? OptionalLong.empty() : OptionalLong.of(ndv);
   }
 
-  public Set<String> getBloomFilterColumns() {
-    if (bloomFilterColumns != null && bloomFilterColumns.size() > 0){
-      return new HashSet<>(bloomFilterColumns);
-    }
-
-    return bloomFilterExpectedDistinctNumbers.keySet();
+  public boolean isBloomFilterEnabled(ColumnDescriptor column) {
+    return bloomFilterEnabled.getValue(column);
   }
 
   public int getMaxBloomFilterBytes() {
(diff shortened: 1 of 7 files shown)


gszadovszky:
@chenjunjiedada, @garawalid, you may be interested in this review.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -61,6 +56,7 @@
   public static final int DEFAULT_STATISTICS_TRUNCATE_LENGTH = Integer.MAX_VALUE;
   public static final int DEFAULT_PAGE_ROW_COUNT_LIMIT = 20_000;
   public static final int DEFAULT_MAX_BLOOM_FILTER_BYTES = 1024 * 1024;
+  public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
```
chenjunjiedada: Why we set the default to false?

gszadovszky:
Setting `parquet.bloom.filter.enabled` is just an option to enable writing bloom filters for every column. Or you can enable for all and then disable some one-by-one using the column-wise format (e.g. `parquet.bloom.filter.enabled#column1 = false`).
The `DEFAULT_BLOOM_FILTER_ENABLED = false` means the writing of bloom filters is disabled by default so the user have to either enable it for all or enable it for some specific columns.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
garawalid: @chenjunjiedada 
You can find more details about the topic in these issues: [PARQUET-1787](https://issues.apache.org/jira/browse/PARQUET-1787) and [PARQUET-1784](https://issues.apache.org/jira/browse/PARQUET-1784)

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -459,26 +454,44 @@ public Builder withMaxBloomFilterBytes(int maxBloomFilterBytes) {
...
+     *
+     * @param columnPath the path of the column (dot-string)
+     * @param ndv the NDV of the column
+     *
+     * @return this builder for method chaining
+     */
+    public Builder withBloomFilterNDV(String columnPath, long ndv) {
+      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %d", columnPath, ndv);
+      this.bloomFilterNDVs.withValue(columnPath, ndv);
+      // Setting and NDV for a column implies writing a bloom filter
```
garawalid: Typo: `and` to `a`

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -459,26 +454,44 @@ public Builder withMaxBloomFilterBytes(int maxBloomFilterBytes) {
...
+      this.bloomFilterEnabled.withValue(columnPath, true);
+      return this;
+    }
+
+    /**
+     * Enable or disable the bloom filter for the columns not specified by
+     * {@link #withBloomFilterEnabled(String, boolean)}.
      *
-     * @param columnToNDVMap the columns which has bloom filter enabled.
+     * @param enabled whether bloom filter shall be enabled
```
garawalid: I'd write
```suggestion
     * @param enabled whether bloom filter shall be enabled for all columns
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -150,8 +146,8 @@
   public static final String ESTIMATE_PAGE_SIZE_CHECK = "parquet.page.size.check.estimate";
   public static final String COLUMN_INDEX_TRUNCATE_LENGTH = "parquet.columnindex.truncate.length";
   public static final String STATISTICS_TRUNCATE_LENGTH = "parquet.statistics.truncate.length";
-  public static final String BLOOM_FILTER_COLUMN_NAMES = "parquet.bloom.filter.column.names";
-  public static final String BLOOM_FILTER_EXPECTED_NDV = "parquet.bloom.filter.expected.ndv";
+  public static final String BLOOM_FILTER_ENABLED = "parquet.bloom.filter.enabled";
+  public static final String BLOOM_FILTER_EXPECTED_NDV = "parquet.bloom.filter.ndv";
```
garawalid: I'd keep it `parquet.bloom.filter.expected.ndv` because it's clear and the NDV still an estimation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -558,30 +556,38 @@ public SELF withPageWriteChecksumEnabled(boolean enablePageWriteChecksum) {
...
-      if (columns != null) {
-        encodingPropsBuilder.withBloomFilterColumnNames(Arrays.asList(columns));
-      }
+    public SELF withBloomFilterEnabled(boolean enabled) {
+      encodingPropsBuilder.withBloomFilterEnabled(enabled);
+      return self();
+    }
 
+    /**
+     * Sets the bloom filter enabled/disabled for the specified column
```
garawalid: Wouldn't it be better to say that the result of this method will be combined with `withBloomFilterEnabled(boolean enabled)`?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
chenjunjiedada: Thanks @garawalid.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
gszadovszky: I wouldn't say combined. If an enabled/disabled state is set for a specific column then that state would be used independently from the "universal" state set. Let me try to make it more clear in the comments.

gszadovszky:
@garawalid, it sounds reasonable to describe the difference of the ndv and the enabled/disabled state configurations. But where should I put it? Currently, there is no proper documentation for the parquet hadoop configs.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
garawalid: I agree, it's better now!

garawalid:
@gszadovszky 
I ve been thinking about it too. I'll put the note within the unit tests in #753

chenjunjiedada:
@gszadovszky , Actually, I think we don't have the case about all columns enabled with bloom filters in the real scenario. Even we want to do that we can config via column-wise way one by one. How about removing this?

gszadovszky:
@chenjunjiedada, it was an idea @garawalid came up with in the comments of PARQUET-1784. It can also be used for the case of having more columns for bloom filters than the others. In this case you may turn on the "root" switch, an turn off one-by-one the others where no bloom filter is required.

@garawalid, we might add a new page to the parquet site (parquet.apache.org) containing all the hadoop configuration available for parquet. However, it would be a separate jira.

garawalid:
@chenjunjiedada 
Let's assume that we have some clients and many system information (simply databases) where they are represented by different IDs, we would like to keep a mapping of a client with its IDs in different systems. We may need to use bloom filter for all id columns.
In my company we have such tables where the number of columns is more than 20 and the size is some GB stored as Parquet.

@gszadovszky 
I like the idea of adding a new page. +1

chenjunjiedada:
@garawalid, Thanks for the explanation, I understood now. 

+1 to have a page to describe the configuration.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -102,10 +98,10 @@ public static WriterVersion fromString(String name) {
   private final int columnIndexTruncateLength;
   private final int statisticsTruncateLength;
 
-  // The key-value pair represents the column name and its expected distinct number of values in a row group.
-  private final Map<String, Long> bloomFilterExpectedDistinctNumbers;
+  // The expected NDV (number of distinct values) for each columns
+  private final ColumnProperty<Long> bloomFilterNDVs;
```
shangxinli: Is it max value?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -292,9 +285,9 @@ public String toString() {
         + "Max row count for page size check is: " + getMaxRowCountForPageSizeCheck() + '\n'
         + "Truncate length for column indexes is: " + getColumnIndexTruncateLength() + '\n'
         + "Truncate length for statistics min/max  is: " + getStatisticsTruncateLength() + '\n'
-        + "Bloom filter enabled column names are: " + getBloomFilterColumns() + '\n'
+        + "Bloom filter enabled: " + bloomFilterEnabled + '\n'
         + "Max Bloom filter size for a column is " + getMaxBloomFilterBytes() + '\n'
-        + "Bloom filter enabled column expected number of distinct values are: " + getBloomFilterColumnExpectedNDVs().values() + '\n'
+        + "Bloom filter expected number of distinct values are: " + bloomFilterNDVs + '\n'
```
shangxinli: are -> is

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -459,26 +454,44 @@ public Builder withMaxBloomFilterBytes(int maxBloomFilterBytes) {
...
+     * {@link #withBloomFilterEnabled(String, boolean)}.
      *
-     * @param columnToNDVMap the columns which has bloom filter enabled.
+     * @param enabled whether bloom filter shall be enabled
      *
      * @return this builder for method chaining
      */
-    public Builder withBloomFilterColumnToNDVMap(Map<String, Long> columnToNDVMap) {
-      this.bloomFilterColumnExpectedNDVs = columnToNDVMap;
+    public Builder withBloomFilterEnabled(boolean enabled) {
```
shangxinli: Can we document the settings of their relations? Here are my understandings. 
1. 'withBloomFilterEnabled(boolean enabled)' would set it globally, which can be overridden by column specific setting of 'withBloomFilterEnabled(String columnPath, boolean enabled)' or 'withBloomFilterNDV(String columnPath, long ndv)'. 2. Now 'withBloomFilterNDV(String columnPath, long ndv)' will override 'withBloomFilterEnabled(String columnPath, boolean enabled)' implicitly or vice versa explicitly. 

I feel it is a bit complicated to use given the above relations. It is also my understanding that NDV should only present when bloomFitler is enabled. If that is the case, can we just use NDV null or not to indicate enabling/disabling to simplify the settings.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
gszadovszky: It depends. `bloomFilterNDVs` is a map like structure that contains an NDV for each columns. The output of this line can be either one default value or several column name - NDV pairs.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
gszadovszky: `ColumnProperty` is created to handle different values per column names. It may contain one value for each column.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
gszadovszky: Bloom filters are not written by default for any columns. You may turn it on for the different columns. The "global" setting is just a shortcut to turn bloom filters on for all columns (and you may turn off for some of them).
NDV can be set for a specific column. If not set (but the bloom filter is turned on for that column) bloom filter can still be written using a default calculation. It is another short cut that if the NDV is set for a column then the user obviously want to write a bloom filter for that column so we set the related flag.

gszadovszky:
@shangxinli, could you please check it again?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
shangxinli: Sounds good. I thought earlier it was just one single number to allow the maximum NDVs, but I was wrong.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
shangxinli: We were on the same page except for the 'default calculation' for NDV. Now I understand better. But my concern is still that it is a little complicated to use these 3 settings including the 2 shortcuts.  How about we rename the two column-specific setting methods like withEnableBloomFilterWithNDV() and withEnableBloomFilterWithDefaultNDV()?

gszadovszky: There is no default NDV. We calculate the optimal size of the bloom filter based on the NDV if it is set and use the maximum size if it is not. (See the related code [here](https://github.com/apache/parquet-mr/pull/763/files#diff-c50835e13bd6059132963b559e57303bR91-R97))

I've tried to explain the current behavior of the properties more precisely in the method comments.

shangxinli: Sounds good

PR #766, opened 2020-03-04
[PARQUET-1808]: Replacing multiple String Object creations with StringBuilder

koiralo:

### Jira

- [x] My PR addresses the following [PARQUET-1808](https://issues.apache.org/jira/browse/PARQUET-1808)
  - https://issues.apache.org/jira/browse/PARQUET-1808

### Tests

- [x] Test Covered with existing tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
+++ b/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
@@ -44,10 +44,10 @@ public class SimpleGroup extends Group {
 
   @Override
   public String toString() {
-    return toString("");
+    return toString(new StringBuilder()).toString();
   }
 
-  public String toString(String indent) {
+  public StringBuilder toString(StringBuilder indent) {
     StringBuilder result = new StringBuilder();
     int i = 0;
     for (Type field : schema.getFields()) {
@@ -60,14 +60,14 @@ public class SimpleGroup extends Group {
             if (value == null) {
               result.append(": NULL\n");
             } else if (value instanceof Group) {
-              result.append("\n").append(((SimpleGroup)value).toString(indent+"  "));
+              result.append("\n").append(((SimpleGroup)value).toString(indent.append("  ")));
             } else {
               result.append(": " ).append(value.toString()).append("\n");
             }
           }
       }
     }
-    return result.toString();
+    return result;
   }
 
   @Override
--- a/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
+++ b/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
@@ -46,7 +46,8 @@ public class SimpleGroup extends Group {
   public String toString() {
     return toString("");
   }
-  private StringBuilder appendToString(StringBuilder builder, String indent){
+
+  private StringBuilder appendToString(StringBuilder builder, String indent) {
     int i = 0;
     for (Type field : schema.getFields()) {
       String name = field.getName();
--- a/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
+++ b/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
@@ -61,7 +61,7 @@ public class SimpleGroup extends Group {
           } else if (value instanceof Group) {
             builder.append("\n").append(((SimpleGroup)value).appendToString(builder,indent+"  "));
           } else {
-            builder.append(": " ).append(value.toString()).append("\n");
+            builder.append(": ").append(value.toString()).append('\n');
           }
         }
       }
--- a/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
+++ b/parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
@@ -56,17 +56,13 @@ public class SimpleGroup extends Group {
       ++i;
       if (values != null && !values.isEmpty()) {
           for (Object value : values) {
-            result.append(indent);
-            result.append(name);
+            result.append(indent).append(name);
             if (value == null) {
               result.append(": NULL\n");
             } else if (value instanceof Group) {
-              result.append("\n");
-              result.append(((SimpleGroup)value).toString(indent+"  "));
+              result.append("\n").append(((SimpleGroup)value).toString(indent+"  "));
             } else {
-              result.append(": " );
-              result.append(value.toString());
-              result.append("\n");
+              result.append(": " ).append(value.toString()).append("\n");
             }
           }
       }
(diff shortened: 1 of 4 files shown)


parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -48,28 +48,30 @@ public String toString() {
...
           for (Object value : values) {
-            result += indent + name;
+            result.append(indent);
+            result.append(name);
             if (value == null) {
-              result += ": NULL\n";
+              result.append(": NULL\n");
             } else if (value instanceof Group) {
-              result += "\n" + ((SimpleGroup)value).toString(indent+"  ");
+              result.append("\n");
```
gszadovszky: nit: Appending a char instead of a string with one char would perform a little bit better
```suggestion
              result.append('\n');
```

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -48,28 +48,30 @@ public String toString() {
...
+              result.append(((SimpleGroup)value).toString(indent+"  "));
```
gszadovszky: As we are doing this change because of performance reasons I would create a private method in `SimpleGroup` that gets the `StringBuilder` object is it can append all the stuff into the same builder.

parquet-column/src/main/java/org/apache/parquet/io/MessageColumnIO.java
```diff
@@ -276,11 +276,11 @@ private void printState() {
 
     private void log(Object message, Object...parameters) {
       if (DEBUG) {
-        String indent = "";
+        StringBuilder indent = new StringBuilder();
```
gszadovszky: You may create the builder with an exact capacity so it does not need to grow during appending values.
Also, you may count in the message itself and appending it instead of concatenating the indent to the message separately.

parquet-column/src/test/java/org/apache/parquet/io/ExpectationValidatingConverter.java
```diff
@@ -151,14 +151,16 @@ public Void getCurrentRecord() {
   }
 
   private String path(List<GroupType> path, Type type) {
-    String pathString = "";
+    StringBuilder pathString = new StringBuilder();
     if (path.size() > 0) {
       for (int i = 1; i < path.size(); i++) {
-        pathString += path.get(i).getName() + ".";
+        pathString.append(path.get(i).getName());
```
gszadovszky: nit: you may keep the chain pattern as `append` returns the builder object.
Do not have a strong opinion on the current or the chained version.

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
koiralo: Isn't it already using same StringBuffer?

gszadovszky: This method is recursive in case of the child is a `Group`. This method appends the string generated by this recursive invocation. So, if you have a deeply nested structure it would create a bunch of `StringBuilder` objects, generate `String` objects from them and appends them to the `StringBuilder` object created at the higher level.
Instead, I would suggest passing the `StringBuilder` object created at the top level to the children's methods so you'll have only one builder and everything will be appended to it directly.

koiralo: Thanks for detail explanation, I didn't take a detail look in recursive. I think changing a method signature and using `StringBuilder `should be enough.

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -44,10 +44,10 @@ public SimpleGroup(GroupType schema) {
 
   @Override
   public String toString() {
-    return toString("");
+    return toString(new StringBuilder()).toString();
   }
 
-  public String toString(String indent) {
+  public StringBuilder toString(StringBuilder indent) {
```
gszadovszky: We shall not modify existing public methods as they are breaking changes. That's why I've mentioned creating a new private method that would be invoked by the existing ones.
Let me put an example about the idea I wanted to describe:
```java
public String toString(String indent) {
  StringBuilder builder = new StringBuilder();
  appendToString(builder, indent);
  return builder.toString();
}

private void appendToString(StringBuilder builder, String indent) {
  // ...
        builder.append(indent).append(name);
        if (value == null) {
          builder.append(": NULL\N");
        } else if (value instanceof Group) {
          builder.append('\n');
          ((SimpleGroup) value).appendToString(builder, indent + "  ");
        } else {
          builder.append(": ").append(value).append('\n');
  // ...
}
```

koiralo: Ok make sense. have done the required changes.

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -44,30 +44,34 @@ public SimpleGroup(GroupType schema) {
 
   @Override
   public String toString() {
-    return toString(new StringBuilder()).toString();
+    return toString("");
   }
-
-  public StringBuilder toString(StringBuilder indent) {
-    StringBuilder result = new StringBuilder();
+  private StringBuilder appendToString(StringBuilder builder, String indent){
```
gszadovszky: nit:
```suggestion

  private StringBuilder appendToString(StringBuilder builder, String indent) {
```

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -44,30 +44,34 @@ public SimpleGroup(GroupType schema) {
...
-              result.append("\n").append(((SimpleGroup)value).toString(indent.append("  ")));
-            } else {
-              result.append(": " ).append(value.toString()).append("\n");
-            }
+        for (Object value : values) {
+          builder.append(indent).append(name);
+          if (value == null) {
+            builder.append(": NULL\n");
+          } else if (value instanceof Group) {
+            builder.append("\n").append(((SimpleGroup)value).appendToString(builder,indent+"  "));
```
gszadovszky: This is not correct this way. You are appending the builder to itself which would result mixed up output.
```suggestion
            builder.append('\n');
            ((SimpleGroup) value).appendToString(builder, indent + "  ");
```

parquet-column/src/main/java/org/apache/parquet/example/data/simple/SimpleGroup.java
```diff
@@ -44,30 +44,34 @@ public SimpleGroup(GroupType schema) {
...
+          } else {
+            builder.append(": " ).append(value.toString()).append("\n");
```
gszadovszky: nit:
```suggestion
            builder.append(": ").append(value.toString()).append('\n');
```

PR #769, opened 2020-03-16
PARQUET-1817: Crypto Properties Factory

shangxinli:

### Jira

### Tests

### Commits

### Documentation


ggershinsky:
Travis fails because the branch is broken (on parquet-format-structures - unhandled mix of encryption and bloom filters). A known problem, I'll send a fix for that today.

ggershinsky:
the branch is fixed by #771 . Once it is merged, your pr should pass CI checks after rebase.

gszadovszky:
#771 is merged, Travis restarted.

ggershinsky:
"[WARNING] Files with unapproved licenses:
  /home/travis/build/apache/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java"

shangxinli:
Fixed it. Thanks Gidon and Gabor!

ggershinsky:
@shangxinli @gszadovszky my pull request is ready, but now it depends on this one. I'll send it after 769 is merged.

shangxinli:
> @shangxinli @gszadovszky my pull request is ready, but now it depends on this one. I'll send it after 769 is merged.

The build is green now. Once @gszadovszky reviewed the code, he or myself can merge it and you are ready to go.

shangxinli:
> I assume we will see some implementations of this class in the unit tests of the crypto feature. Just double check...

Yes.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.parquet.hadoop.api.WriteSupport.WriteContext;
+import org.apache.parquet.hadoop.util.ConfigurationUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+import static org.apache.parquet.Preconditions.checkNotNull;
+
+public abstract class CryptoPropertiesFactory {
```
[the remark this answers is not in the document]
shangxinli: sure

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(CryptoPropertiesFactory.class);
+  public static final String CRYPTO_FACTORY_CLASS_PROPERTY_NAME = "parquet.encryption.factory.class";
+
+  public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOException {
+    if (null == conf) {
```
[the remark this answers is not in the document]
shangxinli: I will remove it.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,70 @@
...
+      return cryptoFactory;
+    } catch (InstantiationException | IllegalAccessException e) {
+      throw new BadConfigurationException("could not instantiate CryptoPropertiesFactory class: "
+        + cryptoPropertiesFactoryClass, e);
+    }
+  }
+
+  public abstract void initialize(Configuration hadoopConfig);
+
+  public abstract FileEncryptionProperties getFileEncryptionProperties(
```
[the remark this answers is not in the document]
shangxinli: We will remove initialize().
Yes, we will document the arguments and their purpose. 
I will let @gidon to explain the usage of it. From what I heard from him "tempFileName is because we want to stress that the file name can change."

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -51,20 +54,32 @@ public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOE
...
-
-  public abstract FileEncryptionProperties getFileEncryptionProperties(
-    Configuration fileHadoopConfig, Path tempFilePath,
-    WriteContext fileWriteContext)  throws IOException;
+  /**
+   * Get FileEncryptionProperties object which is created by the implementation of this interface. Please see
+   * the unit test (TBD) for example
+   *
+   * @param fileHadoopConfig Configuration that is used to pass the needed information, e.g. KMS uri
+   * @param tempFilePath File path of the parquet file
```
shangxinli: @ggershinsky, can you make comments about what to be used by this field? Our implementation doesn't need it so far.

ggershinsky: I'd suggest also to add a comment to the interface itself, to remind about its purpose (this helps to understand the method parameters), something like "CryptoPropertiesFactory interface enables transparent activation of Parquet encryption. It's custom implementations produce encryption and decryption properties for each Parquet file, using the information available in Parquet file writers and readers: file path, file extended schema (in writers only) - and also Hadoop configuration properties that can pass custom parameters required by a crypto factory. A factory implementation can use or ignore any of these parameters."

As for the "tempFilePath" - "File path of the parquet file being written. Can be used for AAD prefix creation, key material management, etc. Implementations must not presume the path is permanent, as the file can be moved or renamed later"

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.hadoop.BadConfigurationException;
+import org.apache.parquet.hadoop.api.WriteSupport.WriteContext;
+import org.apache.parquet.hadoop.util.ConfigurationUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+import static org.apache.parquet.Preconditions.checkNotNull;
```
ggershinsky: this import is not needed

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: Good point

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: removed.

ggershinsky:
strange. All Travis checks are green,
https://travis-ci.org/github/apache/parquet-mr/builds/666364206?utm_source=github_status&utm_medium=notification
but the PR is stuck on yellow.

chenjunjiedada:
@ggershinsky , FYI, Iceberg has a PR that has the same status.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -28,19 +28,39 @@
...
+ *
+ * It's customized implementations produce encryption and decryption properties for each Parquet file, using the information
+ * available in Parquet file writers and readers: file path, file extended schema (in writers only) - and also Hadoop
+ * configuration properties that can pass custom parameters required by a crypto factory. A factory implementation can
+ * use or ignore any of these parameters.
+ *
+ * The example could be as below.
+ *
+ * 1. Write a class to implement CryptoPropertiesFactory.
+ * 2. Set configuration of "parquet.encryption.factory.class" with the full namespace of this class.
```
gszadovszky:
```suggestion
 * 2. Set configuration of "parquet.encryption.factory.class" with the fully qualified name of this class.
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -28,19 +28,39 @@
...
+ *    For example, we can set the configuration in SparkSession as below.
+ *       SparkSession spark = SparkSession
+ *                   .config("parquet.encryption.factory.class",
+ *                    "xxx.xxx.CryptoPropertiesImpl")
+ *
+ *    This class will be invoked when the static method loadFactory() is called.
```
gszadovszky:
```suggestion
 *    The implementation of this interface will be instantiated by {@link #loadFactory(Configuration)}.
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -28,19 +28,39 @@
...
+  Logger LOG = LoggerFactory.getLogger(CryptoPropertiesFactory.class);
+  String CRYPTO_FACTORY_CLASS_PROPERTY_NAME = "parquet.encryption.factory.class";
 
+  /**
+   * Load CryptoPropertiesFactory class specified by CRYPTO_FACTORY_CLASS_PROPERTY_NAME as the path in the configuration
+   *
+   * @param conf Configuration where user specifies the class path
+   * @return return object with class CryptoPropertiesFactory if user specified the class path and invoking of
+   * the class succeeds, null if user doesn't specify the class path. RunTimeException with type BadConfigurationException
+   * will be thrown if invoking the configured class fails
```
gszadovszky:
```suggestion
   * @return object with class CryptoPropertiesFactory if user specified the class path and invoking of
   * the class succeeds, null if user doesn't specify the class path
   * @throws BadConfigurationException if the instantiation of the configured class fails
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -51,20 +71,36 @@ public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOE
...
+   * the unit test (TBD) for example
+   *
+   * @param fileHadoopConfig Configuration that is used to pass the needed information, e.g. KMS uri
+   * @param tempFilePath File path of the parquet file being written.
+   *                     Can be used for AAD prefix creation, key material management, etc.
+   *                     Implementations must not presume the path is permanent,
+   *                     as the file can be moved or renamed later
+   * @param fileWriteContext WriteContext to provide information like schema to build the FileEncryptionProperties
+   * @return
+   * @throws IOException
```
gszadovszky: javadoc generation does not like having these tags empty. Please write something to them as well.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -51,20 +71,36 @@ public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOE
...
+   */
+  FileEncryptionProperties getFileEncryptionProperties(Configuration fileHadoopConfig, Path tempFilePath,
+                                                       WriteContext fileWriteContext)  throws IOException;
```
gszadovszky: Why do we think that an `IOException` might occur? Would a file read/write suppose to happen in an implementation?

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -51,20 +71,36 @@ public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOE
...
-  public abstract FileDecryptionProperties getFileDecryptionProperties(
-    Configuration hadoopConfig, Path filePath)  throws IOException;
+  /**
+   *
+   * @param hadoopConfig Configuration that is used to pass the needed information, e.g. KMS uri
+   * @param filePath File path of the parquet file
+   *                 Can be used for AAD prefix verification, part of key metadata etc
+   * @return
+   * @throws IOException
+   */
```
gszadovszky: Missing main description, empty return and throws tags.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -51,20 +71,36 @@ public static CryptoPropertiesFactory loadFactory(Configuration conf) throws IOE
...
+  FileDecryptionProperties getFileDecryptionProperties(Configuration hadoopConfig, Path filePath)  throws IOException;
```
gszadovszky: See my concerns about `IOException` above.

ggershinsky:
> Last time I've forgot about one major thing. (Sorry for the late idea.)
> parquet-mr library is usually used for either reading or writing at a time but not both. I think, it is not a good practice to expect both encryption and decryption properties while only one would be required.
> What do you guys think?

On one hand, its certainly possible to split this into two interfaces, to make it fit the parquet-mr practices. On the other hand, a person working with encryption in eg Spark shell might want to both write and read the files. He would define two Hadoop config params for that, "parquet.encryption.factory.class" and "parquet.decryption.factory.class", which is a bit cumbersome. Or maybe one parameter is enough - if the class implements both interfaces, and we'll load it as either encryption or decryption as needed (from writer or reader, respectively). If this works in Java (currently can't see why not), then could be the right thing to do.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
ggershinsky: In our implementation, a number of different things can go wrong when creating encryption (decryption) properties, ranging from a failure to communicate with a KMS system, to having a wrong access control token, to missing the required hadoop properties (such as a list of encrypted columns), to failure to write/read separate files that keep the key material. Some of these problems belong to the IO domain, some don't - but my impression was that most of Parquet API exceptions are the IOException, because Parquet is an IO library (file writing/reading).

ggershinsky: There is always a question of what the application (that calls Parquet API) can do differently, in an automatic fashion, for different exception types. In this case, I can't see such automatic differentiated response. If the creation of encr/decr properties fails, the only thing to do is to see exception message, and either fix the KMS comms (restart, fix network, ), or provide a valid access token (often manual procedure), or to configure the required parameters, or to fix the filesystem problems, etc.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.parquet.hadoop.util.ConfigurationUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * CryptoPropertiesFactory interface enables transparent activation of Parquet encryption.
+ *
+ * It's customized implementations produce encryption and decryption properties for each Parquet file, using the information
```
ggershinsky:
```suggestion
 * It's customized implementations produce encryption and decryption properties for each Parquet file, using the input information
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
```diff
@@ -0,0 +1,106 @@
...
+ * available in Parquet file writers and readers: file path, file extended schema (in writers only) - and also Hadoop
+ * configuration properties that can pass custom parameters required by a crypto factory. A factory implementation can
+ * use or ignore any of these parameters.
```
ggershinsky:
```suggestion
 * use or ignore any of these inputs.
```
was a bit misleading, my fault.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
gszadovszky: I see.
What I don't like in IOException in this case is that it is a checked exception so when we call these methods we have to handle them or let them thrown out of the library. I don't think we can do anything useful at handling so we have to throw an exception. We can either let through the IOException in the call stack which can be very painful or even impossible if we don't want to break the existing API or we can wrap the IOException into a runtime exception. 
Usually, the latter is the preferred way in parquet-mr. So why would we expect an IOException from our internal API (even thought it is implemented by our clients) if we will wrap it into our own runtime exception anyway.
I would suggest expecting a `ParquetEncodingException` or a `ParquetDecodingException` or even create specific ones for this case (by extending a `ParquetRuntimeException`).

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: Sounds good

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: Fixed

gszadovszky:
> > Last time I've forgot about one major thing. (Sorry for the late idea.)
> > parquet-mr library is usually used for either reading or writing at a time but not both. I think, it is not a good practice to expect both encryption and decryption properties while only one would be required.
> > What do you guys think?
> 
> On one hand, its certainly possible to split this into two interfaces, to make it fit the parquet-mr practices. On the other hand, a person working with encryption in eg Spark shell might want to both write and read the files. He would define two Hadoop config params for that, "parquet.encryption.factory.class" and "parquet.decryption.factory.class", which is a bit cumbersome. Or maybe one parameter is enough - if the class implements both interfaces, and we'll load it as either encryption or decryption as needed (from writer or reader, respectively). If this works in Java (currently can't see why not), then could be the right thing to do.

I like this idea. It should work just fine.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: fixed

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
ggershinsky: Sounds good. I'm fine with either option.

shangxinli:
> > Last time I've forgot about one major thing. (Sorry for the late idea.)
> > parquet-mr library is usually used for either reading or writing at a time but not both. I think, it is not a good practice to expect both encryption and decryption properties while only one would be required.
> > What do you guys think?
> 
> On one hand, its certainly possible to split this into two interfaces, to make it fit the parquet-mr practices. On the other hand, a person working with encryption in eg Spark shell might want to both write and read the files. He would define two Hadoop config params for that, "parquet.encryption.factory.class" and "parquet.decryption.factory.class", which is a bit cumbersome. Or maybe one parameter is enough - if the class implements both interfaces, and we'll load it as either encryption or decryption as needed (from writer or reader, respectively). If this works in Java (currently can't see why not), then could be the right thing to do.

I agree with Gidon. Aside from that, some ETL transform jobs need to decrypt raw data Parquet files and encrypt transformed Parquet files.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/CryptoPropertiesFactory.java
shangxinli: I will create ParquetCryptoRuntimeException extends ParquetRuntimeException

shangxinli:
@gszadovszky, The latest commit needs to be changed and I think it would be better to add unit tests now than TBD. You can hold off for now and I will let you know when it is ready for review.

shangxinli:
@gszadovszky,  I think the change is ready to review now.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/DecryptionPropertiesFactory.java
```diff
@@ -0,0 +1,86 @@
...
+ * It's customized implementations produce decryption properties for each Parquet file, using the input information
+ * available in Parquet file readers: file path and Hadoop configuration properties that can pass custom parameters
+ * required by a crypto factory. A factory implementation can use or ignore any of these inputs.
+ *
+ * The example usage could be as below.
+ *   1. Write a class to implement DecryptionPropertiesFactory.
+ *   2. Set configuration of "parquet.crypto.factory.class" with the fully qualified name of this class.
+ *      For example, we can set the configuration in SparkSession as below.
+ *         SparkSession spark = SparkSession
+ *                     .config("parquet.encryption.factory.class",
```
gszadovszky:
```suggestion
 *                     .config("parquet.crypto.factory.class",
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/EncryptionPropertiesFactory.java
```diff
@@ -0,0 +1,93 @@
...
+ * available in Parquet file writers: file path, file extended schema - and also Hadoop configuration properties that
+ * can pass custom parameters required by a crypto factory. A factory implementation can use or ignore any of these
+ * inputs.
+ *
+ * The example usage could be as below.
+ *   1. Write a class to implement EncryptionPropertiesFactory.
+ *   2. Set configuration of "parquet.crypto.factory.class" with the fully qualified name of this class.
+ *      For example, we can set the configuration in SparkSession as below.
+ *         SparkSession spark = SparkSession
+ *                     .config("parquet.encryption.factory.class",
```
gszadovszky:
```suggestion
 *                     .config("parquet.crypto.factory.class",
```

parquet-hadoop/src/main/java/org/apache/parquet/crypto/ParquetCryptoRuntimeException.java
```diff
@@ -0,0 +1,42 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.crypto;
+
+import org.apache.parquet.ParquetRuntimeException;
+
+/**
+ * thrown when a encryption or decryption operation problem occured
```
gszadovszky:
```suggestion
 * Thrown when an encryption or decryption operation problem is occurred
```

parquet-hadoop/src/test/java/org/apache/parquet/crypto/SampleDecryptionPropertiesFactory.java
```diff
@@ -0,0 +1,57 @@
...
+  public final static ColumnPath col1 = ColumnPath.fromDotString("col_1");
+  public final static byte[] col1Key = {0x02, 0x03, 0x4, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b,
+    0x0c, 0x0d, 0x0e, 0x0f, 0x10, 0x11};
+  public final static ColumnDecryptionProperties col1DecrProperties = ColumnDecryptionProperties.builder(col1.toDotString())
+    .withKey(col1Key).build();
+  public final static ColumnPath col2 = ColumnPath.fromDotString("col_2");
+  public final static byte[] col2Key = {0x03, 0x4, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c,
+     0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12};
+  public final static ColumnDecryptionProperties col2DecrProperties = ColumnDecryptionProperties.builder(col2.toDotString())
+    .withKey(col2Key).build();
```
gszadovszky: As per java code style the expected order should be `public static final` and the names should be upper case.

parquet-hadoop/src/test/java/org/apache/parquet/crypto/SampleEncryptionPropertiesFactory.java
```diff
@@ -0,0 +1,58 @@
...
+  public final static ColumnPath col1 = ColumnPath.fromDotString("col_1");
+  public final static byte[] col1Key = {0x02, 0x03, 0x4, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b,
+    0x0c, 0x0d, 0x0e, 0x0f, 0x10, 0x11};
+  public final static ColumnEncryptionProperties col1EncrProperties = ColumnEncryptionProperties.builder(
+    col1.toDotString()).withKey(col1Key).build();
+  public final static ColumnPath col2 = ColumnPath.fromDotString("col_2");
+  public final static byte[] col2Key = {0x03, 0x4, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c,
+     0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12};
+  public final static ColumnEncryptionProperties col2EncrProperties = ColumnEncryptionProperties.builder(
+    col2.toDotString()).withKey(col2Key).build();
```
gszadovszky: see above

parquet-hadoop/src/test/java/org/apache/parquet/crypto/SampleDecryptionPropertiesFactory.java
shangxinli: Fixed

parquet-hadoop/src/test/java/org/apache/parquet/crypto/SampleEncryptionPropertiesFactory.java
shangxinli: Fixed

PR #770, opened 2020-03-16
PARQUET-1815: Add union API to BloomFilter interface

garawalid:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -381,7 +381,7 @@ public class BlockSplitBloomFilter implements BloomFilter {
   }
 
   private boolean checkCompatibility(BlockSplitBloomFilter that) {
-    return this.maximumBytes == that.maximumBytes &&
+    return this.getBitsetSize() == that.getBitsetSize() &&
       this.hashStrategy == that.hashStrategy &&
       this.getAlgorithm() == that.getAlgorithm();
   }
@@ -392,8 +392,9 @@ public class BlockSplitBloomFilter implements BloomFilter {
       BlockSplitBloomFilter that = (BlockSplitBloomFilter) bloomFilter;
 
       if (!checkCompatibility(that)) {
-        throw new RuntimeException("Bloom filters are not compatible! Please check"
-          + "if they have the same maximum Bytes, hash strategy and algorithm.");
+        throw new IllegalArgumentException("Bloom filter is not compatible! "
+          + "Bitset size = " + that.getBitsetSize() + ", hash strategy = " + that.hashStrategy
+          + ", algorithm = " + that.getAlgorithm());
       }
 
       for (int i = 0; i < this.bitset.length; i++) {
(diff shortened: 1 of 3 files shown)


chenjunjiedada (approved):
LGTM

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -379,4 +379,32 @@ public long hash(float value) {
...
+          + "if they have the same maximum Bytes, hash strategy and algorithm.");
+      }
+
+      for (int i = 0; i < this.bitset.length; i++) {
+        this.bitset[i] |= that.bitset[i];
+      }
+
+      return new BlockSplitBloomFilter(this.bitset);
+    } else {
+      throw new RuntimeException("Bloom filter should be BlockSplitFilter");
```
gszadovszky: I know currently we only have `BlockSplitBloomFilter` but later on we might have others. Do we want to support different bloom filter implementation in one parquet file? If yes, do we want to support unioning different implementations?
I think, we shall document it properly in the interface and throw the exception accordingly. For example
```java
throw new IllegalArgumentException("Unsupported BloomFilter implementation to union with: " + bloomFilter.getClass());
```
or
```java
throw new UnsupportedOperationException("Unioning BloomFilter with type " + bloomFilter.getClass() + " is not supported yet");
```

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -379,4 +379,32 @@ public long hash(float value) {
...
+  }
+
+  @Override
+  public BloomFilter union(BloomFilter bloomFilter) {
+    if (bloomFilter instanceof BlockSplitBloomFilter) {
+      BlockSplitBloomFilter that = (BlockSplitBloomFilter) bloomFilter;
+
+      if (!checkCompatibility(that)) {
+        throw new RuntimeException("Bloom filters are not compatible! Please check"
+          + "if they have the same maximum Bytes, hash strategy and algorithm.");
```
gszadovszky: I would suggest using `IllegalArgumentException` instead.

I think, the message would be more useful if we add the actual values of max bytes, hash strategy etc. instead of suggesting to check them. When debugging from log files this message would not help. Maybe, `checkCompatibility` itself can throw the specific exception so you would know which value is incorrect and what would be the correct one.

gszadovszky (changes requested):

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
garawalid: I agree!

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
garawalid: You re right! Let me fix it.

gszadovszky (approved):

PR #773, opened 2020-03-18
PARQUET-1743: Add equals API to BloomFilter interface

garawalid:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -337,7 +337,7 @@ public class BlockSplitBloomFilter implements BloomFilter {
     if (object instanceof BlockSplitBloomFilter) {
       BlockSplitBloomFilter that = (BlockSplitBloomFilter) object;
       return Arrays.equals(this.bitset, that.bitset)
-        && this.maximumBytes == that.maximumBytes
+        && this.getAlgorithm() == that.getAlgorithm()
         && this.hashStrategy == that.hashStrategy;
     }
     return false;
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -27,6 +27,7 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.IntBuffer;
+import java.util.Arrays;
 
 /*
  * This Bloom filter is implemented using block-based Bloom filter algorithm from Putze et al.'s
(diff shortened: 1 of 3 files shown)


gszadovszky:
Do we still need this PR since PARQUET-1743 is resolved already?

garawalid:
The old PR was against the bloom-filter branch.  

Are we going to merge the bloom-filter branch to master at the end?  Or we will apply patches with new PR from bloom-filter to master like `PARQUET-41` that contains `PARQUET-1328`, `PARQUET-1391`, `PARQUET-1516`, `PARQUET-1660`?

gszadovszky:
Oh, I see. I've missed that.
Feature branches are usually used because the state of that feature is not stable enough to put to master. Since the feature is already merged to master and these jiras seem to be stable extensions of the feature I would merge them directly to master.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -328,6 +329,20 @@ public long hash(Object value) {
...
   }
 
+  @Override
+  public boolean equals(Object object) {
+    if (object == this) {
+      return true;
+    }
+    if (object instanceof BlockSplitBloomFilter) {
+      BlockSplitBloomFilter that = (BlockSplitBloomFilter) object;
+      return Arrays.equals(this.bitset, that.bitset)
```
shangxinli: Why only these 3 checks while there are several other properties like minimumBytes?

gszadovszky:
@garawalid, could you answer @shangxinli's comment so we can step forward and push this to master?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
garawalid: @shangxinli
Sorry for the late reply.
Other properties like `BYTES_PER_BLOCK` are for internal use, any small change will lead to a different `bitset`.
So I thought no need to check them.

shangxinli: It depends on the usage of this method. BYTES_PER_BLOCK is fine but why minimumBytes is not checked since maximumBytes is chcked?

garawalid: The [implementation](https://github.com/apache/parquet-mr/blob/bloom-filter/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java) of `bloom-filter` branch takes in the constructor only `maximumBytes`. You re right, it's a mistake.   
What about ignoring both the `maximumBytes` and the `minimumBytes`, since they will lead to a different size of bitset?

shangxinli: Again, it depends on the usage of this method. If it is a mistake, I would prefer to fix that mistake and come back with right approach.

garawalid: I’ll see what I can do!

garawalid: What about it this approach?

shangxinli: LGTM

PR #774, opened 2020-03-19
PARQUET-1821: Add 'column-size' command to parquet-cli and parquet-tools

shangxinli:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.cli.commands;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+import org.apache.avro.file.SeekableInput;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.cli.BaseCommand;
+import org.apache.parquet.cli.util.Formats;
+import org.apache.parquet.format.converter.ParquetMetadataConverter;
+import org.apache.parquet.hadoop.ParquetFileReader;
+import org.apache.parquet.hadoop.metadata.BlockMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.slf4j.Logger;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Parameters(commandDescription="Print the column sizes of a parquet file")
+public class ColumnSizeCommand extends BaseCommand {
+
+  public ColumnSizeCommand(Logger console) {
+    super(console);
+  }
+
+  @Parameter(description = "<parquet path>")
+  String target;
+
+  @Parameter(
+    names = {"-c", "--column", "--columns"},
+    description = "List of columns",
+    required = false)
+  List<String> columns;
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public int run() throws IOException {
+    Preconditions.checkArgument(target != null,
+      "A Parquet file is required.");
+
+    Path inputFile = new Path(target);
+    Map<String, Long> columnSizes = getColumnSizeInBytes(inputFile);
+    Map<String, Float> columnPercentage = getColumnPercentage(columnSizes);
+
+    // If user inputted columns, only print out size for those columns
+    if (columns != null && columns.size() > 0) {
+      for (String inputColumn : columns) {
+        long size = 0;
+        float percentage = 0;
+        for (String column : columnSizes.keySet()) {
+          if (column.startsWith(inputColumn)) {
+            size += columnSizes.get(column);
+            percentage += columnPercentage.get(column);
+          }
+        }
+        console.info(inputColumn + "->" + " Size In Bytes: " + size + " Size In Percentage: " + percentage);
+      }
+    } else {
+      for (String column : columnSizes.keySet()) {
+        console.info(column + "->" + " Size In Bytes: " + columnSizes.get(column)
+          + " Size In Percentage: " + columnPercentage.get(column));
+      }
+    }
+
+    return 0;
+  }
+
+  @Override
+  public List<String> getExamples() {
+    return Lists.newArrayList(
+        "# Print every column size in byte and percentage for a Parquet file",
+        "sample.parquet",
+        "sample.parquet col_1 col_2",
+        "sample.parquet col_1 col_2.sub_col_a"
+    );
+  }
+
+  // Make it public to allow some automation tools to call it
+  public Map<String, Long> getColumnSizeInBytes(Path inputFile) throws IOException {
+    Map<String, Long> colSizes = new HashMap<>();
+    ParquetMetadata pmd = ParquetFileReader.readFooter(new Configuration(), inputFile, ParquetMetadataConverter.NO_FILTER);
+
+    for (BlockMetaData block : pmd.getBlocks()) {
+      for (ColumnChunkMetaData column : block.getColumns()) {
+        String colName = column.getPath().toDotString();
+        colSizes.put(colName, column.getTotalSize() + colSizes.getOrDefault(colName, 0L));
+      }
+    }
+
+    return colSizes;
+  }
+
+  // Make it public to allow some automation tools to call it
+  public Map<String, Float> getColumnPercentage(Map<String, Long> colSizes) {
+    long totalSize = colSizes.values().stream().reduce(0L, Long::sum);
+    Map<String, Float> colPercentage = new HashMap<>();
+
+    for (Map.Entry<String, Long> entry : colSizes.entrySet()) {
+      colPercentage.put(entry.getKey(), ((float) entry.getValue()) / ((float) totalSize));
+    }
+
+    return colPercentage;
+  }
+}
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
@@ -68,25 +68,25 @@ public class ColumnSizeCommand extends BaseCommand {
 
     Path inputFile = new Path(target);
     Map<String, Long> columnSizes = getColumnSizeInBytes(inputFile);
-    Map<String, Float> columnPercentage = getColumnPercentage(columnSizes);
+    Map<String, Float> columnRatio = getColumnRatio(columnSizes);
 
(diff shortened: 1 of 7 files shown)


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
```diff
@@ -0,0 +1,132 @@
...
+  @SuppressWarnings("unchecked")
+  public int run() throws IOException {
+    Preconditions.checkArgument(target != null,
+      "A Parquet file is required.");
+
+    Path inputFile = new Path(target);
+    Map<String, Long> columnSizes = getColumnSizeInBytes(inputFile);
+    Map<String, Float> columnPercentage = getColumnPercentage(columnSizes);
+
+    // If user inputted columns, only print out size for those columns
```
gszadovszky:
```suggestion
    // If user defined columns, only print out size for those columns
```

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
```diff
@@ -0,0 +1,132 @@
...
+    if (columns != null && columns.size() > 0) {
+      for (String inputColumn : columns) {
+        long size = 0;
+        float percentage = 0;
+        for (String column : columnSizes.keySet()) {
```
gszadovszky: I think this behavior is misleading. What if the columns in the file are `id` and `identity` and you want to query the size of the column `id`?
I would suggest checking equality and simply print the result instead of summarizing.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
```diff
@@ -0,0 +1,132 @@
...
+    return 0;
+  }
+
+  @Override
+  public List<String> getExamples() {
+    return Lists.newArrayList(
+        "# Print every column size in byte and percentage for a Parquet file",
+        "sample.parquet",
+        "sample.parquet col_1 col_2",
+        "sample.parquet col_1 col_2.sub_col_a"
```
gszadovszky: I think, one of `-c`, `--column` or `--columns` is missing from here

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
```diff
@@ -0,0 +1,132 @@
...
+    return colSizes;
+  }
+
+  // Make it public to allow some automation tools to call it
+  public Map<String, Float> getColumnPercentage(Map<String, Long> colSizes) {
+    long totalSize = colSizes.values().stream().reduce(0L, Long::sum);
+    Map<String, Float> colPercentage = new HashMap<>();
+
+    for (Map.Entry<String, Long> entry : colSizes.entrySet()) {
+      colPercentage.put(entry.getKey(), ((float) entry.getValue()) / ((float) totalSize));
```
gszadovszky: This is not percentage but ratio.

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Map;
+
+public class ColumnSizeCommand extends ArgsOnlyCommand {
+
+  public static final String[] USAGE = new String[] {
+    "<input>",
+    "where <input> is the parquet file to calculate teh column size" +
+    "     [<column> ...] are the columns in the case sensitive dot format" +
+    "     to be calculated, for example a.b.c. If no columns are input, all the" +
+    "     columns will be printed out"
```
gszadovszky:
```suggestion
    "where <input> is the parquet file to calculate the column size" +
    "     [<column> ...] are the columns in the case sensitive dot format" +
    "     to be calculated, for example a.b.c. If no columns are set, all the" +
    "     columns will be printed out"
```

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
```diff
@@ -0,0 +1,121 @@
...
+
+    Map<String, Long> columnSizes = getColumnSizeInBytes(inputFile);
+    Map<String, Float> columnPercentage = getColumnPercentage(columnSizes);
+
+    if (args.size() > 1) {
+      for (String inputColumn : args.subList(1, args.size())) {
+        long size = 0;
+        float percentage = 0;
+        for (String column : columnSizes.keySet()) {
+          if (column.startsWith(inputColumn)) {
```
gszadovszky: See my comment at cli.

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
```diff
@@ -0,0 +1,121 @@
...
+    return colSizes;
+  }
+
+  // Make it public to allow some automation tools to call it
+  public Map<String, Float> getColumnPercentage(Map<String, Long> colSizes) {
+    long totalSize = colSizes.values().stream().reduce(0L, Long::sum);
+    Map<String, Float> colPercentage = new HashMap<>();
+
+    for (Map.Entry<String, Long> entry : colSizes.entrySet()) {
+      colPercentage.put(entry.getKey(), ((float) entry.getValue()) / ((float) totalSize));
```
gszadovszky: See my comment at cli.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
shangxinli: Good finding! What if we change to 'if (column.equals(inputColumn) || column.startsWith(inputColumn + "."))'? 

In some use cases, I see a parent field can have 50+ child fields. If a user wants to see that parent field, it would be cumbersome to input 50+ fields in the command.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
shangxinli: Got it.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
shangxinli: changed to ratio

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
shangxinli: Thanks.

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
shangxinli: Thanks.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
gszadovszky: It sounds good to me. It should be documented in the command line, though.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnSizeCommand.java
```diff
@@ -0,0 +1,134 @@
...
+
+  @Override
+  public List<String> getExamples() {
+    return Lists.newArrayList(
+        "# Print every column size in byte and ratio for a Parquet file",
+        "sample.parquet",
+        "sample.parquet -c col_1",
+        "sample.parquet -column col_2",
+        "sample.parquet -columns col_1 col_2",
+        "sample.parquet -columns col_1 col_2.sub_col_a"
```
gszadovszky:
```suggestion
        "sample.parquet --column col_2",
        "sample.parquet --columns col_1 col_2",
        "sample.parquet --columns col_1 col_2.sub_col_a"
```

shangxinli:
> Please, document the part column path match in the help.



> Please, document the part column path match in the help.
Thanks.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestColumnSizeCommand.java
```diff
@@ -0,0 +1,88 @@
...
+        g.add("DocId", rnd.nextLong());
+        g.add("Num", rnd.nextInt());
+        writer.write(g);
+      }
+    }
+
+    return file;
+  }
+
+  private static String createTempFile(String prefix) {
```
Fokko: Instead of having this file, we can use a JUnit rule: https://github.com/apache/parquet-mr/blob/c3e1a849d02042b324e4f72e1e4bba6c2a8da5bc/parquet-cli/src/test/java/org/apache/parquet/cli/commands/FileTest.java#L42

Then we can do: `temporaryFolder.newFile("test.parquet");`

Would be nice to keep this consistent

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnSizeCommand.java
```diff
@@ -0,0 +1,121 @@
...
+      for (String inputColumn : args.subList(1, args.size())) {
+        long size = 0;
+        float ratio = 0;
+        for (String column : columnSizes.keySet()) {
+          if (column.equals(inputColumn) || column.startsWith(inputColumn + ".")) {
+            size += columnSizes.get(column);
+            ratio += columnRatio.get(column);
+          }
+        }
+        Main.out.println(inputColumn + "->" + " Size In Bytes: " + size + " Size In Ratio: " + ratio);
```
Fokko:
```suggestion
        Main.out.println(inputColumn + "-> Size In Bytes: " + size + " Size In Ratio: " + ratio);
```

parquet-cli/src/test/java/org/apache/parquet/cli/commands/ColumnSizeCommandTest.java
```diff
@@ -0,0 +1,99 @@
...
+        writer.write(g);
+      }
+    }
+
+    return file;
+  }
+
+  private static String createTempFile(String prefix) {
+    try {
+      return Files.createTempDirectory(prefix).toAbsolutePath().toString() + "/test.parquet";
```
Fokko: See comment in TestColumnSizeCommand.java 👍

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestColumnSizeCommand.java
shangxinli: Sounds good!

parquet-cli/src/test/java/org/apache/parquet/cli/commands/ColumnSizeCommandTest.java
shangxinli: Sounds good!

parquet-cli/src/test/java/org/apache/parquet/cli/commands/ColumnSizeCommandTest.java
```diff
@@ -0,0 +1,91 @@
...
+  public void testColumnSize() throws Exception {
+    String inputFile = createParquetFile("input");
+    Map<String, Long> columnSizeInBytes = command.getColumnSizeInBytes(new Path(inputFile));
+    assertEquals(columnSizeInBytes.size(), 2);
+    assertTrue(columnSizeInBytes.get("DocId") > columnSizeInBytes.get("Num"));
+    Map<String, Float> columnRatio = command.getColumnRatio(columnSizeInBytes);
+    assertTrue(columnRatio.get("DocId") > columnRatio.get("Num"));
+  }
+
+  private String createParquetFile(String prefix) throws IOException {
```
gszadovszky: `prefix` is never used and so it is misleading.
Travis fails because `parquetFile()` uses the class name to create the file while there are two test methods invoking it. The file already exist after executing the second test method.
I would suggest correcting the super class implementation by generating a unique file name each time `parquetFile` is invoked.

parquet-tools/src/test/java/org/apache/parquet/tools/command/TestColumnSizeCommand.java
```diff
@@ -0,0 +1,94 @@
...
+  public void testColumnSize() throws Exception {
+    String inputFile = createParquetFile("input");
+    Map<String, Long> columnSizeInBytes = command.getColumnSizeInBytes(new Path(inputFile));
+    assertEquals(columnSizeInBytes.size(), 2);
+    assertTrue(columnSizeInBytes.get("DocId") > columnSizeInBytes.get("Num"));
+    Map<String, Float> columnRatio = command.getColumnRatio(columnSizeInBytes);
+    assertTrue(columnRatio.get("DocId") > columnRatio.get("Num"));
+  }
+
+  private String createParquetFile(String prefix) throws IOException {
```
gszadovszky: `prefix` is not used and so it is misleading.

PR #775, opened 2020-03-19
PARQUET-1381: add parquet block merging feature

brimzi:
This PR addresses [PARQUET-1381](https://issues.apache.org/jira/projects/PARQUET/issues/PARQUET-1381).

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -25,7 +25,6 @@ import static org.apache.parquet.hadoop.ParquetWriter.MAX_PADDING_SIZE_DEFAULT;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -34,7 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.zip.CRC32;
 
 import org.apache.hadoop.conf.Configuration;
@@ -909,9 +907,7 @@ public class ParquetFileWriter {
   }
 
   private List<ColumnChunkMetaData> getColumnsInOrder(BlockMetaData rowGroup, boolean dropColumns) {
-    return getColumnsInOrder(rowGroup, schema, dropColumns).stream()
-      .map(Entry::getValue)
-      .collect(Collectors.toList());
+    return getColumnsInOrder(rowGroup, schema, dropColumns);
   }
 
   /**
@@ -919,31 +915,31 @@ public class ParquetFileWriter {
    * @param schema the schema to use for column ordering
    * @param dropColumns whether we should drop columns that are not defined in the provided schema
    */
-  public static List<Entry<ColumnDescriptor, ColumnChunkMetaData>> getColumnsInOrder(BlockMetaData rowGroup,
-                                                                                     MessageType schema, boolean dropColumns) {
+  public static List<ColumnChunkMetaData> getColumnsInOrder(BlockMetaData rowGroup,
+                                                            MessageType schema, boolean dropColumns) {
     Map<String, ColumnChunkMetaData> columnsToCopy = new HashMap<>();
     for (ColumnChunkMetaData chunk : rowGroup.getColumns()) {
       columnsToCopy.put(chunk.getPath().toDotString(), chunk);
     }
 
-    List<Entry<ColumnDescriptor, ColumnChunkMetaData>> columnsInOrder = new ArrayList<>();
+    List<ColumnChunkMetaData> columnsInOrder = new ArrayList<>();
 
     for (ColumnDescriptor descriptor : schema.getColumns()) {
       String path = ColumnPath.get(descriptor.getPath()).toDotString();
       ColumnChunkMetaData chunk = columnsToCopy.remove(path);
       if (chunk != null) {
-        columnsInOrder.add(new AbstractMap.SimpleEntry<>(descriptor, chunk));
+        columnsInOrder.add(chunk);
       } else {
         throw new IllegalArgumentException(String.format(
-            "Missing column '%s', cannot copy row group: %s", path, rowGroup));
+          "Missing column '%s', cannot copy row group: %s", path, rowGroup));
       }
     }
 
     // complain if some columns would be dropped and that's not okay
     if (!dropColumns && !columnsToCopy.isEmpty()) {
       throw new IllegalArgumentException(String.format(
-          "Columns cannot be copied (missing from target schema): %s",
-          String.join(", ", columnsToCopy.keySet())));
+        "Columns cannot be copied (missing from target schema): %s",
+        String.join(", ", columnsToCopy.keySet())));
     }
     return columnsInOrder;
   }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -25,6 +25,7 @@ import static org.apache.parquet.hadoop.ParquetWriter.MAX_PADDING_SIZE_DEFAULT;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -33,13 +34,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.zip.CRC32;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.Version;
 import org.apache.parquet.bytes.BytesInput;
@@ -70,9 +71,9 @@ import org.apache.parquet.internal.column.columnindex.OffsetIndexBuilder;
 import org.apache.parquet.internal.hadoop.metadata.IndexReference;
 import org.apache.parquet.io.InputFile;
 import org.apache.parquet.io.OutputFile;
-import org.apache.parquet.io.SeekableInputStream;
 import org.apache.parquet.io.ParquetEncodingException;
 import org.apache.parquet.io.PositionOutputStream;
+import org.apache.parquet.io.SeekableInputStream;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.PrimitiveType;
 import org.apache.parquet.schema.TypeUtil;
@@ -840,32 +841,7 @@ public class ParquetFileWriter {
                              boolean dropColumns) throws IOException {
     startBlock(rowGroup.getRowCount());
 
-    Map<String, ColumnChunkMetaData> columnsToCopy =
-        new HashMap<String, ColumnChunkMetaData>();
-    for (ColumnChunkMetaData chunk : rowGroup.getColumns()) {
-      columnsToCopy.put(chunk.getPath().toDotString(), chunk);
-    }
-
-    List<ColumnChunkMetaData> columnsInOrder =
-        new ArrayList<ColumnChunkMetaData>();
-
-    for (ColumnDescriptor descriptor : schema.getColumns()) {
-      String path = ColumnPath.get(descriptor.getPath()).toDotString();
-      ColumnChunkMetaData chunk = columnsToCopy.remove(path);
-      if (chunk != null) {
-        columnsInOrder.add(chunk);
-      } else {
-        throw new IllegalArgumentException(String.format(
-            "Missing column '%s', cannot copy row group: %s", path, rowGroup));
-      }
-    }
-
-    // complain if some columns would be dropped and that's not okay
-    if (!dropColumns && !columnsToCopy.isEmpty()) {
-      throw new IllegalArgumentException(String.format(
-          "Columns cannot be copied (missing from target schema): %s",
-          String.join(", ", columnsToCopy.keySet())));
-    }
+    List<ColumnChunkMetaData> columnsInOrder = getColumnsInOrder(rowGroup, dropColumns);
 
     // copy the data for all chunks
     long start = -1;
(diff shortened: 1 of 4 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -919,6 +895,59 @@ public void appendRowGroup(SeekableInputStream from, BlockMetaData rowGroup,
...
+    }
+
+    List<Entry<ColumnDescriptor, ColumnChunkMetaData>> columnsInOrder = new ArrayList<>();
+
+    for (ColumnDescriptor descriptor : schema.getColumns()) {
+      String path = ColumnPath.get(descriptor.getPath()).toDotString();
+      ColumnChunkMetaData chunk = columnsToCopy.remove(path);
+      if (chunk != null) {
+        columnsInOrder.add(new AbstractMap.SimpleEntry<>(descriptor, chunk));
+      } else {
```
shangxinli: Why we would allow columnsToCopy to have null at first place?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -919,6 +895,59 @@ public void appendRowGroup(SeekableInputStream from, BlockMetaData rowGroup,
...
+   * @param maxRowGroupSize the maximum size in bytes the new created groups can be
+   * @param useV2Writer whether to use a V2 encoding based writer when rewriting dictionary encoded pages
+   * @param compression compression to use when writing
+   * @throws IOException
+   */
+  public void mergeRowGroups(List<InputFile> inputFiles, long maxRowGroupSize, boolean useV2Writer, CompressionCodecName compression) throws IOException {
+    new RowGroupMerger(schema, compression, useV2Writer).merge(inputFiles, maxRowGroupSize, this);
+  }
+
+  private List<ColumnChunkMetaData> getColumnsInOrder(BlockMetaData rowGroup, boolean dropColumns) {
```
shangxinli: Converting to Entry and then back to List would add some cost, comparing with the original implementation. Can you keep it as List<ColumnChunkMetaData> in line of 922 for original use and wrap it up with List<Entry<ColumnDescriptor, ColumnChunkMetaData>> for new methods to use?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+        }
+        return rowCount;
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      }
+    }
+  }
+
+  private static class SizeEstimator {
+    private final boolean targetCompressed;//is destination block compressed
```
shangxinli: Move the comments above

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    private double factor;
+
+    private SizeEstimator(boolean targetCompressed) {
+      this.targetCompressed = targetCompressed;
+    }
+
+    private long estimate(BlockMetaData blockMeta) {
+      if (factor <= 0) {
+        return targetCompressed ? blockMeta.getCompressedSize() : blockMeta.getTotalByteSize();//very naive estimator
```
shangxinli: Move the comments above

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+      }
+      return (long) (factor * blockMeta.getTotalByteSize());
+    }
+
+    private void update(long actualBytes, BlockMetaData blockMeta) {
+      factor = (double) actualBytes / blockMeta.getTotalByteSize();// basically remembers last ratio
```
shangxinli: Move the comments above

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    }
+
+    private long estimate(BlockMetaData blockMeta) {
+      if (factor <= 0) {
+        return targetCompressed ? blockMeta.getCompressedSize() : blockMeta.getTotalByteSize();//very naive estimator
+      }
+      return (long) (factor * blockMeta.getTotalByteSize());
+    }
+
+    private void update(long actualBytes, BlockMetaData blockMeta) {
```
shangxinli: 'factor' should be mandatorily initialized before it can be used. Would it be better to move it to the constructor?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
shangxinli: Move the comments above

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    private MutableMergedBlock(MessageType schema) {
+      this.schema = schema;
+    }
+
+    @Override
+    public long rowCount() {
+      return recordCount;
+    }
+
+    private long getCompressedBytesSize() {
```
shangxinli: I don't have option should we use  getCompressedBytesSize() or getCompressedSize(), but since BlockMetaData already uses getCompressedSize(), can we just reuse the name for consistency?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+            mergedBlock = null;
+          }
+
+          if (mergedBlock == null && estimator.estimate(blockMeta) > maxRowGroupSize) {
+            //save it directly without re encoding it
+            saveBlockTo(ReadOnlyMergedBlock.of(blockMeta, group, schema, compressor), writer);
+            continue;
+          }
+
+          if (mergedBlock == null) {
```
shangxinli: Should this block (line 116 - line 118) be moved above line 105, so that you can avoid checking mergedBlock != null in line 105 and line 110?

shangxinli (commented):
I just started with some piece of the code but not done yet. I will lock down sometime later to have another look.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
brimzi: It is updated on every iteration so that the estimation can use the most recent calculated value. On creation of the estimator we do not have any info to calculate the factor

brimzi:
Thank you @shangxinli, in the mean time I have address your comments above.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+
+          long sizeBeforeMerge = mergedBlock.getCompressedBytesSize();
+          mergedBlock.merge(blockMeta, group);
+          //update our estimator
+          long currentBlockEffect = mergedBlock.getCompressedBytesSize() - sizeBeforeMerge;
+          estimator.update(currentBlockEffect, blockMeta);
+        }
+      }
+    }
+    if (mergedBlock != null)
```
shangxinli, line 128: add '{' and '}' always

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+      saveBlockTo(mergedBlock, writer);
```
shangxinli, line 129: Please add 'mergeBlock = null' after line 129.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+        writer.startColumn(col.getColumnDesc(), col.getValueCount(), col.getCompression());
+
+        col.writeDictionaryPageTo(writer);
+        col.writeDataPagesTo(writer);
+
+        writer.endColumn();
+      }
+
+      writer.endBlock();
+    } catch (IOException e) {
```
shangxinli, line 147: Why we convert to UncheckedIOException? The caller merge() already throws IOException

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+          estimator.update(currentBlockEffect, blockMeta);
+        }
+      }
+    }
+    if (mergedBlock != null)
+      saveBlockTo(mergedBlock, writer);
+  }
+
+
+  private static void saveBlockTo(MergedBlock block, ParquetFileWriter writer) {
```
shangxinli, line 133: As bloom filter feature is implemented, I guess we need to take care of that too.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    return builder.build();
+  }
+
+  private BytesInput compress(BytesInput bytes) {
+    return compress(bytes, compressor);
+  }
+
+  private static BytesInput compress(BytesInput bytes, CompressionCodecFactory.BytesInputCompressor compressor) {
+    try {
+      return BytesInput.copy(compressor.compress(bytes));//we copy as some compressors use shared memory
```
shangxinli: Move comments to above

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
```
shangxinli, line 169: Same comments as above, why converting to UncheckedIOException()?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+              page.getBytes(), page.getStatistics(), page.getIndexRowCount().get(), page.getRlEncoding(),
+              page.getDlEncoding(), page.getValueEncoding());
+
+          } else {
+            writer.writeDataPage(page.getValueCount(), page.getUncompressedSize(),
+              page.getBytes(), page.getStatistics(), page.getRlEncoding(),
+              page.getDlEncoding(), page.getValueEncoding());
+          }
+        } catch (IOException e) {
+          throw new UncheckedIOException(e);
```
shangxinli, line 210: Same question why converting to UncheckedIOException?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
+/*
```
shangxinli, line 1: If this file is only used by tools, can we move it out of parquet-hadoop project?

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -63,28 +104,70 @@ public MergeCommand() {
...
+      .anyMatch(chunk -> {
+        EncodingStats stats = chunk.getEncodingStats();
+        return stats != null && stats.usesV2Pages();
+      });
+
+    List<InputFile> inputFiles = files.stream().map(f -> {
+      try {
+        return HadoopInputFile.fromPath(f, conf);
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
```
shangxinli, line 158: I don't see why converting the exception is needed.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -63,28 +104,70 @@ public MergeCommand() {
...
-
-    // Merge data
+    ParquetMetadata parquetMetadata = mergedMetadata(files);
     ParquetFileWriter writer = new ParquetFileWriter(conf,
-            mergedMeta.getSchema(), outputFile, ParquetFileWriter.Mode.CREATE);
+      parquetMetadata.getFileMetaData().getSchema(), outputFile, ParquetFileWriter.Mode.CREATE);
+    PrintWriter stdOut = new PrintWriter(Main.out, true);
+
+    if (mergeBlocks) {
+      long maxRowGroupSize = options.hasOption('l')? Long.parseLong(options.getOptionValue('l')) * 1024 * 1024 : DEFAULT_BLOCK_SIZE;
```
shangxinli, line 132: Use a constant variable to represent '1024 * 1024 : DEFAULT_BLOCK_SIZE'.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -63,28 +104,70 @@ public MergeCommand() {
...
     // Merge schema and extraMeta
-    FileMetaData mergedMeta = mergedMetadata(inputFiles);
-    PrintWriter out = new PrintWriter(Main.out, true);
-
-    // Merge data
+    ParquetMetadata parquetMetadata = mergedMetadata(files);
     ParquetFileWriter writer = new ParquetFileWriter(conf,
-            mergedMeta.getSchema(), outputFile, ParquetFileWriter.Mode.CREATE);
+      parquetMetadata.getFileMetaData().getSchema(), outputFile, ParquetFileWriter.Mode.CREATE);
+    PrintWriter stdOut = new PrintWriter(Main.out, true);
```
shangxinli, line 129: Unlike cat/dump command, these kinds of parquet tools could be used by some automation jobs. In that case, using log4j might be better than PrintWriter.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -919,6 +895,59 @@ public void appendRowGroup(SeekableInputStream from, BlockMetaData rowGroup,
...
+  /**
+   * Merges adjacent row groups in the supplied files while maintaining that the new groups is no more than the specified
+   * maxRowGroupSize
+   * @param inputFiles input files to merge
+   * @param maxRowGroupSize the maximum size in bytes the new created groups can be
+   * @param useV2Writer whether to use a V2 encoding based writer when rewriting dictionary encoded pages
+   * @param compression compression to use when writing
+   * @throws IOException
+   */
+  public void mergeRowGroups(List<InputFile> inputFiles, long maxRowGroupSize, boolean useV2Writer, CompressionCodecName compression) throws IOException {
```
shangxinli, line 905: I am not sure we want to release a public API to PartquetFileWriter which is heavily used.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -18,27 +18,37 @@
...
+import org.apache.parquet.hadoop.util.HiddenFileFilter;
+import org.apache.parquet.io.InputFile;
 import org.apache.parquet.tools.Main;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.List;
-
 public class MergeCommand extends ArgsOnlyCommand {
```
shangxinli, line 47: Add tests for it

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
shangxinli, line 133: why static?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+      builder.withWriterVersion(ParquetProperties.WriterVersion.PARQUET_2_0);
+    }
+    return builder.build();
+  }
+
+  private BytesInput compress(BytesInput bytes) {
+    return compress(bytes, compressor);
+  }
+
+  private static BytesInput compress(BytesInput bytes, CompressionCodecFactory.BytesInputCompressor compressor) {
```
shangxinli, line 164: Why we have so many static methods?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
```diff
@@ -0,0 +1,634 @@
...
+    }
+
+    void addPage(DataPage data, Dictionary pageDictionary) {
+      DataPage recodePage = recodePage(data, pageDictionary);
+      compressedSizeAccumulator.accept((long) recodePage.getCompressedSize());
+      valueCount += recodePage.getValueCount();
+      pages.add(recodePage);
+    }
+
+    DataPage recodePage(DataPage data, Dictionary pageDictionary) {
```
shangxinli, line 498: This is a pretty long method. Consider refactoring with some helper function.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
brimzi: Good catch, This somehow was carried over after refactoring. I have removed it

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
brimzi: Since we are using page level APIs of ParquetFileWriter(i.e _writeDataPage_ and _writeDataPageV2_), that is handled for us under the hood.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/RowGroupMerger.java
brimzi: I am forced to wrap it here because in the Visitor interface this method does not throw any exception, hence I am supposed to swallow all checked exceptions in this implementation

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
brimzi: Same issue here as above, Functional Interface forces us to do so

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
brimzi: That is a good question. My thinking is that this method is similar(an improvement) to _appendRowGroup_ which is a public API. In most use cases the user of _appendRowGroup_ actually wants the functionality provided by this method.  But it boils down to if _appendRowGroup_ really belonged here in the first place. I have no strong position on this, do you think I should not expose this new method?

shangxinli: I prefer not to unless you strongly think we should.

MaheshGPai:
@shangxinli, @brimzi  Is this PR still active? I can pick this and take this forward if required.
We are looking for the capability to merge rowgroups as well.

brimzi:
@MaheshGPai, this PR stalled. You can pick it up if you like, but I am not sure if this is valid anymore

MaheshGPai:
The tools code has moved over the time. But the fundamental change is still valid. Thanks for that. I'll float a new PR with the change. Hope thats fine.

MaheshGPai:
@shangxinli @brimzi Created https://github.com/apache/parquet-mr/pull/1121

PR #776, opened 2020-04-01
PARQUET-1229: Parquet MR encryption

ggershinsky:

### Jira

### Tests

### Commits

### Documentation


ggershinsky:
@gszadovszky The encryption pull request is submitted, with changes in "hadoop" and "column" modules. Travis fails on "The job exceeded the maximum log length, and has been terminated". All tests up that point pass successfully (including the tests in hadoop and column modules).

gszadovszky:
FYI: Created PARQUET-1832. I'll try to fix this issue ASAP.

ggershinsky:
Thank you

shangxinli:
Is this ready to review? Since there is no comment yet, can you squash it to one single commit to make the review easier?

ggershinsky:
preferably reviewed after the Travis fix is in. @gszadovszky @shangxinli can you apply #777 to the encryption branch.

ggershinsky:
> can you squash it to one single commit to make the review easier?

This can be reviewed as a single commit at
https://github.com/apache/parquet-mr/pull/776/files

ggershinsky:
the PR is ready for review

shangxinli:
> preferably reviewed after the Travis fix is in. @gszadovszky @shangxinli can you apply #777 to the encryption branch.

It seems I still don't have permission to do so. @gszadovszky, I did 'git cherry-pick' and then git push to encryption branch', but got permission error. Do I have to create a PR for it?

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
```diff
@@ -90,6 +102,10 @@
 
   // Update last two bytes with new page ordinal (instead of creating new page AAD from scratch)
   public static void quickUpdatePageAAD(byte[] pageAAD, short newPageOrdinal) {
+    if (newPageOrdinal < 0) {
```
shangxinli, line 105: Since this is a public method, can we validate pageAAD also?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -463,14 +486,29 @@ ConvertedType convertToConvertedType(LogicalTypeAnnotation logicalTypeAnnotation
     }
   }
 
-  private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGroups, BlockMetaData block) {
+  private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGroups, BlockMetaData block,
+      InternalFileEncryptor fileEncryptor) throws IOException {
+    
     //rowGroup.total_byte_size = ;
```
shangxinli, line 492: Can we remove this?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -463,14 +486,29 @@ ConvertedType convertToConvertedType(LogicalTypeAnnotation logicalTypeAnnotation
     }
   }
 
-  private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGroups, BlockMetaData block) {
+  private void addRowGroup(ParquetMetadata parquetMetadata, List<RowGroup> rowGroups, BlockMetaData block,
```
shangxinli, line 489: Should we keep original addRowGroup() intact and just add a new one with InternalFileEncryptor? There is not much duplicate code if doing so and we can refactor the existing code with helper functions.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1185,70 +1275,189 @@ static long getOffset(ColumnChunk columnChunk) {
...
       public FileMetaData visit(RangeMetadataFilter filter) throws IOException {
-        return filterFileMetaDataByMidpoint(readFileMetaData(from), filter);
+        return filterFileMetaDataByMidpoint(readFileMetaData(from, footerDecryptor, encryptedFooterAAD), filter);
       }
     });
     LOG.debug("{}", fileMetaData);
-    ParquetMetadata parquetMetadata = fromParquetMetadata(fileMetaData);
+    
+    if (!encryptedFooter && null != fileDecryptor) {
+      if (!fileMetaData.isSetEncryption_algorithm()) { // Plaintext file
```
shangxinli, line 1338: Move comments up

gszadovszky:
> > preferably reviewed after the Travis fix is in. @gszadovszky @shangxinli can you apply #777 to the encryption branch.
> 
> It seems I still don't have permission to do so. @gszadovszky, I did 'git cherry-pick' and then git push to encryption branch', but got permission error. Do I have to create a PR for it?

@shangxinli, I've done a rebase on the branch `encryption`. Maybe, that's why you were unable to push a commit because after the rebase it would not be a fast forward commit (would need a force-push).
(Because of the rebase the Travis fix is already in.)
If you have any problems with manual push to the repo, let's bring it offline. I'm happy to help.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
ggershinsky: Ok, I'll add this.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
ggershinsky: encryption is isolated there, with _if (null != fileEncryptor)_  and _if (encryptMetaData)_ - similar to the _if (columnIndexRef != null)_  and _if (offsetIndexRef != null)_ in the same function.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
ggershinsky: is there a requirement in the code formatting rules in this community to keep comments in separate lines?

shangxinli: No. I see most of them on up line but a few on the same line. It is not a must.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: Not all the changes are isolated. Generally, adding 'if/else' will add diverge the code and add the complexity. One other thing is regression thinking. If fileEncryptor is null, which would be most of the case, then it just executes the existing method without change. It would be less error prone.

shangxinli:
@ggershinsky Do you have time to review the code?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
ggershinsky: Many functions can be run either with or without encryption. Duplicating them will result in hundreds or thousands of duplicate code lines. This will make code maintenance (changes/fixes)  a headache. Instead, we isolate encryption with _if_ switches, without duplicating the existing code.

The same goes for other recent new features (column indexes and bloom filters) - they are isolated with an _if_ switch, instead of code duplication. See _if (columnIndexRef != null) and if (offsetIndexRef != null)_ in this addRowGroup function.

ggershinsky:
> @ggershinsky Do you have time to review the code?

You probably meant @gszadovszky

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
gszadovszky: Actually, based on parquet-mr README:
> Generally speaking, stick to the Sun Java Code Conventions

Based on the [related section](https://www.oracle.com/java/technologies/javase/codeconventions-comments.html#286) both should be fine.

gszadovszky, line 1338: Is `plaintext` a usual term for _un-encrypted_ files? I don't really like it but fine if it is commonly used in that sense. (Plaintext files for me are the `*.txt` files.)

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1185,70 +1275,189 @@ static long getOffset(ColumnChunk columnChunk) {
...
+        fileDecryptor.setPlaintextFile();
+        // Done to detect files that were not encrypted by mistake
+        if (!fileDecryptor.plaintextFilesAllowed()) {
+          throw new IOException("Applying decryptor on plaintext file");
```
gszadovszky, line 1342: I think, `ParquetCryptoRuntimeException` would fit better here.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
```diff
@@ -68,19 +67,32 @@
 
   public static byte[] createModuleAAD(byte[] fileAAD, ModuleType moduleType, 
       short rowGroupOrdinal, short columnOrdinal, short pageOrdinal) {
```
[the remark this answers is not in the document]
ggershinsky: The links to the discussion on this,
https://github.com/apache/parquet-format/pull/114#discussion_r234022579
https://github.com/apache/parquet-format/pull/114#discussion_r232941138
http://mail-archives.apache.org/mod_mbox/parquet-dev/201901.mbox/%3CCAO4re1kM4xGMNT4CGrjvA43t-QgUmUwLMskTJfd8ivgCfF8rSw%40mail.gmail.com%3E

The parquet-cpp approach to this is to allow for any number of row groups in files without encryption, and to limit it to 32K in encrypted files,
https://github.com/apache/arrow/commit/0c5168cf203ddf94dff01c178d649853323acbb8
"While writing files with so many row groups is a bad idea, people will still do it... This .. enables reading the many-row-group files again. Files with encrypted row group metadata with that many row groups cannot be read"

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndex.java
```diff
@@ -49,6 +49,13 @@
    * @return the index of the first row in the page
    */
   public long getFirstRowIndex(int pageIndex);
+  
+  /**
+   * @param pageIndex
+   *         the index of the page
+   * @return the original ordinal of the page in the column chunk
+   */
+  public short getPageOrdinal(int pageIndex);
```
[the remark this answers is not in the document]
ggershinsky: The background discussion is here,
https://github.com/apache/parquet-mr/pull/776#discussion_r427743861

In the case of pages, encryption becomes an order (or two orders) of magnitude slower if the pages are small. Basically, the hardware acceleration does not kick in with small pages (and there are additional problems). This is another reason not to allow more than 32K pages in a chunk.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
ggershinsky: In cryptography, plaintext is an opposite of ciphertext (the result of plaintext encryption).

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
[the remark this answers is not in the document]
ggershinsky: Also - like with the page numbers in the previous comment, having too many row groups will adversely affect encryption performance. There are per-rowgroup encryption operations, always performed on small  buffers - therefore, very slow. Having dozens/hundreds of thousands (or more) of them will significantly affect the overall encryption time of a file. Moreover, having lots of row groups might lead to having smaller data pages, which decreases the performance further.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndex.java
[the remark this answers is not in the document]
ggershinsky: Plus - the page headers are also encrypted. These are small, so the hardware acceleration is not applied on them. Having dozens/hundreds of thousands (or more) of page headers will significantly affect the overall encryption time of a file.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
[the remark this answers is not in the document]
ggershinsky: Still, the variables in this function (and elsewhere) don't have to be `short`. After looking at the code, it seems `int`s are better suited for managing and checking these parameters (and for enabling any values in unencrypted files). I'll make this change.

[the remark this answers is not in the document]
gszadovszky: I understand that large numbers of pages/row groups or columns would lead to significant performance drawbacks but it should not limit what the spec allows otherwise.

Since it is discussed and approved already, I am fine with using `short` values for these. What I would suggest adding though is to have the conversion from `int` to `short` centralized and and have specific error messages so it is clear that the limit reached is a hard limit for the encryption feature. Also, if we will publish any description/example for the encryption feature these limitations shall be listed there.

One more thing: the check of `intValue > Short.MAX_VALUE` is not complete. In case of `intValue` is negative the cast may result in a valid positive `short` value. I would suggest using `(short) intValue != intValue` instead.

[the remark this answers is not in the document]
ggershinsky: Thanks for the value checking tip, I'll update the code to use it. As for a centralization - I think this function (`createModuleAAD`) is the right place. In the encryption feature, ordinals are used only for integrity verification - performed via AADs, which are calculated here for both encryption and decryption. Everywhere in the code, the ordinals will be an `int`. Since the `createModuleAAD` is called only for encrypted files, an exception will be thrown only for them (if an ordinal exceeds the `Short.MAX_VALUE`).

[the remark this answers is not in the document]
gszadovszky: Sounds good to me.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1021,18 +1148,44 @@ DictionaryPage readDictionary(ColumnChunkMetaData meta) throws IOException {
         !meta.getEncodings().contains(Encoding.RLE_DICTIONARY)) {
       return null;
     }
+    /** TODO Gabor - can be replaced with this?:
+    if (!meta.hasDictionaryPage()) {
+      return null;
+    } */
```
gszadovszky, line 1154: I think, we can replace it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -442,7 +460,13 @@ static ParquetMetadata readSummaryMetadata(Configuration configuration, Path bas
    */
   @Deprecated
   public static final ParquetMetadata readFooter(Configuration configuration, Path file) throws IOException {
-    return readFooter(configuration, file, NO_FILTER);
+    return readFooter(configuration, file, getDecryptionProperties(file, configuration));
+  }
+
+  @Deprecated
+  public static final ParquetMetadata readFooter(Configuration configuration, Path file,
```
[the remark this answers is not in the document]
ggershinsky: These are encrypting versions of the existing `readFooter` functions, already marked as deprecated, but still actively used (eg in parquet-cli and in Spark). The deprecation comment says "@ deprecated will be removed in 2.0.0". Since we are not at parquet 2.0 yet, I've marked the encrypting versions of these functions as deprecated too, in order not to forget to handle them when working on parquet-2.0.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -705,22 +797,39 @@ public ParquetFileReader(Configuration conf, Path file, ParquetMetadata footer)
...
+
+  public ParquetFileReader(InputFile file, ParquetReadOptions options, 
+      FileDecryptionProperties fileDecryptionProperties) throws IOException {
     this.converter = new ParquetMetadataConverter(options);
     this.file = file;
     this.f = file.newStream();
     this.options = options;
+    if ((null == fileDecryptionProperties) && (file instanceof HadoopInputFile)) {
+      HadoopInputFile hadoopFile = (HadoopInputFile) file;
+      fileDecryptionProperties = getDecryptionProperties(hadoopFile.getPath(), hadoopFile.getConfiguration());
```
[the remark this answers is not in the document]
ggershinsky: We need the file `Path` in order to create the file decryption properties. `ParquetReadOptions` and `HadoopReadOptions` don't keep/handle the file paths.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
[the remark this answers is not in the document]
gszadovszky: So, all of these new methods are used inside parquet-mr? If not, then I don't think we need them. If yes, then please, try to refactor the caller part to use the non-deprecated ones instead. If it does not require too much effort.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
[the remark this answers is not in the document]
gszadovszky: `ParquetReadOptions` was created to carry all the required properties for reading a parquet file. If the `Path` is necessary for the decryption then we might add it to the options as well. If we decide to not to add it still, I would use the options object to carry any other decryption properties.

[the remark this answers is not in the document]
ggershinsky: Sounds good. I'll check what can be done here.
Also, will check how the deprecated readFooter functions can be handled for encryption.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
```diff
@@ -241,6 +280,8 @@ public PrimitiveType getPrimitiveType() {
    * @return the stats for this column
    */
   abstract public Statistics getStatistics();
+  
+  abstract public void decryptIfNeededed();
```
gszadovszky, line 284: The design seems to require this method to be only invoked inside this package. So please, make it _package private_.

You may have an empty implementation here so you don't need to add the empty implementations in the Int/Long classes.

Also, a bit too many "_ed_"s:
```suggestion
  abstract public void decryptIfNeeded();
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
```diff
@@ -165,10 +200,10 @@ protected static boolean positiveLongFitsInAnInt(long value) {
     return (value >= 0) && (value + Integer.MIN_VALUE <= Integer.MAX_VALUE);
   }
 
-  private final EncodingStats encodingStats;
+  protected EncodingStats encodingStats;
```
gszadovszky, line 203: These are accessed only inside the same class so you may keep them `private`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
```diff
@@ -141,11 +159,28 @@ public static ColumnChunkMetaData get(
           totalUncompressedSize);
     }
   }
+  
+  public static ColumnChunkMetaData getWithEncryptedMetadata(ParquetMetadataConverter parquetMetadataConverter, ColumnPath path,
```
gszadovszky, line 163: I would expect some comments here why we need this object (to decrypt this metadata lazily instead of simply decrypt it at reading).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestBloomEncryption.java
```diff
@@ -0,0 +1,313 @@
+/*
```
gszadovszky, line 1: I guess, this test and the one for column indexes were created by copy-pasting the original tests and adding the encryption. Since we already use parameterized testing in both I would suggest keeping the original tests and adding another dimension for _plain_/_encrypted_ or similar. This way we would have 4 runs for each tests: (V1 with plain), (V1 with encryption), (V2 with plain) and (V2 with encryption).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestEncryption.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.MessageType;
+import org.junit.rules.TemporaryFolder;
+
+public class TestEncryption {
+
+  @Test
+  public void test() throws Exception {
+    Configuration conf = new Configuration();
+    Path root = new Path("target/tests/TestEncryption/");
```
gszadovszky, line 56: I would suggest creating a temporary dir in temporary space instead of having a relative path depending on the current directory. This way it is a bit error prone.
(There are a couple of ways to create and cleanup temporary directories in junit.)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestEncryption.java
```diff
@@ -0,0 +1,215 @@
...
+            .append("binary_field", "test" + i)
+            .append("int32_field", 32)
+            .append("int64_field", 64l)
+            .append("boolean_field", true)
+            .append("float_field", 1.0f)
+            .append("double_field", 2.0d)
+            .append("flba_field", "foo")
+            .append("int96_field", Binary.fromConstantByteArray(new byte[12])));
+      }
+      writer.close();
```
gszadovszky, line 188: Consider using try-with-resources. That construct would be less error prone (e.g. closing the writer/reader in case of an exception occurs).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestEncryption.java
```diff
@@ -0,0 +1,215 @@
...
+            group.getInt96("int96_field",0));
+      }
+      reader.close();
+    }
+    enforceEmptyDir(conf, root);
+  }
+
+
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
```
gszadovszky, line 213: Yep, that's one good way I was talking about but you should use `temp` instead of hardcoding the directory at the beginning.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestEncryption.java
```diff
@@ -0,0 +1,215 @@
...
+    for (int encryptionMode = 0; encryptionMode < numberOfEncryptionModes; encryptionMode++) {
+      System.out.println("MODE: "+encryptionMode);
+      
+      Path file = new Path(root, "m_" + encryptionMode + ".parquet.encrypted");
+      ParquetWriter<Group> writer = new ParquetWriter<Group>(
+          file,
+          new GroupWriteSupport(),
+          UNCOMPRESSED, 1024, 1024, 512, true, false, ParquetWriter.DEFAULT_WRITER_VERSION, conf, 
+          encryptionPropertiesList[encryptionMode]);
+      for (int i = 0; i < 1000; i++) {
```
gszadovszky, line 176: I don't know if make to much sense to write the same data 1000 times. Most of our tests are working by generating random data in memory (e.g. a `List<Group>`) the write it and then test whether we can read back the same data we have in memory.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
gszadovszky: I agree on validating the arguments is important. But the proper exception to be thrown for a `null` is a `NullPointerException` that would be thrown at line 134 anyway. If you really want to validate the argument for `null` at the first line of the method I would suggest using `java.util.Objects.requireNonNull(Object)`.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCipher.java
```diff
@@ -78,11 +78,22 @@
     if (rowGroupOrdinal < 0) {
       throw new IllegalArgumentException("Wrong row group ordinal: " + rowGroupOrdinal);
     }
-    byte[] rowGroupOrdinalBytes = shortToBytesLE(rowGroupOrdinal);
+    short shortRGOrdinal = (short) rowGroupOrdinal;
+    if (shortRGOrdinal != rowGroupOrdinal) {
+      throw new ParquetCryptoRuntimeException("Encrypted parquet files can't have "
+          + "more than Short.MAX_VALUE row groups: " + rowGroupOrdinal);
```
gszadovszky, line 84: I think, writing the actual value instead of referencing a java constant is more informative.

ggershinsky:
> I know, this is not the whole feature yet but would like to ensure that at the end we will have proper test coverity. So, the question is do we have or planning to have tests to verify the working of semi-encrypted files (some columns and the footer are plaintext) with previous readers.

We test this manually. We also have an automatic test that runs the current reader code on files that were previously created with other writers (such as parquet-cpp with encryption; these files are kept in parquet-testing repo and fetched from there). But I don't know how to automate running of previous readers on the files written by the current code - ?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestEncryption.java
ggershinsky: we're working on a better unitest for encryption, in #782 . it should be possible to drop this one.

ggershinsky:
> Now, we are in sync. I will try to keep it this way. ;)

Appreciated! 
Addressing today's comments will keep me busy for a while :), I'll get back next week.

gszadovszky:
> > I know, this is not the whole feature yet but would like to ensure that at the end we will have proper test coverity. So, the question is do we have or planning to have tests to verify the working of semi-encrypted files (some columns and the footer are plaintext) with previous readers.
> 
> We test this manually. We also have an automatic test that runs the current reader code on files that were previously created with other writers (such as parquet-cpp with encryption; these files are kept in parquet-testing repo and fetched from there). But I don't know how to automate running of previous readers on the files written by the current code - ?

Good question. We should have the tooling to provide such tests but not sure if we already have it. I am fine with testing it manually for now just document it in details (e.g. adding to the git comment of the test PR).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestBloomEncryption.java
ggershinsky: this will be a part of https://github.com/apache/parquet-mr/pull/782 (it will also remove the TestBloomEncryption.java and TestColumnIndexEncryption.java files).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
ggershinsky: these fields are accessed by the `EncryptedColumnChunkMetaData` class, that sets their values after decrypting the column metadata.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -188,7 +189,11 @@ private void checkDeltaByteArrayProblem(FileMetaData meta, Configuration conf, B
       // this is okay if not using DELTA_BYTE_ARRAY with the bug
       Set<Encoding> encodings = new HashSet<Encoding>();
       for (ColumnChunkMetaData column : block.getColumns()) {
-        encodings.addAll(column.getEncodings());
+        try {
+          encodings.addAll(column.getEncodings());
+        } catch (KeyAccessDeniedException e) {
```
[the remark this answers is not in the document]
ggershinsky: removed the suppression

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1114,7 +1299,20 @@ public ColumnIndex readColumnIndex(ColumnChunkMetaData column) throws IOExceptio
       return null;
     }
     f.seek(ref.getOffset());
-    return ParquetMetadataConverter.fromParquetColumnIndex(column.getPrimitiveType(), Util.readColumnIndex(f));
+
+    column.decryptIfNeededed();
```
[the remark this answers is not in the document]
ggershinsky: found a better place for it, inside the `ColumnChunkMetaData` object

ggershinsky:
The current review round is addressed; the checks pass.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
gszadovszky: Yep, but `EncryptedColumnChunkMetaData` is defined inside the class `ColumnChunkMetaData`. It makes them access each other's `private` members. Even the class`EncryptedColumnChunkMetaData` can be declared `private` if you don't want to use the type outside of `EncryptedColumnChunkMetaData`.
See [Nested Classes](https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html) for details.

ggershinsky: hmm, when I change `encodingStats` to `private`, I get a compilation error  
_The field ColumnChunkMetaData.encodingStats is not visible_
at line 619,
`this.encodingStats = shadowColumnChunkMetaData.encodingStats;`

gszadovszky: OK, I got it now. So, these are not nested classes. There are 4 different classes next to each other in the same java file. This is valid since only one class is public but I've never seen such design in production. I've never realized this design in `ColumnChunkMetaData`. There is no reason why one would put multiple classes in the same file but not nesting them.

So, without refactoring this structure I'll accept `encodingStats` to not being `private`. But, we can narrow the visibility by using _package private_ (no modifier).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1071,12 +1229,31 @@ public BloomFilterReader getBloomFilterDataReader(BlockMetaData block) {
    */
   public BloomFilter readBloomFilter(ColumnChunkMetaData meta) throws IOException {
     long bloomFilterOffset = meta.getBloomFilterOffset();
+
+    if (0 == bloomFilterOffset) { // TODO Junjie - is there a better way to handle this?
```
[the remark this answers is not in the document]
gszadovszky: @ggershinsky, please ensure this is correct and don't keep TODOs in the final code. (We usually don't fix them and they will be there forever.)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
ggershinsky: Sure, will change this.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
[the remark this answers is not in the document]
ggershinsky: yep, this works; also, makes sense to me - we should not proceed to read a bloom filter header, if there is no bloom filter in the file; this can be checked via the `bloomFilterOffset` variable - it can't be 0 if a bloom filter is present; and if no bloom filter, it can be 0 only.

I'll remove the TODO comment.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestBloomEncryption.java
andersonm-1: Thank you, @gszadovszky , for the suggestion. We've added it to #782 .

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
shangxinli: It is not a blocking comment and I am fine with it. But generally speaking, adding too much nested if/else diverges the code path and causes the complexity for reading. One way to avoid duplicating is to wrap them up in helper functions. I understand column indexes and bloom filters already did that but if we keep adding features like this, the code will become less and less readable.

PR #778, opened 2020-04-03
PARQUET-1827: UUID type currently not supported by parquet-mr

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-avro/README.md
+++ b/parquet-avro/README.md
@@ -1,0 +1,44 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+Apache Avro integration
+======
+
+**TODO**: Add description and examples how to use parquet-avro
+
+## Available options via Hadoop Configuration
+
+### Configuration for reading
+
+| Name                                    | Type      | Description                                                          |
+|-----------------------------------------|-----------|----------------------------------------------------------------------|
+| `parquet.avro.data.supplier`            | `Class`   | The implementation of the interface org.apache.parquet.avro.AvroDataSupplier. Available implementations in the library: GenericDataSupplier, ReflectDataSupplier, SpecificDataSupplier.<br/>The default value is `org.apache.parquet.avro.SpecificDataSupplier` |
+| `parquet.avro.read.schema`              | `String`  | The Avro schema to be used for reading. It shall be compatible with the file schema. The file schema will be used directly if not set. |
+| `parquet.avro.projection`               | `String`  | The Avro schema to be used for projection.                           |
+| `parquet.avro.compatible`               | `boolean` | Flag for compatibility mode. `true` for materializing Avro `IndexedRecord` objects, `false` for materializing the related objects for either generic, specific, or reflect records.<br/>The default value is `true`. |
+
+### Configuration for writing
+
+| Name                                    | Type      | Description                                                          |
+|-----------------------------------------|-----------|----------------------------------------------------------------------|
+| `parquet.avro.write.data.supplier`      | `Class`   | The implementation of the interface org.apache.parquet.avro.AvroDataSupplier. Available implementations in the library: GenericDataSupplier, ReflectDataSupplier, SpecificDataSupplier.<br/>The default value is `org.apache.parquet.avro.SpecificDataSupplier` |
+| `parquet.avro.schema`                   | `String`  | The Avro schema to be used for generating the Parquet schema of the file. |
+| `parquet.avro.write-old-list-structure` | `boolean` | Flag whether to write list structures in the old way (2 levels) or the new one (3 levels). When writing at 2 levels no null values are available at the element level.<br/>The default value is `true` |
+| `parquet.avro.add-list-element-records` | `boolean` | Flag whether to assume that any repeated element in the schmea is a list element.<br/>The default value is `true`. |
+| `parquet.avro.write-parquet-uuid`       | `boolean` | Flag whether to write the [Parquet UUID logical type](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid) in case of an [Avro UUID type](https://avro.apache.org/docs/current/spec.html#UUID) is present.<br/>The default value is `false`. |
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
@@ -37,10 +37,12 @@ import java.util.Collections;
 
 import static org.apache.avro.Schema.Type.INT;
 import static org.apache.avro.Schema.Type.LONG;
+import static org.apache.avro.Schema.Type.STRING;
 import static org.apache.parquet.avro.AvroTestUtil.field;
 import static org.apache.parquet.avro.AvroTestUtil.optionalField;
 import static org.apache.parquet.avro.AvroTestUtil.primitive;
 import static org.apache.parquet.avro.AvroTestUtil.record;
+import static org.apache.parquet.schema.LogicalTypeAnnotation.uuidType;
 import static org.apache.parquet.schema.OriginalType.DATE;
 import static org.apache.parquet.schema.OriginalType.TIMESTAMP_MICROS;
 import static org.apache.parquet.schema.OriginalType.TIMESTAMP_MILLIS;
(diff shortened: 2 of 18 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -861,6 +871,36 @@ PrimitiveStringifier valueStringifier(PrimitiveType primitiveType) {
     }
   }
 
+  public static class UUIDLogicalTypeAnnotation extends LogicalTypeAnnotation {
```
shangxinli, line 874: Why there is no implementation for hasCode() and equal()?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -421,4 +422,30 @@ private String stringifyWithScale(BigInteger i) {
       }
     };
   }
+
+  static final PrimitiveStringifier UUID_STRINGIFIER = new PrimitiveStringifier("UUID_STRINGIFIER") {
+    private final char[] digit = "0123456789abcdef".toCharArray();
+    @Override
+    public String stringify(Binary value) {
+      byte[] bytes = value.getBytesUnsafe();
```
shangxinli, line 430: Do you want to make 'bytes' final since you are calling getBytesUnsafe()?

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -309,6 +308,35 @@ public void testDecimalStringifier() {
...
+    PrimitiveStringifier stringifier = PrimitiveStringifier.UUID_STRINGIFIER;
+
+    assertEquals("00112233-4455-6677-8899-aabbccddeeff", stringifier.stringify(
+        toBinary(0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff)));
+    assertEquals("00000000-0000-0000-0000-000000000000", stringifier.stringify(
+        toBinary(0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00)));
+    assertEquals("ffffffff-ffff-ffff-ffff-ffffffffffff", stringifier.stringify(
+        toBinary(0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff)));
+
+    assertEquals("0eb1497c-19b6-42bc-b028-b4b612bed141", stringifier.stringify(
```
shangxinli, line 322: Are the 3 test data are randomly chosen? It seems duplicate coverage. 

Could you add some negative tests like incorrect length uuids, invalid characters etc.

parquet-avro/README.md
```diff
@@ -0,0 +1,44 @@
+<!--
```
shangxinli, line 1: Should we add it using separate jira?

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
```diff
@@ -766,6 +768,33 @@ public void testReuseNameInNestedStructureAtSameLevel() throws Exception {
...
+        Arrays.asList(new Schema.Field("uuid", LogicalTypes.uuid().addToSchema(Schema.create(STRING)), null, null)));
+    String parquet = "message myrecord {\n" +
+        "  required binary uuid (STRING);\n" +
+        "}\n";
+    Schema toAvro = Schema.createRecord("myrecord", null, null, false,
+        Arrays.asList(new Schema.Field("uuid", Schema.create(STRING), null, null)));
+
+    testAvroToParquetConversion(fromAvro, parquet);
+    testParquetToAvroConversion(toAvro, parquet);
+  }
```
shangxinli, line 783: Can we have checkReaderWriterCompatibility() to verify if the parquet and avro schema are compatible for UUID?  
There are issues like PARQUET-1681 for avro schema and parquet schema conversion for other types.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
gszadovszky: As this is a singleton the default implementation of `equals(Object)` and `hashCode()` fits perfectly.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
gszadovszky: `final` would protect the reference only and not the values of the array. Making a reference final in a local scope is usually required in situations where it is accessed from e.g. lambda closures.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
gszadovszky: The idea is to have 3 kind-of corner cases and 3 common (random but constant) values. What do you mean by duplicate coverage? (I think, we do not need exhaustive testing for the stringifiers because they only used by our tools for debugging purposes.)

The stringifiers do not validate the data they get because of performance reasons. So, if the array is longer than 16 it would simply stringify the first 16 and skip the others. In case of the length is too short then an `ArrayIndexOutOfBoundsException` would be thrown. Do you think we should test these cases? They would not reach any additional branches in the parquet code.
Invalid characters are not possible. The full set of values of the 16 bytes array is covered in UUID.

parquet-avro/README.md
gszadovszky: Usually we separate jiras in similar cases to not make cherry-picking hard in case of the related changed needs to be on another branch as well. In this case this is only a documentation so should not cause any troubles. It would be cleaner if this documentation would have already been existed and I've had to add the docs of the new keys only (which would clearly be part of this change).
If you have a strong opinion to separate this to another change I'm happy to do so, though.

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
gszadovszky: To be honest I am not too familiar with parquet-avro. I've made the changes based on the implementation/test of other logical types. Could you explain it in more details what you would test exactly?

shangxinli: Basically we found some type of avro schema is not compatible with the parquet schema which it is converted to. This caused problem that the data cannot be read. I have a test here (https://github.com/shangxinli/parquet-mr/commit/f80469f55b83404ea334ee4019f658ecdb5ac575#diff-536ca67880a7870cf8df8f95143bd7d7R814) that reproduce the issue for a nested schema. I know likely UUID type won't have this issue but I better to have a test for it. It is pretty easy to add also.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
shangxinli: By duplicate coverage, I meant the #2, #3 tests seems repeating the same test as #1. The value is different, but when the test executes, they would execute the same code path. So I think they won't provide extra coverage. 

From test perspective, negative test does provide values. In this case, we can test the exception is thrown as expected if it is too short. 

For "if the array is longer than 16 it would simply stringify the first 16 and skip the others", that could cause silent errors, right?

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
shangxinli: If it is too much effort for doing this, it is OK not to do it. It is a lower priority.

parquet-avro/README.md
shangxinli: Sounds good

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
shangxinli: Sounds good!

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
shangxinli: Sounds good!

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
gszadovszky: I'll look into this just did not have time to work on this PR. Thanks a lot for reviewing. :)

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
gszadovszky: I'm happy to add tests for the edge cases like too short or too long inputs. Though, I would not implement additional validations because of performance issues. A `stringify` method would be invoked on each values; an additional check would highly impact performance even if it is only used from the tools and not really in production. A `Stringifier` is associated to the value at schema level which means it shall never happen that the value is invalid. That's why the `Stringifier` implementations do not validate the values.

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
gszadovszky: The [`testRoundTripConversion`](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java#L144) I'm using in `testUUIDTypeWithParquetUUID` is actually stronger than the one you suggested: it checks for equality (in two phases) of the initial and the result avro schemas (and not only for compatibility). For `testUUIDType`, though it is a good idea to check the compatibility of the avro schemas.

shangxinli: Sounds good

PR #788, opened 2020-04-21
PARQUET-1847: Filter out github notification from dev mail list

chenjunjiedada:

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


chenjunjiedada:
@wesm, Could you please help to take a look?

wesm:
github@ doesn't exist for parquet.a.o so either this traffic should be directed to another mailing list, or a new one should be created. 

Pausing for a second, we should probably confirm on dev@ where people want these e-mails to go

chenjunjiedada:
Got it. Maybe we need a vote to choose solutions.

chenjunjiedada:
@wesm , I guess you are familiar with infra setup. Would you please help to list the options and start a vote thread for this?

PR #793, opened 2020-05-21
PARQUET-1866: Replace Hadoop ZSTD with JNI-ZSTD

shangxinli:

### Jira

### Tests

### Commits

### Documentation


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstdCodec.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.parquet.hadoop.codec.ZstdCodec;
+import org.junit.Assert;
+import org.junit.Test;  
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Random;
+
+public class TestZstdCodec {
```
gszadovszky, line 33: I tried to find the code part where we set the hadoop conf to the codec but could not find it. Please, write a high level test where you set compression level and workers in the hadoop conf and executes a file write via e.g. an MR job.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/ZstdCodec.java
```diff
@@ -0,0 +1,112 @@
...
+ * which is already a dependency for Parquet. ZSTD JNI ZstdOutputStream and
+ * ZstdInputStream use Zstd internally. So no need to create compressor and
+ * decompressor in ZstdCodec.
+ */
+public class ZstdCodec implements Configurable, CompressionCodec {
+
+  public final static String PARQUET_COMPRESS_ZSTD_LEVEL = "parquet.compression.codec.zstd.level";
+  public final static int DEFAULT_PARQUET_COMPRESS_ZSTD_LEVEL = 3;
+  public final static String PARQUET_COMPRESS_ZSTD_WORKERS = "parquet.compression.codec.zstd.workers";
+  public final static int DEFAULTPARQUET_COMPRESS_ZSTD_WORKERS = 0;
```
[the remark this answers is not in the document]
shangxinli: Sure

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstdCodec.java
shangxinli: Added test for conf setting with MR

shangxinli:
@karavelov, Do you have time to review the code?

luben:
LGTM

shangxinli:
@gszadovszky Do you have time for another look?

parquet-hadoop/README.md
```diff
@@ -324,9 +324,20 @@ ParquetInputFormat to materialize records. It should be a the descendant class o
...
 ## Class: UnmaterializableRecordCounter
 
 **Property:** `parquet.read.bad.record.threshold`  
 **Description:** The percentage of bad records to tolerate.  
 **Default value:** `0`
+
+## Class: ZstandardCodec
+
+**Property:** `parquet.compression.codec.zstd.level`
+**Description:** The compression level of ZSTD. The valid range is 1~22. Generally the higher compression level, the higher compression ratio can be achieved, but the writing time will be longer.
```
gszadovszky, line 336: In markdown you have to do some extra if you want to force new lines. With a simple linebreak the two lines will be written together. There are some tricks [here](https://gist.github.com/shaunlebron/746476e6e7a4d698b373) for enforcing line breaks.
(You may check the rendered format on [github](https://github.com/shangxinli/parquet-mr/tree/master_new/parquet-hadoop)).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
```diff
@@ -0,0 +1,146 @@
...
+  public void testZstdConfWithMr() throws Exception {
+    JobConf jobConf = new JobConf();
+    Configuration conf = new Configuration();
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_LEVEL, 18);
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_LEVEL, 4);
+    RunningJob mapRedJob = runMapReduceJob(CompressionCodecName.ZSTD, jobConf, conf);
+    assert(mapRedJob.isSuccessful());
+  }
+
+  private RunningJob runMapReduceJob(CompressionCodecName codec, JobConf jobConf, Configuration conf) throws IOException, ClassNotFoundException, InterruptedException {
```
gszadovszky, line 104: Thanks for creating an MR test. What I wanted to test with it is if the ZSTD related configuration takes place if set via the configuration. This is not tested currently. I wanted to test it because I am not sure if we have the code in place to pass configuration properties to the codecs.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
```diff
@@ -0,0 +1,146 @@
...
+    is.close();
+    return decompressed;
+  }
+
+  @Test
+  public void testZstdConfWithMr() throws Exception {
+    JobConf jobConf = new JobConf();
+    Configuration conf = new Configuration();
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_LEVEL, 18);
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_LEVEL, 4);
```
gszadovszky, line 99: You are setting the level twice (and the workers are missing).

parquet-hadoop/README.md
shangxinli: Thanks for pointing out! I think it is OK to keep them in same line unless you have strong opinion we should have a separate line. I looked at the existing lines above and I see some of them are longer than mines.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: Good catch

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: I set breakpoint and checked it. But I don't know how to get back the level/workers at the MR job. Any ideal how to do that? I also verified it in the standalone Spark application. Do you think that is sufficient?

parquet-hadoop/README.md
gszadovszky: I was trying to say that _Property_, _Description_ and _Default value_ should be separate paragraphs. Currently they are not, the _Description_ is rendered just after _Property_ in the same line even if in the markdown they are separated. You should use one of the techniques described under the link to force the new line there.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
gszadovszky: Maybe, if we create two parquet files with different levels (e.g. 1 and 22) but with the exact same data we can expect that the larger level will generate smaller file?
I just want to ensure that the properties really arrive to the codec. If you think the related test requires too much effort or even not feasible I am fine if you can point me to the code path where the properties are passed through to the codecs.

parquet-hadoop/README.md
shangxinli: I see. I added double space in the end of Property, Description sentences. I checked it in Intellij and it works. I also see the above sections used the trailing double space.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: I like the idea to compare the two files with different compression levels. Only thing is that it is not 100% guarantee the higher level always gets higher compression ratio although the chance of that is very low. In our test data, it seems it is OK though. So I will just go ahead to add the file size comparison.

dongjoon-hyun:
Thank you, @shangxinli and all!
cc @dbtsai

dbtsai:
+1 @shangxinli and thank you for this contribution. 

This will allow users who are on order versions of hadoop that don't support native ZSTD to use ZSTD compression in Parquet, and also, users don't have to go through the very complicated hadoop native installation. For developers, we will be able to easily test this out in different local envs.  

cc @rdblue

dbtsai:
@shangxinli do we have benchmark comparing to native hadoop codec both in size and speed? Thanks.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
gszadovszky: Yes, I know, there is no 100% guarantee in normal circumstances. But if you write the test in a deterministic way (as it should be by e.g. using a fix seed for random generation) then without changing the data and the compression codec it shall be fine. I worth a comment, though about it is not the nicest way of testing the work of the properties but currently that's seems to be the only one.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
```diff
@@ -0,0 +1,164 @@
...
+    return decompressed;
+  }
+
+  @Test
+  public void testZstdConfWithMr() throws Exception {
+    long fileSizeLowLevel = runMrWithConf(1);
+    // Clear the cache so that a new codec can be created with new configuration
+    CodecFactory.CODEC_BY_NAME.clear();
+    long fileSizeHighLevel = runMrWithConf(22);
+    assert (fileSizeLowLevel > fileSizeHighLevel);
```
gszadovszky, line 101: Please use the JUnit framework `assert` functions instead of the `assert` keyword. The [`assert` keyword](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html) is not for unit testing and it is not 100% guaranteed that the assertions are enabled during testing.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private long runMrWithConf(int level) throws Exception {
+    JobConf jobConf = new JobConf();
+    Configuration conf = new Configuration();
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_LEVEL, level);
+    jobConf.setInt(ZstandardCodec.PARQUET_COMPRESS_ZSTD_WORKERS, 4);
+    Path path = new Path(Files.createTempDirectory("zstd" + level).toAbsolutePath().toString());
+    RunningJob mapRedJob = runMapReduceJob(CompressionCodecName.ZSTD, jobConf, conf, path);
+    assert(mapRedJob.isSuccessful());
```
gszadovszky, line 111: Same as above

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
```diff
@@ -0,0 +1,164 @@
...
+  private BytesInput decompress(ZstandardCodec codec, BytesInput bytes, int uncompressedSize) throws IOException {
+    BytesInput decompressed;
+    InputStream is = codec.createInputStream(bytes.toInputStream(), null);
+    decompressed = BytesInput.from(is, uncompressedSize);
+    is.close();
+    return decompressed;
+  }
+
+  @Test
+  public void testZstdConfWithMr() throws Exception {
```
gszadovszky, line 96: Please, add some comments that the intent of this test is to verify that the properties are passed through from the config to the codec.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: Sounds good!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: fixed

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: Added. Thanks.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestZstandardCodec.java
shangxinli: Thanks.

shangxinli:
> @shangxinli do we have benchmark comparing to native hadoop codec both in size and speed? Thanks.

Hi @dbtsai, I didn't because I don't have Hadoop host installed with ZSTD. @luben, did you ever compare it with Hadoop ZSTD?

luben:
@shangxinli : I haven't benchmarked

PR #794, opened 2020-05-29
Adding manifest file

s-pedamallu:


manifest.yaml
```diff
@@ -0,0 +1,40 @@
+name: parquetmr
+description: Lyft fork of the Apache Parquet project
```
Fokko, line 2: ?? I don't think this should go in the Apache project.

s-pedamallu: Sorry... This PR should not have been out... let me close this one.

PR #796, opened 2020-06-12
Parquet-1872: Add TransCompression command to parquet-tools

shangxinli:

### Jira

### Tests

### Commits

### Documentation


parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
```diff
@@ -0,0 +1,302 @@
...
+
+  private static final class DumpGroupConverter extends GroupConverter {
+    @Override public void start() {}
+    @Override public void end() {}
+    @Override public Converter getConverter(int fieldIndex) { return new DumpConverter(); }
+  }
+
+  private static final class DumpConverter extends PrimitiveConverter {
+    @Override public GroupConverter asGroupConverter() { return new DumpGroupConverter(); }
+  }
```
gszadovszky, line 282: I think, `Dummy` would be a better naming instead of `Dump`.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
```diff
@@ -0,0 +1,302 @@
...
+    BytesInput data = readBlock(payloadLength, reader);
+    if (isCompressed) {
+      data = decompressor.decompress(data, rawDataLength);
+    }
+    BytesInput newCompressedData = compressor.compress(data);
+    return newCompressedData.toByteArray();
+  }
+
+  private BytesInput readBlock(int length, TransParquetFileReader reader) throws IOException {
+    byte[] data = new byte[length];
```
gszadovszky, line 262: Creating a byte array can be expensive if there are many pages. I would suggest maintaining a cache array that is reused and created only if a block is larger than the cache.

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
```diff
@@ -0,0 +1,271 @@
...
+            Assert.assertEquals(inColumnIndex.getMinValues(), outColumnIndex.getMinValues());
+            Assert.assertEquals(inColumnIndex.getNullCounts(), outColumnIndex.getNullCounts());
+          }
+          if (inOffsetIndex != null) {
+            Assert.assertEquals(inOffsetIndex.getPageCount(), outOffsetIndex.getPageCount());
+            for (int k = 0; k < inOffsetIndex.getPageCount(); k++) {
+              Assert.assertEquals(inOffsetIndex.getFirstRowIndex(k), outOffsetIndex.getFirstRowIndex(k));
+              Assert.assertEquals(inOffsetIndex.getLastRowIndex(k, inChunk.getValueCount()),
+                outOffsetIndex.getLastRowIndex(k, outChunk.getValueCount()));
+            }
```
gszadovszky, line 207: You are not testing `OffsetIndex.getOffset(int)` which is a big problem. These offsets are to reference the exact locations of the pages in the file. It is not trivial to test it but if you were it would fail because you are not rewriting it in the new file. Every other data in the column/offset indexes shall be the same in the source and the target file but these values shall be recalculated as the sizes of the pages will be different after compressed by another codec so the positions shift in the file.

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
```diff
@@ -0,0 +1,271 @@
...
+    String outputFile = createTempFile("output_trans");
+
+    TransCompressionCommand command = new TransCompressionCommand(createLogger());
+    command.setConf(new Configuration());
+    command.input = inputFile;
+    command.output = outputFile;
+    command.codec = destCodec;
+    command.run();
+
+    validateColumns(inputFile, numRecord);
```
gszadovszky, line 124: I think, you should check the output file instead.

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
```diff
@@ -0,0 +1,271 @@
...
+  private void validateColumns(String inputFile, int numRecord) throws IOException {
+    ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), new Path(inputFile)).withConf(conf).build();
+    for (int i = 0; i < numRecord; i++) {
+      Group group = reader.read();
+      assertTrue(group.getLong("DocId", 0) < 1000);
+      assertEquals(group.getBinary("Name", 0).length(), 100);
+      assertEquals(group.getBinary("Gender", 0).length(), 100);
+      Group subGroup = group.getGroup("Links", 0);
+      assertEquals(subGroup.getBinary("Backward", 0).length(), 100);
+      assertEquals(subGroup.getBinary("Forward", 0).length(), 100);
```
gszadovszky, line 166: I don't like these tests. You should test for exact values. You may either keep the generated data in memory so you can match the file content with it or you may read both the source and the target file in the same time and match the values.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
```diff
@@ -0,0 +1,302 @@
...
+
+    Path inPath = new Path(input);
+    Path outPath = new Path(output);
+    CompressionCodecName codecName = CompressionCodecName.valueOf(codec);
+
+    ParquetMetadata metaData = ParquetFileReader.readFooter(getConf(), inPath, NO_FILTER);
+    MessageType schema = metaData.getFileMetaData().getSchema();
+
+    try (TransParquetFileReader reader = new TransParquetFileReader(HadoopInputFile.fromPath(inPath, getConf()), HadoopReadOptions.builder(getConf()).build())) {
+      ParquetFileWriter writer = new ParquetFileWriter(getConf(), schema, outPath, ParquetFileWriter.Mode.CREATE);
```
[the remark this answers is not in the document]
shangxinli: The writer doesn't extend the closable(). If we add, it will report build error.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
```diff
@@ -0,0 +1,302 @@
...
+        throw new IOException("There are more pages " + pageIndex + " found in the column than in the columnIndex " + columnIndex.getNullPages().size());
+      }
+      org.apache.parquet.column.statistics.Statistics.Builder statsBuilder = org.apache.parquet.column.statistics.Statistics.getBuilderForReading(type);
+      statsBuilder.withNumNulls(columnIndex.getNullCounts().get(pageIndex));
+
+      if (!columnIndex.getNullPages().get(pageIndex)) {
+        statsBuilder.withMin(columnIndex.getMinValues().get(pageIndex).array().clone());
+        statsBuilder.withMax(columnIndex.getMaxValues().get(pageIndex).array().clone());
+      }
+      return statsBuilder.build();
```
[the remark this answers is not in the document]
shangxinli: True. The method convertStatistics() will read the statistic ether from page or ColumnIndex. The the returned value of convertStatistics() is passed to writeDataPage(). Then writeDataPage() will write to ColumnIndex only I think, since your change removed the page header's statistics.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
shangxinli: Got it.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
shangxinli: Sounds good! While creating the cache byte array, it uncover an bug(I think it is a bug) in the H1SeeakableInputStream. I added that fix to this PR also.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
gszadovszky: Hm, you're right. `writer.end` closes the file descriptor behind. So, I would suggest invoking `end` in the `finally` block so the file descriptor will be closed even if an exception occurs.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
gszadovszky: You are completely right, I've forgot about my own implementation :)

Then, just one comment for this code part. I think, column index is more reliable so I would use it for generating statistics at the first place.

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
gszadovszky: Update: I was wrong. The offset index is calculated at page write so the offsets should be correct. But testing them still makes sense.

shangxinli: Sure. I will add it.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
shangxinli: Sounds good

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
shangxinli: Since your change removed page statistics, when ColumnIndex is not null, then page statistics must be null, and vice versa(for older version). So the order to check which one shouldn't matter to the result. Correct?

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
shangxinli: Good catch!

parquet-cli/src/test/java/org/apache/parquet/cli/commands/TransCompressionCommandTest.java
shangxinli: I will cache the input data and compare it.

shangxinli:
> There are a lot of code duplications between `cli` and `tools`. I would suggest adding these to `parquet-hadoop` as a utility for this functionality and use it from the two command line tools. The unit test also can be placed in `parquet-hadoop` so you write it once only. (You may keep simple unit tests at the tools side to verify the tool itself but the functionality shall be verified in the module where it is implemented.)

reply:
I was able to move the core port to parquet-hadoop. However, I found it is hard share the test class because between parquet-hadoop and parquet-cli because by default the test class is not package as this article said https://dzone.com/articles/sharing-test-classes-between-multiple-modules-in-a. So I cannot utilize the test methods in parquet-hadoop test. There are several approaches as below. Let me know your thoughts. 
1. Keep as is. Parquet-tool will be removed later. 
2. Move the most parts into parquet-hadoop and implement the unit tests there. In parquet-cli and parquet-tool, we don't add unit test.  Give the code in parquet-cli/tools would be just a simple wrapper, it should be OK.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
gszadovszky: This is true for parquet-mr. I am not sure about other implementations, which one writes page statistics and/or column indexes. During the column index design we were talking about some edge cases of some data types (e.g. NaN for floating point numbers) and turned out that page statistics are not always handled the same way in the different implementations. At least for parquet-mr and Impala we synchronized these for column indexes.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
```diff
@@ -53,7 +53,7 @@ public void readFully(byte[] bytes) throws IOException {
 
   @Override
   public void readFully(byte[] bytes, int start, int len) throws IOException {
-    stream.readFully(bytes);
+    stream.readFully(bytes, start, len);
```
gszadovszky, line 56: Good catch!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/CompressionConveterTest.java
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  private String getString() {
+    char[] chars = {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'x', 'z', 'y'};
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < 100; i++) {
+      sb.append(chars[new Random().nextInt(10)]);
+    }
+    return sb.toString();
+  }
```
gszadovszky, line 274: Instead of creating a `Random` instance or retrieving a `ThreadLocalRandom` instance for each generated values I would suggest creating one `Random` instance in `TestDocs` in the constructor and use it for all the values to be generated in that instance. Also suggest to use a hard-coded seed for `Random` to ensure the test is deterministic.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/TransCompressionCommand.java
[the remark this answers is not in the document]
shangxinli: Sounds good!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/CompressionConveterTest.java
shangxinli: Sounds good!

PR #800, opened 2020-07-13
PARQUET-1884: Encryption branch merge in master

ggershinsky:
Merges encryption feature


gszadovszky:
@ggershinsky, could you resolve the conflicts?

ggershinsky:
> A have some minor findings.
> 
> One more thing that we failed to check before. It is the CRC calculation. (See [TestDataPageV1Checksums](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageV1Checksums.java) and [the thrift file](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L628-L652) for details.) How does the CRC work in case of encrypted pages? Is the definition in the thrift file still stand (CRC is calculated after compression but before encryption)?
> At this point I am not requesting any additional implementation but we shall check at least that the CRC calculation also works fine with encryption.

Per https://github.com/apache/parquet-mr/pull/800/files#diff-d007a18083a2431c30a5416f248e0a4bR169 and https://github.com/apache/parquet-mr/pull/800/files#diff-d007a18083a2431c30a5416f248e0a4bR183, CRC is calculated after compression and encryption of page bytes. Which makes sense, because these are the bytes that are written to the disk. The Thrift comment says CRC is calculated after compression (doesn't mention encryption), but we should have modified this comment. One more point - AEG_GCM mode verifies page integrity (in CPU hardware), so CRC is not needed in this encryption mode.

ggershinsky:
> AEG_GCM mode verifies page integrity (in CPU hardware), so CRC is not needed in this encryption mode.

Btw, this is the default encryption mode. In Java 9 and later, there is no reason to use encryption and CRC together.

gszadovszky:
Currently CRC check is configurable by hadoop conf `parquet.page.verify-checksum.enabled` which is disabled by default. I accept there is no reason to use both encryption and crc together but as the user can configure that we shall ensure it works properly.

ggershinsky:
Per the sync discussion,
- no reason to use both encryption and CRC even in Java 8 or older (GCM overhead is due to integrity verification; same issue as with CRC)
- I'll run a quick test (a variation of the TestDataPageV1Checksums), to make sure Parquet works properly if a user turns both features on, as allowed by the API/config (no need to forbid this combination)

ggershinsky:
@gszadovszky TestDataPageV1Checksums didn't work out, because it uses an old parquet API, with deprecated methods we've decided not to extend for encryption. So instead, I've run the ColumnIndexFiltering test, that uses the valid parquet API. I've modified it to add encryption/decryption, and enablePageWriteChecksum()/usePageChecksumVerification(true) to this test, everything worked ok.

Regarding the travis.yaml and hadoop/readme.md changes - should I send a pull request to the encryption branch, or to the master after this pr is merged?

gszadovszky:
@ggershinsky, thanks a lot for testing it. 

Sorry, I've forgot that this one is not a simple PR that allows easy fixing of the findings. Please, create a JIRA about these two and fix it on master.
I'm merging this PR.

PR #802, opened 2020-07-20
PARQUET-1885: Pass descriptor to ProtoWriteSupport constructor

mauliksoneji:
addresses https://issues.apache.org/jira/browse/PARQUET-1885

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoDescriptorSupport.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoDescriptorSupport.java
@@ -1,0 +1,58 @@
+/* 
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.proto;
+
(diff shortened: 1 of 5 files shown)


Fokko:
@mauliksoneji can you resolve the conflicts?

mauliksoneji:
Hello @Fokko thank you very much for checking the PR.
I have fixed the conflicts and pushed the code with additional tests.

mauliksoneji:
Also, [this previously merged commit](https://github.com/apache/parquet-mr/commit/fadbe6ef326faa5984d4d8d6df581a91c8c6cca2) was throwing an error because of unimported class. It has been fixed in my PR

mauliksoneji:
Hello everyone, can someone please check this PR when you are free. Thanks in advance!

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoDescriptorSupport.java
```diff
@@ -0,0 +1,58 @@
...
+    this.protoMessage = protoMessage;
+  }
+
+  public ProtoDescriptorSupport(Descriptors.Descriptor messageDescriptor) {
+    this.messageDescriptor = messageDescriptor;
+  }
+
+  public Descriptors.Descriptor getMessageDescriptor(Configuration configuration) {
+    // if no protobuf descriptor was given in constructor, load descriptor from configuration (set with setProtobufClass)
+    if (protoMessage == null && messageDescriptor == null) {
```
gszadovszky, line 42: How this is possible while this class have only constructors that expect one of them? I think, it would be more readable and failproof if we would have a parameterless constructor that allows to be both `null` and the existing ones would expect to set them non-`null`. What do you think?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoDescriptorSupport.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.parquet.proto;
+
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Message;
+import com.twitter.elephantbird.util.Protobufs;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.parquet.hadoop.BadConfigurationException;
+
+public class ProtoDescriptorSupport {
```
gszadovszky, line 27: Could you write some comments about the purpose of this class?

gszadovszky (changes requested):

jinyius:
does this pr satisfy the issue as well?

https://github.com/apache/parquet-mr/pull/963/files

PR #815, opened 2020-09-15
PARQUET-1776: NIO wrapper for Output/Input File

HunterL:

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/io/LocalSeekableInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/io/LocalSeekableInputStream.java
@@ -1,0 +1,47 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.io;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class LocalSeekableInputStream extends DelegatingSeekableInputStream {
+  private long pos;
+
+  public LocalSeekableInputStream(InputStream stream) {
+    super(stream);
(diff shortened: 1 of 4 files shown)


belugabehr:
@HunterL 

OK, so I've been recently diving into Java NIO a bit and let me add some thoughts that relate to this task:

Please change the naming convention and drop the "Local" prefix.  I think Nio is probably best.  As an example of why this matters:

```
// This exists in the current PR
return new LocalSeekableInputStream(Files.newInputStream(path));
```

Well that statement `Files.newInputStream(path)` will provide a new InputStream, but the InputStream is totally driven by whatever the path points at.  Things within the JVM could be setup to all the provided path to point at an FTP location for example and that would be a _remote_ read and not a _local_ read.


I also think you will have better luck storing, internal to the Input/Output streams a `SeekableByteChannel` instead of an `OutputStream` or an `InputStream`.  This will allow you to easily track the position aspect of the underlying stream without having to do it yourself.  For example, the current `OutputStream` implementation does not currently implement this feature:

```
  @Override
  public long getPos() throws IOException {
    return 0;
  }
```

Thanks.

belugabehr:
Also requires unit tests.

belugabehr:
Here are a couple of starters:

```
       private final ByteBuffer oneByteBuffer = ByteBuffer.allocate(1);

	@Override
	public int read() throws IOException {
		int read = this.channel.read(this.oneByteBuffer.rewind());
		if (read < 0) {
			return read;
		}
		return (0xFF & this.oneByteBuffer.flip().get());
	}

	@Override
	public void readFully(byte[] bytes) throws IOException {
		ByteBuffer bb = ByteBuffer.allocate(bytes.length);
					while (bb.hasRemaining()) {
						int read = this.channel.read(bb);
						if (read == -1) {
							throw new EOFException();
						}
					}
					bb.flip().get(bytes);
	}
```

parquet-common/src/main/java/org/apache/parquet/io/LocalSeekableInputStream.java
```diff
@@ -0,0 +1,47 @@
...
+    this.pos = 0;
+  }
+
+  @Override
+  public long getPos() throws IOException {
+    return pos;
+  }
+
+  @Override
+  public void seek(long newPos) throws IOException {
```
fommil, line 39: I don't see how this can work, the `pos` needs to be updated anytime the stream is progressed. From my reading of the DelegatingSeekableInputStream it is designed to work only for InputStreams that already track their position and can therefore be consulted. e.g. in

```scala
final class LocalInputFile(file: File) extends InputFile {
  def getLength() = file.length()
  def newStream(): SeekableInputStream = {
    val input = new FileInputStream(file)
    new DelegatingSeekableInputStream(input) {
      def getPos(): Long = input.getChannel().position()
      def seek(bs: Long): Unit = {
        val _ = input.getChannel().position(bs)
      }
    }
  }
}
```

from https://issues.apache.org/jira/browse/PARQUET-1953 which is relying on the underlying `FileInputStream` to keep track of its position in the stream.

PR #819, opened 2020-09-24
PARQUET-1915: Add nullify column

shangxinli:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
@@ -32,7 +32,6 @@ import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 import org.apache.parquet.hadoop.util.ColumnMasker;
 import org.apache.parquet.hadoop.util.ColumnMasker.MaskMode;
 import org.apache.parquet.hadoop.util.CompressionConverter.TransParquetFileReader;
-import org.apache.parquet.hadoop.util.CompressionConverter.TransParquetFileWriter;
 import org.apache.parquet.hadoop.util.HadoopInputFile;
 import org.apache.parquet.schema.MessageType;
 import org.slf4j.Logger;
@@ -90,7 +89,7 @@ public class ColumnMaskingCommand extends BaseCommand {
 
     ParquetMetadata metaData = ParquetFileReader.readFooter(getConf(), inPath, NO_FILTER);
     MessageType schema = metaData.getFileMetaData().getSchema();
-    TransParquetFileWriter writer = new TransParquetFileWriter(getConf(), schema, outPath, ParquetFileWriter.Mode.CREATE);
+    ParquetFileWriter writer = new ParquetFileWriter(getConf(), schema, outPath, ParquetFileWriter.Mode.CREATE);
     writer.start();
 
     try (TransParquetFileReader reader = new TransParquetFileReader(HadoopInputFile.fromPath(inPath, getConf()), HadoopReadOptions.builder(getConf()).build())) {
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.cli.commands;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.HadoopReadOptions;
+import org.apache.parquet.cli.BaseCommand;
+import org.apache.parquet.hadoop.ParquetFileReader;
+import org.apache.parquet.hadoop.ParquetFileWriter;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.hadoop.util.ColumnMasker;
+import org.apache.parquet.hadoop.util.ColumnMasker.MaskMode;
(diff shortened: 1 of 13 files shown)


shangxinli:
@gszadovszky @Fokko  Do you have time to review this change?

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
```diff
@@ -0,0 +1,104 @@
...
+  @Parameter(description = "<mask mode: nullify, hash, redact>")
+  String mode;
+
+  @Parameter(description = "<input parquet file path>")
+  String input;
+
+  @Parameter(description = "<output parquet file path>")
+  String output;
+
+  @Parameter(description = "columns to be repalced with masked value")
```
gszadovszky, line 64:
```suggestion
  @Parameter(description = "columns to be replaced with masked value")
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ColumnMasker.java
```diff
@@ -0,0 +1,274 @@
...
+        // we just write one null for the whole list at the top level, instead of nullify the elements in the list one by one
+        if (rlvl == 0) {
+          cWriter.writeNull(rlvl, dlvl - 1);
+        }
+      } else {
+        cWriter.writeNull(rlvl, dlvl);
+      }
+    }
+
+    BytesInput data = cWriter.concatWriters();
```
gszadovszky, line 150: So the concept is to write the nulls into a "hacked" `ColumnWrite`r and extract the bytes from it then write it as a page. This way you write the whole column chunk into one page. This breaks the concept of column-indexes that we write new pages for a couple of rows (default is 20k) even if the pages are small in bytes.
I think, it would be much better to use a proper `PageWriteStore` instead of a dummy one and let the original implementation decide when to create new pages. This way you do not need to extend the existing interface with `contactWriters` (which also breaks the current concept of the interface that is only a sink for the data).

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
```diff
@@ -0,0 +1,104 @@
...
+public class ColumnMaskingCommand extends BaseCommand {
+
+  private ColumnMasker masker;
+
+  public ColumnMaskingCommand(Logger console) {
+    super(console);
+    masker = new ColumnMasker();
+  }
+
+  @Parameter(description = "<mask mode: nullify, hash, redact>")
```
gszadovszky, line 55: I am not sure it is a good idea the different modes and throw exceptions that some of them are not supported/implemented. We are heading to the next major release and I think it is not a good user experience to have such features list as if they work properly but they are not implemented yet.

gszadovszky (changes requested):
Started reviewing but found a conceptual issue so stopped there and tried to explain.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ColumnMaskingCommand.java
shangxinli: > I am not sure it is a good idea the different modes and throw exceptions that some of them are not supported/implemented. We are heading to the next major release and I think it is not a good user experience to have such features list as if they work properly but they are not implemented yet.

Sounds good! I will remove unsupported ones from the list.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ColumnMasker.java
shangxinli: The problem is we need to make the class ColumnChunkPageWriteStore public because we are under the namespace of org.apache.parquet.hadoop.util which cannot use ColumnChunkPageWriteStore for now.

gszadovszky: We already have a lot of classes that are public because of we cannot fine-tune the java visibility but are for internal use only. I would suggest make it public if it is required for your design and comment that it is not for public use. You may want to use the yetus' `@Private` annotation as well.

shangxinli: sure

gszadovszky (changes requested):
NOTE: Travis failure is unrelated to this change. The problem is with the twitter maven repo (maven.twttr.com).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompressionConverter.java
```diff
@@ -53,10 +56,15 @@
...
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.apache.parquet.hadoop.ParquetWriter.DEFAULT_BLOCK_SIZE;
+import static org.apache.parquet.hadoop.ParquetWriter.MAX_PADDING_SIZE_DEFAULT;
```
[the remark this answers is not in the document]
shangxinli: Sounds good!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/CompressionConverter.java
```diff
@@ -267,5 +275,19 @@ public PageHeader readPageHeader() throws IOException {
     public long getPos() throws IOException {
       return f.getPos();
     }
+
+    public SeekableInputStream getStream() {
+      return f;
+    }
+  }
+
+  public static final class TransParquetFileWriter extends ParquetFileWriter {
```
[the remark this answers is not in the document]
shangxinli: It is a left over. I will remove it.

parquet-column/src/main/java/org/apache/parquet/column/ColumnWriter.java
```diff
@@ -91,5 +92,11 @@
    * @return the number of bytes of memory used to buffer the current data
    */
   long getBufferedSizeInMemory();
+
+  /**
+   * concat all the values writer in the order of repetitionLevel writer, definitionLevel writer, and data writer
+   * @return combined values in the type of ByteInput
+   */
+  BytesInput concatWriters();
```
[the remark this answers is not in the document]
shangxinli: yes, leftovers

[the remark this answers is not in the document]
shangxinli: I will remove

parquet-tools/src/main/java/org/apache/parquet/tools/command/ColumnMaskingCommand.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.parquet.schema.MessageType;
+
+import java.util.List;
+import java.util.Set;
+
+import static org.apache.parquet.format.converter.ParquetMetadataConverter.NO_FILTER;
+import org.apache.parquet.hadoop.util.ColumnMasker.MaskMode;
+import org.apache.parquet.hadoop.util.CompressionConverter.TransParquetFileWriter;
+
+public class ColumnMaskingCommand extends ArgsOnlyCommand {
```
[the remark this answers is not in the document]
shangxinli: Yes. I added the named arguments for cli because of PARQUET-1910. For tools, I see no other commands added. So just follow the pattern.

gszadovszky (approved):

gszadovszky:
Re-triggering travis builds...

PR #820, opened 2020-09-29
PARQUET-1917: Don't write values for oneOf fields that aren't set

dossett:
https://issues.apache.org/jira/browse/PARQUET-1917

### Jira

- [X] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-1917/) issues and references them in the PR title. 

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
@@ -337,6 +337,12 @@ public class ProtoWriteSupport<T extends MessageOrBuilder> extends WriteSupport<
         List<FieldDescriptor> fieldDescriptors = messageDescriptor.getFields();
         for (FieldDescriptor fieldDescriptor : fieldDescriptors) {
           FieldDescriptor.Type type = fieldDescriptor.getType();
+
+          //For a field in a oneOf that isn't set don't write anything
+          if (fieldDescriptor.getContainingOneof() != null && !pb.hasField(fieldDescriptor)) {
+            continue;
+          }
+
           if (!fieldDescriptor.isRepeated() && FieldDescriptor.Type.MESSAGE.equals(type) && !pb.hasField(fieldDescriptor)) {
             continue;
           }
--- a/parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoWriteSupportTest.java
+++ b/parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoWriteSupportTest.java
@@ -22,6 +22,7 @@ import com.google.protobuf.Message;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.junit.Test;
+import static org.junit.Assert.*;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
 import org.apache.parquet.io.api.Binary;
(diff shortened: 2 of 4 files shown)


parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -337,6 +337,12 @@ private void writeAllFields(MessageOrBuilder pb) {
         List<FieldDescriptor> fieldDescriptors = messageDescriptor.getFields();
         for (FieldDescriptor fieldDescriptor : fieldDescriptors) {
           FieldDescriptor.Type type = fieldDescriptor.getType();
+
+          //For a field in a oneOf that isn't set don't write anything
+          if (fieldDescriptor.getContainingOneof() != null && !pb.hasField(fieldDescriptor)) {
+            continue;
+          }
```
dossett, line 344: Maybe this approach is too simplistic, but it does seem like exactly the behavior needed.

belugabehr: What about the read side?  Is there going to be any issue there?  Currently a value is always written, so there is a value to read.  Now that value is missing.

dossett: @belugabehr That's a great question.  Everything seems ok when I tested it out locally. I wrote some data out to a parquet file with this change, read the data back into protobuf messages, and then wrote the data back to a second parquet file.  The two parquet files were identical, so everything round tripped successfully.  

I am struggling a bit to write unit tests for this case, but that's probably me getting up the parquet API learning curve.

dossett: To be clear, I will try to add tests for this to my PR.

belugabehr:
Thanks for the collaboration here.

For the unit tests you are using Java 'assert' functions.  It is required to use the JUnit Assert methods (look at other tests for reference).

belugabehr:
Please add a test for setting field 1 and not setting field 2.  Since setting the fields should reset the values for all other fields in the 'oneof' field, it could be possible that field 2, set to '0', would clear the value of the first field.  If it just so happens that Parquet is loading the fields in the order they are defined, field 2 would be last and would overwrite all the previous values, so we don't really understand here what is happening to those fields.

dossett:
Great feedback as always @belugabehr !

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoWriteSupportTest.java
```diff
@@ -913,4 +918,62 @@ public void testMessageWithExtensions() throws Exception {
...
+
+    TestProto3.OneOfTestMessage.Builder msgBuilder3 = TestProto3.OneOfTestMessage.newBuilder();
+    msgBuilder3.setFirst(42);
+    TestProto3.OneOfTestMessage theMessageFirstSet = msgBuilder3.build();
+
+    //Write them out and read them back
+    Path tmpFilePath = TestUtils.writeMessages(theMessage, theMessageNothingSet, theMessageFirstSet);
+    List<TestProto3.OneOfTestMessage> gotBack = TestUtils.readMessages(tmpFilePath, TestProto3.OneOfTestMessage.class);
+
+    //First message
```
belugabehr, line 964: Bit of a nit here, but if you find yourself putting in comments like "this is test 1, this is test 2, this is test 3, etc." that is a clear sign that you need to break the test into multiple test methods.

If a test fails, it will be much easier to figure out exactly what broke instead of having to step through all this code.  It also ensure that the tests are testing what is intended and that there aren't any side-effects/artifacts of the other test scenarios interfering.

dossett:
@gszadovszky Tagging you on this PR per discussion in the dev list.  If you approve the change I will also clean up the new tests a bit per David's comments.  Thanks!

gszadovszky:
@dossett, it looks good to me. Let me wait for @belugabehr's approval then I'll approve and push this.

belugabehr (approved):

belugabehr:
I looked at the unit test again, it's fine as-is.  +1

Thanks.

gszadovszky (approved):

dossett:
Thank you both!

PR #824, opened 2020-10-07
PARQUET-1920: Fix Parquet writer's memory check interval calculation

SinghAsDev:
Fix Parquet writer's memory check interval calculation and throw helpful message while dealing with too large column chunks.

### Jira

- [x] My PR addresses [PARQUET-1920](https://issues.apache.org/jira/browse/PARQUET-1920).
  - https://issues.apache.org/jira/browse/PARQUET-1920

### Tests

- [ ] My PR does not necessarily needs any specific test as it is removing hard coded values in favor of configs. However, I can think of ways to test if insisted.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
@@ -217,10 +217,10 @@ abstract public class BytesInput {
   public byte[] toByteArray() throws IOException {
     long size = size();
     if (size > Integer.MAX_VALUE) {
-      throw new IOException("Page size, " + size + ", is larger than allowed " + Integer.MAX_VALUE +"." +
+      throw new IOException("Page size, " + size + ", is larger than allowed " + Integer.MAX_VALUE + "." +
         " Usually caused by a Parquet writer writing too big column chunks on encountering highly skewed dataset." +
         " Please set page.size.row.check.max to a lower value on the writer, default value is 10000." +
-        " You can try setting it to " + (10000 / (size/ Integer.MAX_VALUE)) + " or lower.");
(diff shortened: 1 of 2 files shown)


SinghAsDev:
Hey @gszadovszky @shangxinli what do you guys think of this?

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -215,6 +215,13 @@ public static BytesInput copy(BytesInput bytesInput) throws IOException {
    * @throws IOException if there is an exception reading
    */
   public byte[] toByteArray() throws IOException {
+    long size = size();
+    if (size > Integer.MAX_VALUE) {
+      throw new IOException("Page size, " + size + ", is larger than allowed " + Integer.MAX_VALUE +"." +
+        " Usually caused by a Parquet writer writing too big column chunks on encountering highly skewed dataset." +
+        " Please set page.size.row.check.max to a lower value on the writer, default value is 10000." +
+        " You can try setting it to " + (10000 / (size/ Integer.MAX_VALUE)) + " or lower.");
```
gszadovszky, line 223: nit:
```suggestion
      throw new IOException("Page size, " + size + ", is larger than allowed " + Integer.MAX_VALUE + "." +
        " Usually caused by a Parquet writer writing too big column chunks on encountering highly skewed dataset." +
        " Please set page.size.row.check.max to a lower value on the writer, default value is 10000." +
        " You can try setting it to " + (10000 / (size / Integer.MAX_VALUE)) + " or lower.");
```

gszadovszky (changes requested):
Thanks a lot for fixing this. I only found some little formatting issues.

gszadovszky (approved):

PR #831, opened 2020-10-19
PARQUET-1928: Interpret Parquet INT96 type as FIXED[12] AVRO Schema

anantdamle:

Reading Parquet files in Apache Beam using ParquetIO uses `AvroParquetReader` causing it to throw `IllegalArgumentException("INT96 not implemented and is deprecated")`

Customers have large datasets which can't be reprocessed again to convert into a supported type. An easier approach would be to convert into a byte array of 12 bytes, that can then be interpreted by the developer in any way they want to interpret it.

This patch interprets the INT96 parquet type as a byte array of 12-bytes. the developer/user can then handle it appropriate to interpret into a timestamp or simple some bytes.

- [x ] My PR adds the following unit tests `testParquetInt96AsFixed12AvroType` and `testParquetInt96DefaultFail`

https://issues.apache.org/jira/browse/PARQUET-1928

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ * 
  *   http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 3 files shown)


anantdamle:
Adding @rdblue  @tomwhite for review

anantdamle:
Gentle bump up

gszadovszky:
Parquet community was against adding INT96 support to not to encourage our clients to use it. While I understand the requirement of supporting the already written types. (Meanwhile as parquet-avro did not support INT96 ever this change is required for developments of new functionalities depending on the deprecated INT96.)
Anyway, I am fine with this change but I do not really like that it works by default. What do you think about keeping the original behavior by default and introduce a configuration flag to switch it on? (See `writeParquetUUID` as an example.) This way we still not encourage the clients to use INT96 but have the option to do so if it is necessary.

anantdamle:
thanks for the review. I have incorporated the changes requested.

gszadovszky (approved):

anantdamle:
@gszadovszky Thanks for the approval. 
sorry for a noob question. What happens next? how does this PR get merged in master?

gszadovszky:
@anantdamle, thank you for the contribution!
I will do it. Just usually wait 24 hours (because of the weekend this time a bit more) to give a chance for others to comment before it gets merged.

anantdamle:
Thanks @gszadovszky, quick request to kindly squash and merge as there are 2 - useless commits to rectify my IDE's autochange to LICENSE comments.

gszadovszky:
@anantdamle, our usual process is to squash all the changes related to one jira before merging. Thanks a lot for your contribution!

PR #838, opened 2020-11-01
PARQUET-1940: KEK length configuration

ggershinsky:

PR #841, opened 2020-11-03
PARQUET-1939: Fix remote KMS client ambiguity

ggershinsky:
1. rename RemoteKmsClient class to LocalWrapKmsClient; it will be used only in rare situations where in-server wrapping in not supported. 

2. this also means the "wrap.locally" property is not needed anymore, making the API slightly simpler.

### DIFF
--- a/parquet-hadoop/README.md
+++ b/parquet-hadoop/README.md
@@ -418,13 +418,6 @@ If `false`, DEKs are directly encrypted with master keys, KEKs are not used.
 
 ---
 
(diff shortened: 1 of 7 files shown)


ggershinsky:
@gszadovszky Thanks for approving the #838. We have a couple of additional PRs - this one (841) and the #842, will appreciate their review. 
There is nothing else in our pipelines.

gszadovszky (approved):

gszadovszky:
Re-triggering travis builds...

PR #847, opened 2020-12-14
[PARQUET-1951] Allow different strategies to combine key values when …

satishkotha:
…merging parquet files

### Jira

- [ X] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-1951) issues and references them in the PR title.

### Tests

- [X ] My PR adds the following unit tests: TestParquetFileWriter#testMergeMetadataWithNoConflictingKeyValues, and testMergeMetadataWithConflictingKeyValues

### Commits

- [ X] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":

### Documentation

- [ X] In case of new functionality, my PR adds documentation that describes how to use it.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ConcatenatingKeyValueMetadataMergeStrategy.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ConcatenatingKeyValueMetadataMergeStrategy.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.hadoop.metadata;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ *  Strategy to concatenate if there are multiple values for a given key in metadata.
+ *  Note: use this with caution. This is expected to work only for certain use cases.
+ */
+public class ConcatenatingKeyValueMetadataMergeStrategy implements KeyValueMetadataMergeStrategy {
+  private static final String DEFAULT_DELIMITER = ",";
+  
+  private final String delimiter;
+
+  /**
+   * Default constructor.
+   */
+  public ConcatenatingKeyValueMetadataMergeStrategy() {
+    this.delimiter = DEFAULT_DELIMITER;
+  }
+
+  /**
+   * Constructor to use different delimiter for concatenation.
+   * 
+   * @param delim delimiter char sequence.
+   */
+  public ConcatenatingKeyValueMetadataMergeStrategy(String delim) {
+    this.delimiter = delim;
+  }
+  
+  /**
+   * @param keyValueMetaData the merged app specific metadata
+   */
+  public Map<String, String> merge(Map<String, Set<String>> keyValueMetaData) {
+    Map<String, String> mergedKeyValues = new HashMap<String, String>();
+    for (Map.Entry<String, Set<String>> entry : keyValueMetaData.entrySet()) {
+      mergedKeyValues.put(entry.getKey(), entry.getValue().stream().collect(Collectors.joining(this.delimiter)));
+    }
+    return mergedKeyValues;
+  }
+}
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/DefaultKeyValueMetadataMergeStrategy.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/DefaultKeyValueMetadataMergeStrategy.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.hadoop.metadata;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ *  Default strategy to throw errors in metadata if there are multiple values for a given key in metadata.
+ */
+public class DefaultKeyValueMetadataMergeStrategy implements KeyValueMetadataMergeStrategy {
+  /**
+   * @param keyValueMetaData the merged app specific metadata
+   *
+   * @throws NullPointerException if keyValueMetaData is {@code null}
+   */
+  public Map<String, String> merge(Map<String, Set<String>> keyValueMetaData) {
+    Map<String, String> mergedKeyValues = new HashMap<String, String>();
+    for (Map.Entry<String, Set<String>> entry : keyValueMetaData.entrySet()) {
+      if (entry.getValue().size() > 1) {
+        throw new RuntimeException("could not merge metadata: key " + entry.getKey() + " has conflicting values: " + entry.getValue());
+      }
+      mergedKeyValues.put(entry.getKey(), entry.getValue().iterator().next());
+    }
+    return mergedKeyValues;
+  }
+}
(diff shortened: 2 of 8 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/KeyValueMetadataMergeStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+ * 1) Expect metadata to be exactly identical. Otherwise throw error (this is the default strategy implementation)
+ * 2) Custom strategy to merge metadata if the values are not identical. Example: concatenate values.
+ */
+public interface KeyValueMetadataMergeStrategy extends Serializable {
+  /**
+   * @param keyValueMetaData the merged app specific metadata
+   *
+   * @throws NullPointerException if keyValueMetaData is {@code null}
+   */
+  Map<String, String> merge(Map<String, Set<String>> keyValueMetaData);
```
gszadovszky, line 43: I think, this interface does not support all types of the potential merge strategy implementations. For example if you would like to use the value which is common in the most files you cannot do it because you only have the distinct values. What do you think about using a `List<String>` instead of a `Set<String>`?

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -41,6 +46,21 @@
           "   <output> is the destination parquet file"
   };
 
+  private static final String DEFAULT_KEY_VALUE_MERGE_STRATEGY = new DefaultKeyValueMetadataMergeStrategy().getClass().getName();
```
gszadovszky, line 49: You do not need to instantiate an object to get its class. Simply use `DefaultKeyValueMetadataMergeStrategy.class`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/DefaultKeyValueMetadataMergeStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.parquet.hadoop.metadata;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ *  Default strategy to throw errors in metadata if there are multiple values for a given key in metadata.
+ */
+public class DefaultKeyValueMetadataMergeStrategy implements KeyValueMetadataMergeStrategy {
```
gszadovszky, line 28: I do not really like the naming of this class. This is the default implementation, it's fine but the name should suggest what it does. I'm quite sure there is a better naming but what about `StrictKeyValueMetadataMergeStrategy`?

gszadovszky (changes requested):
I have a couple of comments but the change itself looks good.

Meanwhile, I have some problems with the concept. `parquet-tools` is a simple command in most of the environments. It means that it is not clear how to extend it with your own implementation.
I would suggest adding this change to the API only (in `parquet-hadoop`) so you can extend the implementation from the code. If you want to extend `parquet-tools` as well I would suggest adding your implementation to parquet and let the user choose from the existing implementations (by its simple name like `strict` or others not the class name).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/DefaultKeyValueMetadataMergeStrategy.java
satishkotha: Good suggestion. Fixed it. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/KeyValueMetadataMergeStrategy.java
satishkotha: GlobalMetadata uses Map<String, Set<String>> https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/GlobalMetaData.java#L41. So merge method takes in same argument.

Do you think it makes sense to change GlobalMetadata to track items in List<String>? I'm not sure if it would be backward compatible and if there are use cases expecting this to be Set.  

Without changing GlobalMetadata#keyValueMetaData, it may not be useful to change this interface. Let me know what you think.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
satishkotha: My bad. Fixed. Thanks

satishkotha:
> I have a couple of comments but the change itself looks good.
> 
> Meanwhile, I have some problems with the concept. `parquet-tools` is a simple command in most of the environments. It means that it is not clear how to extend it with your own implementation.
> I would suggest adding this change to the API only (in `parquet-hadoop`) so you can extend the implementation from the code. If you want to extend `parquet-tools` as well I would suggest adding your implementation to parquet and let the user choose from the existing implementations (by its simple name like `strict` or others not the class name).

@gszadovszky I added one more strategy for merging key values. I would prefer to keep this in parquet-tools to make this easy to run for one-off debugging use cases.  Let me know if you strong opinion against this.

Also, the actual merge strategy I need is specific to my use-case. So I'd like to keep referencing the class name. Here is an example to help explain:

This is how I run parquet-tools
java -cp  my-app-jar:parquet-tools/target/parquet-tools-1.12.0-SNAPSHOT.jar:<hadoop-jars> merge --s <myClass> f1.parquet f2.parquet merged.parquet

Parquet files we write have additional metadata "range" and "count" (Count is not rowcount, its specific to my usecase tracking distinct values for a column).

1) file f1.parquet metadata will have 

- key: "range", value : "a-d"
- key: "count", value: "12000"

2) file f2.parquet metadata will have 
- key: "range", value: "e-f"
- key: "count", value: "10000"

When we merge, we want to resulting parquet file to have metadata 
- key: "range", value: "a-f"
- key: "count", value: "15000"

So merge strategy is different for different keys we store in parquet footer. This is specific to how we use parquet and very likely not usable for other use cases, so I prefer not to expose this strategy in parquet library.

Let me know if you have any suggestions.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/KeyValueMetadataMergeStrategy.java
shangxinli: People may already rely on the 'set' to do the dedup.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ConcatenatingKeyValueMetadataMergeStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+public class ConcatenatingKeyValueMetadataMergeStrategy implements KeyValueMetadataMergeStrategy {
+  /**
+   * @param keyValueMetaData the merged app specific metadata
+   *
+   * @throws NullPointerException if keyValueMetaData is {@code null}
+   */
+  public Map<String, String> merge(Map<String, Set<String>> keyValueMetaData) {
+    Map<String, String> mergedKeyValues = new HashMap<String, String>();
+    for (Map.Entry<String, Set<String>> entry : keyValueMetaData.entrySet()) {
+      mergedKeyValues.put(entry.getKey(), entry.getValue().stream().collect(Collectors.joining(",")));
```
shangxinli, line 39: The delimiter needs to be parameterized in other non-hudi cases.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/KeyValueMetadataMergeStrategy.java
gszadovszky: Sorry, I've missed this type is already in our API. Agreed to keep the interface as is.

gszadovszky:
@satishkotha, thanks for explaining your use case. I understand that you need to inject your own implementation to make it work. What I would like to achieve is to keep parquet-tools easy to use as a command. 

I do not have a strong opinion to not to allow the runtime extension of parquet-tools functionality but it have to be well documented and easy to use from the command line. Please keep in mind that parquet-tools might be shipped as a command to an environment (e.g. via brew) and the user might not have the knowledge or even the privilege to modify the classpath.

I would suggest adding two separate arguments for merge. One is for selecting a strategy that is available in parquet-mr. For this case we shall list the available options and the descriptions of these options. Another argument might be available for the user's own implementation. For this one it needs to be documented which interface is needed to be implemented and the fact that the implementations needs to be added to the classpath of parquet-tools. (It would be hard to give a step-by-step guide for a beginner as you would not know anything about the environment and paths but I think it is enough to keep this option for experts.)

satishkotha:
> @satishkotha, thanks for explaining your use case. I understand that you need to inject your own implementation to make it work. What I would like to achieve is to keep parquet-tools easy to use as a command.
> 
> I do not have a strong opinion to not to allow the runtime extension of parquet-tools functionality but it have to be well documented and easy to use from the command line. Please keep in mind that parquet-tools might be shipped as a command to an environment (e.g. via brew) and the user might not have the knowledge or even the privilege to modify the classpath.
> 
> I would suggest adding two separate arguments for merge. One is for selecting a strategy that is available in parquet-mr. For this case we shall list the available options and the descriptions of these options. Another argument might be available for the user's own implementation. For this one it needs to be documented which interface is needed to be implemented and the fact that the implementations needs to be added to the classpath of parquet-tools. (It would be hard to give a step-by-step guide for a beginner as you would not know anything about the environment and paths but I think it is enough to keep this option for experts.)

These are great points for usability. I updated MergeCommand to take in additional option and updated documentation. Please take a look. I'm happy to add any additional documentation if needed.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
```diff
@@ -46,19 +47,34 @@
...
+  
   public static final Options OPTIONS;
 
   static {
     OPTIONS = new Options();
     Option mergeStrategy = Option.builder("s")
       .longOpt("mergeStrategy")
       .desc("Strategy to merge (key, value) pairs in metadata if there are multiple values for same key " +
-        "(default: " + DEFAULT_KEY_VALUE_MERGE_STRATEGY + ")")
+        "(default: " + MergeStrategy.STRICT + "). You can provide your custom implementation by specifying " +
```
gszadovszky, line 63: We should list the available options here. (The best way would be do it by iterating over the enum so a potential extension later would not forget about this part.) 

nit: in command line we rarely use upper case arguments. Maybe it would be nicer to convert the enum names to lower case.

gszadovszky (changes requested):
Thanks a lot for your efforts. It seems great. I have one more note, though.

parquet-tools/src/main/java/org/apache/parquet/tools/command/MergeCommand.java
satishkotha: Ok. Done.

gszadovszky (approved):

PR #849, opened 2020-12-15
PARQUET-1954: TCP connection leak in parquet dump

StefanXiepj:

### Jira
My PR addresses the following [PARQUET-1954](https://issues.apache.org/jira/browse/PARQUET-1954T/)

### Tests
does not need testing

### DIFF
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
@@ -203,6 +203,7 @@ public class DumpCommand extends ArgsOnlyCommand {
             } finally {
               if (freader != null) {
                 freader.close();
+                freader = null;
               }
             }
           }
@@ -238,6 +239,7 @@ public class DumpCommand extends ArgsOnlyCommand {
                     out.flushColumns();
                     if (freader != null) {
                         freader.close();
+                        freader = null;
                     }
                 }
             }
--- a/parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
+++ b/parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
@@ -163,31 +163,30 @@ public class DumpCommand extends ArgsOnlyCommand {
             }
         }
 
-        ParquetFileReader freader = null;
         if (showmd) {
           long group = 0;
           for (BlockMetaData block : blocks) {
-            try {
-              if (group != 0)
-                out.println();
-              out.format("row group %d%n", group++);
-              out.rule('-');
+            if (group != 0)
+              out.println();
+            out.format("row group %d%n", group++);
+            out.rule('-');
 
-              List<ColumnChunkMetaData> ccmds = block.getColumns();
-              if (showColumns != null) {
-                ccmds = new ArrayList<ColumnChunkMetaData>();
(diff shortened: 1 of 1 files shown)


StefanXiepj:
@Pavitheran Could you review this PR , thanks.

parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
```diff
@@ -165,48 +165,47 @@ public static void dump(PrettyPrintWriter out, ParquetMetadata meta, MessageType
...
+
+                store = freader.readNextRowGroup();
+              }
+              out.flushColumns();
             } finally {
-                if (freader != null) {
-                    freader.close();
-                }
+              if (freader != null) {
+                freader.close();
```
gszadovszky, line 205: If an I/O exception occurs in the `ParquetFileReader` constructor then you might try to close the previous reader object. I would guess there is no problem with invoking `close` multiple times on the same object but who knows? It might be a good idea to reset the reference so you would not close the same object again. Another idea would be to use try-with-resources for the code part where the reader is really needed.

gszadovszky (changes requested):
Thanks a lot for your contribution. Please, check my notes.

parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
StefanXiepj: This is the memory used before and after fix this bug:
image

StefanXiepj:
@wangyum Could you help me to review this PR ? thanks.

parquet-tools/src/main/java/org/apache/parquet/tools/command/DumpCommand.java
gszadovszky: I did not say your fix is not working. What I suggest is to avoid closing a reader twice if an issue happens in the `ParquetFileReader` constructor.

StefanXiepj: yeah yeah,  thx for your review, i am sorry, please forgive me. Do you mean we can fix it like this:
`
...
            } finally {
              if (freader != null) {
                freader.close();
                // reset current reader is null, avoid closing a reader twice
                freader = null;
              }
            }
`
or
`
try(ParquetFileReader freader = new ParquetFileReader(conf, meta.getFileMetaData(), inpath, rblocks, columns)){
// do something where the freader is really needed.
}
`
is right ?

StefanXiepj: hi, @gszadovszky Could you help me to review this PR ? thanks.

Fokko: The `ParquetFileReader` is Closeable, I would suggest using the try-with-resource pattern: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

StefanXiepj: @Fokko thanks for your suggestion, can you review it again?

gszadovszky (approved):

shangxinli:
Good catch! Thanks for the effort!

PR #852, opened 2021-01-11
PARQUET-1851: fix parquet metadata converter NPE

chenjunjiedada:
This fixes the NPE as below:

```
java.lang.NullPointerExceptionjava.lang.NullPointerException at
org.apache.parquet.format.converter.ParquetMetadataConverter.addRowGroup(ParquetMetadataConverter.java:476) at
org.apache.parquet.format.converter.ParquetMetadataConverter.toParquetMetadata(ParquetMetadataConverter.java:177) at
org.apache.parquet.hadoop.ParquetFileWriter.serializeFooter(ParquetFileWriter.java:914) at
org.apache.parquet.hadoop.ParquetFileWriter.end(ParquetFileWriter.java:864) at
org.apache.iceberg.parquet.ParquetWriter.close(ParquetWriter.java:206) at
org.apache.iceberg.data.TestLocalScan.writeFile(TestLocalScan.java:429)
```

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -861,7 +861,7 @@ public class ParquetFileWriter {
    */
   public void endBlock() throws IOException {
     if (currentRecordCount == 0) {
-      throw new IOException("End block with zero record");
+      throw new ParquetEncodingException("End block with zero record");
     }
 
     state = state.endBlock();
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -130,7 +130,7 @@ public class ParquetFileWriter {
 
   // The Bloom filters
   private final List<Map<String, BloomFilter>> bloomFilters = new ArrayList<>();
-  
+
   // The file encryptor
   private final InternalFileEncryptor fileEncryptor;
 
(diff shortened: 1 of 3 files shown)


shangxinli (approved):
Good catch!

chenjunjiedada:
@gszadovszky Thanks for the comment and it makes sense to me. How about throwing an exception when detecting the row group's count is zero? Does that sound reasonable to you?

PS. the space changes in the PR is automatically fixed by idea IDE. If you don't like it, I can handle it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -860,6 +860,10 @@ public void endColumn() throws IOException {
    * @throws IOException if there is an error while writing
    */
   public void endBlock() throws IOException {
+    if (currentRecordCount == 0) {
+      throw new IOException("End block with zero record");
```
gszadovszky, line 864: I would suggest using a `ParquetEncodingException` instead.

gszadovszky (changes requested):
Thanks a lot for the changes. Please find my only comment in the code. Otherwise it looks good to me.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
chenjunjiedada: SGTM.

gszadovszky (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
vdiravka: The empty parquet files was used in Drill earlier and now this change breaks the possibility to store empty tables (with header schema only) in parquet files. Any suggestions to bypass this and integrate Parquet-1.12.0 to Drill?

gszadovszky: @vdiravka, you might want to create a separate jira about this topic so we can discuss it in a more open way. Please, also describe what the empty parquet files are used for.

vdiravka: @gszadovszky Done
https://issues.apache.org/jira/browse/PARQUET-2026

PR #855, opened 2021-01-19
PARQUET-1964: FOLLOWUP: Avoid constructing useless ArrayList

wangyum:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -851,23 +851,22 @@ public class ParquetFileReader implements Closeable {
   }
 
   private List<BlockMetaData> filterRowGroups(List<BlockMetaData> blocks) throws IOException {
-    // set up data filters based on configured levels
-    List<RowGroupFilter.FilterLevel> levels = new ArrayList<>();
-
-    if (options.useStatsFilter()) {
-      levels.add(STATISTICS);
-    }
-
-    if (options.useDictionaryFilter()) {
-      levels.add(DICTIONARY);
-    }
-
-    if (options.useBloomFilter()) {
-      levels.add(BLOOMFILTER);
-    }
-
     FilterCompat.Filter recordFilter = options.getRecordFilter();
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -827,7 +827,7 @@ public long getRecordCount() {
   }
 
   public long getFilteredRecordCount() {
-    if (!options.useColumnIndexFilter()) {
+    if (!options.useColumnIndexFilter() && options.getRecordFilter() != null) {
```
wangyum, line 830: How to reproduce this issue:
```scala
val hadoopInputFile = HadoopInputFile.fromPath(new Path("/path/to/parquet/000.snappy.parquet"), new Configuration())
val reader = ParquetFileReader.open(hadoopInputFile)
val recordCount = reader.getFilteredRecordCount
reader.close()
```

wangyum:
cc @gszadovszky

gszadovszky:
Thanks a lot for working on this. It is a good catch!

I've had to investigate a bit why we do not catch this in the unit tests. The answer is that in most cases a [NOOP](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/filter2/compat/FilterCompat.java#L62) filter is created instead of having a `null` value. We use this `NOOP` if no filter is specified in the builder/config. Meanwhile it is possible to set a `null` in the builder/config so the NPE may occur.
I've also realized that in older filter implementations (row group level min/max, dictionary) the `NOOP` filter is fine and do not have significant performance costs over a `null` check. Meanwhile for column index of bloom filter based filtering do have performance costs because they read the related data from the file even for `NOOP`.

If you don't mind, @wangyum, I would like to take care of this (including the NPE and the potential performance issues).

wangyum:
@gszadovszky Please take care of this.

wangyum:
cc @gszadovszky

gszadovszky (approved):
Thanks for fixing this. LGTM.

There are a couple unrelated Travis failures for our PRs. Let's wait for a green build.

Fokko (approved):

PR #856, opened 2021-01-19
PARQUET-1964: Properly handle missing/null filter

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -527,7 +527,7 @@ public class ParquetFileReader implements Closeable {
     return readFooter(file, options, f, converter);
   }
 
-  private static final ParquetMetadata readFooter(InputFile file, ParquetReadOptions options, 
+  private static final ParquetMetadata readFooter(InputFile file, ParquetReadOptions options,
       SeekableInputStream f, ParquetMetadataConverter converter) throws IOException {
 
     long fileLen = file.getLength();
@@ -579,7 +579,7 @@ public class ParquetFileReader implements Closeable {
 
     // Regular file, or encrypted file with plaintext footer
     if (!encryptedFooterMode) {
-      return converter.readParquetMetadata(footerBytesStream, options.getMetadataFilter(), fileDecryptor, false, 
+      return converter.readParquetMetadata(footerBytesStream, options.getMetadataFilter(), fileDecryptor, false,
           fileMetadataLength);
     }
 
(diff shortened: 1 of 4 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1046,6 +1046,8 @@ private ColumnIndexStore getColumnIndexStore(int blockIndex) {
   }
 
   private RowRanges getRowRanges(int blockIndex) {
+    assert FilterCompat
```
shangxinli, line 1049: Do we create an obstacle for testing with NOOP filters?

gszadovszky: We already test `NOOP` in many places because this one is the default if no filter is present. `null` filter is only possible if the user explicitly set it to `null` in the `ParquetReadOptions`.
This assert is only to ensure that this private method will not be invoked with a `null` or a `NOOP` filter. No "real" checking is required because the callers are in the same class and they already check the filter.

shangxinli: This private method is only called in two places and both ensure NOOP won't present. That is probably why the tests with NOOP in many places can pass. Given it is a private method, that is fine but it reminds me that the public method calculateRowRanges() doesn't check if the passed-in filter is null/noop. Do we need to check that too? 

Another thing is should we use assert or exception because assert could be turned off. There is some discussion about using assert or exception in stack overflow(https://stackoverflow.com/questions/1276308/exception-vs-assertion#:~:text=Use%20assertions%20for%20internal%20logic,should%20be%20explicit%20using%20exceptions.).

gszadovszky: I think, we do not need to cover the cases of `NOOP` or `null` filters in every places of the API. `ColumnIndexFilter` is not really for public use. I do not see why a client would use it directly. I think, the current handling is correct: NPE is thrown in case of null and the correct filtering is done otherwise.

What I wanted to cover here is to handle the `null` filters properly where the user faces it (by using setting the filter in `ParquetReadOptions`) and also to ensure there are no unnecessary reads (for column index or bloom filter) in case of `NOOP`.

I think this issue is not a bug anyway. There is no correctness issue if the user does not set the filter explicitly to `null`.

About `assert`. I think this is a good tool to ensure something that should happen anyway. I agree it is not suitable for checking anything in a public method but where the caller is properly controlled (e.g. for a private method) I think, it is fine. I think `assert` is fine if you could simply skip it without the risk of any failures. While keeping it in a private method does not have a performance penalty in production environment the unit tests would catch any issue if the callers are modified.

shangxinli:
LGTM

PR #858, opened 2021-01-29
PARQUET-1966: Fix build with JDK11 for JDK8

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -393,6 +393,7 @@
         <!-- Override source and target from the ASF parent -->
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
         <configuration>
           <argLine>-XX:MaxPermSize=256m</argLine>
           <source>${maven.compiler.source}</source>
@@ -624,6 +625,24 @@
         <surefire.logLevel>WARN</surefire.logLevel>
         <surefire.argLine>-Xmx512m -XX:MaxJavaStackTraceDepth=10</surefire.argLine>
       </properties>
+    </profile>
+
+    <profile>
+      <id>jdk9+</id>
+      <activation>
+        <jdk>[1.9,)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <release>8</release>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
     </profile>
 
   </profiles>


dossett:
That's a nice use of profiles.  I'm (non-binding) +1. I don't see any logs for the travis build failures, don't know if they've expired or maybe the tests just failed to launch for some reason.

gszadovszky:
Thanks, @dossett. I am not sure about the Travis status either. I've restarted the build, let's see what happens.

gszadovszky:
Close/re-open to trigger tests.
@shangxinli, if you have time could you please review this? The release is blocked on this one.

wangyum (approved):

pom.xml
```diff
@@ -393,6 +393,7 @@
         <!-- Override source and target from the ASF parent -->
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
```
Fokko, line 396: We could also bump the parent POM to 22: https://github.com/apache/maven-apache-parent/blob/apache-22/pom.xml

shangxinli:
For the failed test "testMemoryManagerUpperLimit", I am not sure it is caused by this change, or the test itself is unstable.

pom.xml
gszadovszky: I would not update it for now (just before the release). I think, it is better to bump such versions for the next release. The bump of the `maven-compiler-plugin` was required to support the `release` config.

gszadovszky:
> For the failed test "testMemoryManagerUpperLimit", I am not sure it is caused by this change, or the test itself is unstable. It seems the failure is because the pool size is larger than expected.
> 
> "should be within 10% of the expected value (expected = 453745044 actual = 505046624)"

Yes, the test is flaky. Based on git history we've had to increase the percentage 3 times already. (I am not sure if this test make sense this way but I did not want to drop it for now.) Meanwhile, I have not seen this test fail until the transition to the github actions. I guess there are some environmental differences between the new actions and Travis.

shangxinli:
> > For the failed test "testMemoryManagerUpperLimit", I am not sure it is caused by this change, or the test itself is unstable. It seems the failure is because the pool size is larger than expected.
> > "should be within 10% of the expected value (expected = 453745044 actual = 505046624)"
> 
> Yes, the test is flaky. Based on git history we've had to increase the percentage 3 times already. (I am not sure if this test make sense this way but I did not want to drop it for now.) Meanwhile, I have not seen this test fail until the transition to the github actions. I guess there are some environmental differences between the new actions and Travis.
> 
> But this change is not related to the flaky test. It is related to the issue caused by the first RC built with jdk11 and failed in a jre8 runtime.

Sounds good

shangxinli (approved):

PR #861, opened 2021-02-02
PARQUET-1970: Make minor releases source compatible

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
+++ b/parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
@@ -32,6 +32,6 @@ public interface OutputFile {
   long defaultBlockSize();
 
   default String getPath() {
-    throw new UnsupportedOperationException();
+    return null;
   }
 }
--- a/parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
+++ b/parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
@@ -31,5 +31,7 @@ public interface OutputFile {
 
   long defaultBlockSize();
 
-  String getPath();
+  default String getPath() {
+    throw new UnsupportedOperationException();
+  }
 }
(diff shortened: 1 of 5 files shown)


parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
```diff
@@ -31,5 +31,7 @@
 
   long defaultBlockSize();
 
-  String getPath();
+  default String getPath() {
+    throw new UnsupportedOperationException();
```
ggershinsky, line 35: We're consuming this function in one of the encryption options:
https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L284
If there is no encryption, or encryption is used with a crypto factory that doesn't need the file path - this exception will break the existing callers. A simple way to resolve this is by catching the exception here, and passing a null file path. This allows for these callers to run ok. The crypto factories that do need the file path sometimes (like the one we have in parquet-mr) should check `path==null` in situations where the file path is used, and throw an exception there. What do you think?

gszadovszky: It sounds good to me. I'll exchange the default implementation to return `null`.

ggershinsky: Cool, thanks. I'll add then the null checks in the crypto-factory (key management).

gszadovszky: I think, I've added them. Please review the latest commit.

ggershinsky: Oh, this is already covered in the commit. Looks good!

gszadovszky:
@shangxinli, if you have time could you please review this? The release is blocked on this one.

parquet-common/src/main/java/org/apache/parquet/io/OutputFile.java
shangxinli: +1 for using null instead of through exception, which is expensive.

shangxinli (approved):

Fokko (approved):

PR #866, opened 2021-02-09
Update Maven plugin versions

martin-g:
Use scala-maven-plugin instead of the old maven-scala-plugin.
Update Scala version to 2.12.12

### Jira

### Tests

### Commits

### Documentation


parquet-scala/pom.xml
```diff
@@ -46,13 +46,13 @@
       <url>https://scala-tools.org/repo-releases</url>
     </pluginRepository>
   </pluginRepositories>
-    
+
```
shangxinli, line 49: Probably IDE does that but let's avoid unnecessary change here. If other pending PRs touch the same line, it would conflict.

martin-g: @shangxinli I've re-introduced the empty spaces!

Fokko (approved):
Looks good. One related comment, I think it is better, in the end, to inherit versions from the Parent Apache POM. WDYT?

martin-g:
>  I think it is better, in the end, to inherit versions from the Parent Apache POM. WDYT?

Let me know if this is what Parquet team prefers and I will update the PR!

parquet-scala/pom.xml
```diff
@@ -75,8 +75,9 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.scala-tools</groupId>
-        <artifactId>maven-scala-plugin</artifactId>
+        <groupId>net.alchim31.maven</groupId>
+        <artifactId>scala-maven-plugin</artifactId>
+        <version>${scala-maven-plugin.version}</version>
```
Fokko, line 80: If you bump the parent pom version: https://github.com/apache/parquet-mr/blob/master/pom.xml#L7. Then this line shouldn't be required, because we then inherit the version from the parent pom.

martin-g: There is no dependencyManagement for this plugin in the parquet-mr/pom.xml. 
Also there is nothing about scala-maven-plugin at https://search.maven.org/artifact/org.apache/apache/23/pom so what you suggest won't work.

I will:
1) update the version of org.apache:apache from 16 to 23
2) add dependencyManagement in parquet-mr/pom.xml

martin-g: Done!
But at the moment some Maven plugins' versions in parquet-mr/pom.xml are newer than the ones in https://search.maven.org/artifact/org.apache/apache/23/pom, so I didn't apply your suggestion because this will actually downgrade them.

martin-g: The local build failed because of https://github.com/apache/parquet-mr/blob/5d8fe214f16829fe258400d0c68ddaacc979b03a/pom.xml#L241 and https://github.com/apache/parquet-mr/blob/5d8fe214f16829fe258400d0c68ddaacc979b03a/parquet-cli/pom.xml#L83. 
How is this supposed to work ?

martin-g: CI failed too: https://github.com/apache/parquet-mr/runs/2388424766?check_suite_focus=true

akauppi:
Looking at Parquet, with Scala 2.13 in mind.

PR #872, opened 2021-02-18
PARQUET-1975 Disable BrotliCodec for non-x86_64

martin-g:
BrotliCodec fails to load on non-x86_64 CPU architectures because MeteoGroup/jbrotli does not provide native binaries
With this change brotli-codec is added as a dependency to parquet-hadoop only for amd64, i.e. x86_64

### DIFF
--- a/parquet-hadoop/pom.xml
+++ b/parquet-hadoop/pom.xml
@@ -98,12 +98,6 @@
       <version>${fastutil.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.github.rdblue</groupId>
-      <artifactId>brotli-codec</artifactId>
-      <version>${brotli-codec.version}</version>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
       <groupId>com.github.luben</groupId>
       <artifactId>zstd-jni</artifactId>
       <version>${zstd-jni.version}</version>
@@ -151,5 +145,24 @@
     </plugins>
   </build>
 
+  <profiles>
+    <!-- PARQUET-1975 Do not add brotli-codec for non-x86_64 architectures -->
+    <profile>
+      <id>x86_64</id>
+      <activation>
+        <os>
+          <arch>amd64</arch>
+        </os>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>com.github.rdblue</groupId>
+          <artifactId>brotli-codec</artifactId>
+          <version>${brotli-codec.version}</version>
+          <optional>true</optional>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
 
 </project>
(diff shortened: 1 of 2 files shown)


gszadovszky (changes requested):
Based on a quick search the system property `os.arch` does not seem to be well-defined. Is there a way to disable brotli testing only in case of arm64 instead of trying to enable it only for x86_64? (I guess it does not hurt to download a dependency if it won't be used so we only have to deal with the test code.)

martin-g:
The problem is org.apache.hadoop.io.compress.CompressionCodecFactory#CODEC_PROVIDERS (hadoop-common:2.10.1) - https://github.com/apache/hadoop/blob/branch-2.10.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/CompressionCodecFactory.java#L42. 
It uses ServiceLoader API to load all implementations of org.apache.hadoop.io.compress.CompressionCodec and brotli-codec fails due to the missing binary.
So, having the dependency on the classpath breaks the **loading** of CompressionCodecFactory class and even code that does not use Brotli but any of the other compressions fails immediately.

gszadovszky:
> The problem is org.apache.hadoop.io.compress.CompressionCodecFactory#CODEC_PROVIDERS (hadoop-common:2.10.1) - https://github.com/apache/hadoop/blob/branch-2.10.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/CompressionCodecFactory.java#L42.
> It uses ServiceLoader API to load all implementations of org.apache.hadoop.io.compress.CompressionCodec and brotli-codec fails due to the missing binary.
> So, having the dependency on the classpath breaks the **loading** of CompressionCodecFactory class and even code that does not use Brotli but any of the other compressions fails immediately.

I see, thanks for the explanation. Then, you may try something like `<arch>!arm64</arch>` instead of using `amd64`. I think `amd64` is too strict and we may end up not testing something without noticing.

martin-g:
brotli-codec provides binaries only for `darwin-x86-amd64` and `linux-x86-amd64` - those are actual folder names in the jar.

I've updated this PR as you requested but if someone tries to build on PPC64, s390x or RISC-V then (s)he will face the same problem.

martin-g:
Should I modify .travis.yml in this PR or you prefer this to be in separate PR ?

gszadovszky (approved):
Thanks for the updates!

> I've updated this PR as you requested but if someone tries to build on PPC64, s390x or RISC-V then (s)he will face the same problem.

I prefer to be on the safe side and exclude tests only if required. I don't know if maven allows to have an `OR` statement or something similar in the profile activation but it is clearly a solvable issue.

> Should I modify .travis.yml in this PR or you prefer this to be in separate PR ?

You've already created the separate jira for this: PARQUET-1980. I would prefer to implement the travis related changes under that jira in a separate PR.

gszadovszky (approved):
Sorry, I've forgotten about this PR. It looks good to me.

PR #876, opened 2021-02-26
PARQUET-1980 Re-introduce TravisCI for testing on ARM64

martin-g:


martin-g:
The build on ARM64 will fail until https://github.com/apache/parquet-mr/pull/872 is merged.

gszadovszky (changes requested):
Travis fails because the .travis.yml file you have restored references scripts that were renamed. See #862 for details.

Also, please add some comments to the travis file about why we need this (and why github actions are not enough).

martin-g:
> Also, please add some comments to the travis file about why we need this (and why github actions are not enough).

Done!

gszadovszky (approved):

PR #878, opened 2021-03-05
PARQUET-1992: Manually download interop files inside the test and move encryption interop test to maven integration-test phase

andersonm-ibm:
This PR is another option of solving the problem "Cannot build from tarball because of git submodules".
1. Encryption interop tests run separately from unit tests - in integration-test phase 
2. Files for interop tests are downloaded manually, so git submodules are totally removed from the project

- `mvn package` - doesn't run interop tests
- `mvn verify/install` - run interop test and download files for interop from GitHub directly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ target/
 .cache
 *~
 mvn_install.log
-parquet-hadoop/parquet-testing
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ target/
 .cache
 *~
 mvn_install.log
+parquet-hadoop/parquet-testing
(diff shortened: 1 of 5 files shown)


gszadovszky (changes requested):
Thanks a lot for working on this. Sorry about the confusion that I did accept the idea of using git submodules before. Haven't realized it would cause issues with the tarball.

You've mentioned in the jira and in this PR as well that integration tests are executed in the phase `verify` and are not in the phases `package/install`. Based on the maven [lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference) `install` does depend on `verify` so the integration tests will be executed. It is not a problem and neither is that you have moved this test from unit tests to IT. Just wanted to mention.

gszadovszky (changes requested):

gszadovszky (approved):

PR #881, opened 2021-03-16
PARQUET-1999: NPE might occur if OutputFile is implemented by the client

gszadovszky:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -281,8 +281,9 @@ public class ParquetWriter<T> implements Closeable {
 
     // encryptionProperties could be built from the implementation of EncryptionPropertiesFactory when it is attached.
     if (encryptionProperties == null) {
+      String path = file == null ? null : file.getPath();
       encryptionProperties = ParquetOutputFormat.createEncryptionProperties(conf,
-          file == null ? null : new Path(file.getPath()), writeContext);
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -281,8 +281,9 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
 
     // encryptionProperties could be built from the implementation of EncryptionPropertiesFactory when it is attached.
     if (encryptionProperties == null) {
+      String path = file == null ? null : file.getPath();
       encryptionProperties = ParquetOutputFormat.createEncryptionProperties(conf,
-          file == null ? null : new Path(file.getPath()), writeContext);
+          path == null ? null : new Path(path), writeContext);
```
shangxinli, line 286: Instead of checking 'null' twice, can we do it like this? 
Path path = (file == null) ? null : new Path(file.getPath());

gszadovszky: We are not checking the same twice. Once, we are checking the argument `file` if it is `null` then we are checking if the value returned by `file.getPath()` is `null`. (The latter check was missing and that caused the issue.)

shangxinli: Got it. Thanks for the clarification.

shangxinli:
LGTM now

PR #890, opened 2021-04-12
PARQUET-2023 Do not pipe the mvn output to pv

martin-g:
Piping to pv looses the exit status of mvn and the builds never fail

Run mvn in quiet mode to prevent Travis error because of too much output

### Jira

https://issues.apache.org/jira/browse/PARQUET-2023

### Tests

No modifications to the source code. Only to the CI config.

### Commits

- [ X ] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":

### Documentation

No new functionality!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,10 @@ install:
   - sudo apt-get install -qq maven
   - java -version
   - mvn -version
-  - mvn install --quiet --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true
+  - travis_wait 60 mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true -Dorg.slf4j.simpleLogger.logFile=mvn-install.log
   
-script: mvn verify --quiet --batch-mode javadoc:javadoc -P ci-test,$HADOOP_PROFILE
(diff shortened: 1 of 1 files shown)


gszadovszky (approved):
The idea behind the original construct was to avoid "too much output" error of Travis while keeping all the logs in case of a failure. `pv` was required to have some output periodically otherwise Travis would think the process is stuck and kill it. 
Meanwhile the issue caused by `pv` is serious; thanks a lot for fixing it.

I am a bit skeptic about the `--quiet` argument because we might not have enough logs in case of a failure. Meanwhile I do not have a better idea and we might think about it more deeply if an issue occurs.

Fokko (approved):
I prefer the approach as well, let's see how the `--quite` works in practice.

martin-g:
I have (maybe?!) a better solution:

`mvn clean ... -Dorg.slf4j.simpleLogger.logFile=simple.log`

This will write all the output to ./simple.log. In combination with `travis_wait` and `after_failure` we could achieve what you wanted to do with `pv`. What do you think ?

gszadovszky:
@martin-g, it sounds great. If you want to invest more in this I am happy to review it.

martin-g:
> If you want to invest more in this I am happy to review it.

Good! I will update the PR!

gszadovszky (approved):
It looks good to me. Unfortunately, such changes can be verified properly only in production.
I'll push this tomorrow if there are no objections.
Thanks a lot for your efforts!

martin-g:
No problem!
Just ping me if you think I could be in help with something else!

PR #900, opened 2021-05-03
PARQUET-2042: Add support for unwrapping common Protobuf wrappers and…

mwong38:
… logical Timestamps, Date, TimeOfDay

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-protobuf/pom.xml
+++ b/parquet-protobuf/pom.xml
@@ -32,6 +32,7 @@
   <properties>
     <elephant-bird.version>4.4</elephant-bird.version>
     <protobuf.version>3.25.1</protobuf.version>
+    <common-protos.version>2.28.0</common-protos.version>  <!-- make sure it's compatible with protobuf.version -->
     <truth-proto-extension.version>1.1.5</truth-proto-extension.version>
   </properties>
 
@@ -66,6 +67,16 @@
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
       <version>${protobuf.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java-util</artifactId>
+      <version>${protobuf.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-common-protos</artifactId>
+      <version>${common-protos.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.parquet</groupId>
@@ -191,6 +202,7 @@
             </goals>
             <configuration>
               <protocArtifact>com.google.protobuf:protoc:${protobuf.version}</protocArtifact>
+              <includeMavenTypes>direct</includeMavenTypes>
               <addSources>test</addSources>
               <addProtoSources>all</addProtoSources>
               <includeMavenTypes>direct</includeMavenTypes>
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
@@ -18,10 +18,21 @@
  */
 package org.apache.parquet.proto;
 
+import com.google.protobuf.BoolValue;
 import com.google.protobuf.ByteString;
+import com.google.protobuf.BytesValue;
 import com.google.protobuf.DescriptorProtos;
 import com.google.protobuf.Descriptors;
+import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.DoubleValue;
+import com.google.protobuf.FloatValue;
+import com.google.protobuf.Int32Value;
+import com.google.protobuf.Int64Value;
 import com.google.protobuf.Message;
+import com.google.protobuf.StringValue;
+import com.google.protobuf.UInt32Value;
+import com.google.protobuf.UInt64Value;
+import com.google.protobuf.util.Timestamps;
 import com.twitter.elephantbird.util.Protobufs;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.column.Dictionary;
@@ -42,6 +53,8 @@ import org.apache.parquet.schema.PrimitiveType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.LocalDate;
+import java.time.LocalTime;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -233,6 +246,21 @@ class ProtoMessageConverter extends GroupConverter {
       public Optional<Converter> visit(LogicalTypeAnnotation.MapLogicalTypeAnnotation mapLogicalType) {
         return of(new MapConverter(parentBuilder, fieldDescriptor, parquetType));
       }
+
+      @Override
+      public Optional<Converter> visit(LogicalTypeAnnotation.TimestampLogicalTypeAnnotation timestampLogicalType) {
+        return of(new ProtoTimestampConverter(parent, timestampLogicalType));
+      }
+
+      @Override
+      public Optional<Converter> visit(LogicalTypeAnnotation.DateLogicalTypeAnnotation dateLogicalType) {
+        return of(new ProtoDateConverter(parent));
+      }
+
+      @Override
+      public Optional<Converter> visit(LogicalTypeAnnotation.TimeLogicalTypeAnnotation timeLogicalType) {
+        return of(new ProtoTimeConverter(parent, timeLogicalType));
+      }
     }).orElseGet(() -> newScalarConverter(parent, parentBuilder, fieldDescriptor, parquetType));
   }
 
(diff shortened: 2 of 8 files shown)


shangxinli:
@mwong38, can you put more information in the Jira on why/what is changed? This is pretty big change and it would help people to review your code.

sheinbergon:
@shangxinli @mwong38  any way I can help with merge? Currently, proto-to-parquet conversions do not support proper timestamp handling, and it looks like this work addresses the issue.

Let me know.

mwong38:
After proto3 made everything optional, there is no way to know whether a primitive has been set or not. That is, you could no longer represent a "nullable" primitive. (They later brought `optional` keyword back, but the damage is done). The solution was to wrap a primitive in a message. For example, to represent a nullable `double`, there is the [DoubleValue](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#doublevalue). It's analogous to Java's boxed `Double`. 

Parquet supports nullable primitives natively. If we were to represent these well-known Protobuf types directly, it will be nested inside a deeper data structure; very inconvenient and a waste of space. What I did here is "unwrap" these primitives and make use of Parquet's nullaibility.

Additionally, I convert other well-known types such as Timestamp and Date to Parquet's. Again, rather than represent the data structure in its raw nested form (seconds/nanos or year, month, day, etc), they are converted to Parquet's native or logical representation of Timestamp and Date. You can turn on or off this features in the `ProtoSchemaConverter` constructor.

sheinbergon:
@shangxinli is there any reason this work shouldn't be merged, given @mwong38 explanation?

shangxinli:
I don't see a reason why this cannot be merge. I will have a look soon.

shangxinli:
Can you squash all the commits?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -427,6 +485,218 @@ public void addBinary(Binary binary) {
...
+    final ParentValueContainer parent;
+
+    public ProtoDateConverter(ParentValueContainer parent) {
+      this.parent = parent;
+    }
+
+    @Override
+    public void addInt(int value) {
+      LocalDate localDate = LocalDate.ofEpochDay(value);
+      com.google.type.Date date = com.google.type.Date.newBuilder()
```
shangxinli, line 525: Can we use import?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
```diff
@@ -427,6 +485,218 @@ public void addBinary(Binary binary) {
...
+
+    final ParentValueContainer parent;
+
+    public ProtoUInt32ValueConverter(ParentValueContainer parent) {
+      this.parent = parent;
+    }
+
+    @Override
+    public void addLong(long value) {
+      parent.add(UInt32Value.of((int) value));
```
shangxinli, line 650: Can we do a safety check when converting from int to long?

sheinbergon:
@mwong38 let me know if you want me to help in any way

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
mwong38: I think in this particular case it's more clear to be explicit that this is the Google Protobuf Date, rather than, say, the Java built-in Date.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
mwong38: Good idea

sheinbergon:
@shangxinli any news about merging this version? Are there still any blockers?

shangxinli:
Will have another look soon.

parquet-protobuf/pom.xml
```diff
@@ -57,6 +58,16 @@
...
       <version>${protobuf.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java-util</artifactId>
+      <version>${protobuf.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-common-protos</artifactId>
```
emkornfield, line 68: why this dependency, i thought well known types shipped with the proto package?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -97,6 +127,46 @@ public MessageType convert(Class<? extends Message> protobufClass) {
 
   private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addField(FieldDescriptor descriptor, final GroupBuilder<T> builder) {
     if (descriptor.getJavaType() == JavaType.MESSAGE) {
+      if (unwrapProtoWrappers) {
+        String typeName = descriptor.getMessageType().getFullName();
+        if (typeName.equals(PROTOBUF_TIMESTAMP_TYPE)) {
+          return builder.primitive(INT64, getRepetition(descriptor)).as(timestampType(true, TimeUnit.NANOS));
```
emkornfield, line 133: should TimeUnit be something configurable?  I know protobuf timestamps support nanosecond resolution but might not be what applications want to store.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -35,22 +35,49 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(ProtoSchemaConverter.class);
   private final boolean parquetSpecsCompliant;
+  private final boolean unwrapProtoWrappers;
 
   public ProtoSchemaConverter() {
-    this(false);
+    this(false, false);
+  }
+
+  public ProtoSchemaConverter(boolean parquetSpecsCompliant) {
```
emkornfield, line 79: please add docs for all new publicly facingin parameters.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -97,6 +127,46 @@ public MessageType convert(Class<? extends Message> protobufClass) {
 
   private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addField(FieldDescriptor descriptor, final GroupBuilder<T> builder) {
     if (descriptor.getJavaType() == JavaType.MESSAGE) {
+      if (unwrapProtoWrappers) {
+        String typeName = descriptor.getMessageType().getFullName();
```
emkornfield, line 131: is there a reason you are using type equality vs descriptor equality?  (I haven't checked but I would think .Equals would work?  Maybe add a comment if this is intentional.

shangxinli:
@mwong38 Can you address the feedback from @emkornfield before we can merge?

sheinbergon:
@mwong38 anyway I can help with finalizing this PR?

parquet-protobuf/pom.xml
mwong38: If you want to pre-packaged common Proto classes, you need to include this.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
mwong38: I don't think it's worth complicating the API. The Timestamp common Proto stores time in nanoseconds. There's no good reason to deviate from that or to truncate the resolution. If the user wishes to do more manipulation, it can be done downstream.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
mwong38: That's a good point. I'll change it to compare the `Descriptor` directly to the Wrapper's Descriptor rather than the name.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
shangxinli: In that case, you can use the default 'TimeUnit.NANOS' while having that configured.

mwong38: Isn't that outside the scope of ProtoSchemaConverter/ParquetWriter? I don't think it should go into the business of doing transformations. If the point of `ProtoSchemaConverter` is to give the closest representation of the Protobuf object in Parquet, then it should be NANOS and nothing more.

shangxinli: @emkornfield  Do you still have a comment on this?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
shangxinli: If there is no collision on the imports, we should use imports generally.

shangxinli:
I think we are close to merge this PR. Resolve the conflict and use the imports , then we can merge.

mwong38:
> I think we are close to merge this PR. Resolve the conflict and use the imports , then we can merge.

Done

zhaolihe:
any reason why this PR hasn't been merged yet?

mwong38:
> I just noticed this PR and sorry to see it does not check in. @mwong38 Could you try rebase it one last time? Thanks!

Done. I really hope it's the last time.

shangxinli (approved):

shangxinli:
Merged

PR #901, opened 2021-05-04
PARQUET-2037: Write INT96 with parquet-avro

gszadovszky:

### Jira

### Tests

### Commits

### Documentation


shangxinli:
In the ticket, you mentioned there are two ways to solve this issue. I see you implemented it using the 2nd way. I wonder what is the reason behind it?  I am not in favor of one over the other, just want to know what are the pros and cons for each.

gszadovszky:
@shangxinli,
The other option was about using the `doc` field of the avro schema field. It would be a misuse of it as anyone can set anything to that field. It is more about documenting the related field for humans instead of taking decisions based on it in the code. On the Parquet sync that was my only idea to resolve this issue and even I did not like it.
I think the 2nd option (covered by this PR) is much better but only realized its possibility after the Parquet sync meeting so we could not discuss it personally.

shangxinli:
Yeah, agree. Just one thing that sometimes it might not be straightforward for the user to know the exact path to manually set in the configuration for some deeply nested schema.  I remember last time when I worked on Avro schema, there are 20+ layers nested in the field and there are 'Type' in the middle with 'name' in it. That is not very human being readable and easy to make mistake. But I have less experience working on Schema, I am not certain this is a real issue.

gszadovszky:
@shangxinli, I agree this is not a perfect solution but I could not come up with any better one. Meanwhile, this feature will not be used widely since INT96 is deprecated. Maybe, it is even better that this feature is not always easy to use :)

PR #902, opened 2021-05-04
PARQUET-1633: Fix integer overflow

eadwright:
This PR addresses this issue: https://issues.apache.org/jira/browse/PARQUET-1633

I have not added unit tests, as to check overflow conditions I would need test data over 2GB in size (on disk, compressed), considerably larger in-memory and thus requiring significant CI resources.

The issue was using an `int` for length field, which for parquet files with very large `row_group_size` (row groups over 2GB) would cause silent integer overflow, manifesting itself as negative length and an attempt to create an ArrayList with negative length.

I have tested reading a 6.6GB parquet file with a huge row group size (works out at over 2GB) which recreated this problem, and with this modification can read the file without any issues.


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1464,7 +1464,7 @@ protected PageHeader readPageHeader(BlockCipher.Decryptor blockDecryptor, byte[]
      */
     private void verifyCrc(int referenceCrc, byte[] bytes, String exceptionMsg) {
       crc.reset();
-      crc.update(bytes);
+      crc.update(bytes, 0, bytes.length);
```
eadwright, line 1467: Changed to adopt a Java 8 API, to be consistent with the pom

shangxinli:
Can you come up with a unit test?

eadwright:
> Can you come up with a unit test?

I want to, and I'm open to ideas. To re-create the problem you need a parquet file which has a row group over 2GB in size, which translates to way over 2GB of memory used. Perhaps 10x more - I've not investigated which field types may work the best. My concern is whether such a test would blow the constraints during CI or put an unreasonable memory burden on anyone wanting to build/test the project. What are reasonable limits there?

eadwright:
Interesting to note ParquetWriter takes an int for the block size, so probably can't create a file which recreates this bug. I can easily recreate this bug by creating a large file with an insane block size in python.

eadwright:
I tried adapting `TestParquetReaderRandomAccess` to create a file with a 2GB block size and 135M records, each with 2 longs, which should spill over into two blocks and recreate the issue. Alas this fails to run even if I give it 12GB of heap. One needs a lot of RAM to reproduce this issue.

gszadovszky:
@eadwright,

I'll try to summarize the issue, please correct me if I'm wrong. Parquet-mr is not able to write such big row groups (>2GB) because of the `int` array size limitation. Meanwhile, both the format and some other implementations allow such big row groups. So, parquet-mr shall be prepared for this issue in some way.
One option is to "simply" read the large row groups. It would require significant efforts to use proper memory handling objects that would properly support reading the large row groups. (A similar effort would also make parquet-mr available to write larger row groups than 2GB.)

The other option is to handle the too large row groups with a proper error message in parquet-mr without allowing silent overflows. This second option would be this effort. It is great to handle to potential int overflows but the main point, I think, would be at the footer conversion (`ParquetMetadataConverter`) where we create our own object structure from the file footer. At this point we can throw the proper error messages if the row group is too large to be handled (for now) in parquet-mr. 
BTW, it might not be enough to check the potential overflows to validate if we can read a row group size. (See e.g. the source code of [ArrayList](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l229).)

About the lack of the unit tests. I can accept in some cases where unit tests are not practically feasible to be implemented. In these cases I usually ask to validate the code offline.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
advancedxy: This is unrelated, I would prefer to update this in another PR.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1763,8 +1763,8 @@ public void addChunk(ChunkDescriptor descriptor) {
     public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOException {
       f.seek(offset);
 
-      int fullAllocations = length / options.getMaxAllocationSize();
-      int lastAllocationSize = length % options.getMaxAllocationSize();
+      int fullAllocations = (int)(length / options.getMaxAllocationSize());
```
advancedxy, line 1766: `(int)(length / options.getMaxAllocationSize())` -> ` Math.toIntExact(length / options.getMaxAllocationSize());`

advancedxy (changes requested):
I met a similar issue: https://issues.apache.org/jira/browse/PARQUET-2045, probably the same one.

I would like this fix got merged.

advancedxy:
> @eadwright,
> 
> I'll try to summarize the issue, please correct me if I'm wrong. Parquet-mr is not able to write such big row groups (>2GB) because of the `int` array size limitation. Meanwhile, both the format and some other implementations allow such big row groups. So, parquet-mr shall be prepared for this issue in some way.
> One option is to "simply" read the large row groups. It would require significant efforts to use proper memory handling objects that would properly support reading the large row groups. (A similar effort would also make parquet-mr available to write larger row groups than 2GB.)
> 
> The other option is to handle the too large row groups with a proper error message in parquet-mr without allowing silent overflows. This second option would be this effort. It is great to handle to potential int overflows but the main point, I think, would be at the footer conversion (`ParquetMetadataConverter`) where we create our own object structure from the file footer. At this point we can throw the proper error messages if the row group is too large to be handled (for now) in parquet-mr.
> BTW, it might not be enough to check the potential overflows to validate if we can read a row group size. (See e.g. the source code of [ArrayList](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l229).)
> 
> About the lack of the unit tests. I can accept in some cases where unit tests are not practically feasible to be implemented. In these cases I usually ask to validate the code offline.

Hi @gszadovszky parquet-mr is able to produce big row groups. We found some files wrote by Spark(which uses parquet-mr) have this problem. See https://issues.apache.org/jira/browse/PARQUET-2045 for details.

There are two options to fix this problem:
1. fail at writer side when creating such large row group/column chunk
2. support at reader side, which is this approach. It would require a lot of resource, but it's feasible.

Either option is fine for me, WDYT?

gszadovszky:
@advancedxy, thanks for explaining.
I think, the best option is 2. It is up to the user to provide enough resources for handling the large row groups or not writing them.
Meanwhile, even though I've written I can accept lacking of unit tests in some situations my concern in this case is I am not sure that every aspect of a large row group is handled properly. So, we clearly need to validate this fix with such large row groups. This test can be even implemented in this source code but we must not include it in the unit tests or integration tests we run regularly.

advancedxy:
> @advancedxy, thanks for explaining.
> I think, the best option is 2. It is up to the user to provide enough resources for handling the large row groups or not writing them.
> Meanwhile, even though I've written I can accept lacking of unit tests in some situations my concern in this case is I am not sure that every aspect of a large row group is handled properly. So, we clearly need to validate this fix with such large row groups. This test can be even implemented in this source code but we must not include it in the unit tests or integration tests we run regularly.


@gszadovszky well understood.

Hi @eadwright, is it possible for you to add the test case(tagged with @Ignore) in your PR, so others like @gszadovszky or me can verify it offline?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
eadwright: Agreed, committed to revert this change

eadwright:
Not a unit test yet, but I have at least found a reliable way to create a parquet file in python which causes Java to break when reading it. For this we require a huge number of rows, and a string column with very high entropy to avoid the use dictionary lookup or compression. Not these conditions can be real-world - records containing long SHA3 hashes for example.

This python code needs more than 16GB of RAM to execute.

```import pandas as pd
import numpy as np

rand_array = np.random.rand(75000000, 4)

df = pd.DataFrame(rand_array, columns=['number1', 'number2', 'number3', 'number4'])
df['string1'] = df['number1'].astype(str) + df['number2'].astype(str) + df['number3'].astype(str)

df.to_parquet("random.parquet", compression='snappy', engine='pyarrow')```

**Comment by eadwright** [comment]
Screenshot from `parquet-tools meta` looking at the generated file. Note the TS (total size) of row group 1 is way over 2GB

image

**Comment by eadwright** [comment]
> @eadwright, what do you mean by "necessary for the rows to spill over into a second row group."? It shall not be possible. Even the pages keep row boundaries but for row groups it is required by the specification.

Sorry I probably didn't phrase that well. I mean, for this bug to occur, you need i) a row group which is taking more than 2GB of space, to get the 2^31 signed-int overflow, and ii) another subsequent row group (any size) so the code with the bug adds to a file offset using a corrupted value.

In the example python code, if I asked it to produce 50M rows instead of 75M, you get a ~3.3GB row group, but no second row group. The file offset addition code path is not executed and the file gets read correctly, the bug is not triggered.

**Comment by gszadovszky** [comment]
Thanks, @eadwright for explaining. I get it now.

**Comment by eadwright** [comment]
Build failed due to some transient connectivity issue - builds fine for my, Java 8 and 11

**Comment by eadwright** [comment]
Correction, we need a column within a row group to be over 2GB in size to cause the issue.

**Comment by eadwright** [comment]
I've tweaked the python to create a test file which Java can't read. The python can now run fine on a 16GB machine.

```import pandas as pd
import numpy as np
rand_array = np.random.rand(48000000, 4)
df = pd.DataFrame(rand_array, columns=['number1', 'number2', 'number3', 'number4'])
df['string1'] = df['number1'].astype(str) + df['number2'].astype(str) + df['number3'].astype(str)
df.drop(["number1", "number2", "number3", "number4"], axis=1, inplace=True)
df.to_parquet("random.parquet", compression="snappy", engine="pyarrow", **{"row_group_size": 47800000})```

It creates 48M records, 47.8M of which are in the first row group, and the data for the `string1` column in the first row group is about 2.1GB in size, over the threshold to cause the Java bug.

**Comment by advancedxy** [comment]
@eadwright can you upload the corrupt parquet file to some public cloud service like google drive or s3? or is it possible for you to add a new test case(not run by default) in `TestParquetReaderRandomAccess`?

@gszadovszky do you have any other concerns about this pr?

**Comment by eadwright** [comment]
I've uploaded the python-generated file to my Google Drive: https://drive.google.com/file/d/1trWjeJaHpqbHlnipnDUrwlwYM-AKFNWg/view?usp=sharing

Still working on a unit test. I'm used to work with Avro schema, but I'm trying to cut it back to bare-bones.

**Comment by eadwright** [comment]
Work in progress looks like this (not committed) - adjust `PATH` to point to the file I uploaded.

```

import org.apache.parquet.ParquetReadOptions;
import org.apache.parquet.column.page.PageReadStore;
import org.apache.parquet.example.data.Group;
import org.apache.parquet.example.data.simple.convert.GroupRecordConverter;
import org.apache.parquet.hadoop.util.HadoopInputFile;
import org.apache.parquet.io.ColumnIOFactory;
import org.apache.parquet.io.MessageColumnIO;
import org.apache.parquet.io.RecordReader;
import org.apache.parquet.schema.PrimitiveType;
import org.junit.Test;

import java.io.IOException;

import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
import static org.apache.parquet.schema.Type.Repetition.REQUIRED;

public class TestParquetReaderLargeColumn {
  private static final Path PATH = new Path("/Volumes/HDD/random6.parquet");

  @Test
  public void test() throws IOException {
    Configuration configuration = new Configuration();
    ParquetReadOptions options = ParquetReadOptions.builder().build();

    MessageType messageType = buildSchema();

    try (ParquetFileReader reader = new ParquetFileReader(HadoopInputFile.fromPath(PATH, configuration), options)) {
      PageReadStore pages;
      while ((pages = reader.readNextRowGroup()) != null) {
        MessageColumnIO columnIO = new ColumnIOFactory().getColumnIO(messageType);
        RecordReader<Group> recordReader = columnIO.getRecordReader(pages, new GroupRecordConverter(messageType));
        long rowCount = pages.getRowCount();

        for (int i = 0; i < rowCount - 1; i++) {
          Group group = recordReader.read();
          group.getString("string", 0);
        }
      }
    }
  }

  private static MessageType buildSchema() {
    return new MessageType("AvroString",
      new PrimitiveType(REQUIRED, BINARY, "string"));
  }
}
```

Alas when I run it, I get this Exception, not sure why yet:

```
java.lang.NoSuchMethodError: org.apache.parquet.format.LogicalType.getSetField()Lshaded/parquet/org/apache/thrift/TFieldIdEnum;

	at org.apache.parquet.format.converter.ParquetMetadataConverter.getLogicalTypeAnnotation(ParquetMetadataConverter.java:1066)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.buildChildren(ParquetMetadataConverter.java:1569)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.fromParquetSchema(ParquetMetadataConverter.java:1524)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.fromParquetMetadata(ParquetMetadataConverter.java:1399)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.readParquetMetadata(ParquetMetadataConverter.java:1370)
	at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:583)
	at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:777)
	at org.apache.parquet.hadoop.TestParquetReaderLargeColumn.test(TestParquetReaderLargeColumn.java:64)
```

eadwright:
Update - I've created some Java code which writes a parquet file nearly big enough to cause the issue, and can successfully read this file. However, two problems:

* If I bump the record count so I have a file big enough to reproduce this bug 1633, another bug causes 32-bit integer overflow. The code as it stands cannot reproduce the file which was created in python.
* The code I'm using to read the data, the potential unit test, does not work against the python-produced file (which has an avro schema) as I get this error (I suspect some classpath issue) `java.lang.NoSuchMethodError: org.apache.parquet.format.LogicalType.getSetField()Lshaded/parquet/org/apache/thrift/TFieldIdEnum;`

So... if any of you can reproduce the original error with the parquet file I posted above, and can validate this 7 line fix addresses it, that'd be great. Open to ideas of course.

gszadovszky:
@eadwright, I'll try to look into this this week and produce a java code to reproduce the issue.

eadwright:
Awesome, appreciated. Also note the file uploaded isn't corrupt as such, it just goes beyond 32-bit limits.

gszadovszky:
@eadwright, I've implemented a [unit test](https://github.com/gszadovszky/parquet-mr/commit/fcaf41269470c03c088b7eb5598558d44013f59d) to reproduce the issue and test your solution. Feel free to use it in your PR. I've left some TODOs for you :)

gszadovszky:
@eadwright, so the CI was [executed](https://github.com/gszadovszky/parquet-mr/actions/runs/879304911) somehow on my private repo and failed due to OoM. So, we may either investigate if we can tweak our configs/CI or disable this test by default.

eadwright:
@gszadovszky I raised a PR to bring your changes in to my fork. Not had time yet alas to address the TODOs. I can say though that I believe to read that example file correctly with the fix requires 10GB of heap or so, probably similar with your test. Agree this test should be disabled by default, too heavy for CI.

gszadovszky:
@eadwright, I've made some changes in the unit test (no more TODOs). See the update [here](https://github.com/gszadovszky/parquet-mr/commit/6dc3f418b537fd5cb7954018243399f39784d81b). The idea is to not skip the tests in "normal" case but catch the OoM and skip. This way no tests should fail on any environments. Most of the modern laptops should have enough memory so this test will be executed on them.

eadwright:
@gszadovszky I had a look at your changes. I feel uncomfortable relying on any behaviour at all after an OOM error. Are you sure this is the right approach?

eadwright:
@gszadovszky Sorry, delay on my part. Have merged your changes in. Even in a testing pipeline I am uncomfortable with catching any kind of `java.lang.Error`, especially an OOM. Should we remove those catch clauses and mark this test to be ignored usually - let it be run manually?

gszadovszky:
@eadwright, I understand your concerns I don't really like it either. Meanwhile, I don't feel good having a test that is not executed automatically. Without regular executions there is no guarantee that this test would be executed ever again and even if someone would execute it it might fail because of the lack of maintenance.

What do you think about the following options? @shangxinli, I'm also curious about your ideas.
* Execute this test separately with a maven profile. I am not sure if the CI allows allocating such large memory but with Xmx options we might give a try and create a separate check for this test only.
* Similar to the previous with the profile but not executing in the CI ever. Instead, we add some comments to the release doc so this test will be executed at least once per release.
* Configuring the CI profile to skip this test but have it in the normal scenario meaning the devs will execute it locally. There are a couple of cons though. There is no guarantee that devs executes all the tests including this one. It also can cause issues if the dev doesn't have enough memory and don't know that the test failure is not related to the current change.

eadwright:
Interesting options @gszadovszky , I have no strong opinion. I'd just like this fix merged once everyone is happy :)

gszadovszky:
@eadwright, sorry, you're right. This is not tightly related to your PR. Please, remove the try-catch blocks for OOE and put an `@Ignore` annotation to the test class for now. I'll open a separate jira to solve the issue of such "not-to-be-tested-by-ci" tests.

eadwright:
@gszadovszky done

gszadovszky (approved):
Thanks a lot for your efforts, @eadwright.

eadwright:
Pleasure @gszadovszky

eadwright:
Note I don't have write access, so I guess someone needs to merge this at some point? Not sure of your workflow.

gszadovszky:
Sure, @eadwright, I'm merging it. I just usually give another day after the approval so others have the chance to comment.

gszadovszky:
@eadwright, do you have a jira account so I can assign this one to you?

eadwright:
@gszadovszky Thanks - just created Jira account `edw_vtxa`.

PR #910, opened 2021-05-21
PARQUET-2052: Integer overflow when writing huge binary using dictionary encoding

sunchao:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
@@ -173,7 +173,7 @@ public abstract class DictionaryValuesWriter extends ValuesWriter implements Req
       BytesInput bytes = concat(BytesInput.from(bytesHeader), rleEncodedBytes);
       // remember size of dictionary when we last wrote a page
       lastUsedDictionarySize = getDictionarySize();
-      lastUsedDictionaryByteSize = (int) dictionaryByteSize;
+      lastUsedDictionaryByteSize = Math.toIntExact(dictionaryByteSize);
       return bytes;
     } catch (IOException e) {
       throw new ParquetEncodingException("could not encode the values", e);
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
@@ -81,7 +81,7 @@ public abstract class DictionaryValuesWriter extends ValuesWriter implements Req
   protected boolean dictionaryTooBig;
 
   /* current size in bytes the dictionary will take once serialized */
-  protected int dictionaryByteSize;
+  protected long dictionaryByteSize;
 
   /* size in bytes of the dictionary at the end of last dictionary encoded page (in case the current page falls back to PLAIN) */
   protected int lastUsedDictionaryByteSize;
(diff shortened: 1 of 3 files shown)


sunchao:
@gszadovszky @shangxinli @ggershinsky could you take a look? thanks!

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
```diff
@@ -173,7 +173,7 @@ public BytesInput getBytes() {
       BytesInput bytes = concat(BytesInput.from(bytesHeader), rleEncodedBytes);
       // remember size of dictionary when we last wrote a page
       lastUsedDictionarySize = getDictionarySize();
-      lastUsedDictionaryByteSize = dictionaryByteSize;
+      lastUsedDictionaryByteSize = (int) dictionaryByteSize;
```
gszadovszky, line 176: This method should not be called in case of `shouldFallBack()` returns true be we would better be on the safe side and check the potential overflow.

gszadovszky (changes requested):
I have only one note about the downcast. Please check.

About the failures. This is indeed an incompatible change since it is done on a protected field in a public class. I can see two options to workaround this.
1. Keep the field as an `int` and check the overflow afterwards at any places where `dictionaryByteSize` is used. I don't really like this idea myself. It might be risky and definitely not clean but it would not trigger the compatibility checker.
2. Exclude the related field from compatibility checker. I don't have too much experience on this but there are examples in the [doc](https://siom79.github.io/japicmp/MavenPlugin.html) of the plugin. If you choose this solution the proper comments for the exclusion (in the pom.xml) would be required. It would also be nice to mention that after a minor release we shall remove that exclusion.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
sunchao: Okay. Changed to `Math.toIntExact` so it will throw exception when overflow.

sunchao:
@gszadovszky I took your second recommendation and disabled the compatibility check for the field. Also added a TODO there to remove it once next minor release is done. Could you take another look? thanks.

advancedxy:
@sunchao Hi, do you encounter this in production?

sunchao:
@advancedxy yes, one of our users wrote very large strings and ended up with corrupted Parquet files because of this issue.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
shangxinli: If that user continues writing that large string, will here throw an exception?  And it will block the writer, right?

shangxinli:
@gszadovszky Do you think we should fall back to no dictionary encoding in case of large byte size?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
sunchao: I don't think so. With the fix, it should've already fallback to PLAIN encoding during `writeBytes` before getting here.

gszadovszky (approved):
Thanks, @sunchao for your efforts!

@shangxinli, this case of large byte size is an overflow meaning the dictionary size would be larger than it could be represented as an `int`. Meanwhile in the format page header both compressed and uncompressed sizes are represented as `i32` values so the current dictionary would be out of its limits anyway.
It is another question if all the encoders used for writing data pages for binary values are prepared for similar situations. I would leave this question for @sunchao since this issue seems a rare situation (not aware of similar jiras while parquet-mr works like this since the beginning) .

sunchao:
Thanks @gszadovszky and @shangxinli for taking a look. 

> It is another question if all the encoders used for writing data pages for binary values are prepared for similar situations.

After fallback from dictionary encoding, I think it will fail later [during writing data page](https://github.com/apache/parquet-mr/blob/apache-parquet-1.12.0/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L163). I think this is still better than generating corrupted page.

gszadovszky:
I agree, @sunchao that failing with a proper exception is always better than committing a corrupt file. Any issues might happen after the fallback should be handled in a separate jira.

PR #918, opened 2021-07-09
PARQUET-2064: Make Range public accessible in RowRanges

shangxinli:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.parquet.internal.filter2.columnindex;
+package org.apache.parquet.filter2.columnindex;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -28,6 +28,8 @@ import java.util.Set;
 
 import org.apache.parquet.filter2.compat.FilterCompat.Filter;
 import org.apache.parquet.internal.column.columnindex.OffsetIndex;
+import org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter;
+import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
 
 /**
  * Class representing row ranges in a row-group. These row ranges are calculated as a result of the column index based
@@ -92,7 +94,7 @@ public class RowRanges {
     }
   }
 
-  static final RowRanges EMPTY = new RowRanges(Collections.emptyList());
+  public static final RowRanges EMPTY = new RowRanges(Collections.emptyList());
 
   private final List<Range> ranges;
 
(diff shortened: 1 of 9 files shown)


sunchao:
I also opened https://issues.apache.org/jira/browse/PARQUET-2061 for similar purpose.

shangxinli:
Thanks Chao for your comments. Can you make a PR for [PARQUET-2061](https://issues.apache.org/jira/browse/PARQUET-2061)? I am thinking we can add a new API to return the RowRanges directly and use this PR to make Range public accessible so that your API can this class? 

@gszadovszky Do you think it makes sense? If yes, I am going to fix this PR.

gszadovszky:
@shangxinli, unfortunately we still struggle with not having proper low level API for our clients. We did not add the ranges and such to the public API because they should not be required there. Since both Spark and Hive uses lower level APIs (that was not designed to be public originally) I don't think we have any other choice for now to make all the necessary classes/methods public.

Meanwhile, I've added `RowRanges` to the package `org.apache.parquet.internal.filter2.columnindex` (note `internal`) to make it clear that even though the class is public it is not for our clients. So, if we really want to make this public we also need to move it to another package.

sunchao:
@shangxinli sure I can make a PR after this one is done - it depends on making the `Range` class public.

gszadovszky (changes requested):
I think, we still have to move `RowRanges` from `org.apache.parquet.internal.filter2.columindex` to `org.apache.parquet.filter2.columnindex`. This move shall also be done to any class we want to make public for our clients.

shangxinli:
@gszadovszky I just made the change but the building error 'METHOD_RETURN_TYPE_CHANGED' was reported by 'japicmp'. I guess we need to change something to let 'japicmp' know the 'METHOD_RETURN_TYPE_CHANGED' is intentional.

gszadovszky:
@shangxinli, when I've said we should move the class `RowRanges` from the internal package I did not realize it would be a breaking change from java API point of view. japicmp is right and it is not easy to suppress such errors. We can exclude the whole rule or the related class from the check but it might lead to uncaught changes in the future.
The easiest and more future proof solution is to leave the class `RowRanges` in the internal package. (Sorry for the confusion.) It is not the best solution because it might be misleading to the API clients but I don't have a better one.

shangxinli:
Agree. Let's keep it internal for now.

gszadovszky (approved):
@shangxinli, do you know anything about the failing Travis builds? It is happening for a while now.

gszadovszky:
@shangxinli, I've created the jira INFRA-22171 about the Travis failures. Since the Travis test is for arm64 validation and most of the changes should be platform independent, we can ignore these failures and step forward with the PRs.

gszadovszky:
Re-triggering CI...

shangxinli:
Thank you so much @gszadovszky! Sorry didn't get time to check your ask yesterday. I see INFRA-22171 is moving forward. Let's see.

gszadovszky:
@shangxinli, so the infra ticket is solved now, everything seems to be working fine. The only failing check here is due to the fact that this PR was existed before the fix. We have another entry for Travis CI that is passing so we can ignore the failing one.

Feel free to push this.
(This fix seems to be required for a patch release so it shall also be backported to the parquet-1.12.x branch.)

PR #919, opened 2021-07-30
PARQUET-2070 replace deprecated syntax in ProtoWriteSupport.java

sv3ndk:
### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2070)

### Tests

- [ ] My PR does not need testing for this extremely good reason: I'm simply removing a private method, no feature has changed, existing unit are sufficient to validate the change.

### Commits

This minor PR:

* replaces usage of the deprecated syntax `TextFormat.printToString(asProto)` with the now recommended and simpler `.toString()` 
* the private method where this code was present was only invoked from one place, where the protobuf Descriptor was already available => I moved the logic directly there, which removes one reflection invocation and makes the code slightly more concise (and probably trivially faster as well)
* I also added a mention of protobuf in the readme, since Thrif and Avro were mentioned already

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,9 @@ Thrift integration is provided by the [parquet-thrift](https://github.com/apache
 ### Avro
 Avro conversion is implemented via the [parquet-avro](https://github.com/apache/parquet-mr/tree/master/parquet-avro) sub-project.
 
+### Protobuf
+Protobuf conversion is implemented via the [parquet-protobuf](https://github.com/apache/parquet-mr/tree/master/parquet-protobuf) sub-project.
+
 ### Create your own objects
 * The ParquetOutputFormat can be provided a WriteSupport to write your own objects to an event based RecordConsumer.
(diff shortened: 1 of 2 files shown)


sv3ndk:
The travis build does not seem to start, is there something I can do to trigger it?

gszadovszky (approved):

sv3ndk:
The unit test executed by Github seem OK, but the Travis build has failed with no logs after being queued for about 6h.
Would we be facing a timeout on Travis side due to lack of resource? 
What's a good way to address this?

gszadovszky:
@sv3ndk, we have similar failures in Travis for other PRs as well. (Created INFRA-22171 about this issue.) Travis test is there to cover arm64 architecture. I don't think this change would hurt arm64 compatibility so I'm pushing this one.

sv3ndk:
Makes sense, thanks for the merge @gszadovszky

PR #921, opened 2021-08-09
PARQUET-2064: Make Range public accessible in RowRanges

shangxinli:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
@@ -37,7 +37,8 @@ import org.apache.parquet.internal.column.columnindex.OffsetIndex;
  * @see ColumnIndexFilter#calculateRowRanges(Filter, ColumnIndexStore, Set, long)
  */
 public class RowRanges {
-  private static class Range {
+  // Make it public because some uppler layer application need to access it
+  public static class Range {
 
     // Returns the union of the two ranges or null if there are elements between them.
     private static Range union(Range left, Range right) {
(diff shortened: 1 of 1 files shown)


parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
```diff
@@ -63,8 +64,8 @@ private static Range intersection(Range left, Range right) {
       return null;
     }
 
-    final long from;
-    final long to;
+    public final long from;
+    public long to;
```
gszadovszky, line 68: `to` should also be final.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
```diff
@@ -299,6 +300,10 @@ public boolean isOverlapping(long from, long to) {
         (r1, r2) -> r1.isBefore(r2) ? -1 : r1.isAfter(r2) ? 1 : 0) >= 0;
   }
 
+  public List<Range> getRanges() {
+    return ranges;
```
gszadovszky, line 304: I don't like returning `ranges` directly. This is a mutable list so the caller can modify it without any control. I would suggest wrapping it into an unmodifiable list.

gszadovszky (changes requested):

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
shangxinli: Fixed it.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
shangxinli: Yeah, that makes sense. After this PR, I will change the master branch also.

gszadovszky (approved):

PR #923, opened 2021-08-14
[PARQUET-1968] FilterApi support In predicate

huaxingao:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-1968

### Tests

### Commits

### Documentation


huaxingao:
@gszadovszky @shangxinli @rdblue Could you please take a look at this PR when you have time? Thanks a lot!

huaxingao:
also cc @chenjunjiedada

pom.xml
```diff
@@ -478,6 +478,7 @@
                 change to fix a integer overflow issue.
                 TODO: remove this after Parquet 1.13 release -->
               <exclude>org.apache.parquet.column.values.dictionary.DictionaryValuesWriter#dictionaryByteSize</exclude>
+              <exclude>org.apache.parquet.filter2.predicate.FilterPredicate</exclude>
```
gszadovszky, line 481: However, it is not what filter2 API is designed for, technically it is possible to the user to implement this interface and by adding new methods to it we really break our API.
What do you think about adding default implementations by throwing `UnsupportedOperationException`? This way we do not need to add this class here.

gszadovszky (commented):
I have some comments in the code. All of them are more to open discussions so I neither approve nor disapprove for now.
Otherwise the code seems good to me. Thanks a lot for working on it!

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -247,6 +250,80 @@ public int hashCode() {
     }
   }
 
+  // base class for In and NotIn
```
shangxinli, line 253: Have a better comment since it is public method

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -247,6 +250,80 @@ public int hashCode() {
...
+    public String toString() {
+      return toString;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      SetColumnFilterPredicate<?> that = (SetColumnFilterPredicate<?>) o;
+      return column.equals(that.column) && values.equals(that.values) && Objects.equals(toString, that.toString);
```
shangxinli, line 294: Is toString comparison still needed here? It seems toString have (values and class). You can just compare class here.

parquet-column/src/main/java/org/apache/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
```diff
@@ -123,6 +124,46 @@ public boolean accept(Visitor visitor) {
     }
   }
 
+  abstract class SetInspector implements IncrementallyUpdatedFilterPredicate {
```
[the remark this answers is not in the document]
huaxingao: Changed. Thanks!

pom.xml
huaxingao: Changed. Thanks!

huaxingao:
@gszadovszky @shangxinli @dbtsai Thank you all very much for reviewing! I have changed the code to generate the visit methods for in/notIn and also added the default by throwing Exception. Will address the rest of the comments tomorrow or the day after tomorrow.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -247,6 +250,80 @@ public int hashCode() {
...
+      String name = getClass().getSimpleName().toLowerCase(Locale.ENGLISH);
+      StringBuilder str = new StringBuilder();
+      int iter = 0;
+      for (T value : values) {
+        if (iter >= 100) break;
+        str.append(value).append(", ");
+        iter++;
+      }
+      String valueStr = values.size() <= 100 ? str.substring(0, str.length() - 2) : str + "...";
+      this.toString = name + "(" + column.getColumnPath().toDotString() + ", " + valueStr + ")";
```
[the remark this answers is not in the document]
viirya: Is it just enough to replace `str + "..."` to `str.append("...").toString`?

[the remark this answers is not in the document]
viirya: `str.substring(0, str.length() - 2)` is still `StringBuilder` operation. Seems fine?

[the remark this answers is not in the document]
viirya: Maybe we can replace line 273 with `StringBuilder` operation too?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -287,6 +291,27 @@ boolean isNullPage(int pageIndex) {
           pageIndex -> nullCounts[pageIndex] > 0 || matchingIndexes.contains(pageIndex));
     }
 
+    @Override
```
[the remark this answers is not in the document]
huaxingao: @gszadovszky I tried this: if the values in a page are <= the max value in the IN set, and >= the min value in the IN set, then the page might contain the values in the IN set. I am not sure if this is want you want so I only changed `In` for now. Please take a look. Thanks!

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
huaxingao: Fixed. Thanks!

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -247,6 +250,80 @@ public int hashCode() {
...
+    private final Column<T> column;
+    private final Set<T> values;
+    private final String toString;
+
+    protected SetColumnFilterPredicate(Column<T> column, Set<T> values) {
+      this.column = Objects.requireNonNull(column, "column cannot be null");
+      this.values = Objects.requireNonNull(values, "values cannot be null");
+      checkArgument(!values.isEmpty(), "values in SetColumnFilterPredicate shouldn't be empty!");
+
+      String name = getClass().getSimpleName().toLowerCase(Locale.ENGLISH);
```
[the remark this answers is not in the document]
huaxingao: Removed.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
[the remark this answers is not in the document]
huaxingao: Fixed. Thanks!

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -247,6 +250,80 @@ public int hashCode() {
...
+
+    @Override
+    public String toString() {
+      return toString;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
```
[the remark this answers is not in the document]
huaxingao: Yes, but just trying to follow the style at https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java#L150

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
huaxingao: Removed toString comparison

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
```diff
@@ -1,14 +1,14 @@
-/*
```
gszadovszky, line 1: We have to validate In and NotIn for the record level filtering as well. All the tests here are validating higher levels only. (Record level filtering is when you actually read the data and check the values one by one to return only those values that are fulfill the filter. This is done by the class `IncrementallyUpdatedFilterPredicateBuilder` generated by this one.) 

In would suggest checking and extending the test classes `TestRecordLevelFilters`, `TestColumnIndexFiltering` and `TestBloomFiltering`.

gszadovszky (changes requested):
I have some more concrete comments in the code. Some more work is needed but I think it is going to the good direction. Thanks a lot for your efforts.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -250,27 +250,16 @@ public Eq(Column<T> column, T value) {
     }
   }
 
-  // base class for In and NotIn
+  // base class for In and NotIn. In is used to filter data based on a list of values. NotIn is used to filter data that
+  // are not in the list of values.
```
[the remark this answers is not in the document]
huaxingao: Changed.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/bloomfilterlevel/BloomFilterImpl.java
```diff
@@ -118,6 +120,45 @@ private ColumnChunkMetaData getColumnChunk(ColumnPath columnPath) {
...
+        for (T value : values) {
+          if (bloomFilter.findHash(bloomFilter.hash(value))) {
+            return BLOCK_MIGHT_MATCH;
+          }
+        }
+        return BLOCK_CANNOT_MATCH;
+      }
+    } catch (RuntimeException e) {
+      LOG.warn(e.getMessage());
+    }
```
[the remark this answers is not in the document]
huaxingao: Removed.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -293,8 +290,48 @@ boolean isNullPage(int pageIndex) {
...
+            for (int i = 0; i < nullCounts.length; i++) {
+              if (nullCounts[i] > 0) {
+                matchingIndexes1.add(i);
+              }
+            }
+          }
+        }
+      }
+
+      IntSet matchingIndexes2 = new IntOpenHashSet();
```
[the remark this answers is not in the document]
huaxingao: I will think more about what is a better way to implement this `visit(In<T> in)`.

parquet-column/src/test/java/org/apache/parquet/internal/filter2/columnindex/TestColumnIndexFilter.java
```diff
@@ -410,6 +410,7 @@ public void testFiltering() {
     Set<Integer> set5 = new HashSet<>();
     set5.add(7);
     set5.add(20);
+    System.out.println(in(intColumn("column5"), set5).toString());
```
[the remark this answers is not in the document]
huaxingao: Sorry, removed.

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
```diff
@@ -67,15 +67,18 @@ public void run() throws IOException {
     add("package org.apache.parquet.filter2.recordlevel;\n" +
         "\n" +
         "import java.util.List;\n" +
+        "import java.util.Set;\n" +
         "\n" +
         "import org.apache.parquet.hadoop.metadata.ColumnPath;\n" +
         "import org.apache.parquet.filter2.predicate.Operators.Eq;\n" +
         "import org.apache.parquet.filter2.predicate.Operators.Gt;\n" +
         "import org.apache.parquet.filter2.predicate.Operators.GtEq;\n" +
+      "import org.apache.parquet.filter2.predicate.Operators.In;\n" +
```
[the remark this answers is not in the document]
huaxingao: fixed

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
huaxingao: I will think about how to achieve this. Thanks for the suggestion.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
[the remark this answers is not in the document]
gszadovszky: It is a nit but javadoc style comments in java starts with `/**`. If the simple `/*` one is used it won't be generated to the javadocs.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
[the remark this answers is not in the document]
huaxingao: @gszadovszky Sorry for the delay. I did the following changes:
1. Removed `TreeSet` to avoid sort the whole set, and added a method to get max and min.
2. Updated `StatisticsFilter` to have the similar range comparison.
3. I actually didn't change `notIn` because I don't think range checking works with `notIn`: if not in the range, `notIn` is true, but if in the range, it doesn't mean `notIn` is false.  For example, if we have `1, 2, 3, 6, 7 8, 9, 10` and the `notIn` predicate has set values `4, 5`, `4, 5` is in the range of 1 to 10, but `1, 2, 3, 6, 7 8, 9, 10` doesn't contain `4, 5`. In `StatisticsFilter` I simply return `BLOCK_MIGHT_MATCH;` for `notIn`. I probably should return `IndexIterator.all(getPageCount());` in `ColumnIndexBuilder` to be consistent with `StatisticsFilter`.

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
huaxingao: @gszadovszky I added a test in `TestRecordLevelFilters.java` to test the new methods in the generated class `IncrementallyUpdatedFilterPredicateBuilder`. I have added tests in `TestColumnIndexFiltering` and `TestBloomFiltering` in my original changes. Do I need more tests in these two files?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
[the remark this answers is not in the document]
gszadovszky: The return types of `StatisticsFilter` and `ColumnIndexBuilder` are different because the first one simply returns whether the related row group can be dropped or not while the second one returns the indexes of the matching pages.

For `noIn` I agree. However there is an edge case you can figure out something when the min and max values are the same. In this case all the values shall be the same in the row group / page so you can say something in the case of there is only one element in `notIn`. It sounds a very rare scenario so I am fine returning "might match".

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
gszadovszky: `TestColumnIndexFiltering` is another test class different from `TestColumnIndexFilter`. The first one is a higher level test. It would be nice if you could add some tests there as well.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
```diff
@@ -146,6 +147,33 @@ public void testAllFilter() throws Exception {
...
+  @Test
+  public void testInFilter() throws Exception {
+    BinaryColumn name = binaryColumn("name");
+
+    HashSet<Binary> nameSet = new HashSet<>();
+    nameSet.add(Binary.fromString("thing2"));
+    nameSet.add(Binary.fromString("thing1"));
+    for (int i = 100; i < 200; i++) {
+      nameSet.add(Binary.fromString("p" + i));
+    }
```
gszadovszky, line 159: I think you should also add some value to the set that are not in the file.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
```diff
@@ -146,6 +147,33 @@ public void testAllFilter() throws Exception {
...
+    expectedNames.add("thing2");
+    for (int i = 100; i < 200; i++) {
+      expectedNames.add("p" + i);
+    }
+
+    assertEquals(expectedNames.get(0), ((Group)(found.get(0))).getString("name", 0));
+    assertEquals(expectedNames.get(1), ((Group)(found.get(1))).getString("name", 0));
+    for (int i = 2; i < 102; i++) {
+      assertEquals(expectedNames.get(i), ((Group)(found.get(i))).getString("name", 0));
+    }
```
gszadovszky, line 174: This is not correct this way. You should validate that all the values returned by the reader fulfills the filter and there are no values left out. So, `"thing1"`, `"thing2"` and from `"p100"` to `"p199"` and nothing else.

gszadovszky (changes requested):
Added a couple of more comments/requests.
Sorry if I am a bit strict here but filtering is not an easy topic and can have serious issues (lost of data).

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
huaxingao: Added. Thanks!

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -326,12 +323,27 @@ boolean isNullPage(int pageIndex) {
...
+      matchingIndex.addAll(matchingIndexesForNull);  // add the matching null pages
+      return IndexIterator.filter(getPageCount(), pageIndex -> matchingIndex.contains(pageIndex));
+    }
+
+    private <T extends Comparable<T>> T getMaxOrMin(boolean isMax, Set<T> values) {
+      T res = null;
+      for (T element : values) {
+        if (res == null) {
+          res = element;
+        } else if (isMax && res != null && element != null && res.compareTo(element) < 0) {
```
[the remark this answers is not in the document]
huaxingao: I changed to `PrimitiveComparator`. I checked `instanceof` and then cast to use the `PrimitiveComparator` for each of the type. Not sure if this is the correct way. Please take a look.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -186,26 +186,36 @@ private boolean hasNulls(ColumnChunkMetaData column) {
...
-          return false;
-      } else {
-        // numNulls is not set. We don't know anything about the nulls in this chunk
-        if (!stats.isNumNullsSet()) {
-          return false;
-        }
-        if (hasNulls(meta)) {
-          return false;
-        }
+  private <T extends Comparable<T>> T getMaxOrMin(boolean isMax, Set<T> values) {
```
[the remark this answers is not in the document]
huaxingao: Done. Thanks!

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -291,32 +291,29 @@ boolean isNullPage(int pageIndex) {
...
               }
             }
           }
         }
       }
 
-      IntSet matchingIndexes2 = new IntOpenHashSet();
-      IntSet matchingIndexes3 = new IntOpenHashSet();
+      IntSet matchingIndexesLessThanMax = new IntOpenHashSet();
+      IntSet matchingIndexesLargerThanMin = new IntOpenHashSet();
```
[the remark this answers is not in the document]
huaxingao: Changed.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -186,26 +186,36 @@ private boolean hasNulls(ColumnChunkMetaData column) {
...
+      }
+    }
+
     // drop if all the element in value < min || all the element in value > max
-    return allElementCanBeDropped(stats, values, meta);
+    if (stats.compareMinToValue(getMaxOrMin(true, values)) <= 0 &&
+      stats.compareMaxToValue(getMaxOrMin(false, values)) >= 0) {
+      return BLOCK_MIGHT_MATCH;
+    }
+    else {
```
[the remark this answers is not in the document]
huaxingao: Done.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
huaxingao: Added. Thanks!

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
huaxingao: I added `assert(found.size() == 102)`. Since I have already checked that `found` contains `"thing1"`, `"thing2"` and from `"p100"` to `"p199"`, I think this assert size is sufficient to check if `found` doesn't contain anything else.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public final class MinMax<T> {
+  private PrimitiveComparator comparator;
+  private Iterator<T> iterator;
+  private T min = null;
+  private T max = null;
+
+  public MinMax(PrimitiveComparator comparator, Iterator<T> iterator) {
+    this.comparator = comparator;
+    this.iterator = iterator;
```
gszadovszky, line 38: You do not need to store the iterator and comparator instances since they are only used at initialization. I would simply pass them to `getMinAndMax`.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * This class calculates the max and min values of a Set.
+ */
+public final class MinMax<T> {
+  private PrimitiveComparator comparator;
+  private Iterator<T> iterator;
+  private T min = null;
+  private T max = null;
+
+  public MinMax(PrimitiveComparator comparator, Iterator<T> iterator) {
```
gszadovszky, line 36: I would expect an `Iterable` instead of an `Iterator`. This way the Set can be passed directly.

gszadovszky: ... and you may use foreach instead of a while loop on the iterator.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -0,0 +1,89 @@
...
+package org.apache.parquet.column;
+
+import java.util.Iterator;
+import java.util.Set;
+
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.PrimitiveComparator;
+
+/**
+ * This class calculates the max and min values of a Set.
```
gszadovszky, line 28: If you are more general in implementation by using an Iterator (or Iterable) you should use the same in the comments.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -0,0 +1,89 @@
...
+          (element instanceof Binary &&
+            ((PrimitiveComparator<Binary>)comparator).compare((Binary)max, (Binary)element) < 0) ||
+          (element instanceof Double &&
+             ((PrimitiveComparator<Double>)comparator).compare((Double)max, (Double)element) < 0) ||
+          (element instanceof Float &&
+             ((PrimitiveComparator<Float>)comparator).compare((Float)max, (Float)element) < 0) ||
+          (element instanceof Boolean &&
+            ((PrimitiveComparator<Boolean>)comparator).compare((Boolean)max, (Boolean)element) < 0) ||
+          (element instanceof Long &&
+            ((PrimitiveComparator<Long>)comparator).compare((Long) max, (Long)element) < 0))
```
gszadovszky, line 67: You do not need to check the instance and cast. You may expect a `PrimitiveComparator<T>` and use it directly.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -0,0 +1,89 @@
...
+  public T getMax() {
+    return max;
+  }
+
+  private void getMinAndMax() {
+    while(iterator.hasNext())  {
+      T element = iterator.next();
+      if (max == null) {
+        max = element;
+      } else if (max != null && element != null) {
```
gszadovszky, line 55: You are already in the else path so do not need to check for `max != null`.

gszadovszky (changes requested):
I have a couple of comments for the new MinMax class but otherwise everything seems great!

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -18,25 +18,17 @@
...
-  private PrimitiveComparator comparator;
-  private Iterator<T> iterator;
   private T min = null;
   private T max = null;
 
-  public MinMax(PrimitiveComparator comparator, Iterator<T> iterator) {
-    this.comparator = comparator;
-    this.iterator = iterator;
-    getMinAndMax();
+  public MinMax(PrimitiveComparator comparator, Iterable<T> iterable) {
```
gszadovszky, line 30: I think,  you should get warnings if you use a generic class without the generic type. `PrimitiveComparator<T>` or even `Comparator<T>` should work fine.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -47,43 +39,22 @@ public T getMax() {
...
-          (element instanceof Double &&
-             ((PrimitiveComparator<Double>)comparator).compare((Double)max, (Double)element) < 0) ||
-          (element instanceof Float &&
-             ((PrimitiveComparator<Float>)comparator).compare((Float)max, (Float)element) < 0) ||
-          (element instanceof Boolean &&
-            ((PrimitiveComparator<Boolean>)comparator).compare((Boolean)max, (Boolean)element) < 0) ||
-          (element instanceof Long &&
-            ((PrimitiveComparator<Long>)comparator).compare((Long) max, (Long)element) < 0))
+      } else if (element != null) {
+        if (comparator.compare(max, element) < 0) {
```
gszadovszky, line 47: nit: You may combine the two with an `&&`.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
```diff
@@ -47,43 +39,22 @@ public T getMax() {
...
-          (element instanceof Double &&
-            ((PrimitiveComparator<Double>)comparator).compare((Double)min, (Double)element) > 0) ||
-          (element instanceof Float &&
-             ((PrimitiveComparator<Float>)comparator).compare((Float)min, (Float)element) > 0) ||
-          (element instanceof Boolean &&
-            ((PrimitiveComparator<Boolean>)comparator).compare((Boolean)min, (Boolean)element) > 0) ||
-          (element instanceof Long &&
-            ((PrimitiveComparator<Long>)comparator).compare((Long)min, (Long)element) > 0))
+      } else if (element != null) {
+        if (comparator.compare(min, element) > 0) {
```
gszadovszky, line 54: See above

gszadovszky (changes requested):
I have some minor issues only. Thanks a lot for your efforts to implement this! I think it is a great improvement for the query engines.

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
huaxingao: Fixed. Thanks!

parquet-column/src/main/java/org/apache/parquet/column/MinMax.java
huaxingao: Done.

gszadovszky (approved):

huaxingao:
@gszadovszky @shangxinli @viirya @dbtsai Thank you so much for all your help!!

gszadovszky:
Thank you for your contribution, @huaxingao! Great work!

PR #925, opened 2021-08-28
PARQUET-2078: Failed to read parquet file after writing with the same …

loudongfeng:
…parquet version

### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET) issues:  
  - https://issues.apache.org/jira/browse/PARQUET-2078

### Tests

- [ ] My PR adds the following unit tests :
TestParquetFileWriter.testWriteReadWithRecordReader
### Commits

- [ ] My commits all reference Jira issues in their subject lines.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -206,8 +206,13 @@ public class ParquetMetadataConverter {
     for (BlockMetaData block : blocks) {
       numRows += block.getRowCount();
       long blockStartPos = block.getStartingPos();
+      // first block
+      if (blockStartPos == 4) {
+        preBlockStartPos = 0;
+        preBlockCompressedSize = 0;
+      }
       if (preBlockStartPos != 0) {
-        assert blockStartPos == preBlockStartPos + preBlockCompressedSize;
+        assert blockStartPos >= preBlockStartPos + preBlockCompressedSize;
       }
       preBlockStartPos = blockStartPos;
       preBlockCompressedSize = block.getCompressedSize();
@@ -1243,14 +1248,9 @@ public class ParquetMetadataConverter {
       if (rowGroup.isSetFile_offset()) {
         assert rowGroup.isSetTotal_compressed_size();
 
-        if (preStartIndex == 0) {
-          //the first block always holds the truth
-          startIndex = rowGroup.getFile_offset();
-        } else {
-          //calculate offset for other blocks
-          startIndex = preStartIndex + preCompressedSize;
-        }
-
+        //the file_offset of first block always holds the truth, while other blocks don't :
+        //see PARQUET-2078 for details
+        startIndex = tryUseFileOffset(rowGroup, preStartIndex, preCompressedSize);
         preStartIndex = startIndex;
         preCompressedSize = rowGroup.getTotal_compressed_size();
 
@@ -1276,6 +1276,24 @@ public class ParquetMetadataConverter {
     return metaData;
   }
 
+  private static long tryUseFileOffset(RowGroup rowGroup, long preStartIndex, long preCompressedSize) {
+    long startIndex = 0;
+    startIndex = rowGroup.getFile_offset();
+    // skip checking the first rowGroup
+    // (in case of summary file, there are multiple first groups from different footers)
+    if (preStartIndex != 0 && preStartIndex <= startIndex) {
+
+      //calculate start index for other blocks
+      long minStartIndex = preStartIndex + preCompressedSize;
+      if (startIndex < minStartIndex) {
+        // a bad offset detected, try first column's offset
+        // can not use minStartIndex in case of padding
+        startIndex = getOffset(rowGroup.getColumns().get(0));
+      }
+    }
+    return startIndex;
+  }
+
   // Visible for testing
   static FileMetaData filterFileMetaDataByStart(FileMetaData metaData, OffsetMetadataFilter filter) {
     List<RowGroup> rowGroups = metaData.getRow_groups();
@@ -1287,14 +1305,9 @@ public class ParquetMetadataConverter {
       if (rowGroup.isSetFile_offset()) {
         assert rowGroup.isSetTotal_compressed_size();
 
-        if (preStartIndex == 0) {
-          //the first block always holds the truth
-          startIndex = rowGroup.getFile_offset();
-        } else {
-          //calculate offset for other blocks
-          startIndex = preStartIndex + preCompressedSize;
-        }
-
+        //the file_offset of first block always holds the truth, while other blocks don't :
+        //see PARQUET-2078 for details
+        startIndex = tryUseFileOffset(rowGroup, preStartIndex, preCompressedSize);
         preStartIndex = startIndex;
         preCompressedSize = rowGroup.getTotal_compressed_size();
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -201,9 +201,18 @@ public class ParquetMetadataConverter {
     List<BlockMetaData> blocks = parquetMetadata.getBlocks();
     List<RowGroup> rowGroups = new ArrayList<RowGroup>();
     long numRows = 0;
+    long preBlockStartPos = 0;
+    long preBlockCompressedSize = 0;
     for (BlockMetaData block : blocks) {
       numRows += block.getRowCount();
+      long blockStartPos = block.getStartingPos();
+      if (preBlockStartPos != 0) {
+        assert blockStartPos == preBlockStartPos + preBlockCompressedSize;
+      }
+      preBlockStartPos = blockStartPos;
+      preBlockCompressedSize = block.getCompressedSize();
       addRowGroup(parquetMetadata, rowGroups, block, fileEncryptor);
+
     }
     FileMetaData fileMetaData = new FileMetaData(
         currentVersion,
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -1226,9 +1226,28 @@ public class ParquetMetadataConverter {
   static FileMetaData filterFileMetaDataByMidpoint(FileMetaData metaData, RangeMetadataFilter filter) {
     List<RowGroup> rowGroups = metaData.getRow_groups();
     List<RowGroup> newRowGroups = new ArrayList<RowGroup>();
+    long preStartIndex = 0;
+    long preCompressedSize = 0;
     for (RowGroup rowGroup : rowGroups) {
       long totalSize = 0;
-      long startIndex = getOffset(rowGroup);
+      long startIndex;
+      if (rowGroup.isSetFile_offset()) {
+        assert rowGroup.isSetTotal_compressed_size();
+
+        if (preStartIndex == 0) {
+          //the first block always holds the truth
+          startIndex = rowGroup.getFile_offset();
+        } else {
+          //calculate offset for other blocks
+          startIndex = preStartIndex + preCompressedSize;
+        }
+
+        preStartIndex = startIndex;
+        preCompressedSize = rowGroup.getTotal_compressed_size();
+
+      } else {
+        startIndex = getOffset(rowGroup.getColumns().get(0));
+      }
 
       if (rowGroup.isSetTotal_compressed_size()) {
         totalSize = rowGroup.getTotal_compressed_size();
@@ -1252,8 +1271,27 @@ public class ParquetMetadataConverter {
   static FileMetaData filterFileMetaDataByStart(FileMetaData metaData, OffsetMetadataFilter filter) {
     List<RowGroup> rowGroups = metaData.getRow_groups();
     List<RowGroup> newRowGroups = new ArrayList<RowGroup>();
+    long preStartIndex = 0;
+    long preCompressedSize = 0;
     for (RowGroup rowGroup : rowGroups) {
-      long startIndex = getOffset(rowGroup);
+      long startIndex;
+      if (rowGroup.isSetFile_offset()) {
+        assert rowGroup.isSetTotal_compressed_size();
+
+        if (preStartIndex == 0) {
+          //the first block always holds the truth
+          startIndex = rowGroup.getFile_offset();
+        } else {
+          //calculate offset for other blocks
+          startIndex = preStartIndex + preCompressedSize;
+        }
+
+        preStartIndex = startIndex;
+        preCompressedSize = rowGroup.getTotal_compressed_size();
+
+      } else {
+        startIndex = getOffset(rowGroup.getColumns().get(0));
+      }
 
       if (filter.contains(startIndex)) {
         newRowGroups.add(rowGroup);
@@ -1264,6 +1302,9 @@ public class ParquetMetadataConverter {
   }
 
   static long getOffset(RowGroup rowGroup) {
+    if (rowGroup.isSetFile_offset()) {
+      return rowGroup.getFile_offset();
+    }
     return getOffset(rowGroup.getColumns().get(0));
   }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -119,6 +119,7 @@ import org.apache.parquet.internal.column.columnindex.BinaryTruncator;
 import org.apache.parquet.internal.column.columnindex.ColumnIndexBuilder;
 import org.apache.parquet.internal.column.columnindex.OffsetIndexBuilder;
 import org.apache.parquet.internal.hadoop.metadata.IndexReference;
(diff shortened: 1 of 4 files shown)


loudongfeng:
This patch fixes both the write path and the read path.
Write path:
fix the currentDictionaryPageOffset reuse issue, then RowGroup.file_offset in parquet file will be correct.
Read path:
supporting read parquet file with wrong RowGroup.file_offset (by ignoring it)
I'm not sure how the read path changes will inflect encryption files written by parquet 1.12.0.
But encryption files‘s RowGroup.file_offset is wrongly setted already.

Another solution for read path:
only ignore file_offset when
1. file version is parquet 1.12.0 .
2. file is not encrypted

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -239,6 +248,82 @@ public void testWriteRead() throws Exception {
     PrintFooter.main(new String[] {path.toString()});
   }
 
+  @Test
+  public void testWriteReadWithRecordReader() throws Exception {
```
gszadovszky, line 252: I think this unit test does not reproduce the original scenario. Please check my comment in the [jira](https://issues.apache.org/jira/browse/PARQUET-2078?focusedCommentId=17406621&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17406621).

gszadovszky (changes requested):
Thanks a lot for working on this, @loudongfeng!

I think it is fine to not use the file offset at all but calculate it by using the offsets of the first column chunk. @ggershinsky, what do you think?

But, you also need to handle the case of invalid dictionary offset in `getOffset(ColumnChunk)`. Please check my comment in the [jira](https://issues.apache.org/jira/browse/PARQUET-2078?focusedCommentId=17406621&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17406621).

Also, please check the whole code for potential usage of the dictionary offset and the file offset. It would be also great if you could validate the new code with the original invalid files.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
gszadovszky: My bad, it seems just fine.

gszadovszky (approved):
Excellent work. Thanks a lot for your efforts!

shangxinli:
@ggershinsky Do you want to have a look?

ggershinsky:
Sure. This won't work if the first column is encrypted and the reader doesn't have its key. Can the "write" part be fixed instead, so the RowGroup offset is set correctly?

gszadovszky:
@ggershinsky, even though this PR fixes the write path as well we have already released 1.12.0 so we have to prepare for the case of `RowGroup.file_offset` is incorrect.

ggershinsky:
Yep, but the current fix perpetuates the situation where some readers can't process encrypted files, even if they have keys for all projected columns; doesn't look like an optimal long-term solution.. I'm just back online after a vacation; will go over the details in this thread, maybe something else can be done here.

gszadovszky:
@ggershinsky, sorry, I've completely missed the fact that `RowGroup.file_offset` is introduced for the encryption feature and it actually required for it. Somehow we shall check if the `file_offset` points to the previous row group. At the worst case we shall check at least if the parquet file is written by 1.12.0. But in this case we only know that the `file_offset` might be wrong.

loudongfeng:
FYI,Maybe we can make use of this information :
RowGroup[n].file_offset = RowGroup[n-1].file_offset + RowGroup[n-1].total_compressed_size
total_compressed_size always holds the truth, while file_offset doesn't.
total_compressed_size is also introduced in for encryption feature.

ggershinsky:
@gszadovszky No problem at all, thank you for helping with this!

ggershinsky:
> FYI,Maybe we can make use of this information :
> RowGroup[n].file_offset = RowGroup[n-1].file_offset + RowGroup[n-1].total_compressed_size
> total_compressed_size always holds the truth, while file_offset doesn't.
> total_compressed_size is also introduced in for encryption feature.

Yep, exactly! If there are no hidden surprises and this works as expected, it would certainly be the optimal solution. While at it, maybe you can also add a check at the [write side](https://github.com/apache/parquet-mr/blob/apache-parquet-1.12.0/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L580), to verify the RG offset values in a similar manner (must be equal to a sum of previous RG sizes, plus the first RG offset; this also runs in a loop). Thanks @loudongfeng !

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1226,12 +1226,25 @@ public ParquetMetadata readParquetMetadata(InputStream from) throws IOException
...
-
       if (rowGroup.isSetFile_offset()) {
-        startIndex = rowGroup.getFile_offset();
+        assert rowGroup.isSetTotal_compressed_size();
+
+        if (preStartIndex == 0) {
+          //the first block always holds the truth
+          startIndex = rowGroup.getFile_offset();
+        } else {
+          //calculate offset for other blocks
```
ggershinsky, line 1241: Could you expand the comment a bit, to briefly explain the problem? (maybe with mentioning the jira number). To help ensure future changes don't revert this to the more intuitive getFile_offset().

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -201,9 +201,23 @@ public FileMetaData toParquetMetadata(int currentVersion, ParquetMetadata parque
     List<BlockMetaData> blocks = parquetMetadata.getBlocks();
     List<RowGroup> rowGroups = new ArrayList<RowGroup>();
     long numRows = 0;
+    long preBlockStartPos = 0;
+    long preBlockCompressedSize = 0;
     for (BlockMetaData block : blocks) {
       numRows += block.getRowCount();
+      long blockStartPos = block.getStartingPos();
+      // first block
+      if (blockStartPos == 4) {
```
ggershinsky, line 210: why this check is necessary, doesn't the first block always start at 4? Or this addresses a file merging usecase?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -201,9 +201,23 @@ public FileMetaData toParquetMetadata(int currentVersion, ParquetMetadata parque
...
+        preBlockStartPos = 0;
+        preBlockCompressedSize = 0;
+      }
+      if (preBlockStartPos != 0) {
+        assert blockStartPos >= preBlockStartPos + preBlockCompressedSize;
```
ggershinsky, line 215: why >= instead of == ?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1254,14 +1276,41 @@ static FileMetaData filterFileMetaDataByMidpoint(FileMetaData metaData, RangeMet
...
+    // skip checking the first rowGroup
+    // (in case of summary file, there are multiple first groups from different footers)
+    if (preStartIndex != 0 && preStartIndex <= startIndex) {
+
+      //calculate start index for other blocks
+      long minStartIndex = preStartIndex + preCompressedSize;
+      if (startIndex < minStartIndex) {
+        // a bad offset detected, try first column's offset
+        // can not use minStartIndex in case of padding
+        startIndex = getOffset(rowGroup.getColumns().get(0));
```
ggershinsky, line 1291: this will throw an exception in encrypted files.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
loudongfeng: There are alignments between row groups. see PARQUET-306 and https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L411

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
loudongfeng: To address _common_metadata file case, which will merge multiple file footers into a single meta file.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
loudongfeng: In case of  encrypted files with wongly setted file_offset, i have no idea how to fix it when alignment padding take place.
If no padding, then we can just use the calculated index.
I didn't find any footer meta about padding position or something else indicating padding or not.

gszadovszky:
@shangxinli, @ggershinsky, please note that I'll be on vacation from today till the end of next week so won't have time for this PR. While this seems to be quite urgent so do not hesitate to push it in and initiate a release for 1.12.1. I think this fix would worth a separate release as quick as possible.
Also, do not forget to approve the unit tests execution for every new commits since @loudongfeng is not a member (or whatever GitHub actions require to be executed automatically).
-- Thanks

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
ggershinsky: got it, will check a few things

ggershinsky: @loudongfeng @gszadovszky @shangxinli @sunchao . Looks like the problem is only in one mode of encryption ("encrypted footer", where the ColumnMetaData is not set). Let me propose the following:


      long minStartIndex = preStartIndex + preCompressedSize;
      if (startIndex < minStartIndex) {
        // a bad offset detected, try first column's offset if available
       ColumnChunk columnChunk = rowGroup.getColumns().get(0);
       if (columnChunk.isSetMeta_data()) {
         startIndex = getOffset(columnChunk);
       } else { // EncryptedFooter mode. Plaintext ColumnMetaData is not available.
         // use minStartIndex (imprecise in case of padding, but good enough for filtering)
         startIndex = minStartIndex;
      }


what do you think?

ggershinsky: that might work for range filtering. as for the explicit offset filtering, the precision might be important. In systems that [use padding](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopOutputFile.java#L37) (HDFS, WEBHDFS, VIEWFS)  - instead of `startIndex = minStartIndex` , we might have to throw an exception when this is called from  `filterFileMetaDataByStart` saying that this file can't be split by offsets, and should be processed without splitting. Again, given the number of conditions for this situation to occur, and the fact that 1.12.0 is not released yet any framework, such exceptions might never be thrown in reality.

ggershinsky: Yep, that should be it (IMO). This will work for all files written by 1.12.1 (naturally), for all regular/unencrypted files files written by 1.12.0 and for most (in practice, for all) of encrypted files written by 1.12.0. Tomorrow, I'll send an update to the snippet above.

loudongfeng: @ggershinsky, your proposal sounds perfect to me.Looking forward to your patch, or shall I update commit following your proposal?

loudongfeng: Another choice: suppose columnChunk.isSetMeta_data() is the same across different row groups, how about using first columns's offset by default, only using file offset with "encrypted footer"?
(And only throw exception when bad file offset detected and called from filterFileMetaDataByStart, as you suggested.)
@ggershinsky

ggershinsky: Thanks @loudongfeng, sure, please update the pr with these changes. Regarding columnChunk.isSetMeta_data() - the result will be the same across different row groups, since the whole file is written in one mode (either with the ColumnMetaData, or without them). If they are unavailable, we can't use them for splits, so will have to throw an exception in case of filterFileMetaDataByStart (only for bad columns where startIndex < minStartIndex)

ggershinsky: and of course, I'm fine with your proposal to change the order - if you can detect the "encrypted footer" mode (eg by checking one columnChunk.isSetMeta_data()), then its ok to start with using first columns's offset

loudongfeng: Commit updated following ggershinsky's suggestions.Thanks.

ggershinsky:
Thanks @loudongfeng , looks good. I'll run the last round of checks with a number of encryption modes early next week.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1226,14 +1240,32 @@ public ParquetMetadata readParquetMetadata(InputStream from) throws IOException
...
         startIndex = rowGroup.getFile_offset();
-      } else {
-        startIndex = getOffset(rowGroup.getColumns().get(0));
+        if (invalidFileOffset(startIndex, preStartIndex, preCompressedSize)) {
+          // use minStartIndex(imprecise in case of padding, but good enough for filtering)
+          startIndex = preStartIndex + preCompressedSize;
+        }
+        preStartIndex = startIndex;
+        preCompressedSize = rowGroup.getTotal_compressed_size();
+
```
ggershinsky, line 1268: nit: empty line

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -201,9 +201,23 @@ public FileMetaData toParquetMetadata(int currentVersion, ParquetMetadata parque
...
+        preBlockStartPos = 0;
+        preBlockCompressedSize = 0;
+      }
+      if (preBlockStartPos != 0) {
+        assert blockStartPos >= preBlockStartPos + preBlockCompressedSize;
+      }
+      preBlockStartPos = blockStartPos;
+      preBlockCompressedSize = block.getCompressedSize();
       addRowGroup(parquetMetadata, rowGroups, block, fileEncryptor);
+
```
ggershinsky, line 220: nit: empty line

ggershinsky (approved):
> the last round of checks with a number of encryption modes early next week.

Everything was ok

sunchao (commented):
@loudongfeng @gszadovszky @ggershinsky it seems there is also a bug in parquet-cpp which causes incorrect file offset to be written, see https://issues.apache.org/jira/browse/SPARK-36696, so we'll want to make sure the solution here work for that case as well.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1226,14 +1239,31 @@ public ParquetMetadata readParquetMetadata(InputStream from) throws IOException
...
       long startIndex;
+      ColumnChunk columnChunk = rowGroup.getColumns().get(0);
+      if (firstColumnWithMetadata) {
+        startIndex = getOffset(columnChunk);
+      } else {
+        assert rowGroup.isSetFile_offset();
+        assert rowGroup.isSetTotal_compressed_size();
 
-      if (rowGroup.isSetFile_offset()) {
+        //the file_offset of first block always holds the truth, while other blocks don't :
```
sunchao, line 1258: I think this is no longer true with the issue we found in parquet-cpp.

loudongfeng: @sunchao Thanks for your information.May be the first rowgroup check is enough for this situation(firstFileOffset==4)? I will submit a new commit.

shangxinli:
@ggershinsky Can you have another look for the new commit?

ggershinsky:
Yep, looks good. The first row group always starts at offset 4.
@loudongfeng Maybe the hardcoded `4` should be replaced with eg `ParquetFileWriter.MAGIC.length`? Or even better, a new constant added at `ParquetFileWriter`, something like 'FIRST_ROW_OFFSET = MAGIC.length`

ggershinsky:
> it seems there is also a bug in parquet-cpp which causes incorrect file offset to be written, see https://issues.apache.org/jira/browse/SPARK-36696, so we'll want to make sure the solution here work for that case as well.

Yep, it does. I've taken the file that was posted at that jira, and read it with Spark with p1.12.0 - this indeed fails. After adding this fix to parquet, the reading worked ok. This happens because for regular files (and most of encrypted files), this fix ignores the `RowGroup.offset` field, and reverts the offset compute to the pre-1.12 behavior.

shangxinli:
Thanks, I will merge it soon.

PR #940, opened 2021-12-04
[PARQUET-2106] Refactoring lexicographic `BinaryComparator` to avoid `ByteBuffer.wrap` in the hot-path

alexeykudinkin:
This refactors `BinaryComparator` to essentially avoid any allocations in the routine of comparing 2 `Binary` objects. 

More details could be found in [PARQUET-2106](https://issues.apache.org/jira/browse/PARQUET-2106)

### Jira

- [ ] My PR addresses [PARQUET-2106](https://issues.apache.org/jira/browse/PARQUET-2106) issues and references them in the PR title. For example, "PARQUET-1234: My Parquet PR"

### Tests

- Does not need testing, since it's a simple refactoring that changes flows that are already covered

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -76,6 +76,12 @@ abstract public class Binary implements Comparable<Binary>, Serializable {
   @Override
   @Deprecated
   abstract public int compareTo(Binary other);
+
+  public abstract int compareToLexicographic(Binary other);
+
+  public abstract int compareToLexicographic(byte[] other, int otherOffset, int otherLength);
+
+  public abstract int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength);
 
   abstract public ByteBuffer toByteBuffer();
 
@@ -189,6 +195,22 @@ abstract public class Binary implements Comparable<Binary>, Serializable {
     @Override
     public int compareTo(Binary other) {
       return PrimitiveComparator.UNSIGNED_LEXICOGRAPHICAL_BINARY_COMPARATOR.compare(this, other);
+    }
+
+    @Override
+    public int compareToLexicographic(Binary other) {
+      // NOTE: We have to flip the sign, since we swap operands sides
+      return -other.compareToLexicographic(value, offset, length);
+    }
+
+    @Override
+    public int compareToLexicographic(byte[] other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(value, offset, length, other, otherOffset, otherLength);
+    }
+
+    @Override
+    public int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(value, offset, length, other, otherOffset, otherLength);
     }
 
     @Override
@@ -328,7 +350,23 @@ abstract public class Binary implements Comparable<Binary>, Serializable {
       return PrimitiveComparator.UNSIGNED_LEXICOGRAPHICAL_BINARY_COMPARATOR.compare(this, other);
     }
 
-   @Override
+    @Override
+    public int compareToLexicographic(Binary other) {
+      // NOTE: We have to flip the sign, since we swap operands sides
+      return -other.compareToLexicographic(value, 0, value.length);
+    }
+
+    @Override
+    public int compareToLexicographic(byte[] other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(this.value, 0, value.length, other, otherOffset, otherLength);
+    }
+
+    @Override
+    public int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(this.value, 0, value.length, other, otherOffset, otherLength);
+    }
+
+    @Override
     public ByteBuffer toByteBuffer() {
       return ByteBuffer.wrap(value);
     }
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -183,10 +183,10 @@ public String toString() {
   private static abstract class BinaryComparator extends PrimitiveComparator<Binary> {
     @Override
     int compareNotNulls(Binary o1, Binary o2) {
-      return compare(o1.toByteBuffer(), o2.toByteBuffer());
+      return compareBinary(o1, o2);
     }
 
-    abstract int compare(ByteBuffer b1, ByteBuffer b2);
+    abstract int compareBinary(Binary b1, Binary b2);
```
gszadovszky, line 189: nit: Maybe `compareBinaries` would be a better naming

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -613,4 +681,60 @@ private static final boolean equals(byte[] array1, int offset1, int length1, byt
     }
     return true;
   }
+
+  // TODO java-doc
```
gszadovszky, line 685: Please, do not leave TODOs in the final code. (javadoc comments wouldn't be required for private methods anyway.)

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -542,6 +606,10 @@ public static Binary fromCharSequence(CharSequence value) {
     return new FromCharSequenceBinary(value);
   }
 
+  public static int compareToLexicographic(Binary one, Binary other) {
```
gszadovszky, line 609: I think, `lexicographicCompare` would be a better naming.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -77,6 +77,12 @@ private Binary() { }
   @Deprecated
   abstract public int compareTo(Binary other);
 
+  abstract int compareToLexicographic(Binary other);
+
+  abstract int compareToLexicographic(byte[] other, int otherOffset, int otherLength);
+
+  abstract int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength);
+
```
gszadovszky, line 85: See the static method for better naming.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -191,6 +197,22 @@ public int compareTo(Binary other) {
...
+    @Override
+    public int compareToLexicographic(byte[] other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(value, offset, length, other, otherOffset, otherLength);
+    }
+
+    @Override
+    public int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(value, offset, length, other, otherOffset, otherLength);
+    }
+
```
gszadovszky, line 215: These methods should be package private.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -328,7 +350,23 @@ public int compareTo(Binary other) {
...
+    @Override
+    public int compareToLexicographic(byte[] other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(this.value, 0, value.length, other, otherOffset, otherLength);
+    }
+
+    @Override
+    public int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(this.value, 0, value.length, other, otherOffset, otherLength);
+    }
```
gszadovszky, line 367: These methods should be package private.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -475,6 +513,32 @@ public int compareTo(Binary other) {
...
+      } else {
+        // NOTE: We have to flip the sign, since we swap operands sides
+        return -Binary.compareToLexicographic(other, otherOffset, otherLength, value, offset, length);
+      }
+    }
+
+    @Override
+    public int compareToLexicographic(ByteBuffer other, int otherOffset, int otherLength) {
+      return Binary.compareToLexicographic(value, offset, length, other, otherOffset, otherLength);
+    }
```
gszadovszky, line 540: These methods should be package private.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -613,4 +681,60 @@ private static final boolean equals(byte[] array1, int offset1, int length1, byt
...
+      if (res != 0) {
+        return res;
+      }
+    }
+
+    return length1 - length2;
+  }
+
+  // TODO java-doc
+  private static final int compareToLexicographic(byte[] array1, int offset1, int length1, ByteBuffer array2, int offset2, int length2) {
```
gszadovszky, line 702: nit: I would suggest renaming `array2` to `buffer` and `array1` to `array`.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -613,4 +681,60 @@ private static final boolean equals(byte[] array1, int offset1, int length1, byt
...
+      if (res != 0) {
+        return res;
+      }
+    }
+
+    return length1 - length2;
+  }
+
+  // TODO java-doc
+  private static final int compareToLexicographic(ByteBuffer array1, int offset1, int length1, ByteBuffer array2, int offset2, int length2) {
```
gszadovszky, line 718: nit: I would suggest renaming `array1` to `buffer1` and `array2` to `buffer2`.

gszadovszky (changes requested):
Thanks a lot for working on this. I have a couple of minor comments.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
alexeykudinkin: The `Binary` refers to a class not an object, hence it feels more natural as `compareBinary`

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
alexeykudinkin: Sorry, didn't push the latest commit. This is fixed.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
alexeykudinkin: It's called `compareToLexicographic` to keep it in-line with base class's generic `compareTo`

gszadovszky: `compareTo` is usually used when `this` is to be compared to the argument. In case of a method comparing the two arguments to each other `compare` is preferred. 
I think, either `lexicographicCompare` or `compareLexicographically` would be better than the current one.

The naming of the package private methods that really compares `this` to the argument is less important to me but still, something like `lexicographicCompareTo` or `compareToLexicographically` would be better.

gszadovszky (changes requested):

alexeykudinkin:
@gszadovszky good points, will address.

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -542,6 +606,10 @@ public static Binary fromCharSequence(CharSequence value) {
     return new FromCharSequenceBinary(value);
   }
 
+  public static int lexicographicCompareTo(Binary one, Binary other) {
```
gszadovszky, line 609: Still contains `To`.
```suggestion
  public static int lexicographicCompare(Binary one, Binary other) {
```

gszadovszky (changes requested):
One naming issue left for the static object-object comparison. Otherwise, it looks good to me.

alexeykudinkin:
My bad. Updated

gszadovszky (approved):

PR #950, opened 2022-03-03
PARQUET-2006: Column resolution by ID

huaxingao:
Co-Authored-By: Huaxin Gao [huaxin_gao@apple.com](mailto:huaxin_gao@apple.com)
Co-Authored-By: Xinli Shang [shangx@uber.com](mailto:shangx@uber.com)

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2006
 
#### Changes in this PR:
Parquet relies on the column name. In a lot of usages e.g. schema resolution, this would be a problem. For example, Iceberg has schema evolution and column name could be changed. 

This PR will add the support of column resolution by Id. The Id we are using is the `field_id` in the Parquet schema (https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L398). After this PR, if the user explicitly requests for column Id resolution, Parquet readers will use the `field_id` to determine which Parquet columns to read. 

The changes are as follows:

In write path, `ParquetOutputFormat.COLUMN_ID_RESOLUTION` is introduced, the default is false. If sets to true, the `field_id` has to be unique in the entire schema, otherwise, an Exception will be thrown. In read path, `ParquetInputFormat.COLUMN_ID_RESOLUTION` is introduced and the default is false. If sets to true, Parquet readers will resolve the column using `field_id`. If there are duplicate id in the schema, an Exception will be thrown. 

For Filters, this PR adds the choice of constructing a filter using column ID instead of column path
for example, originally we only allow
```
IntColumn intColumn = intColumn("a.b.c"); 
FilterPredicate predicate = eq(intColumn, 7);
```
Now we can have
```
IntColumn intColumn = intColumn(new Type.ID(1)); 
FilterPredicate predicate = eq(intColumn, 7);
```

For Filters, the column resolution is done at `SchemaCompatibilityValidator.validateColumn`. In this method, the file's schema is used as the source of truth to resolve and validate the filter's columns. 

For read schema, the column resolution is done at `ParquetFileReader.setRequestedSchema`. Again, the file's schema is used as the source of truth to resolve and validate the requested schema. If `ParquetInputFormat.COLUMN_ID_RESOLUTION` sets to true, the id in the requested schema will be used to figured out the original `ColumnPath`. 

For example, if the file schema has column name `random`, type int, id 1 , column name `name`, type String, id 2 as follows (type is Spark type)

```
      val writeSchema =
        new StructType()
          .add("random", IntegerType, true, withId(1))
          .add("name", StringType, true, withId(0))
```
if after schema evolution, the table schema is changed to column name `a`, type int, id 1 , column name `b`, type String, id 2 as follows (type is Spark type)
```
      val readSchema =
        new StructType()
          .add("a", StringType, true, withId(0))
          .add("b", IntegerType, true, withId(1))
```
the id in the requested schema (the above readSchema) will be used to compared to the file schema to figure out the correct ColumnPath to read.

### Tests

### Commits

### Documentation


shangxinli:
Can you rebase and combine all the commits?

huaxingao:
Rebased. @shangxinli

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,92 @@ public String getFile() {
...
+
+  public MessageType setRequestedSchema(MessageType projection) {
     paths.clear();
-    for (ColumnDescriptor col : projection.getColumns()) {
+    HashSet<Type.ID> ids = new HashSet<>();
+    boolean fileSchemaIdUnique = uniqueId(fileMetaData.getSchema(), ids);
+    ids = new HashSet<>();
+    boolean projectionSchemaIdUnique = uniqueId(projection, ids);
+    MessageType schema = null;
+    // if ids are unique, use id resolution
```
shangxinli, line 918: Instead of figuring out by parquet-mr internals, did you consider letting the caller enable it explicitly? By doing that, we can still check the uniqueness, if not unique, we can throw exceptions.

shangxinli:
Hi. @huaxingao Thanks for working on it.  I just had a first-round review and left some comments.  After we address them, I will have another look.

ggershinsky:
hi @huaxingao , can you describe the lifecycle of the column IDs at a high level, either in the PR description, or in a comment? Where these IDs are stored (if in footer - which struct/field)? How are they set and written? Is the writer app expected to verify the uniqueness, or it can use this PR code for that? How the columns IDs are read and used (is the reader app expected to do anything beyond using this PR code)?
I think the answer to the last question is mostly provided, but it doesn't explicitly say what IDs are used (where they are stored / read from).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
ggershinsky: +1 to an explicit control / visibility

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
```diff
@@ -170,6 +174,24 @@ public Void visit(Not not) {
 
   private <T extends Comparable<T>> void validateColumn(Column<T> column) {
     ColumnPath path = column.getColumnPath();
+    HashSet<Type.ID> ids = new HashSet<>();
```
ggershinsky, line 177: nit: this can be introduced after the `if`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -181,7 +181,7 @@ public void initialize(ParquetFileReader reader, ParquetReadOptions options) {
     this.columnCount = requestedSchema.getPaths().size();
     // Setting the projection schema before running any filtering (e.g. getting filtered record count)
     // because projection impacts filtering
-    reader.setRequestedSchema(requestedSchema);
+    this.requestedSchema = reader.setRequestedSchema(requestedSchema);
```
ggershinsky, line 184: `this.requestedSchema` is set in the line 180; so this line overrides the class field setting. Maybe you can use a temp local variable (with appropriate name) in the line 180

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -472,7 +488,13 @@ public static boolean getPageWriteChecksumEnabled(Configuration conf) {
     LOG.debug("Parquet properties are:\n{}", props);
 
     WriteContext fileWriteContext = writeSupport.init(conf);
-    
+
+    MessageType schema = fileWriteContext.getSchema();
+    HashSet<Type.ID> ids = new HashSet<>();
```
ggershinsky, line 493: this variable is not used outside `checkDuplicateId`; is there a way to hide it?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,92 @@ public String getFile() {
...
+        if (unique) unique = uniqueId(field.asGroupType(), ids);
+      }
+    }
+    return unique;
+  }
+
+  public MessageType setRequestedSchema(MessageType projection) {
     paths.clear();
-    for (ColumnDescriptor col : projection.getColumns()) {
+    HashSet<Type.ID> ids = new HashSet<>();
```
ggershinsky, line 913: this variable is not used outside `uniqueId`; is there a way to hide it?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
huaxingao: Fixed. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
huaxingao: Fixed. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
huaxingao: Updated

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
huaxingao: Good idea! Added `ParquetInputFormat.COLUMN_ID_RESOLUTION` to control this.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
huaxingao: Fixed. Thanks!

huaxingao:
@ggershinsky I updated the description. Please check again to see if it is clear to you. Thanks!

ggershinsky:
Thanks @huaxingao , one more question / clarification. In the writer,
>  field_id has to be unique in the entire schema, otherwise, an Exception will be thrown.

what happens if the column ids are not set by the caller?

In the reader,
> the id in the requested schema will be used to figured out the original ColumnPath.

what happens if the file footer doesn't have the column ids (field_id were not set by the writer)?

shangxinli:
@huaxingao @ggershinsky Will you be able to join tomorrow's meeting to have a discussion on the open issues? We can try to close then in the meeting and move this PR forward.

huaxingao:
@ggershinsky 
I think in write/read, if `COLUMN_ID_RESOLUTION` sets to true but field_id were not set by the caller/writer, we need to throw Exception.

huaxingao:
@shangxinli Yes, I will join the meeting tomorrow.

ggershinsky:
I'll join too.

shangxinli:
> hi @huaxingao , can you describe the lifecycle of the column IDs at a high level, either in the PR description, or in a comment? Where these IDs are stored (if in footer - which struct/field)? How are they set and written? Is the writer app expected to verify the uniqueness, or it can use this PR code for that? How the column IDs are read and used (is the reader app expected to do anything beyond using this PR code)? I think the answer to the last question is mostly provided, but it doesn't explicitly say what IDs are used (where they are stored / read from).

+1, I think we can add it to the design doc

parquet-column/src/main/java/org/apache/parquet/column/ColumnDescriptor.java
```diff
@@ -70,7 +71,20 @@ public ColumnDescriptor(String[] path, PrimitiveTypeName type,
...
+    this(path, null, type, maxRep, maxDef);
+  }
+
+  /**
+   * @param path the path to the leaf field in the schema
+   * @param id the id to the leaf field in the schema
+   * @param type the type of the field
+   * @param maxRep the maximum repetition level for that path
+   * @param maxDef the maximum definition level for that path
+   * @deprecated will be removed in 2.0.0; Use {@link #ColumnDescriptor(String[], PrimitiveType, int, int)}
```
rdblue, line 83: Is this correct? I would expect to deprecate the old one.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -266,6 +269,10 @@ public int getMaxBloomFilterBytes() {
     return maxBloomFilterBytes;
   }
 
+  public boolean getColumnIdResolution() {
```
rdblue, line 272: I prefer to use the verb from the setting, rather than adding `get` as a second verb. Code will read more easily if you use `boolean resolveColumnsByID()`. You can also add "should" or "is" to denote that the return is a boolean if you like that convention.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -511,6 +519,11 @@ public Builder withPageWriteChecksumEnabled(boolean val) {
       return this;
     }
 
+    public Builder withColumnIdResolution(boolean columnIdResolution) {
```
rdblue, line 522: `resolveColumnsByID(boolean shouldResolveById)`?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
```diff
@@ -72,26 +73,50 @@ public static IntColumn intColumn(String columnPath) {
     return new IntColumn(ColumnPath.fromDotString(columnPath));
   }
 
+  public static IntColumn intColumn(Type.ID id) {
```
rdblue, line 76: @huaxingao and @shangxinli, has the Parquet community considered using Iceberg expressions and filters? I know that's a separate change, but it would be a great way to pick up a cleaner filter API that handles expression binding.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -48,10 +55,18 @@ protected Column(ColumnPath columnPath, Class<T> columnType) {
...
 
+    public Type.ID getColumnId() {
+      return columnId;
+    }
+
     public ColumnPath getColumnPath() {
       return columnPath;
     }
 
+    public void setColumnPath(ColumnPath columnPath) {
```
rdblue, line 66: Why does this add a setter for column path?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
```diff
@@ -170,6 +174,24 @@ public Void visit(Not not) {
...
     ColumnPath path = column.getColumnPath();
+    if (path == null) {
+      HashSet<Type.ID> ids = new HashSet<>();
+      Type.ID id = column.getColumnId();
+      List<ColumnDescriptor> columnDescriptors = new ArrayList<>(columnsAccordingToSchema.values());
+      for (ColumnDescriptor columnDescriptor : columnDescriptors) {
+        Type.ID columnId = columnDescriptor.getId();
+        if (columnId != null) {
+          if (ids.contains(columnId)) {
+            throw new RuntimeException("duplicate id");
```
rdblue, line 185: I doubt that this is the right place to catch duplicate column IDs.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/SchemaCompatibilityValidator.java
```diff
@@ -170,6 +174,24 @@ public Void visit(Not not) {
...
+          }
+          ids.add(columnId);
+          if (columnId.intValue() == id.intValue()) {
+            column.setColumnPath(ColumnPath.get(columnDescriptor.getPath()));
```
rdblue, line 189: I appreciate not wanting to change the validation logic, but making the filter `Column` mutable just to be able to add the path seems like a bad idea in general. What happens if the filter is reused and the path changes? It looks like this would use the first path for a column that was encountered.

I think this should validate directly using IDs instead.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,97 @@ public String getFile() {
...
+
+  public MessageType setRequestedSchema(MessageType projection, boolean useColumnId) {
     paths.clear();
-    for (ColumnDescriptor col : projection.getColumns()) {
+    MessageType schema = null;
+    if (useColumnId) {
+      HashSet<Type.ID> ids = new HashSet<>();
+      boolean fileSchemaIdUnique = uniqueId(fileMetaData.getSchema(), ids);
+      if (!fileSchemaIdUnique) {
+        throw new RuntimeException("can't use column id resolution because there are duplicate column ids.");
```
rdblue, line 918: Nit: error messages should use sentence case and begin with a capital letter.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,97 @@ public String getFile() {
     return blocks;
   }
 
-  public void setRequestedSchema(MessageType projection) {
+  private boolean uniqueId(GroupType schema, HashSet<Type.ID> ids) {
```
rdblue, line 883: I don't think it is a good practice to modify a set that's passed in. I would expect this to produce a set. If you want to throw an exception because this finds a duplicate ID, then I think it should just throw an exception in this method.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,97 @@ public String getFile() {
...
+    } else {
+      schema = projection;
+    }
+    for (ColumnDescriptor col : schema.getColumns()) {
       paths.put(ColumnPath.get(col.getPath()), col);
     }
+    return schema;
+  }
+
+  private MessageType resetColumnNameBasedOnId(MessageType schema) {
```
rdblue, line 935: This doesn't seem like the right class to contain these utility methods. I'd recommend a utility class to handle this.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -878,11 +880,97 @@ public String getFile() {
...
+        if (idInFileMetaData != null && id != null && idInFileMetaData.intValue() == id.intValue()) {
+          fieldName = c.getPrimitiveType().getName();
+        }
+      }
+      resetField = new PrimitiveType(field.getRepetition(), field.asPrimitiveType().getPrimitiveTypeName(), fieldName);
+    } else {
+      List<Type> childFields = ((GroupType) field).getFields();
+      List<Type> resetFields = resetColumnNameInFields(childFields);
+      if (resetFields.size() > 0) {
+        resetField = ((GroupType) field).withNewFields(resetFields);
```
rdblue, line 969: It doesn't look like this renames the field itself. So if you have a nested field with a top-level name change this wouldn't work.

rdblue (changes requested):
The approach here seems to be to allow IDs, but to translate back to field names for each file. I think that's going to be a problem for maintainability. It would be better to map column names to IDs and use IDs so you have consistency across all files.

shangxinli:
@rdblue regarding to 'using Iceberg expressions and filters', we agreed to use. We are finding resources to work on it.  Huaxin may be able to work on it after this column resolution by ID is done (didn't confirm with her yet).

iflytek-hmwang5:
@huaxingao  @shangxinli  Hi, how is the project going now? Any updates?

huaxingao:
@iflytek-hmwang5 
I think we don't need this PR any more. The purpose of this PR is to introduce page indexing in Iceberg. However, there's an alternative method for supporting page indexes in Iceberg, as outlined in this [PR](https://github.com/apache/iceberg/pull/6935)

emkornfield:
Per https://github.com/apache/parquet-java/pull/950#issuecomment-2017180916 going to close this as stale.

PR #951, opened 2022-03-09
PARQUET-2134: Fix type checking in HadoopStreams.wrap

7c00:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2134

### Tests

- [x] does not need testing for this extremely good reason:

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
@@ -19,16 +19,15 @@
 
 package org.apache.parquet.hadoop.util;
 
+import org.apache.hadoop.fs.ByteBufferReadable;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.parquet.io.ParquetDecodingException;
-import org.apache.parquet.io.SeekableInputStream;
 import org.apache.parquet.io.PositionOutputStream;
+import org.apache.parquet.io.SeekableInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
+import java.io.InputStream;
 import java.util.Objects;
 
 /**
@@ -37,9 +36,6 @@ import java.util.Objects;
 public class HadoopStreams {
 
   private static final Logger LOG = LoggerFactory.getLogger(HadoopStreams.class);
-
-  private static final Class<?> byteBufferReadableClass = getReadableClass();
-  static final Constructor<SeekableInputStream> h2SeekableConstructor = getH2SeekableConstructor();
 
   /**
    * Wraps a {@link FSDataInputStream} in a {@link SeekableInputStream}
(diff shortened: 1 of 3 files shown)


7c00:
Related issue: https://github.com/prestodb/presto/pull/17435

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -66,6 +67,15 @@ public static SeekableInputStream wrap(FSDataInputStream stream) {
     }
   }
 
+  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
+    InputStream wrapped = stream.getWrappedStream();
+    if (wrapped instanceof FSDataInputStream) {
+      return isWrappedStreamByteBufferReadable(((FSDataInputStream) wrapped));
```
shangxinli, line 73: Is there a corner case that can cause an infinite loop?

7c00: Yes, it could be. But it may be hard to create such a case. As its code shows, FSDataInputStream is a wrapper class of an inputstream. When we check the wrapped inputstream recursively, it would finally reach an inputstream whose type is not FSDataInputStream. A developer could override `getWrappedStream` as `return this`, while this makes no sence.

shangxinli: I understand it would be very rare case but once that happen it would be hard to debug this 'hang' issue. Let's do two things: 1) Add check if it is 'this'; Throw exception if that happens; 2) Add debug log; When it hangs, developer can enable debug log and see what parquet-mr is doing.

7c00: Good suggestions! I have updated as the comment.

shangxinli:
Thanks for adding the check and debug log. LGTM! One more thing(sorry for not asking at first-round review), do you think it makes sense to add tests?

7c00:
> @7c00 Do you have time to look into the last feedback?

@shangxinli Thanks for your comments. Sorry for late to reply. I think it's ok to add some unit tests, and I am going to do that in this weekend. I will ping you when finished. Thank you in advance.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
steveloughran: you can't. the inner stream is set in the constructor, so cannot take the not-yet-constructed class as an argument.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -51,7 +52,7 @@ public class HadoopStreams {
   public static SeekableInputStream wrap(FSDataInputStream stream) {
     Objects.requireNonNull(stream, "Cannot wrap a null input stream");
     if (byteBufferReadableClass != null && h2SeekableConstructor != null &&
-        byteBufferReadableClass.isInstance(stream.getWrappedStream())) {
+        isWrappedStreamByteBufferReadable(stream)) {
```
steveloughran, line 55: this is really going into the internals of the hadoop classes and potentially tricky if there is any dynamic decision making in the inner class. The good news there is I don't see anything doing that.

there is a way to ask (hadoop 3.2+) if a stream does support the API before calling, using the StreamCapabilities interface.
https://issues.apache.org/jira/browse/HDFS-14111

```java
if (stream.hasCapability( "in:readbytebuffer") {
  // stream is confident it has the api
) else {
  // do the checking of the inner class
}

**Review by steveloughran — COMMENTED** [review]
This would be a lot easier if HadoopStreams didn't use reflection to get at ByteBufferReadable. that's an api which came with hadoop 2.0.2, so *everything has it.

File: parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java [hunk 3]
```diff
@@ -66,6 +67,19 @@ public static SeekableInputStream wrap(FSDataInputStream stream) {
     }
   }
 
+  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
+    InputStream wrapped = stream.getWrappedStream();
+    if (wrapped == stream) {
+      throw new ParquetDecodingException("Illegal FSDataInputStream as wrapped itself");
```
steveloughran, line 73: this can't happen. the inner stream is set in the constructor, so cannot take the not-yet-constructed class as an argument...no need to worry about recursion.

steveloughran:
I've taken this PR and added the changes I was suggesting, plus tests. see #971. If you take that extra commit and merge it in here, it should complete this PR

shangxinli:
> I've taken this PR and added the changes I was suggesting, plus tests. see #971. If you take that extra commit and merge it in here, it should complete this PR

@7c00 Are you OK since you are originally created this PR?

7c00:
Thanks @steveloughran @shangxinli . I have cherry-picked the commit from https://github.com/apache/parquet-mr/pull/971

steveloughran:
whoever actually commits this can use the github squash option to combine all commits into one before merging.

FYI, I've just started writing a shim library so that apps compiling against hadoop 3.2.0 wil be able to invoke the 3.3+ API calls when present: [HADOOP-18287](https://issues.apache.org/jira/browse/HADOOP-18287). 

First parquet will need to be able to compile/link against hadoop 3.x: #976 976

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -50,51 +46,45 @@ public class HadoopStreams {
...
+   * In hadoop 3.3.0+, the StreamCapabilities probe can be used to
+   * check this: only those streams which provide the read(ByteBuffer)
+   * semantics MAY return true for the probe "in:readbytebuffer";
+   * FSDataInputStream will pass the probe down to the underlying stream.
+   *
+   * @param stream stream to probe
+   * @return true if it is safe to a H2SeekableInputStream to access the data
+   */
+  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
+    if (stream.hasCapability("in:readbytebuffer")) {
```
shangxinli, line 74: We don't have the Hadoop 3..3.0 yet in Parquet. Does it mean we need to hold of this PR?

steveloughran: no, the StreamCapabilities probe has been around since hadoop 2. it is just in 3.3.0 all streams which implement the api return true for this probe...a probe which gets passed down the wrapped streams. It avoids looking at the wrapped streams as you should be able to trust the response (put differently: if something lied it is in trouble)

shangxinli:
@7c00 and @steveloughran Thank both of you for the great contribution! This PR comes from two authors. Can @7c00 add @steveloughran as the co-author to this PR? [This](https://github.blog/2018-01-29-commit-together-with-co-authors/) is an example.

7c00:
@shangxinli Thank you for reminding me. I have squashed the PR and added @steveloughran as the co-author.

steveloughran:
thanks. created [HADOOP-18336](https://issues.apache.org/jira/browse/HADOOP-18336)
tag FSDataInputStream.getWrappedStream() @Public/@Stable  to make sure that hadoop code knows external libs may be calling the method.

shangxinli:
LGTM

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
sunchao: @steveloughran @shangxinli it looks like the API is not available in Hadoop 2.8.x, so it will create issues for projects that want to use the latest version of Parquet but still want to keep Hadoop 2.8.x.

sunchao: also see related JIRA: https://issues.apache.org/jira/browse/PARQUET-2276

PR #953, opened 2022-03-29
Performance optimizations:  Merged all LittleEndianDataInputStream functionality into ByteBufferInputStream

theosib-amazon:
This PR is all performance optimization. In benchmarking with Trino, we find query performance to improve from 5% to 15%, depending on the query, and that includes all the I/O time from S3.

The main modification is to merge all of LittleEndianDataInputStream functionality into ByteBufferInputStream, which yields the following benefits:
- Elimination of extra layers of abstraction and method call overhead
- Enable the use of intrinsics for readInt, readLong, etc.
- Availability of faster access methods like readFully and skipFully, without the need for helper functions
- Reduces some object creation in the performance critical path

This also includes and enables performance optimizations to:
- ByteBitPackingValuesReader
- PlainValuesReader
- RunLengthBitPackingHybridDecoder

Context:
I've been working on improving Parquet reading performance in Trino, mostly by profiling while running performance benchmarks and TPCDS queries. This PR is a subset of the changes I made that have more than doubled the performance of a lot of TPCDS queries (wall clock time, including the S3 access time). If you are kind enough to accept these changes, I have more I would like to contribute.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPackingValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPackingValuesReader.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -20,6 +20,7 @@ package org.apache.parquet.column.values.bitpacking;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 
 import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.bytes.BytesUtils;
@@ -38,31 +39,37 @@ public class ByteBitPackingValuesReader extends ValuesReader {
   private final int[] decoded = new int[VALUES_AT_A_TIME];
   private int decodedPosition = VALUES_AT_A_TIME - 1;
   private ByteBufferInputStream in;
+  byte[] tempEncode;
 
   public ByteBitPackingValuesReader(int bound, Packer packer) {
     this.bitWidth = BytesUtils.getWidthFromMaxInt(bound);
     this.packer = packer.newBytePacker(bitWidth);
+    // Keep reusing this buffer to eliminate object creation in the critical path
+    tempEncode = new byte[bitWidth];
+  }
+
+  private void readMore() {
+    try {
+      // This eliminates the use of slice(), which is slow because of object creation in the critical path.
+      int avail = in.available();
+      if (avail < bitWidth) {
+        in.read(tempEncode, 0, avail);
+        Arrays.fill(tempEncode, avail, bitWidth, (byte)0);
+      } else {
+        in.read(tempEncode, 0, bitWidth);
+      }
+      packer.unpack8Values(tempEncode, 0, decoded, 0);
+    } catch (IOException e) {
+      throw new ParquetDecodingException("Failed to read packed values", e);
+    }
+    decodedPosition = 0;
   }
 
   @Override
   public int readInteger() {
     ++ decodedPosition;
     if (decodedPosition == decoded.length) {
-      try {
-        if (in.available() < bitWidth) {
-          // unpack8Values needs at least bitWidth bytes to read from,
-          // We have to fill in 0 byte at the end of encoded bytes.
-          byte[] tempEncode = new byte[bitWidth];
-          in.read(tempEncode, 0, in.available());
-          packer.unpack8Values(tempEncode, 0, decoded, 0);
-        } else {
-          ByteBuffer encoded = in.slice(bitWidth);
-          packer.unpack8Values(encoded, encoded.position(), decoded, 0);
-        }
-      } catch (IOException e) {
-        throw new ParquetDecodingException("Failed to read packed values", e);
-      }
-      decodedPosition = 0;
+      readMore();
     }
     return decoded[decodedPosition];
   }
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
@@ -19,6 +19,8 @@
 
 package org.apache.parquet.bytes;
 
+import org.apache.parquet.ShouldNeverHappenException;
+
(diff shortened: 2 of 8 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPackingValuesReader.java
```diff
@@ -1,14 +1,14 @@
...
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
  * regarding copyright ownership.  The ASF licenses this file
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- *
```
rdblue, line 11: Please remove all non-functional changes, like these that modify whitespace. While I like having cleaner code, we don't want to introduce unnecessary commit conflicts.

theosib-amazon: This was an annoying side-effect of using the IDE. Is there any convenient way to change this on github, or do I have to do the whole thing all over again?

theosib-amazon: I think I got them all.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -46,12 +74,19 @@ public long position() {
...
+  has any remaining bytes.
+   */
+
   @Override
   public int read() throws IOException {
-    if (!buffer.hasRemaining()) {
-    	throw new EOFException();
+    try {
+      return buffer.get() & 255;
+    } catch (Exception e) {
```
rdblue, line 87: This needs to be more specific.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -46,12 +74,19 @@ public long position() {
...
+  we catch and turn into an EOFException. This is measured to be faster than explicitly checking if the ByteBuffer
+  has any remaining bytes.
+   */
+
   @Override
   public int read() throws IOException {
-    if (!buffer.hasRemaining()) {
-    	throw new EOFException();
+    try {
+      return buffer.get() & 255;
```
rdblue, line 86: Can you revert the change to the constant?

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -88,6 +132,15 @@ public long skip(long n) {
     return bytesToSkip;
   }
 
+  @Override
+  public void skipFully(long n) throws IOException {
+    try {
+      buffer.position(buffer.position() + (int)n);
```
rdblue, line 138: Looks like this is just trying to avoid the checks that are being done in `skip`. I don't think that's a good idea. This should delegate to skip instead.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -174,4 +248,63 @@ public boolean markSupported() {
   public int available() {
     return buffer.remaining();
   }
+
+  @Override
+  public byte readByte() throws IOException {
```
rdblue, line 253: The changes from here on out look like what you're really trying to do because we want to read directly from the stream. Can you remove all the other changes that aren't needed?

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -203,11 +217,22 @@ public List<ByteBuffer> remainingBuffers() {
       return sliceBuffers(length - position);
     } catch (EOFException e) {
       throw new RuntimeException(
-          "[Parquet bug] Stream is bad: incorrect bytes remaining " +
-              (length - position));
+        "[Parquet bug] Stream is bad: incorrect bytes remaining " +
+          (length - position));
```
rdblue, line 221: Non-functional change.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -248,13 +295,17 @@ public int read() throws IOException {
       throw new EOFException();
     }
 
+    this.position++;
```
rdblue, line 298: Can you revert the unnecessary change here? This was `this.position += 1;` and I don't see a reason to change it more than necessary.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -379,4 +432,117 @@ public void remove() {
...
+    long ch7 = (long)readUnsignedByte() << 56;
+    return ((ch0 + ch1) + (ch2 + ch3)) + ((ch4 + ch5) + (ch6 + ch7));
+  }
+
+  @Override
+  public long readLong() throws IOException {
+    if (current == null) {
+      throw new EOFException();
+    }
+    if (current.remaining() >= 8) {
```
rdblue, line 539: There should be a newline between a control flow block and the next statement.

rdblue (changes requested):
Overall I think most of these changes are good, but there are a few things that should be done before committing this:
* Revert any unnecessary changes, like new constructors and style changes that are non-functional (e.g. using `x++` instead of `x += 1`)
* Separate the `ByteBufferInputStream` additions into a dedicated PR with tests
* Make real changes to `PlainValuesReader` rather than keeping both input streams and changing the reference to `in2`
* Update for project style

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
theosib-amazon: I did this specifically because it's performance-critical. I did a bunch of profiling, and skips are among the operations that have to have minimal overhead. Delegating to skip() would introduce a bunch of checks that the JIT isn't going to be smart enough to remove.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
theosib-amazon: I'm not sure what you're referring to. All of the methods beyond this point are absolutely necessary. We need to be able to read ints and longs and such from the bytebuffer, and this is the only way to get them.

theosib-amazon:
I made all the cosmetic changes you asked for.

I'm not sure why you're asking to separate the ByteBufferInputStream additions into its own PR, since the PR was all about improving performance by moving functionality from LittleEndianDataInputStream into ByteBufferInputStream. The changes to PlainValuesReader rely on all of those changes.

The only reason I kept the reference to LittleEndianDataInputStream in PlainValuesReader is because otherwise the build fails with a compatibility break against 1.12.0. I'm going to go ahead with the change in the hopes that that doesn't cause a check failure.

theosib-amazon:
Also, you mentioned tests. Since I'm not making any functional changes, I'm not sure what to test for. The new code should behave exactly as the old version, just a bit faster.

PR #958, opened 2022-04-19
PARQUET-2138: Add ShowBloomFilterCommand to parquet-cli

WangGuangxin:
Add `ShowBloomFilter` command to parquet-cli. We can leverage it to check whether a block can be filtered by bloom filter.

Example usage:
```
parquet-cli bloom-filter -c test_column -v 1226545,3940309
```

output
```
Row group 0:
--------------------------------------------------------------------------------
value 1226545 maybe exists.
value 3940309 NOT exists.

Row group 1:
--------------------------------------------------------------------------------
value 1226545 NOT exists.
value 3940309 NOT exists.
```

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
@@ -49,7 +49,7 @@ public class ShowBloomFilterCommand extends BaseCommand {
   }
 
   @Parameter(description = "<parquet path>")
-  List<String> files;
+  String file;
 
   @Parameter(
     names = { "-c", "--column" },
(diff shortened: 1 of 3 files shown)


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
```diff
@@ -0,0 +1,135 @@
...
+    description = "Check if the given values match bloom filter",
+    required = true)
+  List<String> testValues;
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public int run() throws IOException {
+    Preconditions.checkArgument(files != null && files.size() >= 1,
+      "A Parquet file is required.");
+    Preconditions.checkArgument(files.size() == 1,
```
shangxinli, line 71: Can we merge it into the above?

shangxinli:
For column encrypted files, does it work? Can you add tests for it? For how to create encrypted test file, you can refer to this file parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/ColumnEncryptorTest.java.

shangxinli:
Yeah, I guess you need to pass the key directly or KMS URL.

shangxinli:
@WangGuangxin Do you still plan to implement the decryption? I don't want to place a blocker if you don't have plan for it.

shangxinli:
Let's merge it now and we can add column decryption later.

PR #960, opened 2022-04-25
Performance optimization: Move all LittleEndianDataInputStream functionality into ByteBufferInputStream

theosib-amazon:
I broke up https://github.com/apache/parquet-mr/pull/953 into more digestible pieces. This new PR is the lowest level set of changes. By themselves, these additions to ByteBufferInputStream don't yield much improvement, so future PRs will include modifications to other source files that take advantage of this new functionality.

The complete set of changes (including subsequent PRs) is for performance optimization. In benchmarking with Trino, we find query performance to improve from 5% to 15%, depending on the query, and that includes all the I/O time from S3.

All of LittleEndianDataInputStream functionality is moved into ByteBufferInputStream, without changing any pre-existing interfaces or functionality. These changes yield the following benefits:
- Elimination of extra layers of abstraction and method call overhead
- Enable the use of intrinsics for readInt, readLong, etc.
- Availability of faster access methods like readFully and skipFully, without the need for helper functions

This PR also marks LittleEndianDataInputStream as deprecated.

Context:
I've been working on improving Parquet reading performance in Trino, mostly by profiling while running performance benchmarks and TPCDS queries. This PR is a subset of the changes I made that have more than doubled the performance of a lot of TPCDS queries (wall clock time, including the S3 access time). If you are kind enough to accept these changes, I look forward to offering further performance improvements.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -49,15 +49,6 @@ public class ByteBufferInputStream extends InputStream {
     }
   }
 
-  public static ByteBufferInputStream wrap(byte[] buf) {
-    return new SingleBufferInputStream(buf);
-  }
-
-  public static ByteBufferInputStream wrap(byte[] buf, int start, int length) {
-    return new SingleBufferInputStream(buf, start, length);
-  }
-
-
   ByteBufferInputStream() {
     delegate = null;
   }
@@ -83,26 +74,11 @@ public class ByteBufferInputStream extends InputStream {
    */
   @Deprecated
   public ByteBufferInputStream(ByteBuffer buffer, int offset, int count) {
-    // This is necessary to pass "TestDeprecatedBufferInputStream"...
     ByteBuffer temp = buffer.duplicate();
     temp.position(offset);
     ByteBuffer byteBuf = temp.slice();
     byteBuf.limit(count);
     delegate = wrap(byteBuf);
-    // ... but it would probably be faster to do this:
-//    delegate = wrap(buffer, offset, count);
-  }
-
-  public ByteBufferInputStream(byte[] inBuf) {
-    delegate = wrap(inBuf);
-  }
-
-  public ByteBufferInputStream(byte[] inBuf, int start, int length) {
-    delegate = wrap(inBuf, start, length);
-  }
-
-  public ByteBufferInputStream(List<ByteBuffer> inBufs) {
-    delegate = wrap(inBufs);
   }
 
   /**
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -28,16 +28,6 @@ import java.util.List;
 
 import org.apache.parquet.ShouldNeverHappenException;
 
-/*
-Changes implemented:
-All of the functionality of LittleEndianDataInputStream has been merged into ByteBufferInputStream and its child
-classes. This has resulted in measurable performance improvements for the following reasons:
-- Elimination of at least one layer of abstraction / method call overhead
-- Enabling support for intrinsics for readInt, readLong, etc.
-- Eliminate the need for the JIT to make inferences that may or may not inline methods from BytesUtils and
-  the InputStream.read() that is called by BytesUtils.
- */
-
 public class ByteBufferInputStream extends InputStream {
 
   // Used to maintain the deprecated behavior of instantiating ByteBufferInputStream directly
@@ -57,10 +47,6 @@ public class ByteBufferInputStream extends InputStream {
     } else {
       return new MultiBufferInputStream(buffers);
     }
-  }
-
-  public static ByteBufferInputStream wrap(ByteBuffer buffer, int offset, int count) {
-    return new SingleBufferInputStream(buffer, offset, count);
   }
 
   public static ByteBufferInputStream wrap(byte[] buf) {
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -28,6 +28,16 @@ import java.util.List;
 
 import org.apache.parquet.ShouldNeverHappenException;
 
+/*
+Changes implemented:
+All of the functionality of LittleEndianDataInputStream has been merged into ByteBufferInputStream and its child
+classes. This has resulted in measurable performance improvements for the following reasons:
+- Elimination of at least one layer of abstraction / method call overhead
+- Enabling support for intrinsics for readInt, readLong, etc.
+- Eliminate the need for the JIT to make inferences that may or may not inline methods from BytesUtils and
+  the InputStream.read() that is called by BytesUtils.
+ */
+
 public class ByteBufferInputStream extends InputStream {
 
   // Used to maintain the deprecated behavior of instantiating ByteBufferInputStream directly
@@ -48,6 +58,19 @@ public class ByteBufferInputStream extends InputStream {
       return new MultiBufferInputStream(buffers);
     }
   }
+
+  public static ByteBufferInputStream wrap(ByteBuffer buffer, int offset, int count) {
+    return new SingleBufferInputStream(buffer, offset, count);
+  }
+
+  public static ByteBufferInputStream wrap(byte[] buf) {
+    return new SingleBufferInputStream(buf);
+  }
+
+  public static ByteBufferInputStream wrap(byte[] buf, int start, int length) {
+    return new SingleBufferInputStream(buf, start, length);
+  }
+
 
   ByteBufferInputStream() {
     delegate = null;
@@ -74,11 +97,26 @@ public class ByteBufferInputStream extends InputStream {
    */
   @Deprecated
   public ByteBufferInputStream(ByteBuffer buffer, int offset, int count) {
+    // This is necessary to pass "TestDeprecatedBufferInputStream"...
     ByteBuffer temp = buffer.duplicate();
     temp.position(offset);
     ByteBuffer byteBuf = temp.slice();
     byteBuf.limit(count);
     delegate = wrap(byteBuf);
+    // ... but it would probably be faster to do this:
+//    delegate = wrap(buffer, offset, count);
+  }
+
+  public ByteBufferInputStream(byte[] inBuf) {
+    delegate = wrap(inBuf);
+  }
+
+  public ByteBufferInputStream(byte[] inBuf, int start, int length) {
+    delegate = wrap(inBuf, start, length);
+  }
+
+  public ByteBufferInputStream(List<ByteBuffer> inBufs) {
+    delegate = wrap(inBufs);
   }
 
   /**
@@ -98,12 +136,12 @@ public class ByteBufferInputStream extends InputStream {
     return delegate.position();
   }
 
+  public void position(int pos) {
+    throw new UnsupportedOperationException();
+  }
+
   public void skipFully(long n) throws IOException {
-    long skipped = skip(n);
-    if (skipped < n) {
-      throw new EOFException(
-          "Not enough bytes to skip: " + skipped + " < " + n);
-    }
+    delegate.skipFully(n);
   }
 
   public int read(ByteBuffer out) {
@@ -119,7 +157,8 @@ public class ByteBufferInputStream extends InputStream {
   }
 
   public ByteBufferInputStream sliceStream(long length) throws EOFException {
-    return ByteBufferInputStream.wrap(sliceBuffers(length));
+    return delegate.sliceStream(length);
+    //return ByteBufferInputStream.wrap(sliceBuffers(length));
   }
 
   public List<ByteBuffer> remainingBuffers() {
@@ -127,7 +166,11 @@ public class ByteBufferInputStream extends InputStream {
   }
 
   public ByteBufferInputStream remainingStream() {
-    return ByteBufferInputStream.wrap(remainingBuffers());
+    return delegate.remainingStream();
+  }
+
+  public ByteBufferInputStream duplicate() {
+    return delegate.duplicate();
   }
 
   public int read() throws IOException {
@@ -138,12 +181,32 @@ public class ByteBufferInputStream extends InputStream {
     return delegate.read(b, off, len);
   }
 
+  public int read(byte[] b) throws IOException {
+    return read(b, 0, b.length);
+  }
+
+  public void readFully(byte b[]) throws IOException {
+    delegate.readFully(b, 0, b.length);
+  }
+
+  public void readFully(byte b[], int off, int len) throws IOException {
+    delegate.readFully(b, off, len);
+  }
+
   public long skip(long n) {
     return delegate.skip(n);
   }
 
+  public int skipBytes(int n) {
+    return (int)skip(n);
+  }
+
   public int available() {
     return delegate.available();
+  }
+
+  public int remaining() {
+    return available();
   }
 
   public void mark(int readlimit) {
@@ -157,4 +220,83 @@ public class ByteBufferInputStream extends InputStream {
   public boolean markSupported() {
     return delegate.markSupported();
   }
+
+  public void close() throws IOException {
+  }
+
+  public boolean readBoolean() throws IOException {
+    return readByte() != 0;
+  }
+
+  public byte readByte() throws IOException {
+    return delegate.readByte();
+  }
+
+  public int readUnsignedByte() throws IOException {
+    return delegate.readUnsignedByte();
+  }
+
+  public short readShort() throws IOException {
+    return delegate.readShort();
+  }
+
+  public int readUnsignedShort() throws IOException {
+    return delegate.readUnsignedShort();
+  }
+
+  public int readInt() throws IOException {
+    return delegate.readInt();
+  }
+
+  public long readLong() throws IOException {
+    return delegate.readLong();
+  }
+
+  public float readFloat() throws IOException {
+    return Float.intBitsToFloat(readInt());
+  }
+
+  public double readDouble() throws IOException {
+    return Double.longBitsToDouble(readLong());
+  }
+
+  public int readIntLittleEndianOnThreeBytes() throws IOException {
+    int ch1 = readUnsignedByte();
+    int ch2 = readUnsignedByte();
+    int ch3 = readUnsignedByte();
+    return ((ch3 << 16) + (ch2 << 8) + (ch1 << 0));
+  }
+
+  public int readIntLittleEndianPaddedOnBitWidth(int bitWidth)
+    throws IOException {
+
+    int bytesWidth = BytesUtils.paddedByteCountFromBits(bitWidth);
+    switch (bytesWidth) {
+      case 0:
+        return 0;
+      case 1:
+        return readUnsignedByte();
+      case 2:
+        return readUnsignedShort();
+      case 3:
+        return readIntLittleEndianOnThreeBytes();
+      case 4:
+        return readInt();
+      default:
+        throw new IOException(
+          String.format("Encountered bitWidth (%d) that requires more than 4 bytes", bitWidth));
+    }
+  }
+
+  public int readUnsignedVarInt() throws IOException {
+    int value = 0;
+    int i = 0;
+    int b;
+    while (((b = readUnsignedByte()) & 0x80) != 0) {
+      value |= (b & 0x7F) << i;
+      i += 7;
+    }
+    return value | (b << i);
+  }
+
 }
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -138,7 +138,7 @@ public class ByteBufferInputStream extends InputStream {
     return read(b, 0, b.length);
   }
 
-  public void readFully(byte b[]) throws IOException {
+  public void readFully(byte[] b) throws IOException {
     readFully(b, 0, b.length);
   }
 
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -186,7 +186,7 @@ public class ByteBufferInputStream extends InputStream {
   }
 
   public void readFully(byte b[]) throws IOException {
-    delegate.readFully(b, 0, b.length);
+    readFully(b, 0, b.length);
   }
 
   public void readFully(byte b[], int off, int len) throws IOException {
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -120,7 +120,6 @@ public class ByteBufferInputStream extends InputStream {
 
   public ByteBufferInputStream sliceStream(long length) throws EOFException {
     return delegate.sliceStream(length);
-    //return ByteBufferInputStream.wrap(sliceBuffers(length));
   }
 
   public List<ByteBuffer> remainingBuffers() {
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
@@ -98,10 +98,6 @@ public class ByteBufferInputStream extends InputStream {
     return delegate.position();
   }
 
-  public void position(int pos) {
-    throw new UnsupportedOperationException();
-  }
-
   public void skipFully(long n) throws IOException {
     delegate.skipFully(n);
   }
@@ -130,10 +126,6 @@ public class ByteBufferInputStream extends InputStream {
     return delegate.remainingStream();
   }
 
-  public ByteBufferInputStream duplicate() {
-    return delegate.duplicate();
-  }
-
   public int read() throws IOException {
     return delegate.read();
   }
@@ -158,16 +150,8 @@ public class ByteBufferInputStream extends InputStream {
     return delegate.skip(n);
   }
 
-  public int skipBytes(int n) {
-    return (int)skip(n);
-  }
-
   public int available() {
     return delegate.available();
-  }
-
-  public int remaining() {
-    return available();
   }
 
   public void mark(int readlimit) {
@@ -180,9 +164,6 @@ public class ByteBufferInputStream extends InputStream {
 
   public boolean markSupported() {
     return delegate.markSupported();
-  }
-
-  public void close() throws IOException {
   }
 
   public boolean readBoolean() throws IOException {
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
@@ -94,7 +94,7 @@ class MultiBufferInputStream extends ByteBufferInputStream {
   @Override
   public void skipFully(long n) throws IOException {
     if (current == null || n > length) {
-      throw new EOFException();
+      throw new EOFException("Not enough bytes to skip: " + length + " < " + n));
     }
     
     skip(n);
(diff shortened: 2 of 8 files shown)


theosib-amazon:
I added some tests for this new code.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -28,6 +28,16 @@
 
 import org.apache.parquet.ShouldNeverHappenException;
 
+/*
+Changes implemented:
```
rdblue, line 32: Can you remove this? Generally, we want PRs that can be merged. You can add comments about what is changing in the PR rather than in the code.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -49,6 +59,19 @@ public static ByteBufferInputStream wrap(List<ByteBuffer> buffers) {
     }
   }
 
+  public static ByteBufferInputStream wrap(ByteBuffer buffer, int offset, int count) {
```
rdblue, line 62: What is the purpose of these methods? We already wrap buffers, why not have the caller handle this?

It's also really strange to pass a buffer with an offset and a "count" (do you mean length?) because buffers already have that information built in.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -74,11 +97,26 @@ public ByteBufferInputStream(ByteBuffer buffer) {
...
   @Deprecated
   public ByteBufferInputStream(ByteBuffer buffer, int offset, int count) {
+    // This is necessary to pass "TestDeprecatedBufferInputStream"...
     ByteBuffer temp = buffer.duplicate();
     temp.position(offset);
     ByteBuffer byteBuf = temp.slice();
     byteBuf.limit(count);
     delegate = wrap(byteBuf);
+    // ... but it would probably be faster to do this:
+//    delegate = wrap(buffer, offset, count);
```
rdblue, line 107: Please remove notes to self.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: Removed

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: This was to make ByteBufferInputStream resemble ByteBuffer. ByteBuffer has methods like this. I've gone ahead and removed them at your request.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: Done.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -274,6 +329,7 @@ public void mark(int readlimit) {
     if (mark >= 0) {
       discardMark();
     }
+
```
rdblue, line 332: Nit: unnecessary whitespace change.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -157,4 +181,83 @@ public void reset() throws IOException {
   public boolean markSupported() {
     return delegate.markSupported();
   }
+
+  public void close() throws IOException {
```
rdblue, line 185: Close delegate?

shangxinli:
Is the '5% to 15%' gain from this change or along with other changes? If it is later, can you share the point to other changes? Like to see the overall changes before committing.

theosib-amazon:
That improvement comes from a larget set of changes. I have a design doc that goes over all those changes plus some more that make it possible to get even more performance improvements.

https://docs.google.com/document/d/1fBGpF_LgtfaeHnPD5CFEIpA2Ga_lTITmFdFIcO9Af-g/edit?usp=sharing

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -157,4 +165,80 @@ public void reset() throws IOException {
...
+        return readIntLittleEndianOnThreeBytes();
+      case 4:
+        return readInt();
+      default:
+        throw new IOException(
+          String.format("Encountered bitWidth (%d) that requires more than 4 bytes", bitWidth));
+    }
+  }
+
+  public int readUnsignedVarInt() throws IOException {
```
shangxinli, line 233: Is it copied from ByteUnites? I wonder why we don't use that directly?>

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -157,4 +165,80 @@ public void reset() throws IOException {
...
+  }
+
+  public int readIntLittleEndianOnThreeBytes() throws IOException {
+    int ch1 = readUnsignedByte();
+    int ch2 = readUnsignedByte();
+    int ch3 = readUnsignedByte();
+    return ((ch3 << 16) + (ch2 << 8) + (ch1 << 0));
+  }
+
+  public int readIntLittleEndianPaddedOnBitWidth(int bitWidth)
```
shangxinli, line 212: Is it copied from ByteUnites? I wonder why we don't use that directly?

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -157,4 +165,80 @@ public void reset() throws IOException {
...
+
+  public float readFloat() throws IOException {
+    return Float.intBitsToFloat(readInt());
+  }
+
+  public double readDouble() throws IOException {
+    return Double.longBitsToDouble(readLong());
+  }
+
+  public int readIntLittleEndianOnThreeBytes() throws IOException {
```
shangxinli, line 205: Is it copied from BytesUtils.java? I wonder why we don't use that directly?

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -238,8 +257,31 @@ public int read(byte[] bytes, int off, int len) {
   }
 
   @Override
-  public int read(byte[] bytes) {
-    return read(bytes, 0, bytes.length);
+  public void readFully(byte[] bytes, int off, int len) throws IOException {
```
shangxinli, line 260: Can you cast a light why we need to add the implementation readFully() here? For performance improvement?

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -238,8 +257,31 @@ public int read(byte[] bytes, int off, int len) {
...
+      
+      return;
+    }
+
+    if (current == null || len > length) {
+      throw new EOFException();
+    }
+
+    int bytesRead = 0;
+    while (bytesRead < len) {
```
shangxinli, line 274: This seems duplicate with above line 244.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -379,4 +427,120 @@ public void remove() {
...
+
+  @Override
+  public long readLong() throws IOException {
+    if (current == null) {
+      throw new EOFException();
+    }
+    
+    if (current.remaining() >= 8) {
+      // If the whole short can be read from the current buffer, use intrinsics
+      this.position += 8;
```
shangxinli, line 539: USE Long.BYTES instead of 8

shangxinli:
@sunchao Can you have a review?

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: Not exactly. The one in BytesUtils calls methods that read one byte at a time. This one can take advantage of faster methods that read whole words at a time. This is a critical-path method, so it's a performance win to eliminate the extra level of abstraction and all the extra overhead fetching individual bytes and shifting.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: See my other comments.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: There are situations where we need to read an exact number of bytes and throw an exception if not enough are available. This is faster than reading maybe enough and then checking, and this is a performance critical path.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: There are two key differences that make it hard to combine them without hurting performance for one, the other, or both, and they're both performance critical.

sunchao:
Is this mostly a refactoring PR? I also don't see `LittleEndianDataInputStream` being marked as deprecated.

theosib-amazon:
> Is this mostly a refactoring PR? I also don't see `LittleEndianDataInputStream` being marked as deprecated.

I initially marked `LittleEndianDataInputStream` as deprecated. But I seem to recall that I was advised by Ryan Blue to do that in a later PR.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -89,6 +91,15 @@ public long skip(long n) {
     return bytesSkipped;
   }
 
+  @Override
+  public void skipFully(long n) throws IOException {
+    if (current == null || n > length) {
+      throw new EOFException();
```
sunchao, line 97: ditto: some info in exception, better retain the original message: `Not enough bytes to skip: " + skipped + " < " + n)`

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
sunchao: Why can't we just track the remaining bytes of the stream on the client side and check before reading any bytes?

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -379,4 +427,120 @@ public void remove() {
...
     }
   }
+
+  @Override
+  public byte readByte() throws IOException {
+    return (byte)readUnsignedByte();
+  }
+
+  @Override
+  public int readUnsignedByte() throws IOException {
```
sunchao, line 437: this looks the same as `read`

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
```diff
@@ -379,4 +427,120 @@ public void remove() {
       second.remove();
     }
   }
+
+  @Override
+  public byte readByte() throws IOException {
+    return (byte)readUnsignedByte();
```
sunchao, line 433: nit: can we add a space after the cast `(byte) readUnsignedByte()`? for all the other places too

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -88,6 +136,21 @@ public long skip(long n) {
...
 
+  @Override
+  public void skipFully(long n) throws IOException {
+    if (n < 0 || n > Integer.MAX_VALUE) {
+      throw new IllegalArgumentException();
+    }
+    
+    try {
+      buffer.position(buffer.position() + (int)n);
+    } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
theosib-amazon: I did it this way on purpose. This way is always faster. Doing the check has to wait until there's profiling info and the C2 compiler gets hold of this code.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -38,6 +39,34 @@ class SingleBufferInputStream extends ByteBufferInputStream {
...
+  SingleBufferInputStream(ByteBuffer buffer, int start, int length) {
+    // duplicate the buffer because its state will be modified
+    this.buffer = buffer.duplicate();
+    this.startPosition = start;
+    this.buffer.position(start);
+    this.buffer.limit(start + length);
+    this.buffer.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+  }
+
+  SingleBufferInputStream(byte[] inBuf) {
```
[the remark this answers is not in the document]
theosib-amazon: Yeah. I put this here I think because I'm trying to make BufferInputStream have the same functionality as multiple other things that I'm combining into one. Some tests would be good. I'll see about writing some tests soon.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -38,6 +39,34 @@ class SingleBufferInputStream extends ByteBufferInputStream {
...
+
+  SingleBufferInputStream(byte[] inBuf, int start, int length) {
+    this.buffer = ByteBuffer.wrap(inBuf, start, length);
+    this.buffer.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+    // This seems to be consistent with HeapByteBuffer.wrap(), which leaves
+    // the internal "offset" at zero and sets the starting position at start.
+    this.startPosition = 0;
+  }
+
+  SingleBufferInputStream(List<ByteBuffer> inBufs) {
```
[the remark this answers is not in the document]
theosib-amazon: IIRC, I used to have a similar constructor for ByteBufferInputStream, but I was advised to remove it, so I had this here for uniformity. We can remove this.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: Normally, the user of the class would read exactly the right number of bytes. These checks and exceptions exist only to catch bugs elsewhere. This is one reason why it's important to minimize the overhead of these checks in such performance-critical methods.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: You're right. They're identical. I think the C1 compiler will eliminate the overhead of read() calling readUnsignedByte().

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -70,9 +105,22 @@ public int read(byte[] bytes, int offset, int length) throws IOException {
 
     return bytesToRead;
   }
-  
+
+  @Override
+  public void readFully(byte[] bytes, int offset, int length) throws IOException {
+    try {
+      buffer.get(bytes, offset, length);
+    } catch (BufferUnderflowException|IndexOutOfBoundsException e) {
```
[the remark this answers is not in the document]
theosib-amazon: Got it.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -88,6 +136,21 @@ public long skip(long n) {
     return bytesToSkip;
   }
 
+  @Override
+  public void skipFully(long n) throws IOException {
+    if (n < 0 || n > Integer.MAX_VALUE) {
+      throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
theosib-amazon: got it

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -70,9 +105,22 @@ public int read(byte[] bytes, int offset, int length) throws IOException {
...
+      buffer.get(bytes, offset, length);
+    } catch (BufferUnderflowException|IndexOutOfBoundsException e) {
+      throw new EOFException(e.getMessage());
+    }
+  }
+
   @Override
   public long skip(long n) {
+    if (n < 0) {
+      throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
theosib-amazon: got it

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
```diff
@@ -174,4 +254,64 @@ public boolean markSupported() {
...
+  public int readInt() throws IOException {
+    try {
+      return buffer.getInt();
+    } catch (BufferUnderflowException e) {
+      throw new EOFException(e.getMessage());
+    }
+  }
+
+  /*
+  Use ByteBuffer.getLonmg(), which takes advantage of platform intrinsics
```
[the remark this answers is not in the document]
theosib-amazon: got it

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: Got it

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: got it

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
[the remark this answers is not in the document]
theosib-amazon: I already don't like the fact that I have to check the argument to make sure it's not negative and not bigger than int max. buffer.positiion() already checks for the position going out of bounds and throws an exception, so it would be redundant to have another check for the exact same thing here. A catch for an exception that never happens is basically always free, while a test for a condition that never happens is not free until profiling gets enough info about it for the C2 compiler to eliminate it.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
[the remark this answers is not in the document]
theosib-amazon: I went ahead and added two tests to cover the unused SingleBufferInputStream constructors. I considered just deleting these constructors, but I decided that it might be valuable to include them as documentation on how to do this in a way that is congruent to the behavior of HeapByteBuffer, just in case anyone wanted to do this in the future. There's also the risk that someone would think they have to wrap an array with ByteBuffer before using SingleBufferInputStream, but it would be better to avoid the overhead of ByteBuffer.duplicate(). (ByteBuffer.duplicate() appears to take constant time by making a reference to the same backing array, but it's a big constant with loads of checks.)

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
shangxinli: I meant the method here: https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/bytes/BytesUtils.java#L120. It is exactly same but just different signature.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
shangxinli: Check this https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/bytes/BytesUtils.java#L110

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -138,6 +134,18 @@ public int read(byte[] b, int off, int len) throws IOException {
     return delegate.read(b, off, len);
   }
 
+  public int read(byte[] b) throws IOException {
+    return read(b, 0, b.length);
+  }
+
+  public void readFully(byte b[]) throws IOException {
```
shangxinli, line 141: Why we need to be different between line 137 ' byte[] b' and 141 ' byte b[]'?

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
shangxinli: The difference between this method and read() is mainly to precheck if there is enough remaining length. I believe this can be done by wrapping up the read() method and adding the prechecks. Duplicating the code makes it harder to maintain.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -138,6 +134,18 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+    readFully(b, 0, b.length);
+  }
+
+  public void readFully(byte b[], int off, int len) throws IOException {
```
shangxinli, line 145: Don't see where it is used. Don't know why it is 'public'.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
shangxinli: Duplicating code is hard to maintain.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
shangxinli: If they are identical, we should remove the duplicate one.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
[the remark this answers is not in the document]
shangxinli: the try/catch is also more expensive than checking. I agree with Chao to have the check instead of try/catch.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: I know about that method. The BytesUtils code always reads one byte at a time. My version will read a whole word at a time for short and int. This is faster.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: See my other comment on this. These two methods have the the same outcome, but mine is faster. I believe this is warranted for a performance critical path.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: I'll make these changes if you insist. But those prechecks are expensive, which is why I'm trying to avoid them when possible in a performance critical path.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: Well, my objective here is to maximize performance. So we have to decide between maintainability and performance. Let's deliberate over this a bit more, and I'll do what you think is best.

parquet-common/src/main/java/org/apache/parquet/bytes/SingleBufferInputStream.java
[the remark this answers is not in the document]
theosib-amazon: A try/catch is basically free when no exception is thrown, while a check is not. I have tested this, and the try/catch is empirically faster, since the no-exception case is the common case. Putting in the check means we have to wait until the C2 compiler produces a trace without the branch. But even then, there's always the overhead of a check somewhere to be able to fall back to interpretation if the condition is not correct for the trace. I'm avoiding all of that entirely, making this faster in of the most common case.

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
theosib-amazon: Looks like I took care of this already.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: This is almost certainly the result of a copy/paste of the method signature, and I just didn't notice. I went ahead and fixed it.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: This is what the method signatures for DataInputStream.readFully look like.

I also have a whole bunch of other performance improvements I want to contribute (https://docs.google.com/document/d/1fBGpF_LgtfaeHnPD5CFEIpA2Ga_lTITmFdFIcO9Af-g/edit?usp=sharing), and I think this might get used in some of that code.

I'm very soon going to publish an open preview of all of my proposed changes to a branch of my own fork, so we'll be able to check this out.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: Wait. Are you referring to readIntLittleEndianPaddedOnBitWidth or readIntLittleEndianOnThreeBytes?

The former is definitely faster. An argument could be made to remove the latter, although it'll take longer for the JIT to hide the extra layers of virtual calls.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: Here's the complete preview of my changes to ParquetMR: https://github.com/theosib-amazon/parquet-mr/tree/batch-read-optimizations

parquet-common/src/main/java/org/apache/parquet/bytes/MultiBufferInputStream.java
wgtmac: > Well, my objective here is to maximize performance. So we have to decide between maintainability and performance. Let's deliberate over this a bit more, and I'll do what you think is best.

I deeply understand this is a difficult trade-off. Do you have any evidence on the performance penalty if we wrap `read` and `readFully` methods to share some common logic? If the penalty is acceptable, we should definitely go for maintainability.

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
```diff
@@ -157,4 +165,80 @@ public void reset() throws IOException {
...
+
+  public byte readByte() throws IOException {
+    return delegate.readByte();
+  }
+
+  public int readUnsignedByte() throws IOException {
+    return delegate.readUnsignedByte();
+  }
+
+  public short readShort() throws IOException {
```
wgtmac, line 181: I like the idea to provide these read functions to enable larger read. BTW, is there any use case to read a batch of shorts (and other numeric types)?

wgtmac (commented):
IMHO, the open source community priorities code maintainability more than performance gain as the project is widely adopted and maintained by many users and developers. However, if there is any concrete benchmark result to provide compelling evidence, it will be helpful to the discussion. @theosib-amazon @shangxinli @sunchao

parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java
theosib-amazon: I news the new batch read methods heavily in some optimizations I made to Trino. As for short, I can't say I recall any uses in Trino of readShorts(). readShort() is used indirectly through a method that reads a variable sized representation.

shangxinli:
@theosib-amazon Thanks again for your contribution! I see the comments are generally around duplicating code, refactoring, and making code maintainable. If you have a measurement of improvement on this change alone, it would help the reviewers

PR #962, opened 2022-04-28
Performance optimization to ByteBitPackingValuesReader

theosib-amazon:
Minor performance improvements for ByteBitPackingValuesReader:

- Move object creation (temporary byte array) out of frequently used code by retaining small temporary buffer as class member.
- Move less frequently used code (readMore) into separate function to encourage JIT to inline more frequently used code (readInteger)
- Avoid use of unpack8Values that takes ByteBuffer, because it's slower.


shangxinli:
Can you add tests?

theosib-amazon:
There is no new functionality here. There is just a performance optimization. It looks like the following tests should already handle this: BitPackingPerfTest, TestBitPackingColumn. There are also a bunch of other tests that use it indirectly, because they test classes that instantiate ByteBitPackingValuesReader.

ggershinsky:
Optimizations like using byte arrays instead of byte buffers, and allocating the byte array once only, instead of per operation. Done in a concise manner, without unnecessary code changes. LGTM.

PR #968, opened 2022-05-16
PARQUET-2149:  Async IO implementation for ParquetFileReader

parthchandra:
### Jira

This PR addresses the following [PARQUET-2149](https://issues.apache.org/jira/browse/PARQUET-2149): Implement async IO for Parquet file reader

### Tests

This PR adds the following unit tests 
   AsyncMultiBufferInputStream.*
   TestMultipleWriteRead.testReadWriteAsync 
   TestColumnChunkPageWriteStore.testAsync
   
The PR is also tested by changing the default configuration to make all reads async and then ensuring all unit tests pass

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/page/PageReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/page/PageReader.java
@@ -18,10 +18,13 @@
  */
 package org.apache.parquet.column.page;
 
+import java.io.Closeable;
+import java.io.IOException;
+
 /**
  * Reader for a sequence a page from a given column chunk
  */
-public interface PageReader {
+public interface PageReader extends Closeable {
 
   /**
    * @return the dictionary page in that chunk or null if none
@@ -37,4 +40,9 @@ public interface PageReader {
    * @return the next page in that chunk or null if after the last page
    */
   DataPage readPage();
+
+  /**
+   * Close the page reader. By default it is no-op.
+   */
+  default void close() throws IOException {}
 }
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
@@ -1,0 +1,166 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.bytes;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.LongAccumulator;
+import java.util.concurrent.atomic.LongAdder;
+import org.apache.parquet.io.SeekableInputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class AsyncMultiBufferInputStream extends MultiBufferInputStream {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AsyncMultiBufferInputStream.class);
+
+  private int fetchIndex = 0;
+  private final SeekableInputStream fileInputStream;
+  private int readIndex = 0;
+  private ExecutorService threadPool;
+  private LinkedBlockingQueue<Future<Void>> readFutures;
+  private boolean closed = false;
+
+  private LongAdder totalTimeBlocked = new LongAdder();
+  private LongAdder totalCountBlocked = new LongAdder();
+  private LongAccumulator maxTimeBlocked = new LongAccumulator(Long::max, 0L);
+
+  AsyncMultiBufferInputStream(
+      ExecutorService threadPool, SeekableInputStream fileInputStream, List<ByteBuffer> buffers) {
+    super(buffers);
+    this.fileInputStream = fileInputStream;
+    this.threadPool = threadPool;
+    readFutures = new LinkedBlockingQueue<>(buffers.size());
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("ASYNC: Begin read into buffers ");
+      for (ByteBuffer buf : buffers) {
+        LOG.debug("ASYNC: buffer {} ", buf);
+      }
+    }
+    fetchAll();
+  }
+
+  private void checkState() {
+    if (closed) {
+      throw new RuntimeException("Stream is closed");
+    }
+  }
+
+  private void fetchAll() {
+    checkState();
+    submitReadTask(0);
+  }
+
+  private void submitReadTask(int bufferNo) {
+    ByteBuffer buffer = buffers.get(bufferNo);
+    try {
+      readFutures.put(threadPool.submit(() -> {
+        readOneBuffer(buffer);
+        if (bufferNo < buffers.size() - 1) {
+          submitReadTask(bufferNo + 1);
+        }
+        return null;
+      }));
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    }
+  }
+
+  private void readOneBuffer(ByteBuffer buffer) {
+    long startTime = System.nanoTime();
+    try {
+      fileInputStream.readFully(buffer);
+      buffer.flip();
+      long readCompleted = System.nanoTime();
+      long timeSpent = readCompleted - startTime;
+      LOG.debug("ASYNC Stream: READ - {}", timeSpent / 1000.0);
+      fetchIndex++;
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  public boolean nextBuffer() {
+    checkState();
+    // hack: parent constructor can call this method before this class is fully initialized.
+    // Just return without doing anything.
+    if (readFutures == null) {
+      return false;
+    }
+    if (readIndex < buffers.size()) {
+      long start = System.nanoTime();
+      try {
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("ASYNC (next): Getting next buffer");
+        }
+        Future<Void> future = readFutures.take();
+        future.get();
+        long timeSpent = System.nanoTime() - start;
+        totalCountBlocked.add(1);
+        totalTimeBlocked.add(timeSpent);
+        maxTimeBlocked.accumulate(timeSpent);
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("ASYNC (next): {}: Time blocked for read {} ns", this, timeSpent);
+        }
+      } catch (InterruptedException | ExecutionException e) {
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+        LOG.error("Async (next): exception while getting next buffer: ", e);
+        throw new RuntimeException(e);
+      }
+      readIndex++;
+    }
+    return super.nextBuffer();
+  }
+
+  public void close() {
+    LOG.debug(
+        "ASYNC Stream: Blocked: {} {} {}",
+        totalTimeBlocked.longValue() / 1000.0,
+        totalCountBlocked.longValue(),
+        maxTimeBlocked.longValue() / 1000.0);
+    Future<Void> readResult;
+    while (!readFutures.isEmpty()) {
+      try {
+        readResult = readFutures.poll();
+        readResult.get();
+        if (!readResult.isDone() && !readResult.isCancelled()) {
+          readResult.cancel(true);
+        } else {
+          readResult.get(1, TimeUnit.MILLISECONDS);
+        }
+      } catch (ExecutionException | InterruptedException | TimeoutException e) {
+        // Do nothing
+      }
+    }
+    closed = true;
+  }
+}
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/SequenceByteBufferInputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/SequenceByteBufferInputStream.java
@@ -1,0 +1,268 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.bytes;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ *   A bare minimum implementation of a {@link java.io.SequenceInputStream} that wraps an
+ *   <i>ordered</i> collection of ByteBufferInputStreams.
+ *   <p>
+ *   This class, as implemented, is intended only for a specific use in the ParquetFileReader and
+ *   throws {@link UnsupportedOperationException} in unimplemented methods to catch any unintended
+ *   use in other cases.
+ *   <p>
+ *   Even though this class is derived from ByteBufferInputStream it explicitly does not support any
+ *   byte buffer related methods like slice. It does, however support sliceBuffers which is a
+ *   curious case of reading data from underlying streams
+ *   <p>
+ *   Even though this class changes the state of the underlying streams (by reading from them)
+ *   it does not own them and so the close method does not close the streams. To avoid resource
+ *   leaks the calling code should close the underlying streams
+ */
+public class SequenceByteBufferInputStream extends ByteBufferInputStream {
+
+  Collection<ByteBufferInputStream> collection;
+  Iterator<ByteBufferInputStream> iterator;
+  ByteBufferInputStream current;
+  long position = 0;
+
+  @Override
+  public String toString() {
+    return "SequenceByteBufferInputStream{" + "collection="
+        + collection + ", current="
+        + current + ", position="
+        + position + '}';
+  }
+
+  public SequenceByteBufferInputStream(Collection<ByteBufferInputStream> collection) {
+    this.collection = collection;
+    iterator = collection.iterator();
+    current = iterator.hasNext() ? iterator.next() : null;
+    if (current == null) {
+      throw new UnsupportedOperationException(
+          "Initializing SequenceByteBufferInputStream with an empty collection is not supported");
+    }
+  }
+
+  @Override
+  public long position() {
+    return position;
+  }
+
+  @Override
+  public int read(ByteBuffer out) {
+    int len = out.remaining();
+    if (len <= 0) {
+      return 0;
+    }
+    if (current == null) {
+      return -1;
+    }
+    int totalBytesRead = 0;
+    while (totalBytesRead < len) {
+      int bytesRead = current.read(out);
+      if (bytesRead == -1) {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          break;
+        }
+      } else {
+        totalBytesRead += bytesRead;
+      }
+    }
+    position += totalBytesRead;
+    return totalBytesRead;
+  }
+
+  @Override
+  public ByteBuffer slice(int length) throws EOFException {
+    throw new UnsupportedOperationException("slice is not supported");
+  }
+
+  @Override
+  /**
+   * This is a blocking call. Use with care when using in asynchronous mode.
+   */
+  public List<ByteBuffer> sliceBuffers(long len) throws EOFException {
+    if (len <= 0) {
+      return Collections.emptyList();
+    }
+
+    if (current == null) {
+      throw new EOFException();
+    }
+
+    List<ByteBuffer> buffers = new ArrayList<>();
+    long bytesAccumulated = 0;
+    while (bytesAccumulated < len) {
+      // This is not strictly according to the input stream contract, but once again the
+      // underlying implementations of ByteBufferInputStream return the available bytes
+      // based on the size of the underlying buffers rather than the bytes currently read
+      // into the buffers. This works for us because the underlying implementations will
+      // actually fill the buffers with the data before returning the slices we ask for
+      // (which is why this is a blocking call)
+      if (current.available() > 0) {
+        int bufLen = (int) Math.min(len - bytesAccumulated, current.available());
+        List<ByteBuffer> currentSlices = current.sliceBuffers(bufLen);
+        buffers.addAll(currentSlices);
+        bytesAccumulated += bufLen;
+
+        // update state; the bytes are considered read
+        this.position += bufLen;
+      } else {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          // there are no more streams
+          throw new EOFException();
+        }
+      }
+    }
+    position += bytesAccumulated;
+    return buffers;
+  }
+
+  @Override
+  public ByteBufferInputStream sliceStream(long length) throws EOFException {
+    throw new UnsupportedOperationException("sliceStream is not supported");
+  }
+
+  @Override
+  public List<ByteBuffer> remainingBuffers() {
+    throw new UnsupportedOperationException("remainingBuffers is not supported");
+  }
+
+  @Override
+  public ByteBufferInputStream remainingStream() {
+    throw new UnsupportedOperationException("remainingStream is not supported");
+  }
+
+  @Override
+  public int read() throws IOException {
+    int val;
+    while (true) {
+      try {
+        val = current.read() & 0xFF; // as unsigned
+        position += 1;
+        break;
+      } catch (EOFException e) {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          return -1;
+        }
+      }
+    }
+    return val;
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException {
+    if (len <= 0) {
+      if (len < 0) {
+        throw new IndexOutOfBoundsException("Read length must be greater than 0: " + len);
+      }
+      return 0;
+    }
+    if (current == null) {
+      return -1;
+    }
+    int totalBytes = 0;
+    while (totalBytes < len) {
+      int bytesRead = current.read(b, off + totalBytes, len - totalBytes);
+      if (bytesRead == -1) {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          break;
+        }
+      } else {
+        totalBytes += bytesRead;
+      }
+    }
+    position += totalBytes;
+    return totalBytes;
+  }
+
+  @Override
+  public long skip(long n) {
+    if (n <= 0) {
+      if (n < 0) {
+        throw new IndexOutOfBoundsException("Skip length must be greater than 0: " + n);
+      }
+      return 0;
+    }
+    if (current == null) {
+      return -1;
+    }
+    long totalBytesSkipped = 0;
+    while (totalBytesSkipped < n) {
+      long bytesSkipped = current.skip(n - totalBytesSkipped);
+      // the contract for skip specifies that skip may return 0 if EOF is reached in which case
+      // we have no good way of knowing the end of stream has been reached. We depend on the
+      // fact that all implementations of ByteBufferInputStream return -1 on EOF
+      if (bytesSkipped == -1) {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          break;
+        }
+      } else {
+        totalBytesSkipped += bytesSkipped;
+      }
+    }
+    position += totalBytesSkipped;
+    return totalBytesSkipped;
+  }
+
+  @Override
+  public int available() {
+    return current.available();
+  }
+
+  @Override
+  public void close() throws IOException {
+    super.close();
+  }
+
+  @Override
+  public void mark(int readlimit) {
+    throw new UnsupportedOperationException("mark is not supported");
+  }
+
+  @Override
+  public void reset() throws IOException {
+    throw new UnsupportedOperationException("reset is not supported");
+  }
+
+  @Override
+  public boolean markSupported() {
+    return false;
+  }
+}
--- a/parquet-hadoop/src/main/java/org/apache/parquet/HadoopReadOptions.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/HadoopReadOptions.java
@@ -20,6 +20,7 @@
 package org.apache.parquet;
 
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.bytes.ByteBufferAllocator;
@@ -44,6 +45,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
       boolean useBloomFilter,
       boolean useOffHeapDecryptBuffer,
       boolean useHadoopVectoredIo,
+      boolean enableAsyncIOReader,
+      boolean enableParallelColumnReader,
       FilterCompat.Filter recordFilter,
       MetadataFilter metadataFilter,
       CompressionCodecFactory codecFactory,
@@ -52,6 +55,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
       Map<String, String> properties,
       Configuration conf,
       FileDecryptionProperties fileDecryptionProperties,
+      ExecutorService ioThreadPool,
+      ExecutorService processThreadPool,
       ParquetMetricsCallback metricsCallback) {
     super(
         useSignedStringMinMax,
@@ -63,6 +68,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
         useBloomFilter,
         useOffHeapDecryptBuffer,
         useHadoopVectoredIo,
+        enableAsyncIOReader,
+        enableParallelColumnReader,
         recordFilter,
         metadataFilter,
         codecFactory,
@@ -70,6 +77,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
         maxAllocationSize,
         properties,
         fileDecryptionProperties,
+        ioThreadPool,
+        processThreadPool,
         metricsCallback,
         new HadoopParquetConfiguration(conf));
     this.conf = conf;
@@ -126,6 +135,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
           useBloomFilter,
           useOffHeapDecryptBuffer,
           useHadoopVectoredIo,
+          enableAsyncIOReader,
+          enableParallelColumnReader,
           recordFilter,
           metadataFilter,
           codecFactory,
@@ -134,6 +145,8 @@ public class HadoopReadOptions extends ParquetReadOptions {
           properties,
           conf,
           fileDecryptionProperties,
+          ioThreadPool,
+          processThreadPool,
           metricsCallback);
     }
   }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -19,14 +19,13 @@
 package org.apache.parquet.hadoop;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
-import java.util.ArrayDeque;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.PrimitiveIterator;
-import java.util.Queue;
+import java.util.concurrent.LinkedBlockingDeque;
 import org.apache.parquet.ParquetReadOptions;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.ByteBufferReleaser;
@@ -54,7 +53,7 @@ import org.slf4j.LoggerFactory;
  * The name is kind of confusing since it references three different "entities"
  * in our format: columns, chunks, and pages
  */
-class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore {
+public class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore {
   private static final Logger LOG = LoggerFactory.getLogger(ColumnChunkPageReadStore.class);
 
   /**
@@ -64,11 +63,13 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
    * This implementation is provided with a list of pages, each of which
    * is decompressed and passed through.
    */
-  static final class ColumnChunkPageReader implements PageReader {
+  public static final class ColumnChunkPageReader implements PageReader {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ColumnChunkPageReader.class);
 
     private final BytesInputDecompressor decompressor;
     private final long valueCount;
-    private final Queue<DataPage> compressedPages;
+    private final LinkedBlockingDeque<Optional<DataPage>> compressedPages;
     private final DictionaryPage compressedDictionaryPage;
     // null means no page synchronization is required; firstRowIndex will not be returned by the pages
     private final OffsetIndex offsetIndex;
@@ -80,31 +81,33 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
     private final ByteBufferReleaser releaser;
+    private final FilePageReader filePageReader;
 
     ColumnChunkPageReader(
         BytesInputDecompressor decompressor,
-        List<DataPage> compressedPages,
+        LinkedBlockingDeque<Optional<DataPage>> compressedPages,
         DictionaryPage compressedDictionaryPage,
         OffsetIndex offsetIndex,
+        long valueCount,
         long rowCount,
         BlockCipher.Decryptor blockDecryptor,
         byte[] fileAAD,
         int rowGroupOrdinal,
         int columnOrdinal,
+        FilePageReader filePageReader,
         ParquetReadOptions options) {
       this.decompressor = decompressor;
-      this.compressedPages = new ArrayDeque<DataPage>(compressedPages);
+      this.compressedPages = compressedPages;
       this.compressedDictionaryPage = compressedDictionaryPage;
-      long count = 0;
-      for (DataPage p : compressedPages) {
-        count += p.getValueCount();
-      }
-      this.valueCount = count;
+      this.valueCount = valueCount;
       this.offsetIndex = offsetIndex;
       this.rowCount = rowCount;
       this.options = options;
       this.releaser = new ByteBufferReleaser(options.getAllocator());
       this.blockDecryptor = blockDecryptor;
+
+      this.filePageReader = filePageReader;
+
       if (null != blockDecryptor) {
         dataPageAAD =
             AesCipher.createModuleAAD(fileAAD, ModuleType.DataPage, rowGroupOrdinal, columnOrdinal, 0);
@@ -114,6 +117,11 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
         dataPageAAD = null;
         dictionaryPageAAD = null;
       }
+    }
+
+    @Override
+    public void close() throws IOException {
+      this.filePageReader.close();
     }
 
     private int getPageOrdinal(int currentPageIndex) {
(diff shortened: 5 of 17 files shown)


parthchandra:
Anyone know why the CI checks are failing with a SocketTimeout exception, and what to do to address this?

dbtsai:
@parthchandra do you have performance benchmark? Thanks

parthchandra:
I have some numbers from an internal benchmark using Spark. I didn't see any benchmarks in the Parquet codebase that I could reuse.

Here are the numbers from my own benchmark - 

 - 10 runs, each run reads all columns from store_sales (the largest table) in the TPC-DS (100G) dataset
  `spark.sql("select * from store_sales")`
  - Sync reader with default 8MB buffer size, Async reader with 1MB buffer size (achieves better pipelining)
  - Run on Macbook Pro, reading from S3. Spark has 6 cores.
  - All times in seconds
   
  | Run | Async | Sync | Async (w/o outliers)| Sync (w/o outliers) |
  | ---:| ---:| ---:| ---:| ---:| 
|1| 84| 102| - | - |		
|2| 90| 366| 90| 366|
|3| 78| 156| - | 156|
|4| 84| 128| 84| - |
|5| 108|402| - | - |		
|6| 90| 432| 90| - |
|7| 84| 378| 84| 378|
|8| 108|324| - | 324|
|9| 90| 318| 90| 318|
|10|90| 282| 90| 282|
|Average| 90.6|	288.8| 88| 304|
|Median| 90| 321| **90**| **321**|    
|StdDev| 9.98| 119.


After removing the two highest and two lowest runs for each case, and taking the median value:

Async: 90 sec
Sync: 321 sec

shangxinli:
Great effort! WIll have a look after the build succeed.

theosib-amazon:
@parthchandra Would you mind having a look at my I/O performance optimization plan for ParquetMR? I think we should coordinate, since we have some ideas that might overlap what we touch.
https://docs.google.com/document/d/1fBGpF_LgtfaeHnPD5CFEIpA2Ga_lTITmFdFIcO9Af-g/edit?usp=sharing

dbtsai:
cc @rdblue @gszadovszky @ggershinsky

parthchandra:
> Great effort! WIll have a look after the build succeed.
@shangxinli I have no idea how to get the failed CI to pass. These  failures appear to be in unrelated areas caused by some infra issues. Is there a way to trigger a rerun?

parthchandra:
> @parthchandra Would you mind having a look at my I/O performance optimization plan for ParquetMR? I think we should coordinate, since we have some ideas that might overlap what we touch.
> https://docs.google.com/document/d/1fBGpF_LgtfaeHnPD5CFEIpA2Ga_lTITmFdFIcO9Af-g/edit?usp=sharing

@theosib-amazon I read your document and went thru  #960. It looks like for the most part, #960 and this PR and complement each other.  The overlap I see is in the changes to `MultiBufferInputStream` where you have added the `readFully`, and `skipFully` APIs. The bulk of my changes for async IO are in a class derived from `MultiBufferInputStream` and the heart of the changes depends on overriding `MultiBufferInputStream.nextBuffer`. In `MultiBufferInputStream.nextBuffer` the assumption is that all the buffers have been read into. In `AsyncMultiBufferInputStream.nextBuffer` this assumption is removed and the call *blocks* only if the next required buffer has not been read into.
Now, `skipFully` and `readFully` are potentially blocking calls because both call `nextBuffer` repeatedly if necessary. To gain  maximum pipelining, you want to make calls to skipFully and readFully such that you never block for too long (or at all) in the call. You will get this if you are skipping or reading less than the number of bytes in a single buffer. This is generally the case as decompression and decoding is at the page level and that is smaller than the size of a single buffer. However, for your optimizations, you should be aware of this behaviour.
From what I see, I don't think there will be a conflict.  I'll pull in your PR and give it a deeper look.

theosib-amazon:
@parthchandra One thing that confuses me a bit is that these buffers have only ByteBuffer inside them. There's no actual I/O, so it's not possible to block. Do you have subclasses that provide some sort of access to real I/O?

parquet-column/src/main/java/org/apache/parquet/column/page/PageReader.java
```diff
@@ -37,4 +39,9 @@ public interface PageReader {
    * @return the next page in that chunk or null if after the last page
    */
   DataPage readPage();
+
+  /**
+   * Close the page reader. By default it is no-op.
+   */
+  default void close() throws IOException {}
```
kbendick, line 46: Should we add `Closeable` as an implemented interface?

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Override
+  public boolean nextBuffer() {
+    checkState();
+    // hack: parent constructor can call this method before this class is fully initialized.
+    // Just return without doing anything.
+    if (readFutures == null) {
+      return false;
+    }
```
kbendick, line 129: Because `checkState` has synchronization, would it be safe to move the `checkState` before this somehow or add some kind of less expensive boolean check that we can set to `true` immediately after the `super()` call?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -21,12 +21,15 @@
 import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.PrimitiveIterator;
 import java.util.Queue;
```
kazuyukitanimura, line 29: I think `Queue` is no longer used with this change. Should we remove the import?

kazuyukitanimura (commented):
Thank you @parthchandra A few cosmetic comments

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+  final SeekableInputStream fileInputStream;
+  int fetchIndex = 0;
+  int readIndex = 0;
+  ExecutorService threadPool;
+  LinkedBlockingQueue<Future<Void>> readFutures;
+  boolean closed = false;
+
+  LongAdder totalTimeBlocked = new LongAdder();
+  LongAdder totalCountBlocked = new LongAdder();
+  LongAccumulator maxTimeBlocked = new LongAccumulator(Long::max, 0L);
```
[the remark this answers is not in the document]
parthchandra: Done

parquet-column/src/main/java/org/apache/parquet/column/page/PageReader.java
kbendick: Cool. I know a lot of times I check if instances of closeable and then call close.

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+    long startTime = System.nanoTime();
+    try {
+      fileInputStream.readFully(buffer);
+      buffer.flip();
+      long readCompleted = System.nanoTime();
+      long timeSpent = readCompleted - startTime;
+      LOG.debug("ASYNC Stream: READ - {}", timeSpent / 1000.0);
+      long putStart = System.nanoTime();
+      long putCompleted = System.nanoTime();
+      LOG.debug("ASYNC Stream: FS READ (output) BLOCKED - {}",
```
steveloughran, line 103: how does this work?

parquet-common/src/main/java/org/apache/parquet/bytes/SequenceByteBufferInputStream.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  @Override
+  public long position() {
+    return position;
+  }
+
+  @Override
+  public int read(ByteBuffer out) {
```
steveloughran, line 79: are you 100% confident that all uses of read() in parquet are in a single thread? it is not unknown for apps to read across threads, even though the java APIs say "don't"

parquet-common/src/main/java/org/apache/parquet/bytes/SequenceByteBufferInputStream.java
```diff
@@ -0,0 +1,269 @@
...
+    while (true) {
+      try {
+        val = current.read() & 0xFF; // as unsigned
+        position += 1;
+        break;
+      } catch (EOFException e) {
+        if (iterator.hasNext()) {
+          current = iterator.next();
+        } else {
+          throw new EOFException("End of streams");
```
steveloughran, line 179: InputStream.read mandates return -1 on eof

steveloughran (commented):
minor comments. i would recommend looking at https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/filesystem/fsdatainputstream.html and lifting the tests in hadoop-common to make sure they work with this stream. because, correct or not, the semantics there are what applications often expect.

parthchandra:
@steveloughran thank you very much for taking the time to review and provide feedback! 

> 1. whose s3 client was used for testing here -if the s3a one, which hadoop release?

I was working with s3a -
  Spark 3.2.1
  Hadoop (Hadoop-aws) 3.3.2
  AWS SDK 1.11.655
  

> 2. the azure abfs and gcs connectors do async prefetching of the next block, but are simply assuming that code will read sequentially; if there is another seek/readFully to a new location, those prefetches will be abandoned. there is work in s3a to do prefetching here with caching, so as to reduce the penalty of backwards seeks. https://issues.apache.org/jira/browse/HADOOP-18028

I haven't worked with abfs or gcs. If the connectors do async pre-fetching, that would be great. Essentially,  the time the Parquet reader would have to block in the file system API would reduce substantially. In such a case, we could turn the async reader on/off  and rerun the benchmark to compare. From past experience with the MaprFS which had very aggressive read ahead in its hdfs client, I would still expect better parquet speeds. The fact that the prefetch is turned off when a seek occurs is usual behaviour, but we may see no benefit from the connector in that case. So a combination of async reader and async connector might end up being a great solution (maybe at a slightly greater CPU utilization). We would still have to do a benchmark to see the real effect.
The async version in this PR takes care of the sequential read requirement by a) opening a new stream for each column and ensuring every column is read sequentially. Footers are read using a separate stream. Except for the footer, no other stream ever seeks to a new location. b) The amount of data to be read is predetermined so there is never a read ahead that is discarded.

> 
> hadoop is adding a vectored IO api intended for libraries like orc and parquet to be able to use, where the application provides an unordered list of ranges, a bytebuffer supplier and gets back a list of futures to wait for. the base implementation simply reads using readFully APi. s3a (and later abfs) will do full async retrieval itself, using the http connection pool. https://issues.apache.org/jira/browse/HADOOP-18103
> 
> both vectored io and s3a prefetching will ship this summer in hadoop 3.4.0. i don't see this change conflicting with this, though they may obsolete a lot of it.

Yes, I became aware of this recently. I'm discussing integration of these efforts in a separate channel. At the moment I see no conflict, but have yet to determine how much of this async work would need to be changed. I suspect we may be able to eliminate or vastly simplify `AsyncMultiBufferInputStream`. 

> have you benchmarked this change with abfs or google gcs connectors to see what difference it makes there?

No I have not. Would love help from anyone in the community with access to these. I only have access to S3.

steveloughran:
> I was working with s3a
> Spark 3.2.1
> Hadoop (Hadoop-aws) 3.3.2
> AWS SDK 1.11.655

thanks., that means you are current with all shipping improvments. the main one extra is to use openFile(), passing in length and requesting randomio. this guarantees ranged GET requests and cuts the initial HEAD probe for existence/size of file.

>> have you benchmarked this change with abfs or google gcs connectors to see what difference it makes there?

> No I have not. Would love help from anyone in the community with access to these. I only have access to S3.

that I have. FWIW, with the right tuning of abfs prefetch (4 threads, 128 MB blocks) i can get full FTTH link rate from a remote store; 700 mbit/s . that's to the base station. once you add wifi the bottlenecks move.

parthchandra:
> thanks., that means you are current with all shipping improvments. the main one extra is to use openFile(), passing in length and requesting randomio. this guarantees ranged GET requests and cuts the initial HEAD probe for existence/size of file.

By `openFile()` do you mean `FileSystem.openFileWithOptions(Path,OpenFileParameters)`?
While looking I realized the Parquet builds with a [much older version of hadoop](https://github.com/apache/parquet-mr/blob/a2da156b251d13bce1fa81eb95b555da04880bc1/pom.xml#L79)  
> > > have you benchmarked this change with abfs or google gcs connectors to see what difference it makes there?
> 
> > No I have not. Would love help from anyone in the community with access to these. I only have access to S3.
> 
> that I have. FWIW, with the right tuning of abfs prefetch (4 threads, 128 MB blocks) i can get full FTTH link rate from a remote store; 700 mbit/s . that's to the base station. once you add wifi the bottlenecks move.

Wow! That is nearly as fast as local HDD. At this point the bottlenecks in parquet begin to move towards decompression and decoding but IO remains the slowest link in the chain.  One thing we get with my PR is that the ParquetFileReader had assumptions built in that all data must be read before downstream can proceed. Some of my changes are related to removing these assumptions and ensuring that downstream processing does not block until an entire column is read so we get efficient pipelining. 
What does the 128 MB block mean? Is this the amount prefetched for a stream? The read API does not block until the entire block is filled, I presume. 
With my PR, parquet IO is reading 8MB at a time (default) and downstream is processing 1MB at a time (default) and several such streams (one per column) are in progress at the same time. Hopefully, this read pattern would work with the prefetch.

parthchandra:
> Is byte (and arrays and buffers of bytes) the only datatype you support? My PR is optimizing code paths that pull ints, longs, and other sizes out of the data buffers. Are those not necessary for any of the situations where you're using an async buffer?
The input stream API is generally unaware of the datatypes of its contents and so those are the only apis I use. The other reason is that the ParquetFileReader returns Pages which basically contain metadata and ByteBuffers of _compressed_ data. The decompression and decoding into types comes much later in a downstream thread.
For your PR, I don't think the AsyncMultibufferInputStream is every going to be in play in the paths you're optimizing. But just in case it is, your type aware methods will work as is because AsyncMultibufferInputStream is derived from MultiBufferInputStream and will inherit those methods.

steveloughran:
> At this point the bottlenecks in parquet begin to move towards decompression and decoding but IO remains the slowest link in the chain.

Latency is the killer; in an HTTP request you want read enough but not discard data or break an http connection if the client suddenly does a seek() or readFully() somewhere else. file listings, existence checks etc.

> One thing we get with my PR is that the ParquetFileReader had assumptions built in that all data must be read before downstream can proceed. Some of my changes are related to removing these assumptions and ensuring that downstream processing does not block until an entire column is read so we get efficient pipelining.

That'd be great. now, if you could also handle requesting different columns in parallel and processing them out of order.

> What does the 128 MB block mean? Is this the amount prefetched for a stream? The read API does not block until the entire block is filled, I presume.

this was the abfs client set to do four GET requests of 128MB each. this would be awful for columns stores where smaller ranges are often requested/processed before another seek is made, but quite often parquet does do more back to back reads than just one read/readFully request

> With my PR, parquet IO is reading 8MB at a time (default) and downstream is processing 1MB at a time (default) and several such streams (one per column) are in progress at the same time. Hopefully, this read pattern would work with the prefetch.

be good to think about vectored IO.

and yes, updating parquet dependencies would be good, hadoop 3.3.0 should be the baseline.

just sketched out my thoughts on this. I've played with some of this in my own branch. I think the next step would be for me to look at the benchmark code to make it targetable elsewhere.


https://docs.google.com/document/d/1y9oOSYbI6fFt547zcQJ0BD8VgvJWdyHBveaiCHzk79k/

theosib-amazon:
This is interesting, because when I did profiling of Trino, I found that although I/O (from S3, over the network no less) was significant, even more time was spent in compute. Maybe you're getting improved performance because you're increasing *parallelism* between I/O and compute.

parthchandra:
> Latency is the killer; in an HTTP request you want read enough but not discard data or break an http connection if the client suddenly does a seek() or readFully() somewhere else. file listings, existence checks etc.
> 
> That'd be great. now, if you could also handle requesting different columns in parallel and processing them out of order.

I do. The Parquet file reader api that reads row groups in sync mode reads all columns in sequence. In async mode, it fires off a task for every column blocking only to read the first page of every column before returning. This part also uses a different thread pool from the IO tasks so that IO tasks never wait because there are no available threads in the thread pool.

> 
> be good to think about vectored IO.

I think I know how to integrate this PR with the vectored IO, but this is only after a cursory look. 

> 
> and yes, updating parquet dependencies would be good, hadoop 3.3.0 should be the baseline.

Who can drive this  (presumably) non-trivial change? I myself have no karma points :(
 
> just sketched out my thoughts on this. I've played with some of this in my own branch. I think the next step would be for me to look at the benchmark code to make it targetable elsewhere.
> 
> https://docs.google.com/document/d/1y9oOSYbI6fFt547zcQJ0BD8VgvJWdyHBveaiCHzk79k/

This is great. I now have much more context of where you are coming from (and going to) !

theosib-amazon:
> > Is byte (and arrays and buffers of bytes) the only datatype you support? My PR is optimizing code paths that pull ints, longs, and other sizes out of the data buffers. Are those not necessary for any of the situations where you're using an async buffer?
> > The input stream API is generally unaware of the datatypes of its contents and so those are the only apis I use. The other reason is that the ParquetFileReader returns Pages which basically contain metadata and ByteBuffers of _compressed_ data. The decompression and decoding into types comes much later in a downstream thread.
> > For your PR, I don't think the AsyncMultibufferInputStream is every going to be in play in the paths you're optimizing. But just in case it is, your type aware methods will work as is because AsyncMultibufferInputStream is derived from MultiBufferInputStream and will inherit those methods.

I'm still learning Parquet's structure. So it sounds to me like these buffer input streams are used twice. Once to get data and decompress it, and then once again to decode it into data structures. Is that correct? So it sounds like you're optimizing one layer of processing, and I'm optimizing the next layer up, and it's kindof a coincidence that we're touching some of the same classes just because code reuse has been possible here.

parthchandra:
BTW, adding more tests for the InputStream implementations.

parthchandra:
> So it sounds like you're optimizing one layer of processing, and I'm optimizing the next layer up, and it's kindof a coincidence that we're touching some of the same classes just because code reuse has been possible here.
Yeah, kind of cool :)

parthchandra:
> This is interesting, because when I did profiling of Trino, I found that although I/O (from S3, over the network no less) was significant, even more time was spent in compute. Maybe you're getting improved performance because you're increasing _parallelism_ between I/O and compute.

It may be because I was using Spark's vectorized parquet decoding which is an order or magnitude faster than parquet library's row by row decoding (see [Spark benchmarks](https://github.com/apache/spark/blob/master/sql/core/benchmarks/DataSourceReadBenchmark-results.txt)). If trino is not doing vectorized decoding (I took a very quick look and I don't think it is), I would suggest you can look into that next. All the cool kids are doing it.

theosib-amazon:
That batch reader in Presto reminds me of some of the experimental changes I made in Trino. I modified PrimitiveColumnReader to work out how many of each data item it needs to read from the data source and requests all of them at once in an array. This doubled the performance of some TPCDS queries. This is why I have array access methods planned for ParquetMR. (https://docs.google.com/document/d/1fBGpF_LgtfaeHnPD5CFEIpA2Ga_lTITmFdFIcO9Af-g/edit?usp=sharing) Requesting data in bulk saves a lot on function call overhead for each data item.

parquet-common/src/main/java/org/apache/parquet/bytes/SequenceByteBufferInputStream.java
steveloughran: good to know.

steveloughran:
I've started work on a fs-api-shim library, with the goal of "apps compile against hadoop 3.2.0 can get access to the 3.3 and 3.4 APIs when available either with transparent fallback (openFile()) or ability to probe the API before trying to invoke

https://github.com/steveloughran/fs-api-shim

openfile takes the seek & status params, falls back to open()
: https://github.com/steveloughran/fs-api-shim/blob/main/fs-api-shim-library/src/main/java/org/apache/hadoop/fs/shim/FileSystemShim.java#L87

ByteBufferReadable will raise UnsupportedException if not found, there is a check for it
https://github.com/steveloughran/fs-api-shim/blob/main/fs-api-shim-library/src/main/java/org/apache/hadoop/fs/shim/FSDataInputStreamShim.java

Vector IO SHALL be available the same way

Adopt 3.2.0 then and we will help give the library the ability to use the newer api calls, even stuff not yet shipped in apache releases.

(I want to release this as an asf artifact with oversight by hadoop project. lets us maintain it)

steveloughran:
bq.  perhaps check if the ByteBufferReadable interface is implemented in the stream?

The requirement for the `hasCapability("in:readbytebuffer")` to return true postdates the API; there's no way to be confident that if the probe returns false (or hasPathCapability() isn't available) that the method *isn't actually there*

see #951 for a design which will trust a `true` response, falling back to looking at the wrapped stream. Note that as it calls getWrapped() it is calling methods tagged LimitedPrivate. it should really do that...at the very least hadoop needs a PR saying "we need to do this because..." and that tag can be changed

steveloughran:
(i could of course add those probes into the shim class, so at least that access of internals was in one place)

parquet-hadoop/src/main/java/org/apache/parquet/HadoopReadOptions.java
```diff
@@ -61,9 +65,10 @@ private HadoopReadOptions(boolean useSignedStringMinMax,
                             Configuration conf,
                             FileDecryptionProperties fileDecryptionProperties) {
     super(
-        useSignedStringMinMax, useStatsFilter, useDictionaryFilter, useRecordFilter, useColumnIndexFilter,
-        usePageChecksumVerification, useBloomFilter, recordFilter, metadataFilter, codecFactory, allocator,
-        maxAllocationSize, properties, fileDecryptionProperties
+      useSignedStringMinMax, useStatsFilter, useDictionaryFilter, useRecordFilter,
```
shangxinli, line 68: it seems two spaces were removed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1796,5 +1882,314 @@ public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOEx
...
+
+      if (null != headerBlockDecryptor) {
+        dataPageHeaderAAD = AesCipher.createModuleAAD(aadPrefix, ModuleType.DataPageHeader,
+          rowGroupOrdinal,
+          columnOrdinal, chunk.getPageOrdinal(dataPageCountReadSoFar));
+      }
+      if (null != pageBlockDecryptor) {
+        dataPageAAD = AesCipher.createModuleAAD(aadPrefix, ModuleType.DataPage, rowGroupOrdinal,
+          columnOrdinal, 0);
+        dictionaryPageAAD = AesCipher.createModuleAAD(aadPrefix, ModuleType.DictionaryPage,
```
shangxinli, line 1956: The 'dictionaryPageAAD' is only used in toString()?

shangxinli: @ggershinsky Can you review all the decryption-related changes?

shangxinli (commented):
I reviewed half of the changes. I will continue when I get time.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -126,6 +127,42 @@ public class ParquetFileReader implements Closeable {
 
   public static String PARQUET_READ_PARALLELISM = "parquet.metadata.read.parallelism";
 
+  public static int numProcessors = Runtime.getRuntime().availableProcessors();
```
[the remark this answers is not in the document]
steveloughran: dynamically changing the number of threads/buffer sizes/cache sizes is a recurrent source of pain in past work, as once you get to 128 core systems they often end up asking for too much of a limited resource

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,158 @@
...
+      long start = System.nanoTime();
+      try {
+        LOG.debug("ASYNC (next): Getting next buffer");
+        Future<Void> future = readFutures.take();
+        future.get();
+        long timeSpent = System.nanoTime() - start;
+        totalCountBlocked.add(1);
+        totalTimeBlocked.add(timeSpent);
+        maxTimeBlocked.accumulate(timeSpent);
+        LOG.debug("ASYNC (next): {}: Time blocked for read {} ns", this, timeSpent);
```
ggershinsky, line 125: should `if (LOG.isDebugEnabled()) {` be added here and in 118? This check is performed in the constructor (line 58); `nextBuffer()` is called with higher(/same) frequency.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
ggershinsky: sure

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1796,5 +1882,314 @@ public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOEx
...
+        ", length=" + length +
+        ", chunks=" + chunks +
+        '}';
+    }
   }
+
+  /**
+   * Encapsulates the reading of a single page.
+   */
+  public class PageReader implements Closeable {
```
ggershinsky, line 1899: we already have a PageReader (interface). Could you rename this class.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1796,5 +1882,314 @@ public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOEx
...
+            readResult.cancel(true);
+          } else {
+            readResult.get(1, TimeUnit.MILLISECONDS);
+          }
+        } catch (Exception e) {
+          // Do nothing
+        }
+      }
+      String mode = isParallelColumnReaderEnabled()?"ASYNC" : "SYNC";
+      LOG.debug("READ PAGE: {}, {}, {}, {}, {}, {}, {}", mode,
```
ggershinsky, line 2152: this sounds as if logged in a readPage call.

ggershinsky (changes requested):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
ggershinsky: maybe can also be separated from the ParquetFileReader, this is a chance to reduce the size of the latter :)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
ggershinsky: Yep, the `dictionaryPageAAD` is not necessary here. This is a significant code change, more than just moving the current logic of
```java
public ColumnChunkPageReader readAllPages(BlockCipher.Decryptor headerBlockDecryptor, BlockCipher.Decryptor pageBlockDecryptor, byte[] aadPrefix, int rowGroupOrdinal, int columnOrdinal)
```

I'll have a closer look at the details, but we need a unitest (proposed in my other comment) to make sure decryption works ok with async io and parallel column reading.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1796,5 +1882,314 @@ public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOEx
...
+    private final LinkedBlockingDeque<Optional<DataPage>> pagesInChunk = new LinkedBlockingDeque<>();
+    private DictionaryPage dictionaryPage = null;
+    private int pageIndex = 0;
+    private long valuesCountReadSoFar = 0;
+    private int dataPageCountReadSoFar = 0;
+
+    // derived
+    private final PrimitiveType type;
+    private final byte[] dataPageAAD;
+    private final byte[] dictionaryPageAAD;
```
ggershinsky, line 1918: probably not needed

parthchandra (commented):
Took me a while to get around to addressing all these.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -126,6 +127,42 @@ public class ParquetFileReader implements Closeable {
...
+  // Thread pool to read column chunk data from disk. Applications should call setAsyncIOThreadPool
+  // to initialize this with their own implementations.
+  // Default initialization is useful only for testing
+  public static ExecutorService ioThreadPool = Executors.newCachedThreadPool(
+    r -> new Thread(r, "parquet-io"));
+
+  // Thread pool to process pages for multiple columns in parallel. Applications should call
+  // setAsyncProcessThreadPool to initialize this with their own implementations.
+  // Default initialization is useful only for testing
+  public static ExecutorService processThreadPool = Executors.newCachedThreadPool(
```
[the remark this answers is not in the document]
ggershinsky: not sure; looks like many tests use copy/paste, rather than extension..

parthchandra:
@wgtmac thank you for looking at this. I don't have any more TODOs on this PR. 

> * Adopt the incoming Hadoop vectored io api.

This should be part of another PR. There is a draft PR (#999 ) open for this. Once that is merged in, I can revisit the async I/O code and incorporate the vectored io api. 
In other experiments I have seen that async io gives better results over slower networks. With faster network connections, as  is the case where we are reading from S3 within an AWS environment, reading in parallel (as the vector io api does), starts to give better results. 
I believe, that both should be available as options. 

> * Benchmark against remote object stores from different cloud providers.

The numbers I posted earlier were for reading from AWS/S3 over a 1 Gbps line. Reading from within AWS shows lesser improvement. I don't have an account with other cloud providers. Any help here would be appreciated. 

> IMO, switching `ioThreadPool` and `processThreadPool` the reader instance level will make it more flexible.

I've changed the thread pool so that it is not initialized by default but I left them as static members. Ideally, there should be a single IO thread pool that handles all the IO for a process and the size of the pool is determined by the bandwidthof the underlying storage system. 
Making them per instance is not an issue though. The calling code can decide to set the same thread pool for all instances and achieve the same result. 
Let me update this. 

Also, any changes you want to make are fine with me, and the help is certainly appreciated !

parthchandra:
> > IMO, switching `ioThreadPool` and `processThreadPool` the reader instance level will make it more flexible.
> 
> I've changed the thread pool so that it is not initialized by default but I left them as static members. Ideally, there should be a single IO thread pool that handles all the IO for a process and the size of the pool is determined by the bandwidthof the underlying storage system. Making them per instance is not an issue though. The calling code can decide to set the same thread pool for all instances and achieve the same result. Let me update this.
> 
> Also, any changes you want to make are fine with me, and the help is certainly appreciated !

I'm thinking of merging the thread pools into a single ioThreadPool and making the ioThreadPool settable thru `ParquetReadOptions` (like the allocator is). The work being done by the processThreadPool is rather small and maybe we can do away with it. 
Adding the pool via `ParquetReadOptions`  makes it easier to use with `ParquetReader` (used a lot in unit tests).
WDYT?

wgtmac:
> > > IMO, switching `ioThreadPool` and `processThreadPool` the reader instance level will make it more flexible.
> > 
> > 
> > I've changed the thread pool so that it is not initialized by default but I left them as static members. Ideally, there should be a single IO thread pool that handles all the IO for a process and the size of the pool is determined by the bandwidthof the underlying storage system. Making them per instance is not an issue though. The calling code can decide to set the same thread pool for all instances and achieve the same result. Let me update this.
> > Also, any changes you want to make are fine with me, and the help is certainly appreciated !
> 
> I'm thinking of merging the thread pools into a single `ioThreadPool` and making it settable thru `ParquetReadOptions` (like the allocator is). The work being done by the `processThreadPool` is rather small and maybe we can do away with it. Adding the pool via `ParquetReadOptions` makes it easier to use with `ParquetReader` (used a lot in unit tests). WDYT?

Sorry for my late reply.

Setting the thread pools via `ParquetReadOptions` is a good idea and that is exactly the way I want to do them away with static members. Merging `ioThreadPool` and `processThreadPool` into a single pool should work if the tasks in the `processThreadPool` do not wait for the return of tasks in the `ioThreadPool`. I will look into the detail later.

BTW, I don't have the permission to directly update your PR in place as I am not yet a maintainer of the repo. I may need to open a new one by copying what you have done here and add you as a co-author. WDYT? If that sounds good to you, I can proceed. @parthchandra

parthchandra:
Let me verify that merging the thread pools is not dangerous and post an update. After that if be great if you can take it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/FilePageReader.java
```diff
@@ -0,0 +1,361 @@
...
+            );
+          // Copy crc to new page, used for testing
+          if (pageHeader.isSetCrc()) {
+            compressedDictionaryPage.setCrc(pageHeader.getCrc());
+          }
+          dictionaryPage = compressedDictionaryPage;
+          break;
+        case DATA_PAGE:
+          DataPageHeader dataHeaderV1 = pageHeader.getData_page_header();
+          pageBytes = chunk.readAsBytesInput(compressedPageSize);
```
wgtmac, line 204: IIUC, the task in the `processThreadPool` is blocked here to wait for the completion of task in the `ioThreadPool`.

wgtmac: `chunk.readAsBytesInput(compressedPageSize)` is running inside `processThreadPool ` and what it does is as below:
```
    public BytesInput readAsBytesInput(int size) throws IOException {
      if (LOG.isDebugEnabled()) {
        String mode = (isAsyncIOReaderEnabled()) ? "ASYNC" : "SYNC";
        LOG.debug("{} READ BYTES INPUT: stream {}", mode, stream);
      }
      return BytesInput.from(stream.sliceBuffers(size));
    }
```

The `stream.sliceBuffers(size)` is a blocking call to `SequenceByteBufferInputStream.sliceBuffers()` which calls `AsyncMultiBufferInputStream.sliceBuffers()` internally and waits for return of `AsyncMultiBufferInputStream.nextBuffer()`. Please note that `AsyncMultiBufferInputStream.nextBuffer()` is running in the `ioThreadPool `.

wgtmac: In the current implementation, tasks of `ioThreadPool` are submitted first and then tasks of `processThreadPool` get submitted. If we merge these two thread pools into a single one, problem will happen due to uncertainty of the execution order in the ExecutorService. When the tasks that are previously submitted to `processThreadPool` are running first and have exhausted all the threads in the pool, they will be unable to proceed because none of the I/O tasks manage to get an available thread. Please correct me if I am wrong. Thanks! @parthchandra

parthchandra: You're correct. There is the possibility of a deadlock because the process threads do need to wait on the output of the io threads. So better to have different pools. I realized this myself after I had posted the above, but I'm glad you double checked!

parthchandra:
Updated. The thread pools are now set thru ParquetReadOptions

wgtmac (approved):

wgtmac:
> Updated. The thread pools are now set thru ParquetReadOptions

Thanks for addressing this!

I agree that adopting Hadoop vectored IO is an orthogonal topic and can be a separate patch when the API is ready. So I believe this patch is complete and have voted +1. Any improvement can be added via follow-up patches. I may request other maintainers for a final approval. @shangxinli

parthchandra:
@wgtmac  Thank you!

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+      try {
+        readResult = readFutures.poll();
+        readResult.get();
+        if (!readResult.isDone() && !readResult.isCancelled()) {
+          readResult.cancel(true);
+        } else {
+          readResult.get(1, TimeUnit.MILLISECONDS);
+        }
+      } catch (Exception e) {
+        // Do nothing
```
shangxinli, line 156: Do you think add a log make sense?

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+      fetchIndex++;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public boolean nextBuffer() {
+    checkState();
+    // hack: parent constructor can call this method before this class is fully initialized.
```
shangxinli, line 110: I see the comment with 'hack'. What is the proper implementation?

shangxinli:
@kazuyukitanimura @steveloughran @kbendick @ggershinsky @wgtmac @theosib-amazon Do you still have comments?

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
parthchandra: It's not really useful to log anything here because there is nothing to be done. Some of the calls in the try block throw exceptions and we catch them to prevent them from propagating. We are shutting down and cleaning up the futures so it doesn't matter if there was an exception or not.

parquet-common/src/main/java/org/apache/parquet/bytes/AsyncMultiBufferInputStream.java
parthchandra: The fix would be to change behavior of the parent class. Basically the parent class implicitly assumes that there will be no derived classes and called a private method `nextBuffer` in the constructor. I extended the class and made next buffer protected so it is no longer correct to call it in the constructor. 
Changing the parent class behavior broke too many unit tests. I could have fixed the unit tests as well, but wasn't sure that downstream projects might be depending on the behavior. 
In the end, it seemed that handling this in the derived class was a safer approach. 
It's really a workaround to make things easier to implement. Perhaps I should not have called it a hack.

wgtmac:
> @kazuyukitanimura @steveloughran @kbendick @ggershinsky @wgtmac @theosib-amazon Do you still have comments?

It looks good to me. Please feel free to merge as you see fit. @shangxinli

whcdjj:
Hi, I am very interested in this optimization and just have some questiones when testing in a cluster with 4nodes/96 cores using spark3.1.  Unfortunately， I see little improvement.
I am confused than whether it is neccessary to keep spark.sql.parquet.enableVectorizedReader = false in spark when testing with spark 3.2 and how can i set the parquet buffer size. Sincerely ask for advice @parthchandra

parthchandra:
> Hi, I am very interested in this optimization and just have some questions when testing in a cluster with 4nodes/96 cores using spark3.1. Unfortunately， I see little improvement.

You're likely to see improvement in cases where file i/o is the bottleneck. Most TPC-DS queries are join heavy and you will see little improvement there. You might do better with TPC-H. 

> I am confused than whether it is necessary to keep spark.sql.parquet.enableVectorizedReader = false in spark when testing with spark 3.1 and how can I set the parquet buffer size. 

It's probably best to keep the parquet (read) buffer size untouched.

You should keep `spark.sql.parquet.enableVectorizedReader = true` irrespective of this. This feature improves I/O speed of reading raw data. The Spark vectorized reader kicks in after data is read from storage and converts the raw data into Spark's internal columnar representation and is faster than the row based version.

whcdjj:
> My test is spark.sql("select * from store_sales order by ss_customer_sk limit 10")， store_sales is table of 1TB TP-CDS.

Parquet-io  and parquet-process  threads is hardcoded given like this in ParquetReadOptions.java
public static class Builder {
    protected ExecutorService ioThreadPool = Executors.newFixedThreadPool(4);
    protected ExecutorService processThreadPool = Executors.newFixedThreadPool(4);
}

I also take the follow test with local filesystem using 100GB TP-CDS store_sales table，and I see there is a degradation with async io feature. 
test("parquet reader select") {
    val sc = SparkSession.builder().master("local[4]").getOrCreate()
    val df = sc.read.parquet("file:///D:\\work\\test\\tpcds\\store_sales")
    df.createOrReplaceTempView("table")
    val start = System.currentTimeMillis()
    sc.sql("select * from table order by ss_customer_sk limit 10").show()
    val end = System.currentTimeMillis()
    System.out.println("time: " + (end - start))
  }
without this feature -> time: 7240
with this feature -> time: 19923

Threads are as expected 
image

What process did I go wrong and can you show me the correct way to use this feature?

parthchandra:
Looks correct to me. Couple of questions, are you running this on a cluster or on local system? Also, is the data on SSD's? If you are on a single machine, there might not be enough CPU and the async threads may be contending for time with the processing threads. We'll need some profiling info to get a better diagnosis. Also, with SSD's, reads are so fast from the file system that the async feature might show very little improvement. 
You could turn on the debug logging level for FilePageReader and AsyncMultibufferInputStream.
We can continue this in a thread outside this PR.

whcdjj:
> Looks correct to me. Couple of questions, are you running this on a cluster or on local system? Also, is the data on SSD's? If you are on a single machine, there might not be enough CPU and the async threads may be contending for time with the processing threads. We'll need some profiling info to get a better diagnosis. Also, with SSD's, reads are so fast from the file system that the async feature might show very little improvement. You could turn on the debug logging level for FilePageReader and AsyncMultibufferInputStream. We can continue this in a thread outside this PR.

OK，I will test with the debug logging for more details these days.

parthchandra:
Also try a query like 
```
  select
        SUM(length(IFNULL(ss_sold_date_sk, ' '))),
        SUM(length(IFNULL(ss_sold_time_sk, ' '))),
        SUM(length(IFNULL(ss_item_sk, ' '))),
        SUM(length(IFNULL(ss_customer_sk, ' '))),
        SUM(length(IFNULL(ss_cdemo_sk, ' '))),
        SUM(length(IFNULL(ss_hdemo_sk, ' '))),
        SUM(length(IFNULL(ss_addr_sk, ' '))),
        SUM(length(IFNULL(ss_store_sk, ' '))),
        SUM(length(IFNULL(ss_promo_sk, ' '))),
        SUM(length(IFNULL(ss_ticket_number, ' '))),
        SUM(ss_quantity),
        SUM(ss_wholesale_Cost),
        SUM(ss_list_price ),
        SUM(ss_sales_price),
        SUM(ss_ext_discount_amt),
        SUM(ss_ext_sales_price),
        SUM(ss_ext_wholesale_cost),
        SUM(ss_ext_list_price),
        SUM(ss_ext_tax),
        SUM(ss_coupon_amt),
        SUM(ss_net_paid),
        SUM(ss_net_paid_inc_tax),
        SUM(ss_net_profit)
from store_sales

```
which avoids the expensive sort

hazelnutsgz:
Sorry to interrupt, just wondering if this PR can work with [S3AsyncClient](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3AsyncClient.html)  by any chance? @parthchandra

parthchandra:
> Sorry to interrupt, just wondering if this PR can work with [S3AsyncClient](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3AsyncClient.html) by any chance? @parthchandra
> 
> Thanks!

This PR uses the hdfs interface to s3 (s3a) if the url is s3a. I don't think that the s3a implementation uses any of the implementations of S3AsyncClient.

wgtmac:
@parthchandra Do you have time to resolve the conflicts? I think it would be nice to be included in the next release.

parthchandra:
> @parthchandra Do you have time to resolve the conflicts? I think it would be nice to be included in the next release.

Done

parthchandra:
FWIW the hadoop 3.3.5 vector io changes might make this PR redundant.

Fokko:
@steveloughran 🤚🏻 Looking forward to your talk!

parquet-hadoop/src/main/java/org/apache/parquet/HadoopReadOptions.java
```diff
@@ -52,18 +55,23 @@ private HadoopReadOptions(boolean useSignedStringMinMax,
                             boolean useColumnIndexFilter,
                             boolean usePageChecksumVerification,
                             boolean useBloomFilter,
+                            boolean enableAsyncIOReader,
```
steveloughran, line 58: life would be a lot simpler if this code moved to a parameter object containing all the args: adding new options would be straightforward. We did this for the S3ClientFactory to stop external implementations having incompatible signatures when new stuff went in

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -118,16 +124,21 @@ public long getTotalValueCount() {
 
     @Override
     public DataPage readPage() {
-      final DataPage compressedPage = compressedPages.poll();
+      final DataPage compressedPage;
+      try {
+        compressedPage = compressedPages.take().orElse(null);
+      } catch (InterruptedException e) {
+        throw new RuntimeException("Error reading parquet page data.") ;
```
steveloughran, line 131: nit: include caught exception

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/FilePageReader.java
steveloughran: i've had to do the pool pair in other code to avoid this. what we also do is (and i forget where...) collect statistics on the time spent waiting for an executor. This helps identify when pool size is limiting performance.

steveloughran (commented):
finally sat down to scan through this; I think I'd need to look some more at this to understand it and compare with vector IO.

parthchandra:
Thanks for the comments @steveloughran. Not sure if we are still thinking of merging this. I'm not sure how well this will play with vector io.
I'll spend some time on this in the next few days.

steveloughran:
thanks. really looking at how to maximise parquet cloud read perf right now. IF you have traces, flame graphs etc that'd be good, especially for queries with different types and selectivity.

parthchandra:
I'll be able to spend some time on this this week (I bet it will not rebase on top of latest without a ton of conflicts). 
Re flamegraphs etc. I don't have any with async io, but from past efforts, I see too much noise from the execution engine. Mostly what you see in the flame graph is the time spent is in waiting and copying data into read buffers. The benefits from parallelism and from threads not waiting for input are not so obvious. I'll give it another shot and share.

steveloughran:
> Mostly what you see in the flame graph is the time spent is in waiting and copying data into read buffers. 

on heap or off-heap buffers?

* vector IO through s3a or local fs will read into bytebuffers direct
* https://github.com/apache/hadoop/pull/6686 will export ByteBufferPositionedReadable and its readFully() command, which could speed up HDFS reads

Recent profiling of Hive Tez has shown that because it closes all fs instances in worker process after sub-query execution, time to instantiate S3A filesystem and negotiate TLS connections become significant
https://github.com/apache/hadoop/pull/6892 aims to cut startup/shutdown time in half *provided nothing calls rename()*, but really they should either stop deleting the instances or cache them until the next sub query come in, and if its from the same user: recycle, else delete asynchronously.

parthchandra:
> > Mostly what you see in the flame graph is the time spent is in waiting and copying data into read buffers.
> 
> on heap or off-heap buffers?
> 
On heap.

parthchandra:
@steveloughran I rebased this so now we can read using either vectorIO and async reader (but not both). Haven't been able to find time to compare.

steveloughran:
I think it'd be great to have a benchmark which could target remote stores; the current one does local where the latencies are less so the differences less tangible (though vector IO is fast with SSD as you can do parallel reads directly into buffers)

PR #969, opened 2022-05-18
PARQUET-2148: Enable uniform decryption with plaintext footer

ggershinsky:
Currently, uniform decryption is not enabled in the plaintext footer mode - for no good reason. Column metadata is available, we just need to decrypt and use it.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -21,10 +21,12 @@ package org.apache.parquet.format.converter;
 import static java.util.Optional.empty;
 
 import static java.util.Optional.of;
+import static org.apache.parquet.format.Util.readColumnMetaData;
 import static org.apache.parquet.format.Util.readFileMetaData;
 import static org.apache.parquet.format.Util.writeColumnMetaData;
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1556,25 +1558,32 @@ public ParquetMetadata fromParquetMetadata(FileMetaData parquetMetadata,
...
+                throw new ParquetCryptoRuntimeException("Column encrypted with footer key: No keys available");
               }
               if (null == metaData) {
                 throw new ParquetCryptoRuntimeException("ColumnMetaData not set in Encryption with Footer key");
               }
-              if (null == fileDecryptor) {
-                throw new ParquetCryptoRuntimeException("Column encrypted with footer key: No keys available");
-              }
               columnPath = getPath(metaData);
+              if (!encryptedFooter) { // Unencrypted footer. Decrypt full column metadata, using footer key
```
shangxinli, line 1568: I am not following why we use the footer key to decrypt? Wasn't the column key used before?

ggershinsky: with uniform encryption, we leverage the "encrypt_with_footer_key" mode in parquet-format

PR #975, opened 2022-06-12
PARQUET-2157: add bloom filter fpp config

huaxingao:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2157

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -19,6 +19,7 @@
 package org.apache.parquet.column;
 
 import java.util.Objects;
+import java.util.OptionalDouble;
 import java.util.OptionalLong;
 
 import org.apache.parquet.Preconditions;
@@ -57,6 +58,7 @@ public class ParquetProperties {
   public static final int DEFAULT_PAGE_ROW_COUNT_LIMIT = 20_000;
   public static final int DEFAULT_MAX_BLOOM_FILTER_BYTES = 1024 * 1024;
   public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
+  public static final double DEFAULT_BLOOM_FILTER_FPP = 0.01;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
@@ -100,6 +102,7 @@ public class ParquetProperties {
 
   // The expected NDV (number of distinct values) for each columns
   private final ColumnProperty<Long> bloomFilterNDVs;
+  private final ColumnProperty<Double> bloomFilterFPPs;
   private final int maxBloomFilterBytes;
   private final ColumnProperty<Boolean> bloomFilterEnabled;
   private final int pageRowCountLimit;
@@ -122,6 +125,7 @@ public class ParquetProperties {
     this.columnIndexTruncateLength = builder.columnIndexTruncateLength;
     this.statisticsTruncateLength = builder.statisticsTruncateLength;
     this.bloomFilterNDVs = builder.bloomFilterNDVs.build();
+    this.bloomFilterFPPs = builder.bloomFilterFPPs.build();
     this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
     this.pageRowCountLimit = builder.pageRowCountLimit;
@@ -258,6 +262,11 @@ public class ParquetProperties {
     return ndv == null ? OptionalLong.empty() : OptionalLong.of(ndv);
   }
 
+  public OptionalDouble getBloomFilterFPP(ColumnDescriptor column) {
+    Double fpp = bloomFilterFPPs.getValue(column);
+    return fpp == null ? OptionalDouble.empty() : OptionalDouble.of(fpp);
+  }
+
   public boolean isBloomFilterEnabled(ColumnDescriptor column) {
     return bloomFilterEnabled.getValue(column);
   }
@@ -288,6 +297,7 @@ public class ParquetProperties {
         + "Bloom filter enabled: " + bloomFilterEnabled + '\n'
         + "Max Bloom filter size for a column is " + getMaxBloomFilterBytes() + '\n'
         + "Bloom filter expected number of distinct values are: " + bloomFilterNDVs + '\n'
+        + "Bloom filter false positive probabilities are: " + bloomFilterFPPs + '\n'
         + "Page row count limit to " + getPageRowCountLimit() + '\n'
         + "Writing page checksums is: " + (getPageWriteChecksumEnabled() ? "on" : "off");
   }
@@ -305,6 +315,7 @@ public class ParquetProperties {
     private int columnIndexTruncateLength = DEFAULT_COLUMN_INDEX_TRUNCATE_LENGTH;
     private int statisticsTruncateLength = DEFAULT_STATISTICS_TRUNCATE_LENGTH;
     private final ColumnProperty.Builder<Long> bloomFilterNDVs;
+    private final ColumnProperty.Builder<Double> bloomFilterFPPs;
     private int maxBloomFilterBytes = DEFAULT_MAX_BLOOM_FILTER_BYTES;
     private final ColumnProperty.Builder<Boolean> bloomFilterEnabled;
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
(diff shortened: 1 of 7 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -39,13 +39,17 @@
 import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
 import net.openhft.hashing.LongHashFunction;
+import org.apache.commons.lang3.RandomStringUtils;
```
dongjoon-hyun, line 48: Please add this as a test dependency to `parquet-hadoop/pom.xml`.
```
    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-lang3</artifactId>
      <version>3.9</version>
      <scope>test</scope>
    </dependency>
```

huaxingao:
cc @chenjunjiedada @ggershinsky @shangxinli

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -282,6 +286,63 @@ public void testParquetFileWithBloomFilter() throws IOException {
     }
   }
 
+  @Test
+  public void testParquetFileWithBloomFilterWithFpp() throws IOException {
+    final int totalCount = 100000;
```
chenjunjiedada, line 291: Nit: Why do we need `final`?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -282,6 +286,63 @@ public void testParquetFileWithBloomFilter() throws IOException {
...
+    for (int i = 0; i < testFpp.length; i++) {
+      File file = temp.newFile();
+      file.delete();
+      Path path = new Path(file.getAbsolutePath());
+      try (ParquetWriter<Group> writer = ExampleParquetWriter.builder(path)
+        .withPageRowCountLimit(10)
+        .withConf(conf)
+        .withDictionaryEncoding(false)
+        .withBloomFilterEnabled("name", true)
+        .withBloomFilterNDV("name", 100000l)
```
chenjunjiedada, line 316: Nit: Can we use `TotalCount`?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
huaxingao: Removed.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
huaxingao: Fixed. Thanks!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -282,6 +286,63 @@ public void testParquetFileWithBloomFilter() throws IOException {
...
+        int exist = 0;
+        while (distinctStrings.size() < totalCount) {
+          String str = RandomStringUtils.randomAlphabetic(10);
+          if (distinctStrings.add(str) &&
+            bloomFilter.findHash(LongHashFunction.xx(0).hashBytes(Binary.fromString(str).toByteBuffer()))) {
+            exist++;
+          }
+        }
+        // The exist should be less than totalCount * fpp. Add 10% here for error space.
+        assertTrue(exist < totalCount * (testFpp[i] * 1.1));
```
ggershinsky, line 341: just curious if `totalCount` is sufficient; how often `exist` > 0?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -282,6 +286,63 @@ public void testParquetFileWithBloomFilter() throws IOException {
...
+
+      try (ParquetFileReader reader = ParquetFileReader.open(HadoopInputFile.fromPath(path, new Configuration()))) {
+        BlockMetaData blockMetaData = reader.getFooter().getBlocks().get(0);
+        BloomFilter bloomFilter = reader.getBloomFilterDataReader(blockMetaData)
+          .readBloomFilter(blockMetaData.getColumns().get(0));
+
+        // The exist counts the number of times FindHash returns true.
+        int exist = 0;
+        while (distinctStrings.size() < totalCount) {
+          String str = RandomStringUtils.randomAlphabetic(10);
```
ggershinsky, line 334: the original values are 12 char long. Are we supposed to find a 10-char string among them?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -471,6 +484,12 @@ public Builder withBloomFilterNDV(String columnPath, long ndv) {
       return this;
     }
 
+    public Builder withBloomFilterFPP(String columnPath, double fpp) {
```
ggershinsky, line 487: what happens if this value is set, but the BF is not enabled? (general / per-column)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
ggershinsky: Two related questions:
- what should be the `totalCount` to reliably ensure that a) `exist > 0` b) `exist < totalCount * (testFpp[i] * 1.1)` ? Depending on the `fpp` value, we can get a random assert exception if `totalCount` is too low (also, `exist` could be just 0 then). If `totalCount` is high, the unitest could take a very long time. 
- how long does this unitest run on your laptop? (with the current `totalCount` of 100000).

huaxingao: Basically `exist` > 0 is false positive. which happens when any given hash value that was never inserted into the bloom filter causes the check to return true. I don't think there is a simple closed-form calculation of this probability, but setting `totalCount` to be `100000` seems to be a pretty safe number for the test to pass.

I am thinking we probably should disallow the Bloom filter's size to be unreasonably small. We currently only have the 
maximum bytes of the Bloom filter. Shall we also have the minimum bytes of the Bloom filter? What do you think? @chenjunjiedada 

The test takes about 2300 milli seconds on my laptop.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
huaxingao: Changed. Thanks!

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
huaxingao: This value will be silently ignored.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
chenjunjiedada: The size of the bloom filter is computed with `ndv` and `fpp`. So even the size is "unreasonable" small it should be enough to handle the given situation. Right?

ggershinsky:
> The test takes about 2300 milli seconds on my laptop.

Ok, this is reasonable. If this time is sufficient for reliably testing the upper limit of FPPs, it should be good enough to also check the lower limit, eg `exist > totalCount * (testFpp[i] * 0.9)` , or `exist > totalCount * (testFpp[i] * 0.5)` , or even `exist > 0`. What do you think? This way, we'll be certain the test passes not because `exist` is just 0.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
huaxingao: Yes. Agree.

huaxingao:
> it should be good enough to also check the lower limit, eg exist > totalCount * (testFpp[i] * 0.9) , or exist > totalCount * (testFpp[i] * 0.5) , or even exist > 0. What do you think? This way, we'll be certain the test passes not because exist is just 0.

Thanks for the suggestion! I can't find a reliable number for the lower limit. I put `exist > 0`.

ggershinsky (approved):

shangxinli:
LGTM

huaxingao:
Thank you all very much! @chenjunjiedada @dongjoon-hyun @ggershinsky @shangxinli

dongjoon-hyun:
Could you resolve JIRA please? I realized that the JIRA is still open although this is delivered.
- https://issues.apache.org/jira/browse/PARQUET-2157

PR #976, opened 2022-06-13
PARQUET-2158: Upgrade Hadoop dependency to version 3.2.0

steveloughran:
This updates Parquet's Hadoop dependency to 3.2.0.
This version adds compatibility with Java 11, as well
as many other features and bug fixes.

### Jira

### Tests!

  it's a dependency update.

### Commits

### Documentation

### DIFF
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/PathGlobPattern.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/PathGlobPattern.java
@@ -1,4 +1,4 @@
-/* 
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -20,8 +20,8 @@ package org.apache.parquet.thrift.projection.deprecated;
 
 import org.apache.hadoop.fs.GlobPattern;
 
-import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
+import com.google.re2j.Pattern;
+import com.google.re2j.PatternSyntaxException;
 
 /**
  * Enhanced version of GlobPattern class that is defined in hadoop with extra capability of matching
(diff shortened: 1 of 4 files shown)


steveloughran:
thrift module doesn't compile is using an hadoop internal class tagged as private & which made an incompatible change in hadoop 3. see  HADOOP-12436

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project parquet-thrift: Compilation failure
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/PathGlobPattern.java:[55,49] incompatible types: com.google.re2j.Pattern cannot be converted to java.util.regex.Pattern
```

the good news, the class is deprecated, which explains why nobody has seen it in the wild. Any attempt to use that class would fail with hadoop 3.x on the classpath.

steveloughran:
This PR fixes Parquet to build/link against Hadoop 3.2.0 and higher. It would be cleaner to remove the deprecated class causing compatibility issues -the fact that nobody has ever reported linkage errors implies it is not in active use

pom.xml
```diff
@@ -76,7 +76,7 @@
     <jackson-databind.version>2.13.2.2</jackson-databind.version>
     <japicmp.version>0.14.2</japicmp.version>
     <shade.prefix>shaded.parquet</shade.prefix>
-    <hadoop.version>2.10.1</hadoop.version>
+    <hadoop.version>3.2.0</hadoop.version>
```
sunchao, line 79: hmm why 3.2.0, not 3.3.1/3.3.2?

shangxinli: +1 for the question

parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/PathGlobPattern.java
```diff
@@ -20,8 +20,8 @@
 
 import org.apache.hadoop.fs.GlobPattern;
 
-import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
+import com.google.re2j.Pattern;
```
sunchao, line 23: I think this may not work for projects like Spark who are using Hadoop shaded client, since the `GlobPattern.compiled` is relocated to `org.apache.hadoop.shaded.com.google.re2j.Pattern`.

It might be easier to just remove the class as it has been marked as deprecated since Parquet 1.8.0, 2015

steveloughran: +1 for cutting. i will update the patch

pom.xml
steveloughran: I was being unambitious. move to this, the oldest 3.x release working on java11 ensures that anything else on a version >= to this should link properly.

if you do want to be more current, well, spark is on 3.3.3, hive is trying to move to 3.3.x and I will be doing a 3.3.4 release in a week's time, which is just some security changes mostly of relevance to servers

steveloughran:
i will do a separate PR to remove `PathGlobPattern`; not this week though. 

It is used in DeprecatedFieldProjectionFilter, and that is used in org.apache.parquet.hadoop.thrift.ThriftReadSupport if "parquet.thrift.column.filter" is set. that use would have to be cut and rather than just print a deprecation warning, actually fail.

nobody must be using this on anything with ASF hadoop binaries 3.2+ or they would have complained about linkage errors by now.

shangxinli:
LGTM

steveloughran:
thanks.

PR #979, opened 2022-07-12
PARQUET-2165: Remove deprecated PathGlobPattern class

steveloughran:
Remove the deprecated classes PathGlobPattern and
DeprecatedFieldProjectionFilter so that Parquet will
compile against hadoop 3.x.

If a thrift reader is configured to use the now-deleted filter,
by setting the filter in "parquet.thrift.column.filter",
a ThriftProjectionException will be thrown.

### Jira

### Tests

It modifies the test `TestParquetToThriftReadWriteAndProjection` to switch to the strict filter in all test cases where the old one was being used.

*these tests now all fail with `ThriftProjectionException: No columns have been selected`

I could cut the tests "obsolete" but it would seem to me that moving the tests to the strict filter would be better. I will just need help doing this.

### Commits

### Documentation


Fokko:
@steveloughran is this something that you still want to get in?

Fokko:
Seems to conflict with https://github.com/apache/parquet-mr/pull/1076

steveloughran:
This needs to be disposed of for building against more recent hadoop versions, so someone has to do it..

PR #982, opened 2022-08-08
PARQUET-2160: Close ZstdInputStream to free off-heap memory in time.

zhongyujiang:

### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2160) issues and references them in the PR title. For example, "PARQUET-1234: My Parquet PR"
  - https://issues.apache.org/jira/browse/PARQUET-2160

### Tests

### Commits

### Documentation

### Benchmark

**Benchmark:** 
The benchmark data consists of 500 string columns * 1_000_000  rows(10.85GB under ZSTD compression), and the benchmark is simply reading all the data out and pass them to the black hole.

**JMH Configuration:**
3 rounds of warmup iterations
5 rounds of  measurenemt iterations

**Notes:** 
> - Closing the input stream means reading the decompressed data into the heap early and closing the input stream, which is the method proposed in this pr. Not closing the stream means leaving the stream to the GC. 
>- The JMH Score in the table below is the average of multiple rounds of measurement, while GC Events and GC Time are the statistical values of the total measurement process (3 rounds of Warmup + 5 rounds of Measurement)

According to the benchmark results below, _closing stream_ does not bring additional heap overload, but improves GC instead. From the comparison of the first two results of below table, we can see that _closing stream_ improves the read performance and reduces the GC load.

I also measured the impact of heap size (GC) on read performance. After reducing heap memory to 3GB, GC Events increased, but the read performance is not decreased; However, after increasing heap memory to 5GB, GC Events decreased by about 30% , but the GC time has increased by about 2 times, and the read performance has also been significantly reduced, which may verify the previous guess in Jira: Untimely GC may exacerbate the off-heap memory fragmentation issue. (Of course, this may require more in-depth testing to corroborate)

<div class="okr-block-clipboard" data-okr="%7B%22okrDelta%22%3A%5B%7B%22lineType%22%3A%22unsupport%22%2C%22lineOptions%22%3A%7B%7D%2C%22lineContent%22%3A%5B%5D%7D%5D%2C%22businessKey%22%3A%22lark-doc%22%7D"></div><div data-zone-id="0" data-line-index="0" style="white-space: pre;">

Compression | Close Inputstream | Heap Size(GB) | JMH Score ± Error (s/op) | GC Events | GC Time(s)
-- | -- | -- | -- | -- | --
ZSTD | No | 4 | 53.374 ± 3.303 | 1591 | 40.069
ZSTD | Yes | 4 | 49.613 ± 2.564 | 928 | 17.194
ZSTD | No | 3 | 53.303 ± 1.431 | 2174 | 46.121
ZSTD | No | 5 | 75.993 ± 10.919 | 950 | 116.184

</div>

Out of curiosity, I also tested the effect of _closing stream_ on Snappy and Gzip, the benchmark results show that in such simple read scenario _closing stream_ also has some positive effect on both Snappy and Gzip.

<div class="okr-block-clipboard" data-okr="%7B%22okrDelta%22%3A%5B%7B%22lineType%22%3A%22unsupport%22%2C%22lineOptions%22%3A%7B%7D%2C%22lineContent%22%3A%5B%5D%7D%5D%2C%22businessKey%22%3A%22lark-doc%22%7D"></div><div data-zone-id="0" data-line-index="0" style="white-space: pre;">

Compression | Close Inputstream | Heap Size(GB) | JMH Score ± Error (s/op) | GC Events | GC Time(s)
-- | -- | -- | -- | -- | --
Snappy | No | 4 | 38.835 ± 7.505 | 1618 | 40.193
Snappy | Yes | 4 | 33.796 ± 2.871 | 872 | 12.449
Gzip | No | 4 | 94.902 ± 5.833 | 1539 | 39.630
Gzip | Yes | 4 | 86.160 ± 2.329 | 843 | 12.153

</div>

For more details, please refer to the following pictures:

**ZSTD + Not close + 4GB**
image
image

**ZSTD + Close + 4GB**
image
image

**ZSTD + Not close + 3GB**
image
image

**ZSTD + Not close + 5GB**
image
image

**Snappy + Not close + 4GB**
image
image

**Snappy + Close + 4GB**
image
image

**Gzip + Not Close + 4GB**
image
image

**Gzip + Close + 4GB**
image
image

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.compression.CompressionCodecFactory;
+import org.apache.parquet.hadoop.codec.ZstandardCodec;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 
 public class CodecFactory implements CompressionCodecFactory {
@@ -109,7 +110,12 @@ public class CodecFactory implements CompressionCodecFactory {
           decompressor.reset();
         }
         InputStream is = codec.createInputStream(bytes.toInputStream(), decompressor);
-        decompressed = BytesInput.from(is, uncompressedSize);
+        if (codec instanceof ZstandardCodec) {
+          decompressed = BytesInput.copy(BytesInput.from(is, uncompressedSize));
+          is.close();
+        } else {
+          decompressed = BytesInput.from(is, uncompressedSize);
+        }
       } else {
         decompressed = bytes;
       }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
@@ -110,6 +110,11 @@ public class CodecFactory implements CompressionCodecFactory {
           decompressor.reset();
         }
         InputStream is = codec.createInputStream(bytes.toInputStream(), decompressor);
+
+        // We need to explicitly close the ZstdDecompressorStream here to release the resources it holds to avoid
+        // off-heap memory fragmentation issue, see https://issues.apache.org/jira/browse/PARQUET-2160.
+        // This change will load the decompressor stream into heap a little earlier, since the problem it solves
+        // only happens in the ZSTD codec, so this modification is only made for ZSTD streams.
         if (codec instanceof ZstandardCodec) {
           decompressed = BytesInput.copy(BytesInput.from(is, uncompressedSize));
           is.close();


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -109,7 +110,12 @@ public BytesInput decompress(BytesInput bytes, int uncompressedSize) throws IOEx
           decompressor.reset();
         }
         InputStream is = codec.createInputStream(bytes.toInputStream(), decompressor);
-        decompressed = BytesInput.from(is, uncompressedSize);
+        if (codec instanceof ZstandardCodec) {
```
zhongyujiang, line 113: This looks a little weird, but considering doing so will load the decompressor stream into heap in advance, and only zstd has this problem currently, so I made this modification only for zstd stream.

zhongyujiang: Maybe we can consider closing the decompressed stream after it has been read:
https://github.com/apache/parquet-mr/blob/0819356a9dafd2ca07c5eab68e2bffeddc3bd3d9/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java#L283-L288
But I'm not sure if there is a situation where the decompressed stream is read more than once.

sunchao: The change looks OK to me, we probably should add some comments explaining why ZSTD deserves the special treatment here. 

The change on `BytesInput` looks more intrusive since it is used not only for decompression but other places like compression. For instance, `BytesInput.copy` calls `toByteArray` underneath, and after the call the original object should still be valid.

sunchao (commented):
cc @shangxinli @ggershinsky too

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
zhongyujiang: Added comment.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -109,7 +110,17 @@ public BytesInput decompress(BytesInput bytes, int uncompressedSize) throws IOEx
...
         }
         InputStream is = codec.createInputStream(bytes.toInputStream(), decompressor);
-        decompressed = BytesInput.from(is, uncompressedSize);
+
+        // We need to explicitly close the ZstdDecompressorStream here to release the resources it holds to avoid
+        // off-heap memory fragmentation issue, see https://issues.apache.org/jira/browse/PARQUET-2160.
+        // This change will load the decompressor stream into heap a little earlier, since the problem it solves
+        // only happens in the ZSTD codec, so this modification is only made for ZSTD streams.
+        if (codec instanceof ZstandardCodec) {
+          decompressed = BytesInput.copy(BytesInput.from(is, uncompressedSize));
```
shangxinli, line 119: I understand we had the discussion in the Jira that ByteInput.copy() just loads into a heap in advance but not add extra overall. Can we have a benchmark on the heap/GC(Heap size, GC time etc). I just want to make sure we fix one problem while introducing another problem.

zhongyujiang: > Can we have a benchmark on the heap/GC(Heap size, GC time etc).

Sure, will do a benchmark.

zhongyujiang: Hi @shangxinli , very sorry about the big delay, I was a little busy last week. The benchmark result and detailed data has been posted in the PR describe block, also cc @sunchao.

shangxinli: Thanks for working on it!

shangxinli:
LGTM

sunchao (approved):
LGTM too, thanks @zhongyujiang !

camper42:
wondering why this PR has not been merged

we have face this problem and are looking for a solution (spark 3.2.2+zstd-parquet)

sunchao:
cc @ggershinsky @gszadovszky @shangxinli

alexeykudinkin:
@gszadovszky @ggershinsky @shangxinli 

Folks, do we have an approximate timeline for the next patch release that will be including this patch? 

This is a severe problem that does affect our ability to use Parquet w/ Zstd and i'm aware of at least of handful of similar occasions and issues occurring to others.

shangxinli:
@alexeykudinkin We might release a new patch in the next 2 or 3 months.  

Can you elaborate why "this is a severe problem that does affect our ability to use Parquet w/ Zstd"? I understand it is an issue but we have been running Parquet w/ Zstd for 2+ years in production.

alexeykudinkin:
Totally @shangxinli  

We have running Spark clusters in production _ingesting_ from 100s of Apache Hudi tables (using Parquet and Zstd) and writing into other ones. We switched from gzip to zstd slightly over a month ago and we started to have OOM issues almost immediately. It took us a bit of triaging to zero in on zstd, but now we're confident that it's not mis-calibration of our configs but slow-bleeding leak of the native memory.

The crux of the problem is very particular type of the job -- one that reads a lot of Zstd compressed Parquet (therefore triggering the affected path). Other jobs not reading Parquet are not affected.

shangxinli:
Thanks @alexeykudinkin for the explanation.

Kimahriman:
same, we have certain jobs that can't function without a patched jar. Seems to get worse with the more columns you read. Our worst offender (table with thousands of columns), can easily hit 90+ GiB of resident memory on a 25 GiB heap executor in an hour or so (running on Yarn without memory limit enforcement)

pan3793:
I also encountered this memory leak when migrating data from parquet/snappy to parquet/zstd, Spark executors always occupy unreasonable off-heap memory and have a high risk of being killed by NM, this patch can solve this problem. It's a critical issue in my case, wish the upstream publish a patched version so that we don't need to maintain the internal version.

wangzzu:
@zhongyujiang 
We also encountered the same problem, here I have another question, why this problem only occurs in the parquet/zstd read

PR #983, opened 2022-08-08
WiP: parquet to use openfile api and some other performance enhancements

steveloughran:
This is me looking at what minimal changes could be made to
boost IO performance working with the cloud stores.

Compiles against hadoop 3.3.3; will need hadoop 3.3.5 for some
of the benefits

* not expecting any reviews here, just seeing how well it works for now *

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
@@ -27,6 +27,8 @@ import org.apache.parquet.io.SeekableInputStream;
 import org.apache.parquet.io.InputFile;
 import java.io.IOException;
 
+import static org.apache.hadoop.util.functional.FutureIO.awaitFuture;
+
 public class HadoopInputFile implements InputFile {
 
   private final FileSystem fs;
@@ -54,7 +56,7 @@ public class HadoopInputFile implements InputFile {
   public Configuration getConfiguration() {
     return conf;
   }
-  
+
   public Path getPath() {
     return stat.getPath();
   }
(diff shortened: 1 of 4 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -66,7 +68,13 @@ public long getLength() {
 
   @Override
   public SeekableInputStream newStream() throws IOException {
-    return HadoopStreams.wrap(fs.open(stat.getPath()));
+    // specify random access always
+    return HadoopStreams.wrap(
+      awaitFuture(fs.openFile(stat.getPath())
```
sunchao, line 73: hmm won't this cause issue if the actual hadoop version is below 3.3? I think we may need to check the version through `VersionInfo.getVersion`.

steveloughran: like I said, WiP 😄 

the shim library I'm doing does retrofit this to older versions, but I'm doing this to see what opportunities there are with basic adoption of 3.3.x apis. (https://github.com/steveloughran/fs-api-shim)

Mukund intends to add vectored stream reads to that shim lib too falling back to the positioned read operations, for versions <= 3.3.4

sunchao: Ah I see. Looking forward to that!

steveloughran: people should still be upgrading all their project dependencies though. 3.3.2 should be the minimum.

steveloughran:
not going to finish this

PR #987, opened 2022-08-18
Parquet-MR Encryption - Modify to true to encrypt

renshangtao:
I use iceberg to encryption parquet file，Then i find It will return "No encryption setup found for column [c1]". Both full table encryption and column encryption have this problem

Looking at the code, I can see that this parameter is fixed to false, which causes the encryption process to throw an exception directly


renshangtao:
@ggershinsky please review it. Thank you.

ggershinsky:
This breaks the parquet columnar encryption mode. We use the parquet "uniform" encryption mode instead for file encryption in Iceberg. Please have a look at https://github.com/apache/iceberg/pull/2639

PR #993, opened 2022-09-04
PARQUET-2184: Improve the allocation behavior of SnappyCompressor

abaranec:
This PR improves the allocation behavior of SnappyCompressor.  Previously when more buffer space was needed, it would only allocate enough for the new data to be written.  Now, it will double the internal buffer size up to 8MB, and then afterwards increase size in 1MB increments.

No additional unit tests are added, as the existing unit tests for SnappyCodec and other already verify correctness.  I have personally verified the performance gains using JMH benchmarks.


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/SnappyCompressor.java
```diff
@@ -32,6 +32,10 @@
  * entire input in setInput and compresses it as one compressed block.
  */
 public class SnappyCompressor implements Compressor {
+  // Double up to an 8 mb write buffer,  then switch to 1MB linear allocation
+  private static final int DOUBLING_ALLOC_THRESH =  8 << 20;
```
shangxinli, line 36: use 1 << 23 is more meaningful

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/SnappyCompressor.java
```diff
@@ -96,21 +100,40 @@ public synchronized void setInput(byte[] buffer, int off, int len) {
...
-      tmp.put(inputBuffer);
-      ByteBuffer oldBuffer = inputBuffer;
-      inputBuffer = tmp;
-      CleanUtil.cleanDirectBuffer(oldBuffer);
-    } else {
-      inputBuffer.limit(inputBuffer.position() + len);
+      resizeInputBuffer(inputBuffer.position() + len);
     }
 
+    inputBuffer.limit(inputBuffer.position() + len);
```
shangxinli, line 106: The original code doesn't call limit if (inputBuffer.capacity() - inputBuffer.position() < len)  is true

shangxinli:
I wonder how much benefit get can gain of this fix?

shangxinli:
@abaranec can you resolve the conflict?

abaranec:
@shangxinli Sorry it took so long to do this.  I resolved the conflicts.  The changes all essentially moved into NonBlockedCompressor,  I also incorporated the two changes you suggested.

One other thing worth discussing,  For the first alloc, I'm just using the requested size as the initial.  It occurs to me that it might be better to use a little more memory to guarantee that we start with, and continue with an 8-byte aligned buffer size.   Maybe starting at 16 or 32 bytes.  What do you think?

PR #995, opened 2022-09-08
PARQUET-1711: support recursive proto schemas by limiting recursion depth

jinyius:
- This is an alternative approach to supporting recursion to apache#445 and apache#988.
- This approach could address the other recursion related issues (PARQUET-129, PARQUET-554).
- TODO: ReadSupport

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-1711

### Tests

  - `ProtoSchemaConverterTest#test*Recursion`
  - `ProtoWriteSupportTest#test*Recursion`

### Commits

### Documentation

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.proto;
 
+import com.google.common.collect.ImmutableSetMultimap;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor.JavaType;
@@ -34,6 +35,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
+import javax.annotation.Nullable;
 
 import static org.apache.parquet.schema.LogicalTypeAnnotation.enumType;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.listType;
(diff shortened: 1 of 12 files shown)


jinyius:
ping

jinyius:
fixed missing dep issue.  can someone approve the ci flow?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -124,35 +164,61 @@ private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addR
...
 
     return result.named("element").named("list");
   }
 
-  private <T> GroupBuilder<GroupBuilder<T>> addMessageField(FieldDescriptor descriptor, final GroupBuilder<T> builder) {
+  private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addMessageField(FieldDescriptor descriptor, final GroupBuilder<T> builder, ImmutableSetMultimap<String, Integer> seen, int depth) {
+    // Prevent recursion by terminating with optional proto bytes.
+    depth += 1;
+    String typeName = getInnerTypeName(descriptor);
+    LOG.trace("addMessageField: " + descriptor.getFullName() + " type: " + typeName + " depth: " + depth);
```
emkornfield, line 182: its been a while since I've done java but doesn't the string concatenation as a parameter, incur overhead even if the log level is is above trace (i.e. doesn't there need to be some string formatting that takes the string arguments as separate parameters?)

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -124,35 +164,61 @@ private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addR
...
+      }
+    }
+
     if (descriptor.isMapField() && parquetSpecsCompliant) {
       // the old schema style did not include the MAP wrapper around map groups
-      return addMapField(descriptor, builder);
+      return addMapField(descriptor, builder, seen, depth);
     }
+
+    seen = ImmutableSetMultimap.<String, Integer>builder().putAll(seen).put(typeName, depth).build();
```
emkornfield, line 194: if this gets modified every time through this method, is immutability useful?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -99,9 +139,9 @@ private Type.Repetition getRepetition(FieldDescriptor descriptor) {
     }
   }
 
-  private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addField(FieldDescriptor descriptor, final GroupBuilder<T> builder) {
+  private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addField(FieldDescriptor descriptor, final GroupBuilder<T> builder, ImmutableSetMultimap<String, Integer> seen, int depth) {
```
emkornfield, line 142: would it make sense to consolidate seen and depth into a single data-structure that can be passed through and abstract some of the direct access to the multimap?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
jinyius: you're correct.  however, proto schema conversion shouldn't happen repeatedly in the greater flow of a processing job (ideally, just once), so this overhead isn't too bad.

i'll move to the parameterized/formatted logging calls in the files i touch here.  i would suggest the rest of the codebase do the same to avoid this penalty as well, but it's beyond the scope of this pr.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
jinyius: it's actually not as costly as you think.  guava's immutable structures are written to simply remove method access what not needed, and takes tries its best to avoid memory reallocations when using copyOf or builder patterns [[1](https://github.com/google/guava/wiki/ImmutableCollectionsExplained)][[2](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableSetMultimap.java#L365)][[3](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableSetMultimap.java#L306)][[4](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableSetMultimap.java#L291)].... [generally](https://stackoverflow.com/questions/1284727/mutable-or-immutable-class).  it's pretty [efficient](https://github.com/DimitrisAndreou/memory-measurer/blob/master/ElementCostInDataStructures.txt).  because of depth first traversal, we do want to "go back" and let the previous state of counts start again as the basis for other branch traversals.  this is exactly the benefit as it helps in avoiding defensive copying of mutable data structures or clearing of fields trying to use a single instance when traversing and going back up the stack.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
jinyius: ?

the seen map does encode the depth as a single datastructure.  the depth being a separate arg is important b/c it's the current depth in the traversal, and is used to update the seen data structure.

jinyius:
thanks for the review.  updated to handle the logging perf concern as well as fixing the javadoc errors.

jinyius:
ping

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoSchemaConverterTest.java
```diff
@@ -82,264 +93,447 @@ public void testConvertAllDatatypes() throws Exception {
...
-        "    repeated group key_value {\n" +
-        "      required int64 key;\n" +
-        "      optional group value {\n" +
-        "        optional int32 someId = 3;\n" +
-        "      }\n" +
-        "    }\n" +
-        "  }\n" +
-        "}";
+  public void testProto3ConvertAllDatatypes() {
+    String expectedSchema = JOINER.join(
```
emkornfield, line 97: is it possible to separate this tpe of code style cleanup from functional changes?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
emkornfield: right, I was thinking of encapsulating this logic into its own class, so they can be recorded and updated together, to  1.  Reduce additional parameters that have to be passed through.
2.  Encapsulate the logic behind more mnemonic method names (e.g. AddRecursiveStep())

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -559,7 +564,14 @@ final void writeRawValue(Object value) {
...
     final void writeRawValue(Object value) {
-      ByteString byteString = (ByteString) value;
+      // Non-ByteString values can happen when recursions gets truncated.
+      ByteString byteString = value instanceof ByteString
+          ? (ByteString) value
+          // TODO: figure out a way to use MessageOrBuilder
+          : value instanceof Message
+          ? ((Message) value).toByteString()
+          // Worst-case, just dump as plain java string.
+          : ByteString.copyFromUtf8(value.toString());
```
emkornfield, line 574: is this actually an intended state?  If not it is probably better to raise an exception then writing data that could possibly be hard to recover.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
```diff
@@ -559,7 +564,14 @@ final void writeRawValue(Object value) {
   class BinaryWriter extends FieldWriter {
     @Override
     final void writeRawValue(Object value) {
-      ByteString byteString = (ByteString) value;
+      // Non-ByteString values can happen when recursions gets truncated.
+      ByteString byteString = value instanceof ByteString
+          ? (ByteString) value
+          // TODO: figure out a way to use MessageOrBuilder
+          : value instanceof Message
+          ? ((Message) value).toByteString()
```
emkornfield, line 572: does recordconsumer offer a stream API or something else to avoid the additional array/bytestring copies?

parquet-protobuf/src/test/resources/BinaryTree.par
```diff
@@ -0,0 +1,50 @@
+message Trees.BinaryTree {
+  optional group value = 1 {
```
emkornfield, line 2: Aren't groups [deprecated](https://developers.google.com/protocol-buffers/docs/proto#groups)?

emkornfield: or is par not proto?

parquet-protobuf/src/test/resources/Trees.proto
```diff
@@ -0,0 +1,37 @@
...
+syntax = "proto3";
+
+package Trees;
+
+import "google/protobuf/any.proto";
+
+option java_package = "org.apache.parquet.proto.test";
+
+message BinaryTree {
+    google.protobuf.Any value = 1;
```
emkornfield, line 29: it would be good to verify that something like:

message WrappedTree {
   google.protobuf.Any non_recursive = 1;
   BinaryTree tree = 2;
}

Also gives expected results (non_recursive doesn't accidentally trigger any of the recursio logic). 
}

emkornfield:
Mostly looks reasonable, I'm not too familiar with parquet-mr @shangxinli can you recommend someone who might be able to give a better review?

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
jinyius: i'm not sure encapsulation helps with readability or protection in this case.  they are really tracking different things, and should be understood by readers of the traversal code to know how each piece of state is used.

parquet-protobuf/src/test/java/org/apache/parquet/proto/ProtoSchemaConverterTest.java
jinyius: wdym by "tpe"?

if this isn't blocking, i'd rather avoid the busy-work to undo and redo in a different branch.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
jinyius: this is intended.  for a real-time, production pipeline i'm working on, losing data as it passes through or killing the job b/c of an uncaught exception is problematic as it could lead to data loss and down time.  this way, there's some way to know what the problematic data was and fix it properly asap.

parquet-protobuf/src/test/resources/BinaryTree.par
jinyius: this is parquet schema, not proto.  protos should/would have a .proto suffix.

parquet-protobuf/src/test/resources/Trees.proto
jinyius: i think the existing non-recursive proto tests exercise the existing and newly added (the skipping behavior) code paths.

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoWriteSupport.java
jinyius: no, afaict: https://www.javadoc.io/doc/org.apache.parquet/parquet-column/latest/org/apache/parquet/io/api/RecordConsumer.html

shangxinli:
@ggershinsky Can you have a look?

ggershinsky (approved):
I'm ok with the current state of the PR, and would like to thank its reviewers.

ggershinsky:
I would also like to recommend adding @matthieun as a co-author to this PR, per the discussion in the parallel PR.

jinyius:
> I would also like to recommend adding @matthieun as a co-author to this PR, per the discussion in the parallel PR.

how do you do this?

jinyius:
can someone retry the github actions?  there seemed to have been a transient issue that caused one of the test/build targets to fail.  i'd like to get this change in this week.

emkornfield:
@ggershinsky what is the process to merge this?  Does parquet-mr just use the github UI?

ggershinsky:
yep, just the squash/merge button.

jinyius:
>

jinyius:
@ggershinsky 

i'd love to just hit the button.  i don't see it.  the workflow for travis ci had a failure due to a transient connection issue, and so it wasn't giving me the option to merge.  the ui messaging also states that "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests."

ggershinsky:
@shangxinli are you ok with this PR in its current form?

jinyius:
yeah, i still don't see a button to merge.  it now shows everything approved, checks passed, and no conflicts.

i think a committer needs to merge.

ggershinsky (approved):
Thanks @jinyius

shangxinli:
LGTM

PR #998, opened 2022-09-23
PARQUET-2195: Add scan command to parquet-cli

wgtmac:
This PR enhances parquet-cli by adding a scan command to go through all records without printing them. This is useful when users need to verify if the parquet file is corrupted.

Added ScanCommandTest to make sure no error happens. Test it manually with local parquet files.

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
@@ -36,7 +36,7 @@ import java.util.List;
 public class ScanCommand extends BaseCommand {
 
   @Parameter(description = "<file>")
-  List<String> sourceFiles;
+  String sourceFile;
 
   @Parameter(
     names = {"-c", "--column", "--columns"},
@@ -50,17 +50,14 @@ public class ScanCommand extends BaseCommand {
   @Override
   public int run() throws IOException {
     Preconditions.checkArgument(
-      sourceFiles != null && !sourceFiles.isEmpty(),
+      sourceFile != null && !sourceFile.isEmpty(),
       "Missing file name");
-    Preconditions.checkArgument(sourceFiles.size() == 1,
-      "Only one file can be given");
 
-    final String source = sourceFiles.get(0);
-    Schema schema = getAvroSchema(source);
+    Schema schema = getAvroSchema(sourceFile);
     Schema projection = Expressions.filterSchema(schema, columns);
 
     long startTime = System.currentTimeMillis();
-    Iterable<Object> reader = openDataFile(source, projection);
+    Iterable<Object> reader = openDataFile(sourceFile, projection);
     boolean threw = true;
     long count = 0;
     try {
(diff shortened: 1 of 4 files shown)


wgtmac:
@shangxinli Please take a look when you have time. Thanks!

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
```diff
@@ -0,0 +1,94 @@
...
+    super(console);
+  }
+
+  @Override
+  public int run() throws IOException {
+    Preconditions.checkArgument(
+      sourceFiles != null && !sourceFiles.isEmpty(),
+      "Missing file name");
+    Preconditions.checkArgument(sourceFiles.size() == 1,
+      "Only one file can be given");
```
shangxinli, line 56: Why not define it as String instead of List<Striing>?

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
```diff
@@ -0,0 +1,94 @@
...
+
+    final String source = sourceFiles.get(0);
+    Schema schema = getAvroSchema(source);
+    Schema projection = Expressions.filterSchema(schema, columns);
```
shangxinli, line 60: What do we do if the file that doesn't have Avro Schema,

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
```diff
@@ -0,0 +1,94 @@
...
+
+    long startTime = System.currentTimeMillis();
+    Iterable<Object> reader = openDataFile(source, projection);
+    boolean threw = true;
+    long count = 0;
+    try {
+      for (Object record : reader) {
+        count += 1;
```
shangxinli, line 68: If your goal is only to get count, why not get it from the metadata? Iterate each record is an expensive operation.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
wgtmac: I believe its naming is a little bit confusing. It supports getting schema from parquet, avro and avsc files. 

Please check here for detail: https://github.com/apache/parquet-mr/blob/master/parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java#L397

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
wgtmac: It serves the same purpose as https://github.com/apache/arrow/blob/master/cpp/tools/parquet/parquet_scan.cc to validate the data integrity of a parquet file.

wgtmac: This is very handy when we want to quickly check if any file is corrupted in production.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ScanCommand.java
wgtmac: Fixed

shangxinli:
lgtm

shangxinli (approved):

PR #1001, opened 2022-09-28
PARQURT-2197: Document uniform encryption

ggershinsky:
Document the hadoop parameter for uniform encryption


ggershinsky:
This is a minor patch, documenting a hadoop parameter that already exists. But if there are any comments, let me know.

PR #1010, opened 2022-11-10
PARQUET-2213: add InputFile.newStream with a read range

sunchao:
This adds a new API, `InputFile.newStream`, that allow specifying a range for the returned input stream. This allows the implementations (especially for cloud storages like S3) to only download the data within the specified range, instead of the whole object.

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/io/InputFile.java
+++ b/parquet-common/src/main/java/org/apache/parquet/io/InputFile.java
@@ -41,4 +41,16 @@ public interface InputFile {
    */
   SeekableInputStream newStream() throws IOException;
 
+  /**
+   * Open a new {@link SeekableInputStream} for the underlying data file,
+   * in the range of '[offset, offset + length)'
+   *
+   * @param offset the offset in the file to read from
+   * @param length the total number of bytes to read
+   * @return a new {@link SeekableInputStream} to read the file
+   * @throws IOException if the stream cannot be opened
+   */
+  default SeekableInputStream newStream(long offset, long length) throws IOException {
+    return newStream();
+  }
 }


parquet-common/src/main/java/org/apache/parquet/io/InputFile.java
```diff
@@ -41,4 +41,16 @@ public interface InputFile {
...
+  /**
+   * Open a new {@link SeekableInputStream} for the underlying data file,
+   * in the range of '[offset, offset + length)'
+   *
+   * @param offset the offset in the file to read from
+   * @param length the total number of bytes to read
+   * @return a new {@link SeekableInputStream} to read the file
+   * @throws IOException if the stream cannot be opened
+   */
+  default SeekableInputStream newStream(long offset, long length) throws IOException {
```
wgtmac, line 53: If we need to read multiple part of a parquet file (e.g. different row groups, page index, footer, etc.), should we call it multiple times for each individual part?

steveloughran:
I would prefer if Parquet used the same opt(key, value) builder pattern that we use in the new hadoop FS api calls. https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/fsdatainputstreambuilder.html

This allows for future addition of new options. The reader could then take them and, where appropriate, map them to the hadoop openfile options org.apache.hadoop.fs.Options.OpenFileOptions#FS_OPTION_OPENFILE_STANDARD_OPTIONS which can then get picked up by the connector.

passing in split/start end and file length is good.
file length: used by s3a to skip the HEAD when opening; abfs and gcs could copy. abfs will take a FileStatus in the withFileStatus() parameter
split start: where to begin that read
split end: should be used by prefetchers to know where to stop prefetching

parquet should set the read policy itself, i'd go for "random, adaptive" as the ordered list, with "vectored" in front of that when vectored IO is to be used.

parquet-common/src/main/java/org/apache/parquet/io/InputFile.java
steveloughran: this is what the vectored IO API of hadoop 3.3.5, an API for which we intend to provide in a compatibility library for apps running against hadoop 3.2.x+. https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/site/markdown/filesystem/fsdatainputstream.md#default-void-readvectoredlist-extends-filerange-ranges-intfunctionbytebuffer-allocate

steveloughran: problem with length here is it *split length* or *file length*? as with splittable text formats different tasks may get their own split and are allowed to read past it. this is why the fs builder api has two different options for file len and split start/end and we can't use split end as the value for file length.

steveloughran:
added comments, especially on a builder api.

one option to be able to set would be whether the footer was required or not. as gcs/abfs will both fetch and cache that if they think it would be good; s3a can do the same. knowing if it was wanted (or ideally, how many MB of footer you'd like) would help shape this policy

parquet-common/src/main/java/org/apache/parquet/io/InputFile.java
sunchao: > If we need to read multiple part of a parquet file (e.g. different row groups, page index, footer, etc.), should we call it multiple times for each individual part?

Yea that is the use case we are looking at, ideally we can do parallel downloading for different `Chunk`s in `ParquetFileReader`. Although this PR doesn't cover that part of the changes yet. 

For S3 in particular, we can use this API to only download a range of the file (with `fs.s3a.experimental.input.fadvise` set to `random`), which could give better performance.

> problem with length here is it split length or file length? as with splittable text formats different tasks may get their own split and are allowed to read past it. this is why the fs builder api has two different options for file len and split start/end and we can't use split end as the value for file length.

This is more like "range length". The idea is to break up a row group into smaller ranges (think of column chunks, or consecutive pages in case of column index) which can be processed in parallel.

steveloughran: you should go with the hadoop https://issues.apache.org/jira/browse/HADOOP-16202 options; s3a fs now reads them and it lines up abfs/gcs for the same. you can declare split start/end as well as file length so that
* length => client can skip existence probes, they know the file limit
* spilt range: they know to not prefetch past the end of the split, if they prefetch
* read policy: standard set of policies and a parse policy of "csv list of policies -pick the first one you recognise". again, can be used by all the stores

sunchao:
Thanks @steveloughran . I like your idea of introducing a builder api, although we need to think how it can be used by different implementations of `InputFile`, not only for Hadoop.

For instance Iceberg has this [`S3InputFile`](https://github.com/apache/iceberg/blob/master/aws/src/main/java/org/apache/iceberg/aws/s3/S3InputFile.java) which knows nothing about file status or read policy.

steveloughran:
> For instance Iceberg has this [S3InputFile](https://github.com/apache/iceberg/blob/master/aws/src/main/java/org/apache/iceberg/aws/s3/S3InputFile.java) which knows nothing about file status or read policy.

the builder api we use has opt(key, val) being something implementations can ignore, must(key, val) says "you must recognise the key and 'val' must be something you understand".

so opening a file with a length and read policy would be ignored by iceberg automatically if in opt()
```
fc.openFile(p)
        .opt("fs.option.openfile.read.policy",
            "random")
        .opt("fs.option.openfile.length",
            Long.toString(status.getLen()))
```
put them in must() and the implementation must recognise the keys, even if it *chooses* to not actually do anything with the length/policy.

PR #1011, opened 2022-11-15
PARQUET-2159: Vectorized BytePacker decoder using Java VectorAPI

jiangjiguang:
The PR includes 3 aspects：
1. Use java17 vector api to decode bit-packing ,  the performance gain is 4x ~ 8x according to the microbenchmark 
2. Upgrade the project to java17 to support java vector api
3. Add ParquetReadRouter to compatible with different platform when computing engines(such as spark) read parquet.

### Jira
- https://issues.apache.org/jira/projects/PARQUET/issues/PARQUET-2159?filter=allopenissues

### Tests

- Add unit tests org.apache.parquet.column.values.bitpacking.TestByteBitPackingVectorLE

### Commits

### Documentation
- The PR adds maven profile vector to enable generate java17 vector bit-packing decode opt. code, and run junit tests: mvn clean install -P vector
- The PR needs Intel Ice Lake CPU to run junit tests
- The contributors are jiangjiguang  jiyu1021 guangzegu  Fang-Xie  , and co-contributor is  jatin-bhateja


jiangjiguang:
@wangyum @gszadovszky This PR always fails to build, I do not know why.   Is the reason of failure "1 workflow awaiting approval" ? please help me, this is my first PR to parquet-mr community, thanks!

jiangjiguang:
@wangyum Could you review the PR ? thanks

gszadovszky:
@jiangjiguang, Apache introduced a process to require approvals for test executions to keep resources under control. I've approved it so the tests are running. Not sure though when I will have time to properly review it.

jiangjiguang:
@gszadovszky I resubmitted the code， can you approve the workflow again ?

jiangjiguang:
@gszadovszky Could you review the PR ? thanks

gszadovszky:
@jiangjiguang, however I agree on upgrading to newer java versions (and to have performance benefits like this one) it is not always an easy thing to do. For example our main historical user Hadoop is still on java 1.8 which means they would not be able to upgrade parquet after releasing this one. I think it worth a discussion on the dev list and maybe even a formal vote on it.

jiangjiguang:
@gszadovszky I agree on it. As far as I know some products(such as hadoop presto flink ) are working towards upgrading to java17.  Spark already supports java17, trino requires a minimum java version is java17.0.3,  So I think parquet  should support java17 as soon as possible and be compatible with java8, because of java17 vector can bring 4x ~ 8x performance gain for parquet encode/decode. The PR uses "maven profile -P vector"  and  "code gen" to support java17 and is compatible with java8

gszadovszky:
@jiangjiguang, 

What do you mean by this change is compatible with java8? The vectorized encoding/decoding requires java17 runtime, right? If yes, we will need to have a way release artifacts for java17 and keep java8 compatibility with others. We need to have an agreement (at community level) how we want to achieve this.

By running through the change related to the compile it seems that if you have a jdk17 in the compile environment it will be compiled for a java17 runtime automatically. Since we do not have a specific environment for making a release (shame on us) it highly depends on the environment of the dev how creates the release. So, we shall either add a `release` configuration for `8` to the `jdk7` profile (if it makes sense) or do not have the `jdk17` profile activated automatically. Otherwise we might release artifacts that are not compatible < java17.

jiangjiguang:
> @gszadovszky That what I said supports java17 and is compatible with java8 depends on compile environment. Can you start a discussion on how to support java17 ?

gszadovszky:
@jiangjiguang,

With you current implementation if parquet-mr is compiled with jdk17 than it will not be compatible with java8 while any jdk under 17 will generate a code compatible with java8 since `release` is set to 8. My problem is this exact behaviour. I would suggest keep generating java8 compatible artifacts with using any versions of jdks (including 17) by default and let it compile for java17 (and the vectorized stuff) when specific profile is activated manually.

About starting the discussion about java17 compatibility. I am currently not really active in the community and not even following the dev list. Maybe, @shangxinli or @ggershinsky would like to drive this topic?

shangxinli:
Thank @gszadovszky a lot for helping with this PR! 

+1 for what @gszadovszky said. The mainstream runtime JDK is still 1.8. Parquet is one of the underlying building blocks for many big data applications.  The bare minimum, for now, is to keep java8 compatible. Otherwise forcing applications to upgrade to jdk17 because of Parquet is disruptive and impacts adoptions. 

@jiangjiguang, I am very happy to see you have this PR to help the Parquet community.  Would you mind starting an email discussion to dev@parquet.apache.org for this topic?

parquet-generator/src/main/resources/ByteBitPackingVectorLE
```diff
@@ -0,0 +1,3215 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.column.values.bitpacking;
```
wgtmac, line 20: Should we add a comment here to let reader know this is an auto-generated source file and should not edit it directly? Just in case.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.apache.parquet.column.values.bitpacking;
+
+import org.apache.parquet.bytes.ByteBufferInputStream;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+public class ParquetReadRouter {
```
wgtmac, line 27: It seems that this class is not used anywhere.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
```diff
@@ -105,4 +114,15 @@ public void unpack8Values(final byte[] input, final int inPos, final int[] outpu
   public void unpack32Values(byte[] input, int inPos, int[] output, int outPos) {
     unpack32Values(ByteBuffer.wrap(input), inPos, output, outPos);
   }
+
+  public void unpackValuesVector(final byte[] input, final int inPos, final int[] output, final int outPos) {
+
+  }
```
wgtmac, line 120: Please add an empty line between method bodies.

parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingVectorGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.encoding.bitpacking;
+
+import java.io.*;
+
+/**
+ * This class generates vector bit packers that pack the most significant bit first.
+ * The result of the generation is checked in. To regenerate the code run this class and check in the result.
```
wgtmac, line 25: Why is this intentional?

parquet-generator/src/main/java/org/apache/parquet/encoding/GeneratorVector.java
```diff
@@ -0,0 +1,31 @@
...
+ * under the License.
+ */
+package org.apache.parquet.encoding;
+
+import org.apache.parquet.encoding.bitpacking.ByteBasedBitPackingVectorGenerator;
+
+/**
+ * main class for code generation hook in build for encodings vector generation
+ */
+public class GeneratorVector {
```
wgtmac, line 26: Should we add a separate package like `vectorbitpacking` and rename this class to `Generator` so that it is consistent with other Generators in the parquet-generator?

parquet-generator/src/main/java/org/apache/parquet/encoding/GeneratorVector.java
```diff
@@ -0,0 +1,31 @@
...
+  public static void main(String[] args) throws Exception {
+    Generator.main(args);
```
wgtmac, line 28: Why mixing these two generators together? IMHO, making generators to do their jobs separately is less error-prone to users.

wgtmac (commented):
This work looks promising! It would be great if you can add some micro-benchmark to parquet-benchmarks.

parquet-generator/src/main/resources/ByteBitPackingVectorLE
jiangjiguang: Good idea!

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
jiangjiguang: Yes, this class will be used for big data applications (spark flink ...) as they use now,  I will add some unit test.

parquet-generator/src/main/java/org/apache/parquet/encoding/GeneratorVector.java
jiangjiguang: because bitpacking java17 vector decode depends on the current decode way, so I think it have to run Generator before GeneratorVector

parquet-generator/src/main/java/org/apache/parquet/encoding/bitpacking/ByteBasedBitPackingVectorGenerator.java
jiangjiguang: This class is used to generate ByteBitPackingVectorLE class, which unpackValues/packValues with java17 vector api.
If use  java17 vector api directly, errors will occur when compiling or developing with java8(<java17).

jiangjiguang:
> This work looks promising! It would be great if you can add some micro-benchmark to parquet-benchmarks.

@wgtmac I have add the micro-benchmark to parquet-benchmarks, this is the result:

image

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
```diff
@@ -105,4 +116,16 @@ public void unpack8Values(final byte[] input, final int inPos, final int[] outpu
   public void unpack32Values(byte[] input, int inPos, int[] output, int outPos) {
     unpack32Values(ByteBuffer.wrap(input), inPos, output, outPos);
   }
+
+  public void unpackValuesVector(final byte[] input, final int inPos, final int[] output, final int outPos) {
```
jatin-bhateja, line 120: Please add appropriate comments over newly added definitions, like the ones over top over scalar routines.

parquet-generator/src/main/resources/ByteBitPackingVectorLE
```diff
@@ -0,0 +1,3218 @@
...
+    public final void unpack32Values(final byte[] in, final int inPos, final int[] out, final int outPos) {
+    }
+
+    public final void unpack32Values(final ByteBuffer in, final int inPos, final int[] out, final int outPos) {
+    }
+
+    public final void unpackValuesVector(final byte[] input, final int inPos, final int[] output, final int outPos) {
+    }
+
+    public final void unpackValuesVector(final ByteBuffer input, final int inPos, final int[] output, final int outPos) {
```
jatin-bhateja, line 106: All these empty definitions can be removed if we introduce a new class ByteVectorPacker which inherit from existing BytePacker.

jatin-bhateja: We also need to ensure that we invoke both scalar generators and vector generators for jdk17 profile.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ByteBitPackingVectorBenchmarks.java
```diff
@@ -0,0 +1,83 @@
...
+    BytePacker bytePacker = Packer.LITTLE_ENDIAN.newBytePacker(bitWidth);
+    for (int i = 0, j = 0; i < input.length; i += bitWidth, j += 8) {
+      bytePacker.unpack8Values(input, i, output, j);
+    }
+  }
+
+  @Benchmark
+  public void testUnpackVector() {
+    BytePacker bytePacker = Packer.LITTLE_ENDIAN.newBytePacker(bitWidth);
+    BytePacker bytePackerVector = Packer.LITTLE_ENDIAN.newBytePackerVector(bitWidth);
```
jatin-bhateja, line 65: As suggested in my following comments, it will be useful if we are able to access all the routines through one packer instance.

jiangjiguang:
@wgtmac  I have resubmitted the PR, can you review it again ?

jiangjiguang:
> Thank @gszadovszky a lot for helping with this PR!
> 
> +1 for what @gszadovszky said. The mainstream runtime JDK is still 1.8. Parquet is one of the underlying building blocks for many big data applications. The bare minimum, for now, is to keep java8 compatible. Otherwise forcing applications to upgrade to jdk17 because of Parquet is disruptive and impacts adoptions.
> 
> @jiangjiguang, I am very happy to see you have this PR to help the Parquet community. Would you mind starting an email discussion to [dev@parquet.apache.org](mailto:dev@parquet.apache.org) for this topic?
> 
> cc @ggershinsky @wgtmac

@shangxinli 

> 

@gszadovszky @shangxinli @wgtmac I have started the discussion about how to upgrade java17 over a month, but nobody involved!  So I have updated the PR, it does not  involve how to upgrade java17.
The default compilation is java8
Just add maven build parameters  -P java17-target -P vector and get the expected jars  when people want to use java17 vector to speed up parquet decode

gszadovszky:
Sorry for the bad experience, @jiangjiguang. Unfortunately, I don't have too much time for the parquet community.
@wgtmac, @jatin-bhateja, since you were the ones who actively reviewed this PR I would like to ask for your approvals. I won't have time to properly review this but I am happy to push it if you agree.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public static Boolean supportVector() {
+    if (vector != null) {
+      return vector;
+    }
+    synchronized (ParquetReadRouter.class) {
+      if (vector == null) {
+        synchronized (ParquetReadRouter.class) {
+          vector = avx512Flag();
```
wgtmac, line 100: I assume the java vector api is agnostic to specific instruction set. So this function name is a little bit misleading to me since it binds the vector to AVX-512. Better to organize the code to be more extensible for other instruction sets.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+      }
+    }
+    return vector;
+  }
+
+  private static boolean avx512Flag() {
```
wgtmac, line 107: Looks like these routines will be executed every time once `supportVector()` is called.

jiangjiguang:
> @jiangjiguang, please check test failures.

@gszadovszky 

> @jiangjiguang, please check test failures.

@gszadovszky Thank you for your attention about the PR, I have fixed it and the test successes.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,132 @@
...
+          }
+          line = line.substring(index + 1);
+          Set<String> flagsSet = Arrays.stream(line.split(" ")).collect(Collectors.toSet());
+          if (flagsSet.contains("avx512vbmi") && flagsSet.contains("avx512_vbmi2")) {
+            return true;
+          }
+        }
+      }
+    } catch (Exception ex) {
+      LOG.warn("Not getting CPU info error");
```
wgtmac, line 128:
```suggestion
      LOG.warn("Failed to gett CPU info");
```

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,132 @@
...
+  public static void readBatchVector(int bitWidth, ByteBufferInputStream in, int currentCount, int[] currentBuffer) throws IOException {
+    BytePacker packer = Packer.LITTLE_ENDIAN.newBytePacker(bitWidth);
+    BytePacker packerVector = Packer.LITTLE_ENDIAN.newBytePackerVector(bitWidth);
+    int valueIndex = 0;
+    int byteIndex = 0;
+    int unpackCount = packerVector.getUnpackCount();
+    int inputByteCountPerVector = packerVector.getUnpackCount() / 8 * bitWidth;
+    int totalByteCount = currentCount * bitWidth / 8;
+
+    // register of avx512 are 512 bits, and can load up to 64 bytes
```
wgtmac, line 62: Can we make 64 bytes to be an input parameter? My intention is not to fix it so we can support other instruction sets as well. In my experience, sometimes AVX2 outperforms AVX512 in certain workloads, especially for bitpacking.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
```diff
@@ -31,6 +31,13 @@ public abstract class BytePacker {
 
   private final int bitWidth;
 
+  /**
+   * unpack the output int values at a time.
```
wgtmac, line 35:
```suggestion
   * Number of integer values to be unpacked at a time.
```

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
```diff
@@ -105,4 +116,26 @@ public void unpack8Values(final byte[] input, final int inPos, final int[] outpu
...
   }
+
+  /**
+   * unpack bitWidth bytes from input at inPos into {unpackCount} values in output at outPos using Java Vector API.
+   * @param input the input bytes
+   * @param inPos where to read from in input
+   * @param output the output values
+   * @param outPos where to write to in output
+   */
+  public void unpackValuesVector(final byte[] input, final int inPos, final int[] output, final int outPos) {
```
wgtmac, line 127: Do we have a standard norm to define a function when it leverages the Vector API? Does it sound good to rename it to `unpackValuesUsingVector` or `vectorUnpackValues`? The current name is misleading.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
```diff
@@ -105,4 +116,26 @@ public void unpack8Values(final byte[] input, final int inPos, final int[] outpu
...
+
+  /**
+   * unpack bitWidth bytes from input at inPos into {unpackCount} values in output at outPos using Java Vector API.
+   * @param input the input bytes
+   * @param inPos where to read from in input
+   * @param output the output values
+   * @param outPos where to write to in output
+   */
+  public void unpackValuesVector(final ByteBuffer input, final int inPos, final int[] output, final int outPos) {
+
```
wgtmac, line 139: It is better to throw here and above?

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/Packer.java
```diff
@@ -101,6 +121,10 @@ private static Object getStaticField(String className, String fieldName) {
    */
   public abstract BytePacker newBytePacker(int width);
 
+  public BytePacker newBytePackerVector(int width) {
+    throw new RuntimeException("This function must be implemented by subclasses!");
```
wgtmac, line 125:
```suggestion
    throw new RuntimeException("newBytePackerVector must be implemented by subclasses!");
```

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGeneratorVector.java
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.encoding.vectorbitpacking;
+
+import java.io.*;
```
wgtmac, line 21: Please do not use `*` in the import.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGeneratorVector.java
```diff
@@ -0,0 +1,63 @@
...
+ * This class generates vector bit packers that pack the most significant bit first.
+ * The result of the generation is checked in. To regenerate the code run this class and check in the result.
+ */
+public class BitPackingGeneratorVector {
+  private static final String CLASS_NAME_PREFIX_FOR_INT = "ByteBitPackingVector";
+  private static final String CLASS_NAME_PREFIX_FOR_LONG = "ByteBitPackingVectorForLong";
+
+  public static void main(String[] args) throws Exception {
+    String basePath = args[0];
+    //TODO: Int for Big Endian
```
wgtmac, line 33: Would you mind opening the JIRAs to break down the work for better tracking?

pom.xml
```diff
@@ -659,5 +666,74 @@
...
+              </compilerArgs>
+            </configuration>
+          </plugin>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <argLine>${surefire.argLine} ${extraJavaTestArgs} ${extraJavaVectorTestArgs}</argLine>
+              <systemPropertyVariables>
```
wgtmac, line 690: Why are these needed?

wgtmac (commented):
Sorry for the delay. I have left some comments and the implementation is overall looking good. Thanks @jiangjiguang for your effort!

My main concern is the extensibility to support other instruction sets. In addition, it seems to me that the java vector api is still incubating. As I am not a java expert, do we have the risk of unstable API?

jiangjiguang:
> Sorry for the delay. I have left some comments and the implementation is overall looking good. Thanks @jiangjiguang for your effort!
> 
> My main concern is the extensibility to support other instruction sets. In addition, it seems to me that the java vector api is still incubating. As I am not a java expert, do we have the risk of unstable API?

@wgtmac Jatin is a java expert， @jatin-bhateja Can you help give an answer? thanks.

pom.xml
jiangjiguang: extraJavaVectorTestArgs is only for Java17 Vector module, I have rename it to extraJavaVectorArgs, and move it to profile vector

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGeneratorVector.java
jiangjiguang: OK, I will create JIRAs to  break down the work.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGeneratorVector.java
jiangjiguang: OK, I will remote it next update

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGeneratorVector.java
jiangjiguang: I have created 3 sub-tasks(including Int for Big Endian、Long for Big Endian、Long for Little Endian) with https://issues.apache.org/jira/browse/PARQUET-2159

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
jiangjiguang: About "the java vector api is agnostic to specific instruction set." @jatin-bhateja Can you help with details ? thanks.
@wgtmac The PR using only avx512 vector to optimize parquet decode. so I have renamed related classes and methods to to be more extensible for other instruction sets(such as AVX2)

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
jiangjiguang: @wgtmac yes, I have added some comments to guide users how to use this class.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
jiangjiguang: @wgtmac I have reviewed the Java Vector Module name norm, and renamed related classed and methods, unpackValuesUsingVector looks good, so I used it.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
jiangjiguang: @wgtmac It is very difficult to make 64 bytes to be an input parameter.  The PR using only AVX512 Vector to optimize parquet decode. As you said, sometimes AVX2 outperforms better, but we should know the specific workloads for AVX2 so that we can make a choice to use AVX2 or AVX512.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
jiangjiguang: @wgtmac I don't understand it, not throw exception here. Can you you describe in detail ？ thanks.

jiangjiguang:
@wgtmac PTAK again  thanks.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,141 @@
...
+public class ParquetReadRouter {
+  private static final Logger LOG = LoggerFactory.getLogger(ParquetReadRouter.class);
+
+  private static final int BITS_PER_BYTE = 8;
+
+  // register of avx512 are 512 bits, and can load up to 64 bytes
+  private static final int BYTES_PER_512VECTOR = 64;
+
+  // values are bit packed 8 at a time, so reading bitWidth will always work
+  private static final int VALUES_PER_PACKED = 8;
```
wgtmac, line 49:
```suggestion
  private static final int NUM_VALUES_TO_PACK = 8;
```

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,141 @@
...
+
+  private static volatile Boolean vector;
```
wgtmac, line 51: Does it make sense to use an enum instead of boolean to represent longest available vector support? Something like this:
```
enum VectorSupport {
  NONE,
  VECTOR_128,
  VECTOR_256,
  VECTOR_512
}
```
This makes it more extensible.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,141 @@
...
+ * Utility class for big data applications (such as Apache Spark and Apache Flink).
+ * For Intel CPU, Flags containing avx512vbmi and avx512_vbmi2 can have better performance gains.
+ */
+public class ParquetReadRouter {
+  private static final Logger LOG = LoggerFactory.getLogger(ParquetReadRouter.class);
+
+  private static final int BITS_PER_BYTE = 8;
+
+  // register of avx512 are 512 bits, and can load up to 64 bytes
+  private static final int BYTES_PER_512VECTOR = 64;
```
wgtmac, line 46:
```suggestion
  private static final int BYTES_PER_VECTOR_512 = 64;
```

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,141 @@
...
+
+  // values are bit packed 8 at a time, so reading bitWidth will always work
+  private static final int VALUES_PER_PACKED = 8;
+
+  private static volatile Boolean vector;
+
+  // The is just a logical method, most call readBatchUsing512Vector or readBatch directly given that you should know your computer system
```
wgtmac, line 53:
```suggestion
  // Dispatches to use vector when available. Directly call readBatchUsing512Vector() if you are sure about it.
```

parquet-column/src/test/java/org/apache/parquet/column/values/bitpacking/TestParquetReadRouter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.ByteBuffer;
+
+import static org.junit.Assert.assertArrayEquals;
+
+public class TestParquetReadRouter {
+
+  /**
+   * The range of bitWidth is 1 ~ 32, change it directly if test other bitWidth.
+   */
+  private static final int bitWidth = 7;
```
wgtmac, line 34: It would be good to add cases of different bit widths.

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/BytePacker.java
wgtmac: I mean why is it an empty function? Shouldn't it throw an unimplemented exception?

parquet-generator/src/main/resources/ByteBitPacking512VectorLE
```diff
@@ -0,0 +1,3095 @@
...
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorShuffle;
+import jdk.incubator.vector.VectorSpecies;
+
+import java.nio.ByteBuffer;
+
+/**
+ * This is an auto-generated source file and should not edit it directly.
+ */
+public abstract class ByteBitPacking512VectorLE {
```
wgtmac, line 37: Should we remove this generated file from git? I assume it will be automatically created at maven build time, like other generators do.

wgtmac (commented):
Thanks for addressing the feedback. Left some minor comments.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
jiangjiguang: Good idea!

parquet-column/src/test/java/org/apache/parquet/column/values/bitpacking/TestParquetReadRouter.java
jiangjiguang: Good idea!

parquet-generator/src/main/resources/ByteBitPacking512VectorLE
jiangjiguang: This is a good question! 
    Why I put the generated file in the resources directory? Due to use Java Vector, it needs rearrange/shuffle/lanewise operations, so it is difficult to create automatically ByteBitPacking512VectorLE like other generators do. On the contrary,  It is relatively simple to code directly instead of generating.
    I think more important is the finished code than how that code was generated. 
    In fact, I've done part of the work to generate ByteBitPacking512VectorLE like other generators do. But is the generating valuable if it if it increases the workload ?
@wgtmac Do you think it is necessary to generate ByteBitPacking512VectorLE like other generators do ?  thanks.

jiangjiguang:
@wgtmac PTAK again

parquet-generator/src/main/resources/ByteBitPacking512VectorLE
wgtmac: OK, I got your point now. Thanks!

wgtmac:
> @wgtmac PTAK again

Generally this patch looks good to me now. Thanks @jiangjiguang for working on it!

Could you approve the workflow and take another pass? @gszadovszky @shangxinli @ggershinsky

jatin-bhateja:
> > Sorry for the delay. I have left some comments and the implementation is overall looking good. Thanks @jiangjiguang for your effort!
> > My main concern is the extensibility to support other instruction sets. In addition, it seems to me that the java vector api is still incubating. As I am not a java expert, do we have the risk of unstable API?
> 
> @wgtmac Jatin is a java expert， @jatin-bhateja Can you help give an answer? thanks.

Hi @wgtmac , our patch vectorizes unpacking algorithm for various decode bit sizes, entire new functionality is exposed through a plugin interface **ParquetReadRouter**, in order to prevent any performance regressions over other targets we have enabled the new functionality only for X86 targets with valid features, this limitation can be removed over time.

VectorAPI made its appearance in JDK16 and has been maturing since then with each successive release. I do not have a firm timeline for you at this point on its incubation exit and being exposed as a preview feature.  Intent here is to enable parquet-mr community developers to make use of the plugin in parquet reader and provide us with early feedback, we are also in process of vectorizing packer algorithm.

Being a large project we plan to do this incrementally, we seek your guidance here in pushing this patch through.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ByteBitPackingVectorBenchmarks.java
wgtmac: Could you elaborate more? @jatin-bhateja

parquet-generator/src/main/resources/ByteBitPacking512VectorLE
wgtmac: Do you have any script to generate the code here? If true, it would be great to commit it as well.

wgtmac:
> > > Sorry for the delay. I have left some comments and the implementation is overall looking good. Thanks @jiangjiguang for your effort!
> > > My main concern is the extensibility to support other instruction sets. In addition, it seems to me that the java vector api is still incubating. As I am not a java expert, do we have the risk of unstable API?
> > 
> > 
> > @wgtmac Jatin is a java expert， @jatin-bhateja Can you help give an answer? thanks.
> 
> Hi @wgtmac , our patch vectorizes unpacking algorithm for various decode bit sizes, entire new functionality is exposed through a plugin interface **ParquetReadRouter**, in order to prevent any performance regressions over other targets we have enabled the new functionality only for X86 targets with valid features, this limitation can be removed over time.
> 
> VectorAPI made its appearance in JDK16 and has been maturing since then with each successive release. I do not have a firm timeline for you at this point on its incubation exit and being exposed as a preview feature. Intent here is to enable parquet-mr community developers to make use of the plugin in parquet reader and provide us with early feedback, we are also in process of vectorizing packer algorithm.
> 
> Being a large project we plan to do this incrementally, we seek your guidance in pushing this patch through either on master or a separate development branch.

Thanks for your explanation @jatin-bhateja! 

So when vector API is finalized in the future java release, we may need to change the VM options to enable it accordingly.

BTW, I may not be able to verify the generated code line by line. Please advice the best practice to test them according to your experience. Thanks @jatin-bhateja

jiangjiguang:
> 

@wgtmac I know your concern: 
 1. I will keep the content of the PR updated if needed when java changed.
 2. I have coded a test to verify generated code, org.apache.parquet.column.values.bitpacking.TestByteBitPacking512VectorLE
 3. I have finished the TPC-H integrated Testing with spark, maybe I can write a document to give best practice to test them

parquet-generator/src/main/resources/ByteBitPacking512VectorLE
jiangjiguang: @wgtmac I have the script, but it generate only the code partly. 
It needs hard work and lots of time to complete the script(I don't think it is necessary). 
In fact, the code is completed mostly by manually.
Should I commit the script which is partly completed ?

wgtmac: In this case, the script is not necessary. Manual bit-unpacking code is error-prone, we really rely on the quality and coverage of test cases.

jiangjiguang: > In this case, the script is not necessary. Manual bit-unpacking code is error-prone, we really rely on the quality and coverage of test cases.

@wgtmac  I strongly agree with you, so I try my best to cover all aspects(different aspects from bitWidth 1 to 32) in class TestByteBitPacking512VectorLE. Besides, I have done the TPC-H testing, and compared the query result with before optimization. In short, I have done more work to ensure the code quality.

jiangjiguang:
@wgtmac I added doc about how big data applications use Java Vector API

README.md
```diff
@@ -83,6 +83,16 @@ Parquet is a very active project, and new features are being added quickly. Here
 * Column stats
 * Delta encoding
 * Index pages
+* Java Vector API support
+
+## Java Vector API support
```
wgtmac, line 88: @dongjoon-hyun @sunchao Could you please take a look and advise if it is possible to integrate it with Apache Spark? Thanks!

sunchao: It should be possible but we need to make sure it is backward compatible since Spark also compiles with older JDK versions like JDK8.

You can take a look at https://github.com/apache/spark/pull/30810 which does something similar.

jiangjiguang: @wgtmac @sunchao @dongjoon-hyun   Yes, I have integrated it with Apache Spark, and finished TPC-H testing. Because of Parquet ByteBitPacking512VectorLE class only will be loaded when used, we can add configuration such as "spark.sql.parquet.vector.enable", if false, any java vector api will not load, so it is backward compatible with older JDK versions like JDK8.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,133 @@
...
+public class ParquetReadRouter {
+  private static final Logger LOG = LoggerFactory.getLogger(ParquetReadRouter.class);
+
+  private static volatile Boolean vector;
+
+  public static void read(int bitWidth, ByteBufferInputStream in, int currentCount, int[] currentBuffer) throws IOException {
+    if (supportVector()) {
+      readBatchVector(bitWidth, in, currentCount, currentBuffer);
+    } else {
+      readBatchVector(bitWidth, in, currentCount, currentBuffer);
```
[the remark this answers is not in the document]
jiangjiguang: @jatin-bhateja please task the updated code

parquet-generator/src/main/resources/ByteBitPackingVectorLE
jiangjiguang: Yes, I have verified it

README.md
sunchao: @jiangjiguang sounds good, could you share the TPC-H benchmark results too?

WangYuxing0924: @sunchao , [here](https://issues.apache.org/jira/projects/PARQUET/issues/PARQUET-2159) shows the micro-benchmark of bitpack function and test report from Spark vectorizedparquetRecordReader (scan operatories). Most tpch queries are join-related operatories so the hotspot lies in the join/shuffle stage. bitpack optimization would be beneficial for SQL filter query

Fang-Xie: @sunchao , [here](https://issues.apache.org/jira/projects/PARQUET/issues/PARQUET-2159) shows the micro-benchmark of bitpack function and test report from Spark vectorizedparquetRecordReader (scan operatories). Most tpch queries are join-related operatories so the hotspot lies in the join/shuffle stage. bitpack optimization would be beneficial for SQL filter query

sunchao: Ah thanks! this looks promising and looking forward to the Spark PR!

jiangjiguang:
> > @wgtmac PTAK again
> 
> Generally this patch looks good to me now. Thanks @jiangjiguang for working on it!
> 
> Could you approve the workflow and take another pass? @gszadovszky @shangxinli @ggershinsky

@gszadovszky @shangxinli @ggershinsky Can you take a look the pr ? thanks

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * This class generates vector bit packers that pack the most significant bit first.
+ * The result of the generation is checked in. To regenerate the code run this class and check in the result.
+ */
```
gszadovszky, line 30: Is this comment still stands? I cannot find the generated file checked in. BTW, I don't think it is a good practice to check in generated files. It is better to automatically generate them in `generate-sources` phase and then compile them with the other source files.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
```diff
@@ -0,0 +1,67 @@
...
+    generateScheme(false, false, basePath);
+
+    //TODO: Long for Big Endian
+    //generateScheme(true, true, basePath);
+
+    //TODO: Long for Little Endian
+    //generateScheme(true, false, basePath);
+  }
+
+  private static void generateScheme(boolean isLong, boolean msbFirst,
```
gszadovszky, line 50: I do not really get how this generation works. Some description would be great.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
```diff
@@ -0,0 +1,67 @@
...
+    //generateScheme(false, true, basePath);
+
+    // Int for Little Endian
+    generateScheme(false, false, basePath);
+
+    //TODO: Long for Big Endian
+    //generateScheme(true, true, basePath);
+
+    //TODO: Long for Little Endian
+    //generateScheme(true, false, basePath);
```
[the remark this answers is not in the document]
jiangjiguang: @gszadovszky As far as I know, most JVM depends on Little Endian, I haven't see that JVM depends on Big Endian, So I only finished the code of Little Endian. "Long for Little Endian" affects on parquet page v2 version, I have finished the code, I will submit it to parquet-mr community after this PR.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
```diff
@@ -0,0 +1,133 @@
...
+          if (index < 0) {
+            continue;
+          }
+          line = line.substring(index + 1);
+          Set<String> flagsSet = Arrays.stream(line.split(" ")).collect(Collectors.toSet());
+          if (flagsSet.contains("avx512vbmi") && flagsSet.contains("avx512_vbmi2")) {
+            return VectorSupport.VECTOR_512;
+          }
+        }
+      }
```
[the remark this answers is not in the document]
jiangjiguang: @gszadovszky avx512vbmi and avx512_vbmi2 are CPU Flags which contains some instruction set. The code of this PR  has used the instruction set. Without them, we cannot get benefit from Java Vector API.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
jiangjiguang: @gszadovszky  I have changed the comments.
Due to use Java Vector API, it needs rearrange/shuffle/lanewise operations, so it is very difficult to create automatically ByteBitPacking512VectorLE like other generators do. On the contrary,  It is relatively simple to code directly instead of generating.
I think more important is the finished code than how that code was generated. 
Besides, the class is a plugin based on java17 to make it compatible with java8

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
jiangjiguang: @gszadovszky I have added the maven profile named vector, when execute mvn clean package -P vector, the class will be run and reads ByteBitPacking512VectorLE file to generate ByteBitPacking512VectorLE.java

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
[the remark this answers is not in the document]
gszadovszky: @jiangjiguang, I understand how it works. What I would have expected from the Vector API is to implement some help to check on the CPU flags.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
gszadovszky: So by generating this class simply copies the file form `resources` to a source dir under the name of `.java`. Why do we need code generation then?

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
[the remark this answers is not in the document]
gszadovszky: Let's clean up the TODOs in the next PR then. TODOs in the code in their own mean nothing. Nobody will pick them up to fix them. We either remove the TODOs or at least create a jira for them.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
gszadovszky: But if this code is written by hand (which is completely find) why don't we check it in instead of having the generator thing?

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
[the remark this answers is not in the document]
wgtmac: +1 for the request for official support to reduce maintenance burden.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
gszadovszky: I would suggest to have a separate source directory for java17 that gets "activated" only in case of the related profile is activated. This generator solution is misleading to me.

To keep this part of code clean I would also suggest to include java17 and the related class compile/unit test execution in the github actions so they would be executed on the PRs. (The only difference remains is we won't ship these in our releases.)

What do you think, @wgtmac, @jiangjiguang?

wgtmac: +1 for @gszadovszky 

The core idea is to isolate the new feature as an individual unit and minimize the burden of future maintenance.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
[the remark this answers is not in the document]
jiangjiguang: @jatin-bhateja Can you help give an answer ? thanks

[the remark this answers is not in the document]
jatin-bhateja: > @jiangjiguang, I understand how it works. What I would have expected from the Vector API is to implement some help to check on the CPU flags.

Hi @gszadovszky , vector APIs are just in time compiled into efficient instruction sequence, this instruction sequence can vary based on CPU features set of the target platform,  by introducing additional feature checks for vectorized unpacking we are trying to prevent any perf degradations over non-X86 platforms, over the time these features checks may be relaxed once we quantify the performance improvements on other targets.

One can use JVM option -XX:+PrintIntrinsics to determine if  VectorAPI used in the source is intrinsified for a particular target.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
jiangjiguang: @gszadovszky @wgtmac  I add a new module named parquet-encoding-vector on directory plugins, the new module implements all code about Vector. I think it can keep code clean about java17 and  related class compile/unit test execution.

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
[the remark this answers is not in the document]
gszadovszky: @jatin-bhateja, thanks for the explanation. This is fine to check CPU flags. I was wondering, though if there is an existing 3rd party implementation to check things like that so we don't take the burden of maintaining this code.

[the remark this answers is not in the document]
jatin-bhateja: >  I was wondering, though if there is an existing 3rd party implementation to check things like that

Hi @gszadovszky , Any implementation using VectorAPI will need some fine tuning since algorithm may not be performant if any API does not get intrinsified for a specified target.

>  burden of maintaining this code.

I am not aware of any 3rd party java based library which vectorizes parquet-mr bit packing.  Intel has also contributed a VectorAPI base version of [PForR Java library](https://github.com/lemire/JavaFastPFOR/pull/51) which does integral bit packing which is other encoding scheme being used in parquet-mr but its a different format.

PFOR_VS_PARQUET_FORMAT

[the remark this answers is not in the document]
gszadovszky: Sorry, @jatin-bhateja, I was misleading. I was talking about this very code about checking the CPU flags not the whole vectorization. I accept we need to check the CPU flags. I only think that the way we are doing is quite low level. It might be implemented already by a 3rd party that might support other OS etc.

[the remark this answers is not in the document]
jatin-bhateja: @jiangjiguang , can you please respond to @gszadovszky  query.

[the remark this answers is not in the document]
jiangjiguang: @gszadovszky Even though the java vector API can check the CPU Flags(low level check), but if users use java vector api on computer which dose not contain avx512vbmi && avx512_vbmi2, the java vector API how to do ?  throw exception?  fall back ?  

Besides,  the CPU Flags are becoming more and more,  and  Java vector api is also increasing. maybe Java vector api dose not know which CPU Flags  the users need. 
I think  users should have already known their computer(including CPU Flags). They can decide whether to  enable parquet java17 vector optimization. 

For example, velox(https://github.com/facebookincubator/velox) shows only what CPU Flags needed.
image

pom.xml
```diff
@@ -151,6 +151,9 @@
     <module>parquet-scala</module>
     <module>parquet-thrift</module>
     <module>parquet-hadoop-bundle</module>
+    <!--
+    <module>plugins/parquet-encoding-vector</module>
```
wgtmac, line 155: This means user has to remove the comment to enable it which seems not very friendly. We can use `<profile>` to enable module optionally.

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ByteBitPackingVectorBenchmarks.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * This class uses the java17 vector API, add VM options --add-modules=jdk.incubator.vector
+ */
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@Warmup(iterations = 1, batchSize = 100000)
+@Measurement(iterations = 1, batchSize = 100000)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class ByteBitPackingVectorBenchmarks {
```
wgtmac, line 45: If the parquet-encoding-vector does not build, what will happen to this file? Does it fail?

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/Packer.java
```diff
@@ -86,6 +105,7 @@ private static Object getStaticField(String className, String fieldName) {
   static IntPackerFactory leIntPackerFactory = getIntPackerFactory("LemireBitPackingLE");
   static BytePackerFactory beBytePackerFactory = getBytePackerFactory("ByteBitPackingBE");
   static BytePackerFactory leBytePackerFactory = getBytePackerFactory("ByteBitPackingLE");
+  static BytePackerFactory leBytePacker512VectorFactory = null;
```
wgtmac, line 108: Please add a comment to explain leBytePacker512VectorFactory will be initialized lazily.

plugins/parquet-encoding-vector/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>parquet-encoding-vector</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Apache Parquet Encodings Vector</name>
+  <url>https://parquet.apache.org</url>
+
+  <properties>
+    <maven.compiler.source>17</maven.compiler.source>
```
wgtmac, line 38: If we add a new profile in the root pom file, then we can set compiler version there and avoid hardcode here.

plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/VectorSupport.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bitpacking;
+
+/**
+ * Vector bits which has be supported
```
wgtmac, line 22:
```suggestion
 * Supported bit widths to use Vector API.
```

plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
```diff
@@ -0,0 +1,3010 @@
...
+import jdk.incubator.vector.Vector;
+import jdk.incubator.vector.VectorMask;
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorShuffle;
+import jdk.incubator.vector.VectorSpecies;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
```
wgtmac, line 35: Remove it or add something?

README.md
```diff
@@ -83,6 +83,16 @@ Parquet is a very active project, and new features are being added quickly. Here
...
+Parquet-MR has supported Java Vector API to speed up reading, to enable the function:
+* Java 17+, 64-bit
+* For Intel CPU, Flags containing avx512vbmi and avx512_vbmi2 can have better performance gains(ICE Lake or newer processor).
+* mvn clean package -P java17-target -P vector
```
wgtmac, line 92: Please update the document to reflect the latest change.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
wgtmac: +1 for the option 2. But why not enable it by a new profile?

gszadovszky: I agree with @wgtmac. Option 2 with maven profile support seems the best way to go.

README.md
jiangjiguang: I have updated the document

plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
jiangjiguang: I have added comments

parquet-encoding/src/main/java/org/apache/parquet/column/values/bitpacking/Packer.java
jiangjiguang: I have added some comments

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/ByteBitPackingVectorBenchmarks.java
jiangjiguang: Yes, it will fail, so I new a new module plugins/parquet-plugins-benchmarks  and put the class into the module.

plugins/parquet-encoding-vector/pom.xml
jiangjiguang: I have removed the property  and use <release>17</release> like parent pom.xml
plugins/parquet-encoding-vector is independent and it needs some configurations(such as extraJavaVectorArgs), but other modules will not need, so I think it is best placed in its own module

pom.xml
jiangjiguang: I have removed it, user need go in plugins/parquet-encoding-vector and build with maven

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
jiangjiguang: @gszadovszky @wgtmac I agree with you, but I can't solve how to exclude a module(parquet-encoding-vector) in the profile by default.
In fact, The current solution is also simply and user only go to the plugins/parquet-encoding-vector and build with maven

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
jiangjiguang: I have checked it in instead of having the generator thing

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/ParquetReadRouter.java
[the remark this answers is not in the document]
jiangjiguang: @gszadovszky I got it. As far as I know, there is no such library available for java. I think we can try to achieve implement a library like that with JNI in the future. But there is no time line. If I have finished it, I will update here.

parquet-generator/src/main/java/org/apache/parquet/encoding/vectorbitpacking/BitPackingGenerator512Vector.java
wgtmac: Not sure whether this answer works in this case: https://stackoverflow.com/a/13383092

jiangjiguang: @wgtmac It makes sense. I have added profile plugins.  mvn clean package -P plugins  to enable this feature.

pom.xml
```diff
@@ -659,5 +662,13 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>plugins</id>
```
wgtmac, line 667: `plugins` is a little bit generic to me. Rename it to `encoding-plugin` or `vector-plugins`? Any suggestion?  @gszadovszky 

In addition, please rename `plugins` folder to `parquet-plugins` to follow naming of other sub-directories.

wgtmac (commented):
Looks good. A minor comment mainly on naming.

pom.xml
jiangjiguang: @wgtmac  @gszadovszky  vector-plugins +1, and it can show the feature of this PR. 
 I have renamed plugins to parquet-plugins.

gszadovszky (changes requested):
It looks good to me. I've added a note for the README, though.

One more thing. We need to add testing of this profile automatically with github actions. (The current test results for this PR have no meaning since these tests are not executed.)

README.md
```diff
@@ -83,6 +83,20 @@ Parquet is a very active project, and new features are being added quickly. Here
 * Column stats
 * Delta encoding
 * Index pages
+* Java Vector API support
+
+## Java Vector API support
+Parquet-MR has supported Java Vector API to speed up reading, to enable this feature:
```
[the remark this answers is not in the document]
jiangjiguang: I have added the experimental flag

jiangjiguang:
> 
@gszadovszky @wgtmac  This feature need avx512vbmi and avx512_vbmi2 instruction set, so it needs github action runners with intel ice lake. I do not know how to select runners with Intel Ice Lake ? So I have submitted the help (https://github.com/actions/runner/issues/2467).

jiangjiguang:
> > 
> 
> @gszadovszky @wgtmac This feature need avx512vbmi and avx512_vbmi2 instruction set, so it needs github action runners with intel ice lake. I do not know how to select runners with Intel Ice Lake ? So I have submitted the help ([actions/runner#2467](https://github.com/actions/runner/issues/2467)).

@gszadovszky  @wgtmac

jiangjiguang:
@gszadovszky @wgtmac  I have added a new workflow named Vector-plugins, can you run it ?

wgtmac:
> @gszadovszky @wgtmac I have added a new workflow named Vector-plugins, can you run it ?

It seems that this PR is closed. Could you please reopen it and see if it can run automatically?

jiangjiguang:
> > @gszadovszky @wgtmac I have added a new workflow named Vector-plugins, can you run it ?
> 
> It seems that this PR is closed. Could you please reopen it and see if it can run automatically?

@wgtmac sorry, may be my wrong click, I have reopened it

wgtmac:
> @wgtmac sorry, may be my wrong click, I have reopened it

NP. Seems it is running. https://github.com/apache/parquet-mr/actions/runs/4300429406/jobs/7496638049

.github/workflows/vector-plugins.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '17' ]
+        codes: [ 'uncompressed,brotli', 'gzip,snappy' ]
```
wgtmac, line 30: To reduce some resource consumption, `uncompressed` seems enough.

.github/workflows/vector-plugins.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: before_install
+        env:
+          CI_TARGET_BRANCH: $GITHUB_HEAD_REF
+        run: |
+          bash dev/ci-before_install.sh
+      - name: install
+        run: |
+          EXTRA_JAVA_TEST_ARGS=$(mvn help:evaluate -Dexpression=extraJavaTestArgs -q -DforceStdout)
+          export MAVEN_OPTS="$MAVEN_OPTS $EXTRA_JAVA_TEST_ARGS"
+          mvn install --batch-mode -Pvector-plugins -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true -Djava.version=${{ matrix.java }} -pl -parquet-hadoop,-parquet-arrow,-parquet-avro,-parquet-benchmarks,-parquet-cli,-parquet-column,-parquet-hadoop-bundle,-parquet-jackson,-parquet-pig,-parquet-pig-bundle,-parquet-protobuf,-parquet-thrift
```
wgtmac, line 48: Why is `-pl -parquet-hadoop,-parquet-arrow,-parquet-avro,-parquet-benchmarks,-parquet-cli,-parquet-column,-parquet-hadoop-bundle,-parquet-jackson,-parquet-pig,-parquet-pig-bundle,-parquet-protobuf,-parquet-thrift` required here?

jiangjiguang: because they have been run in the Test workflow. I think vector-plugins should run only the module associated with it

.github/workflows/vector-plugins.yml
jiangjiguang: it makes sense

.github/workflows/vector-plugins.yml
jiangjiguang: @wgtmac I updated the vector-plugins workflow, it only specifies modules needed to execute

wgtmac (approved):
LGTM. Thanks @jiangjiguang !

wgtmac:
I'd request sign off from @gszadovszky @shangxinli

.github/workflows/vector-plugins.yml
```diff
@@ -0,0 +1,56 @@
...
+          export MAVEN_OPTS="$MAVEN_OPTS $EXTRA_JAVA_TEST_ARGS"
+          mvn install --batch-mode -Pvector-plugins -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true -Djava.version=${{ matrix.java }} -pl parquet-encoding,parquet-plugins/parquet-encoding-vector,parquet-plugins/parquet-plugins-benchmarks
+      - name: verify
+        env:
+          TEST_CODECS: ${{ matrix.codes }}
+          JAVA_VERSION: ${{ matrix.java }}
+        run: |
+          EXTRA_JAVA_TEST_ARGS=$(mvn help:evaluate -Dexpression=extraJavaTestArgs -q -DforceStdout)
+          export MAVEN_OPTS="$MAVEN_OPTS $EXTRA_JAVA_TEST_ARGS"
+          mvn verify --batch-mode javadoc:javadoc -Pci-test
```
gszadovszky, line 56: Shouldn't we use `-Pvector-plugins` here as well? Without it we won't actually run the related unit tests only compile the code in the `install` phase.
BTW, do we want to execute every tests in this workflow? I've thought the intent is to compile only the necessary modules to execute the tests of `vector-plugins` only.

parquet-plugins/parquet-encoding-vector/src/test/java/org/apache/parquet/column/values/bitpacking/TestByteBitPacking512VectorLE.java
```diff
@@ -0,0 +1,172 @@
...
+import static org.junit.Assert.assertArrayEquals;
+
+public class TestByteBitPacking512VectorLE {
+  private static final Logger LOG = LoggerFactory.getLogger(TestByteBitPacking512VectorLE.class);
+  @Test
+  public void unpackValuesUsingVector() {
+    if (ParquetReadRouter.getSupportVectorFromCPUFlags() != VectorSupport.VECTOR_512) {
+      LOG.info("avx512vbmi and avx512_vbmi2 are not supported, skip this test.");
+      return;
+    }
```
gszadovszky, line 39: This way this is not a quality gate. I think we should not allow bypassing this test here but somehow ensure that the environment is capable of executing it. What do you think?

parquet-plugins/parquet-encoding-vector/src/test/java/org/apache/parquet/column/values/bitpacking/TestParquetReadRouter.java
```diff
@@ -0,0 +1,61 @@
...
+      }
+      ByteBufferInputStream inputStream = ByteBufferInputStream.wrap(ByteBuffer.wrap(input));
+
+      ParquetReadRouter.read(bitWidth, inputStream, 0, output);
+      ParquetReadRouter.readBatch(bitWidth, inputStream, 0, outputBatch);
+      assertArrayEquals(output, outputBatch);
+      if (ParquetReadRouter.getSupportVectorFromCPUFlags() != VectorSupport.VECTOR_512) {
+        LOG.info("avx512vbmi and avx512_vbmi2 are not supported, skip this test.");
+        return;
+      }
```
gszadovszky, line 56: See my previous comment in the other test class.

gszadovszky (changes requested):
I've added some comments about testing. @jiangjiguang, do you have any info about selecting the proper CPU environment for these tests?

parquet-plugins/parquet-encoding-vector/src/test/java/org/apache/parquet/column/values/bitpacking/TestByteBitPacking512VectorLE.java
jiangjiguang: @gszadovszky I agree with you. 
1、I have verified with `lscpu command` that there are not avx512vbmi and avx512_vbmi2 instruction set on actions runner.
2、I have checked the docs(https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), github actions do not support to select runner with specific instruction set.
3、I have resubmitted the help on how to select specific runner(https://github.com/orgs/community/discussions/48955)
I think there are two ways to fix it:
a. skip vector related tests until github actions support to select specific runner.
b. I will add a self-hosted runner for parquet-mr repo with cloud instance  to run vector related tests. But the runner is not long time since I have to pay money for it.

What do you think ?  @gszadovszky

gszadovszky: @jiangjiguang, I wouldn't expect you to spend money on it. I accept there is no proper solution for now. Since this is more for the developers who work on this part and it does not part of the official parquet-mr distribution I am fine with not enforcing these tests to be executed.
I would suggest to use `Assume` from `junit`. This way the tests would be marked as _skipped_ instead of having them _passed_.

.github/workflows/vector-plugins.yml
jiangjiguang: fix it

parquet-plugins/parquet-encoding-vector/src/test/java/org/apache/parquet/column/values/bitpacking/TestByteBitPacking512VectorLE.java
jiangjiguang: @gszadovszky thanks, I have used assume from junit and added vector-plugins in vector-plugins.yml.

please take a look.

gszadovszky (approved):
It looks good to me.
Thank you for working on it, @jiangjiguang!

jatin-bhateja:
Hi @gszadovszky , @wgtmac Thanks for your reviews and guidance during this long review process. May we request to you please merge this PR by adding following people as  co-authors ( jiangjiguang jiyu1021 guangzegu Fang-Xie jatin-bhateja) alternatively make @jiangjiguang a committer so that he can merge on our behalf.

jiangjiguang:
> Hi @gszadovszky , @wgtmac Thanks for your reviews and guidance during this long review process. May we request to you please merge this PR by adding following people as co-authors ( @jiangjiguang @jiyu1021 @guangzegu @Fang-Xie @jatin-bhateja) alternatively make @jiangjiguang a committer so that he can merge on our behalf.

@jatin-bhateja I have added them as co-authors on the first commit.
image

wgtmac:
I have merged it. Thanks @jiangjiguang @jatin-bhateja for the contribution and @gszadovszky for the review!

LuciferYang:
@jiangjiguang Sorry to bother you, I am not sure if the use of the AVX 512 will still make other CPU cores to downshift frequency. If so, is there a way to manually turn off this feature now? Otherwise, the new version will be difficult to promote and use in the enterprise.

CascadingCloud:
> @jiangjiguang Sorry to bother you, I am not sure if the use of the AVX 512 will still make other CPU cores to downshift frequency. If so, is there a way to manually turn off this feature now? Otherwise, the new version may be difficult to promote and use in the enterprise.

spark.sql.parquet.vector512.read.enabled

PR #1014, opened 2022-12-14
PARQUET-2227: Refactor several file rewriters to use a new unified ParquetRewriter implementation

wgtmac:
### Jira

- This patch aims to solve the first step of [PARQUET-2075](https://issues.apache.org/jira/browse/PARQUET-2075).

### Tests

- Make sure all tasks pass, especially CompressionConverterTest, ColumnPrunerTest, ColumnMaskerTest, and ColumnEncryptorTest.

### Commits

 - A new ParquetRewriter is introduced to unify rewriting logic.
 - RewriteOptions is defined to provide essential settings.
 - CompressionConverter, ColumnPruner, ColumnMasker, and ColumnEncryptor have been refactored and marked as deprecated.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
@@ -16,27 +16,30 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.parquet.hadoop.util;
+package org.apache.parquet.hadoop.rewrite;
 
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.Path;
-import org.apache.parquet.hadoop.rewrite.ParquetRewriter;
-import org.apache.parquet.hadoop.rewrite.RewriteOptions;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+public enum MaskMode {
+  NULLIFY("nullify"),
+  HASH("hash"),
+  REDACT("redact");
 
-import java.io.IOException;
-import java.util.List;
+  private String mode;
 
-@Deprecated
-public class ColumnPruner {
-
-  private static final Logger LOG = LoggerFactory.getLogger(ColumnPruner.class);
-
-  public void pruneColumns(Configuration conf, Path inputFile, Path outputFile, List<String> cols) throws IOException {
-    RewriteOptions options = new RewriteOptions.Builder(conf, inputFile, outputFile).prune(cols).build();
-    ParquetRewriter rewriter = new ParquetRewriter(options);
-    rewriter.processBlocks();
-    rewriter.close();
+  MaskMode(String text) {
+    this.mode = text;
   }
+
+  public String getMode() {
+    return this.mode;
+  }
+
+  public static MaskMode fromString(String mode) {
+    for (MaskMode b : MaskMode.values()) {
+      if (b.mode.equalsIgnoreCase(mode)) {
+        return b;
+      }
+    }
+    return null;
+  }
+
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
@@ -18,6 +18,8 @@
  */
 package org.apache.parquet.hadoop.rewrite;
 
+import org.apache.parquet.Preconditions;
+
 public enum MaskMode {
   NULLIFY("nullify"),
   HASH("hash"),
@@ -26,20 +28,11 @@ public enum MaskMode {
   private String mode;
 
   MaskMode(String text) {
+    Preconditions.checkArgument(text != null, "Text of mask mode is required");
     this.mode = text;
   }
 
   public String getMode() {
     return this.mode;
   }
-
-  public static MaskMode fromString(String mode) {
-    for (MaskMode b : MaskMode.values()) {
-      if (b.mode.equalsIgnoreCase(mode)) {
-        return b;
-      }
-    }
-    return null;
-  }
-
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
@@ -21,6 +21,7 @@ package org.apache.parquet.hadoop.rewrite;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.HadoopReadOptions;
+import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.column.ColumnReader;
@@ -95,7 +96,7 @@ public class ParquetRewriter implements Closeable {
   private ParquetMetadata meta;
   private MessageType schema;
   private String createdBy;
-  private CompressionCodecName codecName = null;
+  private CompressionCodecName newCodecName = null;
   private List<String> pruneColumns = null;
   private Map<ColumnPath, MaskMode> maskColumns = null;
   private Set<ColumnPath> encryptColumns = null;
@@ -106,13 +107,13 @@ public class ParquetRewriter implements Closeable {
     Path outPath = options.getOutputFile();
     Configuration conf = options.getConf();
 
-    // TODO: set more member variables
-    codecName = options.getCodecName();
+    newCodecName = options.getNewCodecName();
     pruneColumns = options.getPruneColumns();
 
     // Get file metadata and full schema from the input file
     meta = ParquetFileReader.readFooter(conf, inPath, NO_FILTER);
     schema = meta.getFileMetaData().getSchema();
+    // TODO: find a better way to keep original createdBy info and add ParquetRewriter info to it.
     createdBy = meta.getFileMetaData().getCreatedBy();
 
     // Prune columns if specified
@@ -139,7 +140,6 @@ public class ParquetRewriter implements Closeable {
     if (options.getEncryptColumns() != null && options.getFileEncryptionProperties() != null) {
       this.encryptColumns = convertToColumnPaths(options.getEncryptColumns());
       this.encryptMode = true;
-      // TODO: make sure options.getFileEncryptionProperties() is set
     }
 
     reader = new TransParquetFileReader(
@@ -167,7 +167,7 @@ public class ParquetRewriter implements Closeable {
     this.meta = meta;
     this.schema = schema;
     this.createdBy = createdBy == null ? meta.getFileMetaData().getCreatedBy() : createdBy;
-    this.codecName = codecName;
+    this.newCodecName = codecName;
     if (maskColumns != null && maskMode != null) {
       this.maskColumns = new HashMap<>();
       for (String col : maskColumns) {
@@ -194,7 +194,7 @@ public class ParquetRewriter implements Closeable {
       BlockMetaData blockMetaData = meta.getBlocks().get(blockId);
       List<ColumnChunkMetaData> columnsInOrder = blockMetaData.getColumns();
 
-      for (int i = 0, columnId = 0; i < columnsInOrder.size(); i += 1) {
+      for (int i = 0, columnId = 0; i < columnsInOrder.size(); i++) {
         ColumnChunkMetaData chunk = columnsInOrder.get(i);
         ColumnDescriptor descriptor = descriptorsMap.get(chunk.getPath());
 
@@ -210,12 +210,14 @@ public class ParquetRewriter implements Closeable {
         }
 
         reader.setStreamPosition(chunk.getStartingPos());
-        CompressionCodecName newCodecName = codecName == null ? chunk.getCodec() : codecName;
-        EncryptorRunTime encryptorRunTime = null;
-        if (this.encryptMode) {
-          encryptorRunTime = new EncryptorRunTime(writer.getEncryptor(), chunk, blockId, columnId);
+        CompressionCodecName newCodecName = this.newCodecName == null ? chunk.getCodec() : this.newCodecName;
+        ColumnChunkEncryptorRunTime columnChunkEncryptorRunTime = null;
+        boolean encryptColumn = false;
+        if (encryptMode) {
+          columnChunkEncryptorRunTime =
+                  new ColumnChunkEncryptorRunTime(writer.getEncryptor(), chunk, blockId, columnId);
+          encryptColumn = encryptColumns != null && encryptColumns.contains(chunk.getPath());
         }
-        boolean encryptColumn = encryptColumns != null && encryptColumns.contains(chunk.getPath());
 
         if (maskColumns != null && maskColumns.containsKey(chunk.getPath())) {
           // Mask column and compress it again.
@@ -226,14 +228,22 @@ public class ParquetRewriter implements Closeable {
               throw new IOException(
                       "Required column [" + descriptor.getPrimitiveType().getName() + "] cannot be nullified");
             }
-            nullifyColumn(descriptor, chunk, crStore, writer, schema, newCodecName, encryptorRunTime, encryptColumn);
+            nullifyColumn(
+                    descriptor,
+                    chunk,
+                    crStore,
+                    writer,
+                    schema,
+                    newCodecName,
+                    columnChunkEncryptorRunTime,
+                    encryptColumn);
           } else {
             throw new UnsupportedOperationException("Only nullify is supported for now");
           }
-        } else if (encryptMode || codecName != null) {
+        } else if (encryptMode || this.newCodecName != null) {
           // Translate compression and/or encryption
           writer.startColumn(descriptor, crStore.getColumnReader(descriptor).getTotalValueCount(), newCodecName);
-          processChunk(chunk, newCodecName, encryptorRunTime, encryptColumn);
+          processChunk(chunk, newCodecName, columnChunkEncryptorRunTime, encryptColumn);
           writer.endColumn();
         } else {
           // Nothing changed, simply copy the binary data.
@@ -254,7 +264,7 @@ public class ParquetRewriter implements Closeable {
 
   private void processChunk(ColumnChunkMetaData chunk,
                             CompressionCodecName newCodecName,
-                            EncryptorRunTime encryptorRunTime,
+                            ColumnChunkEncryptorRunTime columnChunkEncryptorRunTime,
                             boolean encryptColumn) throws IOException {
     CompressionCodecFactory codecFactory = HadoopCodecs.newFactory(0);
     CompressionCodecFactory.BytesInputDecompressor decompressor = null;
@@ -272,13 +282,13 @@ public class ParquetRewriter implements Closeable {
     byte[] dataPageAAD = null;
     byte[] dictPageHeaderAAD = null;
     byte[] dataPageHeaderAAD = null;
-    if (encryptorRunTime != null) {
-      metaEncryptor = encryptorRunTime.getMetaDataEncryptor();
-      dataEncryptor = encryptorRunTime.getDataEncryptor();
-      dictPageAAD = encryptorRunTime.getDictPageAAD();
-      dataPageAAD = encryptorRunTime.getDataPageAAD();
-      dictPageHeaderAAD = encryptorRunTime.getDictPageHeaderAAD();
-      dataPageHeaderAAD = encryptorRunTime.getDataPageHeaderAAD();
+    if (columnChunkEncryptorRunTime != null) {
+      metaEncryptor = columnChunkEncryptorRunTime.getMetaDataEncryptor();
+      dataEncryptor = columnChunkEncryptorRunTime.getDataEncryptor();
+      dictPageAAD = columnChunkEncryptorRunTime.getDictPageAAD();
+      dataPageAAD = columnChunkEncryptorRunTime.getDataPageAAD();
+      dictPageHeaderAAD = columnChunkEncryptorRunTime.getDictPageHeaderAAD();
+      dataPageHeaderAAD = columnChunkEncryptorRunTime.getDataPageHeaderAAD();
     }
 
     ColumnIndex columnIndex = reader.readColumnIndex(chunk);
@@ -535,7 +545,7 @@ public class ParquetRewriter implements Closeable {
       }
     }
 
-    currentPath.remove(fieldName);
+    currentPath.remove(currentPath.size() - 1);
     return prunedField;
   }
 
@@ -553,10 +563,10 @@ public class ParquetRewriter implements Closeable {
                              ParquetFileWriter writer,
                              MessageType schema,
                              CompressionCodecName newCodecName,
-                             EncryptorRunTime encryptorRunTime,
+                             ColumnChunkEncryptorRunTime columnChunkEncryptorRunTime,
                              boolean encryptColumn) throws IOException {
-    // TODO: support encryption
-    if (encryptorRunTime != null) {
+    // TODO: support nullifying and encrypting same column.
+    if (columnChunkEncryptorRunTime != null) {
       throw new RuntimeException("Nullifying and encrypting column is not implemented yet");
     }
     long totalChunkValues = chunk.getValueCount();
@@ -657,7 +667,7 @@ public class ParquetRewriter implements Closeable {
     @Override public GroupConverter asGroupConverter() { return new DummyGroupConverter(); }
   }
 
-  private static class EncryptorRunTime {
+  private static class ColumnChunkEncryptorRunTime {
     private final InternalColumnEncryptionSetup colEncrSetup;
     private final BlockCipher.Encryptor dataEncryptor;
     private final BlockCipher.Encryptor metaDataEncryptor;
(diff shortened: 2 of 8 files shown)


wgtmac:
Can you please take a look when you have time? @shangxinli @gszadovszky @ggershinsky

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -0,0 +1,144 @@
...
+  final CompressionCodecName codecName;
+  final Map<String, MaskMode> maskColumns;
+  final List<String> encryptColumns;
+  final FileEncryptionProperties fileEncryptionProperties;
+
+  private RewriteOptions(Configuration conf,
+                         Path inputFile,
+                         Path outputFile,
+                         List<String> pruneColumns,
+                         CompressionCodecName codecName,
```
shangxinli, line 44: Maybe use the variable 'newCodecName'?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -0,0 +1,144 @@
...
+      return this;
+    }
+
+    public Builder encryptionProperties(FileEncryptionProperties fileEncryptionProperties) {
+      this.fileEncryptionProperties = fileEncryptionProperties;
+      return this;
+    }
+
+    public RewriteOptions build() {
+      // TODO: validate any conflict setting
```
shangxinli, line 132: Can you open a Jira to track it? If you can do it in this PR, it is even better.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+    Configuration conf = options.getConf();
+
+    // TODO: set more member variables
+    codecName = options.getCodecName();
+    pruneColumns = options.getPruneColumns();
+
+    // Get file metadata and full schema from the input file
+    meta = ParquetFileReader.readFooter(conf, inPath, NO_FILTER);
+    schema = meta.getFileMetaData().getSchema();
+    createdBy = meta.getFileMetaData().getCreatedBy();
```
shangxinli, line 116: There was some discussion should we carry over the old author or replace it with this rewriter author. Or we can append 'reCreatedBy' but it needs a specification (parquet-format) change. Any thoughts on it?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+
+    // Prune columns if specified
+    if (pruneColumns != null && !pruneColumns.isEmpty()) {
+      List<String> paths = new ArrayList<>();
+      getPaths(schema, paths, null);
```
shangxinli, line 121: Should we move the line 120 to be inside the getPath()?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+            Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+
+    int blockId = 0;
+    while (store != null) {
+      writer.startBlock(store.getRowCount());
+
+      BlockMetaData blockMetaData = meta.getBlocks().get(blockId);
+      List<ColumnChunkMetaData> columnsInOrder = blockMetaData.getColumns();
+
+      for (int i = 0, columnId = 0; i < columnsInOrder.size(); i += 1) {
```
shangxinli, line 197: Can we use i++?

shangxinli:
I just left some comments initially. I will spend more time on it. 

@ggershinsky If you have time, can you have a look too?

shangxinli:
If you can add more unit tests, particularly the combinations of prune, mask, trans-compression etc.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+          throw new IOException("Column " + chunk.getPath().toDotString() + " is already encrypted");
+        }
+
+        reader.setStreamPosition(chunk.getStartingPos());
+        CompressionCodecName newCodecName = codecName == null ? chunk.getCodec() : codecName;
+        EncryptorRunTime encryptorRunTime = null;
+        if (this.encryptMode) {
+          encryptorRunTime = new EncryptorRunTime(writer.getEncryptor(), chunk, blockId, columnId);
+        }
+        boolean encryptColumn = encryptColumns != null && encryptColumns.contains(chunk.getPath());
```
shangxinli, line 218: Can we apply this.encryptMode to this check also?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+      } else {
+        List<Type> childFields = ((GroupType) field).getFields();
+        List<Type> prunedFields = pruneColumnsInFields(childFields, currentPath, prunePaths);
+        if (prunedFields.size() > 0) {
+          prunedField = ((GroupType) field).withNewFields(prunedFields);
+        }
+      }
+    }
+
+    currentPath.remove(fieldName);
```
shangxinli, line 538: This line seems a bug(even in the original code). We should remove by index instead of by name. Something like this: currentPath.remove(currentPath.size() - 1)

ggershinsky:
> I just left some comments initially. I will spend more time on it.
> 
> @ggershinsky If you have time, can you have a look too?

sure, will do

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+    @Override public void start() {}
+    @Override public void end() {}
+    @Override public Converter getConverter(int fieldIndex) { return new DummyConverter(); }
+  }
+
+  private static final class DummyConverter extends PrimitiveConverter {
+    @Override public GroupConverter asGroupConverter() { return new DummyGroupConverter(); }
+  }
+
+  private static class EncryptorRunTime {
```
ggershinsky, line 660: This is a per-column-chunk collection, maybe the class name should reflect that?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+    private byte[] dataPageHeaderAAD;
+    private byte[] dataPageAAD;
+    private byte[] dictPageHeaderAAD;
+    private byte[] dictPageAAD;
+
+    public EncryptorRunTime(InternalFileEncryptor fileEncryptor,
+                            ColumnChunkMetaData chunk,
+                            int blockId,
+                            int columnId) throws IOException  {
+      if (fileEncryptor == null) {
```
ggershinsky, line 675: should this object be created at all if the file encryptor is null?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+        this.dictPageHeaderAAD = createAAD(colEncrSetup, ModuleType.DictionaryPageHeader, blockId, columnId);
+        this.dictPageAAD = createAAD(colEncrSetup, ModuleType.DictionaryPage, blockId, columnId);
+      }
+    }
+
+    private byte[] createAAD(InternalColumnEncryptionSetup colEncrSetup,
+                             ModuleType moduleType,
+                             int blockId,
+                             int columnId) {
+      if (colEncrSetup != null && colEncrSetup.isEncrypted()) {
```
ggershinsky, line 702: should this be checked here, upon every call? also, what's the point of creating an EncryptorRunTime if the column is not encrypted?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -0,0 +1,733 @@
...
+      this.maskColumns = new HashMap<>();
+      for (Map.Entry<String, MaskMode> col : options.getMaskColumns().entrySet()) {
+        maskColumns.put(ColumnPath.fromDotString(col.getKey()), col.getValue());
+      }
+    }
+
+    if (options.getEncryptColumns() != null && options.getFileEncryptionProperties() != null) {
+      this.encryptColumns = convertToColumnPaths(options.getEncryptColumns());
+      this.encryptMode = true;
+      // TODO: make sure options.getFileEncryptionProperties() is set
```
ggershinsky, line 142: what is left to do here?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: IMHO, reusing the old `createdBy` is a bad idea which makes it difficult to reason about the origin of the file. What about concatenating the old `createdBy` with the writer version of `ParquetRewriter` so we can keep both old and new creator in the same field?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: This is a good question. The code was just moved from https://github.com/apache/parquet-mr/blob/1adc22804a700d78f8480667d083e91d6147339f/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ColumnEncryptor.java#L84. I think it should throw if the EncryptorRuntime is required but the fileEncryptor is null.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: If `EncryptorRunTime` was not created for non-encrypted columns, generating the file footer will get some error. It can be reproduced by the ColumnEncryptorTest.

ggershinsky: not sureI understand. The EncryptorRunTime object is created per column, so no need to run this check for every module aad; this boolean can be built in the EncryptorRunTime constructor. Also, it would be more logical to use EncryptorRunTime objects only for encrypted columns.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
```diff
@@ -0,0 +1,45 @@
...
+
+public enum MaskMode {
+  NULLIFY("nullify"),
+  HASH("hash"),
+  REDACT("redact");
+
+  private String mode;
+
+  MaskMode(String text) {
+    this.mode = text;
```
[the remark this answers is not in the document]
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
wgtmac: Fixed and removed the TODO

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: I'd rather keep the same behavior for now and fix it in a separate JIRA.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/MaskMode.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public String getMode() {
+    return this.mode;
+  }
+
+  public static MaskMode fromString(String mode) {
```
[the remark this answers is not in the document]
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Renamed it to ColumnChunkEncryptorRunTime

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed it by checking `colEncrSetup` only once per creation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: I forgot to remove this TODO.

wgtmac:
> If you can add more unit tests, particularly the combinations of prune, mask, trans-compression etc, it would be better.

I have added some test cases in the `ParquetRewriterTest` to mix config of prune, nullify and trans-compression. 

There are two outstanding issues as below:
- Masking and encrypting same column is not supported yet.
- Fix `createdBy` of rewritten file to keep both original info and new rewriter info.

I am inclined to fix them in separate JIRAs because this patch is simply a refactoring patch to unify different rewriters without changing any behavior, and it is large enough.

@ggershinsky @shangxinli Please review again when you have time. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -0,0 +1,178 @@
...
+          for (String pruneColumn : pruneColumns) {
+            Preconditions.checkArgument(!maskColumns.containsKey(pruneColumn),
+                    "Cannot prune and mask same column");
+          }
+        }
+
+        if (encryptColumns != null) {
+          for (String pruneColumn : pruneColumns) {
+            Preconditions.checkArgument(!encryptColumns.contains(pruneColumn),
+                    "Cannot prune and mask same column");
```
ggershinsky, line 149: Cannot prune and encrypt same column?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -0,0 +1,178 @@
...
+      // TODO: support masking and encrypting same columns
+      if (maskColumns != null && encryptColumns != null) {
+        for (String encryptColumn : encryptColumns) {
+          Preconditions.checkArgument(!maskColumns.containsKey(encryptColumn),
+                  "Cannot encrypt and mask same column");
+        }
+      }
+
+      if (encryptColumns != null && !encryptColumns.isEmpty()) {
+        Preconditions.checkArgument(fileEncryptionProperties != null,
```
ggershinsky, line 163: what if `fileEncryptionProperties` was set, but the `encryptColumns` is null or `encryptColumns.isEmpty()`?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -0,0 +1,308 @@
...
+import org.apache.parquet.internal.column.columnindex.OffsetIndex;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.*;
```
ggershinsky, line 52: please don't use * (also in lines 57, 58)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -0,0 +1,308 @@
...
+public class ParquetRewriterTest {
+
+  private final int numRecord = 100000;
+  private Configuration conf = new Configuration();
+  private EncryptionTestFile inputFile = null;
+  private String outputFile = null;
+  private ParquetRewriter rewriter = null;
+
+  @Test
+  public void testPruneSingleColumnAndTranslateCodec() throws Exception {
```
ggershinsky, line 69: do we also have an encryption rewrite unitest?

shangxinli:
@gszadovszky I Just want to check if you have time to have a look. @wgtmac Just be nice to take over the work that we discussed earlier to have an aggregated rewriter.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
wgtmac: Yes, pruned column does not exist in the rewritten file. Then it does not make sense to encrypt the missing column any more.

gszadovszky:
> @gszadovszky I Just want to check if you have time to have a look. @wgtmac just be nice to take over the work that we discussed earlier to have an aggregated rewriter.

@shangxinli, I'll try to take a look this week.

shangxinli:
Thanks a lot @gszadovszky

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
wgtmac: I have added checking to disable this case.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
wgtmac: Fixed

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
wgtmac: Added a test case to prune, transcode and encrypt columns.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
ggershinsky: Yep. I meant the exception text in the line 149, it is identical to the masking check (line 142); I guess the encryption check would print "Cannot prune and encrypt same column"

wgtmac: Sorry, I misunderstood your meaning. Fixed the error message. Please take a look again. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
gszadovszky: Since we have only one field for _created by_ and it's content is more or less specified, we should write the current version of parquet-mr there. It is also a good idea to keep the original _created by_ value, though. What do you think about adding it to the `key_value_metadata` with a specific key? Even though this field would not be specified at least we won't lose the info.

gszadovszky (changes requested):
I think it is a great refactor. Thanks a lot for working on it, @wgtmac!
In the other hand I've thought about PARQUET-2075 as a request for a new feature in `parquet-cli` that can be used to convert from one parquet file to another with specific configurations. (Later on we might extend it to allow multiple parquet files to be merged/rewritten to one specified and the tool would decide which level of deserialization/serialization is required.)
I am fine with handling it in a separate jira but let's make it clear. Either create another jira for this refactor as a prerequisite of PARQUET-2075 or rephrase PARQUET-2075 and create a new for `parquet-cli`.
@shangxinli, what do you think?

wgtmac:
> I think it is a great refactor. Thanks a lot for working on it, @wgtmac! In the other hand I've thought about PARQUET-2075 as a request for a new feature in `parquet-cli` that can be used to convert from one parquet file to another with specific configurations. (Later on we might extend it to allow multiple parquet files to be merged/rewritten to one specified and the tool would decide which level of deserialization/serialization is required.) I am fine with handling it in a separate jira but let's make it clear. Either create another jira for this refactor as a prerequisite of PARQUET-2075 or rephrase PARQUET-2075 and create a new for `parquet-cli`. @shangxinli, what do you think?

Thanks for your review @gszadovszky 

- I'd prefer creating a new JIRA for this refactor to be a prerequisite. Merging multiple files to a single one with customized pruning, encryption, and codec is also in my mind and will be supported later. I will create separate JIRAs as sub-tasks of PARQUET-2075 and work on them progressively.
- Putting the original `created_by` into `key_value_metadata` is a good idea. However, it is tricky if a file has been rewritten for several times. What about adding a key named `original_created_by` to `key_value_metadata` and concatenating all old `created_by`s to it?

gszadovszky:
> * I'd prefer creating a new JIRA for this refactor to be a prerequisite. Merging multiple files to a single one with customized pruning, encryption, and codec is also in my mind and will be supported later. I will create separate JIRAs as sub-tasks of PARQUET-2075 and work on them progressively.

Perfect! :)

> * Putting the original `created_by` into `key_value_metadata` is a good idea. However, it is tricky if a file has been rewritten for several times. What about adding a key named `original_created_by` to `key_value_metadata` and concatenating all old `created_by`s to it?

It sounds good to me. Maybe have the latest one at the beginning and use the separator `'\n'`?

wgtmac:
> > * I'd prefer creating a new JIRA for this refactor to be a prerequisite. Merging multiple files to a single one with customized pruning, encryption, and codec is also in my mind and will be supported later. I will create separate JIRAs as sub-tasks of PARQUET-2075 and work on them progressively.
> 
> Perfect! :)
> 
> > * Putting the original `created_by` into `key_value_metadata` is a good idea. However, it is tricky if a file has been rewritten for several times. What about adding a key named `original_created_by` to `key_value_metadata` and concatenating all old `created_by`s to it?
> 
> It sounds good to me. Maybe have the latest one at the beginning and use the separator `'\n'`?

I am afraid some implementations may drop characters after `'\n'` when displaying the string content. Let me do some investigation.

gszadovszky:
> I am afraid some implementations may drop characters after `'\n'` when displaying the string content. Let me do some investigation.

I do not have a strong opinion for `'\n'` only that we need a character that probably won't be used by any systems writing parquet files.

wgtmac:
> > I am afraid some implementations may drop characters after `'\n'` when displaying the string content. Let me do some investigation.
> 
> I do not have a strong opinion for `'\n'` only that we need a character that probably won't be used by any systems writing parquet files.

As we are discussing a new entry (`original.created.by`) to the key value metadata, I need to raise two related issues once we have supported rewriting (merging) several files into one:
- We need to merge `original.created.by` from all input files, making it difficult to tell which created_by comes from which input file. Therefore, `original.created.by` should be dropped in this case.
- Is there any key value metadata that will conflict from different input files and should be dealt with by the rewriter? For now we simply keep all the old key value metadata from the old file.

@gszadovszky @ggershinsky @shangxinli Thoughts?

If this behavior requires further discussion, I'd suggest to keep the current state of `created_by` unchanged in this pull request which is large enough. All rewriters (ColumnPruner, CompressionConverter, ColumnMasker, and ColumnEncrypter) have dropped original `created_by` and store the current writer version to the footer.

gszadovszky:
I agree that merging the key-value metadata is not an easy question. Let's discuss it separately as it is not related to this PR.

I also agree to store the current writer (parquet-mr) in `created_by` in case of rewriting.  It is not easy to decide what would be the proper solution anyway. `created_by` is usually used for handling potential erroneous writes. Let's say there was an issue in parquet-mr at the version 1.2.3 that written a specific encoding of integers wrongly (not according to spec). What if we rewrite the file but do not re-encode the pages? Can we still handle the original issue? What if the rewriter re-encodes the related pages?
Let's store the original writer in `original.created.by` for now. Let's discuss this topic separately however, I am not sure if we can find a proper solution.

wgtmac:
> I agree that merging the key-value metadata is not an easy question. Let's discuss it separately as it is not related to this PR.
> 
> I also agree to store the current writer (parquet-mr) in `created_by` in case of rewriting. It is not easy to decide what would be the proper solution anyway. `created_by` is usually used for handling potential erroneous writes. Let's say there was an issue in parquet-mr at the version 1.2.3 that written a specific encoding of integers wrongly (not according to spec). What if we rewrite the file but do not re-encode the pages? Can we still handle the original issue? What if the rewriter re-encodes the related pages? Let's store the original writer in `original.created.by` for now. Let's discuss this topic separately however, I am not sure if we can find a proper solution.

I agree. Now I have updated this PR to preserve the old writer version into `original.created.by` and added a test to make sure it is preserved. Please take a look. Thanks!

wgtmac:
Gentle ping. @gszadovszky @ggershinsky @shangxinli

Any chance to take another look?

gszadovszky (approved):

gszadovszky:
Thank you, @wgtmac for working on this! It looks good to me.
Since there were other reviewers, I would wait a bit to give a chance to them for add feedback. I'll push it after a couple of days. (Feel free to ping me if I forget.)

shangxinli:
Thanks @wgtmac for working on this and Thanks @gszadovszky and @ggershinsky for reviewing it! I am a little late for the comments discussion but I see we are in the right direction. Let's address it in a separate discussion. If it turns out that changing the parquet-format is the right way to solve it, we can make the proposal and I can help for the approval. 

My comments are all addressed. I don't have further comments.

ggershinsky (approved):
My comments are addressed too. Thanks for working on this PR.

wgtmac:
Thank you @gszadovszky @ggershinsky @shangxinli

PR #1016, opened 2023-01-04
PARQUET-2223: Parquet Data Masking Enhancement for Column Encryption

zhangjiashen:
### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2223

### Tests

- Unit Tests

### Commits

- Add Data Masking functionalities to support skipping reading some columns

### Documentation

- [Design](https://docs.google.com/document/d/1K0juUXOg0wWXBlTSBiZYs4snCKeY5MsoVYxWEixL2qU/edit#heading=h.1kzvc0jyhlyb)

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/GroupReadSupport.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/GroupReadSupport.java
@@ -25,8 +25,10 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.example.data.Group;
 import org.apache.parquet.example.data.simple.convert.GroupRecordConverter;
 import org.apache.parquet.hadoop.api.ReadSupport;
+import org.apache.parquet.hadoop.util.DataMaskingUtil;
 import org.apache.parquet.io.api.RecordMaterializer;
 import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.MessageTypeParser;
 
 public class GroupReadSupport extends ReadSupport<Group> {
 
@@ -35,8 +37,21 @@ public class GroupReadSupport extends ReadSupport<Group> {
       Configuration configuration, Map<String, String> keyValueMetaData,
       MessageType fileSchema) {
     String partialSchemaString = configuration.get(ReadSupport.PARQUET_READ_SCHEMA);
-    MessageType requestedProjection = getSchemaForRead(fileSchema, partialSchemaString);
-    return new ReadContext(requestedProjection);
+    String removeColumns = configuration.get(DataMaskingUtil.DATA_MASKING_COLUMNS);
+
+    if (removeColumns == null) {
+      MessageType requestedProjection = getSchemaForRead(fileSchema, partialSchemaString);
+      return new ReadContext(requestedProjection);
+    }
+
+    if (partialSchemaString == null) {
+      return new ReadContext(DataMaskingUtil.removeColumnsInSchema(fileSchema, removeColumns));
+    } else {
+      MessageType updatedSchema = DataMaskingUtil.removeColumnsInSchema(
+        MessageTypeParser.parseMessageType(partialSchemaString), removeColumns);
+      MessageType requestedProjection = getSchemaForRead(fileSchema, updatedSchema.toString());
+      return new ReadContext(requestedProjection);
+    }
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/DataMaskingUtil.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.parquet.schema.Type;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class DataMaskingUtil
+{
+  public final static String DATA_MASKING_COLUMNS = "parquet.data.masking.columns";
```
wgtmac, line 33: Should we update the [doc](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/README.md#class-readsupport) to reflect the new config?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/DataMaskingUtil.java
```diff
@@ -0,0 +1,95 @@
...
+  public final static String DELIMITER = ",";
+
+  public static MessageType removeColumnsInSchema(MessageType schema, String removeColumns)
+  {
```
wgtmac, line 37: ditto

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/DataMaskingUtil.java
```diff
@@ -0,0 +1,95 @@
...
+    if (removeColumns == null || removeColumns.isEmpty()) {
+      return schema;
+    }
+
+    Set<ColumnPath> paths = getColumnPaths(removeColumns);
+    List<String> currentPath = new ArrayList<>();
+    List<Type> prunedFields = removeColumnsInFields(schema.getFields(), currentPath, paths);
+    return new MessageType(schema.getName(), prunedFields);
```
wgtmac, line 45: What if all columns have been pruned and an empty schema is returned?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/GroupReadSupport.java
```diff
@@ -35,8 +37,21 @@ public org.apache.parquet.hadoop.api.ReadSupport.ReadContext init(
       Configuration configuration, Map<String, String> keyValueMetaData,
       MessageType fileSchema) {
     String partialSchemaString = configuration.get(ReadSupport.PARQUET_READ_SCHEMA);
-    MessageType requestedProjection = getSchemaForRead(fileSchema, partialSchemaString);
-    return new ReadContext(requestedProjection);
+    String removeColumns = configuration.get(DataMaskingUtil.DATA_MASKING_COLUMNS);
```
wgtmac, line 40: It seems that only the example `ReadSupport` has enabled skipping masked columns. Does it mean that other ReadSupport implementations are required to apply the same approach? If yes, does the `AvroReadSupport` class require the change, too?

shangxinli:
@ggershinsky Do you have time to have a look?

ggershinsky:
As far as I understand, _data masking_ replaces content of sensitive columns; it does not remove the columns (schema and content). The latter is done by _column pruning_ - when re-writing a file. All of that is not related to _column encryption_. So I'm not fully sure what is the goal of the mechanism in this PR. Maybe we can start with a googledoc that describes the problem, the goals and the solution design?

ggershinsky:
I found the doc. Could you provide me with a "comment" access, so we'll discuss the goals and design there? Thanks.

zhangjiashen:
> I found the doc. Could you provide me with a "comment" access, so we'll discuss the goals and design there? Thanks.

@ggershinsky thanks for looking at this, I have added permission for you, feel free to add questions/comments!

shangxinli:
@zhangjiashen The current change is incomplete. You only port the change of utilizes to hide the columns in schema, but you need to actually hide it in the readFooter(). And before that, you need to mark those columns as hidden when access denied is thrown from KMS.

shangxinli:
@zhangjiashen Do you have time to work on this?

zhangjiashen:
> 

Created a new PR for data masking-> https://github.com/apache/parquet-mr/pull/1112

zhangjiashen:
Created a new PR for data masking-> https://github.com/apache/parquet-mr/pull/1112

PR #1018, opened 2023-01-08
PARQUET-2219: ParquetFileReader skips empty row group

wgtmac:
### Jira

My PR addresses the [PARQUET-2219](https://issues.apache.org/jira/browse/PARQUET/PARQUET-2219).

### Tests

My PR adds the following unit test to read parquet file with empty row group:
- parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetReaderEmptyBlock.java

Two test parquet files are created by C++ parquet writer from Apache Arrow.
- Single empty row group: parquet-hadoop/src/test/resources/test-empty-row-group_1.parquet
- Three row groups with empty one in the middle: parquet-hadoop/src/test/resources/test-empty-row-group_2.parquet
- For row groups with two empty ones in the middle: parquet-hadoop/src/test/resources/test-empty-row-group_3.parquet

### Commits

The parquet specs does not forbid empty row group and some implementations are able to generate files with empty row group. The commit aims to make ParquetFileReader robust by skipping empty row group while reading.


wgtmac:
@gszadovszky @ggershinsky @shangxinli @sunchao Could you please take a look when you have time?

cc @emkornfield

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetReaderEmptyBlock.java
```diff
@@ -0,0 +1,83 @@
...
+      Assert.assertNull(r.readNextRowGroup());
+    }
+  }
+
+  @Test
+  public void testSkipEmptyBlock() throws IOException {
+    Configuration conf = new Configuration();
+    ParquetMetadata readFooter = ParquetFileReader.readFooter(conf, EMPTY_BLOCK_FILE_2);
+
+    // The parquet file contains three row groups, the second one is empty
```
gszadovszky, line 64: I think, it would be nice to test the case of multiple empty row groups next to each other.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1038,7 +1044,9 @@ public PageReadStore readNextFilteredRowGroup() throws IOException {
     }
     BlockMetaData block = blocks.get(currentBlock);
     if (block.getRowCount() == 0L) {
-      throw new RuntimeException("Illegal row group of 0 rows");
+      // Skip the empty block
+      advanceToNextBlock();
+      return readNextFilteredRowGroup();
```
gszadovszky, line 1049: There is a warning log in case of `readNextRowGroup` but here we don't log anything.

gszadovszky (changes requested):
Thanks you for fixing this. I've added some comments.
Also, could you add a similar test for the filtered row groups?

shangxinli:
@gszadovszky Nice to see you are back!

wgtmac:
> Thanks you for fixing this. I've added some comments. Also, could you add a similar test for the filtered row groups?

Thanks for your review @gszadovszky !

I have addressed all of your comments. Please take a look again.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -927,7 +925,15 @@ public PageReadStore readRowGroup(int blockIndex) throws IOException {
    * @return the PageReadStore which can provide PageReaders for each column.
    */
   public PageReadStore readNextRowGroup() throws IOException {
-    ColumnChunkPageReadStore rowGroup = internalReadRowGroup(currentBlock);
+    ColumnChunkPageReadStore rowGroup = null;
+    try {
+      rowGroup = internalReadRowGroup(currentBlock);
+    } catch (ParquetEmptyBlockException e) {
+      LOG.warn("Read empty block at index {}", currentBlock);
```
shangxinli, line 932: Any way to add file path?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1038,7 +1044,10 @@ public PageReadStore readNextFilteredRowGroup() throws IOException {
     }
     BlockMetaData block = blocks.get(currentBlock);
     if (block.getRowCount() == 0L) {
-      throw new RuntimeException("Illegal row group of 0 rows");
+      LOG.warn("Read empty block at index {}", currentBlock);
```
shangxinli, line 1047: Same comments as above

gszadovszky:
> @gszadovszky Nice to see you are back!

@shangxinli, I wouldn't say I'm back, unfortunately. I'm a bit closer to Parquet at Dremio but actually not working on it. We'll see if I will have some spare time for it. :)

gszadovszky (approved):
It is good from my side

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
wgtmac: Added the file path to the log. Please take a look again. Thanks!

shangxinli (approved):

PR #1020, opened 2023-01-14
PARQUET-2226 Support merge bloom filters

yabola:
Bloom filters support merge operation. We need to collect Parquet's bloom filters of multiple files, and then generate a more comprehensive bloom filter for common use.  Guava also supports similar api
https://guava.dev/releases/31.0.1-jre/api/docs/src-html/com/google/common/hash/BloomFilter.html#line.252

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -400,14 +400,14 @@ public class BlockSplitBloomFilter implements BloomFilter {
   public void merge(BloomFilter otherBloomFilter) throws IOException {
     Preconditions.checkArgument(otherBloomFilter != null, "Cannot merge a null BloomFilter");
     Preconditions.checkArgument((getAlgorithm() == otherBloomFilter.getAlgorithm()),
-      String.format("BloomFilters must have the same algorithm (%s != %s)",
-        getAlgorithm(), otherBloomFilter.getAlgorithm()));
+      "BloomFilters must have the same algorithm (%s != %s)",
+        getAlgorithm(), otherBloomFilter.getAlgorithm());
     Preconditions.checkArgument((getHashStrategy() == otherBloomFilter.getHashStrategy()),
-      String.format("BloomFilters must have the same hashStrategy (%s != %s)",
-        getHashStrategy(), otherBloomFilter.getHashStrategy()));
+      "BloomFilters must have the same hashStrategy (%s != %s)",
+        getHashStrategy(), otherBloomFilter.getHashStrategy());
     Preconditions.checkArgument((getBitsetSize() == otherBloomFilter.getBitsetSize()),
-      String.format("BloomFilters must have the same size of bitsets (%s != %s)",
-        getBitsetSize(), otherBloomFilter.getBitsetSize()));
+      "BloomFilters must have the same size of bitsets (%s != %s)",
+        getBitsetSize(), otherBloomFilter.getBitsetSize());
     ByteArrayOutputStream otherOutputStream = new ByteArrayOutputStream();
     otherBloomFilter.writeTo(otherOutputStream);
     byte[] otherBits = otherOutputStream.toByteArray();
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -22,6 +22,7 @@ package org.apache.parquet.column.values.bloomfilter;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.io.api.Binary;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
@@ -393,5 +394,22 @@ public class BlockSplitBloomFilter implements BloomFilter {
   @Override
   public long hash(Binary value) {
     return hashFunction.hashBytes(value.getBytes());
+  }
+
+  @Override
+  public void putAll(BloomFilter otherBloomFilter) throws IOException {
+    Preconditions.checkArgument((otherBloomFilter.getAlgorithm() == getAlgorithm()),
+      "BloomFilter algorithm should be same");
+    Preconditions.checkArgument((otherBloomFilter.getHashStrategy() == getHashStrategy()),
+      "BloomFilter hashStrategy should be same");
+    Preconditions.checkArgument((otherBloomFilter.getBitsetSize() == getBitsetSize()),
+      "BloomFilter bitset size should be same");
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    otherBloomFilter.writeTo(outputStream);
+    byte[] otherBits = outputStream.toByteArray();
+    for (int i = 0; i < otherBits.length; i++) {
+      byte otherBit = otherBits[i];
+      bitset[i] |= otherBit;
+    }
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -397,16 +397,16 @@ public class BlockSplitBloomFilter implements BloomFilter {
   }
 
   @Override
-  public void putAll(BloomFilter otherBloomFilter) throws IOException {
+  public void merge(BloomFilter otherBloomFilter) throws IOException {
     Preconditions.checkArgument((otherBloomFilter.getAlgorithm() == getAlgorithm()),
       "BloomFilter algorithm should be same");
     Preconditions.checkArgument((otherBloomFilter.getHashStrategy() == getHashStrategy()),
       "BloomFilter hashStrategy should be same");
     Preconditions.checkArgument((otherBloomFilter.getBitsetSize() == getBitsetSize()),
       "BloomFilter bitset size should be same");
-    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-    otherBloomFilter.writeTo(outputStream);
-    byte[] otherBits = outputStream.toByteArray();
+    ByteArrayOutputStream otherOutputStream = new ByteArrayOutputStream();
+    otherBloomFilter.writeTo(otherOutputStream);
+    byte[] otherBits = otherOutputStream.toByteArray();
     for (int i = 0; i < otherBits.length; i++) {
       byte otherBit = otherBits[i];
       bitset[i] |= otherBit;
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -398,18 +398,21 @@ public class BlockSplitBloomFilter implements BloomFilter {
 
   @Override
   public void merge(BloomFilter otherBloomFilter) throws IOException {
-    Preconditions.checkArgument((otherBloomFilter.getAlgorithm() == getAlgorithm()),
-      "BloomFilter algorithm should be same");
-    Preconditions.checkArgument((otherBloomFilter.getHashStrategy() == getHashStrategy()),
-      "BloomFilter hashStrategy should be same");
-    Preconditions.checkArgument((otherBloomFilter.getBitsetSize() == getBitsetSize()),
-      "BloomFilter bitset size should be same");
+    Preconditions.checkArgument(otherBloomFilter != null, "Cannot merge a null BloomFilter");
+    Preconditions.checkArgument((getAlgorithm() == otherBloomFilter.getAlgorithm()),
+      String.format("BloomFilters must have the same algorithm (%s != %s)",
+        getAlgorithm(), otherBloomFilter.getAlgorithm()));
+    Preconditions.checkArgument((getHashStrategy() == otherBloomFilter.getHashStrategy()),
+      String.format("BloomFilters must have the same hashStrategy (%s != %s)",
+        getHashStrategy(), otherBloomFilter.getHashStrategy()));
+    Preconditions.checkArgument((getBitsetSize() == otherBloomFilter.getBitsetSize()),
+      String.format("BloomFilters must have the same size of bitsets (%s != %s)",
+        getBitsetSize(), otherBloomFilter.getBitsetSize()));
     ByteArrayOutputStream otherOutputStream = new ByteArrayOutputStream();
     otherBloomFilter.writeTo(otherOutputStream);
     byte[] otherBits = otherOutputStream.toByteArray();
     for (int i = 0; i < otherBits.length; i++) {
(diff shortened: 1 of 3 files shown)


yabola:
@shangxinli @gszadovszky Can you help take a look if it is suitable~

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BloomFilter.java
```diff
@@ -176,4 +176,10 @@ public String toString() {
    * @return compress algorithm that the bloom filter apply
    */
   Compression getCompression();
+
+  /**
+   * Combines this Bloom filter with another Bloom filter by performing a bitwise OR of the underlying data
+   * @param otherBloomFilter The Bloom filter to combine this Bloom filter with.
+   */
+  void putAll(BloomFilter otherBloomFilter) throws IOException;
```
wgtmac, line 184: Rename to `merge`?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -394,4 +395,21 @@ public long hash(float value) {
   public long hash(Binary value) {
     return hashFunction.hashBytes(value.getBytes());
   }
+
+  @Override
+  public void putAll(BloomFilter otherBloomFilter) throws IOException {
```
wgtmac, line 400: Could you add some tests to verify if the merged BF is as expected?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BloomFilter.java
wgtmac: IMHO, user needs to know if two BFs are compatible to merge before calling this function. So an utility function to test compatibility of two BFs is also required. WDYT?

yabola: > IMHO, user needs to know if two BFs are compatible to merge before calling this function. So an utility function to test compatibility of two BFs is also required. WDYT?

I had added some check before merging bloomFilter. I see that guava does not seem to have additional utility function.If necessary, I can add an utility function.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: > Could you add some tests to verify if the merged BF is as expected?

Sure, I will add some UT later

gszadovszky:
Thanks, @yabola for working on this and also to @wgtmac for reviewing. I do not have much experience with bloom filters so I will rely on your review. Ping me if you have a +1.
@chenjunjiedada, any thoughts?

gszadovszky:
One more thing, @yabola. The compatibility tests fail because you have added a new method to a public interface. Even though this interface is not supposed to be implemented by our clients the current project structure does not allow filtering these classes out from the compat checks. The easiest way would be to add a default implementation to the interface which throws an exception (not implemented).

shangxinli:
@chenjunjiedada Do you still have time to review this change?

yabola:
@wgtmac I had added unit test, please take a look~ thanks!

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -394,4 +395,21 @@ public long hash(float value) {
   public long hash(Binary value) {
     return hashFunction.hashBytes(value.getBytes());
   }
+
+  @Override
+  public void merge(BloomFilter otherBloomFilter) throws IOException {
+    Preconditions.checkArgument((otherBloomFilter.getAlgorithm() == getAlgorithm()),
```
wgtmac, line 401: It would be more user-friendly to print the mismatched values on both sides.

wgtmac (changes requested):
Thanks @yabola for adding tests. I just left some comments. Please let me know what do you think.

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
```diff
@@ -181,6 +182,60 @@ public void testBloomFilterNDVs(){
...
   }
 
+  @Test
+  public void testMergeBloomFilter() throws IOException {
+    Random random = new Random();
+    int numBytes = BlockSplitBloomFilter.optimalNumOfBits(1024 * 1024, 0.01) / 8;
+    BloomFilter otherBloomFilter = new BlockSplitBloomFilter(numBytes);
+    BloomFilter mergedBloomFilter = new BlockSplitBloomFilter(numBytes);
+
+    Set<String> originStrings = new HashSet<>();
```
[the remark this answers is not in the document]
yabola: I have splitted different types with different parameter, I am not sure if is suitable.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BloomFilter.java
```diff
@@ -176,4 +176,13 @@ public String toString() {
...
    */
   Compression getCompression();
+
+  /**
+   * Combines this Bloom filter with another Bloom filter by performing a bitwise OR of the underlying bits
+   *
+   * @param otherBloomFilter The Bloom filter to combine this Bloom filter with.
+   */
+  default void merge(BloomFilter otherBloomFilter) throws IOException {
+    throw new UnsupportedOperationException("Not supported merge operation.");
```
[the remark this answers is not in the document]
yabola: done

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: done

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
```diff
@@ -181,6 +182,60 @@ public void testBloomFilterNDVs(){
...
+      testDoubles.add(testDouble);
+      otherBloomFilter.insertHash(otherBloomFilter.hash(testDouble));
+
+      float testFloat = random.nextFloat();
+      testFloats.add(testFloat);
+      otherBloomFilter.insertHash(otherBloomFilter.hash(testFloat));
+    }
+    mergedBloomFilter.merge(otherBloomFilter);
+    for (String testString : originStrings) {
+      assertTrue(mergedBloomFilter.findHash(mergedBloomFilter.hash(Binary.fromString(testString))));
```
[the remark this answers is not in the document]
yabola: Because I added  random value, if the BloomFilter to be merged is not empty, there is a small probability that the two BloomFilters will be inconsistent when judging whether there is a hash value.
But if the BloomFilter to be merged is empty in the beginning, the result from these two BloomFilter should be always the same.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -398,18 +398,21 @@ public long hash(Binary value) {
...
   @Override
   public void merge(BloomFilter otherBloomFilter) throws IOException {
-    Preconditions.checkArgument((otherBloomFilter.getAlgorithm() == getAlgorithm()),
-      "BloomFilter algorithm should be same");
-    Preconditions.checkArgument((otherBloomFilter.getHashStrategy() == getHashStrategy()),
-      "BloomFilter hashStrategy should be same");
-    Preconditions.checkArgument((otherBloomFilter.getBitsetSize() == getBitsetSize()),
-      "BloomFilter bitset size should be same");
+    Preconditions.checkArgument(otherBloomFilter != null, "Cannot merge a null BloomFilter");
+    Preconditions.checkArgument((getAlgorithm() == otherBloomFilter.getAlgorithm()),
```
wgtmac, line 402: nit: you don't have to write String.format explicitly
```
    Preconditions.checkArgument((getAlgorithm() == otherBloomFilter.getAlgorithm()),
            "BloomFilters must have the same algorithm (%s != %s)",
                    getAlgorithm(), otherBloomFilter.getAlgorithm());
```

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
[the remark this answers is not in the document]
wgtmac: In this case, I'd suggest not use random values.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -394,4 +395,24 @@ public long hash(float value) {
...
+        getAlgorithm(), otherBloomFilter.getAlgorithm()));
+    Preconditions.checkArgument((getHashStrategy() == otherBloomFilter.getHashStrategy()),
+      String.format("BloomFilters must have the same hashStrategy (%s != %s)",
+        getHashStrategy(), otherBloomFilter.getHashStrategy()));
+    Preconditions.checkArgument((getBitsetSize() == otherBloomFilter.getBitsetSize()),
+      String.format("BloomFilters must have the same size of bitsets (%s != %s)",
+        getBitsetSize(), otherBloomFilter.getBitsetSize()));
+    ByteArrayOutputStream otherOutputStream = new ByteArrayOutputStream();
+    otherBloomFilter.writeTo(otherOutputStream);
+    byte[] otherBits = otherOutputStream.toByteArray();
```
wgtmac, line 413: should we check the equality of bitset.length and otherBits.length?

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
```diff
@@ -181,6 +182,83 @@ public void testBloomFilterNDVs(){
     assertTrue(bytes < 5 * 1024 * 1024);
   }
 
+  @Test
+  public void testMergeEmptyBloomFilter() throws IOException {
```
wgtmac, line 186: Probably we need a test to expect throwing exception when two BFs are not compatible.

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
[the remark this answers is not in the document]
wgtmac: Actually it does not matter what the data type is. We can simplify the test by writing two lists of hash values to create two BFs. Then we are pretty sure the test result of each value. What do you think?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: Thanks, done

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: I had checked the `getBitsetSize` before, so we may don't need to checke the bitset.length ?

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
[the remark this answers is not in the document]
yabola: > Actually it does not matter what the data type is. We can simplify the test by writing two lists of hash values to create two BFs. Then we are pretty sure the test result of each value. What do you think?

I removed the random number logic and change~

parquet-column/src/test/java/org/apache/parquet/column/values/bloomfilter/TestBlockSplitBloomFilter.java
yabola: I added a test for two BFs are not compatible.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -394,4 +395,24 @@ public long hash(float value) {
   public long hash(Binary value) {
     return hashFunction.hashBytes(value.getBytes());
   }
+
+  @Override
+  public void merge(BloomFilter otherBloomFilter) throws IOException {
```
wgtmac, line 400: I still think it is helpful to extract the check logic to a separate method like `boolean canMergeFrom(BloomFilter other)` because user has to check these parameters anyway before calling the `merge` method. Then `merge` implementation can simply call `canMergeFrom`.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -394,4 +395,24 @@ public long hash(float value) {
...
+    Preconditions.checkArgument(otherBloomFilter != null, "Cannot merge a null BloomFilter");
+    Preconditions.checkArgument((getAlgorithm() == otherBloomFilter.getAlgorithm()),
+      "BloomFilters must have the same algorithm (%s != %s)",
+        getAlgorithm(), otherBloomFilter.getAlgorithm());
+    Preconditions.checkArgument((getHashStrategy() == otherBloomFilter.getHashStrategy()),
+      "BloomFilters must have the same hashStrategy (%s != %s)",
+        getHashStrategy(), otherBloomFilter.getHashStrategy());
+    Preconditions.checkArgument((getBitsetSize() == otherBloomFilter.getBitsetSize()),
```
wgtmac, line 408: Maybe we should put this check to the 1st place to make it fail fast. Usually parameters above do not change.

wgtmac (commented):
The tests look good to me. Left some comments.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: I have added `canMergeFrom` method, please take a look if it is suitable~

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
yabola: done

wgtmac (approved):
Thanks @yabola! Looking good from my side.

Could you take another pass? @gszadovszky @gerashegalov @shangxinli

gszadovszky (approved):
Thank you @yabola and @wgtmac for working on this!

gszadovszky:
@yabola, what is your jira account? I'd like to assign the jira to you before closing.

yabola:
@wgtmac Thank you for your detailed review and @gszadovszky help.
My jira id is miracle

gszadovszky:
Sure. :)
Please double-check the jira if I assigned it to the correct one.

PR #1022, opened 2023-02-03
PARQUET-831: fix estimate page size check overflow corrupting parquet

jianchun:
When used memSize is 0, ColumnWriterV1 runs into floating point div by 0 issue, resulting in valueCountForNextSizeCheck become about 1G. (floating point div by 0 is Infinity, casting to int becomes MAX_INT.) Then it will keep buffering data and won't emit a page until 1G rows or close().

This potential too much buffering can easily overflow underlying CapacityByteArrayOutputStream, which uses int type as size count and does not check for overflow. When CapacityByteArrayOutputStream size overflows int it becomes negative. Parquet write path does not check negative size and writes asis into V1PageHeader uncompressed_page_size.

This corrupts parquet file, as parquet reader allocates uncompressed_page_size read buffer, and throws
NegativeArraySizeException when it's negative.

Fixed by applying an upper bound to valueCountForNextSizeCheck, also added overflow check.

This bug exists in 1.10.x and before.

ColumnWriteStoreV2 has a small parenthesis misplacement bug, resulting in unintended casting to float then back to long before division. So it is an integer division instead of intended floating point division. It doesn't cause problem because it applies other min/max constraints. Fixed as well. (This issue also exists in 1.11.x)

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/test/java/org/apache/parquet/column/impl/TestColumnWriterV1.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/impl/TestColumnWriterV1.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.impl;
+
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.ParquetProperties;
+import org.apache.parquet.column.page.PageWriter;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
+import org.apache.parquet.schema.Type.Repetition;
+import org.junit.Test;
+
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.anyInt;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class TestColumnWriterV1 {
+
+  @Test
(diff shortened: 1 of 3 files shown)


wgtmac:
Thanks for your contribution @jianchun 

Could you please fix the CI check first?

jianchun:
@wgtmac Thanks for looking. This PR is targeting legacy 1.10.x branch. Don't know if project still maintains 1.10.

The CI check is build config issue. From similar issue https://github.com/scylladb/scylla-tools-java/issues/137, need to fix build.properties to use https instead of http.

```
mvn.ant.tasks.download:
      [get] Getting: http://repo1.maven.org/maven2/org/apache/maven/maven-ant-tasks/2.1.3/maven-ant-tasks-2.1.3.jar
      [get] To: /home/travis/build/apache/parquet-mr/thrift-0.9.3/lib/java/build/tools/maven-ant-tasks-2.1.3.jar
      [get] Error opening connection java.io.IOException: Server returned HTTP response code: 501 for URL: http://repo1.maven.org/maven2/org/apache/maven/maven-ant-tasks/2.1.3/maven-ant-tasks-2.1.3.jar
```

parquet-column/src/test/java/org/apache/parquet/column/impl/TestColumnWriterV1.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.anyInt;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class TestColumnWriterV1 {
+
+  @Test
+  public void testEstimateNextSizeCheckOnZeroUsedMem() throws Exception {
```
wgtmac, line 38: Does the new test case fail in the master branch without your fix?

jianchun: The new test case is not applicable in master branch as later code has significantly refactored. After refactor ColumnWriterV1 is no longer responsible to slice and flush pages itself. Other components will call its "writePage" method instead.

In 0.10 and prior, ColumnWriterV1 is responsible for slicing pages where this bug resides and this test targets.

wgtmac: Thanks for the explanation!

IMHO, we may not have any plan to make a release based on the legacy version. Please advise if I have missed anything. @shangxinli

jianchun: @shangxinli @wgtmac Since this is only applicable to legacy that has no more releases, should we abandon this PR?

wgtmac: Yes, I think it does not make sense to fix something that will not be released.

jianchun: Closed, thanks

PR #1023, opened 2023-02-04
PARQUET-2237 Improve performance when filters in RowGroupFilter can match exactly

yabola:
If we can accurately judge by the minMax status, we don’t need to load the dictionary from filesystem and compare one by one anymore.

Similarly , Bloomfilter needs to load from filesystem, it may costs time and memory. If we can  exactly determine the existence/nonexistence of the value from minMax or dictionary filters , then we can avoid using Bloomfilter to Improve performance.

 
For example, 
1. read data greater than `x1` in the block, if minMax in status is all greater than `x1`, then we don't need to read dictionary from filesystem and compare one by one.
2. If we already have page dictionaries and  have compared one by one,  we don't need to read BloomFilter from filesystem and compare.
3. some more cases...


wgtmac (commented):
Thanks @yabola for the fix! The idea of the patch is good and the algorithm should be correct (at least I cannot come up with a counter example yet). However, we still need to be careful just in case. My main concern is the test coverage. In addition, we may also need an option to toggle it off just in case.

yabola:
@wgtmac Thanks for review. I will address your comments and I updated my PR description to explain in more detail.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -98,16 +99,19 @@ public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredic
 
     for (BlockMetaData block : blocks) {
       boolean drop = false;
+      // Whether one filter can exactly determine the existence/nonexistence of the value.
+      // If true then we can skip the remaining filters to save time and space.
+      AtomicBoolean canExactlyDetermine = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
yabola: It used to be for the convenience of fetching the returned results. But I will change my codes in another implemention later

parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
```diff
@@ -792,6 +793,16 @@ public void testInverseUdpMissingColumn() throws Exception {
       canDrop(LogicalInverseRewriter.rewrite(not(userDefined(fake, nullRejecting))), ccmd, dictionaries));
   }
 
+  @Test
+  public void testCanSkipOtherFilters() {
```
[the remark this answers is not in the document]
yabola: I will add more UT

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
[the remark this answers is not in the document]
yabola: I will change my codes

yabola:
please also cc @gszadovszky if you have time , I will add more UT and improve my codes.

wgtmac:
cc @gszadovszky @ggershinsky @shangxinli

yabola:
It seems code compatibility check error caused by modifying visitor return value...

yabola:
> It seems code compatibility check error caused by modifying visitor return value... can we remove that restrictions ? I will add more UT later

If we have to keep the code compatibility, can I add some flag in `DictionaryFilter` and other filters to complete such function?

yabola:
It seems code compatibility check error caused by modifying visitor return value... can we remove that restrictions ? or I should keep the code compatibility and add new flag to mark it?
I will add more UT later

wgtmac:
Unfortunately we cannot modify the signature of any public methods. My suggestion was to make the new enum serves as an internal state of the visitor (and probably use it to terminate evaluation early). Then add a new method to return the final state. Does it work?

shangxinli:
+1, let's not modify the signature.

yabola:
I thought of a way to use Boolean objects to distinguish the different types, It doesn't need to modify the visitor return type now.  But it seems Jenkins machine was broken?...

yabola:
retest this please

yabola:
@wgtmac @shangxinli I thought of a way to avoid interface modification and distinguish by Boolean objects. Please take a look

yabola:
Emmmm, If this way is not suitable, I can use the filter internal variable to represent and keep compatibility

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+ * under the License.
+ */
+
+package org.apache.parquet.filter2.compat;
+
+import org.apache.parquet.filter2.predicate.FilterPredicate;
+import org.apache.parquet.filter2.predicate.Operators;
+
+/**
+ * Used in Filters to mark whether the block data matches the condition.
```
wgtmac, line 26: We'd better explain explicitly that the evaluation decision it whether to drop the row group. That's why BLOCK_MIGHT_MATCH is false and the AND expression uses `||` in the implementation. This is counter-intuitive at the first glance.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+ * If we cannot decide whether the block matches, it will be always safe to return BLOCK_MIGHT_MATCH.
+ *
+ * We use Boolean Object here to distinguish the value type, please do not modify it.
+ */
+public class PredicateEvaluation {
+  /* The block might match, but we cannot decide yet, will check in the other filters. */
+  public static final Boolean BLOCK_MIGHT_MATCH = new Boolean(false);
```
wgtmac, line 33:
```suggestion
  public static final boolean BLOCK_MIGHT_MATCH = false;
```

wgtmac: It will deal with boxing and unboxing automatically.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+  /* The block can match for sure. */
+  public static final Boolean BLOCK_MUST_MATCH = new Boolean(false);
+  /* The block can't match for sure */
+  public static final Boolean BLOCK_CANNOT_MATCH = new Boolean(true);
+
+  public static Boolean evaluateAnd(Operators.And and, FilterPredicate.Visitor<Boolean> predicate) {
+    Boolean left = and.getLeft().accept(predicate);
```
wgtmac, line 40: In the current implementation, left and right predicates are always evaluated. We can short cut the evaluation if `left == BLOCK_CANNOT_MATCH` and skip evaluation of right predicate. It would save some cost if the right expr will read dictionary.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+      // if left and right operation must need the block, then we must take the block
+      return BLOCK_MUST_MATCH;
+    } else {
+      return BLOCK_MIGHT_MATCH;
+    }
+  }
+
+  public static Boolean evaluateOr(Operators.Or or, FilterPredicate.Visitor<Boolean> predicate) {
+    Boolean left = or.getLeft().accept(predicate);
+    Boolean right = or.getRight().accept(predicate);
```
wgtmac, line 54: ditto

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      return BLOCK_MIGHT_MATCH;
+    }
+  }
+
+  public static Boolean isDeterminedPredicate(Boolean predicate) {
+    return predicate == BLOCK_MUST_MATCH || predicate == BLOCK_CANNOT_MATCH;
+  }
+
+  public static void checkPredicate(Boolean predicate) {
```
wgtmac, line 69:
```suggestion
  public static void checkPredicate(boolean predicate) {
```

wgtmac: If `BLOCK_CANNOT_MATCH` is defined as `Boolean` but not as `boolean`, the equality check here may fail if the input predicate is created in other places. Either replace `==` with `.equals()` or use `boolean` as your choice.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,76 @@
...
+      return BLOCK_CANNOT_MATCH;
+    } else if (left == BLOCK_MUST_MATCH || right == BLOCK_MUST_MATCH) {
+      // if left or right operation must need the block, then we must take the block
+      return BLOCK_MUST_MATCH;
+    } else {
+      return BLOCK_MIGHT_MATCH;
+    }
+  }
+
+  public static Boolean isDeterminedPredicate(Boolean predicate) {
```
wgtmac, line 65: Please add some comments to the public methods.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/bloomfilterlevel/BloomFilterImpl.java
```diff
@@ -28,26 +33,29 @@
...
-
 public class BloomFilterImpl implements FilterPredicate.Visitor<Boolean>{
   private static final Logger LOG = LoggerFactory.getLogger(BloomFilterImpl.class);
-  private static final boolean BLOCK_MIGHT_MATCH = false;
-  private static final boolean BLOCK_CANNOT_MATCH = true;
 
   private final Map<ColumnPath, ColumnChunkMetaData> columns = new HashMap<ColumnPath, ColumnChunkMetaData>();
 
-  public static boolean canDrop(FilterPredicate pred, List<ColumnChunkMetaData> columns, BloomFilterReader bloomFilterReader) {
+  public static Boolean evaluate(FilterPredicate pred, List<ColumnChunkMetaData> columns, BloomFilterReader bloomFilterReader) {
```
wgtmac, line 49: Why use `Boolean` everywhere? It seems an overkill to me.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -97,21 +100,37 @@ public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredic
...
-      boolean drop = false;
+      Boolean predicate = BLOCK_MIGHT_MATCH;
+
+      if (levels.contains(FilterLevel.STATISTICS)) {
+
+        predicate = StatisticsFilter.evaluate(filterPredicate, block.getColumns());
 
-      if(levels.contains(FilterLevel.STATISTICS)) {
-        drop = StatisticsFilter.canDrop(filterPredicate, block.getColumns());
+        if(predicate == BLOCK_MUST_MATCH) {
```
wgtmac, line 109:
```suggestion
        if (predicate == BLOCK_MUST_MATCH) {
```

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -97,21 +100,37 @@ public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredic
     List<BlockMetaData> filteredBlocks = new ArrayList<BlockMetaData>();
 
     for (BlockMetaData block : blocks) {
-      boolean drop = false;
+      Boolean predicate = BLOCK_MIGHT_MATCH;
+
+      if (levels.contains(FilterLevel.STATISTICS)) {
+
```
wgtmac, line 106: Please remove this blank line.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
wgtmac: Same comment for style of lines below.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -97,21 +100,37 @@ public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredic
...
+        }
+        if(isDeterminedPredicate(predicate)) {
+          continue;
+        }
       }
 
-      if(!drop && levels.contains(FilterLevel.DICTIONARY)) {
-        drop = DictionaryFilter.canDrop(filterPredicate, block.getColumns(), reader.getDictionaryReader(block));
+      if (levels.contains(FilterLevel.DICTIONARY)) {
+
```
wgtmac, line 118: ditto

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -392,15 +420,15 @@ public <T extends Comparable<T>> Boolean visit(In<T> in) {
...
+        return predicate(dictSet, values);
       }
     } catch (IOException e) {
       LOG.warn("Failed to process dictionary for filter evaluation.", e);
     }
     return BLOCK_MIGHT_MATCH; // cannot drop the row group based on this dictionary
   }
 
-  private <T extends Comparable<T>> Boolean drop(Set<T> dictSet, Set<T> values) {
+  private <T extends Comparable<T>> Boolean predicate(Set<T> dictSet, Set<T> values) {
```
wgtmac, line 431:
```suggestion
  private <T extends Comparable<T>> Boolean evaluate(Set<T> dictSet, Set<T> values) {
```

wgtmac: Use a verb instead of noun.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -129,8 +139,9 @@ public <T extends Comparable<T>> Boolean visit(Eq<T> eq) {
         return BLOCK_MIGHT_MATCH;
       }
       // we are looking for records where v eq(null)
-      // so drop if there are no nulls in this chunk
-      return !hasNulls(meta);
+      // so if there are no nulls in this chunk, we can drop it,
+      // if there has nulls in this chunk, we must take it
+      return !hasNulls(meta) ? BLOCK_CANNOT_MATCH : BLOCK_MUST_MATCH;
```
wgtmac, line 144: I suspect that hasNull will always return precise information. When null_count is missing, hasNulls also returns true. Replacing `BLOCK_MUST_MATCH` with `BLOCK_MIGHT_MATCH` here makes more sense to me.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -289,8 +320,14 @@ public <T extends Comparable<T>> Boolean visit(Lt<T> lt) {
...
 
-    // drop if value <= min
-    return stats.compareMinToValue(value) >= 0;
+    // we are looking for records where v < someValue
+    if (stats.compareMinToValue(value) >= 0) {
+      // drop if value <= min
+      return BLOCK_CANNOT_MATCH;
+    } else {
+      // if value > min, we must take it
+      return BLOCK_MUST_MATCH;
```
wgtmac, line 329: Sometimes the min/max values are actually lower/upper bounds. Does this optimization still work for that case?

wgtmac (commented):
The new methods look reasonable to me. Could you please add more tests?

yabola:
@wgtmac Sorry, `Boolean` type has to be used here, so that we can distinguish the `BLOCK_MIGHT_MATCH` and `BLOCK_MUST_MATCH`. This is example:
```
Boolean b1 = new Boolean(true);
Boolean b2 = new Boolean(true);
boolean b3 = true;
boolean b4 = true;

assert b1 != b2;
assert b1.equals(b2);
assert b2 == b3 == b4;
```

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
yabola: Yes, thanks

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
yabola: I think  when `Statistics#hasNonNullValue` is true, it means that minMax is generated by the data content, so it can represent the real data minMax ( `Statistics#hasNonNullValue` has been processed before.)
I think if we can use minMax to judge the `BLOCK_CANNOT_MATCH` ,  we can also judge the `BLOCK_MUST_MATCH ` in some case.

yabola:
@wgtmac @shangxinli I add new UT, if you have time, please take a look, thanks!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestRowGroupFilterExactly.java
```diff
@@ -0,0 +1,303 @@
...
+    assertCorrectFiltering(or(eq(binaryColumn("name"), Binary.fromString("noneExistName")),
+      lt(longColumn("id"), 1234L)));
+  }
+
+  private void assertCorrectFiltering(FilterPredicate filter) throws IOException {
+    ParquetReadOptions readOptions = ParquetReadOptions.builder()
+      .withRecordFilter(FilterCompat.get(filter)).build();
+
+    // simulate the previous behavior, only skip other filters when predicate is BLOCK_CANNOT_MATCH
+    testEvaluation.setTestExactPredicate(Collections.singletonList(BLOCK_CANNOT_MATCH));
```
yabola, line 174: simulate the previous behavior, only skip other filters when predicate is BLOCK_CANNOT_MATCH

yabola (commented):
@wgtmac @shangxinli I add new UT `TestRowGroupFilterExactly` , if you have time, please take a look, thanks!

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,105 @@
...
+  public static final Boolean BLOCK_MIGHT_MATCH = new Boolean(false);
+  /* The block can match for sure. */
+  public static final Boolean BLOCK_MUST_MATCH = new Boolean(false);
+  /* The block can't match for sure */
+  public static final Boolean BLOCK_CANNOT_MATCH = new Boolean(true);
+
+  public static Boolean evaluateAnd(Operators.And and, FilterPredicate.Visitor<Boolean> predicate) {
+    Boolean left = and.getLeft().accept(predicate);
+    if (left == BLOCK_CANNOT_MATCH) {
+      // seems unintuitive to put an || not an && here but we can
```
wgtmac, line 45: The comment does not match the code now.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,105 @@
...
+      // drop a chunk of records if we know that either the left or
+      // the right predicate agrees that no matter what we don't
+      // need this chunk.
+      return BLOCK_CANNOT_MATCH;
+    }
+    Boolean right = and.getRight().accept(predicate);
+    if (right == BLOCK_CANNOT_MATCH) {
+      return BLOCK_CANNOT_MATCH;
+    } else if (left == BLOCK_MUST_MATCH && right == BLOCK_MUST_MATCH) {
```
wgtmac, line 54:
```suggestion
    } else if (left == BLOCK_MUST_MATCH || right == BLOCK_MUST_MATCH) {
```

wgtmac: Both left and right are not `BLOCK_CANNOT_MATCH`, so it can return `BLOCK_MUST_MATCH ` if either side is `BLOCK_MUST_MATCH`

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
```diff
@@ -0,0 +1,105 @@
...
+    if (left == BLOCK_MUST_MATCH) {
+      // if left or right operation must need the block, then we must take the block
+      return BLOCK_MUST_MATCH;
+    }
+    Boolean right = or.getRight().accept(predicate);
+    if (right == BLOCK_MUST_MATCH) {
+      // if left or right operation must need the block, then we must take the block
+      return BLOCK_MUST_MATCH;
+    } else if (left == BLOCK_CANNOT_MATCH && right == BLOCK_CANNOT_MATCH) {
+      // seems unintuitive to put an && not an || here
```
wgtmac, line 73: The comment does not match the code.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
yabola: thanks for your review,  added more UT for `And` `Or` to cover

yabola: Thanks for your review, I update comments and add more UT for `And` `Or`

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
yabola: if left is `BLOCK_MUST_MATCH` , right is might `BLOCK_MIGHT_MATCH` , left & right should be `BLOCK_MIGHT_MATCH`.
Because in the next filter may right may be `BLOCK_CANNOT_MATCH ` and we should drop it

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
yabola: done

wgtmac:
cc @zhongyujiang 
Not sure if you are interested in reviewing this PR.

yabola:
@gszadovszky @shangxinli If you have time, please also take a look, thanks~

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
yabola: @gszadovszky Thank you for your reply!
In the original implementation, `BLOCK_MUST_MATCH` can be judged using minMax. 
So if we follow the specification, can we using minMax as the data result of an enlarged range. So we can accurately judge that when the data is not in this range.

yabola: In fact, what I wanted to do at the beginning was to avoid the use of `BloomFilter` through `minMax` and `dictionary`(if column have) as much as possible, because the former is more accurate and `BloomFilter` may cost time and memory.

gszadovszky: If we allow min/max values to be lower/upper bounds and not part of the real data (as @wgtmac suggested) then there are only two scenarios we can be sure the requested value is in the row group:
* Searching for `null` and the number of nulls (is specified and) is greater than `0`
* The min value and the max value are both equal to the value we are searching for
If we expect min/max values to be part of the row groups then we can extend the related statements but it might not be a safe choice.

Dictionary filter is much more straightforward. Bloom filter should be used only if the requested value is not in the dictionary (or there is no dictionary at all) and there are pages that are not dictionary encoded. So your change clearly makes sense to differenciate the three potential results and move forward to bloom filters only in case of `BLOCK_MIGHT_MATCH`.

yabola: @gszadovszky Thank you very much for your advice.
But this will reduce some usage scenarios. From the code, `minMax` and `dictionary` is basically the minMax of real data. But the specification doesn't say.
Can we add a config (default false, don't use `BLOCK_MUST_MATCH`) to optimize it?

zhongyujiang: >  From the code, minMax and dictionary is basically the minMax of real data. But the specification doesn't say.

IIUC, there may be other implementation libraries where min-max is not part of the row-group data set, because the specification does not define it clearly. Although I don't know of any library that does this.

> Can we add a config (default false, don't use BLOCK_MUST_MATCH) to optimize it?

From my point of view, it is doable to introduce such a config, but I am a little concerned that it will be difficult for users to understand.

wgtmac: I'm not aware of any implementation use lower/upper bounds instead of min/max values. But the specs does allow it to happen and `ColumnIndex` has stated about it [(link)](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L956) explicitly as mentioned by @gszadovszky 

I prefer to be conservative and make it easy to use:
- Regard statistics as lower/upper bounds unless `min == max`.
- Do not add any config for an aggressive optimization.

gszadovszky: FYI: for column index we specified it explicitly to allow shortening the min/max values (e.g. string are truncated to 64bytes by default). 
The older min/max statistics for the row groups are trickier since it is not properly specified. The issue is if only one implementation of parquet in the wild would do the similar truncate we may loose data when filtering if we use this "aggressive optimization". 
Dictionary is a completely different thing. It contains the very values from the pages themselves. It is not designed to be used for filtering (but we use anyway) but to save space.

yabola: > Dictionary contains the very values from the pages themselves
@gszadovszky do you think it is suitable to use `BLOCK_MUST_MATCH` when all page are encoded in `DictionaryFilter`?
For example,
1. when search for data = x1 and dictionary has one element = x1  , then return `BLOCK_MUST_MATCH`
2. when search for data > x1 and every elements in dictionary > x1  , then return `BLOCK_MUST_MATCH`

gszadovszky: @yabola, I think you misunderstand how dictionary/filtering works. The dictionary contains all of the values which the dictionary encoded pages may contain. These pages actually not contain the values but the indices referencing to the related values in the dictionary.

So, if a searched element can be found in the dictionary you may return `BLOCK_MUST_MATCH` even if only one page is dictionary encoded. For example if the filter is `x > 1` then any element in the dictionary `> 1` would fulfill the filter so the `BLOCK_MUST_MATCH`.
If the dictionary does not contain any of the searched elements (for the previous example every elements are `<= 1`) then you may return `BLOCK_CANNOT_MATCH` only if all the related pages are dictionary encoded. Otherwise you return `BLOCK_MIGHT_MATCH` since you don't know anything about the not dictionary encoded pages based on the dictionary.

wgtmac: > @yabola, I think you misunderstand how dictionary/filtering works. The dictionary contains all of the values which the dictionary encoded pages may contain. These pages actually not contain the values but the indices referencing to the related values in the dictionary.
> 
> So, if a searched element can be found in the dictionary you may return `BLOCK_MUST_MATCH` even if only one page is dictionary encoded. For example if the filter is `x > 1` then any element in the dictionary `> 1` would fulfill the filter so the `BLOCK_MUST_MATCH`. If the dictionary does not contain any of the searched elements (for the previous example every elements are `<= 1`) then you may return `BLOCK_CANNOT_MATCH` only if all the related pages are dictionary encoded. Otherwise you return `BLOCK_MIGHT_MATCH` since you don't know anything about the not dictionary encoded pages based on the dictionary.

CMIW, the `DictionaryFilter` is only enabled when all data pages are dictionary-encoded in a certain column chunk. So @yabola 's statement seems work to me.

gszadovszky: @wgtmac, do you refering e.g. [this](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java#L139-L143)? But this comment is only true if we are about to return `BLOCK_MIGHT_MATCH` or `BLOCK_CANNOT_MATCH`. If we want to potentially return `BLOCK_MUST_MATCH` we need to read the dictionary and check the values. So we might spend more time on dictionary filtering to potentially skip using bloom filters. But what if we don't have bloom filters? (We should not write bloom filters if all the pages are dictionary encoded anyway.)

wgtmac: @gszadovszky Maybe I didn't fully understand your point. Are you concerned about the performance or correctness? The `DictionaryFilter` is already decoding the dictionary to evaluate predicates like `lt` to return `BLOCK_CANNOT_MATCH`. This patch does not introduce more overhead but supports returning `BLOCK_MUST_MATCH` at the same time.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/RowGroupFilter.java
```diff
@@ -97,21 +100,32 @@ public List<BlockMetaData> visit(FilterCompat.FilterPredicateCompat filterPredic
...
 
     for (BlockMetaData block : blocks) {
-      boolean drop = false;
-
-      if(levels.contains(FilterLevel.STATISTICS)) {
-        drop = StatisticsFilter.canDrop(filterPredicate, block.getColumns());
+      Boolean predicate = BLOCK_MIGHT_MATCH;
+      if (levels.contains(FilterLevel.STATISTICS)) {
+        predicate = StatisticsFilter.predicate(filterPredicate, block.getColumns());
+        if(isExactPredicate(predicate)) {
```
wgtmac, line 106: nit: name of `isExactPredicate` is a little bit unclear.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/compat/PredicateEvaluation.java
wgtmac: I didn't get your point here. @yabola 

If the current expression is `A and B`, then following result applies regardless of other expressions:
- A is BLOCK_MUST_MATCH and B is BLOCK_MUST_MATCH => BLOCK_MUST_MATCH
- A is BLOCK_MUST_MATCH and B is BLOCK_MIGHT_MATCH => BLOCK_MUST_MATCH
- A is BLOCK_MIGHT_MATCH and B is BLOCK_MUST_MATCH => BLOCK_MUST_MATCH
- A is BLOCK_MIGHT_MATCH and B is BLOCK_MIGHT_MATCH => BLOCK_MIGHT_MATCH
- A is BLOCK_CANNOT_MATCH or/and B is BLOCK_CANNOT_MATCH => BLOCK_CANNOT_MATCH

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
yabola: > We should not write bloom filters if all the pages are dictionary encoded anyway

@gszadovszky CMIW, in the existing implementation, `BloomFilter` will still generate even if all pages are encoded have dictionaries.

wgtmac: I think @gszadovszky is correct.

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L884

yabola: @wgtmac @gszadovszky I think it might be bug here because we use `PLAIN_DICTIONARY` in parquet v1? 
Bloomfilter will avoid generating only when the dictionary type is all `RLE_DICTIONARY`?
 https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L884

wgtmac: Good catch! I am not familiar with the old story. Does format v1 support bloom filter?

yabola: > Good catch! I am not familiar with the old story. Does format v1 support bloom filter?

Yes, Spark use parquet v1 by default

gszadovszky: Godd catch indeed, @yabola! Could you open a separate jira and maybe a PR for this finding?

@wgtmac, performance. Let's see the following scenario. We have dictionary encoding but not for all the pages. We also have Bloom filter. Does it worth reading the dictionary to check if a value is in there knowing if it doesn't we still want to check the Bloom filter? I don't know the answer, maybe yes. But if it is a no, then the whole concept of this PR is questionable.
For the case of all the pages are dictionary encoded we should not have Bloom filters therefore it doesn't really matter if we return `BLOCK_MIGHT_MATCH` or `BLOCK_MUST_MATCH` in case we find the interested values in the dictionary.
Since we might already written some Bloom filters for fully dictionary encoded column chunks we should handle this scenario. But we can do it easily buy skipping reading Bloom filters in this case completely.

wgtmac: > We have dictionary encoding but not for all the pages. We also have Bloom filter.

Yes, that's true.

> Does it worth reading the dictionary to check if a value is in there knowing if it doesn't we still want to check the Bloom filter?

In this case, the dictionary will not be read via `expandDictionary(meta)` by `DictionaryFilter` if `hasNonDictionaryPages(meta)` returns true and will not make performance worse. e.g. https://github.com/apache/parquet-mr/blob/261f7d2679407c833545b56f4c85a4ae8b5c9ed4/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java#L388

> For the case of all the pages are dictionary encoded we should not have Bloom filters therefore it doesn't really matter if we return BLOCK_MIGHT_MATCH or BLOCK_MUST_MATCH in case we find the interested values in the dictionary.

It is difficult to make the trade-off here. If we only have one predicate, then the dictionary will read any way, either by the DictionaryFilter or by reading the data later if the row group cannot be dropped. However, if we have other predicates that can drop the row group, then reading the dictionary here by DictionaryFilter is worthless.

yabola: @gszadovszky @wgtmac @zhongyujiang Thank you very much for working on it.  I have some thoughts.

> We can improve (?) the case when not all the pages are dictionary encoded

1. I can't make sure if it is suitable to load dictionary even if pages are not all decoded. (I may choose not to change this behavior)

2. However considering the origin `BloomFilter` bug in parquet v1, we might have to do something to avoid using `BloomFilter`(even if pages are all encoded).  
In the code implementation we may have to use some flag to mark if dictionary `DictionaryFilter#expandDictionary` successfully (method will throw `IOException` and we can't `expandDictionary` again in `BloomFilterImpl`).
Or we could also use `BLOCK_MUST_MATCH` like this PR.

> StatisticsFilter: Because of the lower/upper bound issue we cannot really improve this (except for the specific case when min=max)

If we only use it when min=max, I think it might not worth doing it.

yabola:
@wgtmac @gszadovszky 
I have a proposal to  automatically build BloomFilter with a more precise size. I create a jira https://issues.apache.org/jira/browse/PARQUET-2254 and  I hope to get your opinions, thank you.

> Now the usage is to specify the size, and then build BloomFilter. In general scenarios, it is actually not sure how much the distinct value is.
If BloomFilter can be automatically generated according to the data, the file size can be reduced and the reading efficiency can also be improved.

I have an idea that the user can specify a maximum BloomFilter filter size, then we build several BloomFilter at the same time, we can use the largest BloomFilter as a counting tool( If there is no hit when inserting a value, the counter will be +1, of course this may be imprecise but enough)
Then at the end of the write, choose a BloomFilter of a more appropriate size when the file is finally written.

I want to implement this feature and

gszadovszky:
Thanks @yabola for coming up with this idea. Let's continue the discussion about the BloomFilter building idea in the jira.

Meanwhile, I've been thinking about the actual problem as well. Currently, for row group filtering we are checking the min/max values first which is correct since this is the most fast thing to do. Then the dictionary and then the bloom filter. The ordering of the latter two is not obvious to me in every scenarios. At the time of filtering we did not start reading the actual row group so there is no advantage in I/O to read the dictionary first. Furthermore, searching something in the bloom filter is much faster than in the dictionary. And the size of the bloom filter is probably less than the size of the dictionary. Though, it would require some measurements to prove if it is a good idea to get the bloom filter before the dictionary. What do you think?

wgtmac:
> Thanks @yabola for coming up with this idea. Let's continue the discussion about the BloomFilter building idea in the jira.
> 
> Meanwhile, I've been thinking about the actual problem as well. Currently, for row group filtering we are checking the min/max values first which is correct since this is the most fast thing to do. Then the dictionary and then the bloom filter. The ordering of the latter two is not obvious to me in every scenarios. At the time of filtering we did not start reading the actual row group so there is no advantage in I/O to read the dictionary first. Furthermore, searching something in the bloom filter is much faster than in the dictionary. And the size of the bloom filter is probably less than the size of the dictionary. Though, it would require some measurements to prove if it is a good idea to get the bloom filter before the dictionary. What do you think?

What I did in production is to issue async I/Os of dictionaries (if all data pages are dictionary-encoded in that column chunk and the dictionary is not big) and bloom filters of selected row groups in advance. The reason is to eliminate blocking I/O when pushing down the predicates. However, the parquet specs only records the offset to bloom filter. So I also added the length of each bloom filter in the key value metadata section (probably a good reason to add to the specs as well?)

yabola:
> Thanks @yabola for coming up with this idea. Let's continue the discussion about the BloomFilter building idea in the jira.
> 
> Meanwhile, I've been thinking about the actual problem as well. Currently, for row group filtering we are checking the min/max values first which is correct since this is the most fast thing to do. Then the dictionary and then the bloom filter. The ordering of the latter two is not obvious to me in every scenarios. At the time of filtering we did not start reading the actual row group so there is no advantage in I/O to read the dictionary first. Furthermore, searching something in the bloom filter is much faster than in the dictionary. And the size of the bloom filter is probably less than the size of the dictionary. Though, it would require some measurements to prove if it is a good idea to get the bloom filter before the dictionary. What do you think?

It is a good idea to adjust filter order and prefer the use of lighter filters first to judge.
But I have some concern (not sure if it is correct): 
In parquet dictionary will generate only in low-base data( see `parquet.dictionary.page.size` 1 MB), and BloomFilter is usually used in high base columns(?) . So ideally only one of these two will be used(?)

And ideally we should only use one of these two (don't judge both of them). If there is a BloomFilter and filter is `=` or `in`, only use the BloomFilter , otherwise use the dictionary.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -52,14 +53,24 @@ public class DictionaryFilter implements FilterPredicate.Visitor<Boolean> {
   private static final boolean BLOCK_MIGHT_MATCH = false;
   private static final boolean BLOCK_CANNOT_MATCH = true;
 
-  public static boolean canDrop(FilterPredicate pred, List<ColumnChunkMetaData> columns, DictionaryPageReadStore dictionaries) {
+  public static boolean canDrop(FilterPredicate pred, List<ColumnChunkMetaData> columns,
+    DictionaryPageReadStore dictionaries, AtomicBoolean canExactlyDetermine) {
```
shangxinli: This is a public method. We are changing the signature now. Can you add another one to support callers that are still calling the old one?

PR #1025, opened 2023-02-11
PARQUET-2241: Fix ByteStreamSplitValuesReader with nulls

wgtmac:
ByteStreamSplitValuesReader depends on page.num_values which includes null values to compute the total stream length. Then it throws if it fails to read enough bytes from the page buffer. This certainly happens if the page contains null values.

### Jira

[PARQUET-2241](https://issues.apache.org/jira/browse/PARQUET-2241)

### Tests

Add test `org.apache.parquet.avro.TestByteStreamSplitE2E` to write and read floating values with BYTE_STREAM_SPLIT encoding.

### Commits

`ByteStreamSplitValuesReader` strictly depends on remaining stream length to get actual number of encoded values before decoding.


wgtmac:
Could you please take a look at this? Thanks in advance!

@shangxinli @ggershinsky @gszadovszky

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -58,15 +58,19 @@ protected void gatherElementDataFromStreams(byte[] gatheredData)
   public void initFromPage(int valuesCount, ByteBufferInputStream stream) throws ParquetDecodingException, IOException {
```
shangxinli, line 58: It seems the input value valueCount is not used anymore

wgtmac: Added a check to use `valuesCount` as upper bound.

wgtmac:
@gszadovszky Have time to take a look? The fix is pretty trivial.

gszadovszky (approved):
I'll push it tomorrow. @shangxinli, gentle ping if you want to check it again.

shangxinli (approved):

PR #1026, opened 2023-02-13
PARQUET-2228: ParquetRewriter supports more than one input file

wgtmac:
### Jira

https://issues.apache.org/jira/browse/PARQUET-2228

### Tests

- Refactor and add various cases to `ParquetRewriterTest` for merging files.

### Commits

- RewriteOptions supports merging more than one input file.
- Enforce schema equality of all input files.
- ParquetRewriter supports various rewrite operations on several input files.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
@@ -45,7 +45,6 @@ import org.apache.parquet.format.PageHeader;
 import org.apache.parquet.format.converter.ParquetMetadataConverter;
 import org.apache.parquet.hadoop.CodecFactory;
 import org.apache.parquet.hadoop.ColumnChunkPageWriteStore;
-import org.apache.parquet.hadoop.ParquetFileReader;
 import org.apache.parquet.hadoop.ParquetFileWriter;
 import org.apache.parquet.hadoop.metadata.BlockMetaData;
 import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
@@ -63,6 +62,7 @@ import org.apache.parquet.io.api.Converter;
 import org.apache.parquet.io.api.GroupConverter;
 import org.apache.parquet.io.api.PrimitiveConverter;
 import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.InvalidSchemaException;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.PrimitiveType;
 import org.apache.parquet.schema.Type;
@@ -74,15 +74,16 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.apache.parquet.column.ParquetProperties.DEFAULT_COLUMN_INDEX_TRUNCATE_LENGTH;
 import static org.apache.parquet.column.ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH;
 import static org.apache.parquet.crypto.ModuleCipherFactory.ModuleType;
-import static org.apache.parquet.format.converter.ParquetMetadataConverter.NO_FILTER;
 import static org.apache.parquet.hadoop.ParquetWriter.DEFAULT_BLOCK_SIZE;
 import static org.apache.parquet.hadoop.ParquetWriter.MAX_PADDING_SIZE_DEFAULT;
 
@@ -93,33 +94,37 @@ public class ParquetRewriter implements Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(ParquetRewriter.class);
   private final int pageBufferSize = ParquetProperties.DEFAULT_PAGE_SIZE * 2;
   private final byte[] pageBuffer = new byte[pageBufferSize];
-  private TransParquetFileReader reader;
-  private ParquetFileWriter writer;
-  private ParquetMetadata meta;
-  private MessageType schema;
-  private String originalCreatedBy;
+  // Configurations for the new file
   private CompressionCodecName newCodecName = null;
   private List<String> pruneColumns = null;
   private Map<ColumnPath, MaskMode> maskColumns = null;
   private Set<ColumnPath> encryptColumns = null;
   private boolean encryptMode = false;
   private Map<String, String> extraMetaData = new HashMap<>();
+  // Writer to rewrite the input files
+  private ParquetFileWriter writer;
+  // Number of blocks written which is used to keep track of the actual row group ordinal
+  private int numBlocksRewritten = 0;
+  // Reader and relevant states of the in-processing input file
+  private Queue<TransParquetFileReader> inputFiles = new LinkedList<>();
+  // Schema of input files (should be the same) and to write to the output file
+  private MessageType schema = null;
+  // The reader for the current input file
+  private TransParquetFileReader reader = null;
+  // The metadata of current reader being processed
+  private ParquetMetadata meta = null;
+  // created_by information of current reader being processed
+  private String originalCreatedBy = "";
+  // Unique created_by information from all input files
+  private Set<String> allOriginalCreatedBys = new HashSet<>();
 
   public ParquetRewriter(RewriteOptions options) throws IOException {
-    Preconditions.checkArgument(options.getInputFiles().size() == 1, "Only support one input file");
-    Path inPath = options.getInputFiles().get(0);
-    Path outPath = options.getOutputFile();
     Configuration conf = options.getConf();
+    openInputFiles(options.getInputFiles(), conf);
+    initNextReader();
 
     newCodecName = options.getNewCodecName();
     pruneColumns = options.getPruneColumns();
-
-    // Get file metadata and full schema from the input file
-    meta = ParquetFileReader.readFooter(conf, inPath, NO_FILTER);
-    schema = meta.getFileMetaData().getSchema();
-    originalCreatedBy = meta.getFileMetaData().getCreatedBy();
-    extraMetaData.putAll(meta.getFileMetaData().getKeyValueMetaData());
-    extraMetaData.put(ORIGINAL_CREATED_BY_KEY, originalCreatedBy);
 
     // Prune columns if specified
     if (pruneColumns != null && !pruneColumns.isEmpty()) {
@@ -127,7 +132,7 @@ public class ParquetRewriter implements Closeable {
       getPaths(schema, paths, null);
       for (String col : pruneColumns) {
         if (!paths.contains(col)) {
-          LOG.warn("Input column name {} doesn't show up in the schema of file {}", col, inPath.getName());
+          LOG.warn("Input column name {} doesn't show up in the schema of file {}", col, reader.getFile());
         }
       }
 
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -183,12 +186,61 @@ public ParquetRewriter(TransParquetFileReader reader,
...
+          }
+        }
+        this.allOriginalCreatedBys.add(reader.getFooter().getFileMetaData().getCreatedBy());
+        this.inputFiles.add(reader);
+      } catch (IOException e) {
+        throw new IllegalArgumentException("Failed to open input file: " + inputFile, e);
+      }
+    }
+
+    extraMetaData.put(ORIGINAL_CREATED_BY_KEY, String.join("\n", allOriginalCreatedBys));
```
wgtmac, line 213: We have discussed the issue to consolidate original created_by when merging several input files. Now I use a HashSet to make sure there is no duplicate and they are concatenated by `\n`. Does this sound good? @gszadovszky

wgtmac:
@ggershinsky @shangxinli PTAL, thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
shangxinli: Do we do dedup?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -183,12 +186,61 @@ public ParquetRewriter(TransParquetFileReader reader,
     }
   }
 
+  // Open all input files to validate their schemas are compatible to merge
+  private void openInputFiles(List<Path> inputFiles, Configuration conf) {
+    Preconditions.checkArgument(inputFiles != null && !inputFiles.isEmpty(), "No input files");
+
+    for (Path inputFile : inputFiles) {
```
shangxinli, line 193: Do we want to set a max size?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: > Do we do dedup?

Yes, `allOriginalCreatedBys` is a HashSet which does the job.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: It is tricky to set a max size, whether in terms of number of files or total file sizes. Based on my knowledge, the performance varies when input files have different number of rows, number of columns, number of blocks, block sizes, etc. So it would be better for the caller to do the planning.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -183,12 +186,61 @@ public ParquetRewriter(TransParquetFileReader reader,
...
+      try {
+        TransParquetFileReader reader = new TransParquetFileReader(
+                HadoopInputFile.fromPath(inputFile, conf), HadoopReadOptions.builder(conf).build());
+        MessageType inputFileSchema = reader.getFooter().getFileMetaData().getSchema();
+        if (this.schema == null) {
+          this.schema = inputFileSchema;
+        } else {
+          // Now we enforce equality of schemas from input files for simplicity.
+          if (!this.schema.equals(inputFileSchema)) {
+            throw new InvalidSchemaException("Input files have different schemas");
```
ggershinsky, line 203: maybe some info (eg file path)?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -183,12 +186,61 @@ public ParquetRewriter(TransParquetFileReader reader,
...
+      } catch (IOException e) {
+        throw new IllegalArgumentException("Failed to open input file: " + inputFile, e);
+      }
+    }
+
+    extraMetaData.put(ORIGINAL_CREATED_BY_KEY, String.join("\n", allOriginalCreatedBys));
+  }
+
+  // Routines to get reader of next input file.
+  // Returns true if there is a next file to read, false otherwise.
```
ggershinsky, line 217: this function returns void

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -484,15 +673,22 @@ private List<Long> getOffsets(TransParquetFileReader reader, ColumnChunkMetaData
...
-    assertEquals(inFMD.getCreatedBy(), outFMD.getCreatedBy());
-    assertNull(inFMD.getKeyValueMetaData().get(ParquetRewriter.ORIGINAL_CREATED_BY_KEY));
+    FileMetaData outFMD = getFileMetaData(outputFile, null).getFileMetaData();
+    assertEquals(inputCreatedBy, outFMD.getCreatedBy());
 
     String originalCreatedBy = outFMD.getKeyValueMetaData().get(ParquetRewriter.ORIGINAL_CREATED_BY_KEY);
     assertNotNull(originalCreatedBy);
-    assertEquals(inFMD.getCreatedBy(), originalCreatedBy);
+    assertEquals(inputCreatedBy, originalCreatedBy);
   }
```
ggershinsky, line 693: nit: empty line

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Fixed

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
wgtmac: Fixed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
gszadovszky: Sounds good to me. Thanks, @wgtmac.

gszadovszky:
@wgtmac, by supporting multiple files to rewrite them into one we will end up with the same number of row-groups, right? Therefore, this tool is not ment to be used to solve the "small files problem". I am highlighting this because we've had issues with users misunderstanding the purpose of features like this. Maybe, we should add some notes about it to the help of parquet-cli.

wgtmac:
> @wgtmac, by supporting multiple files to rewrite them into one we will end up with the same number of row-groups, right? Therefore, this tool is not ment to be used to solve the "small files problem". I am highlighting this because we've had issues with users misunderstanding the purpose of features like this. Maybe, we should add some notes about it to the help of parquet-cli.

You're right. We might add an option to force rewriting the input files record by record so row groups are regenerated by the writer. Does that sound good?

wgtmac:
> > @wgtmac, by supporting multiple files to rewrite them into one we will end up with the same number of row-groups, right? Therefore, this tool is not ment to be used to solve the "small files problem". I am highlighting this because we've had issues with users misunderstanding the purpose of features like this. Maybe, we should add some notes about it to the help of parquet-cli.
> 
> You're right. We might add an option to force rewriting the input files record by record so row groups are regenerated by the writer. Does that sound good? @gszadovszky

Something like https://github.com/apache/parquet-mr/blob/master/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ConvertCommand.java

gszadovszky:
> You're right. We might add an option to force rewriting the input files record by record so row groups are regenerated by the writer. Does that sound good? @gszadovszky

It sounds great, @wgtmac, but I am fine implementing it separately or in this PR as you prefer. However, we still need to highlight somehow to the user that in the other cases the user should not expect performance improvements in case of merging several files into one. (Moreover it'll increase the footer size which might also generate additional issues.) What do you think?

wgtmac:
> > You're right. We might add an option to force rewriting the input files record by record so row groups are regenerated by the writer. Does that sound good? @gszadovszky
> 
> It sounds great, @wgtmac, but I am fine implementing it separately or in this PR as you prefer. However, we still need to highlight somehow to the user that in the other cases the user should not expect performance improvements in case of merging several files into one. (Moreover it'll increase the footer size which might also generate additional issues.) What do you think?

I agree. Let me add some comments to explain the issue at the moment.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -101,37 +103,121 @@ public static class Builder {
...
       this.outputFile = outputFile;
     }
 
+    /**
+     * Create a builder to create a RewriterOptions.
+     * <p>
+     * Please note that if merging more than one file, the schema of all files must be the same.
+     * Otherwise, the rewrite will fail.
+     * <p>
+     * The rewrite will keep original row groups from all input files. This may not be optimal
```
wgtmac, line 125: I have added some comment to elaborate the small row group problem. Please check @gszadovszky

wgtmac:
I saw a test failure below from the [GHA](https://github.com/apache/parquet-mr/actions/runs/4195487917/jobs/7275103509) which is unstable:
```
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.637 s <<< FAILURE! - in org.apache.parquet.hadoop.TestParquetWriter
Error:  testParquetFileWithBloomFilterWithFpp(org.apache.parquet.hadoop.TestParquetWriter)  Time elapsed: 0.368 s  <<< FAILURE!
java.lang.AssertionError
	at org.apache.parquet.hadoop.TestParquetWriter.testParquetFileWithBloomFilterWithFpp(TestParquetWriter.java:342)
```

shangxinli (approved):

shangxinli:
Let me know if you have more feedbacks  @ggershinsky @gszadovszky

ggershinsky (approved):
LGTM, thanks

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
gszadovszky: Thank you, @wgtmac

gszadovszky (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -183,12 +189,69 @@ public ParquetRewriter(TransParquetFileReader reader,
...
+          if (!this.schema.equals(inputFileSchema)) {
+            LOG.error("Input files have different schemas, expect: {}, input: {}, current file: {}",
+                    this.schema, inputFileSchema, inputFile);
+            throw new InvalidSchemaException("Input files have different schemas, current file: " + inputFile);
+          }
+        }
+        this.allOriginalCreatedBys.add(reader.getFooter().getFileMetaData().getCreatedBy());
+        this.inputFiles.add(reader);
+      } catch (IOException e) {
+        throw new IllegalArgumentException("Failed to open input file: " + inputFile, e);
```
vectorijk, line 214: shall we handle the close of connection opened by `TransParquetFileReader` when the exception is thrown here?
in `initNextReader` shall we also consider to close previous reader explicitly if it is not null

wgtmac: That sounds reasonable. Would you like to create a PR with your proposed change? @vectorijk

vectorijk: sounds good

PR #1028, opened 2023-02-15
PARQUET-2244: Fix notIn for columns with null values

zhongyujiang:
Jira: [PARQUET-2244](https://issues.apache.org/jira/browse/PARQUET-2244)
This fixes a bug similar to [PARQUET-1510](https://issues.apache.org/jira/browse/PARQUET-1510).

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2244

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
@@ -447,6 +447,14 @@ public class DictionaryFilter implements FilterPredicate.Visitor<Boolean> {
       return BLOCK_MIGHT_MATCH;
     }
 
+    boolean mayContainNull = (meta.getStatistics() == null
+      || !meta.getStatistics().isNumNullsSet()
+      || meta.getStatistics().getNumNulls() > 0);
+    // The column may contain nulls and the values set contains no null, so the row group cannot be eliminated.
+    if (mayContainNull) {
+      return BLOCK_MIGHT_MATCH;
+    }
+
     // if the chunk has non-dictionary pages, don't bother decoding the
     // dictionary because the row group can't be eliminated.
     if (hasNonDictionaryPages(meta)) {
--- a/parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
@@ -35,6 +35,7 @@ import org.apache.parquet.example.data.Group;
 import org.apache.parquet.example.data.simple.SimpleGroupFactory;
 import org.apache.parquet.filter2.predicate.FilterPredicate;
 import org.apache.parquet.filter2.predicate.LogicalInverseRewriter;
+import org.apache.parquet.filter2.predicate.Operators;
 import org.apache.parquet.filter2.predicate.Operators.BinaryColumn;
 import org.apache.parquet.filter2.predicate.Operators.DoubleColumn;
 import org.apache.parquet.filter2.predicate.Operators.FloatColumn;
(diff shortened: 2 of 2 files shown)


zhongyujiang:
@huaxingao @gszadovszky Can you help review this? Thanks!

gszadovszky (approved):

gszadovszky:
@shangxinli, it might require a backport and releases on the branches `In` and `NotIn` were released.

zhongyujiang:
@gszadovszky Thanks for reviewing and the quick merge!

huaxingao:
I did a quick test using Spark

```
          Seq("A", "A", null).toDF("column").repartition(1).write.mode("overwrite").parquet("t")
          spark.read.parquet("t").where("NOT (column <=> 'A')").show    // this returns null
          spark.read.parquet("t").where("NOT (column = 'A')").show    // this returns empty
          spark.read.parquet("t").where("column IN ('A')").show    // this returns A A
          spark.read.parquet("t").where("column NOT IN ('A')").show    // this returns empty
```

If we only has `A` and `null` for `column` and we have predicate `column not in ('A')`, should we return empty instead of null?

wgtmac:
> I did a quick test using Spark
> 
> ```
>           Seq("A", "A", null).toDF("column").repartition(1).write.mode("overwrite").parquet("t")
>           spark.read.parquet("t").where("NOT (column <=> 'A')").show    // this returns null
>           spark.read.parquet("t").where("NOT (column = 'A')").show    // this returns empty
>           spark.read.parquet("t").where("column IN ('A')").show    // this returns A A
>           spark.read.parquet("t").where("column NOT IN ('A')").show    // this returns empty
> ```
> 
> If we only has `A` and `null` for `column` and we have predicate `column not in ('A')`, should we return empty instead of null?

IIUC, 
- `col IN (A, B)` is equal to `col = A OR col = B`
- `col NOT IN (A, B)` is equal to `col <> A AND col <> B`. where `col <> A` means `col IS NOT NULL and col != A`

So my answer to your question above is empty. @huaxingao 

It seems that we lose the chance to skip the row group with this fix. @gszadovszky

zhongyujiang:
I haven't encountered any troubles caused by this situation in practice. I found this while looking at the code, when evaluating  `notIn`, dictionary filter returns `BLOCK_MIGHT_MATCH` when the column isn't in the file which means all values are null(see L450-L453), but it does not consider whether there will be a null value when the column really exists. I think it's inconsistent so opened this fix.

zhongyujiang:
Didn't think about comparisons with non-null values before submitting this PR.  I don't know if there is a downstream that relies on Parquet judge `value <> null` as TRUE instead of UNKNOW, I guess that might be in some non-ansi standard engines. If there is no such situation, I think this PR can be reverted safely.

gszadovszky:
It seems I pushed it too quickly. Sorry for not giving the time to give feedback, @huaxingao and @wgtmac.
@zhongyujiang, feel free to put up another PR with the revert.

zhongyujiang:
>  I don't know if there is a downstream that relies on Parquet judge value <> null as TRUE instead of UNKNOW, I guess that might be in some non-ansi standard engines.

I just want to make sure it's safe to revert this, I did some search and find this: https://stackoverflow.com/questions/129077/null-values-inside-not-in-clause
Seems SQL Server 2005 has a switch `ansi_null` to control wheather the result of  `value <> null` is `TRUE` or `UNKNOWN`,  when `ansi_nulls` is off, `3 <> null` is true. Though I didn't find such a switch in engines like Hive or Spark. Do you have any comments on this? Or do you think it's worth caring about?  @gszadovszky @wgtmac @huaxingao.

huaxingao:
I don't have a strong opinion on whether to keep or revert the fix. The fix won't cause any correctness issue on the engine side because engine will filter again.

wgtmac:
> I don't have a strong opinion on whether to keep or revert the fix. The fix won't cause any correctness issue on the engine side because engine will filter again.

Same here.

This is not a correctness issue. Rather we should be careful with NULL behavior since different engines may have different assumptions. I'd say we might lose some optimization with this fix but it is much safer now.

zhongyujiang:
Then I will not open a revert PR, thanks everyone for commenting on this!

PR #1029, opened 2023-02-15
PARQUET-2245: Improve dictionary filter evaluating notEq

zhongyujiang:
JIRA: [PARQUET-2245](https://issues.apache.org/jira/browse/PARQUET-2245)

This is a minor improvement for evaluating `notEq`. When evaluating `notEq`, if the column may contain nulls and the `notEq` value is non-null, the row-group must not be skipped. In such scenario reading dictionary and compare values is not necessary.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
@@ -179,6 +179,15 @@ public class DictionaryFilter implements FilterPredicate.Visitor<Boolean> {
       return BLOCK_MIGHT_MATCH;
     }
 
+    boolean mayContainNull = (meta.getStatistics() == null
+      || !meta.getStatistics().isNumNullsSet()
+      || meta.getStatistics().getNumNulls() > 0);
+    // The column may contain nulls and the value is non-null, don't bother decoding the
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -187,10 +196,7 @@ public <T extends Comparable<T>> Boolean visit(NotEq<T> notEq) {
 
     try {
       Set<T> dictSet = expandDictionary(meta);
-      boolean mayContainNull = (meta.getStatistics() == null
-          || !meta.getStatistics().isNumNullsSet()
-          || meta.getStatistics().getNumNulls() > 0);
-      if (dictSet != null && dictSet.size() == 1 && dictSet.contains(value) && !mayContainNull) {
+      if (dictSet != null && dictSet.size() == 1 && dictSet.contains(value)) {
```
wgtmac, line 199: I just noticed that the `FilterPredicate` does not provide an entry for `IS NULL` or `IS NOT NULL`. This confuses me because `col IS NOT NULL` is not equal to `col != NULL`.

CMIW, `col NOT EQ A` has two meanings as below:
- If A is NULL, it should return an empty list. Because NULL cannot be compared to any value including another NULL.
- Otherwise, it should return a list of values excluding A and NULL.

cc @huaxingao @gszadovszky @shangxinli

PR #1031, opened 2023-02-16
PARQUET-2247: Fail-fast if CapacityByteArrayOutputStream write overflow

cxzl25:
### Jira

My PR addresses the following [PARQUET-2247](https://issues.apache.org/jira/browse/PARQUET-2247) 

### Tests

Before this PR, the size function returns a negative number.  
Need a relatively large heap ( -Xmx5g) to run this test.
```java
  @Test
  public void testWriteBigArray() throws Throwable {
    CapacityByteArrayOutputStream capacityByteArrayOutputStream = newCapacityBAOS(10);

    byte[] val = new byte[1024];
    capacityByteArrayOutputStream.write(val);
    byte[] bigVal = new byte[Integer.MAX_VALUE - 16];
    capacityByteArrayOutputStream.write(bigVal);

    assertEquals(-2147482641, capacityByteArrayOutputStream.size());
  }
```

###  After this PR throw exception

```java
org.apache.parquet.ParquetByteArrayOutputOverflowException: Cannot write byte data larger than Integer.MAX_VALUE bytes: -2147482641
	at org.apache.parquet.bytes.CapacityByteArrayOutputStream.write(CapacityByteArrayOutputStream.java:225)
	at java.base/java.io.OutputStream.write(OutputStream.java:122)
	at org.apache.parquet.bytes.TestCapacityByteArrayOutputStream.testWriteBigArray(TestCapacityByteArrayOutputStream.java:231)
```

### Commits

When `CapacityByteArrayOutputStream#size` and `total_uncompressed_size` are negative, 
it fails quickly to avoid writing success but reading failure.

PR #1032, opened 2023-02-20
PARQUET-2164: Check size of buffered data to prevent page data from overflowing

parthchandra:
This PR addresses the following [PARQUET-2164](https://issues.apache.org/jira/browse/PARQUET-2164) 
The configuration parameters 
```
    parquet.page.size.check.estimate=false
    parquet.page.size.row.check.min=<check_size>
    parquet.page.size.row.check.max=<check_size>
```
address the reported problem. However the issue can still be hit because the default value of `parquet.page.size.check.estimate` is `true`.
This PR simply adds a check to make sure that `CapacityByteArrayOutputStream` cannot overflow but will instead throw an exception.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
@@ -164,6 +164,12 @@ public class CapacityByteArrayOutputStream extends OutputStream {
   private void addSlab(int minimumSize) {
     int nextSlabSize;
 
+    if (bytesUsed + minimumSize < 0) {
+      // This is interpreted as a request for a value greater than Integer.MAX_VALUE
+      // We throw OOM because that is what java.io.ByteArrayOutputStream also does
+      throw new OutOfMemoryError("Size of data exceeded 2GB");
+    }
+
     if (bytesUsed == 0) {
       nextSlabSize = initialSlabSize;
     } else if (bytesUsed > maxCapacityHint / 5) {
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
@@ -164,10 +164,13 @@ public class CapacityByteArrayOutputStream extends OutputStream {
   private void addSlab(int minimumSize) {
     int nextSlabSize;
 
-    if (bytesUsed + minimumSize < 0) {
+    // check for overflow 
+    try {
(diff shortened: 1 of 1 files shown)


cxzl25:
Is it also similar to #1031 ?

parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -164,6 +164,12 @@ public CapacityByteArrayOutputStream(int initialSlabSize, int maxCapacityHint, B
   private void addSlab(int minimumSize) {
     int nextSlabSize;
 
+    if (bytesUsed + minimumSize < 0) {
```
wgtmac, line 167: nit: replace it with `Math.addExact` to hand over overflow check. Then we get the chance to check the 2GB hard limit before it becoming negative

parthchandra: Updated

parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -164,6 +164,15 @@ public CapacityByteArrayOutputStream(int initialSlabSize, int maxCapacityHint, B
   private void addSlab(int minimumSize) {
     int nextSlabSize;
 
+    // check for overflow 
+    try {
+      Math.addExact(bytesUsed, minimumSize);
+    } catch (ArithmeticException e) {
+      // This is interpreted as a request for a value greater than Integer.MAX_VALUE
+      // We throw OOM because that is what java.io.ByteArrayOutputStream also does
+      throw new OutOfMemoryError("Size of data exceeded 2GB (" + e.getMessage() + ")");
```
wgtmac, line 173: The error message mismatches the check above. Should we compare with 2GB explicitly or change the message?

parthchandra: The exception thrown by Math.exact will happen only if integer overflow occurs which happens when the data exceeds 2GB (i.e greater than `Integer.MAX_VALUE` ). 
What message were you thinking?

wgtmac: If we simply do an overflow check here, then the error message should say `Integer.MAX_VALUE` instead of `2GB`. Otherwise, we should explicitly check if the addition result exceeds 2GB. WDYT?

parthchandra: Let me just change the message.

PR #1033, opened 2023-02-24
PARQUET-2251 Avoid generating Bloomfilter when all pages of a column are encoded by dictionary in parquet v1

yabola:
In parquet pageV1, even all pages of a column are encoded by dictionary, it will still generate BloomFilter. Actually it is unnecessary and it costs time and occupies storage. Parquet pageV2 doesn't have this problem.

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestStoreBloomFilter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestStoreBloomFilter.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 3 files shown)


yabola:
@wgtmac @gerashegalov Please take a look, thank you~ 
And I will update  [PR](https://github.com/apache/parquet-mr/pull/1023) to skip bloomfilter when all pages are encoded in dictionary (because there are already many wrong parquet files generated)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestStoreBloomFilter.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public TestStoreBloomFilter(Path file, String pageVersion) {
+    this.file = file;
+    this.pageVersion = pageVersion;
+  }
+
+  @Parameterized.Parameters(name = "Run {index}: parquet {1}")
+  public static Collection<Object[]> params() {
+    return Arrays.asList(
+      new Object[]{FILE_V1, "pageV1"},
```
wgtmac, line 66: nit: the name is a little bit confusing. file format version is not same as data page version.

wgtmac (approved):
LGTM.

cc @gszadovszky

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestStoreBloomFilter.java
yabola: thanks, I modified the description

gszadovszky (approved):
Thanks, @yabola, @wgtmac.

yabola:
@wgtmac @gszadovszky Thank you for your review and help

PR #1034, opened 2023-02-24
PARQUET-2230: Add a new rewrite command powered by ParquetRewriter

wgtmac:
### Jira

https://issues.apache.org/jira/PARQUET-2230

### Tests

Added `RewriteCommandTest`.

### Commits

Implemented RewriteCommand to parquet-cli to support rewriting one or more parquet files with prune, mask and transform-codec.

### Documentation

Updated README.md to reflect the new command.

### DIFF
--- a/parquet-cli/README.md
+++ b/parquet-cli/README.md
@@ -117,6 +117,8 @@ Usage: parquet [options] [command] [command options]
         Check bloom filters for a Parquet column
     scan
         Scan all records from a file
+    rewrite
+        Rewrite one or more Parquet files to a new Parquet file
 
   Examples:
 
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/Main.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/Main.java
@@ -34,6 +34,7 @@ import org.apache.parquet.cli.commands.ConvertCSVCommand;
 import org.apache.parquet.cli.commands.ConvertCommand;
 import org.apache.parquet.cli.commands.ParquetMetadataCommand;
 import org.apache.parquet.cli.commands.PruneColumnsCommand;
+import org.apache.parquet.cli.commands.RewriteCommand;
 import org.apache.parquet.cli.commands.ScanCommand;
 import org.apache.parquet.cli.commands.SchemaCommand;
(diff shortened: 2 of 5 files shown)


wgtmac:
Please take a look when you get the chance. Thanks!
@gszadovszky @ggershinsky @shangxinli

gszadovszky (approved):
Thanks a lot for working on this, @wgtmac!

Are we planning to deprecate the other tools covered by this one? (Or if they were not released yet we might simply remove them?)

Next potential topic around parquet-cli if you're intereseted :)
There are some implementations around the hadoop conf in parquet-cli but I do not fully understand how it works. If it works as is we should document it otherwise we should make it work somehow. It would be great if the different read/write flags could be used in the tools. Like setting the zstd compression ratio for rewrite. Or using a non-default encoding. What do you think?

wgtmac:
> Are we planning to deprecate the other tools covered by this one? (Or if they were not released yet we might simply remove them?)

Let me take a look. If those tools and classes are not released yet, it would be a good time to remove them.

> Next potential topic around parquet-cli if you're intereseted :) There are some implementations around the hadoop conf in parquet-cli but I do not fully understand how it works. If it works as is we should document it otherwise we should make it work somehow. It would be great if the different read/write flags could be used in the tools. Like setting the zstd compression ratio for rewrite. Or using a non-default encoding. What do you think?

I am not familiar with it either. I will dig into it to find how it works and what can be done later.

It is Friday now. Will get back to it next week. :)

wgtmac:
@gszadovszky Unfortunately, those commands and classes have been released in v1.12.0. I have already annotated `deprecated` to those classes. Do you think it makes sense to deprecate those commands as well?

gszadovszky:
@wgtmac, I think it would be nice to inform the cli users as well about the deprecation. There is no harm keeping these commands only that if the related implementation is deprecated we won't maintain the functionality of them.

wgtmac:
> @wgtmac, I think it would be nice to inform the cli users as well about the deprecation. There is no harm keeping these commands only that if the related implementation is deprecated we won't maintain the functionality of them.

Agreed. I will submit a PR to address this. Thanks!

PR #1035, opened 2023-02-26
PARQUET-2202: Review usage and implementation of Preconditions.checkargument method

jerolba:
The [original Jira ticket ](https://issues.apache.org/jira/browse/PARQUET-2202) references to a concrete bad usage of the `Preconditions.checkArgument` method, where a String is calculated before validating the check, creating an overhead when in theory 100% of the cases the composed String will not be used.

The proposed solution inlines the call to `Preconditions.checkArgument`, but the correct approach is to call the `checkArgument` method with a String to format if the argument check is not valid.

A similar issue ocurrs in the [constructor](https://github.com/apache/parquet-mr/blob/62b774cd0f0c60cfbe540bbfa60bee15929af5d4/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java#L153) of the `CapacityByteArrayOutputStream` class, where the error message is always created using the `String.format`, instead of passing the template String and its params to the `checkArgument` method. This performance issue is also visible in a profiling.

The MR fixes both cases, and reviews the usage of `Preconditions` class, to ensure that error messages are not calculated before checking the boolean expression.

I've also reviewed `Preconditions` methods to improve the performance. When you call to a method with a varargs argument, Java internally allocates an array containing all values. To avoid this allocation when the number of params is very low, is recomended to overload the method with versions of the method with different number of arguments. This approach is heavily used in logging frameworks, or in [Guava Preconditions implementation](https://github.com/google/guava/blob/4312d949967f3fb245636f66437a00dd8c346d38/guava/src/com/google/common/base/Preconditions.java#L118)
Because nearly 100% of cases are going to check the condition as valid, the Array associated with varargs call to `Preconditions.strings` method will never be created.

Because all changes are related with `Preconditions.checkargument`, I've created a single PR. I can split it in multiple PRs if needed. Each commit of the PR makes a type of change, but I can squash into a single commit if needed.

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references them in the PR title. For example, "PARQUET-2202: My Parquet PR"
  - https://issues.apache.org/jira/browse/PARQUET-2202

### Tests

- [x] My PR adds test cases to the following unit tests:
  1. https://github.com/jerolba/parquet-mr/blob/review_usage_of_preconditions_checkargument/parquet-common/src/test/java/org/apache/parquet/TestPreconditions.java

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -477,7 +477,7 @@ public class ParquetProperties {
      * @return this builder for method chaining
      */
     public Builder withBloomFilterNDV(String columnPath, long ndv) {
-      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %d", columnPath, ndv);
+      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %s", columnPath, ndv);
       this.bloomFilterNDVs.withValue(columnPath, ndv);
       // Setting an NDV for a column implies writing a bloom filter
       this.bloomFilterEnabled.withValue(columnPath, true);
@@ -485,7 +485,7 @@ public class ParquetProperties {
     }
 
     public Builder withBloomFilterFPP(String columnPath, double fpp) {
-      Preconditions.checkArgument(fpp > 0.0 && fpp < 1.0, "Invalid FPP for column \"%s\": %d", columnPath, fpp);
+      Preconditions.checkArgument(fpp > 0.0 && fpp < 1.0, "Invalid FPP for column \"%s\": %s", columnPath, fpp);
       this.bloomFilterFPPs.withValue(columnPath, fpp);
       return this;
     }
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -520,7 +520,7 @@ public class ParquetProperties {
     }
 
     public Builder withPageRowCountLimit(int rowCount) {
-      Preconditions.checkArgument(rowCount > 0, "Invalid row count limit for pages: " + rowCount);
+      Preconditions.checkArgument(rowCount > 0, "Invalid row count limit for pages: %s", rowCount);
       pageRowCountLimit = rowCount;
       return this;
     }
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
@@ -392,7 +392,7 @@ class AvroIndexedRecordConverter<T extends IndexedRecord> extends GroupConverter
       @Override
       public Converter getConverter(int fieldIndex) {
         Preconditions.checkArgument(
-            fieldIndex == 0, "Illegal field index: " + fieldIndex);
+            fieldIndex == 0, "Illegal field index: %s", fieldIndex);
         return elementConverter;
       }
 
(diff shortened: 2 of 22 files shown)


parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -477,15 +477,15 @@ public Builder withMaxBloomFilterBytes(int maxBloomFilterBytes) {
...
+      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %s", columnPath, ndv);
       this.bloomFilterNDVs.withValue(columnPath, ndv);
       // Setting an NDV for a column implies writing a bloom filter
       this.bloomFilterEnabled.withValue(columnPath, true);
       return this;
     }
 
     public Builder withBloomFilterFPP(String columnPath, double fpp) {
-      Preconditions.checkArgument(fpp > 0.0 && fpp < 1.0, "Invalid FPP for column \"%s\": %d", columnPath, fpp);
+      Preconditions.checkArgument(fpp > 0.0 && fpp < 1.0, "Invalid FPP for column \"%s\": %s", columnPath, fpp);
```
wgtmac, line 488: Is `%f` better for `fpp`?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -477,15 +477,15 @@ public Builder withMaxBloomFilterBytes(int maxBloomFilterBytes) {
      * @return this builder for method chaining
      */
     public Builder withBloomFilterNDV(String columnPath, long ndv) {
-      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %d", columnPath, ndv);
+      Preconditions.checkArgument(ndv > 0, "Invalid NDV for column \"%s\": %s", columnPath, ndv);
```
wgtmac, line 480: Why use `%s` for long?

wgtmac (commented):
Is the change exhaustive?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
jerolba: The changed code was incorrect and in case of failure in the check, the code throws a `java.util.IllegalFormatConversionException`

The existing [code behind checkArgument method](https://github.com/jerolba/parquet-mr/blob/5078cc629074046b5699544c33256ca980ab2761/parquet-common/src/main/java/org/apache/parquet/Preconditions.java#L243) converts all parameters to String before applying the String.format.

If String.format finds an incorrect type, throws `IllegalFormatConversionException`. 

For example try this:
```java
String output = String.format("Integer value: %d", "10");
```

Then, existing Preconditions implementation forces to use Strings messages with %s.

I reviewed all calls to `checkArgument` to fix the incorrect format.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
jerolba: The same logic than previous comment. Fails in case of check failure.

jerolba:
> Is the change exhaustive?

Yes, I reviewed all calls to `checkArgument` to fix incorrect usage of %f and %d

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
wgtmac: Thanks for the explanation. SGTM.

wgtmac (approved):
cc @gszadovszky @ggershinsky @shangxinli

gszadovszky (changes requested):
Could you please also check the usages of `Preconditions.checkState`? There are a couple of places where string concatenation still happens (e.g. [Types](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/schema/Types.java#L478-L481)).

gszadovszky (approved):

PR #1038, opened 2023-03-01
PARQUET-2252: Make some methods public to allow external projects to …

zhongyujiang:
…implement page skipping.

Issue: [PARQUET-2252](https://issues.apache.org/jira/browse/PARQUET-2252)

This PR makes some methods public(methods in `RowRanges` and `IndexIterator` for calculating `RowRanges`, and `getColumnIndexStore` in `ParquetFileReader` for getting column index store), Iceberg needs these to build its own column index filtering.  Since Iceberg is going to calculate `RowRanges` itself, this also adds a public method in `ParquetFileReader` that allows users to pass in `RowRanges` to read filtered row group. Use of these changes can refer to this [PR](https://github.com/apache/iceberg/pull/6967), currently it uses reflection as a workaround.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1038,7 +1038,7 @@ public class ParquetFileReader implements Closeable {
 
     BlockMetaData block = blocks.get(blockIndex);
     if (block.getRowCount() == 0L) {
-      throw new ParquetEmptyBlockException("Illegal row group of 0 rows");
+      return null;
     }
 
     long rowCount = rowRanges.rowCount();
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -46,6 +46,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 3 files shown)


zhongyujiang:
@wgtmac @rdblue Can you please help review this?

wgtmac:
@gszadovszky @shangxinli Do you have any concern?

gszadovszky:
Since these are already used in iceberg I think it is better to have them public and maintain backward compatibility.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1011,6 +1012,35 @@ public PageReadStore readFilteredRowGroup(int blockIndex) throws IOException {
...
+        "[%s, %s]", blockIndex, 0, blocks.size() - 1));
+    }
+
+    if (Objects.isNull(rowRanges)) {
+      throw new IllegalArgumentException("RowRanges must not be null");
+    }
+
+    BlockMetaData block = blocks.get(blockIndex);
+    if (block.getRowCount() == 0L) {
+      throw new ParquetEmptyBlockException("Illegal row group of 0 rows");
```
wgtmac, line 1041: Now the reader simply skips empty row groups instead of throw. Could you change this to be consistent?

wgtmac (changes requested):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
zhongyujiang: I checked PARQUET-2291, seems we only skip empty row group when using reader as a iterator, right? We skip empty row group in `readNextRowGroup()` , but not when the user passes in a `blockIndex`, and  this newly introduced method also requires the user pass in a `blockIndex`.

wgtmac: That makes sense.

wgtmac (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
rdblue: I don't see why this would throw an exception. This method is intended to allow building an external iterator. I don't think anyone would ever want to fail if there were an empty row group, even if the reader thinks it shouldn't have been written. I think this should return null.

rdblue (approved):
Overall looks fine, but I agree that the row group count check should just return null instead of throwing an exception.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
zhongyujiang: Currently, `readRowGroup(int blockIndex)` and `readFilteredRowGroup(int blockIndex)` also throw an exception when handling empty row groups, should we make them also return null instead of throwing an exception to be consistent?

PR #1039, opened 2023-03-02
PARQUET-2237 Improve performance by skipping BloomFilter  when there is already a dictionary filter

yabola:


yabola:
@wgtmac @gszadovszky If you have time, please take a look, thank you

yabola:
The implementation of `and` `or` in this PR is incorrect, I want to change to the previous implementation in https://github.com/apache/parquet-mr/pull/1023

PR #1041, opened 2023-03-17
Uniformizing booleans naming in ParquetWriter

hamza-tam:

### Jira

### Tests

- [x] My PR does not need testing for this extremely good reason: Only variable renaming

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -47,8 +47,8 @@ public class ParquetProperties {
 
   public static final int DEFAULT_PAGE_SIZE = 1024 * 1024;
   public static final int DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
-  public static final boolean DEFAULT_IS_DICTIONARY_ENABLED = true;
-  public static final boolean DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED = false;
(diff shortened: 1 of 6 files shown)


parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -130,7 +130,7 @@ private ParquetProperties(Builder builder) {
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
-    this.enableByteStreamSplit = builder.enableByteStreamSplit;
+    this.byteStreamSplitEnabled = builder.enableByteStreamSplit;
```
yabola, line 133: `builder.enableByteStreamSplit` should also rename as `byteStreamSplitEnabled`

yabola (approved):

yabola (changes requested):
Thanks for working on it. Overall looks well. it can help to unify the code style.

But you modified the public variable, which may lead to code  forward incompatibility. This may be unacceptable

shangxinli:
Yeah, public ones are tricky and we need to make it compatible. But overall, this is a great effort to improve our code.

Fokko:
Unfortunately, we cannot change public static methods, since other projects might be using them. If we want this we first need to deprecate the old values.

Fokko:
Closing this since we have to go through the deprecation cycle which seems a bit excessive for just a rename

PR #1042, opened 2023-03-19
PARQUET-2254 Support building bloom filter that adapts to the data

yabola:
Now the usage of bloom filter is to specify the NDV(number of distinct values) or max bytes, and then build BloomFilter. In general scenarios, it is actually not sure how much the distinct value is.
If BloomFilter can be automatically generated according to the data, the file size can be reduced and the reading efficiency can also be improved.

`AdaptiveBlockSplitBloomFilter` contains multiple `BlockSplitBloomFilter` as candidates and inserts values in
 the candidates at the same time. Finally we will choose the smallest candidate to write out.

### Does this PR introduce any user-facing change?
add new configuration:
`parquet.bloom.filter.adaptive.enabled` : default false, Whether to enable writing adaptive bloom filter.  
If it is true, the bloom filter will be generated with the optimal bit size according to the number of real data distinct values. If it is false, it will not take effect.
Note that the maximum bytes of the bloom filter will not exceed `parquet.bloom.filter.max.bytes` configuration (if it is 
set too small, the generated bloom filter will not be efficient).

`parquet.bloom.filter.candidates.number`: default 5, the number of candidate bloom filters written at the same time.  
When `parquet.bloom.filter.adaptive.enabled` is true, multiple candidate bloom filters will be inserted 
at the same time, finally a bloom filter with the optimal bit size will be selected and written to the file.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -108,7 +108,7 @@ public class ParquetProperties {
   private final int maxBloomFilterBytes;
   private final ColumnProperty<Boolean> bloomFilterEnabled;
   private final ColumnProperty<Boolean> adaptiveBloomFilterEnabled;
-  private final ColumnProperty<Integer> bloomFilterCandidatesNum;
+  private final ColumnProperty<Integer> numBloomFilterCandidates;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final boolean enableByteStreamSplit;
@@ -133,7 +133,7 @@ public class ParquetProperties {
     this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
     this.adaptiveBloomFilterEnabled = builder.adaptiveBloomFilterEnabled.build();
-    this.bloomFilterCandidatesNum = builder.bloomFilterCandidatesNum.build();
+    this.numBloomFilterCandidates = builder.numBloomFilterCandidates.build();
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
     this.enableByteStreamSplit = builder.enableByteStreamSplit;
@@ -286,7 +286,7 @@ public class ParquetProperties {
   }
 
   public int getBloomFilterCandidatesCount(ColumnDescriptor column) {
-    return bloomFilterCandidatesNum.getValue(column);
+    return numBloomFilterCandidates.getValue(column);
   }
 
   public static Builder builder() {
@@ -332,7 +332,7 @@ public class ParquetProperties {
     private final ColumnProperty.Builder<Double> bloomFilterFPPs;
     private int maxBloomFilterBytes = DEFAULT_MAX_BLOOM_FILTER_BYTES;
     private final ColumnProperty.Builder<Boolean> adaptiveBloomFilterEnabled;
-    private final ColumnProperty.Builder<Integer> bloomFilterCandidatesNum;
+    private final ColumnProperty.Builder<Integer> numBloomFilterCandidates;
     private final ColumnProperty.Builder<Boolean> bloomFilterEnabled;
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
     private boolean pageWriteChecksumEnabled = DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED;
@@ -344,7 +344,7 @@ public class ParquetProperties {
       bloomFilterNDVs = ColumnProperty.<Long>builder().withDefaultValue(null);
       bloomFilterFPPs = ColumnProperty.<Double>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_FPP);
       adaptiveBloomFilterEnabled = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_ADAPTIVE_BLOOM_FILTER_ENABLED);
-      bloomFilterCandidatesNum = ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER);
+      numBloomFilterCandidates = ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER);
     }
 
     private Builder(ParquetProperties toCopy) {
@@ -363,7 +363,7 @@ public class ParquetProperties {
       this.bloomFilterFPPs = ColumnProperty.<Double>builder(toCopy.bloomFilterFPPs);
       this.bloomFilterEnabled = ColumnProperty.<Boolean>builder(toCopy.bloomFilterEnabled);
       this.adaptiveBloomFilterEnabled = ColumnProperty.<Boolean>builder(toCopy.adaptiveBloomFilterEnabled);
-      this.bloomFilterCandidatesNum = ColumnProperty.<Integer>builder(toCopy.bloomFilterCandidatesNum);
+      this.numBloomFilterCandidates = ColumnProperty.<Integer>builder(toCopy.numBloomFilterCandidates);
       this.maxBloomFilterBytes = toCopy.maxBloomFilterBytes;
       this.enableByteStreamSplit = toCopy.enableByteStreamSplit;
     }
@@ -543,7 +543,7 @@ public class ParquetProperties {
      */
     public Builder withBloomFilterCandidatesNumber(String columnPath, int number) {
       Preconditions.checkArgument(number > 0, "Invalid candidates number for column \"%s\": %d", columnPath, number);
-      this.bloomFilterCandidatesNum.withDefaultValue(number);
+      this.numBloomFilterCandidates.withDefaultValue(number);
       return this;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -60,7 +60,7 @@ public class ParquetProperties {
   public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
   public static final double DEFAULT_BLOOM_FILTER_FPP = 0.01;
   public static final boolean DEFAULT_ADAPTIVE_BLOOM_FILTER_ENABLED = false;
-  public static final int DEFAULT_BLOOM_FILTER_CANDIDATE_SIZE = 5;
+  public static final int DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER = 5;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
@@ -108,7 +108,7 @@ public class ParquetProperties {
   private final int maxBloomFilterBytes;
   private final ColumnProperty<Boolean> bloomFilterEnabled;
   private final ColumnProperty<Boolean> adaptiveBloomFilterEnabled;
-  private final ColumnProperty<Integer> bloomFilterCandidateSize;
+  private final ColumnProperty<Integer> bloomFilterCandidatesNum;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final boolean enableByteStreamSplit;
@@ -133,7 +133,7 @@ public class ParquetProperties {
     this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
     this.adaptiveBloomFilterEnabled = builder.adaptiveBloomFilterEnabled.build();
-    this.bloomFilterCandidateSize = builder.bloomFilterCandidateSize.build();
+    this.bloomFilterCandidatesNum = builder.bloomFilterCandidatesNum.build();
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
     this.enableByteStreamSplit = builder.enableByteStreamSplit;
@@ -285,8 +285,8 @@ public class ParquetProperties {
     return adaptiveBloomFilterEnabled.getValue(column);
   }
 
-  public int getBloomFilterCandidateSize(ColumnDescriptor column) {
-    return bloomFilterCandidateSize.getValue(column);
+  public int getBloomFilterCandidatesCount(ColumnDescriptor column) {
+    return bloomFilterCandidatesNum.getValue(column);
   }
 
   public static Builder builder() {
@@ -332,7 +332,7 @@ public class ParquetProperties {
     private final ColumnProperty.Builder<Double> bloomFilterFPPs;
     private int maxBloomFilterBytes = DEFAULT_MAX_BLOOM_FILTER_BYTES;
     private final ColumnProperty.Builder<Boolean> adaptiveBloomFilterEnabled;
-    private final ColumnProperty.Builder<Integer> bloomFilterCandidateSize;
+    private final ColumnProperty.Builder<Integer> bloomFilterCandidatesNum;
     private final ColumnProperty.Builder<Boolean> bloomFilterEnabled;
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
     private boolean pageWriteChecksumEnabled = DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED;
@@ -344,7 +344,7 @@ public class ParquetProperties {
       bloomFilterNDVs = ColumnProperty.<Long>builder().withDefaultValue(null);
       bloomFilterFPPs = ColumnProperty.<Double>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_FPP);
       adaptiveBloomFilterEnabled = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_ADAPTIVE_BLOOM_FILTER_ENABLED);
-      bloomFilterCandidateSize = ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATE_SIZE);
+      bloomFilterCandidatesNum = ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER);
     }
 
     private Builder(ParquetProperties toCopy) {
@@ -363,7 +363,7 @@ public class ParquetProperties {
       this.bloomFilterFPPs = ColumnProperty.<Double>builder(toCopy.bloomFilterFPPs);
       this.bloomFilterEnabled = ColumnProperty.<Boolean>builder(toCopy.bloomFilterEnabled);
       this.adaptiveBloomFilterEnabled = ColumnProperty.<Boolean>builder(toCopy.adaptiveBloomFilterEnabled);
-      this.bloomFilterCandidateSize = ColumnProperty.<Integer>builder(toCopy.bloomFilterCandidateSize);
+      this.bloomFilterCandidatesNum = ColumnProperty.<Integer>builder(toCopy.bloomFilterCandidatesNum);
       this.maxBloomFilterBytes = toCopy.maxBloomFilterBytes;
       this.enableByteStreamSplit = toCopy.enableByteStreamSplit;
     }
@@ -539,11 +539,11 @@ public class ParquetProperties {
      * When `AdaptiveBloomFilter` is enabled, set how many bloom filter candidates to use.
      *
      * @param columnPath the path of the column (dot-string)
-     * @param size the candidate size
+     * @param number the number of candidates
      */
-    public Builder withBloomFilterCandidateSize(String columnPath, int size) {
-      Preconditions.checkArgument(size > 0, "Invalid candidate size for column \"%s\": %d", columnPath, size);
-      this.bloomFilterCandidateSize.withDefaultValue(size);
+    public Builder withBloomFilterCandidatesNumber(String columnPath, int number) {
+      Preconditions.checkArgument(number > 0, "Invalid candidates number for column \"%s\": %d", columnPath, number);
+      this.bloomFilterCandidatesNum.withDefaultValue(number);
       return this;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -59,7 +59,7 @@ public class ParquetProperties {
   public static final int DEFAULT_MAX_BLOOM_FILTER_BYTES = 1024 * 1024;
   public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
   public static final double DEFAULT_BLOOM_FILTER_FPP = 0.01;
-  public static final boolean DEFAULT_DYNAMIC_BLOOM_FILTER_ENABLED = true;
+  public static final boolean DEFAULT_DYNAMIC_BLOOM_FILTER_ENABLED = false;
   public static final int DEFAULT_BLOOM_FILTER_CANDIDATE_SIZE = 5;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
@@ -528,10 +528,9 @@ public class ParquetProperties {
      * `parquet.bloom.filter.max.bytes`.
      * If NDV (number of distinct values) for a specified column is set, it will be ignored
      *
-     * @param columnPath the path of the column (dot-string)
-     * @param enabled whether to write bloom filter for the column
+     * @param enabled whether to use dynamic bloom filter
      */
-    public Builder withDynamicBloomFilterEnabled(String columnPath, boolean enabled) {
+    public Builder withDynamicBloomFilterEnabled(boolean enabled) {
       this.dynamicBloomFilterEnabled.withDefaultValue(enabled);
       return this;
     }
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -18,6 +18,8 @@
  */
 package org.apache.parquet.column;
 
+import static org.apache.parquet.bytes.BytesUtils.getWidthFromMaxInt;
+
 import java.util.Objects;
 import java.util.OptionalDouble;
 import java.util.OptionalLong;
@@ -38,8 +40,6 @@ import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWriter;
 import org.apache.parquet.schema.MessageType;
 
-import static org.apache.parquet.bytes.BytesUtils.getWidthFromMaxInt;
-
 /**
  * This class represents all the configurable Parquet properties.
  */
@@ -59,6 +59,8 @@ public class ParquetProperties {
   public static final int DEFAULT_MAX_BLOOM_FILTER_BYTES = 1024 * 1024;
   public static final boolean DEFAULT_BLOOM_FILTER_ENABLED = false;
   public static final double DEFAULT_BLOOM_FILTER_FPP = 0.01;
+  public static final boolean DEFAULT_DYNAMIC_BLOOM_FILTER_ENABLED = true;
+  public static final int DEFAULT_BLOOM_FILTER_CANDIDATE_SIZE = 5;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
@@ -105,6 +107,8 @@ public class ParquetProperties {
   private final ColumnProperty<Double> bloomFilterFPPs;
   private final int maxBloomFilterBytes;
   private final ColumnProperty<Boolean> bloomFilterEnabled;
+  private final ColumnProperty<Boolean> dynamicBloomFilterEnabled;
+  private final ColumnProperty<Integer> bloomFilterCandidateSize;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final boolean enableByteStreamSplit;
@@ -128,6 +132,8 @@ public class ParquetProperties {
     this.bloomFilterFPPs = builder.bloomFilterFPPs.build();
     this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
     this.maxBloomFilterBytes = builder.maxBloomFilterBytes;
+    this.dynamicBloomFilterEnabled = builder.dynamicBloomFilterEnabled.build();
+    this.bloomFilterCandidateSize = builder.bloomFilterCandidateSize.build();
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
     this.enableByteStreamSplit = builder.enableByteStreamSplit;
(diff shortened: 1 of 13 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,317 @@
...
+  /**
+   * Given the maximum acceptable bytes size of bloom filter, generate candidates according
+   * to the max expected distinct values. The size of the candidate bytes needs to be a
+   * square number of 2. Therefore, set the candidate size according to `maxBytes` of `1/2`, `1/4`, `1/8`, etc.
+   *
+   * @param maxBytes      the maximum acceptable bit size
+   * @param candidatesNum the number of candidates
+   * @param fpp           the false positive probability
+   */
+  private void initCandidates(int maxBytes, int candidatesNum, double fpp) {
```
yabola, line 110: `https://github.com/apache/parquet-mr/blob/1235003e742e6a76bf6cb8f7ed33e942fa12d0d5/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L197-L219`

yabola:
https://github.com/apache/parquet-mr/blob/1235003e742e6a76bf6cb8f7ed33e942fa12d0d5/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L197-L219

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola, line 110: In `BlockSplitBloomFilter` , the `byteSize` will be set to the size of the power of 2, therefore here the candidates are also divided according to the power of 2 (`maxBytes` of `1/2`, `1/4`, `1/8`, etc )

https://github.com/apache/parquet-mr/blob/1235003e742e6a76bf6cb8f7ed33e942fa12d0d5/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L206-L219

yabola: In the original implementation,  `Integer.highestOneBit(numBytes)` got the largest power of two less than `numBytes`.  And I think `<< 1` here makes result larger to achieve the effect of receiving NVD values.

yabola (commented):
@wgtmac @gszadovszky If you have time, please take a look , thank you

yabola:
5 candidates are set by default, and 2M storage is occupied by default (the previous default was 1M):
`[expectedNDV=540000, numBytes=65536],[expectedNDV=108000, numBytes=131072],[expectedNDV=216500, numBytes=262144],
[expectedNDV=433000, numBytes=524288],[expectedNDV=866000, numBytes=1048576]`

yabola:
Calculate the size of nvd according to the size of bytes, I also saw similar logic in impala
https://github.com/apache/impala/blob/2c779939dc302be9ee5dd97ddf374bb043040891/be/src/kudu/util/block_bloom_filter.cc#L258-L288

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,316 @@
...
+  @Override
+  public long hash(Binary value) {
+    return maxCandidate.bloomFilter.hash(value);
+  }
+
+  protected class BloomFilterCandidate implements Comparable<BloomFilterCandidate> {
+    // the bloom filter candidate
+    private BlockSplitBloomFilter bloomFilter;
+    // the expected NDV (number of distinct value) in the candidate
+    private int expectedNDV;
```
wgtmac, line 273: Mark it final?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  @Override
+  public boolean findHash(long hash) {
+    return maxCandidate.bloomFilter.findHash(hash);
+  }
+
+  @Override
+  public long hash(Object value) {
+    return maxCandidate.bloomFilter.hash(value);
```
wgtmac, line 226: Does the hash value change from different candidates? IIUC, the hash value is fixed regardless of bitset size.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  /**
+   * Used at the end of the insertion, select the candidate of the smallest size.
+   * At least one of the largest candidates will be kept when inserting data.
+   *
+   * @return the smallest and optimal candidate
+   */
+  protected BloomFilterCandidate optimalCandidate() {
+    return candidates.stream()
```
wgtmac, line 177: IMO, using TreeSet here is an overkill because the bitset size is fixed. We can simply use an ArrayList with allocated size in the descending order. Removing the last elements if its size is not enough for the current NDV.

wgtmac (commented):
I just took a preliminary look. Would be good to discuss this once the related PR gets merged.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: Yes, hash value is the same, it depends on the `hashFunction`, the `hashFunction` of the candidates is the same.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: thank you for your review, done

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: I changed to use `ArrayList`

yabola:
I changed Dynamic BloomFilter to false by default, keeping the same behavior as before

wgtmac:
Sorry for the delay as getting busy these days. Will try to take a look later.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+import org.apache.parquet.io.api.Binary;
+
+/**
+ * `DynamicBlockBloomFilter` contains multiple `BlockSplitBloomFilter` as candidates and inserts values in
+ * the candidates at the same time.
+ * The purpose of this is to finally generate a bloom filter with the optimal bit size according to the number
+ * of real data distinct values. Use the largest bloom filter as an approximate deduplication counter, and then
+ * remove incapable bloom filter candidate during data insertion.
+ */
+public class DynamicBlockBloomFilter implements BloomFilter {
```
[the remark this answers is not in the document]
wgtmac: It could be either `DynamicBloomFilter` or `DynamicBlockSplitBloomFilter`, but not in between. Please also fix the comment above.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(DynamicBlockBloomFilter.class);
+
+  // multiple candidates, inserting data at the same time. If the distinct values are greater than the
+  // expected NDV of candidates, it will be removed. Finally we will choose the smallest candidate to write out.
+  private final List<BloomFilterCandidate> candidates = new ArrayList<>();
+
+  // the largest among candidates and used as an approximate deduplication counter
+  private BloomFilterCandidate maxCandidate;
```
wgtmac, line 56:
```suggestion
  private BloomFilterCandidate largestCandidate;
```

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+
+  // the accumulator of the number of distinct values that have been inserted so far
+  private int distinctValueCounter = 0;
```
wgtmac, line 59: Is `int` enough?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+  // the hash strategy used in this bloom filter.
+  private final HashStrategy hashStrategy;
+  // the column to build bloom filter
+  private ColumnDescriptor column;
+
+  public DynamicBlockBloomFilter(int numBytes, int candidatesNum, double fpp, ColumnDescriptor column) {
+    this(numBytes, LOWER_BOUND_BYTES, UPPER_BOUND_BYTES, HashStrategy.XXH64, fpp, candidatesNum, column);
+  }
+
+  public DynamicBlockBloomFilter(int numBytes, int maximumBytes, int candidatesNum, double fpp, ColumnDescriptor column) {
```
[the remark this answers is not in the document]
wgtmac: I know these constructors are following the same style of `BlockSplitBloomFilter`, but I really doubt if this one is necessary at all. Specifically, how do we know `numBytes` in advance?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+  /**
+   * Given the maximum acceptable bytes size of bloom filter, generate candidates according
+   * to the bytes size. The bytes size of the candidate needs to be a
+   * power of 2. Therefore, set the candidate size according to `maxBytes` of `1/2`, `1/4`, `1/8`, etc.
+   *
+   * @param maxBytes      the maximum acceptable bit size
+   * @param candidatesNum the number of candidates
+   * @param fpp           the false positive probability
+   */
+  private void initCandidates(int maxBytes, int candidatesNum, double fpp) {
```
wgtmac, line 112:
```suggestion
  private void initCandidates(int maxBytes, int numCandidates, double fpp) {
```

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+    for (int i = 1; i <= candidatesNum; i++) {
+      int candidateExpectedNDV = expectedNDV(candidateByteSize, fpp);
+      // `candidateByteSize` is too small, just drop it
+      if (candidateExpectedNDV <= 0) {
+        break;
+      }
+      BloomFilterCandidate candidate =
+        new BloomFilterCandidate(candidateExpectedNDV, candidateByteSize, minimumBytes, maximumBytes, hashStrategy);
+      candidates.add(candidate);
+      candidateByteSize = calculateTwoPowerSize(candidateByteSize / 2);
```
wgtmac, line 123: If `candidateByteSize` is power of two already in the line 113, can be simply apply `candidateByteSize /= 2` here?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+    Optional<BloomFilterCandidate> maxBloomFilter = candidates.stream().max(BloomFilterCandidate::compareTo);
+    if (maxBloomFilter.isPresent()) {
+      maxCandidate = maxBloomFilter.get();
+    } else {
+      throw new IllegalArgumentException("`maximumBytes` is too small to create one valid bloom filter");
```
wgtmac, line 129: Should we simply skip this candidate and add a warn log here? It may not be a fatal error.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+   * The expected number result may be slightly smaller than what `numBytes` can support.
+   *
+   * @param numBytes the bytes size
+   * @param fpp      the false positive probability
+   * @return the expected number of distinct values
+   */
+  private int expectedNDV(int numBytes, double fpp) {
+    int expectedNDV = 0;
+    int optimalBytes = 0;
+    int step = 500;
```
wgtmac, line 144: Should we put the magic number as a static variable and define it on top of the class? And why 500 instead of a power of two?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+    return expectedNDV;
+  }
+
+  /**
+   * BloomFilter bitsets size should be power of 2, see [[BlockSplitBloomFilter#initBitset]]
+   *
+   * @param numBytes the bytes size
+   * @return the largest power of 2 less or equal to numBytes
+   */
+  private int calculateTwoPowerSize(int numBytes) {
```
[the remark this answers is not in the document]
wgtmac: We can simply rename it as `calculateBoundedPowerOfTwo`?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  /**
+   * Used at the end of the insertion, select the candidate of the smallest size.
+   * At least one of the largest candidates will be kept when inserting data.
+   *
+   * @return the smallest and optimal candidate
+   */
+  protected BloomFilterCandidate optimalCandidate() {
+    return candidates.stream().min(BloomFilterCandidate::compareTo).get();
```
wgtmac, line 184: Is `candidates.get(0)` enough? Should we throw if candidates is empty?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  protected List<BloomFilterCandidate> getCandidates() {
+    return candidates;
+  }
+
+  @Override
+  public void writeTo(OutputStream out) throws IOException {
+    finalized = true;
```
wgtmac, line 193: Check `finalized` first in case of re-entry.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+
+  /**
+   * Insert an element to the multiple bloom filter candidates and remove the bad candidate
+   * if the number of distinct values exceeds its expected size.
+   *
+   * @param hash the hash result of element.
+   */
+  @Override
+  public void insertHash(long hash) {
+    Preconditions.checkArgument(!finalized,
```
wgtmac, line 210: Should we check `finalized` and `candidates.isEmpty()` ?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+      distinctValueCounter++;
+    }
+    // distinct values exceed the expected size, remove the bad bloom filter (leave at least the max bloom filter candidate)
+    candidates.removeIf(candidate -> candidate.getExpectedNDV() < distinctValueCounter && candidate != maxCandidate);
+    candidates.forEach(candidate -> candidate.getBloomFilter().insertHash(hash));
+  }
+
+  @Override
+  public int getBitsetSize() {
+    return optimalCandidate().getBloomFilter().getBitsetSize();
```
wgtmac, line 222: What if optimal does not exist? Will it throw in the `optimalCandidate()`?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  @Override
+  public boolean findHash(long hash) {
+    return maxCandidate.bloomFilter.findHash(hash);
```
wgtmac, line 227: Wrap `maxCandidate` in a function and throw if none of candidates are alive. Same for functions below.

wgtmac (commented):
Sorry for the delay. I have left some comments.

yabola:
@wgtmac Thank you for your review! I will address your comments recently.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -97,7 +97,13 @@ abstract class ColumnWriterBase implements ColumnWriter {
       int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
       this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
     } else {
-      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
+      boolean useDynamicBloomFilter = props.getDynamicBloomFilterEnabled(path);
+      if(useDynamicBloomFilter) {
```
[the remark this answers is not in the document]
yabola: done~

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -503,6 +523,30 @@ public Builder withBloomFilterEnabled(boolean enabled) {
...
+     *
+     * @param enabled whether to use dynamic bloom filter
+     */
+    public Builder withDynamicBloomFilterEnabled(boolean enabled) {
+      this.dynamicBloomFilterEnabled.withDefaultValue(enabled);
+      return this;
+    }
+
+    /**
+     * When `DynamicBloomFilter` is enabled, set how many bloomFilters to split as candidates.
```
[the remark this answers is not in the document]
yabola: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -152,6 +153,8 @@ public static enum JobSummaryLevel {
   public static final String BLOOM_FILTER_EXPECTED_NDV = "parquet.bloom.filter.expected.ndv";
   public static final String BLOOM_FILTER_MAX_BYTES = "parquet.bloom.filter.max.bytes";
   public static final String BLOOM_FILTER_FPP = "parquet.bloom.filter.fpp";
+  public static final String DYNAMIC_BLOOM_FILTER_ENABLED = "parquet.bloom.filter.dynamic.enabled";
```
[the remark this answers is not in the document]
yabola: @wgtmac In the end, at least one BloomFilter will be retained, because I won't remove the `largestCandidate` (please see `AdaptiveBlockSplitBloomFilter#insertHash`)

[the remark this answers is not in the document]
yabola: I changed the name to `AdaptiveBlockSplitBloomFilter`, how do you feel?~

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -19,6 +19,11 @@
 package org.apache.parquet.hadoop;
 
 import static java.util.Arrays.asList;
+import static org.apache.parquet.schema.LogicalTypeAnnotation.stringType;
```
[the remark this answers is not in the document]
yabola: sorry, revert these changes

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -152,6 +153,8 @@ public static enum JobSummaryLevel {
...
+  public static final String BLOOM_FILTER_CANDIDATE_SIZE = "parquet.bloom.filter.candidate.size";
```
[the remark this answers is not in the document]
yabola: emmm, I want to align BloomFilter's configurations format `parquet.bloom.filter.xxx`

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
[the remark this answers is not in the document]
yabola: done

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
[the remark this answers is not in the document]
yabola: I deleted some unimportant configurations and  I changed to name `numBytes` as `maximumBytes` (means  the maximum bit size of candidate)

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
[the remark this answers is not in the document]
yabola: done

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: changed to use long

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: The `calculateTwoPowerSize` method also has the function of controlling the lower limit

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: we need to make sure we have at least one BloomFilter candidate

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: `step` is used to increase number of distinct value(NDV), not for numBytes, so I don't use power of two

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: Actually I want to preserve the ability to output multiple times to `OutputStream`.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/DynamicBlockBloomFilter.java
yabola: `candidates` won't be empty

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
```diff
@@ -0,0 +1,307 @@
...
+      BloomFilterCandidate candidate =
+        new BloomFilterCandidate(candidateExpectedNDV, candidateByteSize, minimumBytes, maximumBytes, hashStrategy);
+      candidates.add(candidate);
+      candidateByteSize = calculateBoundedPowerOfTwo(candidateByteSize / 2);
+    }
+    Optional<BloomFilterCandidate> maxBloomFilter = candidates.stream().max(BloomFilterCandidate::compareTo);
+    if (maxBloomFilter.isPresent()) {
+      largestCandidate = maxBloomFilter.get();
+    } else {
+      throw new IllegalArgumentException("`maximumBytes` is too small to create one valid bloom filter");
```
wgtmac, line 123: My previous comment here means we should not fail the writer if bloom filter cannot be generated. This is not a fatal error. 

If the writer will fail due to a misconfiguration, we should probably document this so users will not be surprised?

WDYT?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -152,6 +153,8 @@ public static enum JobSummaryLevel {
   public static final String BLOOM_FILTER_EXPECTED_NDV = "parquet.bloom.filter.expected.ndv";
   public static final String BLOOM_FILTER_MAX_BYTES = "parquet.bloom.filter.max.bytes";
   public static final String BLOOM_FILTER_FPP = "parquet.bloom.filter.fpp";
+  public static final String ADAPTIVE_BLOOM_FILTER_ENABLED = "parquet.bloom.filter.adaptive.enabled";
```
wgtmac, line 156: Please document new configs here: https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/README.md

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
```diff
@@ -0,0 +1,307 @@
...
+   */
+  @Override
+  public void insertHash(long hash) {
+    Preconditions.checkArgument(!finalized,
+      "AdaptiveBlockSplitBloomFilter insertion has been mark as finalized, no more data is allowed!");
+    if (!largestCandidate.bloomFilter.findHash(hash)) {
+      distinctValueCounter++;
+    }
+    // distinct values exceed the expected size, remove the bad bloom filter (leave at least the max bloom filter candidate)
+    candidates.removeIf(candidate -> candidate.getExpectedNDV() < distinctValueCounter && candidate != largestCandidate);
```
wgtmac, line 209: In the worst case where the maximum size is too small and many distinct values have been inserted, the `largestCandidate` will set TRUE to almost all bits making it less effective. Why not simply drop all the candidates?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
```diff
@@ -38,57 +38,51 @@
...
-  public DynamicBlockBloomFilter(int numBytes, int candidatesNum, double fpp, ColumnDescriptor column) {
-    this(numBytes, LOWER_BOUND_BYTES, UPPER_BOUND_BYTES, HashStrategy.XXH64, fpp, candidatesNum, column);
-  }
-
-  public DynamicBlockBloomFilter(int numBytes, int maximumBytes, int candidatesNum, double fpp, ColumnDescriptor column) {
-    this(numBytes, LOWER_BOUND_BYTES, maximumBytes, HashStrategy.XXH64, fpp, candidatesNum, column);
+  /**
+   * Given the maximum acceptable bytes size of bloom filter, generate candidates according it.
+   *
+   * @param maximumBytes  the maximum bit size of candidate
```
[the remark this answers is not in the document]
yabola: It means byte size and I have changed variable name, thank you

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
yabola: I agree with you, this shouldn't be a fatal error. 
However, if  follow the ordinary bloom filter, no matter how small the bytes setting is, there will be a lower bound bloom filter generation (32 bytes size). So I modified it to have at least one minimum BloomFilter(32 bytes size) generation to align with the original implementation. How do you feel?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
yabola: done, and I changed `parquet.bloom.filter.candidate.size` to `parquet.bloom.filter.candidates.number`.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
yabola: In the original BloomFilter, it will generate even if the deduplication value exceeds expectations, so I want to maintain this behavior. WDYT?

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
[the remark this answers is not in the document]
yabola: @wgtmac sorry for late reply, I am little busy these days.
It means byte size and I have changed variable name, thank you

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
```diff
@@ -0,0 +1,307 @@
...
+   * if the number of distinct values exceeds its expected size.
+   *
+   * @param hash the hash result of element.
+   */
+  @Override
+  public void insertHash(long hash) {
+    Preconditions.checkArgument(!finalized,
+      "AdaptiveBlockSplitBloomFilter insertion has been mark as finalized, no more data is allowed!");
+    if (!largestCandidate.bloomFilter.findHash(hash)) {
+      distinctValueCounter++;
```
wgtmac, line 206: From the comment at line 57, it says `distinctValueCounter` is `the accumulator of the number of distinct values that have been inserted so far`. Actually this is the counter for distinct hash values instead of original values. Of course, the increment of `distinctValueCounter` can indicate a new value has been discovered. However, a found hash may hide a new value if the hash value conflicts. We may need to document this behavior. Probably `distinctValueCounter` can be renamed to `numDistinctHashValues`.

parquet-hadoop/README.md
```diff
@@ -211,6 +211,23 @@ conf.set("parquet.bloom.filter.enabled#column.path", false);
 
 ---
 
+**Property:** `parquet.bloom.filter.adaptive.enabled`  
+**Description:** Whether to enable writing adaptive bloom filter.  
+If it is true, the bloom filter will be generated with the optimal bit size 
+according to the number of real data distinct values. If it is false, it will not take effect.
+Note that the maximum bytes of the bloom filter will not exceed `parquet.bloom.filter.max.bytes` configuration.
```
wgtmac, line 218: nit: would be good to explain the generated bloom filter will not be efficient if `parquet.bloom.filter.max.bytes` is too small.

wgtmac (commented):
Thanks, I left two minor comments. Rest LGTM.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/AdaptiveBlockSplitBloomFilter.java
yabola: Yeah, I modified the variable name, thank you.

parquet-hadoop/README.md
yabola: done

wgtmac (approved):
Thanks, LGTM

yabola:
@wgtmac Thank you for your detailed review

PR #1043, opened 2023-03-19
PARQUET-2260 Bloom filter size shouldn't be larger than maxBytes in the configuration

yabola:
Before this PR: If `parquet.bloom.filter.max.bytes` configuration is not a power of 2 value,  the size of the bloom filter generated will exceed this value. For example, if set `parquet.bloom.filter.max.bytes` as 1024 * 1024+1= 1048577 , the bytes size of bloom filter generated will be 1024 * 1024 * 2 = 2097152. This does not match the definition of the parameter

After this PR: If `parquet.bloom.filter.max.bytes` configuration is not a power of 2, and the size of `parquet.bloom.filter.max.bytes` is the final bloom filter bytes size.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
@@ -97,7 +97,7 @@ abstract class ColumnWriterBase implements ColumnWriter {
       int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
       this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
     } else {
-      this.bloomFilter = BlockSplitBloomFilter.of(maxBloomFilterSize);
+      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
     }
   }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
@@ -97,7 +97,7 @@ abstract class ColumnWriterBase implements ColumnWriter {
       int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
       this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
     } else {
-      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize);
+      this.bloomFilter = BlockSplitBloomFilter.of(maxBloomFilterSize);
     }
   }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -86,22 +86,6 @@ public class BlockSplitBloomFilter implements BloomFilter {
     0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
 
   /**
-   * Create a Bloom filter within the maximum bytes size. The bloom filter bytes size
-   * will be a largest power of 2 less than the maximum bytes size.
-   *
-   * @param maxBytes The max number of bytes for Bloom filter.
-   * @return
-   */
-  public static BlockSplitBloomFilter of(int maxBytes) {
-    if (maxBytes < LOWER_BOUND_BYTES) {
-      maxBytes = LOWER_BOUND_BYTES;
-    }
-    int upperBound = Math.min(maxBytes, UPPER_BOUND_BYTES);
-    int byteSize = largestTwoPowerInRange(LOWER_BOUND_BYTES, upperBound);
-    return new BlockSplitBloomFilter(byteSize);
-  }
-
-  /**
    * Constructor of block-based Bloom filter.
    *
    * @param numBytes The number of bytes for Bloom filter bitset. The range of num_bytes should be within
@@ -232,19 +216,6 @@ public class BlockSplitBloomFilter implements BloomFilter {
     }
     this.bitset = new byte[numBytes];
     this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.LITTLE_ENDIAN).asIntBuffer();
-  }
-
-  /**
-   * Calculate the largest power of 2 in the range [lowerBound,upperBound].
-   */
-  public static int largestTwoPowerInRange(int lowerBound, int upperBound) {
-    int result = upperBound;
-    // if `upperBound` is not power of 2, get the next largest power of two less than `upperBound`
-    if ((result & (result - 1)) != 0) {
-      result = Integer.highestOneBit(result);
-    }
-    result = Math.max(result, lowerBound);
-    return result;
   }
 
   @Override
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -97,7 +97,7 @@ public class BlockSplitBloomFilter implements BloomFilter {
       maxBytes = LOWER_BOUND_BYTES;
     }
     int upperBound = Math.min(maxBytes, UPPER_BOUND_BYTES);
-    int byteSize = largestTwoPowerNumberInRange(LOWER_BOUND_BYTES, upperBound);
+    int byteSize = largestTwoPowerInRange(LOWER_BOUND_BYTES, upperBound);
     return new BlockSplitBloomFilter(byteSize);
   }
 
@@ -237,15 +237,13 @@ public class BlockSplitBloomFilter implements BloomFilter {
   /**
    * Calculate the largest power of 2 in the range [lowerBound,upperBound].
    */
-  public static int largestTwoPowerNumberInRange(int lowerBound, int upperBound) {
+  public static int largestTwoPowerInRange(int lowerBound, int upperBound) {
     int result = upperBound;
     // if `upperBound` is not power of 2, get the next largest power of two less than `upperBound`
     if ((result & (result - 1)) != 0) {
       result = Integer.highestOneBit(result);
     }
-    if (result < lowerBound) {
-      result = lowerBound;
-    }
+    result = Math.max(result, lowerBound);
     return result;
   }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
@@ -86,6 +86,22 @@ public class BlockSplitBloomFilter implements BloomFilter {
     0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
 
   /**
+   * Create a Bloom filter within the maximum bytes size. The bloom filter bytes size
+   * will be a largest power of 2 less than the maximum bytes size.
+   *
+   * @param maxBytes The max number of bytes for Bloom filter.
+   * @return
+   */
+  public static BlockSplitBloomFilter of(int maxBytes) {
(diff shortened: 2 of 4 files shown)


yabola:
@wgtmac @ggershinsky If you have time, please take a look, thank you

yabola:
@wgtmac @gszadovszky If you have time, please take a look, thank you

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -85,6 +85,22 @@ public class BlockSplitBloomFilter implements BloomFilter {
...
     0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
 
+  /**
+   * Create a Bloom filter within the maximum bytes size. The bloom filter bytes size
+   * will be a largest power of 2 less than the maximum bytes size.
+   *
+   * @param maxBytes The max number of bytes for Bloom filter.
+   * @return
+   */
+  public static BlockSplitBloomFilter of(int maxBytes) {
```
wgtmac, line 95: Why adding a new function but not modify the constructor directly? It still has the risk if user uses the old constructor.

parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java
```diff
@@ -218,6 +234,19 @@ private void initBitset(int numBytes) {
     this.intBuffer = ByteBuffer.wrap(bitset).order(ByteOrder.LITTLE_ENDIAN).asIntBuffer();
   }
 
+  /**
+   * Calculate the largest power of 2 in the range [lowerBound,upperBound].
+   */
+  public static int largestTwoPowerInRange(int lowerBound, int upperBound) {
```
wgtmac, line 240:
```suggestion
  private static int largestPowerOfTwoInRange(int lowerBound, int upperBound) {
```

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -97,7 +97,7 @@ abstract class ColumnWriterBase implements ColumnWriter {
       int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
       this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
     } else {
-      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize);
+      this.bloomFilter = BlockSplitBloomFilter.of(maxBloomFilterSize);
```
wgtmac, line 100: I found that there is a constructor doing the similar thing:
```java
  /**
   * Constructor of block-based Bloom filter.
   *
   * @param numBytes The number of bytes for Bloom filter bitset. The range of num_bytes should be within
   *                 [DEFAULT_MINIMUM_BYTES, maximumBytes], it will be rounded up/down
   *                 to lower/upper bound if num_bytes is out of range. It will also be rounded up to a power
   *                 of 2. It uses XXH64 as its default hash function.
   * @param maximumBytes The maximum bytes of the Bloom filter.
   */
  public BlockSplitBloomFilter(int numBytes, int maximumBytes) {
    this(numBytes, LOWER_BOUND_BYTES, maximumBytes, HashStrategy.XXH64);
  }
```

wgtmac, line 100:
```suggestion
      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
```

wgtmac: Does this change solve the problem?

yabola: Yes, but it guarantees the maximum bytes size, but it cannot be guaranteed to be a power of 2. I am afraid that there will be a problem here.

yabola: If I understand something wrong, please correct me. 
In the existing implementation, the value of `numBytes` can be adjusted up to a power of 2 as long as it does not exceed `maximumBytes`(probably for performance purposes)
But there is no guarantee that `maximumBytes` must be a power of 2, which is quite strange. My existing implementation guarantees that `numBytes` is the size of the power of 2 as much as possible and don't break this rule.

wgtmac: https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L146

It tries to find the next power of two and also guarantees not to exceed the max size.

BTW, the specs does not require the size to be power of two. Is there any issue you have seen if the size is not power of two?

yabola: If I understand something wrong, please correct me. I found this part of code, need `bitset.length / BYTES_PER_BLOCK` result to be integer, need power of 2, extra bit is useless, but also no error.
https://github.com/apache/parquet-mr/blob/1235003e742e6a76bf6cb8f7ed33e942fa12d0d5/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L245-L258

wgtmac: OK, I got your point. According to the delegate constructor below, `LOWER_BOUND_BYTES` and `UPPER_BOUND_BYTES` are hard limits. What about checking if `minimumBytes` and `maximumBytes` are power of two and throw if not?

```java
  public BlockSplitBloomFilter(int numBytes, int minimumBytes, int maximumBytes, HashStrategy hashStrategy) {
    if (minimumBytes > maximumBytes) {
      throw new IllegalArgumentException("the minimum bytes should be less or equal than maximum bytes");
    }

    if (minimumBytes > LOWER_BOUND_BYTES && minimumBytes < UPPER_BOUND_BYTES) {
      this.minimumBytes = minimumBytes;
    }

    if (maximumBytes > LOWER_BOUND_BYTES && maximumBytes < UPPER_BOUND_BYTES) {
      this.maximumBytes = maximumBytes;
    }

    initBitset(numBytes);

    cacheBuffer.order(ByteOrder.LITTLE_ENDIAN);

    switch (hashStrategy) {
      case XXH64:
        this.hashStrategy = hashStrategy;
        hashFunction = new XxHash();
        break;
      default:
        throw new RuntimeException("Unsupported hash strategy");
    }
  }
```

yabola: I fear there will be forward compatibility issues if throw exception when they are not power of two.
Actually here if the bitSet size is not a power of 2, there will be no error, but some storage is wasted.

wgtmac: I think that is fine. It is well explained in the comment of the constructor. WDYT @gszadovszky

yabola: > If I understand something wrong, please correct me. I found this part of code, need `bitset.length / BYTES_PER_BLOCK` result to be integer, need power of 2, extra bit is useless, but also no error.

@wgtmac @ggershinsky Sorry, I may have misunderstood. `bitset.length` is valid as long as it is a multiple of `BYTES_PER_BLOCK`(32) (not necessarily a power of 2) . 
The reason for the power of 2 here is more complicated, I think there are performance reasons. 
Most of the existing parquet uses need to specify the NDV value or default maxBytes 1MB. In these cases, `numbytes` will be always initialized to a power of 2.
@chenjunjiedada could you take a look at this? thank you.

yabola: > If I understand something wrong, please correct me. I found this part of code, need `bitset.length / BYTES_PER_BLOCK` result to be integer, need power of 2, extra bit is useless, but also no error.

@wgtmac @gszadovszky  Sorry, I may have misunderstood. `bitset.length` is valid as long as it is a multiple of `BYTES_PER_BLOCK`(32) (not necessarily a power of 2) . 
The reason for the power of 2 here is more complicated, I think there are performance reasons. 
Most of the existing parquet uses need to specify the NDV value or default maxBytes 1MB. In these cases, `numbytes` will be always initialized to a power of 2.
@chenjunjiedada could you take a look at this? thank you.

yabola: I refer to the documentation here https://github.com/apache/parquet-format/blob/master/BloomFilter.md#technical-approach and learned that the implementation in parquet is based on the paper http://algo2.iti.kit.edu/documents/cacheefficientbloomfilters-jea.pdf 
The power of 2 is probably for performance requirements 

Impala has the same implementation but i'm not familiar with it... I saw similar implementation logic here.
https://github.com/apache/impala/blob/2c779939dc302be9ee5dd97ddf374bb043040891/be/src/kudu/util/block_bloom_filter.h#L88-L92

wgtmac: Yes, I agree that power of two is solely for best performance.

IMHO, fixing the delegate constructor is the cleanest way to address this. I think it is OK to break users if the actual size exceeds `parquet.bloom.filter.max.bytes` as this is a bug.

gszadovszky: I agree with @wgtmac. Fixing a defect might lead to breaking compatibility.

yabola: @wgtmac @gszadovszky sorry, I didn't understand the meaning, do you mean to use `BlockSplitBloomFilter(int numBytes, int maximumBytes)` API (even if `numBytes` is not power of 2)?

wgtmac: The goal is to fix the issue that bloom filter size can be larger than `parquet.bloom.filter.max.bytes`. The problematic constructors used are as below:

```java
    int maxBloomFilterSize = props.getMaxBloomFilterBytes();

    OptionalLong ndv = props.getBloomFilterNDV(path);
    OptionalDouble fpp = props.getBloomFilterFPP(path);
    // If user specify the column NDV, we construct Bloom filter from it.
    if (ndv.isPresent()) {
      int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
      this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
    } else {
      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize);
    }
```

They are all delegated by `public BlockSplitBloomFilter(int numBytes, int minimumBytes, int maximumBytes, HashStrategy hashStrategy)`. 
```java
  public BlockSplitBloomFilter(int numBytes) {
    this(numBytes, LOWER_BOUND_BYTES, UPPER_BOUND_BYTES, HashStrategy.XXH64);
  }

  public BlockSplitBloomFilter(int numBytes, int maximumBytes) {
    this(numBytes, LOWER_BOUND_BYTES, maximumBytes, HashStrategy.XXH64);
  }
```

So it seems to me that the issue comes from `public BlockSplitBloomFilter(int numBytes)` where `maxBloomFilterSize` is passed to `numBytes` and the actual size can be adjusted to values as large as `UPPER_BOUND_BYTES`.

Is it possible to fix it like below?
```java
  public BlockSplitBloomFilter(int numBytes) {
    this(numBytes, LOWER_BOUND_BYTES, numBytes, HashStrategy.XXH64);
  }
```

yabola: @wgtmac Thank you for the clear explanation! 
I worry that if we modify it like this, it will reduces the ability of one constructor, that is, the initial size is `numbytes`, but the maximum does not exceed `UPPER_BOUND_BYTES`.
In fact, I think it might be a problem of calling the constructor incorrectly.  As you said before, we can change to call `BlockSplitBloomFilter(int numBytes, int maximumBytes)` instead of `BlockSplitBloomFilter(int numBytes)`. 
What do you think about it?

wgtmac: I agree with you. @yabola

wgtmac (approved):
LGTM. Thanks @yabola!

PR #1046, opened 2023-04-04
PARQUET-2263: Upgrade maven-shade-plugin to 3.4.1

wangyum:
Upgrade maven-shade-plugin to 3.4.1 to fix it does not work under Java 17:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (default) on project parquet-format-structures: Error creating shaded jar: null: IllegalArgumentException -> [Help 1]
```
Please see MSHADE-289 for more details.

### DIFF
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -54,11 +54,6 @@
       <version>${avro.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-      <version>${slf4j.version}</version>
-    </dependency>
-    <dependency>
       <groupId>it.unimi.dsi</groupId>
       <artifactId>fastutil</artifactId>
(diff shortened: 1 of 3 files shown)


wangyum:
cc @gszadovszky

gszadovszky:
I am fine with this change as is but I guess, the shade plugin version comes from the apache parqent pom. Don't you think it would worth to upgrade that instead?

Fokko:
I ran into the same issue in https://github.com/apache/parquet-mr/pull/1052. Looks like the latest pom ships with `maven-shade-plugin` with `3.4.1`: https://repo1.maven.org/maven2/org/apache/apache/29/apache-29.pom

Fokko:
@wangyum and @gszadovszky I'm moving this forward if you're okay with that. Thanks @wangyum for opening this PR! 👏🏻

PR #1049, opened 2023-04-11
PARQUET-2265: Don't set default Model in AvroParquetWriter

clairemcginty:
This PR removes the default `model` setting in `AvroParquetWriter.Builder`. The reasons for this change are:

1) `AvroWriteSupport` already sets a default `model` value (also to `SpecificDataSupplier`) in [AvroWriteSupport::init](https://github.com/apache/parquet-mr/blob/d38044f5395494e1543581a4b763f624305d3022/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L126-L135), so setting a default in `AvroParquetWriter` is redundant. 

2) It created some confusion for me, since I expected that the class I set in [parquet.avro.write.data.supplier](https://github.com/apache/parquet-mr/blob/d38044f5395494e1543581a4b763f624305d3022/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L54) would automatically be used as the `model`. This change ensures that if no data model is supplied in the `AvroParquetWriter.Builder`, then `AvroWriteSupport` will select a data model based on the `parquet.avro.write.data.supplier` config value, if present, and `SpecificDataSupplier` if not present.

3. `AvroParquetReader` [sets](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetReader.java#L133) `model = null`, so this PR makes the behavior of AvroParquetReader|Writer consistent.

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2265

### Tests

- [x] **My PR adds the following unit tests** __OR__ does not need testing for this extremely good reason:
     - Added a test to verify that AvroParquetWriter allows AvroWriteSupport to parse model from Configuration if .withDataModel is not set
     - Since AvroWriteSupport still sets `model` to a default value of `SpecificDataSupplier` in [init](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L403-L406), the existing SpecificRecord writer tests continuing to pass should effectively test that the default behavior is unchanged

### Commits

### Documentation

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetWriter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetWriter.java
@@ -160,7 +160,7 @@ public class AvroParquetWriter<T> extends ParquetWriter<T> {
 
   public static class Builder<T> extends ParquetWriter.Builder<T, Builder<T>> {
     private Schema schema = null;
-    private GenericData model = SpecificData.get();
+    private GenericData model = null;
 
     private Builder(Path file) {
       super(file);
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
@@ -27,6 +27,8 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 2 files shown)


wgtmac (approved):
The fix makes sense. LGTM. cc @gszadovszky

gszadovszky (approved):
LGTM

PR #1051, opened 2023-04-13
PARQUET-2267: Add dependabot to update dependencies

Fokko:
Make life easier, and makes sure that we're using the latest dependencies.

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,32 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,32 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
wgtmac, line 22: Thanks for adding this!

I just checked the Apache Orc has added a set of `ignores` here: https://github.com/apache/orc/blob/main/.github/dependabot.yml

I am not sure if we should do similar things to decrease the review burden.

cc @dongjoon-hyun

dongjoon-hyun: Yes, after enabling this, you can exclude one by one later if needed. I'm +1 for enabling.

wgtmac: I agree we can exclude one by one later. Thanks @dongjoon-hyun !

wgtmac (approved):

PR #1075, opened 2023-04-19
PARQUET-2289: Avoid using `hasCapability`

Fokko:
We should avoid using `hasCapability` to maintain compatibility with Hadoop 2. I think we can remove the whole check since Hadoop 1 isn't supported anymore.

This allows us to use Hadoop 2. Relates to https://github.com/apache/parquet-mr/pull/951 

cc @steveloughran 

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H1SeekableInputStream.java
@@ -26,7 +26,9 @@ import java.io.IOException;
 /**
  * SeekableInputStream implementation that implements read(ByteBuffer) for
  * Hadoop 1 FSDataInputStream.
+ * @deprecated will be removed in 2.0.0.
  */
+@Deprecated
 class H1SeekableInputStream extends DelegatingSeekableInputStream {
 
   private final FSDataInputStream stream;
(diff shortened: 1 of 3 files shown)


steveloughran:
StreamCapabilities was added to hadoop branch 2 *six years ago* in https://issues.apache.org/jira/browse/HDFS-11644 and shipped in hadoop-2.9.0

anyone complaining about it not being there is trying to run a version of hadoop which predates this, and is not the version of hadoop-2 parquet compiles with. There is no guarantee *anything* works with that older version: method signatures, dependencies. etc.

oh, and if you really want to claim hadoop 2.8.x support, you have to compile with java7 too, as hadoop 2.9 is the first java8 compatible release

 
If you really want to support it, you need to modify the build to
* compile with java lang == 7
* require a java7 jdk
* build against hadoop 2.8

or you tell the user complaining to upgrade to a release of hadoop which has shipped recently.

Fokko:
@steveloughran thanks for the insight. I thought this would be rather trivial, but this clearly isn't the case. Adding support for Java 7 is something that doesn't make any sense. Let me close this for now.

steveloughran:
aah, no worries. i just despair when people ask for things backported to 5+ year old releases because they don't want the hassle of upgrading their own installation, e.g [HADOOP-18600](https://issues.apache.org/jira/browse/HADOOP-18600). My PoV there is "you are free to fork your own versions of things and maintain them" -the maintenance being what is expected.

Whoever doesn't upgrade also gets to maintain their CVE fix process, both from hadoop and all the dependencies, so I would use the latest hadoop 2.9.x release, as at least there we fix all the hadoop source fixes we can, even though transitive JAR updates are out of scope. A six year old release has 6 years worth of security related JAR updates to worry about.

PR #1078, opened 2023-04-21
PARQUET-2292: Default SpecificRecord model reflects from MODEL$ field

clairemcginty:
Context:

AvroWriteSupport/AvroReadSupport can improve the precision of their default `model` selection. Currently they default to new SpecificDataSupplier().get()[0]. This means that SpecificRecord classes that contain logical types will fail out-of-the-box unless a specific DATA_SUPPLIER is configured that contains logical type conversions.

This PR attempts to derive a `model` by reflecting the `MODEL$` field from the SpecificRecordBase implementation, which in Avro 1.11 comes pre-loaded with all necessary logical type conversions for that Specific Avro class. (For Avro 1.8 compatibility, we have to add the conversions manually from the `conversions` field.)

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -74,6 +74,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-annotations</artifactId>
+      <version>${hadoop.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
@@ -95,6 +101,7 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
       <version>${slf4j.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.parquet</groupId>
@@ -104,8 +111,20 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.23.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.powermock</groupId>
       <artifactId>powermock-module-junit4</artifactId>
+      <version>${powermock.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-core</artifactId>
       <version>${powermock.version}</version>
       <scope>test</scope>
     </dependency>
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -77,7 +77,6 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
@@ -96,7 +95,6 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
       <version>${slf4j.version}</version>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.parquet</groupId>
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -105,6 +105,18 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>${powermock.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito2</artifactId>
+      <version>${powermock.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/parquet-avro/pom.xml
+++ b/parquet-avro/pom.xml
@@ -74,12 +74,6 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-annotations</artifactId>
-      <version>${hadoop.version}</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -134,7 +134,7 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
       avroSchema = new AvroSchemaConverter(configuration).convert(parquetSchema);
     }
 
-    GenericData model = getDataModel(configuration);
+    GenericData model = getDataModel(configuration, avroSchema);
     String compatEnabled = metadata.get(AvroReadSupport.AVRO_COMPATIBILITY);
     if (compatEnabled != null && Boolean.valueOf(compatEnabled)) {
       return newCompatMaterializer(parquetSchema, avroSchema, model);
@@ -149,12 +149,22 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
         parquetSchema, avroSchema, model);
   }
 
-  private GenericData getDataModel(Configuration conf) {
+  private GenericData getDataModel(Configuration conf, Schema schema) {
     if (model != null) {
       return model;
     }
+
+    if (conf.get(AVRO_DATA_SUPPLIER) == null) {
+      final GenericData modelForSchema = AvroRecordConverter.getModelForSchema(schema);
+
+      if (modelForSchema != null) {
+        return modelForSchema;
+      }
+    }
+
     Class<? extends AvroDataSupplier> suppClass = conf.getClass(
-        AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
-    return ReflectionUtils.newInstance(suppClass, conf).get();
+      AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
+
+      return ReflectionUtils.newInstance(suppClass, conf).get();
   }
 }
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -27,6 +27,8 @@ import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.parquet.hadoop.api.ReadSupport;
 import org.apache.parquet.io.api.RecordMaterializer;
 import org.apache.parquet.schema.MessageType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Avro implementation of {@link ReadSupport} for avro generic, specific, and
@@ -36,6 +38,8 @@ import org.apache.parquet.schema.MessageType;
  * @param <T> the Java type of records created by this ReadSupport
  */
 public class AvroReadSupport<T> extends ReadSupport<T> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroReadSupport.class);
 
   public static String AVRO_REQUESTED_PROJECTION = "parquet.avro.projection";
   private static final String AVRO_READ_SCHEMA = "parquet.avro.read.schema";
@@ -155,7 +159,14 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
     }
 
     if (conf.get(AVRO_DATA_SUPPLIER) == null && schema != null) {
-      final GenericData modelForSchema = AvroRecordConverter.getModelForSchema(schema);
+      GenericData modelForSchema;
+      try {
+        modelForSchema = AvroRecordConverter.getModelForSchema(schema);
+      } catch (Exception e) {
+        LOG.warn(String.format("Failed to derive data model for Avro schema %s. Parquet will use default " +
+          "SpecificData model for reading from source.", schema), e);
+        modelForSchema = null;
+      }
 
       if (modelForSchema != null) {
         return modelForSchema;
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -154,7 +154,7 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
       return model;
     }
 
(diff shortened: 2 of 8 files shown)


parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
     }
   }
 
+  /**
+   * Returns the specific data model for a given SpecificRecord schema by reflecting the underlying
+   * Avro class's `MODEL$` field, or Null if the class is not on the classpath or reflection fails.
+   */
+  static SpecificData getModelForSchema(Schema schema) {
```
clairemcginty, line 179: The reflection logic here is all based on the existing Avro behavior for its model cache: https://github.com/apache/avro/blob/release-1.11.1/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L76-L86

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
...
+
+      model = (SpecificData) modelField.get(null);
+    } catch (Exception e) {
+      return null;
+    }
+
+    try {
+      final String avroVersion = Schema.Parser.class.getPackage().getImplementationVersion();
+      // Avro 1.8 doesn't include conversions in the MODEL$ field
+      if (avroVersion.startsWith("1.8.")) {
```
clairemcginty, line 201: For reference, Avro 1.11 compiles SpecificRecord implementations that look like:

```java
@org.apache.avro.specific.AvroGenerated
public class LogicalTypesTest extends org.apache.avro.specific.SpecificRecordBase implements org.apache.avro.specific.SpecificRecord {
  ...
  private static final SpecificData MODEL$ = new SpecificData();
  static {
    MODEL$.addLogicalTypeConversion(new org.apache.avro.data.TimeConversions.TimestampMillisConversion());
  }
}
```

Whereas for Avro 1.8, there is no static initialization, but there _is_ a `conversions` field:

```java
@org.apache.avro.specific.AvroGenerated
public class LogicalTypesTest extends org.apache.avro.specific.SpecificRecordBase implements org.apache.avro.specific.SpecificRecord {
  ...
  private static final SpecificData MODEL$ = new SpecificData();
  
  protected static final org.apache.avro.data.TimeConversions.TimestampConversion TIMESTAMP_CONVERSION = new org.apache.avro.data.TimeConversions.TimestampConversion();

  private static final org.apache.avro.Conversion<?>[] conversions =
      new org.apache.avro.Conversion<?>[] {
      TIMESTAMP_CONVERSION,
      null
  };
}
```

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -400,9 +400,18 @@ private Binary fromAvroString(Object value) {
     return Binary.fromCharSequence(value.toString());
   }
 
-  private static GenericData getDataModel(Configuration conf) {
+  private static GenericData getDataModel(Configuration conf, Schema schema) {
+    if (conf.get(AVRO_DATA_SUPPLIER) == null) {
+      final GenericData modelForSchema = AvroRecordConverter.getModelForSchema(schema);
```
wgtmac, line 405: What if `schema == null`?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
```diff
@@ -149,12 +149,22 @@ private static <T> RecordMaterializer<T> newCompatMaterializer(
...
+        return modelForSchema;
+      }
+    }
+
     Class<? extends AvroDataSupplier> suppClass = conf.getClass(
-        AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
-    return ReflectionUtils.newInstance(suppClass, conf).get();
+      AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
+
+      return ReflectionUtils.newInstance(suppClass, conf).get();
```
wgtmac, line 168: Could you please revert these format changes?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
...
+    } else {
+      return null;
+    }
+
+    final SpecificData model;
+    try {
+      final Field modelField = clazz.getDeclaredField("MODEL$");
+      modelField.setAccessible(true);
+
+      model = (SpecificData) modelField.get(null);
```
wgtmac, line 193: Should it return if model is set successfully?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
wgtmac: @gszadovszky Could you help review this? I am not that familiar with Avro.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
clairemcginty: From where it's called in the code, `schema` should never be null -- but I guess it would be safest to add a null-check here to future-proof it a bit. Added!

parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
clairemcginty: done, sorry about that!

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
...
+    } catch (Exception e) {
+      return null;
+    }
+
+    try {
+      final String avroVersion = Schema.Parser.class.getPackage().getImplementationVersion();
```
RustedBones, line 199: This won't work for avro pre 1.10 where the implementation is not defined, and will return `null`
Here is a [trick](https://github.com/sbt/sbt-avro/blob/ec5f8aa064d0314a97a876bfe1d2d1bdb26dcbb8/src/main/scala/com/github/sbt/avro/SbtAvro.scala#L28) to access the avro version for previous artifact

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
clairemcginty: sorry, can you expand on this a bit?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
...
+      return null;
+    }
+
+    final SpecificData model;
+    try {
+      final Field modelField = clazz.getDeclaredField("MODEL$");
+      modelField.setAccessible(true);
+
+      model = (SpecificData) modelField.get(null);
+    } catch (Exception e) {
```
gszadovszky, line 194: I think, we should log this exception. I would be also nice to use specific exceptions instead of catching everything. WDYT?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
gszadovszky: TBH I do not have a strong opinion on any. I am fine with the current one if it works.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -169,6 +172,46 @@ public void add(Object value) {
...
+      final String avroVersion = Schema.Parser.class.getPackage().getImplementationVersion();
+      // Avro 1.8 doesn't include conversions in the MODEL$ field
+      if (avroVersion.startsWith("1.8.")) {
+        final Field conversionsField = clazz.getDeclaredField("conversions");
+        conversionsField.setAccessible(true);
+
+        final Conversion<?>[] conversions = (Conversion<?>[]) conversionsField.get(null);
+        Arrays.stream(conversions).filter(Objects::nonNull).forEach(model::addLogicalTypeConversion);
+      }
+    } catch (Exception e) {
```
gszadovszky, line 208: Similar to the previous comment about logging and catching specific exceptions.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
gszadovszky: Since we are using reflections on private members there are no compatibility guarantees. We shall be very careful here. What about avro versions prior to 1.8? Also, what if it breaks in the future? Will the related unit test fail for a future Avro releases (in case of upgrading the Avro version in the pom)?

parquet-avro/src/test/java/org/apache/parquet/avro/TestSpecificReadWrite.java
```diff
@@ -237,6 +242,43 @@ public void testAvroReadSchema() throws IOException {
     }
   }
 
+  @Test
```
gszadovszky, line 245: We need to test both of `getModelForSchema` related to the avro version. If the version check gets more complicated, maybe more versions are to cover.

gszadovszky (changes requested):
Sorry, @clairemcginty for the late response. I've added some comments.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
clairemcginty: yeah - it's unfortunate that this isn't solvable without reflection. I've manually tested this with earlier versions of Avro... let me see if I can reshape them into automated tests for `parquet-avro` 👀

clairemcginty: ok, I've added a new test suite, `TestAvroRecordConverter`, that thoroughly tests `getModelForSchema` with a variety of schema variants and with Avro versions 1.{7,8,9,10,11}. To do this, I used Powermock to mock the static invocation used to get Avro runtime version, then I ad-hoc compiled Avro generated classes using compiler versions 1.{7,8,9,10,11}. Happy to take a different approach with testing if you prefer -- this just seemed like the most straightforward way to do it :)

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
clairemcginty: Added clearer logging & null-/error-handling.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
clairemcginty: > Since we are using reflections on private members there are no compatibility guarantees. We shall be very careful here. What about avro versions prior to 1.8? Also, what if it breaks in the future? Will the related unit test fail for a future Avro releases (in case of upgrading the Avro version in the pom)?

so I've tested 1.7 and 1.8; since 1.9 Avro has stably used the `MODEL$` field to hold all conversions, so I feel reasonably confident about relying on this. If that changes, we'll catch it in the new unit tests 👍 

If you want, I can surround invocations of `getModelForSchema` in a try/catch (in `AvroReadSupport`/`AvroWriteSupport`), and just use the default SpecificDataSupplier if they throw anything. That way any unexpected behavior would just result in logical types not being used.

parquet-avro/pom.xml
```diff
@@ -105,6 +110,30 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.23.0</version>
```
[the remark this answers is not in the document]
wgtmac: I think this is fine. BTW, is it compatible to upgrade this in the root pom as well?

wgtmac (commented):
Thanks @gszadovszky for the review!

wgtmac:
@clairemcginty Could you make the CIs happy?

parquet-avro/pom.xml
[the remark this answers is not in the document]
clairemcginty: It should be, but it does require an artifact migration in a few modules (parquet-column and parquet-hadoop depend on mockito-all, which has moved to mockito-core in 2.x). I would be happy to do it, just not sure if it's in scope for this PR :)

clairemcginty:
@wgtmac, sorry about that -- `mvn verify` was happy on my machine, but I wasn't running `install`. Should be good now!

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
gszadovszky: > so I've tested 1.7 and 1.8; since 1.9 Avro has stably used the `MODEL$` field to hold all conversions, so I feel reasonably confident about relying on this. If that changes, we'll catch it in the new unit tests +1
This sounds perfect to me. Thanks a lot for the additional work!

> If you want, I can surround invocations of `getModelForSchema` in a try/catch (in `AvroReadSupport`/`AvroWriteSupport`), and just use the default SpecificDataSupplier if they throw anything. That way any unexpected behavior would just result in logical types not being used.
Yes, I think this fallback mechanism sounds reasonable to me.

clairemcginty: Great! I added code wrapping `getModelForSchema` in a try/catch at each call site.

gszadovszky (approved):
Thank you, @clairemcginty for your efforts. It looks good to me.

clairemcginty:
thanks for the review @gszadovszky! very excited to start using this in Parquet :)

gszadovszky:
I wouldn't like to make you sad, @clairemcginty, but we just released `1.13.0` last month and the previous one was almost a year ago. Meanwhile, nothing says we cannot do releases for tiny features, it is more about the effort and who has the time to do it. I usually don't, unfortunately.
Maybe if you ask nicely, @wgtmac would do another one... :wink:

clairemcginty:
Oh no, I'm sorry I missed the boat! 😭 Is there anything that I (or my organization) could do to help out with the release, @wgtmac? If it helps at all, having this fix released would drastically help us drive Parquet adoption, and lead to more OSS contributions down the line :)

wgtmac:
The parquet community is discussing a new `1.13.1` release to address some issues introduced by `1.13.0`: https://lists.apache.org/thread/1mjvdcmwqjcblmfkfgpd9ob2yodx7tom

As the release manager, what do you think if we include this PR to the `1.13.1` release? @Fokko

gszadovszky:
@clairemcginty, could you reply on the thread that you would like to have this one included? It is better to have a broader audience for discussing the release content. (Note: We should usually not include features in patch releases to lower the potential risks. But it is up to the community to decide.)

clairemcginty:
@gszadovszky I wasn't previously subscribed to the dev mailing list, so I just subscribed -- I think I need to wait until someone pings the thread for it to land in my inbox so I can reply :)

gszadovszky:
@clairemcginty, I think, there is an option to reply from you own mail client in ponymail (the thread link above).

clairemcginty:
thanks @gszadovszky! I used the Ponymail link from the thread, but I fear it's just created a new email thread - apologies

Fokko:
Thanks for pining me here @wgtmac I just replied on the dev-list. I think we can include this one.

PR #1084, opened 2023-04-29
PARQUET-2276: Bring back support for Hadoop 2.7.3

Fokko:

Complements the work of https://github.com/apache/parquet-mr/pull/951. This PR removes the breaking change that caused Parquet to be incompatible with Hadoop <2.9. In this change, we dynamically check if the `hasCapabilities` method is available, and if this is the case, it will use it. Otherwise, it will fall back to the previous implementation, with the addition that it will also check the wrapped streams.

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2276

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
@@ -67,7 +67,7 @@ public class HadoopStreams {
       }
     }
 
-    return isWrappedStreamByteBufferReadableLegacy(stream);
+    return unwrapByteBufferReadableLegacy(stream);
   }
 
   /**
@@ -80,11 +80,11 @@ public class HadoopStreams {
    * @param stream stream to probe
    * @return A H2SeekableInputStream to access, or H1SeekableInputStream if the stream is not seekable
    */
-  private static SeekableInputStream isWrappedStreamByteBufferReadableLegacy(FSDataInputStream stream) {
+  private static SeekableInputStream unwrapByteBufferReadableLegacy(FSDataInputStream stream) {
     InputStream wrapped = stream.getWrappedStream();
     if (wrapped instanceof FSDataInputStream) {
       LOG.debug("Checking on wrapped stream {} of {} whether is ByteBufferReadable", wrapped, stream);
-      return isWrappedStreamByteBufferReadableLegacy(((FSDataInputStream) wrapped));
+      return unwrapByteBufferReadableLegacy(((FSDataInputStream) wrapped));
     }
     if (stream.getWrappedStream() instanceof ByteBufferReadable) {
       return new H2SeekableInputStream(stream);
@@ -117,9 +117,9 @@ public class HadoopStreams {
       return null;
     }
 
-    Boolean hasCapabilities = hasCapabilitiesMethod.invoke(stream, "in:readbytebuffer");
+    boolean isByteBufferReadable = hasCapabilitiesMethod.invoke(stream, "in:readbytebuffer");
 
-    if (hasCapabilities) {
+    if (isByteBufferReadable) {
       // stream is issuing the guarantee that it implements the
       // API. Holds for all implementations in hadoop-*
       // since Hadoop 3.3.0 (HDFS-14111).
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
@@ -20,6 +20,7 @@
 package org.apache.parquet.hadoop.util;
 
 import org.apache.hadoop.fs.ByteBufferReadable;
+import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.parquet.io.PositionOutputStream;
@@ -27,7 +28,10 @@ import org.apache.parquet.io.SeekableInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.io.InputStream;
+import java.lang.reflect.Method;
 import java.util.Objects;
 
 /**
@@ -36,6 +40,39 @@ import java.util.Objects;
 public class HadoopStreams {
 
   private static final Logger LOG = LoggerFactory.getLogger(HadoopStreams.class);
+
+  private static final Class<?> byteBufferReadableClass = getReadableClass();
+  static final Constructor<SeekableInputStream> h2SeekableConstructor = getH2SeekableConstructor();
+
+  private static Class<?> getReadableClass() {
+    try {
+      return Class.forName("org.apache.hadoop.fs.ByteBufferReadable");
+    } catch (ClassNotFoundException | NoClassDefFoundError e) {
+      return null;
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Class<SeekableInputStream> getH2SeekableClass() {
+    try {
+      return (Class<SeekableInputStream>) Class.forName(
+        "org.apache.parquet.hadoop.util.H2SeekableInputStream");
+    } catch (ClassNotFoundException | NoClassDefFoundError e) {
+      return null;
+    }
+  }
+
+  private static Constructor<SeekableInputStream> getH2SeekableConstructor() {
+    Class<SeekableInputStream> h2SeekableClass = getH2SeekableClass();
+    if (h2SeekableClass != null) {
+      try {
+        return h2SeekableClass.getConstructor(FSDataInputStream.class);
+      } catch (NoSuchMethodException e) {
+        return null;
+      }
+    }
+    return null;
+  }
 
   /**
    * Wraps a {@link FSDataInputStream} in a {@link SeekableInputStream}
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -37,6 +41,39 @@ public class HadoopStreams {
 
   private static final Logger LOG = LoggerFactory.getLogger(HadoopStreams.class);
 
+  private static final Class<?> byteBufferReadableClass = getReadableClass();
+  static final Constructor<SeekableInputStream> h2SeekableConstructor = getH2SeekableConstructor();
```
rdblue, line 45: Rather than using two static methods, you can use `DynConstructors` instead to make this one expression and reduce error handling code:

```java
private static final DynConstructors.Ctor<SeekableInputStream> h2streamCtor =
    DynConstructors.Builder(SeekableInputStream.class)
    .impl("org.apache.parquet.hadoop.util.H2SeekableInputStream", FSDataInputStream.class)
    .orNull()
    .build()

...
if (h2streamCtor != null) {
  return h2streamCtor.newInstance(stream);
}
```

rdblue: Looks like you'd need to add the `orNull` handling. I don't see it here: https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/util/DynConstructors.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,41 +83,91 @@ public class HadoopStreams {
...
-  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
-    if (stream.hasCapability("in:readbytebuffer")) {
-      // stream is issuing the guarantee that it implements the
-      // API. Holds for all implementations in hadoop-*
-      // since Hadoop 3.3.0 (HDFS-14111).
-      return true;
+  private static Boolean isWrappedStreamByteBufferReadableHasCapabilities(FSDataInputStream stream) {
+    Method methodHasCapabilities;
+    try {
+      methodHasCapabilities = stream.getClass().getMethod("hasCapability", String.class);
```
rdblue, line 153: You can use DynMethods to get an unbound `hasCapability` method. That can be done statically, so all you need to do is check whether it is present and call it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
rdblue: Huh, I guess this was how it was before? Nevermind on the refactoring then.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,41 +83,91 @@ public class HadoopStreams {
...
+   * This logic is only used for Hadoop <2.9.x, and <3.x.x
+   *
+   * @param stream stream to probe
+   * @return A H2SeekableInputStream to access, or H1SeekableInputStream if the stream is not seekable
+   */
+  private static SeekableInputStream isWrappedStreamByteBufferReadableLegacy(FSDataInputStream stream) {
+    InputStream wrapped = stream.getWrappedStream();
+    if (wrapped instanceof FSDataInputStream) {
+      LOG.debug("Checking on wrapped stream {} of {} whether is ByteBufferReadable", wrapped, stream);
+      return isWrappedStreamByteBufferReadableLegacy(((FSDataInputStream) wrapped));
```
rdblue, line 116: Why would a FSDataInputStream have another one inside?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,41 +83,91 @@ public class HadoopStreams {
...
+    }
+    try {
+      if ((Boolean) methodHasCapabilities.invoke(stream, "in:readbytebuffer")) {
+        // stream is issuing the guarantee that it implements the
+        // API. Holds for all implementations in hadoop-*
+        // since Hadoop 3.3.0 (HDFS-14111).
+        return true;
+      }
+    } catch (IllegalAccessException | InvocationTargetException e) {
+      return null;
```
rdblue, line 165: Why not return false?

rdblue (approved):
I think this should work after comparing it with older code, but it seems like there are some easy improvements to me.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
Fokko: This came from the issue from Presto: https://github.com/prestodb/presto/pull/17435

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
Fokko: Yes, I copied most from the old code to avoid refactoring. I think we can greatly simplify it because it was still taking Hadoop1 into account. We still have to check if the wrapped stream is ByteBufferReadable: https://github.com/apache/hadoop/blob/release-2.4.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FSDataInputStream.java#L142-L148

The `hasCapabilities does the same but in a more elegant way.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
Fokko: Added it

Fokko:
> I think this should work after comparing it with older code, but it seems like there are some easy improvements to me.

@rdblue I agree, I did some cleaning up. Let me know what you think.

Fokko:
FWIW, I also ran the Iceberg tests and it ran fine (except the bloom filter ones, more details [here](https://github.com/apache/iceberg/pull/7301#issuecomment-1528874719)).

steveloughran:
I repeat my stance on this: to claim hadoop 2.7 runtime compatibility you should be building against java7. if you don't, well, make clear its a fairly qualified support "hadoop 2.7.3 on java8 only" and not worry about the bits of hadoop which break if they try to do that (kerberos, s3a, anything with joda-time, ...)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
steveloughran: you can if you try hard, it's just really unusual

you can never wrap an instance by itself.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,7 +54,39 @@ public class HadoopStreams {
...
+   *
+   * This logic is only used for Hadoop <2.9.x, and <3.x.x
+   *
+   * @param stream stream to probe
+   * @return A H2SeekableInputStream to access, or H1SeekableInputStream if the stream is not seekable
+   */
+  private static SeekableInputStream isWrappedStreamByteBufferReadableLegacy(FSDataInputStream stream) {
+    InputStream wrapped = stream.getWrappedStream();
+    if (wrapped instanceof FSDataInputStream) {
+      LOG.debug("Checking on wrapped stream {} of {} whether is ByteBufferReadable", wrapped, stream);
```
shangxinli, line 86: It is good to add this debug log in case recursive methods cause an infinite loop, we can enable this logging and debug.

shangxinli (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
Fokko: Yes, I kept it in there.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -55,23 +95,31 @@ public static SeekableInputStream wrap(FSDataInputStream stream) {
...
    * FSDataInputStream will pass the probe down to the underlying stream.
    *
    * @param stream stream to probe
-   * @return true if it is safe to a H2SeekableInputStream to access the data
+   * @return true if it is safe to a H2SeekableInputStream to access
+   * the data, null when it cannot be determined because of missing hasCapabilities
    */
-  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
-    if (stream.hasCapability("in:readbytebuffer")) {
+  private static Boolean isWrappedStreamByteBufferReadableHasCapabilities(FSDataInputStream stream) {
```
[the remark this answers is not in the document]
Fokko: I agree, I've changed it to `isWrappedStreamByteBufferReadable`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,7 +54,39 @@ public class HadoopStreams {
...
+   * Is the inner stream byte buffer readable?
+   * The test is 'the stream is not FSDataInputStream
+   * and implements ByteBufferReadable'
+   *
+   * This logic is only used for Hadoop <2.9.x, and <3.x.x
+   *
+   * @param stream stream to probe
+   * @return A H2SeekableInputStream to access, or H1SeekableInputStream if the stream is not seekable
+   */
+  private static SeekableInputStream isWrappedStreamByteBufferReadableLegacy(FSDataInputStream stream) {
```
rdblue, line 83: Usually, `isSomething` methods return a boolean. This is unwrapping, so I'd prefer naming it `unwrapByteBufferReadableLegacy` or something to be more clear.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -46,7 +54,39 @@ public class HadoopStreams {
...
+    InputStream wrapped = stream.getWrappedStream();
+    if (wrapped instanceof FSDataInputStream) {
+      LOG.debug("Checking on wrapped stream {} of {} whether is ByteBufferReadable", wrapped, stream);
+      return isWrappedStreamByteBufferReadableLegacy(((FSDataInputStream) wrapped));
+    }
+    if (stream.getWrappedStream() instanceof ByteBufferReadable) {
```
rdblue, line 89: I prefer using the same whitespace conventions as in Iceberg, although that's a bit more relaxed over here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -55,23 +95,31 @@ public static SeekableInputStream wrap(FSDataInputStream stream) {
...
    */
-  private static boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
-    if (stream.hasCapability("in:readbytebuffer")) {
+  private static Boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
+    if (hasCapabilitiesMethod.isNoop()) {
+      // When the method is not available, just return a null
+      return null;
+    }
+
+    Boolean hasCapabilities = hasCapabilitiesMethod.invoke(stream, "in:readbytebuffer");
```
rdblue, line 120: Is this a boxed boolean? If so, should we update the check to handle null?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -55,23 +95,31 @@ public static SeekableInputStream wrap(FSDataInputStream stream) {
...
+
+    if (hasCapabilities) {
```
rdblue, line 122: This variable would be more clear if it were called `isByteBufferReadable` since that's the capability we are checking for.

rdblue (approved):
Merge when you're ready. I think this is correct.

Fokko:
Thanks @shangxinli, @wgtmac, @rdblue, and @steveloughran for the review.

Steve, I'm aware of your stance and I also respect it. Unfortunately, a lot of companies are still using (internally heavy patched versions) of Hadoop, and to get traction in downstream projects we still have to maintain compatibility.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
Fokko: I was assuming that it needed to be an object, but a primitive works as well, so changed that. Thanks for catching this!

PR #1089, opened 2023-05-04
PARQUET-2297: Skip delta problem check for encrypted files

ggershinsky:
https://issues.apache.org/jira/browse/PARQUET-2297

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
@@ -173,7 +173,10 @@ public class ParquetRecordReader<T> extends RecordReader<Void, T> {
       }
     }
 
-    if (!reader.getRowGroups().isEmpty()) {
+    if (!reader.getRowGroups().isEmpty() &&
+      // Encrypted files (parquet-mr 1.12+) can't have the delta encoding problem (resolved in parquet-mr 1.8)
+      reader.getFileMetaData().getEncryptionType() != FileMetaData.EncryptionType.ENCRYPTED_FOOTER &&
+      reader.getFileMetaData().getEncryptionType() != FileMetaData.EncryptionType.PLAINTEXT_FOOTER) {
       checkDeltaByteArrayProblem(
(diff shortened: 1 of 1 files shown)


wgtmac:
Should we include this fix to the next 1.13.1 release: https://lists.apache.org/thread/1mjvdcmwqjcblmfkfgpd9ob2yodx7tom ?  @ggershinsky

ggershinsky:
SGTM, I'll send a PR to the parquet-1.13.x branch too

wgtmac (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -173,7 +173,10 @@ private void initializeInternalReader(ParquetInputSplit split, Configuration con
       }
     }
 
-    if (!reader.getRowGroups().isEmpty()) {
+    if (!reader.getRowGroups().isEmpty() &&
+      // Encrypted files (parquet-mr 1.12+) can't have the delta encoding problem (resolved in parquet-mr 1.8)
```
Fokko, line 177: Could we add a test for this?

Fokko (approved):
LGTM

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
ggershinsky: - with delta encoding problem: basically impossible to reproduce :), it was resolved in 1.8
- without this problem: I've had a look at the existing unitests, unfortunately none can be used as a basis for adding a function for this particular situation. This will require building a new unitest from scratch. However, given that a) the patch is small and straightforward b) Spark stopped using this parquet read path - building a full unitest can be an overkill. But if you have a different opinion, please let me know.

Fokko: Thanks for the explanation, I'm fine with leaving out a unit test. Just curious if it would be easy to modify existing tests to make sure that we hit the code.

PR #1111, opened 2023-06-12
PARQUET-1822: Avoid requiring Hadoop installation for reading/writing

amousavigourabi:

### Jira

### Tests

### Commits

### Documentation


wernerdaehn:
I don't want to sound too greedy, but the next level of this feature would be if the classes in question have no imports of Hadoop in them.
Something like: Parquet(with Hadoop) -> Parquet(with java.nio.File)
And the lower level classes are a jar of their own.

Just dreaming...

wernerdaehn:
And I greatly appreciate your work!

wgtmac:
@gszadovszky @shangxinli @Fokko Do you have time to take a look? This has been discussed in the mailing list: https://lists.apache.org/thread/d33757j99xqn63hrfz415sq60v3x9hmy

gszadovszky (changes requested):
Thanks a lot for working on this @amousavigourabi!

About the method comments. I would only add a method comment that overrides another if there are changes in the behavior comparing to the definition of the super one.

parquet-common/src/main/java/org/apache/parquet/io/DiskInputFile.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.ByteBuffer;
+import java.nio.file.Path;
+
+/**
+ * {@code DiskInputFile} is an implementation needed by Parquet to read
+ * data files from disk using {@link SeekableInputStream} instances.
+ */
+public class DiskInputFile implements InputFile {
```
gszadovszky, line 30: I think "disk" here is not the best since it is up the `Path` implementation and the OS whether it really reads data from the disk. What do you think about "local"?

gszadovszky (changes requested):
Sorry @amousavigourabi, the comments related to the naming "disk" intended to be posted with my previous review. Somehow these comments got lost from it.

parquet-common/src/main/java/org/apache/parquet/io/LocalInputFile.java
```diff
@@ -0,0 +1,99 @@
...
+  private final Path path;
+
+  public LocalInputFile(Path file) {
+    path = file;
+  }
+
+  @Override
+  public long getLength() throws IOException {
+    RandomAccessFile file = new RandomAccessFile(path.toFile(), "r");
+    long length = file.length();
```
wgtmac, line 41: Should it be cached in case of repeated read?

Or would `path.toFile().length()` do the same thing?

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
```diff
@@ -0,0 +1,100 @@
...
+  public LocalOutputFile(Path file) {
+    path = file;
+  }
+
+  @Override
+  public PositionOutputStream create(long buffer) throws IOException {
+    return new PositionOutputStream() {
+
+      private final BufferedOutputStream stream =
+        new BufferedOutputStream(Files.newOutputStream(path), (int) buffer);
```
wgtmac, line 43: Does this support overwrite?

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+package org.apache.parquet.io;
+
+import java.io.BufferedOutputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * {@code DiskOutputFile} is an implementation needed by Parquet to write
```
wgtmac, line 27:
```suggestion
 * {@code LocalOutputFile} is an implementation needed by Parquet to write
```

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
wgtmac: I would expect `create` fails if file with same name exists.

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Override
+  public boolean supportsBlockSize() {
+    return true;
+  }
+
+  @Override
+  public long defaultBlockSize() {
+    return 512;
```
wgtmac, line 100: Is 512 too small? Any chance to make it configurable?

wgtmac: OK, that sounds good!

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
```diff
@@ -0,0 +1,107 @@
...
+    private final BufferedOutputStream stream;
+    private long pos = 0;
+
+    public LocalPositionOutputStream(int buffer, StandardOpenOption... openOption) throws IOException {
+      stream = new BufferedOutputStream(Files.newOutputStream(path, openOption), buffer);
+    }
+
+    @Override
+    public long getPos() {
+      return pos;
```
wgtmac, line 44: Can we get the position from underlying BufferedOutputStream?

parquet-common/src/main/java/org/apache/parquet/io/LocalOutputFile.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public void write(int data) throws IOException {
+      pos++;
```
wgtmac, line 49: Is this correct? int may occupy 4 bytes.

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
```diff
@@ -117,19 +120,52 @@ public void testEmptyArray() throws Exception {
     }
   }
 
+  @Test
+  public void testEmptyArrayLocal() throws Exception {
```
wgtmac, line 124: This duplicates `testEmptyArray()`. Can we parameterize them to share common code?

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWrite.java
```diff
@@ -145,6 +181,39 @@ public void testEmptyMap() throws Exception {
     }
   }
 
+  @Test
+  public void testEmptyMapLocal() throws Exception {
```
wgtmac, line 185: ditto

parquet-common/src/test/java/org/apache/parquet/io/TestLocalInputOutput.java
```diff
@@ -0,0 +1,83 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+public class TestLocalInputOutput {
+
+  Path pathFileExists = Paths.get("src/test/resources/disk_output_file_create_overwrite.parquet");
```
wgtmac, line 34: It seems that we don't need to add a test file. What about using `createTempFile()` to create a path in your tests here?

wgtmac (commented):
Thanks @amousavigourabi! Comments are mostly for testing.

wgtmac:
@amousavigourabi Will you have any update on this?

wgtmac (approved):
LGTM. Thanks!

wgtmac:
@gszadovszky Do you want to take another pass?

gszadovszky (approved):

eyeyar03:
Hi @wgtmac , just new to exploring and parsing parquet in Java.. I've been trying the sample code here but I can't make it work because of the LocalInputFile not yet in the current version published https://mvnrepository.com/artifact/org.apache.parquet/parquet-common.

This PR seems to be merged to master already though, any reason why I am not seeing the changes in the pushed jar in maven?

Thanks for the help. :)

wgtmac:
@eyeyar03 We haven't released the next major version 1.14.0 yet, so that's why you cannot see it from there. Usually we don't backport any new feature to a minor release, so the next minor version 1.13.2 will not have it too.

I am not sure what is the best time for the next major release. Could you please advise? @gszadovszky @shangxinli

gszadovszky:
@wgtmac, in terms of semantic versioning we usually do minor releases (e.g. `1.14.0`) for new features and bugfix/patch releases (e.g. `1.13.2`) to fix regressions. (A major release would mean `2.0.0` which may contain breaking changes but it is not planned any time soon.)
There is no exact rules for a release. We usually do minor releases if the community feels there are enough features/improvements in master since the last minor release. For bugfix/patch release, we shall do so if a serious regression has been introduced in the last minor releases and we have the fix for it.

amousavigourabi:
@gszadovszky @wgtmac if the next minor release is still far away (1.12.0 and 1.13.0 had over two years between them, with 1.13.0 released a few months ago), I wouldn't mind hosting the two relevant implementations in their own little Maven artifact in the meantime, as there does seem to be some demand.

gszadovszky:
@amousavigourabi, I would suggest to join the mailing list dev@parquet.apache.org and start a discussion about a potential minor release in the near future.

eyeyar03:
> @eyeyar03 We haven't released the next major version 1.14.0 yet, so that's why you cannot see it from there. Usually we don't backport any new feature to a minor release, so the next minor version 1.13.2 will not have it too.
> 
> I am not sure what is the best time for the next major release. Could you please advise? @gszadovszky @shangxinli

Thanks @wgtmac . This is noted. Guess we'll have to find an alternative solution for now while waiting for the next major release.

drealeed:
Our project needs this feature as well, is there a date for the next major release?

amousavigourabi:
> Our project needs this feature as well, is there a date for the next major release?

@drealeed if you just need to be able to drop the Hadoop Path dependency, you might want to consider copying the InputFile, OutputFile implementations from this pull request before the next release is out. If you need to fully drop Hadoop, this is still being worked on.

drealeed:
@amousavigourabi , that's actually what I did and it's working for us now. Thanks

chadselph:
I'm a big fan of avoiding hadoop dependencies but this change did cause an issue for us. Instantiating the `LocalOutputFile` creates an outputstream of 128MB, which was causing OOM exceptions for us on our lower-memory deployments. The fix was easy, just set `.withRowGroupSize` to a smaller value but the old code using hadoop Path did not require this much memory.

wgtmac:
Hi @chadselph, thanks for reporting the issue. Did you try to debug what's the root cause? Does this relate to compression? This should be unexpected. cc @amousavigourabi

amousavigourabi:
Hi @chadselph, thanks for the report. Like @wgtmac said this should be unexpected. If you could set up an example of where it shows this behaviour that'd be a good start for us to address the issue.

chadselph:
@amousavigourabi @wgtmac sorry for the delay, here's an example 

```java
import java.io.IOException;
import java.nio.file.Files;
import org.apache.avro.SchemaBuilder;
import org.apache.parquet.avro.AvroParquetWriter;
import org.apache.parquet.io.LocalOutputFile;

public class UseMemory {
  public static void main(String[] args) throws IOException {
    var schema =
        SchemaBuilder.builder()
            .record("Record")
            .fields()
            .nullableBoolean("maybe", false)
            .endRecord();
    var memStart = Runtime.getRuntime().freeMemory();
    var temp = Files.createTempFile("parquet-mem", ".parquet");
    Files.delete(temp.toAbsolutePath());
    var writer =
        AvroParquetWriter.<GenericData.Record>builder(new LocalOutputFile(temp.toAbsolutePath()))
            .withCompressionCodec(CompressionCodecName.SNAPPY)
            .withSchema(schema)
            .build();
    System.out.println(
        "Used " + (memStart - Runtime.getRuntime().freeMemory()) / (1024 * 1024) + " MBytes.");
    writer.close();
    Files.delete(temp.toAbsolutePath());
  }
}

```

I realize this is a crude way to display memory usage, but in case you're in doubt, you can run it in intellij in the debugger and calculate the retained size of the objects, `writer` is 134MB and all of it is from the output stream.

Screenshot 2024-03-13 at 14 49 59

wgtmac:
Thanks for the info! @chadselph 

Do you have time to take an initial look? @amousavigourabi

PR #1128, opened 2023-08-03
Remove redundant code

hadonchen:
There has redundant code in **org.apache.parquet.hadoop.MemoryManager,** method **updateAllocation** calculate _**maxColCount**_ but not use. So I think we can remove it to improve performance.


Fokko (approved):
Less is more, thanks @hadonchen for noticing and fixing this.

wgtmac (approved):

PR #1134, opened 2023-08-07
PARQUET-2336: Add caching key to CodecFactory

Fokko:

The `CODEC_BY_NAME` is static and may be used across different configurations. If a codec is initialized it will be re-used no matter what the configuration is. This is a problem when there are different compression levels used.

https://github.com/apache/parquet-mr/blob/515734c373f69b5250e8b63eb3d1c973da893b63/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java#L45-L46

Therefore we need to cache per compression level as well.

### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references them in the PR title. For example, "PARQUET-2336: My Parquet PR"

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
@@ -239,6 +239,7 @@ public class CodecFactory implements CompressionCodecFactory {
     if (codec != null) {
       return codec;
     }
+
     try {
       Class<?> codecClass;
       try {
@@ -266,10 +267,6 @@ public class CodecFactory implements CompressionCodecFactory {
         break;
       case ZSTD:
         level = configuration.get("parquet.compression.codec.zstd.level");
-        if (level == null) {
-          // keep "io.compression.codec.zstd.level" for backwards compatibility
-          level = configuration.get("io.compression.codec.zstd.level");
-        }
         break;
       default:
         // compression level is not supported; ignore it
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
@@ -173,10 +173,10 @@ public class CodecFactory implements CompressionCodecFactory {
           // null compressor for non-native gzip
           compressor.reset();
         }
-        CompressionOutputStream cos = codec.createOutputStream(compressedOutBuffer, compressor);
-        bytes.writeAllTo(cos);
-        cos.finish();
-        cos.close();
+        try (CompressionOutputStream cos = codec.createOutputStream(compressedOutBuffer, compressor)) {
+          bytes.writeAllTo(cos);
+          cos.finish();
+        }
         compressedBytes = BytesInput.from(compressedOutBuffer);
       }
       return compressedBytes;
@@ -234,11 +234,11 @@ public class CodecFactory implements CompressionCodecFactory {
     if (codecClassName == null) {
       return null;
     }
-    CompressionCodec codec = CODEC_BY_NAME.get(codecClassName);
+    String codecCacheKey = this.cacheKey(codecName);
+    CompressionCodec codec = CODEC_BY_NAME.get(codecCacheKey);
     if (codec != null) {
       return codec;
     }
-
     try {
       Class<?> codecClass;
       try {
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -234,11 +234,11 @@ protected CompressionCodec getCodec(CompressionCodecName codecName) {
     if (codecClassName == null) {
       return null;
     }
-    CompressionCodec codec = CODEC_BY_NAME.get(codecClassName);
+    String codecCacheKey = this.cacheKey(codecName);
+    CompressionCodec codec = CODEC_BY_NAME.get(codecCacheKey);
```
rdblue, line 238: Since `CODEC_BY_NAME` is `protected`, I think this _could_ break something that is relying on the cache, although I'm not sure why someone would access it directly. Maybe that visibility is an accident?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -234,11 +234,11 @@ protected CompressionCodec getCodec(CompressionCodecName codecName) {
...
     if (codec != null) {
       return codec;
     }
-
```
rdblue, line 241: Nit: unnecessary whitespace change.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
wgtmac: If that is a concern, we can cache the old key (w/o level) as well.

wgtmac (approved):

wgtmac:
> Overall this looks good. It changes the cache, but I can't think of why anyone would use it directly.
> 
> @Fokko, do we have a long-term plan for getting off of Hadoop codecs? It seems like that is a good idea. I think the main blocker is that we are still using these codecs. Otherwise we would be able to remove Hadoop dependencies fairly easily.

I have added the aircompressor library when I supported the LZ4_RAW codec. Not sure if this makes it easier. @rdblue

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
Fokko: Personally, I'm not too worried about this, I don't see anyone doing this. At least nobody in the Apache org: https://github.com/search?q=org%3Aapache%20CODEC_BY_NAME&type=code :)

Fokko:
Thanks @wgtmac for jumping in here

> @Fokko, do we have a long-term plan for getting off of Hadoop codecs? It seems like that is a good idea. I think the main blocker is that we are still using these codecs. Otherwise, we would be able to remove Hadoop dependencies fairly easily.

I have a PR that I need to revisit https://github.com/apache/iceberg/pull/7369. It requires some awkward changes to make it work. One issue with the Aircompressor codec it doesn't provide Brolti due to licensing issues.

wgtmac:
Just curious, is brotli widely adopted? It seems that it does not have an official java encoder implementation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -248,13 +248,36 @@ protected CompressionCodec getCodec(CompressionCodecName codecName) {
...
+        level = configuration.get("zlib.compress.level");
+        break;
+      case BROTLI:
+        level = configuration.get("compression.brotli.quality");
+        break;
+      case ZSTD:
+        level = configuration.get("parquet.compression.codec.zstd.level");
+        if (level == null) {
+          // keep "io.compression.codec.zstd.level" for backwards compatibility
+          level = configuration.get("io.compression.codec.zstd.level");
```
zhongyujiang, line 271: Is it necessary to cache the old config levels ？ It's already been deprecated and currently only the [new config](https://github.com/apache/parquet-mr/blob/a3a1ad4e58518a469970b45ccef2fb64695c1894/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/ZstandardCodec.java#L112) is being considered.

Fokko: Thanks, let's remove it then 👍🏻

Fokko:
Thanks everyone for the reviews!

PR #1139, opened 2023-09-13
PARQUET-2171: Support Hadoop vectored IO

steveloughran:

### Jira

### Tests

- [X] My PR adds the following unit tests 
      TestVectorIOBridge
      TestFileRangeBridge

It also parameterizes existing tests run with/without vector IO enabled.
    `TestParquetFileWriter`
    `TestInputFormatColumnProjection`
    `TestInputOutputFormat`

### Commits

### Documentation


parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -105,4 +107,21 @@ public abstract class SeekableInputStream extends InputStream {
    */
   public abstract void readFully(ByteBuffer buf) throws IOException;
 
+  /**
+   * Read a set of file ranges in a vectored manner.
```
danielcweeks, line 111: Can we get some param docs here?  The behavior of the `allocate` param wasn't immediately clear to me.

danielcweeks:
@steveloughran This looks really great!  I think my only comment would be about wether we can expose the implementation in a way that might be more pluggable.  In Iceberg we have a similar parallel to the InputFile and SeekableStream, but it's not apparent to me that we would be able to adapt our IO implementation to leverage vectored reads.

Open to thoughts on how we might do that as well.

shangxinli:
@steveloughran Thanks a lot for creating this PR! This is an important feature that we improve the reading performance of Parquet. I just took a brief look and they look great! I will spend some time later to review it.

steveloughran:
@shangxinli looking forward to your comments -anything you can do to test will be wonderful too!

steveloughran:
@danielcweeks that's a good point about pluggability.

1. an interface/implementation split in parquet would line you up later to choose the back end, maybe?
2. I've done an initial pass at an shim library to use vectored IO operations if a stream/hadoop version had it, but fall back to usual blocking reads if not (along with the same for everything else). but just getting the base vector io stuff into parquet is a lot simpler. I don't know if that would be useful for iceberg https://github.com/apache/hadoop-api-shim
3. video on the whole topic

getting iceberg to pass down which stripes it wants to read is critical for this to work best with s3, abfs and gcs. how are you reading the files at present?

gszadovszky (changes requested):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/BindingUtils.java
```diff
@@ -0,0 +1,211 @@
...
+  public static <T> T awaitFuture(final Future<T> future,
+    final long timeout,
+    final TimeUnit unit)
+    throws InterruptedIOException, IOException, RuntimeException,
+    TimeoutException {
+    try {
+      LOG.debug("Awaiting future");
+      return future.get(timeout, unit);
+    } catch (InterruptedException e) {
+      throw (InterruptedIOException) new InterruptedIOException(e.toString())
```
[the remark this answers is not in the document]
steveloughran: because of the initCause() afterwards converts it to Exception again

parquet-hadoop/pom.xml
```diff
@@ -79,6 +79,11 @@
       <version>${hadoop.version}</version>
       <scope>provided</scope>
     </dependency>
+<!--    <dependency>
+      <groupId>org.apache.hadoop.extensions</groupId>
+      <artifactId>hadoop-api-shim</artifactId>
+      <scope>provided</scope>
+    </dependency>-->
```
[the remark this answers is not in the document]
steveloughran: no, will cut. that was how I migrated

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/BindingUtils.java
```diff
@@ -0,0 +1,211 @@
...
+      throw (RuntimeException) cause;
+    } else if (cause instanceof Error) {
+      throw (Error) cause;
+    } else if (cause != null) {
+      // other type: wrap with a new IOE
+      return new IOException(cause);
+    } else {
+      // this only happens if there was no cause.
+      return new IOException(e);
+    }
```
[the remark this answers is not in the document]
steveloughran: will do

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -89,10 +89,13 @@
...
 import org.apache.parquet.hadoop.example.GroupWriteSupport;
 import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@RunWith(Parameterized.class)
 public class TestParquetFileWriter {
```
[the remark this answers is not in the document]
steveloughran: 1. because it does a read afterwards.
2. and when getting the pr to pass all the tests with vector IO turned on for all of them, this was one which found a bug. Parameterization ensures it stays fixed and with better coverage.

steveloughran:
@gszadovszky thanks for your comments, will update the PR

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/BindingUtils.java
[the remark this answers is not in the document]
gszadovszky: I missed that, thanks!

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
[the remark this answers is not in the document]
gszadovszky: It make sense, thanks. However, it is a bit misleading to me to set the vector IO related config for the writer as well. What do you think about separate the `Configuration` instances for the writer and the reader and only enable vector IO for the reader?

parthchandra:
> @danielcweeks that's a good point about pluggability.
> I don't know if that would be useful for iceberg https://github.com/apache/hadoop-api-shim

Iceberg can use the base Parquet File reader out of the box so should be able to use vector IO as it is. 

> getting iceberg to pass down which stripes it wants to read is critical for this to work best with s3, abfs and gcs. how are you reading the files at present?

However if the S3FileIO feature is enabled, Iceberg provides its own InputStream and InputFile implementation that use AWS SDK V2. Maybe an option to provide your own input stream to vector io might work.

parthchandra:
@mukund-thakur @steveloughran this is a great PR! Some numbers from an independent benchmark. I used Spark to parallelize the reading of all rowgroups (just the reading of the raw data) from TPC-DS/SF10000/store_sales using various APIS and here are some numbers for you. 

32 executors, 16 cores
`fs.s3a.threads.max` = 20

Reader | Avg Time (minutes) | Median | vs Baseline
-- | -- | -- | --
Parquet | 10.32 | 10 | 1
Parquet Vector IO | 2.02 | 2 | 5.1
AWS SDK V2 | 9.86 | 10 | 1
AWS SDK V2 Async | 9.66 | 9.6 | 1.1
AWS SDK V2 AsyncCrt | 9.76 | 10 | 1.1
AWS SDK V2 S3TransferManager | 9.58 | 9.5 | 1.1
AWS SDK V2 Async CRT Http Client | 10.8 | 11 | 1

Summary - The various V2 SDK clients provide lower latency and better upload speeds but for raw data scans, they are all pretty much the same.
Increasing the parallelism as, vector IO does, has maximum benefit.

mukund-thakur:
@parthchandra  Thanks for running the benchmarks. the numbers are impressive.

ahmarsuhail:
@parthchandra just wanted to check, are these numbers with Iceberg and S3FileIO?

parthchandra:
@ahmarsuhail No these numbers are not with iceberg and S3FileIO. 
I used a modified (lots of stuff removed) version of the ParquetFileReader and a custom benchmark program that reads all the row groups in parallel and records the time spent in each read from S3. The modified version of ParquetFileReader can switch between the various methods of reading from S3.
The entry `AWS SDK V2` is a near copy of the Iceberg S3FileIO code though. 
I saw issues with the CRT client when running at scale causing JVM crashes. And the V2 transfer manager did not do range reads properly. Do share your experience.

steveloughran:
Thanks for the numbers; I am deep in the aws v2 sdk migration right now and haven't had a chance to work on this.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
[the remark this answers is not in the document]
steveloughran: started this, but it doesn't work. because filesystems are cached on path; once a writer has retrieved the fs instance for a path, the reader gets it back too. sorry

steveloughran:
...back on this. @parthchandra you know that hadoop trunk is built on the v2 sdk now?

steveloughran:
OK, I've tried to address the changes as well as merge with master

The one thing I'm yet to do is the one by @danielcweeks : have an interface for which the hadoop vector IO would be just one implementation.

We effectively have that in SeekableInputStream; two new default methods: one a probe for the api availability and the other an invocation.
```

Would you be able to wire up the iceberg reader to that? And if not, what changes are needed?

One thing we would need to make sure was good is the awaitFuture stuff; that's a copy of what's in hadoop to handle async IO operations. There's also a hard coded timeout of 300s to wait for the results; I don't know/recall where that number came from but it's potentially dubious as it won't recover from network problems.

File: parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/BindingUtils.java [hunk 6]
```diff
@@ -0,0 +1,209 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.apache.parquet.util.DynMethods;
+
+/**
+ * Binding utils.
+ */
+public final class BindingUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(BindingUtils.class);
+
```
wgtmac, line 40:
```suggestion
```

parquet-hadoop/README.md
```diff
@@ -501,3 +501,11 @@ If `false`, key material is stored in separate new files, created in the same fo
 **Description:** Length of key encryption keys (KEKs), randomly generated by parquet key management tools. Can be 128, 192 or 256 bits.  
 **Default value:** `128`
 
+---
+
+**Property:** `parquet.hadoop.vectored.io.enabled`
```
wgtmac, line 506: Thanks for adding the doc!

wgtmac (commented):
Thanks for the change! I've just scanned it for the 1st pass. Do you have any outstanding work on this PR?

parquet-hadoop/README.md
```diff
@@ -501,3 +501,11 @@ If `false`, key material is stored in separate new files, created in the same fo
 **Description:** Length of key encryption keys (KEKs), randomly generated by parquet key management tools. Can be 128, 192 or 256 bits.  
 **Default value:** `128`
 
+---
+
+**Property:** `parquet.hadoop.vectored.io.enabled`  
+**Description:** Flag to enable use of the FileSystem Vector IO API on Hadoop releases which support the feature.
+If `true` then an attempt will be made to dynamically load the relevant classes;
```
wangyum, line 508: Does this feature need to upgrade NameNode and DataNode?

steveloughran:
Code wise, no, other than reviews from others about what is the best place for things, such as that awaitFuture stuff or any other suggestions which people who know the parquet codebase think is best. Code works and we have been testing this through Amazon S3 Express storage for extra speed up. To be ruthless: there's no point paying the premium for that until you've embraced the extra speed ups you get from this first

steveloughran:
@Fokko that's japicmp getting its logic wrong because it's a new file; thought I'd edited the build rules so it would ignore that.


anyway, need to fix the merge as something (#1209?) has just broken it

parquet-hadoop/README.md
steveloughran: no, hdfs doesn't support it. Native IO does, so if you use file:// URLS you get direct NIO vectored IO into buffers (yay! hadoop APIs move to the 2010s!). S3A supports it with multiple parallel GET with some range coalescing in between. Would love ABFS connector to support it too...

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
```diff
@@ -219,6 +230,7 @@ public static class Builder {
     protected boolean useStatsFilter = STATS_FILTERING_ENABLED_DEFAULT;
     protected boolean useDictionaryFilter = DICTIONARY_FILTERING_ENABLED_DEFAULT;
     protected boolean useRecordFilter = RECORD_FILTERING_ENABLED_DEFAULT;
+    protected boolean useHadoopVectoredIo = HADOOP_VECTORED_IO_ENABLED_DEFAULT;
```
pan3793, line 233:
```suggestion
    protected boolean useHadoopVectoredIO = HADOOP_VECTORED_IO_ENABLED_DEFAULT;
```

steveloughran:
Thanks. I'm away from my laptop until 2024 but really do want to get this in.

parquet-common/src/main/java/org/apache/parquet/io/ParquetFileRange.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class ParquetFileRange {
+
+  /**
+   * Start position in file.
+   */
+  private final long offset;
+
+  /**
+   * Length of data to be read from position.
```
shangxinli, line 37: To be more specific, can we change data to bytes?

parquet-common/src/main/java/org/apache/parquet/io/ParquetFileRange.java
```diff
@@ -0,0 +1,71 @@
...
+    return dataReadFuture;
+  }
+
+  public void setDataReadFuture(CompletableFuture<ByteBuffer> dataReadFuture) {
+    this.dataReadFuture = dataReadFuture;
+  }
+
+  @Override
+  public String toString() {
+    return "range[" + this.offset + "," + (this.offset + (long)this.length) + ")";
```
shangxinli, line 69: The [ and ) don't pair. Is it intentional?

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
shangxinli: +1

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -105,4 +107,21 @@ public abstract class SeekableInputStream extends InputStream {
...
+   * @throws UnsupportedOperationException if not available in this class/runtime.
+   */
+  public void readVectored(List<ParquetFileRange> ranges,
+    IntFunction<ByteBuffer> allocate) throws IOException {
+
+    throw new UnsupportedOperationException("Vectored IO is not supported for " + this);
```
shangxinli, line 117: Do you have any idea what 'this' will print out?  Like to know if it is useful to user to troubleshot in case of exception

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1063,6 +1068,69 @@ public ColumnChunkPageReadStore readFilteredRowGroup(int blockIndex, RowRanges r
...
+
+  /**
+   * Vectored IO doesn't support reading ranges of size greater than
+   * Integer.MAX_VALUE.
+   * @param allParts all parts to read.
+   * @return true or false.
+   */
+  private boolean partsLengthValidForVectoredIO(List<ConsecutivePartList> allParts) {
+    for (ConsecutivePartList consecutivePart :  allParts) {
+      if (consecutivePart.length >= Integer.MAX_VALUE) {
```
shangxinli, line 1099: Remove extra space before allParts

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1063,6 +1068,69 @@ public ColumnChunkPageReadStore readFilteredRowGroup(int blockIndex, RowRanges r
...
 
+  /**
+   * Read data in all parts via either vectored IO or serial IO.
+   * @param allParts all parts to be read.
+   * @param builder used to build chunk list to read the pages for the different columns.
+   * @throws IOException any IOE.
+   */
+  private void readAllPartsVectoredOrNormal(List<ConsecutivePartList> allParts, ChunkListBuilder builder)
+    throws IOException {
+    boolean isVectoredIO = options.useHadoopVectoredIO()
```
shangxinli, line 1079: Move the 3 conditions into a method and make a good document to call out what exact conditions are needed for the user to use vectorized IO. And also we should log it clearly whether or not vectoredIO is used.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1063,6 +1068,69 @@ public ColumnChunkPageReadStore readFilteredRowGroup(int blockIndex, RowRanges r
...
+    List<ParquetFileRange> ranges = new ArrayList<>(allParts.size());
+    for (ConsecutivePartList consecutiveChunks : allParts) {
+      Preconditions.checkArgument(consecutiveChunks.length < Integer.MAX_VALUE,
+        "Invalid length %s for vectored read operation. It must be less than max integer value.",
+        consecutiveChunks.length);
+      ranges.add(new ParquetFileRange(consecutiveChunks.offset, (int) consecutiveChunks.length));
+    }
+    LOG.debug("Doing vectored IO for ranges {}", ranges);
+    ByteBufferAllocator allocator = options.getAllocator();
+    //blocking or asynchronous vectored read.
```
shangxinli, line 1125: The comment is not very clear.  Can you move it to be above the definition of the readVectorized(...) where you can make more comments to make it more clear

steveloughran:
...rebased; now got some (brief) time to look at the comments -don't review just yet

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
steveloughran: ok. will log at info with some more details about the request (number of ranges, total amount of data).

We could also report on the total duration of the vectored read...would that be useful?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
steveloughran: this should be good now

ggershinsky:
@steveloughran Thanks for this PR! When can we expect it to be merge-ready?

steveloughran:
let me have another look at your comments and make sure I've answered them all

steveloughran:
Update: I've been looking at overlapping ranges.
* The base implementation doesn't care.
* the specification and contract tests don't say they MUST be separate
* I'm now worried that a bit of the s3a range coalescing code is getting overlaps wrong.

Here's what I propose
* for hadoop api we say "no overlap" and explicitly raise IOE subclass in the invocation if so. (i.e. always fail fast)
* add tests
* for parquet, we do the same validation so behaviour is consistent across all impls
* If there is an overlap, what to do there? fail or downgrade?

The reason i'd like to fail fast in overlaps is because it lines up nicely with ranged HTTP Get requests. S3 doesn't support that, but others might already and if not, could be persuaded. Currently the s3a client will coalesce "close enough" ranges into single GETs, discarding the data in between. This is more efficient for http connection use and latency -even costs less for in-EC2 queries. 

I do want to add azure abfs vector reads this year.

parthchandra:
> * for parquet, we do the same validation so behaviour is consistent across all impls

I think that is the correct behaviour. 

> * If there is an overlap, what to do there? fail or downgrade?

IMHO, it seems to me that the Parquet side should validate that there are no overlapping ranges. So if the implementation detects an overlapping range it would really be an internal Parquet error and fail fast would be preferable.

ggershinsky:
Per the [thread](https://lists.apache.org/thread/kttwbl5l7opz6nwb5bck2gghc2y3td0o), it'd be good to have this patch in 1.14.0 :) Otherwise, can take a very long time till the next one..

parthchandra:
> Per the [thread](https://lists.apache.org/thread/kttwbl5l7opz6nwb5bck2gghc2y3td0o), it'd be good to have this patch in 1.14.0 :) Otherwise, can take a very long time till the next one..

Note that this PR is for vectorized IO while the thread referred to is talking about vectorized reading (i.e. decoding). The two are not mutually exclusive and both provide performance gains independent of each other.

ggershinsky:
this PR is already covered there. But I mostly meant the thread title.

parthchandra:
> Now, what do people think about fallbacks? we've been using this and to date parquet hasn't ever issued an overlapping request, but there's still the future to think about. How about, when hadoop vector IO requests fail during parquet-side range checks, there's something logged (level?) and fallback to classic PositionedReadable.

I think this is reasonable (INFO level, I think. This is similar to `INFO S3AInputStream: Switching to Random IO seek policy` . Though I'm also fine if an exception is thrown if there are overlapping requests.

parthchandra:
@steveloughran @mukund-thakur do you guys have any information on how much (if any) this impacts the peak memory utilization in the parquet file reader? The total memory allocated while reading a row group should remain the same, but if vector io allocates intermediate buffers (or even if the http request to S3, say, allocates additional memory) then we could see an increase in peak memory usage. 
Have you folks encountered any issues with that?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1148,12 +1148,17 @@ private void readAllPartsVectoredOrNormal(List<ConsecutivePartList> allParts, Ch
...
-      readVectored(allParts, builder);
-    } else {
-      for (ConsecutivePartList consecutiveChunks : allParts) {
-        consecutiveChunks.readAll(f, builder);
+      try {
+        readVectored(allParts, builder);
+        return;
+      } catch (IOException | IllegalArgumentException | UnsupportedOperationException e) {
+        // possible failure modes.
+        LOG.warn("readVectored() failed; falling back to normal IO", e);
```
mukund-thakur, line 1156: Do we need to support fallback? What if some parts were read successfully and failed in later stages?

mukund-thakur:
> @steveloughran @mukund-thakur do you guys have any information on how much (if any) this impacts the peak memory utilization in the parquet file reader? The total memory allocated while reading a row group should remain the same, but if vector io allocates intermediate buffers (or even if the http request to S3, say, allocates additional memory) then we could see an increase in peak memory usage. Have you folks encountered any issues with that?

Hey, we haven't done any memory profiling or seen any memory-related issues during our internal benchmarking.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1132,6 +1138,109 @@ public ColumnChunkPageReadStore readFilteredRowGroup(int blockIndex, RowRanges r
...
+
+  /**
+   * Validated the parts for vectored IO.
+   * Vectored IO doesn't support reading ranges of size greater than
+   * Integer.MAX_VALUE.
+   * @param allParts all parts to read.
+   * @return true or false.
+   */
+  private boolean arePartsValidForVectoredIO(List<ConsecutivePartList> allParts) {
+    for (ConsecutivePartList consecutivePart : allParts) {
```
mukund-thakur, line 1194: I thought we decided to add a check for overlapping ranges explicitly in Parquet. where is that?

steveloughran: not yet ready; i want to make sure the hadoop-side validation does what I want and then will copy and paste.

One thing there is to decide how best to deal with negative offsets. EOF vs IllegalArgument?

wgtmac: +1 for IllegalArgument on my side. IMO, fallback for overlapping ranges is an overkill. We can simply throw and let the user to turn off vectored io if something is wrong.

parthchandra:
@steveloughran are you planning to incorporate the read metrics added in https://github.com/apache/parquet-mr/commit/2e0cd1925546d2560f7658086251851e6fa68559 ?
I can add them after this is merged so as not to hold up this PR.

steveloughran:
@parthchandra can we do that together after?

parthchandra:
> @parthchandra can we do that together after?
Of course. Let's get this PR done.

steveloughran:
> @steveloughran @mukund-thakur do you guys have any information on how much (if any) this impacts the peak memory utilization in the parquet file reader? The total memory allocated while reading a row group should remain the same, but if vector io allocates intermediate buffers (or even if the http request to S3, say, allocates additional memory) then we could see an increase in peak memory usage.

@parthchandra for S3A reading into on-heap buffers (which is the only option this PR supports) we allocate in the SDK and just stream it in from the GET request. For reading into a merged range we discard the stuff in between.

At the end of reading all the ranges then, they are all on heap and there shouldn't be extra overhead. During the read more connections are used in a shared thread pool; connection pool size shoulds to be increased as should the number of threads. We increase the defaults in 3.4.0; it's generally good to have big numbers. That is the biggest intermediate load penalty, as far as I can tell.

For raw local: uses java nio API, so we hope they've got that right.

And for the base implementation, it's just readFully() with offsets and range: nothing unusual.

The off/heap story is a bit more complicated, which is partially why it is not supported here. For s3a data is streamed to heap in 2MB blocks and then copied off heap, so memory cost for no tangible gain. Java nio is better for local, but it becomes tricky to manage the lifespan of those buffers once range coalescing gets involved.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
steveloughran: That is actually a valid point, but it gets complicated fast. Why don't you submit a follow-up PR? You will need to split out the successful from the unsuccessful reads and only execute the new ones.

mukund-thakur: Yes, it will get complicated and that is why I was thinking is that an overkill or should we just fail in case of failure? 
In the meantime, I will try to see how complicated it gets code wise.

steveloughran:
ahh, doesnt build without the hadoop version of guava preconditions (we added after getting fed up with brittleness of guava)
```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/VectorIOBridge.java:[22,37] cannot find symbol
  symbol:   class Preconditions
  location: package org.apache.hadoop.util
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/VectorIOBridge.java:[22,1] static import only from classes and interfaces
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/VectorIOBridge.java:[345,5] cannot find symbol
  symbol:   method checkArgument(boolean,java.lang.String,org.apache.parquet.io.ParquetFileRange)
  location: class org.apache.parquet.hadoop.util.vectorio.VectorIOBridge
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/VectorIOBridge.java:[367,5] cannot find symbol
  symbol:   method checkArgument(boolean,java.lang.String)
  location: class org.apache.parquet.hadoop.util.vectorio.VectorIOBridge
Error:  /home/runner/work/parquet-mr/parquet-mr/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/vectorio/VectorIOBridge.java:[379,11] cannot find symbol
  symbol:   method checkArgument(boolean,java.lang.String,org.apache.parquet.io.ParquetFileRange,org.apache.parquet.io.ParquetFileRange)
  location: class org.apache.parquet.hadoop.util.vectorio.VectorIOBridge
[INFO] 5 errors 
```
will fix

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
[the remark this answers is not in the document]
steveloughran: summary. cantfix

mukund-thakur (commented):
LGTM +1. I think this is good to go in now.

parthchandra:
+1 from me as well.

steveloughran:
tested across the different profiles, all good. i am seeing hadoop 3.4 mr test failures, but that happens on master too.

One thing I do want to change now is actually propose changing the package name from vectorio to something more like "wrapped io"

the reason for this is that in https://github.com/apache/hadoop/pull/6686
I'm adding an "easy use with reflection" wrapper to openFile(), making it easy to
* pass down filestatus already collected (saves a HEAD on the stores)
* pass in read policy as "vectored, random"

Having a unified package for this stuff makes sense, and vectorio is too narrow?

steveloughran:
Update
* moved package to `org.apache.parquet.hadoop.util.wrappedio`
* pulled copied hadoop.util.FutureIO methods into one of the same name

We could consider making BindingUtils package private.

parthchandra:
> We could consider making BindingUtils package private.
It's pretty generic and might be useful elsewhere so I feel it is ok to keep it public.

wgtmac (commented):
Just reviewed for another pass. I've left some minor comments. Overall LGTM, thanks @steveloughran!

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
steveloughran: this has been covered with validation of overlaps and invalid ranges; hadoop is consistent with this too.

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
steveloughran: will do, and rename getters/setters

wgtmac:
> @steveloughran This looks really great! I think my only comment would be about wether we can expose the implementation in a way that might be more pluggable. In Iceberg we have a similar parallel to the InputFile and SeekableStream, but it's not apparent to me that we would be able to adapt our IO implementation to leverage vectored reads.
> 
> Open to thoughts on how we might do that as well.
> 
> Maybe this would be possible by implementing the vectoredIO methods in the [Iceberg adaptor class here](https://github.com/apache/iceberg/blob/c07f2aabc0a1d02f068ecf1514d2479c0fbdd3b0/parquet/src/main/java/org/apache/iceberg/parquet/ParquetIO.java#L105)?

Do you want to double check this again? @danielcweeks @Fokko

steveloughran:
Right, I think I have addressed all the issues except for the design of a public API for other input sources.

I do think that would be good, not just for iceberg integration, but because it will let local filesystem clients do this through LocalInputFile. This is already in hadoop, where it only takes 100 LoC to implement (https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java#L319). (The hadoop API used java.io.nio when designing ours, so this isn't surprising).

Can we get this in and then I can collaborate with you and anyone else interested on having a broader API? I would obviously like what we've done to be ~the reference design, but the binding, testing and including this in the benchmarks are areas I need help from the experts in the project to understand how to fit it in the best.

Incidentally, you should be aware that the next bit of parquet work I want to do is to pass in seek policy and file status when opening files. This is trivial to do on Hadoop 3.3+ saves a HEAD request through the ABFS and S3A connectors, and let them tune prefetching/caching/GET range policy. However, trying to get that builder API through reflection is *very* complicated. 

What I am doing is create some reflection-friendly static methods (https://github.com/apache/hadoop/pull/6686), ship this in hadoop 3.3.9 and 3.4.1 and use the BindingUtils and DynMethods code to invoke. Parquet always calls getFileStatus() before opening a file, so the HEAD is completely superfluous. Note: that hadoop PR lifts and repackages the DynMethods classes so that for testing there we can be confident it works.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
```diff
@@ -104,4 +106,25 @@ public abstract class SeekableInputStream extends InputStream {
...
+   * @throws UnsupportedOperationException if not available in this class/runtime (default)
+   */
+  public void readVectored(List<ParquetFileRange> ranges, final ByteBufferAllocator allocator) throws IOException {
+
+    throw new UnsupportedOperationException("Vectored IO is not supported for " + this);
+  }
+
+  /**
+   * Is the {@link #readVectored(List, ByteBufferAllocator)} method available?
+   * @param allocator the allocator to use for allocating ByteBuffers
```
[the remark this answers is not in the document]
wgtmac: What's your opinion in this? @steveloughran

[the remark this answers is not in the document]
steveloughran: aah, you've noticed that! It's because I've realised that on s3/abfs/gcs network failures which don't recover after retries, we need a way to return the buffer to the pool. See https://issues.apache.org/jira/browse/HADOOP-19105

I actually need to do some more in general on s3a read() recovery: we retry on the GET calls and then keep the connection active only as long as it takes to read the data -so no risk of stale connections is low- but I do need to add the full resilience logic there
* abort http connections considered unrecoverable, rather than return to pool
* retry outstanding entries in each range
* maybe actually handle failures by decombining adjacent ranges.

Having parquet pass the allocator as far down as it does means that supporting releases with the "return to pool" callback doesn't change the rest of the code, instead the bridge would first look for the new method, falling back to the original one if not found.

parquet-common/src/main/java/org/apache/parquet/io/ParquetFileRange.java
```diff
@@ -0,0 +1,71 @@
...
+    return dataReadFuture;
+  }
+
+  public void setDataReadFuture(CompletableFuture<ByteBuffer> dataReadFuture) {
+    this.dataReadFuture = dataReadFuture;
+  }
+
+  @Override
+  public String toString() {
+    return "range[" + this.offset + " - " + (this.offset + (long) this.length) + "]";
```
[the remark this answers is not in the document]
steveloughran: why this change, given range starts with the "[" symbol? I'm happy to merge it, just not sure why unless you want to delineate that the second value in the range is inclusive, which is isn't, now I look at it. maybe it should be 

```java
return String.format("range[%,d to %,d]",
   offset, 
   (offset + (length > 0 ? (length - 1 ) : 0));
```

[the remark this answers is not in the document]
wgtmac: Yes, my intention is to let readers know whether each end is exclusive or inclusive.

parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java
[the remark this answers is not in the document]
wgtmac: That makes sense. Thanks for the explanation!

wgtmac (approved):
+1

wgtmac:
@gszadovszky @shangxinli @ggershinsky Could you help review this for another pass? Thanks!

parquet-common/src/main/java/org/apache/parquet/io/ParquetFileRange.java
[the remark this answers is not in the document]
steveloughran: ok, will tweak

steveloughran:
@wgtmac tweaked that FileRange.toString(); now

```
String.format("range[%,d - %,d)", offset, offset + length);
```

...that shows that the range is inclusive and gives , separators on large ranges.

This is just for debugging -but that doesn't mean it shouldn't be meaningful

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
steveloughran: missed this, will tweak

wgtmac:
I will merge this by the end of this week if no objection. We need to release parquet 1.14.0 and this is a must-have feature.

steveloughran:
thanks! FWIW ORC shipped with this feature recently too.
they also made their release java 17+ only which is quite aggressive and something I'm jealous of...java 8 is way past its EOL

gszadovszky (approved):

steveloughran:
thanks for the approval. I have more plans. but first want to make it really easy for parquet to pick up hadoop-3.2/3.3 APIs through reflection; https://github.com/apache/hadoop/pull/6686

get that into 3.3.9/3.4.1 and then parquet will be able to
* open files with explicit read policy and skipping the head call
* maybe use ByteBufferPositionedReadable.readFully() to read buffers. that's only currently in hdfs though, so the existing code is still used in most places.
+ more

if we can do vectored IO through reflection, then other stuff is there for the taking too.

parthchandra:
@wgtmac it would be great to get this in. I'm waiting with a followup to add simple metrics for the vector io read path and it would be nice if that might make it to the release as well.

wgtmac:
I just merged it. Thanks all!

steveloughran:
thanks! looking forward to other people's results. 

I also have plans for more...

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -152,6 +161,35 @@ public class TestParquetFileWriter {
   @Rule
   public final TemporaryFolder temp = new TemporaryFolder();
 
+  @Parameterized.Parameters(name = "vectored : {0}")
```
mukund-thakur, line 164: Somehow we need to add a validation that when tests with vectored IO enabled run, vectored IO actually gets triggered.

PR #1141, opened 2023-09-16
PARQUET-2347: Add interface layer between Parquet and Hadoop Configuration

amousavigourabi:

### Jira

### Tests

       Additional parameters to run the read/write tests using the new interface-using methods.

### Commits

### Documentation

---

jacicmp exclusions have been added for the following classes: `org.apache.parquet.hadoop.CodecFactory`, `org.apache.parquet.hadoop.ParquetReader`. When these exclusions are removed, the following incompatibilities are detected:
```
There is at least one incompatibility: org.apache.parquet.hadoop.CodecFactory.configuration:FIELD_TYPE_CHANGED,org.apache.parquet.hadoop.ParquetReader$Builder.conf:FIELD_TYPE_CHANGED
```

This PR is part of an effort that has been [discussed on the dev mailing list](https://lists.apache.org/thread/4wl0l3d9dkpx4w69jx3rwnjk034dtqr8).

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -170,10 +170,6 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
         parquetSchema, avroSchema, model);
   }
 
-  private GenericData getDataModel(Configuration conf, Schema schema) {
-    return getDataModel(new HadoopParquetConfiguration(conf), schema);
-  }
-
   private GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
     if (model != null) {
       return model;
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
@@ -24,7 +24,10 @@ import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.util.ReflectionUtils;
+import org.apache.parquet.conf.HadoopParquetConfiguration;
+import org.apache.parquet.conf.ParquetConfiguration;
 import org.apache.parquet.hadoop.api.ReadSupport;
+import org.apache.parquet.hadoop.util.ConfigurationUtil;
 import org.apache.parquet.io.api.RecordMaterializer;
 import org.apache.parquet.schema.MessageType;
 import org.slf4j.Logger;
@@ -95,6 +98,13 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
   public ReadContext init(Configuration configuration,
                           Map<String, String> keyValueMetaData,
                           MessageType fileSchema) {
+    return init(new HadoopParquetConfiguration(configuration), keyValueMetaData, fileSchema);
+  }
+
+  @Override
+  public ReadContext init(ParquetConfiguration configuration,
+                          Map<String, String> keyValueMetaData,
+                          MessageType fileSchema) {
     MessageType projection = fileSchema;
     Map<String, String> metadata = new LinkedHashMap<String, String>();
 
@@ -120,6 +130,13 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
   public RecordMaterializer<T> prepareForRead(
       Configuration configuration, Map<String, String> keyValueMetaData,
       MessageType fileSchema, ReadContext readContext) {
+    return prepareForRead(new HadoopParquetConfiguration(configuration), keyValueMetaData, fileSchema, readContext);
+  }
+
+  @Override
+  public RecordMaterializer<T> prepareForRead(
+    ParquetConfiguration configuration, Map<String, String> keyValueMetaData,
+    MessageType fileSchema, ReadContext readContext) {
     Map<String, String> metadata = readContext.getReadSupportMetadata();
     MessageType parquetSchema = readContext.getRequestedSchema();
     Schema avroSchema;
@@ -154,6 +171,10 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
   }
 
   private GenericData getDataModel(Configuration conf, Schema schema) {
+    return getDataModel(new HadoopParquetConfiguration(conf), schema);
+  }
+
+  private GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
     if (model != null) {
       return model;
     }
@@ -175,6 +196,6 @@ public class AvroReadSupport<T> extends ReadSupport<T> {
 
     Class<? extends AvroDataSupplier> suppClass = conf.getClass(
         AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
-    return ReflectionUtils.newInstance(suppClass, conf).get();
+    return ReflectionUtils.newInstance(suppClass, ConfigurationUtil.createHadoopConfiguration(conf)).get();
   }
 }
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
@@ -412,10 +412,6 @@ public class AvroWriteSupport<T> extends WriteSupport<T> {
     return Binary.fromCharSequence(value.toString());
   }
 
-  private static GenericData getDataModel(Configuration conf, Schema schema) {
-    return getDataModel(new HadoopParquetConfiguration(conf), schema);
-  }
-
   private static GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
     if (conf.get(AVRO_DATA_SUPPLIER) == null && schema != null) {
       GenericData modelForSchema;
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
@@ -34,7 +34,10 @@ import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.avro.util.Utf8;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.parquet.conf.HadoopParquetConfiguration;
+import org.apache.parquet.conf.ParquetConfiguration;
 import org.apache.parquet.hadoop.api.WriteSupport;
+import org.apache.parquet.hadoop.util.ConfigurationUtil;
 import org.apache.parquet.io.api.Binary;
 import org.apache.parquet.io.api.RecordConsumer;
 import org.apache.parquet.schema.GroupType;
@@ -129,6 +132,11 @@ public class AvroWriteSupport<T> extends WriteSupport<T> {
 
   @Override
   public WriteContext init(Configuration configuration) {
+    return init(new HadoopParquetConfiguration(configuration));
+  }
+
+  @Override
+  public WriteContext init(ParquetConfiguration configuration) {
     if (rootAvroSchema == null) {
       this.rootAvroSchema = new Schema.Parser().parse(configuration.get(AVRO_SCHEMA));
       this.rootSchema = new AvroSchemaConverter(configuration).convert(rootAvroSchema);
@@ -405,6 +413,10 @@ public class AvroWriteSupport<T> extends WriteSupport<T> {
   }
 
   private static GenericData getDataModel(Configuration conf, Schema schema) {
+    return getDataModel(new HadoopParquetConfiguration(conf), schema);
+  }
+
+  private static GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
     if (conf.get(AVRO_DATA_SUPPLIER) == null && schema != null) {
       GenericData modelForSchema;
       try {
@@ -423,7 +435,7 @@ public class AvroWriteSupport<T> extends WriteSupport<T> {
 
     Class<? extends AvroDataSupplier> suppClass = conf.getClass(
         AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
-    return ReflectionUtils.newInstance(suppClass, conf).get();
+    return ReflectionUtils.newInstance(suppClass, ConfigurationUtil.createHadoopConfiguration(conf)).get();
   }
 
   private abstract class ListWriter {
--- a/parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
+++ b/parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
@@ -1,0 +1,62 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.conf;
+
+import java.util.Map;
+
+/**
+ * Configuration interface with the methods necessary to configure for Parquet applications.
+ */
+public interface ParquetConfiguration extends Iterable<Map.Entry<String,String>> {
+
+  void set(String name, String value);
+
+  void setLong(String name, long value);
+
+  void setInt(String name, int value);
+
+  void setBoolean(String name, boolean value);
+
+  void setStrings(String name, String... value);
+
+  void setClass(String name, Class<?> value, Class<?> xface);
+
+  String get(String name);
+
+  String get(String name, String defaultValue);
+
+  long getLong(String name, long defaultValue);
+
+  int getInt(String name, int defaultValue);
+
+  boolean getBoolean(String name, boolean defaultValue);
+
+  String getTrimmed(String name);
+
+  String getTrimmed(String name, String defaultValue);
+
+  String[] getStrings(String name, String[] defaultValue);
+
+  Class<?> getClass(String name, Class<?> defaultValue);
+
+  <U> Class<? extends U> getClass(String name, Class<? extends U> defaultValue, Class<U> xface);
+
+  Class<?> getClassByName(String name) throws ClassNotFoundException;
+}
--- a/parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
+++ b/parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
@@ -22,41 +22,157 @@ package org.apache.parquet.conf;
 import java.util.Map;
 
 /**
- * Configuration interface with the methods necessary to configure for Parquet applications.
+ * Configuration interface with the methods necessary to configure Parquet applications.
  */
 public interface ParquetConfiguration extends Iterable<Map.Entry<String,String>> {
 
+  /**
+   * Sets the value of the name property.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   */
   void set(String name, String value);
 
+  /**
+   * Sets the value of the name property to a long.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   */
   void setLong(String name, long value);
 
+  /**
+   * Sets the value of the name property to an integer.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   */
   void setInt(String name, int value);
 
+  /**
+   * Sets the value of the name property to a boolean.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   */
   void setBoolean(String name, boolean value);
 
+  /**
+   * Sets the value of the name property to an array of comma delimited values.
+   *
+   * @param name the property to set
+   * @param value the values to set the property to
+   */
   void setStrings(String name, String... value);
 
+  /**
+   * Sets the value of the name property to a class.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   * @param xface the interface implemented by the value
+   */
   void setClass(String name, Class<?> value, Class<?> xface);
 
+  /**
+   * Gets the value of the name property. Returns null if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @return the value of the property, or null if it does not exist
+   */
   String get(String name);
 
+  /**
+   * Gets the value of the name property. Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property, or the default value if it does not exist
+   */
   String get(String name, String defaultValue);
 
+  /**
+   * Gets the value of the name property as a long. Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as a long, or the default value if it does not exist
+   */
   long getLong(String name, long defaultValue);
 
+  /**
+   * Gets the value of the name property as an integer. Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as an integer, or the default value if it does not exist
+   */
   int getInt(String name, int defaultValue);
 
+  /**
+   * Gets the value of the name property as a boolean. Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as a boolean, or the default value if it does not exist
+   */
   boolean getBoolean(String name, boolean defaultValue);
 
+  /**
+   * Gets the trimmed value of the name property. Returns null if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @return the trimmed value of the property, or null if it does not exist
+   */
   String getTrimmed(String name);
 
+  /**
+   * Gets the trimmed value of the name property as a boolean.
+   * Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the trimmed value of the property, or the default value if it does not exist
+   */
   String getTrimmed(String name, String defaultValue);
 
+  /**
+   * Gets the value of the name property as an array of {@link String}s.
+   * Returns the default value if no such value exists.
+   * Interprets the stored value as a comma delimited array.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as an array, or the default value if it does not exist
+   */
   String[] getStrings(String name, String[] defaultValue);
 
+  /**
+   * Gets the value of the name property as a class. Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as a class, or the default value if it does not exist
+   */
   Class<?> getClass(String name, Class<?> defaultValue);
 
+  /**
+   * Gets the value of the name property as a class implementing the xface interface.
+   * Returns the default value if no such value exists.
+   *
+   * @param name the property to retrieve the value of
+   * @param defaultValue the default return if no value is set for the property
+   * @return the value of the property as a class, or the default value if it does not exist
+   */
   <U> Class<? extends U> getClass(String name, Class<? extends U> defaultValue, Class<U> xface);
 
+  /**
+   * Load a class by name.
+   *
+   * @param name the name of the {@link Class} to load
+   * @return the loaded class
+   * @throws ClassNotFoundException when the specified class cannot be found
+   */
   Class<?> getClassByName(String name) throws ClassNotFoundException;
 }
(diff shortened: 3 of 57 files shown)


wgtmac:
Thanks for working on this! I will take a look by the end of this week.

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
```diff
@@ -333,13 +406,17 @@ public Builder copy(ParquetReadOptions options) {
 
     public ParquetReadOptions build() {
       if (codecFactory == null) {
-        codecFactory = HadoopCodecs.newFactory(0);
+        if (conf == null) {
+          codecFactory = HadoopCodecs.newFactory(0);
+        } else {
+          codecFactory = HadoopCodecs.newFactory(conf, 0);
```
wgtmac, line 412: It seems that original ParquetReadOptions does not require any Configuration parameter. Should we avoid adding this by using ParquetConfiguration internally in the HadoopCodecs?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -167,13 +169,13 @@ public float getProgress() throws IOException, InterruptedException {
...
     // copy custom configuration to the Configuration passed to the ReadSupport
-    Configuration conf = new Configuration();
-    if (options instanceof HadoopReadOptions) {
-      conf = ((HadoopReadOptions) options).getConf();
-    }
+    ParquetConfiguration conf = Objects.requireNonNull(options).getConfiguration();
     for (String property : options.getPropertyNames()) {
       conf.set(property, options.getProperty(property));
     }
+    for (Map.Entry<String, String> property : new Configuration()) {
```
wgtmac, line 176: CMIW, we do not need to address the Hadoop dependency in the parquet-hadoop module, right?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
```diff
@@ -261,7 +263,7 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
 
         LOG.debug("read value: {}", currentValue);
       } catch (RuntimeException e) {
-        throw new ParquetDecodingException(format("Can not read value at %d in block %d in file %s", current, currentBlock, reader.getPath()), e);
+        throw new ParquetDecodingException(format("Can not read value at %d in block %d in file %s", current, currentBlock, reader.getFile()), e);
```
wgtmac, line 266: If my previous statement is correct, then perhaps we do not need change in this line?

pom.xml
```diff
@@ -547,6 +547,8 @@
             </excludeModules>
             <excludes>
               <exclude>${shade.prefix}</exclude>
+              <exclude>org.apache.parquet.hadoop.CodecFactory</exclude>
```
wgtmac, line 550: These lines worth adding some comments.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -423,7 +435,7 @@ private static GenericData getDataModel(Configuration conf, Schema schema) {
 
     Class<? extends AvroDataSupplier> suppClass = conf.getClass(
         AVRO_DATA_SUPPLIER, SpecificDataSupplier.class, AvroDataSupplier.class);
-    return ReflectionUtils.newInstance(suppClass, conf).get();
+    return ReflectionUtils.newInstance(suppClass, ConfigurationUtil.createHadoopConfiguration(conf)).get();
```
wgtmac, line 438: Does it mean that we still cannot get rid of the hadoop dependency? Anyway we have to depend on Configuration class here.

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java
```diff
@@ -254,29 +273,64 @@ public RecordMaterializer<T> prepareForRead(Configuration configuration,
...
+  }
+
+  private static <T> ThriftRecordConverter<T> getRecordConverterInstance(
+      String converterClassName, Class<T> thriftClass,
+      MessageType requestedSchema, StructType descriptor, ParquetConfiguration conf) {
+    return getRecordConverterInstance(converterClassName, thriftClass, requestedSchema, descriptor, conf, ParquetConfiguration.class);
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <T1, T2> ThriftRecordConverter<T1> getRecordConverterInstance(
```
wgtmac, line 311: nit: could we give T1 and T2 better names?

wgtmac (commented):
Thanks for doing this! I have left some inline comments. 

BTW, I am also curious of the following things.
- It seems that we still have some compatibility issues. Can you confirm? If yes, could you please write them out explicitly?
- Is there any follow-up work item to do? Would be good if we can know the whole picture in advance.
- Is it possible to add a simple test case to prove that a simple writer and reader roundtrip can work successfully without Hadoop dependency?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -246,9 +262,9 @@ protected CompressionCodec getCodec(CompressionCodecName codecName) {
         codecClass = Class.forName(codecClassName);
       } catch (ClassNotFoundException e) {
         // Try to load the class using the job classloader
-        codecClass = configuration.getClassLoader().loadClass(codecClassName);
+        codecClass = new Configuration(false).getClassLoader().loadClass(codecClassName);
```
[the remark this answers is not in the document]
amousavigourabi: Here, the code tries to load a Hadoop codec from its classloader. For this, the Hadoop classloader is needed. This is not available by just using getClassLoader() on any ParquetConfiguration.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -246,9 +262,9 @@ protected CompressionCodec getCodec(CompressionCodecName codecName) {
...
       }
-      codec = (CompressionCodec) ReflectionUtils.newInstance(codecClass, configuration);
+      codec = (CompressionCodec) ReflectionUtils.newInstance(codecClass, ConfigurationUtil.createHadoopConfiguration(configuration));
```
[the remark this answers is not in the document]
amousavigourabi: This PR removes _part_ of the Hadoop dependency. For at least the codecs we still rely on them OOTB (see PARQUET-2353). AFAIK, using uncompressed codec should not lead to the Hadoop classes being loaded (which would require the runtime dependency).

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
amousavigourabi: For now this seems necessary to me in order not to introduce any breaking changes. AFAIK clients _could_ make suppClass be their own implementation, which under Parquet prior to this change would involve invoking setConf(Configuration) if that class were Hadoop Configurable (which is what Hadoop's ReflectionUtils.newInstance(Class, Configuration) also does under the hood).

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/AbstractThriftWriteSupport.java
```diff
@@ -40,14 +42,22 @@
...
+  public static void setGenericThriftClass(ParquetConfiguration configuration, Class<?> thriftClass) {
     conf = configuration;
     configuration.set(PARQUET_THRIFT_CLASS, thriftClass.getName());
   }
 
   public static Class getGenericThriftClass(Configuration configuration) {
+    return getGenericThriftClass(new HadoopParquetConfiguration(configuration));
+  }
+
+  public static Class<?> getGenericThriftClass(ParquetConfiguration configuration) {
```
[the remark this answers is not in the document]
amousavigourabi: `Class` is a generic type, raw use of which is discouraged. This is why I chose to give the generic `Class<?>` with a wildcard here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/ConfigurationUtil.java
```diff
@@ -41,4 +49,18 @@ public static Class<?> getClassFromConfig(Configuration configuration, String co
...
   }
 
+  public static Configuration createHadoopConfiguration(ParquetConfiguration conf) {
+    if (conf == null) {
+      return new Configuration();
+    }
+    if (conf instanceof HadoopParquetConfiguration) {
+      return ((HadoopParquetConfiguration) conf).getConfiguration();
+    }
+    Configuration configuration = new Configuration();
```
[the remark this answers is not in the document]
amousavigourabi: When using a HadoopParquetConfiguration, the user did not yet decouple from Hadoop as it is just a wrapper for Configuration. When the user wants to decouple from Hadoop, they can implement their own ParquetConfiguration, which does not rely on Hadoop's Configuration (or a simple implementation is added afterwards, this PR was getting a bit large for that already). There is still some code right now, mainly around the codecs which _needs_ a Hadoop Configuration. It is therefore important that while we're still removing these last references to Hadoop, we can get such an instance from a ParquetConfiguration, in order not to break anything.

parquet-pig/src/main/java/org/apache/parquet/pig/TupleReadSupport.java
```diff
@@ -154,9 +172,9 @@ private static FieldSchema union(FieldSchema mergedFieldSchema, FieldSchema newF
 
   @Override
   public ReadContext init(InitContext initContext) {
-    Schema pigSchema = getPigSchema(initContext.getConfiguration());
-    RequiredFieldList requiredFields = getRequiredFields(initContext.getConfiguration());
-    boolean columnIndexAccess = initContext.getConfiguration().getBoolean(PARQUET_COLUMN_INDEX_ACCESS, false);
+    Schema pigSchema = getPigSchema(initContext.getConfig());
```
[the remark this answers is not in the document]
amousavigourabi: Agreed

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java
amousavigourabi: Done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
amousavigourabi: The Hadoop specific stuff I agree that it would be a bit silly. However, this class `InternalParquetRecordReader` is part of the read/write API for which we're trying to address it. That the read/write API is part of parquet-hadoop is somewhat unfortunate, but changing it at this point would break stuff.

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
amousavigourabi: We still need a way to get the ParquetConfiguration to the HadoopCodecs, for this I used a field in ParquetReadOptions. To this effect, I would also like to start a conversation in the future about the options classes, as it seems to me we can replace them with the more versatile ParquetConfiguration in most, if not all places. This as I do think this current flow of passing the ParquetConfiguration through the ParquetReadOptions is a bit hacky.

amousavigourabi:
@wgtmac thanks a lot for the review! It's quite a big one so I really appreciate you took the time for it.

To address your concerns:

> * It seems that we still have some compatibility issues. Can you confirm? If yes, could you please write them out explicitly?

Correct, japicmp comes up with two incompatibilities. These are in the classes CodecFactory and ParquetReader. The incompatibilities it points to are both changes of private and protected field types from Configuration to ParquetConfiguration. These changes are strictly necessary for the effort to unhadoop the read/write API.

> * Is there any follow-up work item to do? Would be good if we can know the whole picture in advance.

After this, the following steps that will need to be taken are: 1. the creation of a simple unhadooped implementation of the ParquetConfiguration interface, and 2. adding a simple way for users to avoid the Hadoop codecs, as the OOTB implementations of everything still rely very heavily on Hadoop classes. These changes should allow users to drop the Hadoop runtime dependency. The Hadoop client API dependency will still be necessary.

> * Is it possible to add a simple test case to prove that a simple writer and reader roundtrip can work successfully without Hadoop dependency?

We do not yet have any serious ways for users to use the API without Hadoop dependencies. The added parameters to the TestReadWrite fixture make sure the read/write API should still function when using the ParquetConfiguration interface.

pom.xml
```diff
@@ -547,6 +547,8 @@
             </excludeModules>
             <excludes>
               <exclude>${shade.prefix}</exclude>
+              <exclude>org.apache.parquet.hadoop.CodecFactory</exclude> <!-- change field type from Configuration to ParquetConfiguration -->
```
wgtmac, line 550: Just curious: can we remove this in a future PR? Only this PR introduces the breaking change here.

wgtmac (approved):
Thanks for the detail explanation! This LGTM. Probably you can also create JIRA issues in advance for all planned work items.

I would listen to feedback from more maintainers to proceed. @gszadovszky @shangxinli @ggershinsky @Fokko

pom.xml
amousavigourabi: I'm not that well versed in how japicmp works exactly, but I reckon this can be removed after the next minor release.

wgtmac:
@shangxinli @gszadovszky Do you have any comment? If not, I plan to merge this next week. Thanks!

@amousavigourabi Could you please resolve the conflicts? Thank you!

gszadovszky:
@wgtmac, @amousavigourabi, sorry, but I won't have the time to properly review this PR in the upcoming days.
The concept looks good to me so I am giving a +0 for this patch.

parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.apache.parquet.conf;
+
+import java.util.Map;
+
+/**
+ * Configuration interface with the methods necessary to configure for Parquet applications.
+ */
+public interface ParquetConfiguration extends Iterable<Map.Entry<String,String>> {
```
ConeyLiu, line 27: Maybe add some docs for the methods?

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
```diff
@@ -19,9 +19,12 @@
 
 package org.apache.parquet;
 
+import org.apache.hadoop.conf.Configuration;
```
ConeyLiu, line 22: unused import

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
```diff
@@ -185,6 +232,31 @@ public static class Builder {
     protected int maxAllocationSize = ALLOCATION_SIZE_DEFAULT;
     protected Map<String, String> properties = new HashMap<>();
     protected FileDecryptionProperties fileDecryptionProperties = null;
+    protected ParquetConfiguration conf;
+
+    public Builder() {
```
ConeyLiu, line 237: Seems this constructor is not necessary.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DirectZstd.java
```diff
@@ -134,6 +144,10 @@ BytesInput getBytesInput() {
   }
 
   private static BufferPool getPool(Configuration conf) {
+    return getPool(new HadoopParquetConfiguration(conf));
+  }
+
+  private static BufferPool getPool(ParquetConfiguration conf) {
```
ConeyLiu, line 150: It should be safe to modify the existing method since it's a private method.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -211,6 +213,19 @@ private static UnboundRecordFilter getUnboundRecordFilterInstance(Configuration
     }
   }
 
+  private static UnboundRecordFilter getUnboundRecordFilterInstance(ParquetConfiguration configuration) {
```
ConeyLiu, line 216: Same here, we could change `getUnboundRecordFilterInstance` directly as well.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -276,15 +279,40 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
...
+      rowGroupSize,
+      validating,
+      new HadoopParquetConfiguration(conf),
+      maxPaddingSize,
+      encodingProps,
+      encryptionProperties);
+  }
+
+  ParquetWriter(
+    OutputFile file,
```
ConeyLiu, line 296: indent, it would be better to keep the same as the above method.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/ReadSupport.java
```diff
@@ -75,14 +76,32 @@ public ReadContext init(
...
+   * called in {@link org.apache.hadoop.mapreduce.InputFormat#getSplits(org.apache.hadoop.mapreduce.JobContext)} in the front end
+   *
+   * @param configuration    the configuration
+   * @param keyValueMetaData the app specific metadata from the file
+   * @param fileSchema       the schema of the file
+   * @return the readContext that defines how to read the file
+   *
+   * @deprecated override {@link ReadSupport#init(InitContext)} instead
+   */
+  @Deprecated
```
ConeyLiu, line 89: Do we still need to add a `deprecated` method?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/ReadSupport.java
```diff
@@ -101,6 +120,24 @@ abstract public RecordMaterializer<T> prepareForRead(
...
+   * @param fileSchema       the schema of the file
+   * @param readContext      returned by the init method
+   * @return the recordMaterializer that will materialize the records
+   */
+  public RecordMaterializer<T> prepareForRead(
+      ParquetConfiguration configuration,
+      Map<String, String> keyValueMetaData,
+      MessageType fileSchema,
+      ReadContext readContext) {
+    throw new UnsupportedOperationException("Override prepareForRead(ParquetConfiguration, Map<String, String>, MessageType, ReadContext)");
```
ConeyLiu, line 138: The error message should be more meaningful.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/WriteSupport.java
```diff
@@ -105,6 +106,15 @@ public Map<String, String> getExtraMetaData() {
    */
   public abstract WriteContext init(Configuration configuration);
 
+  /**
+   * called first in the task
+   * @param configuration the job's configuration
+   * @return the information needed to write the file
+   */
+  public WriteContext init(ParquetConfiguration configuration) {
+    throw new UnsupportedOperationException("Override init(ParquetConfiguration)");
```
ConeyLiu, line 115: Same here

parquet-avro/src/main/java/org/apache/parquet/avro/AvroReadSupport.java
```diff
@@ -154,6 +171,10 @@ private static <T> RecordMaterializer<T> newCompatMaterializer(
   }
 
   private GenericData getDataModel(Configuration conf, Schema schema) {
+    return getDataModel(new HadoopParquetConfiguration(conf), schema);
+  }
+
+  private GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
```
ConeyLiu, line 177: Same here, we could update the existing private method directly.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -405,6 +413,10 @@ private Binary fromAvroString(Object value) {
   }
 
   private static GenericData getDataModel(Configuration conf, Schema schema) {
+    return getDataModel(new HadoopParquetConfiguration(conf), schema);
+  }
+
+  private static GenericData getDataModel(ParquetConfiguration conf, Schema schema) {
```
ConeyLiu, line 419: same here

parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftRecordConverter.java
```diff
@@ -855,6 +857,18 @@ public ThriftRecordConverter(ThriftReader<T> thriftReader, String name, MessageT
    * @param conf a Configuration
    */
   public ThriftRecordConverter(ThriftReader<T> thriftReader, String name, MessageType requestedParquetSchema, ThriftType.StructType thriftType, Configuration conf) {
+    this(thriftReader, name, requestedParquetSchema, thriftType, new HadoopParquetConfiguration(conf));
+  }
+
+  /**
+   *
```
ConeyLiu, line 864: blank line

ConeyLiu:
@amousavigourabi we have added so many public methods to keep the backward compatibility. So which one is preferred (I think it should be the`ParquetConfiguration`)? Should we deprecate the old one?

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
amousavigourabi: As this is public, removing the implicit no-args constructor _could_ potentially cause problems for users.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/ReadSupport.java
amousavigourabi: This PR is focussed on transitioning from `Configuration` to the `ParquetConfiguration` interface. This included some calls to deprecated methods which I could not very quickly transition away from. I would consider this out-of-scope for this PR.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/ReadSupport.java
amousavigourabi: I follow the example set by `ReadSupport#init(Configuration, Map, MessageType)`. As this error will not occur unless you are implementing your own `ReadSupport` class, I am not sure whether there needs to be that much more information in the exception. I'll add a reference to the `ReadSupport` class though.

amousavigourabi:
> @amousavigourabi we have added so many public methods to keep the backward compatibility. So which one is preferred (I think it should be the`ParquetConfiguration`)? Should we deprecate the old one?

the `ParquetConfiguration` methods are indeed preferred. I would be in favour of deprecating the Hadoop `Configuration` methods. However, I do think deprecating these methods might warrant a separate discussion on the mailing list (as their deprecation and subsequent removal would go further than what we have previously discussed, it would be a step in the direction of eliminating the need for `hadoop-client-api` altogether instead of just `hadoop-client-runtime`) and does not necessarily need to be a part of this PR.

wgtmac:
Thank you for the review @ConeyLiu. This PR is one of the efforts to use the parquet library without the Hadoop dependency. At the moment, we do not want to deprecate Hadoop dependency yet.

parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
ConeyLiu: I see the usage, that makes sense.

ConeyLiu (approved):
Thanks, @amousavigourabi for the response and your efforts. LGTM.

> This PR is one of the efforts to use the parquet library without the Hadoop dependency. At the moment, we do not want to deprecate Hadoop dependency yet.

@wgtmac This is a good start. It would be nice to have a best practice guide doc to help users understand the different configuration usage.

wgtmac:
I totally agree with you. @ConeyLiu 
We might need to consolidate different docs for configuration like this (https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/README.md) into a better space.

parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Sets the value of the name property.
+   *
+   * @param name the property to set
+   * @param value the value to set the property to
+   */
+  void set(String name, String value);
```
shangxinli, line 35: setString()?

amousavigourabi: I just went with the way Hadoop's `Configuration` does it for this and other methods in the interface, in order to not have to change references to conf.set() everywhere.

wgtmac (approved):

wgtmac:
I will merge this if there is no more comment by the end of this week. Thanks @amousavigourabi for working on this!

ConeyLiu:
This broke the CI checking.

parquet-common/src/main/java/org/apache/parquet/util/Reflection.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Lifted from Hadoop's org.apache.hadoop.util.ReflectionUtils.
+ */
+public class Reflection {
+
+  private static final Class<?>[] EMPTY_ARRAY = new Class[]{};
+  private static final Map<Class<?>, Constructor<?>> CONSTRUCTOR_CACHE = new ConcurrentHashMap<Class<?>, Constructor<?>>();
+
+  @SuppressWarnings("unchecked")
+  public static <T> T newInstance(Class<T> theClass) {
```
ConeyLiu, line 35: Not sure why this breaks the API checking. However, this class seems redundant since there is not any usage.

wgtmac: Hmm, this is strange. Should we exclude this class in the japicmp config? @amousavigourabi

amousavigourabi: As there is no current usage, we can get away with deleting this class for now. My main concern with the CI failure is the incompatibility at `org.apache.parquet.conf.ParquetConfiguration.getClass(java.lang.String,java.lang.Class,java.lang.Class):CLASS_GENERIC_TEMPLATE_CHANGED` japicmp complains about. Especially as this did not come up before and the interface is completely new I have no idea why this is happening. Neither of the matters japicmp flags should be an issue AFAIK, so that might be something to investigate in our configuration?

ConeyLiu: I think this should be a bug of japicmp. Because both of the methods are generic template methods. And it should be safe to suppress the checking.

amousavigourabi: I'll open a new PR shortly, deleting the unused stuff and suppressing the warnings.

amousavigourabi: It has also started failing at other completely new methods in the Parquet Hadoop module: `org.apache.parquet.conf.HadoopParquetConfiguration.getClass(java.lang.String,java.lang.Class,java.lang.Class):CLASS_GENERIC_TEMPLATE_CHANGED,org.apache.parquet.hadoop.util.SerializationUtil.readObjectFromConfAsBase64(java.lang.String,org.apache.parquet.conf.ParquetConfiguration):CLASS_GENERIC_TEMPLATE_CHANGED`. Might be something to report to the japicmp maintainer?

ConeyLiu: I tested 0.16.0 has no problem. A little strange.

wgtmac: Probably the current 0.18.1 is more aggressive than 0.16.0

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -48,7 +51,7 @@ public class CodecFactory implements CompressionCodecFactory {
   private final Map<CompressionCodecName, BytesCompressor> compressors = new HashMap<>();
   private final Map<CompressionCodecName, BytesDecompressor> decompressors = new HashMap<>();
 
-  protected final Configuration configuration;
```
Fokko, line 51: This causes the Iceberg build to break when bumping to 1.14.0-SNAPSHOT:

```
> Task :iceberg-parquet:compileJava FAILED
/Users/fokkodriesprong/Desktop/iceberg/parquet/src/main/java/org/apache/iceberg/parquet/ParquetCodecFactory.java:60: error: cannot find symbol
        codecClass = configuration.getClassLoader().loadClass(codecClassName);
                                  ^
  symbol:   method getClassLoader()
  location: variable configuration of type ParquetConfiguration
/Users/fokkodriesprong/Desktop/iceberg/parquet/src/main/java/org/apache/iceberg/parquet/ParquetCodecFactory.java:62: error: incompatible types: ParquetConfiguration cannot be converted to Configuration
      codec = (CompressionCodec) ReflectionUtils.newInstance(codecClass, configuration);
```

wgtmac: Does it make sense to fix this on the iceberg side?

wgtmac: Anyway this is a breaking change. We have to fix it before the release.

Fokko: For Iceberg this is not an issue, since we would remove this class with the 1.14 release (that contains https://github.com/apache/parquet-mr/pull/1134): 
https://github.com/apache/iceberg/blob/866021d7d34f274349ce7de1f29d113395e7f28c/parquet/src/main/java/org/apache/iceberg/parquet/ParquetCodecFactory.java#L28-L32

In Iceberg we shade Parquet, so this will not interfere with other libraries that use a different version of Parquet. I'm curious what if we also have this issue on the Spark side (cc @vinooganesh).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -193,18 +195,13 @@ public static Class<?> getUnboundRecordFilter(Configuration configuration) {
...
-    if (clazz == null) { return null; }
-
+    if (clazz == null) {
+      return null;
+    }
     try {
-      UnboundRecordFilter unboundRecordFilter = (UnboundRecordFilter) clazz.newInstance();
-
-      if (unboundRecordFilter instanceof Configurable) {
-        ((Configurable)unboundRecordFilter).setConf(configuration);
```
shangxinli, line 204: @amousavigourabi  Can you share more details on why this piece of code is removed?

amousavigourabi: looking back, i haven't a clue

PR #1156, opened 2023-09-27
Bump Elephantbird

Fokko:
Not sure if we just should remove it (that would break quite a few APIs, probably something for Parquet 2.0), or update it like this PR. It uses an ancient version:

```
4.4  -> Thrift 0.7.0 😱
4.17 -> Thrift 0.10.0
```

`>4.4` makes a lot of test helpers private, so I had to remove a few tests.

### Jira

### Tests

### Commits

### Documentation


wgtmac:
Does an old version of elephantbird prohibit us from upgrading thrift in the future? Even the latest 4.17 is pretty old. Removing these tests may add more risk to break things though I'm not sure whether these modules are still actively used.

Fokko:
It blocks the upgrade. Somewhere in a static block an exception is being thrown. 

>  Removing these tests may add more risk to break things though I'm not sure whether these modules are still actively used.

The artifact is [littered with CVE's](https://mvnrepository.com/artifact/com.twitter.elephantbird/elephant-bird/4.17), so I hope no-one is using this still.

`apache-thrift` does not seem to be used anymore (usages of 0 since 1.12.2):
image

`apache-protobuf` does have some usage:
image

So we can leave that in for now. Let me prepare a PR and also sent out an email on the dev-list.

Fokko:
I don't think this is a feasible option, let me close it

PR #1157, opened 2023-09-27
PARQUET-2351: Set options with Configuration

amousavigourabi:

### Jira

### Tests

       Parameterizes existing fixture to instantiate writer options using Configuration instead.

### Commits

### Documentation

---

The functionality is implemented in a way in which options set using the `builder.withXXX(...)` method will _always_ override the options passed by the configuration. This is done in order to not make things break in unexpected ways. Having a configuration which used to not have any effect on these options override an explicit option would be a bit odd after all.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
@@ -52,11 +52,11 @@ import org.slf4j.LoggerFactory;
 
 /**
  * OutputFormat to write to a Parquet file
- *
+ * <p>
  * It requires a {@link WriteSupport} to convert the actual records to the underlying format.
  * It requires the schema of the incoming records. (provided by the write support)
  * It allows storing extra metadata in the footer (for example: for schema compatibility purpose when converting from a different schema language).
- *
+ * <p>
  * The format configuration settings in the job configuration:
  * <pre>
  * # The block size is the size of a row group being buffered in memory
@@ -106,7 +106,7 @@ import org.slf4j.LoggerFactory;
 public class ParquetOutputFormat<T> extends FileOutputFormat<Void, T> {
   private static final Logger LOG = LoggerFactory.getLogger(ParquetOutputFormat.class);
 
-  public static enum JobSummaryLevel {
+  public enum JobSummaryLevel {
     /**
      * Write no summary files
      */
@@ -130,7 +130,7 @@ public class ParquetOutputFormat<T> extends FileOutputFormat<Void, T> {
   public static final String ENABLE_JOB_SUMMARY   = "parquet.enable.summary-metadata";
 
   /**
-   * Must be one of the values in {@link JobSummaryLevel} (case insensitive)
+   * Must be one of the values in {@link JobSummaryLevel} (case-insensitive)
    */
   public static final String JOB_SUMMARY_LEVEL = "parquet.summary.metadata.level";
   public static final String BLOCK_SIZE           = "parquet.block.size";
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -358,9 +358,29 @@ public abstract static class Builder<T, SELF extends Builder<T, SELF>> {
     private long rowGroupSize = DEFAULT_BLOCK_SIZE;
     private int maxPaddingSize = MAX_PADDING_SIZE_DEFAULT;
     private boolean enableValidation = DEFAULT_IS_VALIDATING_ENABLED;
-    private ParquetProperties.Builder encodingPropsBuilder =
+    private final ParquetProperties.Builder encodingPropsBuilder =
         ParquetProperties.builder();
 
+    private boolean isPageSizeSet = false;
```
wgtmac, line 364: TBH, these would make the code less maintainable. Not sure if Optional would make them more organized.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
```diff
@@ -195,10 +200,28 @@ private static void prepareFile(WriterVersion version, Path file) throws IOExcep
     writeData(f, writer);
   }
 
+  private static void prepareFileWithConf(WriterVersion version, Path file) throws IOException {
+    Configuration configuration = new Configuration();
```
wgtmac, line 204: It seems that these specific configurations defined in the ParquetOutputFormat are solely used for ParquetOutputFormat to create a RecordWriter, which actually puts all of them into a ParquetProperties.

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java#L457-L484

IIUC, relying on settings from Hadoop configuration is discouraged, we should use ParquetProperties to set all those things.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
amousavigourabi: I agree that these booleans aren't amazing. Optional fields would not solve the issue of making withXXX methods always override the configuration, next to this, Optionals shouldn't be used as fields IMO, as that is [not their intended use](https://mail.openjdk.org/pipermail/jdk8-dev/2013-September/003274.html). Maybe an approach with a Set would work better?

parquet-hadoop/src/test/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilterTest.java
amousavigourabi: These configurations are used to build the `encodingProps` ParquetProperties within ParquetWriter's Builder, which are used elsewhere as well. While the usage of Configuration in this way may be discouraged, the current situation of settings sometimes not being picked up by the ParquetWriter is inconsistent with both ParquetReader behaviour and user expectations.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -341,7 +341,7 @@ public static void setMaxPaddingSize(Configuration conf, int maxPaddingSize) {
     conf.setInt(MAX_PADDING_BYTES, maxPaddingSize);
   }
 
-  private static int getMaxPaddingSize(Configuration conf) {
+  public static int getMaxPaddingSize(Configuration conf) {
```
shangxinli, line 344: Changing method from private to public would need to add documentation and unit tests

amousavigourabi: 👍 I'll add it for the rest of the publics in the class as well while I'm at it.

PR #1165, opened 2023-10-03
PARQUET-2357: Modest refactor of CapacityByteArrayOutputStream

fengjiajie:
Optimization for the CapacityByteArrayOutputStream class:

1. The functionality of **currentSlabIndex** is the same as **currentSlab.position()**, so there is no need to maintain the **currentSlabIndex** variable.
2. When writing an array of length equal to the remaining capacity of the buffer, there is no need to expand to a new buffer.
3. If the **addSlab** operation has already implemented safeguards using **Math.addExact** to prevent overflow of **bytesAllocated** and **bytesUsed**, it is unnecessary to perform additional checks during the **write** operation.

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references them in the PR title. For example, "PARQUET-2357: Modest refactor of CapacityByteArrayOutputStream"
  - https://issues.apache.org/jira/browse/PARQUET-2357

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
@@ -61,7 +61,6 @@ public class CapacityByteArrayOutputStream extends OutputStream {
   private final List<ByteBuffer> slabs = new ArrayList<ByteBuffer>();
 
   private ByteBuffer currentSlab;
-  private int currentSlabIndex;
   private int bytesAllocated = 0;
   private int bytesUsed = 0;
   private ByteBufferAllocator allocator;
@@ -193,7 +192,6 @@ public class CapacityByteArrayOutputStream extends OutputStream {
     this.currentSlab = allocator.allocate(nextSlabSize);
     this.slabs.add(currentSlab);
     this.bytesAllocated = Math.addExact(this.bytesAllocated, nextSlabSize);
-    this.currentSlabIndex = 0;
   }
 
   @Override
@@ -201,9 +199,7 @@ public class CapacityByteArrayOutputStream extends OutputStream {
     if (!currentSlab.hasRemaining()) {
       addSlab(1);
     }
-    currentSlab.put(currentSlabIndex, (byte) b);
-    currentSlabIndex += 1;
-    currentSlab.position(currentSlabIndex);
+    currentSlab.put((byte) b);
     bytesUsed = Math.addExact(bytesUsed, 1);
   }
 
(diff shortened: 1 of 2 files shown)


parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -193,18 +192,15 @@ private void addSlab(int minimumSize) {
...
   public void write(int b) {
     if (!currentSlab.hasRemaining()) {
       addSlab(1);
     }
-    currentSlab.put(currentSlabIndex, (byte) b);
-    currentSlabIndex += 1;
-    currentSlab.position(currentSlabIndex);
-    bytesUsed = Math.addExact(bytesUsed, 1);
+    currentSlab.put((byte) b);
+    bytesUsed += 1;
```
wgtmac, line 203:
```suggestion
    bytesUsed = Math.addExact(bytesUsed, 1);
```

parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
```diff
@@ -214,21 +210,16 @@ public void write(byte b[], int off, int len) {
...
       addSlab(length2);
       currentSlab.put(b, off + length1, length2);
-      currentSlabIndex = length2;
-      bytesUsed = Math.addExact(bytesUsed, length2);
     } else {
       currentSlab.put(b, off, len);
-      currentSlabIndex += len;
-      bytesUsed = Math.addExact(bytesUsed, len);
     }
+    bytesUsed += len;
```
wgtmac, line 222:
```suggestion
    bytesUsed = Math.addExact(bytesUsed, len);
```

parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
wgtmac: It would be good to keep the overflow check, just in case.

parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
wgtmac: ditto

wgtmac (changes requested):
Thanks for the change! I have left a minor comment.

fengjiajie:
@wgtmac  Thank you for your review. 

The reason why write() no longer checks for overflow is because addSlab() ensures that bytesAllocated will not overflow.
In the subsequent write() function, it is guaranteed that **bytesAllocated >= bytesUsed**, and thus bytesUsed will not overflow.

But I agree that it would be advantageous to add the check here as well, I have changed the commit.

wgtmac (approved):
Thanks! LGTM

PR #1170, opened 2023-10-14
PARQUET-2361: Reduce failure rate of unit test

fengjiajie:
Reduce failure rate of unit test testParquetFileWithBloomFilterWithFpp

Change-Id: Ic230f197b0996333a082bb05bd201963d05d862e

```
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    TestParquetWriter.testParquetFileWithBloomFilterWithFpp:342
[INFO]  
```

Multiple different PR triggered this failure:

1. https://github.com/apache/parquet-mr/pull/1062
2. https://github.com/apache/parquet-mr/actions/runs/5420924489/job/14680382407
3. https://github.com/apache/parquet-mr/actions/runs/6336014897
4. https://github.com/apache/parquet-mr/actions/runs/6381223319
5. https://github.com/apache/parquet-mr/actions/runs/6394826826/job/17357106390

I found two issues that may cause the failures to occur easily:

1. This may be a bug. The 'distinctStrings' is used to generate test files, but it is cleared in the first round. Then it is used for probability testing, so in the next round of generating 'fpp' files, it uses the data from the previous round of testing, and the length of the filled strings (length 10 for test not in bloomfilter) is also different from the initial one (strings length 12 for build bloomfilter). 
2. The number of test iterations is insufficient, resulting in an unstable probability.

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2361

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
@@ -288,14 +288,15 @@ public class TestParquetWriter {
 
   @Test
   public void testParquetFileWithBloomFilterWithFpp() throws IOException {
-    int totalCount = 100000;
-    double[] testFpp = {0.01, 0.05, 0.10, 0.15, 0.20, 0.25};
+    int buildBloomFilterCount = 100000;
+    double[] testFpps = {0.01, 0.05, 0.10, 0.15, 0.20, 0.25};
     int randomStrLen = 12;
+    final int testBloomFilterCount = 200000;
 
-    Set<String> distinctStrings = new HashSet<>();
-    while (distinctStrings.size() < totalCount) {
+    Set<String> distinctStringsForFileGenerate = new HashSet<>();
+    while (distinctStringsForFileGenerate.size() < buildBloomFilterCount) {
       String str = RandomStringUtils.randomAlphabetic(randomStrLen);
-      distinctStrings.add(str);
+      distinctStringsForFileGenerate.add(str);
     }
 
     MessageType schema = Types.buildMessage().
@@ -305,7 +306,7 @@ public class TestParquetWriter {
     GroupWriteSupport.setSchema(schema, conf);
 
     GroupFactory factory = new SimpleGroupFactory(schema);
-    for (int i = 0; i < testFpp.length; i++) {
+    for (double testFpp : testFpps) {
       File file = temp.newFile();
       file.delete();
       Path path = new Path(file.getAbsolutePath());
@@ -314,32 +315,32 @@ public class TestParquetWriter {
         .withConf(conf)
         .withDictionaryEncoding(false)
         .withBloomFilterEnabled("name", true)
-        .withBloomFilterNDV("name", totalCount)
-        .withBloomFilterFPP("name", testFpp[i])
+        .withBloomFilterNDV("name", buildBloomFilterCount)
+        .withBloomFilterFPP("name", testFpp)
         .build()) {
-        java.util.Iterator<String> iterator = distinctStrings.iterator();
-        while (iterator.hasNext()) {
-          writer.write(factory.newGroup().append("name", iterator.next()));
+        for (String str : distinctStringsForFileGenerate) {
+          writer.write(factory.newGroup().append("name", str));
         }
       }
-      distinctStrings.clear();
 
       try (ParquetFileReader reader = ParquetFileReader.open(HadoopInputFile.fromPath(path, new Configuration()))) {
         BlockMetaData blockMetaData = reader.getFooter().getBlocks().get(0);
         BloomFilter bloomFilter = reader.getBloomFilterDataReader(blockMetaData)
           .readBloomFilter(blockMetaData.getColumns().get(0));
 
-        // The exist counts the number of times FindHash returns true.
-        int exist = 0;
-        while (distinctStrings.size() < totalCount) {
-          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 2);
-          if (distinctStrings.add(str) &&
+        // The false positive counts the number of times FindHash returns true.
+        int falsePositive = 0;
+        Set<String> distinctStringsForProbe = new HashSet<>();
+        while (distinctStringsForProbe.size() < testBloomFilterCount) {
+          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 1);
+          if (distinctStringsForProbe.add(str) &&
             bloomFilter.findHash(LongHashFunction.xx(0).hashBytes(Binary.fromString(str).toByteBuffer()))) {
-            exist++;
+            falsePositive++;
           }
         }
-        // The exist should be less than totalCount * fpp. Add 10% here for error space.
-        assertTrue(exist < totalCount * (testFpp[i] * 1.1) && exist > 0);
+        // The false positive should be less than totalCount * fpp. Add 15% here for error space.
+        double expectedFalsePositiveMaxCount = Math.floor(testBloomFilterCount * (testFpp * 1.15));
+        assertTrue(falsePositive < expectedFalsePositiveMaxCount && falsePositive > 0);
       }
     }
   }


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -305,41 +306,41 @@ public void testParquetFileWithBloomFilterWithFpp() throws IOException {
     GroupWriteSupport.setSchema(schema, conf);
 
     GroupFactory factory = new SimpleGroupFactory(schema);
-    for (int i = 0; i < testFpp.length; i++) {
+    for (double testFpp : testFpps) {
       File file = temp.newFile();
-      file.delete();
+      assertTrue(file.delete());
```
amousavigourabi, line 311: Do we need this assertion? AFAIK this would only trigger when there is some I/O failure and does not help evaluate the correctness of the code.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -305,41 +306,41 @@ public void testParquetFileWithBloomFilterWithFpp() throws IOException {
...
+          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 1);
+          if (distinctStringsForProbe.add(str) &&
             bloomFilter.findHash(LongHashFunction.xx(0).hashBytes(Binary.fromString(str).toByteBuffer()))) {
-            exist++;
+            falsePositive++;
           }
         }
         // The exist should be less than totalCount * fpp. Add 10% here for error space.
-        assertTrue(exist < totalCount * (testFpp[i] * 1.1) && exist > 0);
+        int expectedFalsePositiveMaxCount = (int) Math.floor(testBloomFilterCount * (testFpp * 1.1));
```
amousavigourabi, line 342: We could avoid the cast and redundant parentheses as follows:

```suggestion
        double expectedFalsePositiveMaxCount = Math.floor(testBloomFilterCount * testFpp * 1.1);
```

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -305,41 +306,41 @@ public void testParquetFileWithBloomFilterWithFpp() throws IOException {
...
         BlockMetaData blockMetaData = reader.getFooter().getBlocks().get(0);
         BloomFilter bloomFilter = reader.getBloomFilterDataReader(blockMetaData)
           .readBloomFilter(blockMetaData.getColumns().get(0));
 
         // The exist counts the number of times FindHash returns true.
-        int exist = 0;
-        while (distinctStrings.size() < totalCount) {
-          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 2);
-          if (distinctStrings.add(str) &&
+        int falsePositive = 0;
```
amousavigourabi, line 332: Could we revert the name for this back to `exist` ? I don't think these are necessarily _all_ false positives.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -305,41 +306,41 @@ public void testParquetFileWithBloomFilterWithFpp() throws IOException {
...
+        Set<String> distinctStringsForProbe = new HashSet<>();
+        while (distinctStringsForProbe.size() < testBloomFilterCount) {
+          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 1);
+          if (distinctStringsForProbe.add(str) &&
             bloomFilter.findHash(LongHashFunction.xx(0).hashBytes(Binary.fromString(str).toByteBuffer()))) {
-            exist++;
+            falsePositive++;
           }
         }
         // The exist should be less than totalCount * fpp. Add 10% here for error space.
```
amousavigourabi, line 341: Could you update the comments that refer to old variable names?

amousavigourabi (changes requested):

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
fengjiajie: I think the variable name should not be "exist." This unit test is to test if "fpp" meets the expected criteria. Here, all satisfying conditions are false positives of the bloom filter, then compare if "fpp" and "fp" are consistent.

amousavigourabi (approved):

fengjiajie:
After this modification, the unit tests are still failing. It seems that we need to explore other approaches, such as fixing the random seed or increasing the tolerance over 10% (current fpp * 1.1)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -314,32 +315,32 @@ public void testParquetFileWithBloomFilterWithFpp() throws IOException {
...
-        // The exist counts the number of times FindHash returns true.
-        int exist = 0;
-        while (distinctStrings.size() < totalCount) {
-          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 2);
-          if (distinctStrings.add(str) &&
+        // The false positive counts the number of times FindHash returns true.
+        int falsePositive = 0;
+        Set<String> distinctStringsForProbe = new HashSet<>();
+        while (distinctStringsForProbe.size() < testBloomFilterCount) {
+          String str = RandomStringUtils.randomAlphabetic(randomStrLen - 1);
```
amousavigourabi, line 335: Is there any reason this cannot be `randomStrLen` by the way?

fengjiajie: @amousavigourabi  The purpose of unit testing is to ensure that the false positive rate of the bloom filter meets expectations. The current approach involves generating a bloom filter using many strings of length 12. Any data with a length other than 12 is guaranteed to not exist in the original data. For this data (length != 12), if the bloom filter returns true, it is considered a false positive. The false positive rate is then calculated by examining these cases. Alternatively, if we want to use strings of length 12 for testing, we need to randomly generate strings and check if they exist in the original data. Only the ones that do not exist can be used to test the false positive rate.

amousavigourabi:
@fengjiajie As the test is now evaluating the false positive rate with significantly more samples than what we use to build the filter, or are provided as NDV, might it not be the case that this will increase the false positive rate of the bloom filter to more than the FPP? Perhaps we could try increasing the NDV, or maybe an adaptive bloom filter might be more appropriate? WDYT?

amousavigourabi:
> @fengjiajie As the test is now evaluating the false positive rate with significantly more samples than what we use to build the filter, or are provided as NDV, might it not be the case that this will increase the false positive rate of the bloom filter to more than the FPP? Perhaps we could try increasing the NDV, or maybe an adaptive bloom filter might be more appropriate? WDYT?

Nevermind, I don't think this should be an issue. I'll be running the test locally in a loop to see if I can reproduce the flake and check how often it might occur.

amousavigourabi:
> > @fengjiajie As the test is now evaluating the false positive rate with significantly more samples than what we use to build the filter, or are provided as NDV, might it not be the case that this will increase the false positive rate of the bloom filter to more than the FPP? Perhaps we could try increasing the NDV, or maybe an adaptive bloom filter might be more appropriate? WDYT?
> 
> Nevermind, I don't think this should be an issue. I'll be running the test locally in a loop to see if I can reproduce the flake and check how often it might occur.

I got 4 failures in 10k runs. This would mean 4 failures in 1250 full actions runs. These failures were all very slightly out of the expected range for the 0.01 fpp case. Given that we take 200k samples, this might indicate a flaw in the code, as 200k samples of some i.i.d. random variable ~ Bern(0.01) really should not have over 2200 hits that often, if ever. If we want to just fix this test I suggest raising the tolerance to 15%. That _should_ keep it from failing.

fengjiajie:
> > > @fengjiajie As the test is now evaluating the false positive rate with significantly more samples than what we use to build the filter, or are provided as NDV, might it not be the case that this will increase the false positive rate of the bloom filter to more than the FPP? Perhaps we could try increasing the NDV, or maybe an adaptive bloom filter might be more appropriate? WDYT?
> > 
> > 
> > Nevermind, I don't think this should be an issue. I'll be running the test locally in a loop to see if I can reproduce the flake and check how often it might occur.
> 
> I got 4 failures in 10k runs. This would mean 4 failures in 1250 full actions runs. These failures were all very slightly out of the expected range for the 0.01 fpp case. Given that we take 200k samples, this might indicate a flaw in the code, as 200k samples of some i.i.d. random variable ~ Bern(0.01) really should not have over 2200 hits that often, if ever. If we want to just fix this test I suggest raising the tolerance to 15%. That _should_ keep it from failing.

@amousavigourabi  I agree with increasing fault tolerance. Thank you very much for your review and testing.

wgtmac (approved):

PR #1173, opened 2023-10-17
PARQUET-2365 : Fixes NPE when rewriting column without column index

ConeyLiu:

### Jira

### Tests

### Commits

### Documentation

The ColumnIndex could be null in some scenes, for example, the float/double column contains NaN or the size has exceeded the expected value. And the page header statistics are not written anymore after we supported ColumnIndex. So we will get NPE when rewriting the column without ColumnIndex due to we will get NULL page statistics when converted from the ColumnIndex(NULL) or page header statistics(NULL). Such as the following:
```java
java.lang.NullPointerException
    at org.apache.parquet.hadoop.ParquetFileWriter.writeDataPage(ParquetFileWriter.java:727)
    at org.apache.parquet.hadoop.ParquetFileWriter.innerWriteDataPage(ParquetFileWriter.java:663)
    at org.apache.parquet.hadoop.ParquetFileWriter.writeDataPage(ParquetFileWriter.java:650)
    at org.apache.parquet.hadoop.rewrite.ParquetRewriter.processChunk(ParquetRewriter.java:453)
    at org.apache.parquet.hadoop.rewrite.ParquetRewriter.processBlocksFromReader(ParquetRewriter.java:317)
    at org.apache.parquet.hadoop.rewrite.ParquetRewriter.processBlocks(ParquetRewriter.java:250)

```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
@@ -451,15 +451,12 @@ public abstract class ColumnIndexBuilder {
     }
   };
 
+  private PrimitiveType type;
   private final BooleanList nullPages = new BooleanArrayList();
   private final LongList nullCounts = new LongArrayList();
-  private final IntList pageIndexes = new IntArrayList();
-
-  private PrimitiveType type;
   private long minMaxSize;
+  private final IntList pageIndexes = new IntArrayList();
   private int nextPageIndex;
-
-  protected boolean invalid;
 
   /**
    * @return a no-op builder that does not collect statistics objects and therefore returns {@code null} at
@@ -546,11 +543,6 @@ public abstract class ColumnIndexBuilder {
    *          the statistics to be added
    */
   public void add(Statistics<?> stats) {
-    if (stats.isEmpty()) {
-      invalid = true;
-      return;
-    }
-
     if (stats.hasNonNullValue()) {
       nullPages.add(false);
       Object min = stats.genericGetMin();
@@ -611,7 +603,7 @@ public abstract class ColumnIndexBuilder {
   }
 
   private ColumnIndexBase<?> build(PrimitiveType type) {
-    if (nullPages.isEmpty() || invalid) {
+    if (nullPages.isEmpty()) {
       return null;
     }
     ColumnIndexBase<?> columnIndex = createColumnIndex(type);
--- a/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
@@ -451,12 +451,15 @@ public abstract class ColumnIndexBuilder {
     }
   };
 
-  private PrimitiveType type;
   private final BooleanList nullPages = new BooleanArrayList();
   private final LongList nullCounts = new LongArrayList();
-  private long minMaxSize;
   private final IntList pageIndexes = new IntArrayList();
+
+  private PrimitiveType type;
+  private long minMaxSize;
   private int nextPageIndex;
+
+  protected boolean invalid;
 
   /**
    * @return a no-op builder that does not collect statistics objects and therefore returns {@code null} at
@@ -543,6 +546,11 @@ public abstract class ColumnIndexBuilder {
    *          the statistics to be added
    */
   public void add(Statistics<?> stats) {
+    if (stats.isEmpty()) {
+      invalid = true;
+      return;
+    }
+
     if (stats.hasNonNullValue()) {
       nullPages.add(false);
       Object min = stats.genericGetMin();
@@ -603,7 +611,7 @@ public abstract class ColumnIndexBuilder {
   }
 
   private ColumnIndexBase<?> build(PrimitiveType type) {
-    if (nullPages.isEmpty()) {
+    if (nullPages.isEmpty() || invalid) {
       return null;
     }
     ColumnIndexBase<?> columnIndex = createColumnIndex(type);
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -152,6 +152,7 @@ public class ParquetFileWriter {
   private long uncompressedLength;
   private long compressedLength;
   private Statistics<?> currentStatistics; // accumulated in writePage(s)
+  private boolean currentStatisticsAreValid;
   private ColumnIndexBuilder columnIndexBuilder;
   private OffsetIndexBuilder offsetIndexBuilder;
 
@@ -474,6 +475,7 @@ public class ParquetFileWriter {
     uncompressedLength = 0;
     // The statistics will be copied from the first one added at writeDataPage(s) so we have the correct typed one
     currentStatistics = null;
+    currentStatisticsAreValid = true;
 
     columnIndexBuilder = ColumnIndexBuilder.getBuilder(currentChunkType, columnIndexTruncateLength);
     offsetIndexBuilder = OffsetIndexBuilder.getBuilder();
@@ -722,14 +724,7 @@ public class ParquetFileWriter {
     LOG.debug("{}: write data page content {}", out.getPos(), compressedPageSize);
     bytes.writeAllTo(out);
 
-    // Copying the statistics if it is not initialized yet so we have the correct typed one
-    if (currentStatistics == null) {
-      currentStatistics = statistics.copy();
-    } else {
-      currentStatistics.mergeStatistics(statistics);
-    }
-
-    columnIndexBuilder.add(statistics);
+    mergeColumnStatistics(statistics);
 
     encodingStatsBuilder.addDataEncoding(valuesEncoding);
     currentEncodings.add(rlEncoding);
@@ -867,13 +862,8 @@ public class ParquetFileWriter {
     this.uncompressedLength += uncompressedSize + headersSize;
     this.compressedLength += compressedSize + headersSize;
 
-    if (currentStatistics == null) {
-      currentStatistics = statistics.copy();
-    } else {
-      currentStatistics.mergeStatistics(statistics);
-    }
+    mergeColumnStatistics(statistics);
 
-    columnIndexBuilder.add(statistics);
     currentEncodings.add(dataEncoding);
     encodingStatsBuilder.addDataEncoding(dataEncoding);
 
@@ -1000,6 +990,7 @@ public class ParquetFileWriter {
     currentStatistics = totalStatistics;
     // Invalid the ColumnIndex
     columnIndexBuilder = ColumnIndexBuilder.getNoOpBuilder();
+    currentStatisticsAreValid = true;
     endColumn();
   }
 
@@ -1010,6 +1001,7 @@ public class ParquetFileWriter {
   public void endColumn() throws IOException {
     state = state.endColumn();
     LOG.debug("{}: end column", out.getPos());
+    Preconditions.checkState(currentStatisticsAreValid, "Column statistics should be valid");
     if (columnIndexBuilder.getMinMaxSize() > columnIndexBuilder.getPageCount() * MAX_STATS_SIZE) {
       currentColumnIndexes.add(null);
     } else {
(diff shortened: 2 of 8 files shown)


ConeyLiu:
Hi @wgtmac, please help to review this when you are free. Thanks a lot.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -543,6 +546,11 @@ public static ColumnIndex build(
    *          the statistics to be added
    */
   public void add(Statistics<?> stats) {
+    if (stats.isEmpty()) {
```
ConeyLiu, line 549: It should be invalid if the stats are empty. Previously we set it as a null page. @gszadovszky please correct me if I am wrong.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -988,6 +988,21 @@ void writeColumnChunk(ColumnDescriptor descriptor,
...
   }
 
+  /**
+   * Overwrite the column total statistics. This special used when the column total statistics
+   * is known while all the page statistics are invalid, for example when rewriting the column.
+   *
+   * @param totalStatistics the column total statistics
+   * @throws IOException if there is an error while writing
+   */
+  public void endColumn(Statistics<?> totalStatistics) throws IOException {
```
ConeyLiu, line 998: Exposing this to public is not a good idea. Other good suggestions are welcome.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
gszadovszky: I am not sure how column index writing worked for the "rewriter". This rewriting thing is newer than the column index. In the normal writing scenario `stats` cannot be empty since we are just creating these objects during the write path.

ConeyLiu: This is especially used when the `ColumnIndex` is null during rewriting. And we pass empty statistics to the `ColumnIndexBuilder` to avoid NPE.

wgtmac: Let me try to understand what happens here. `convertStatistics` is used to recover page statistics from ColumnIndex or original page header if the ColumnIndex is unavailable. The problem emerges when ColumnIndex is unavailable. Am I correct? If true, then why do we need those changes in the ColumnIndexBuilder?

ConeyLiu: The problem happens when both the `ColumnIndex` and the page header `Statistics` are null. Because we get `null` returned from the `convertStatistics`. However, the `ParquetFileWriter.writeDataPage` needs the page statistics. So here we pass invalid page statistics to avoid the NPE and overwrite the column statistics in the end. Otherwise, we need to add some methods that don't need page statistics.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -612,13 +612,13 @@ public void writeDataPage(
...
-    Encoding rlEncoding,
-    Encoding dlEncoding,
-    Encoding valuesEncoding) throws IOException {
+      int valueCount, int uncompressedPageSize,
+      BytesInput bytes,
+      Statistics<?> statistics,
+      long rowCount,
+      Encoding rlEncoding,
+      Encoding dlEncoding,
+      Encoding valuesEncoding) throws IOException {
```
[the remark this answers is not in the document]
ConeyLiu: Sure, will revert it.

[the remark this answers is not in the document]
ConeyLiu: reverted

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
wgtmac, line 549: This is used by the rewriter to rebuild the ColumnIndex. The change enables it to invalidate the ColumnIndex if any statistics is invalid, right? This is also used by the general file writer. Does it affect the behavior? If yes, probably we need add a test case.

wgtmac: From what you have said, it seems that the problem comes from the input file which has valid aggregate statistics for the column chunk but does not write page statistics in the page header. Should we just fix the NPE in the page header and leave other parts as is?

ConeyLiu: > Should we just fix the NPE in the page header and leave other parts as is?

Updated the implementation.

ConeyLiu: Reverted these changes

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -988,6 +988,21 @@ void writeColumnChunk(ColumnDescriptor descriptor,
...
+   * is known while all the page statistics are invalid, for example when rewriting the column.
+   *
+   * @param totalStatistics the column total statistics
+   * @throws IOException if there is an error while writing
+   */
+  public void endColumn(Statistics<?> totalStatistics) throws IOException {
+    Preconditions.checkArgument(totalStatistics != null, "Column total statistics can not be null");
+    currentStatistics = totalStatistics;
+    // Invalid the ColumnIndex
+    columnIndexBuilder = ColumnIndexBuilder.getNoOpBuilder();
```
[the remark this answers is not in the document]
ConeyLiu: Must do this since the column statistics may not match with the built column index.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1317,6 +1324,27 @@ private int toIntWithCheck(long size) {
     return (int)size;
   }
 
+  private void mergeColumnStatistics(Statistics<?> statistics) {
+    if (!currentStatisticsAreValid) {
```
wgtmac, line 1328: Are we able to reuse `currentStatistics.isEmpty()` in this case? We can simply set currentStatistics to empty if any unexpected statistics are to merge.

ConeyLiu: Do we support the case where the input statistics are empty? In that case, the `currentStatistics` will be empty after merging the first statistics. Then we have no idea whether the statistics are valid or not.

wgtmac: From the comment below, I think it serves the same purpose to mark a statistics as invalid. Merging an empty stats should invalidate aggregate statistics and ColumnIndexBuilder, IMO.

```
  /**
   * Returns a boolean specifying if the Statistics object is empty,
   * i.e does not contain valid statistics for the page/column yet
   * @return true if object is empty, false otherwise
   */
  public boolean isEmpty() {
    return !hasNonNullValue && !isNumNullsSet();
  }

```

ConeyLiu: OK, I updated the implementation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -722,14 +722,7 @@ public void writeDataPage(
...
 
-    // Copying the statistics if it is not initialized yet so we have the correct typed one
-    if (currentStatistics == null) {
-      currentStatistics = statistics.copy();
-    } else {
-      currentStatistics.mergeStatistics(statistics);
-    }
-
-    columnIndexBuilder.add(statistics);
+    mergeColumnStatistics(statistics);
```
wgtmac, line 725: Do we need to fix here and below: https://github.com/apache/parquet-mr/blob/514cc6c257fe8e618b100a19d86d304f6442cb94/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L215

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
wgtmac: Since this is used only for invalid stats, is it better to add a `void invalidateStatistics()` which simply include line 988 and 990? Then you just need to call invalidateStatistics() and endColumn() in this case.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -392,6 +402,7 @@ private void processChunk(ColumnChunkMetaData chunk,
     DictionaryPage dictionaryPage = null;
     long readValues = 0;
     Statistics<?> statistics = null;
+    boolean needOverwriteColumnStatistics = false;
```
wgtmac, line 405:
```suggestion
    boolean isColumnStatisticsMalformed = false;
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -324,15 +324,20 @@ private void processBlocksFromReader(IndexCache indexCache) throws IOException {
...
                   newCodecName,
                   columnChunkEncryptorRunTime,
                   encryptColumn,
                   indexCache.getBloomFilter(chunk),
                   indexCache.getColumnIndex(chunk),
                   indexCache.getOffsetIndex(chunk));
-          writer.endColumn();
+          if (needOverwriteStatistics) {
+            // All the column statistics are invalid, so we need to overwrite the column statistics
+            writer.endColumn(chunk.getStatistics());
```
wgtmac, line 337: Is it cleaner to handle this inside processChunk()? Then we don't have to bother with endColumn()

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: This is used to set the column aggregated statistics. Do you mean to modify the `public void endColumn(Statistics<?> totalStatistics)` to `public void invalidateStatistics(Statistics<?> totalStatistics)`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: I will update it. There are many duplicated codes between `ParquetFileWriter` and `ColumnChunkPageWriteStore`. Maybe we should refactor it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
ConeyLiu: I think that would be better. And just need to call the API: `public void invalidateStatistics(Statistics<?> totalStatistics)` in the `processChunk`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: Replaced with `invalidateStatistics `

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: Fixed as well.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
ConeyLiu: Updated

wgtmac (approved):
Thanks. LGTM

wgtmac:
It seems that a rebase is required to make CIs happy.

ConeyLiu:
Just rebased

wgtmac (approved):

ConeyLiu:
Thanks @wgtmac @gszadovszky

ConeyLiu:
Hi, @wgtmac @gszadovszky  should we port this to 1.13.x branch? And is there any plan for a new release?

wgtmac:
IMO, there isn't any plan for 1.13.2 release. But you may port it to 1.13.x branch just in case.

ConeyLiu:
OK, I will port it.

PR #1174, opened 2023-10-17
PARQUET-2366: Optimize random seek during rewriting

ConeyLiu:

### Jira

### Tests

### Commits

### Documentation

The `ColunIndex`, `OffsetIndex`, and `BloomFilter` are stored at the end of the file. We need to randomly seek 4 times when rewriting a column chunk. We found this could impact the rewrite performance heavily for files with a number of columns(~1000). In this PR, we read the `ColumnIndex`, `OffsetIndex`, and `BloomFilter` into a cache to avoid the random seek. We got about 60 times performance improvement in production environments for the files with about one thousand columns.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.hadoop;
+
+import org.apache.parquet.column.values.bloomfilter.BloomFilter;
+import org.apache.parquet.hadoop.metadata.BlockMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
+import org.apache.parquet.internal.column.columnindex.ColumnIndex;
+import org.apache.parquet.internal.column.columnindex.OffsetIndex;
+
+import java.io.IOException;
+import java.util.Set;
+
+/**
+ * A cache for caching indexes(including: ColumnIndex, OffsetIndex and BloomFilter)
+ */
+public interface IndexCache {
+
+  enum CacheStrategy {
+    NONE, /* No cache */
+    PRECACHE_BLOCK /* Precache for block indexes */
+  }
+
+  /**
+   * Create an index cache for the given file reader
+   *
+   * @param fileReader the file reader
+   * @param columns the columns that need to do cache
+   * @param cacheStrategy the cache strategy, supports NONE and PRECACHE_BLOCK
+   * @return the index cache
+   */
+  static IndexCache create(
+      ParquetFileReader fileReader,
+      Set<ColumnPath> columns,
+      CacheStrategy cacheStrategy) {
+    if (cacheStrategy == CacheStrategy.NONE) {
+      return new NoneIndexCache(fileReader);
+    } else if (cacheStrategy == CacheStrategy.PRECACHE_BLOCK) {
+      return new PrefetchIndexCache(fileReader, columns);
+    } else {
+      throw new UnsupportedOperationException("Unknown cache strategy: " + cacheStrategy);
+    }
+  }
+
+  /**
+   * Set the current BlockMetadata
+   */
+  void setBlockMetadata(BlockMetaData currentBlockMetadata) throws IOException;
+
+  /**
+   * Get the ColumnIndex for the given column in the set row group.
+   *
+   * @param chunk the given column chunk
+   * @return the ColumnIndex for the given column
+   * @throws IOException if any I/O error occurs during get the ColumnIndex
+   */
+  ColumnIndex getColumnIndex(ColumnChunkMetaData chunk) throws IOException;
+
+  /**
+   * Get the OffsetIndex for the given column in the set row group.
+   *
+   * @param chunk the given column chunk
+   * @return the OffsetIndex for the given column
+   * @throws IOException if any I/O error occurs during get the OffsetIndex
+   */
+  OffsetIndex getOffsetIndex(ColumnChunkMetaData chunk) throws IOException;
+
+  /**
+   * Get the BloomFilter for the given column in the set row group.
+   *
+   * @param chunk the given column chunk
+   * @return the BloomFilter for the given column
+   * @throws IOException if any I/O error occurs during get the BloomFilter
+   */
+  BloomFilter getBloomFilter(ColumnChunkMetaData chunk) throws IOException;
+
+  /**
+   * Clean the cache
+   */
+  void clean();
+}
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
@@ -35,7 +35,7 @@ public interface IndexCache {
 
   enum CacheStrategy {
     NONE, /* No cache */
-    PRECACHE_BLOCK /* Precache for block indexes */
+    PREFETCH_BLOCK /* Prefetch block indexes */
   }
 
   /**
@@ -43,7 +43,7 @@ public interface IndexCache {
    *
    * @param fileReader the file reader
    * @param columns the columns that need to do cache
-   * @param cacheStrategy the cache strategy, supports NONE and PRECACHE_BLOCK
+   * @param cacheStrategy the cache strategy, supports NONE and PREFETCH_BLOCK
    * @param freeCacheAfterGet whether free the given index cache after calling the given get method
    * @return the index cache
    */
@@ -54,7 +54,7 @@ public interface IndexCache {
       boolean freeCacheAfterGet) {
     if (cacheStrategy == CacheStrategy.NONE) {
       return new NoneIndexCache(fileReader);
-    } else if (cacheStrategy == CacheStrategy.PRECACHE_BLOCK) {
+    } else if (cacheStrategy == CacheStrategy.PREFETCH_BLOCK) {
       return new PrefetchIndexCache(fileReader, columns, freeCacheAfterGet);
     } else {
       throw new UnsupportedOperationException("Unknown cache strategy: " + cacheStrategy);
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
@@ -44,16 +44,18 @@ public interface IndexCache {
    * @param fileReader the file reader
    * @param columns the columns that need to do cache
    * @param cacheStrategy the cache strategy, supports NONE and PRECACHE_BLOCK
+   * @param freeCacheAfterGet whether free the given index cache after calling the given get method
    * @return the index cache
    */
   static IndexCache create(
       ParquetFileReader fileReader,
       Set<ColumnPath> columns,
-      CacheStrategy cacheStrategy) {
+      CacheStrategy cacheStrategy,
+      boolean freeCacheAfterGet) {
     if (cacheStrategy == CacheStrategy.NONE) {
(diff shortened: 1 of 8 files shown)


ConeyLiu:
Hi, @wgtmac @gszadovszky please help to review this when you are free. Thanks a lot.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -265,6 +265,10 @@ private void processBlocksFromReader() throws IOException {
       BlockMetaData blockMetaData = meta.getBlocks().get(blockId);
       List<ColumnChunkMetaData> columnsInOrder = blockMetaData.getColumns();
 
+      List<ColumnIndex> columnIndexes = readAllColumnIndexes(reader, columnsInOrder, descriptorsMap);
```
ConeyLiu, line 268: I could add an option to it if someone is concerned about memory usage. This only caches the metadata for only one block and should be smaller than doing file writing which needs to cache all blocks' metadata.

wgtmac: Thanks for adding this! The change looks reasonable to me. I would suggest adding a new class to specifically cache and read these indexes. The new class have methods like `readBloomFilter()`, `readColumnIndex()` and `readOffsetIndex()` for a specific column path, and can be configured to cache required columns in advance. With this new class, we can do more optimizations including evict consumed items out of cache and use async I/O prefetch to load items. We can split them into separate patches. For the first one, we may simply add the new class without any caching (i.e. no behavior change). WDYT?

ConeyLiu: @wgtmac thanks for your suggestions. Do you mean to read the indexes column by column to reduce memory footprint? The suggested way should have less memory usage. The indexes are stored as the following from my understanding:
```
// column index
block1_col1_column_index
...
block1_coln_column_index
block2_col1_column_index
...
block2_coln_column_index
...

// offset index
block1_col1_offset_index
...
block1_coln_offset_index
block2_col1_offset_index
...
block2_coln_offset_index
...

// bloom index
block1_col1_bloom_index
...
block1_coln_bloom_index
block2_col1_bloom_index
...
block2_coln_bloom_index
...
```

So the problem would be we still need to do random seek for a single rowgroup(3 * number of columns). The async I/O should be helpful for the random seek performance. With this PR, we only need 3 times random seek (except the column pruning) for a single rowgroup.

wgtmac: > Do you mean to read the indexes column by column to reduce memory footprint?

No, my suggested interface does not restrict any implementation detail, at least they should be ready at the `readXXX()` call. You can still read all indexes at once (controlled by a config). We can configurated it to release any consumed index object to reduce memory footprint.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/IndexCacher.java
```diff
@@ -0,0 +1,147 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A cacher for caching file indexes(ColumnIndex, OffsetIndex, BloomFilter)
+ */
+class IndexCacher {
```
ConeyLiu, line 37: @wgtmac Abstracted a cacher class and we could enrich it in the future.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -92,16 +95,21 @@ public FileEncryptionProperties getFileEncryptionProperties() {
...
-    private Path outputFile;
+    private final Configuration conf;
+    private final List<Path> inputFiles;
+    private final Path outputFile;
     private List<String> pruneColumns;
     private CompressionCodecName newCodecName;
     private Map<String, MaskMode> maskColumns;
     private List<String> encryptColumns;
     private FileEncryptionProperties fileEncryptionProperties;
+    private boolean prefetchBlockAllIndexes;
```
ConeyLiu, line 112: Disabled by default to keep existing behaviors.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/IndexCacher.java
wgtmac: Thanks for adding this! Do you think we can move it out of the rewrite module so that the general parquet reader can leverage it in the future?

wgtmac: BTW, what about switching it into an interface? We can possibly provide this as a default concrete implementation. A new enum class (e.g. IndexCacheStrategy) for selecting implementations would be beneficial to extensibility.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/IndexCacher.java
```diff
@@ -0,0 +1,147 @@
...
+  void setCurrentBlockMetadata(BlockMetaData blockMetaData) throws IOException {
+    if (prefetchBlockAllIndexes) {
+      free();
+      this.columnIndexCache = readAllColumnIndexes(blockMetaData);
+      this.offsetIndexCache = readAllOffsetIndexes(blockMetaData);
+      this.bloomIndexCache = readAllBloomFilters(blockMetaData);
+    }
+  }
+
+  ColumnIndex getColumnIndex(ColumnChunkMetaData chunk) throws IOException {
```
wgtmac, line 65: nit: add some comment for these APIs.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -213,6 +221,19 @@ public Builder addInputFile(Path path) {
...
 
+    /**
+     * Whether enable prefetch block indexes into cache.
+     * <p>
+     * This could reduce the random seek while rewriting, disabled by default.
+     *
+     * @param prefetchBlockAllIndexes enable or not
+     * @return self
+     */
+    public Builder prefetchBlockAllIndex(boolean prefetchBlockAllIndexes) {
```
wgtmac, line 232: An enum class would be more extensible in the future. Please see my above comments.

wgtmac (changes requested):
Thanks for the quick response! I have left some comments. Let me know what you think.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/IndexCacher.java
ConeyLiu: > Thanks for adding this! Do you think we can move it out of the rewrite module so that the general parquet reader can leverage it in the future?

Good idea, this may be useful for file reading. Moved to package `org.apache.parquet.hadoop`

> BTW, what about switching it into an interface? We can possibly provide this as a default concrete implementation. A new enum class (e.g. IndexCacheStrategy) for selecting implementations would be beneficial to extensibility.

Changed to interface and added two strategies: `NONE` and `PRECACHE_BLOCK`.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestIndexCache.java
```diff
@@ -0,0 +1,168 @@
...
+      .withNumRecord(numRecords)
+      .withCodec("UNCOMPRESSED")
+      .withRowGroupSize(1024L)
+      .withBloomFilterEnabled(bloomFilterEnabledColumns)
+      .withWriterVersion(writerVersion)
+      .build()
+      .getFileName();
+  }
+
+  private void validateColumnIndex(ColumnIndex expected, ColumnIndex target) {
```
ConeyLiu, line 145: `ColumnIndex` has not customized `equals`, maybe we could implement it to simplify the check here.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestIndexCache.java
```diff
@@ -0,0 +1,168 @@
...
+      Assert.assertEquals(expected.getClass(), target.getClass());
+      Assert.assertEquals(expected.getMinValues(), target.getMinValues());
+      Assert.assertEquals(expected.getMaxValues(), target.getMaxValues());
+      Assert.assertEquals(expected.getBoundaryOrder(), target.getBoundaryOrder());
+      Assert.assertEquals(expected.getNullCounts(), target.getNullCounts());
+      Assert.assertEquals(expected.getNullPages(), target.getNullPages());
+    }
+  }
+
+  private void validateOffsetIndex(OffsetIndex expected, OffsetIndex target) {
```
ConeyLiu, line 159: Sames as `OffsetIndex`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/PrefetchIndexCache.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.parquet.internal.column.columnindex.ColumnIndex;
+import org.apache.parquet.internal.column.columnindex.OffsetIndex;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This index cache will prefetch those columns' indexes when calling {@link #setBlockMetadata(BlockMetaData)}.
```
wgtmac, line 35:
```suggestion
 * This index cache will prefetch indexes of all columns when calling {@link #setBlockMetadata(BlockMetaData)}.
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/PrefetchIndexCache.java
```diff
@@ -0,0 +1,156 @@
...
+    ColumnPath columnPath = chunk.getPath();
+    if (columns.contains(columnPath)) {
+      Preconditions.checkState(
+        columnIndexCache.containsKey(columnPath),
+        "Not found cached ColumnIndex for column: %s with cache strategy: %s",
+        columnPath.toDotString(),
+        CacheStrategy.PRECACHE_BLOCK);
+    }
+
+    return columnIndexCache.remove(columnPath);
```
wgtmac, line 73: This does not support repeated read, we should either document them in the comment or make it configurable. In some cases (e.g. filter pushdown), page index may be repeated accessed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
```diff
@@ -0,0 +1,98 @@
...
+      return new NoneIndexCache(fileReader);
+    } else if (cacheStrategy == CacheStrategy.PRECACHE_BLOCK) {
+      return new PrefetchIndexCache(fileReader, columns);
+    } else {
+      throw new UnsupportedOperationException("Unknown cache strategy: " + cacheStrategy);
+    }
+  }
+
+  /**
+   * Set the current BlockMetadata
```
wgtmac, line 63: Will calling this invalidate all previously cached items?

wgtmac (commented):
Thanks for the change! This is looking good to me.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/PrefetchIndexCache.java
ConeyLiu: I added the parameter: `freeCacheAfterGet` to enable it or not.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
ConeyLiu: Yes, I documented it in the class header.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  static IndexCache create(
+      ParquetFileReader fileReader,
+      Set<ColumnPath> columns,
+      CacheStrategy cacheStrategy,
+      boolean freeCacheAfterGet) {
+    if (cacheStrategy == CacheStrategy.NONE) {
+      return new NoneIndexCache(fileReader);
+    } else if (cacheStrategy == CacheStrategy.PRECACHE_BLOCK) {
+      return new PrefetchIndexCache(fileReader, columns, freeCacheAfterGet);
```
wgtmac, line 58: nit: should we rename PRECACHE_BLOCK to PREFETCH_BLOCK?

wgtmac (approved):
Thanks! LGTM.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/IndexCache.java
ConeyLiu: renamed

wgtmac (approved):

wgtmac:
cc @gszadovszky @ggershinsky @shangxinli if this interests you.

gszadovszky (approved):

ConeyLiu:
Thanks @wgtmac @gszadovszky

PR #1175, opened 2023-10-18
PARQUET-2355: Deprecate `parquet-thrift`

Fokko:

`parquet-thrift` relies on Elephantbird: https://github.com/twitter/elephant-bird

The current version of `elephant-bird` that we use is still on Thrift 0.7.0, and this doesn't work with 0.19.0 that we're trying to upgrade to. Updating is non-trivial since a lot of the code that was used for testing has been made private: https://github.com/apache/parquet-mr/pull/1156

I would suggest removing this from the repository since it looks like no one is using it anymore:

`apache-thrift` does not seem to be used anymore on Maven Central (usages of 0 since 1.12.2): 

image

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2355

### Tests

### Commits

### Documentation


pom.xml
```diff
@@ -544,6 +544,17 @@
               <!-- japicmp is overly aggressive on interface types in signatures, a type was changed to a supertype but this still triggers it -->
               <exclude>org.apache.parquet.hadoop.ColumnChunkPageWriteStore</exclude>
               <exclude>org.apache.parquet.hadoop.ParquetRecordWriter</exclude>
+              <!-- Already deprecated classes/methods/constants -->
+              <exclude>org.apache.parquet.hadoop.thrift.ThriftReadSupport#setProjectionPushdown(org.apache.hadoop.mapred.JobConf,java.lang.String)</exclude>
+              <exclude>org.apache.parquet.hadoop.thrift.ThriftReadSupport#THRIFT_COLUMN_FILTER_KEY</exclude>
```
Fokko, line 549: https://github.com/apache/parquet-mr/blob/5ee513341158bc3962df80260c345b80076dbf3e/parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java#L52-L57

pom.xml
```diff
@@ -544,6 +544,17 @@
...
+              <exclude>org.apache.parquet.thrift.projection.deprecated.DeprecatedFieldProjectionFilter</exclude>
```
Fokko, line 550: https://github.com/apache/parquet-mr/blob/5ee513341158bc3962df80260c345b80076dbf3e/parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/DeprecatedFieldProjectionFilter.java#L33-L34

pom.xml
```diff
@@ -544,6 +544,17 @@
...
+              <!--
+              Got this from:
+                org.apache.parquet.thrift.projection.FieldProjectionFilter[org.apache.parquet.thrift.projection.FieldProjectionFilter]:INTERFACE_REMOVED
+
+              But it didn't change, only the @Deprecated annotation was added
```
Fokko, line 555: Maybe it is a bug, let's try to bump it: https://github.com/apache/parquet-mr/pull/1176

Fokko: Nice, the bump fixed it :)

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java
```diff
@@ -41,21 +41,13 @@
...
 
+@Deprecated
 public class ThriftReadSupport<T> extends ReadSupport<T> {
   private static final Logger LOG = LoggerFactory.getLogger(ThriftReadSupport.class);
-
-  /**
-   * Deprecated. Use {@link #STRICT_THRIFT_COLUMN_FILTER_KEY}
-   * Accepts a ";" delimited list of globs in the syntax implemented by {@link DeprecatedFieldProjectionFilter}
-   */
-  @Deprecated
```
wgtmac, line 56: Why not keeping them as-is?

parquet-thrift/src/main/java/org/apache/parquet/thrift/projection/deprecated/DeprecatedFieldProjectionFilter.java
```diff
@@ -1,106 +0,0 @@
...
-import org.apache.parquet.thrift.projection.FieldsPath;
-import org.apache.parquet.thrift.projection.FieldProjectionFilter;
-import org.apache.parquet.thrift.projection.ThriftProjectionException;
-
-/**
- * Filter thrift attributes using glob syntax.
- * This is used for parsing values assigned to ThriftReadSupport.THRIFT_COLUMN_FILTER_KEY
- */
-@Deprecated
-public class DeprecatedFieldProjectionFilter implements FieldProjectionFilter {
```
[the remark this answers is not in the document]
Fokko: These classes are [marked as deprecated in 2015](https://github.com/apache/parquet-mr/commit/7fc7998398373a14b4cdc0ce18abdeb221b1ccf9), and released in Parquet 1.8.0 :) Removing these was also suggested by @steveloughran in https://github.com/apache/parquet-mr/pull/979

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java
Fokko: These are marked as deprecated in 2015. I think removing these is a first step before removing the whole class.

Fokko:
PTAL @shangxinli @gszadovszky @SinghAsDev @rdblue @julienledem

SinghAsDev:
Hi @Fokko thanks for bringing this up. We at Pinterest still use Parquet thrift quite extensively. How urgent is thrift 0.19.0 upgrade and how much time can we wait before we absolutely need to upgrade thrift? Also, since elephantbird is not maintained anymore, one option would be to just get rid of elephantbird dependency completely, even if it means moving some code in parquet-mr. Thoughts?

SinghAsDev:
@tlazaro, @ttim, @piyushnarang, @isnotinvain in case you all still use parquet-thrift, this discussion maybe of interest to you as well.

tlazaro:
Thanks so much for pinging, though I'm no longer at Twitter, I will try to find someone who might be interested in the conversation.

Fokko:
@SinghAsDev and @tlazaro I think we can make it work with the latest version of Elephantbird. Let me close this for now.

parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftReadSupport.java
steveloughran: part of my attempts to move up to later hadoop versions hit real problems here. and along with thrift in general -i had to grab a copy of an x86 MBP for my mac m1, but even there homebrew had decided to upgrade it, so it was a backup in a corner I ended up with. At which point I discovered that homebrew isn't a real package manager as you can't roll back...

Fokko: I'm also on a M1, and it takes quite a bit of effort to get Thrift compiled. I got a patch ready to update Thrift to the latest version. I'm not sure if we need to update `parquet-format` first, since the release is still on an older version. I'll give it a try in a moment.

steveloughran: that or stick the old version into the repo...

PR #1177, opened 2023-10-19
PARQUET-2261: Implement SizeStatistics

wgtmac:

### Jira

  - https://issues.apache.org/jira/browse/PARQUET-2261

### Tests

### Commits

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.impl;
+
+import java.util.OptionalDouble;
+import java.util.OptionalLong;
+import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.ParquetProperties;
+import org.apache.parquet.column.statistics.SizeStatistics;
+import org.apache.parquet.column.statistics.Statistics;
+import org.apache.parquet.column.values.bloomfilter.AdaptiveBlockSplitBloomFilter;
+import org.apache.parquet.column.values.bloomfilter.BlockSplitBloomFilter;
+import org.apache.parquet.column.values.bloomfilter.BloomFilter;
+import org.apache.parquet.column.values.bloomfilter.BloomFilterWriter;
+import org.apache.parquet.io.api.Binary;
+
+// An internal class to collect column values to build column statistics and bloom filter.
+class ColumnValueCollector {
+
+  private final ColumnDescriptor path;
+  private final BloomFilterWriter bloomFilterWriter;
+  private final BloomFilter bloomFilter;
+  private Statistics<?> statistics;
+  private SizeStatistics.Builder sizeStatisticsBuilder;
+
+  ColumnValueCollector(ColumnDescriptor path, BloomFilterWriter bloomFilterWriter, ParquetProperties props) {
+    this.path = path;
+    resetPageStatistics();
+
+    this.bloomFilterWriter = bloomFilterWriter;
+    if (bloomFilterWriter == null) {
+      this.bloomFilter = null;
+      return;
+    }
+    int maxBloomFilterSize = props.getMaxBloomFilterBytes();
+
+    OptionalLong ndv = props.getBloomFilterNDV(path);
+    OptionalDouble fpp = props.getBloomFilterFPP(path);
+    // If user specify the column NDV, we construct Bloom filter from it.
+    if (ndv.isPresent()) {
+      int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
+      this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
+    } else {
+      if (props.getAdaptiveBloomFilterEnabled(path)) {
+        int numCandidates = props.getBloomFilterCandidatesCount(path);
+        this.bloomFilter =
+            new AdaptiveBlockSplitBloomFilter(maxBloomFilterSize, numCandidates, fpp.getAsDouble(), path);
+      } else {
+        this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
+      }
+    }
+  }
+
+  void resetPageStatistics() {
+    this.statistics = Statistics.createStats(path.getPrimitiveType());
+    this.sizeStatisticsBuilder = SizeStatistics.newBuilder(
+        path.getPrimitiveType(), path.getMaxRepetitionLevel(), path.getMaxDefinitionLevel());
+  }
+
+  void writeNull(int repetitionLevel, int definitionLevel) {
+    statistics.incrementNumNulls();
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+  }
+
+  void write(boolean value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+  }
+
+  void write(int value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+    if (bloomFilter != null) {
+      bloomFilter.insertHash(bloomFilter.hash(value));
+    }
+  }
+
+  void write(long value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+    if (bloomFilter != null) {
+      bloomFilter.insertHash(bloomFilter.hash(value));
+    }
+  }
+
+  void write(float value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+    if (bloomFilter != null) {
+      bloomFilter.insertHash(bloomFilter.hash(value));
+    }
+  }
+
+  void write(double value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+    if (bloomFilter != null) {
+      bloomFilter.insertHash(bloomFilter.hash(value));
+    }
+  }
+
+  void write(Binary value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel, value);
+    if (bloomFilter != null) {
+      bloomFilter.insertHash(bloomFilter.hash(value));
+    }
+  }
+
+  void finalizeColumnChunk() {
+    if (bloomFilterWriter != null && bloomFilter != null) {
+      bloomFilterWriter.writeBloomFilter(bloomFilter);
+    }
+  }
+
+  Statistics<?> getStatistics() {
+    return statistics;
+  }
+
+  SizeStatistics getSizeStatistics() {
+    return sizeStatisticsBuilder.build();
+  }
+}
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
@@ -18,6 +18,8 @@
  */
 package org.apache.parquet.column.impl;
 
+import java.io.IOException;
+import java.io.OutputStream;
 import java.util.OptionalDouble;
 import java.util.OptionalLong;
 import org.apache.parquet.column.ColumnDescriptor;
@@ -34,37 +36,15 @@ import org.apache.parquet.io.api.Binary;
 class ColumnValueCollector {
 
   private final ColumnDescriptor path;
-  private final BloomFilterWriter bloomFilterWriter;
-  private final BloomFilter bloomFilter;
+  private BloomFilterWriter bloomFilterWriter;
+  private BloomFilter bloomFilter;
   private Statistics<?> statistics;
   private SizeStatistics.Builder sizeStatisticsBuilder;
 
   ColumnValueCollector(ColumnDescriptor path, BloomFilterWriter bloomFilterWriter, ParquetProperties props) {
     this.path = path;
     resetPageStatistics();
-
-    this.bloomFilterWriter = bloomFilterWriter;
-    if (bloomFilterWriter == null) {
-      this.bloomFilter = null;
-      return;
-    }
-    int maxBloomFilterSize = props.getMaxBloomFilterBytes();
-
-    OptionalLong ndv = props.getBloomFilterNDV(path);
-    OptionalDouble fpp = props.getBloomFilterFPP(path);
-    // If user specify the column NDV, we construct Bloom filter from it.
-    if (ndv.isPresent()) {
-      int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
-      this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
-    } else {
-      if (props.getAdaptiveBloomFilterEnabled(path)) {
-        int numCandidates = props.getBloomFilterCandidatesCount(path);
-        this.bloomFilter =
-            new AdaptiveBlockSplitBloomFilter(maxBloomFilterSize, numCandidates, fpp.getAsDouble(), path);
-      } else {
-        this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
-      }
-    }
+    initBloomFilter(bloomFilterWriter, props);
   }
 
   void resetPageStatistics() {
@@ -86,45 +66,119 @@ class ColumnValueCollector {
   void write(int value, int repetitionLevel, int definitionLevel) {
     statistics.updateStats(value);
     sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
-    if (bloomFilter != null) {
-      bloomFilter.insertHash(bloomFilter.hash(value));
-    }
+    bloomFilter.insertHash(bloomFilter.hash(value));
   }
 
   void write(long value, int repetitionLevel, int definitionLevel) {
     statistics.updateStats(value);
     sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
-    if (bloomFilter != null) {
-      bloomFilter.insertHash(bloomFilter.hash(value));
-    }
+    bloomFilter.insertHash(bloomFilter.hash(value));
   }
 
   void write(float value, int repetitionLevel, int definitionLevel) {
     statistics.updateStats(value);
     sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
-    if (bloomFilter != null) {
-      bloomFilter.insertHash(bloomFilter.hash(value));
-    }
+    bloomFilter.insertHash(bloomFilter.hash(value));
   }
 
   void write(double value, int repetitionLevel, int definitionLevel) {
     statistics.updateStats(value);
     sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
-    if (bloomFilter != null) {
-      bloomFilter.insertHash(bloomFilter.hash(value));
-    }
+    bloomFilter.insertHash(bloomFilter.hash(value));
   }
 
   void write(Binary value, int repetitionLevel, int definitionLevel) {
     statistics.updateStats(value);
     sizeStatisticsBuilder.add(repetitionLevel, definitionLevel, value);
-    if (bloomFilter != null) {
-      bloomFilter.insertHash(bloomFilter.hash(value));
+    bloomFilter.insertHash(bloomFilter.hash(value));
+  }
+
+  void initBloomFilter(BloomFilterWriter bloomFilterWriter, ParquetProperties props) {
+    this.bloomFilterWriter = bloomFilterWriter;
+    if (bloomFilterWriter == null) {
+      this.bloomFilter = new BloomFilter() {
+        @Override
+        public void writeTo(OutputStream out) throws IOException {}
+
+        @Override
+        public void insertHash(long hash) {}
+
+        @Override
+        public boolean findHash(long hash) {
+          return false;
+        }
+
+        @Override
+        public int getBitsetSize() {
+          return 0;
+        }
+
+        @Override
+        public long hash(int value) {
+          return 0;
+        }
+
+        @Override
+        public long hash(long value) {
+          return 0;
+        }
+
+        @Override
+        public long hash(double value) {
+          return 0;
+        }
+
+        @Override
+        public long hash(float value) {
+          return 0;
+        }
+
+        @Override
+        public long hash(Binary value) {
+          return 0;
+        }
+
+        @Override
+        public long hash(Object value) {
+          return 0;
+        }
+
+        @Override
+        public HashStrategy getHashStrategy() {
+          return null;
+        }
+
+        @Override
+        public Algorithm getAlgorithm() {
+          return null;
+        }
+
+        @Override
+        public Compression getCompression() {
+          return null;
+        }
+      };
+      return;
+    }
+
+    int maxBloomFilterSize = props.getMaxBloomFilterBytes();
+    OptionalLong ndv = props.getBloomFilterNDV(path);
+    OptionalDouble fpp = props.getBloomFilterFPP(path);
+    // If user specify the column NDV, we construct Bloom filter from it.
+    if (ndv.isPresent()) {
+      int optimalNumOfBits = BlockSplitBloomFilter.optimalNumOfBits(ndv.getAsLong(), fpp.getAsDouble());
+      this.bloomFilter = new BlockSplitBloomFilter(optimalNumOfBits / 8, maxBloomFilterSize);
+    } else if (props.getAdaptiveBloomFilterEnabled(path)) {
+      int numCandidates = props.getBloomFilterCandidatesCount(path);
+      this.bloomFilter =
+          new AdaptiveBlockSplitBloomFilter(maxBloomFilterSize, numCandidates, fpp.getAsDouble(), path);
+    } else {
+      this.bloomFilter = new BlockSplitBloomFilter(maxBloomFilterSize, maxBloomFilterSize);
     }
   }
 
   void finalizeColumnChunk() {
-    if (bloomFilterWriter != null && bloomFilter != null) {
+    if (bloomFilterWriter != null) {
       bloomFilterWriter.writeBloomFilter(bloomFilter);
     }
   }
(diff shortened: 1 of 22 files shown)


wgtmac:
I have drafted the POC to read/write SizeStatistics. The feature implementation should be complete and associated tests will be added progressively. Please take a look when you have time. Thanks! @emkornfield 

cc @mapleFU

etseidl:
Thanks @wgtmac, this looks great! I'm not sure if this is in scope for this PR, but it would be nice if the CLI was aware of the changes. Specifically, it would be great if the `column-index` command could write out the unencoded sizes and histograms.  The former is pretty straightforward, but I'm not entirely sure what the right approach is for the histograms. Also, `rewrite` currently ignores the new statistics. For a straight copy, all that's needed is to add `chunk.getSizeStatistics()` to the arg list of `ColumnChunkMetadata.get()` [here](https://github.com/wgtmac/parquet-mr/blob/45e8ee427e8766489e4a84246f452178ceefb75a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L1358).

wgtmac:
> Thanks @wgtmac, this looks great! I'm not sure if this is in scope for this PR, but it would be nice if the CLI was aware of the changes. Specifically, it would be great if the `column-index` command could write out the unencoded sizes and histograms. The former is pretty straightforward, but I'm not entirely sure what the right approach is for the histograms. Also, `rewrite` currently ignores the new statistics. For a straight copy, all that's needed is to add `chunk.getSizeStatistics()` to the arg list of `ColumnChunkMetadata.get()` [here](https://github.com/wgtmac/parquet-mr/blob/45e8ee427e8766489e4a84246f452178ceefb75a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L1358).

Thanks for the suggestion! Yes, it definitely should be done. @etseidl

emkornfield:
@wgtmac took a scan through and this generally seems like what I expected.  Thank you for doing it.  Agree unit tests are needed.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Merge two SizeStatistics of the same column.
+   * It is used to merge size statistics from all pages of the same column chunk.
+   */
+  public void mergeStatistics(SizeStatistics other) {
+    Preconditions.checkArgument(type.equals(other.type), "Cannot merge SizeStatistics of different types");
+    Preconditions.checkArgument(repetitionLevelHistogram.size() != other.repetitionLevelHistogram.size(),
+      "Cannot merge repetitionLevelHistogram with different sizes");
+    Preconditions.checkArgument(definitionLevelHistogram.size() != other.definitionLevelHistogram.size(),
```
etseidl, line 167:
```suggestion
    Preconditions.checkArgument(repetitionLevelHistogram.size() == other.repetitionLevelHistogram.size(),
      "Cannot merge repetitionLevelHistogram with different sizes");
    Preconditions.checkArgument(definitionLevelHistogram.size() == other.definitionLevelHistogram.size(),
```
This one was hard to track down :sweat_smile:

wgtmac: Thanks for pointing it out! I will add more teats later.

wgtmac:
I have just rebased on the latest master branch and fixed all CI falures. As this PR gets too large, I will add print cli command and rewriter support for `SizeStatistics` in follow-up PRs. This is now ready for review. @emkornfield @etseidl @gszadovszky @shangxinli

wgtmac:
cc @ConeyLiu as I have modified mergeColumnStatistics method which you've just refactored.

ConeyLiu:
Thank @wgtmac for your notification.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+ * A structure for capturing metadata for estimating the unencoded,
+ * uncompressed size of data written. This is useful for readers to estimate
+ * how much memory is needed to reconstruct data in their memory model and for
+ * fine-grained filter push down on nested structures (the histograms contained
+ * in this structure can help determine the number of nulls at a particular
+ * nesting level and maximum length of lists).
+ */
+public class SizeStatistics {
+
+  private PrimitiveType type;
```
ConeyLiu, line 40: final

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * When present, there is expected to be one element corresponding to each
+   * repetition (i.e. size=max repetition_level+1) where each element
+   * represents the number of times the repetition level was observed in the
+   * data.
+   *
+   * This field may be omitted if max_repetition_level is 0 without loss
+   * of information.
+   */
+  private List<Long> repetitionLevelHistogram;
```
ConeyLiu, line 68: final

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Same as repetition_level_histogram except for definition levels.
+   *
+   * This field may be omitted if max_definition_level is 0 or 1 without
+   * loss of information.
+   */
+  private List<Long> definitionLevelHistogram;
```
ConeyLiu, line 75: final

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+
+  /**
+   * Create a SizeStatistics.
+   *
+   * @param type physical type of the column associated with this statistics
+   * @param unencodedByteArrayDataBytes number of physical bytes stored for BYTE_ARRAY data values assuming no encoding
+   * @param repetitionLevelHistogram histogram for all repetition levels if non-empty
+   * @param definitionLevelHistogram histogram for all definition levels if non-empty
+   */
+  public SizeStatistics(PrimitiveType type,
```
ConeyLiu, line 169: should this be private since we have the class builder?

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+   * It is true by default. Only set to false while it fails to merge statistics.
+   */
+  private boolean hasValue = true;
+
+  /**
+   * Builder to create a SizeStatistics.
+   */
+  public static class Builder {
+    private PrimitiveType type;
+    private long unencodedByteArrayDataBytes;
```
ConeyLiu, line 89: the other three fields should be final

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+     * @param type physical type of the column associated with this statistics
+     * @param maxRepetitionLevel maximum repetition level of the column
+     * @param maxDefinitionLevel maximum definition level of the column
+     */
+    public Builder(PrimitiveType type, int maxRepetitionLevel, int maxDefinitionLevel) {
+      this.type = type;
+      this.unencodedByteArrayDataBytes = 0L;
+      repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
+      definitionLevelHistogram = new long[maxDefinitionLevel + 1];
+      Arrays.fill(repetitionLevelHistogram, 0L);
```
ConeyLiu, line 105: Does this need? The array should be filled with zero by default.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,254 @@
...
+        "repetitionLevel %s is out of range [0, %s]", repetitionLevel, repetitionLevelHistogram.length - 1);
+      Preconditions.checkArgument(definitionLevel < definitionLevelHistogram.length,
+        "definitionLevel %s is out of range [0, %s]", definitionLevel, definitionLevelHistogram.length - 1);
+      repetitionLevelHistogram[repetitionLevel]++;
+      definitionLevelHistogram[definitionLevel]++;
+    }
+
+    /**
+     * Add repetition and definition level of a value to the statistics.
+     * It is called when value is null, or the column is not of BYTE_ARRAY type.
```
ConeyLiu, line 127: I think here the column could be `BYTE_ARRAY`. The `BYTE_ARRAY` is `Binary`, right?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndex.java
```diff
@@ -57,4 +57,16 @@ public interface ColumnIndex extends Visitor<PrimitiveIterator.OfInt> {
    */
   public List<ByteBuffer> getMaxValues();
 
+  /**
+   * @return the unmodifiable list of the repetition level histograms for each page concatenated together; used for
+   * converting to the related thrift object
+   */
+  public List<Long> getRepetitionLevelHistogram();
```
ConeyLiu, line 64: The `public` is redundant, but I think you want to keep the code alignment.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndexBuilder.java
```diff
@@ -80,11 +90,22 @@ public void add(int compressedPageSize, long rowCount) {
     @Override
     public void add(long offset, int compressedPageSize, long rowCount) {
     }
+
+    @Override
+    public OffsetIndex build() {
```
ConeyLiu, line 95: Why add these two changes? Does this perform as same as before?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndexBuilder.java
```diff
@@ -116,11 +137,28 @@ private OffsetIndexBuilder() {
...
+   *          the number of rows in the page
+   * @param unencodedDataBytes
+   *          the number of bytes of unencoded data of BYTE_ARRAY type
+   */
+  public void add(int compressedPageSize, long rowCount, Optional<Long> unencodedDataBytes) {
+    add(previousOffset + previousPageSize, compressedPageSize, previousRowIndex + previousRowCount, unencodedDataBytes);
     previousRowCount = rowCount;
   }
 
+
```
ConeyLiu, line 161: blank line

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -87,6 +90,7 @@ private static final class ColumnChunkPageWriter implements PageWriter, BloomFil
     private ColumnIndexBuilder columnIndexBuilder;
     private OffsetIndexBuilder offsetIndexBuilder;
     private Statistics totalStatistics;
+    private SizeStatistics totalSizeStatistics;
```
ConeyLiu, line 93: final

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -316,7 +346,14 @@ private int toIntWithCheck(long size) {
       return (int)size;
     }
 
-    private void mergeColumnStatistics(Statistics<?> statistics) {
+    private void mergeColumnStatistics(Statistics<?> statistics, SizeStatistics sizeStatistics) {
+      Preconditions.checkState(totalSizeStatistics != null, "Aggregate size statistics should not be null");
```
ConeyLiu, line 350: If the `totalSizeStatistics` is set as final. The checking here is not needed anymore.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -316,7 +346,14 @@ private int toIntWithCheck(long size) {
...
+      totalSizeStatistics.mergeStatistics(sizeStatistics);
```
ConeyLiu, line 351: Can the size of these two `sizeStatistics` be different?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -775,6 +848,51 @@ public void writeDataPageV2(
...
+   * @param compressedData compressed data bytes
+   * @param uncompressedDataSize the size of uncompressed data
+   * @param statistics the statistics of the page
+   * @param metadataBlockEncryptor encryptor for block data
+   * @param pageHeaderAAD pageHeader AAD
+   * @throws IOException if any I/O error occurs during writing the file
+   */
+  @Deprecated
+  public void writeDataPageV2(
+    int rowCount,
```
ConeyLiu, line 873: indent

ConeyLiu:
Hi @wgtmac thanks for this great work. Could this influence the rewriter? How could we rebuild the `SizeStatistics` during rewriting?

wgtmac:
> Hi @wgtmac thanks for this great work. Could this influence the rewriter? How could we rebuild the `SizeStatistics` during rewriting?

Thanks for your review! IMO, the rewriter can simply preserve SizeStatistics from the original ColumnMetaData and Page Index.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
wgtmac: The check here simply ignores the Binary value from FIXED_SIZE_BYTE_ARRAY type which we do not care.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndex.java
wgtmac: Yes, I don't want to add outliers here ...

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndexBuilder.java
wgtmac: If I don't, `TestParquetFileWriter.testAlignmentWithPadding` and `TestParquetFileWriter.testAlignmentWithNoPadding` will fail in the Github CI (but not on my laptop) due to writing offset index from NoOpOffsetIndexBuilder, which is an undefined behavior.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
wgtmac: That should not happen. But it is checked in the mergeStatistics.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
wgtmac: This is used to restore it from a serialized thrift object.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -389,7 +400,14 @@ void writePage() {
     this.rowsWrittenSoFar += pageRowCount;
     if (DEBUG) LOG.debug("write page");
     try {
-      writePage(pageRowCount, valueCount, statistics, repetitionLevelColumn, definitionLevelColumn, dataColumn);
+      writePage(
+          pageRowCount,
+          valueCount,
+          statistics,
+          sizeStatisticsBuilder.build(),
```
shangxinli, line 407: Can we have parity of line 406 and 407? You can use a varaiible in line 407

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -409,4 +428,14 @@ abstract void writePage(
       ValuesWriter definitionLevels,
       ValuesWriter values)
       throws IOException;
+
+  abstract void writePage(
+      int rowCount,
+      int valueCount,
+      Statistics<?> statistics,
+      SizeStatistics sizeStatistics,
```
shangxinli, line 436: There could be some confusion of the two names that sieStatistics is one type of statistics but we are separating them

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
wgtmac: Fixed

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
wgtmac: I agree. As they are also defined in the thrift file, I think it is not a big problem.

wgtmac:
Gentle ping @emkornfield

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,269 @@
...
+        repetitionLevelHistogram.size() == other.repetitionLevelHistogram.size(),
+        "Cannot merge repetitionLevelHistogram with different sizes");
+    Preconditions.checkArgument(
+        definitionLevelHistogram.size() == other.definitionLevelHistogram.size(),
+        "Cannot merge definitionLevelHistogram with different sizes");
+    unencodedByteArrayDataBytes = Math.addExact(unencodedByteArrayDataBytes, other.unencodedByteArrayDataBytes);
+    for (int i = 0; i < repetitionLevelHistogram.size(); i++) {
+      repetitionLevelHistogram.set(
+          i,
+          repetitionLevelHistogram.get(i)
```
emkornfield, line 213: should these be addExact as well

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndex.java
```diff
@@ -64,4 +66,11 @@ public default long getLastRowIndex(int pageIndex, long rowGroupRowCount) {
     int nextPageIndex = pageIndex + 1;
     return (nextPageIndex >= getPageCount() ? rowGroupRowCount : getFirstRowIndex(nextPageIndex)) - 1;
   }
+
+  /**
+   * @param pageIndex
+   *          the index of the page
+   * @return unencoded/uncompressed size for BYTE_ARRAY types; or empty for other types.
```
emkornfield, line 73: It might be worth clarifying that even for BYTE_ARRAY types this value might not have been written?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndexBuilder.java
```diff
@@ -70,6 +72,14 @@ public long getFirstRowIndex(int pageIndex) {
     public int getPageOrdinal(int pageIndex) {
       return pageIndex;
     }
+
+    @Override
+    public Optional<Long> getUnencodedByteArrayDataBytes(int pageIndex) {
+      if (unencodedByteArrayDataBytes == null || unencodedByteArrayDataBytes.length == 0) {
```
emkornfield, line 78: i'm not sure the comparison to 0 here is correct?  Can't a page be composed of entirely empty strings?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -241,6 +265,30 @@ public void writePageV2(
...
+      writePageV2(
+          rowCount,
+          nullCount,
+          valueCount,
+          repetitionLevels,
+          definitionLevels,
+          dataEncoding,
+          data,
+          statistics,
+          null);
```
emkornfield, line 277: not sure of coding conventions in parquet-mr but I find parameter comments (`/*size_statistics=*/null`)  can help readability.

pom.xml
```diff
@@ -593,9 +593,14 @@
               <exclude>org.apache.parquet.hadoop.thrift.TBaseWriteSupport#setThriftClass(org.apache.parquet.conf.ParquetConfiguration,java.lang.Class)</exclude>
               <exclude>org.apache.parquet.proto.ProtoParquetReader#builder(org.apache.hadoop.fs.Path,boolean)</exclude>
               <exclude>org.apache.parquet.proto.ProtoParquetReader#builder(org.apache.parquet.io.InputFile,boolean)</exclude>
-
               <!-- Due to the removal of deprecated methods -->
               <exclude>org.apache.parquet.arrow.schema.SchemaMapping</exclude>
+              <!-- Some new methods are added to interface to support SizeStatistics-->
```
emkornfield, line 598: This is for API breaking changes?

emkornfield:
> Gentle ping @emkornfield

Took another pass through, I'm less familiar with Parquet MR but overall looks ok to me (with the exception of confirming if the one place I found we should be changing length to 0)

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/OffsetIndexBuilder.java
wgtmac: Here the length means number of pages. If there is no page, then there is no unencodedByteArrayDataBytes either.

pom.xml
wgtmac: Yes, japicmp plugin deems adding new method to public interface is a breaking change.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -401,6 +420,7 @@ void writePage() {
     pageRowCount = 0;
   }
 
+  @Deprecated
```
ConeyLiu, line 423: Should we add the specific version that will remove it?

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
ConeyLiu: It seems like you missed this

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,268 @@
...
+     * Add repetition and definition level of a value to the statistics.
+     * It is called when value is null, or the column is not of BYTE_ARRAY type.
+     *
+     * @param repetitionLevel repetition level of the value
+     * @param definitionLevel definition level of the value
+     * @param value value of to be added
+     */
+    public void add(int repetitionLevel, int definitionLevel, Binary value) {
+      add(repetitionLevel, definitionLevel);
+      if (type.getPrimitiveTypeName() == PrimitiveType.PrimitiveTypeName.BINARY) {
```
ConeyLiu, line 137: I mean the doc here is not matched with code.  The value could not be NULL, otherwise will get NPE.

shangxinli:
LGTM

@ConeyLiu @etseidl @emkornfield Do you still have pending comments?

etseidl:
Looks good to me too. I'd still like to see the CLI changed at some point to print the new statistics, but if no one else has cycles to work on that, I could try cleaning up what I have locally.

wgtmac:
@etseidl I have filed https://issues.apache.org/jira/browse/PARQUET-2433 and https://issues.apache.org/jira/browse/PARQUET-2434 as follow-up work items. Will work on them once this PR gets merged.

ConeyLiu (approved):
+1 thanks for the great work.

emkornfield:
I think all of my suggestions have been addressed.  Thanks @wgtmac !

wgtmac:
@gszadovszky It would be good if you can take a look if possible.

pom.xml
gszadovszky: We can avoid exclusions by adding default implementations to the interfaces. We can even add throwing exceptions to enforce implementation. Exclusions tend to remain after a release and makes it hard to catch potential issues in the excluded parts later.

gszadovszky (changes requested):
I have some comments but LGTM overall.

This change is about writing these new statistics. Are there any benefits in actually using them at reading? Do we plan to implement?

Side note: This is yet another statistics that we gather during writing data. We already have min/max statistics, null counts, bloom filter, and now some additional ones. I think, we should implement a centralized builder that we call once for each value/dl/rl and can generate the statistics we need. We need to implement the gathering part as optimal as it can be. One check less can significantly decrease write performance.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -0,0 +1,268 @@
...
+     * It is called when the column is of BYTE_ARRAY type.
+     *
+     * @param repetitionLevel repetition level of the value
+     * @param definitionLevel definition level of the value
+     * @param value value of to be added
+     */
+    public void add(int repetitionLevel, int definitionLevel, Binary value) {
+      add(repetitionLevel, definitionLevel);
+      if (type.getPrimitiveTypeName() == PrimitiveType.PrimitiveTypeName.BINARY && value != null) {
+        unencodedByteArrayDataBytes = Math.addExact(unencodedByteArrayDataBytes, value.length());
```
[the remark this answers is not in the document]
wgtmac: I tend to keep this check just in case.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2255,27 +2269,47 @@ public static org.apache.parquet.internal.column.columnindex.ColumnIndex fromPar
...
     for (int i = 0, n = offsetIndex.getPageCount(); i < n; ++i) {
       pageLocations.add(new PageLocation(
           offsetIndex.getOffset(i), offsetIndex.getCompressedPageSize(i), offsetIndex.getFirstRowIndex(i)));
+      Optional<Long> unencodedByteArrayDataType = offsetIndex.getUnencodedByteArrayDataBytes(i);
+      if (unencodedByteArrayDataType.isPresent() && unencodedByteArrayDataTypes.size() == i) {
+        unencodedByteArrayDataTypes.add(unencodedByteArrayDataType.get());
+      }
+    }
+    OffsetIndex parquetOffsetIndex = new OffsetIndex(pageLocations);
+    if (unencodedByteArrayDataTypes.size() == pageLocations.size()) {
```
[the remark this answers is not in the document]
wgtmac: This method converts an internal OffsetIndex to a generated thrift OffsetIndex object which does not provide a builder. Therefore we need to check if the unencodedByteArrayDataTypes list is malformed since it may be used in the rewriter where the OffsetIndex was created somewhere else (e.g. from other implementations).

pom.xml
wgtmac: Thanks for the suggestion! Fixed.

wgtmac:
Thanks for the feedback! I've addressed all the comments and added a new internal `ColumnValueCollector` class for common stats collections. @gszadovszky

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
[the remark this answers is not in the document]
gszadovszky: In most cases, this good have checks like this one. But it can significantly hit performance when used in places called regularly. This method is invoked for every values. We shall be as effective as possible.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
```diff
@@ -0,0 +1,139 @@
...
+
+  void write(boolean value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+  }
+
+  void write(int value, int repetitionLevel, int definitionLevel) {
+    statistics.updateStats(value);
+    sizeStatisticsBuilder.add(repetitionLevel, definitionLevel);
+    if (bloomFilter != null) {
```
gszadovszky, line 89: What do you think about using a no-op BloomFilter implementation instead of a null-check? I am not sure if would perform better, though.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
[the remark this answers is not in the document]
wgtmac: Make sense. I have removed the overflow check.

gszadovszky (approved):
Thank you, @wgtmac!

PR #1181, opened 2023-10-31
PARQUET-2371: Resolve japicmp CI failure

amousavigourabi:

### Jira

### Tests

        No user facing changes, only resolves CI failure.

### Commits

### Documentation

---

Japicmp failed on newly introduced classes, during review this did not cause a failure. This likely has something to do with the japicmp 0.18.1 update we pushed since then (for this, see #1176). Includes bumping japicmp to 0.18.2.
The false positives were all under the banner of `CLASS_GENERIC_TEMPLATE_CHANGED`. Similar problems seem to have been reported before at siom79/japicmp#345.


amousavigourabi:
@ConeyLiu @wgtmac, this is a patch for the problems we discussed at #1141.

pom.xml
```diff
@@ -68,7 +68,7 @@
     <jackson.package>com.fasterxml.jackson</jackson.package>
     <jackson.version>2.15.2</jackson.version>
     <jackson-databind.version>2.15.2</jackson-databind.version>
-    <japicmp.version>0.18.1</japicmp.version>
+    <japicmp.version>0.18.2</japicmp.version>
```
ConeyLiu, line 71: After reverting the version back to 0.16.0, the problem is gone.

ConeyLiu (approved):
I am fine with these changes if we want to keep the `japicmp` version. Or else we should revert the version change.

wgtmac (approved):
+1 on my side.

@Fokko @gszadovszky Do you have any comment?

pom.xml
```diff
@@ -546,6 +546,13 @@
               <!-- japicmp is overly aggressive on interface types in signatures, a type was changed to a supertype but this still triggers it -->
               <exclude>org.apache.parquet.hadoop.ColumnChunkPageWriteStore</exclude>
               <exclude>org.apache.parquet.hadoop.ParquetRecordWriter</exclude>
+              <!-- likely japicmp bug, triggers on new interface methods after updating to 0.18.1 -->
+              <exclude>org.apache.parquet.conf.ParquetConfiguration#getClass(java.lang.String,java.lang.Class,java.lang.Class)</exclude>
+              <exclude>org.apache.parquet.hadoop.util.SerializationUtil#readObjectFromConfAsBase64(java.lang.String,org.apache.parquet.conf.ParquetConfiguration)</exclude>
+              <exclude>org.apache.parquet.conf.HadoopParquetConfiguration#getClass(java.lang.String,java.lang.Class,java.lang.Class)</exclude>
+              <exclude>org.apache.parquet.avro.AvroParquetReader#builder(org.apache.parquet.io.InputFile,org.apache.parquet.conf.ParquetConfiguration)</exclude>
+              <exclude>org.apache.parquet.util.Reflection#newInstance(java.lang.Class)</exclude>
```
wgtmac, line 554: This class is removed in this PR. Should this line be deleted as well?

wgtmac, line 554:
```suggestion
```

wgtmac:
I'll merge this when the CI is green to unblock CIs of other PRs.

gszadovszky:
Sorry @wgtmac, @ConeyLiu, for the late response.
I don't think it is a bug in japicmp to fail with a new method in an existing interface. It would fail a 3rd party if it implements the interface. To avoid keep excluding files we might add a default implementation for the new methods instead, if possible.

amousavigourabi:
@gszadovszky, the issue at hand was not that adding methods to already existing interfaces caused failures. As you noted, this would be perfectly fine and expected behaviour. Instead, japicmp failed on wholly new interfaces (see `org.apache.parquet.conf.ParquetConfiguration`), and new method implementations (see `org.apache.parquet.hadoop.thrift.TBaseWriteSupport`), under quite a vague motivation `CLASS_GENERIC_TEMPLATE_CHANGED`.

gszadovszky:
Thanks for the clarification, @amousavigourabi. We cannot do anything but excluding these classes, then, I'm afraid. But the comment is a bit misleading.
Have you tried creating a bug for japicmp about it?

PR #1187, opened 2023-11-10
PARQUET-2374: Add metrics support for parquet file reader

parthchandra:

### Jira

This PR addresses the following: [PARQUET-2374](https://issues.apache.org/jira/browse/PARQUET-2374)
Add metrics support for parquet file reader
### Tests
No new tests. Adds a new public interface for reporting metrics from the parquet file reader

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -306,7 +306,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
 
     private void setDecompressMetrics(BytesInput bytes, long start) {
       if (metricsCallback != null) {
-        long time = System.nanoTime() - start;
+        long time = Math.max(System.nanoTime() - start, 0);
         long len = bytes.size();
         double throughput = ((double) len / time) * ((double) 1000_000_000L) / (1024 * 1024);
         LOG.debug(
@@ -314,7 +314,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
             len / (1024 * 1024),
             time / 1000_000L,
             throughput);
-        metricsCallback.setValueLong(ParquetFileReaderMetrics.DecompressTime.name(), time);
+        metricsCallback.setDuration(ParquetFileReaderMetrics.DecompressTime.name(), time);
         metricsCallback.setValueLong(ParquetFileReaderMetrics.DecompressSize.name(), len);
         metricsCallback.setValueDouble(ParquetFileReaderMetrics.DecompressThroughput.name(), throughput);
       }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -77,7 +77,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
     private final BlockCipher.Decryptor blockDecryptor;
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
-    ParquetMetricsCallback metricsCallback;
+    private final ParquetMetricsCallback metricsCallback;
 
     ColumnChunkPageReader(
         BytesInputDecompressor decompressor,
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -77,7 +77,6 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
     private final BlockCipher.Decryptor blockDecryptor;
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
-//     private final ParquetMetricsCallback metricsCallback;
 
     ColumnChunkPageReader(
         BytesInputDecompressor decompressor,
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -77,6 +77,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
     private final BlockCipher.Decryptor blockDecryptor;
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
+    ParquetMetricsCallback metricsCallback;
 
     ColumnChunkPageReader(
         BytesInputDecompressor decompressor,
@@ -88,7 +89,8 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
         byte[] fileAAD,
         int rowGroupOrdinal,
         int columnOrdinal,
-        ParquetReadOptions options) {
+        ParquetReadOptions options,
+        ParquetMetricsCallback callback) {
       this.decompressor = decompressor;
       this.compressedPages = new ArrayDeque<DataPage>(compressedPages);
       this.compressedDictionaryPage = compressedDictionaryPage;
@@ -110,6 +112,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
         dataPageAAD = null;
         dictionaryPageAAD = null;
       }
+      this.metricsCallback = callback;
     }
 
     private int getPageOrdinal(int currentPageIndex) {
@@ -156,11 +159,13 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
 
               ByteBuffer decompressedBuffer =
                   options.getAllocator().allocate(dataPageV1.getUncompressedSize());
+              long start = System.nanoTime();
               decompressor.decompress(
                   byteBuffer,
                   (int) compressedSize,
                   decompressedBuffer,
                   dataPageV1.getUncompressedSize());
+              setDecompressMetrics(bytes, start);
 
               // HACKY: sometimes we need to do `flip` because the position of output bytebuffer is
               // not reset.
@@ -172,7 +177,9 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
               if (null != blockDecryptor) {
                 bytes = BytesInput.from(blockDecryptor.decrypt(bytes.toByteArray(), dataPageAAD));
               }
+              long start = System.nanoTime();
               decompressed = decompressor.decompress(bytes, dataPageV1.getUncompressedSize());
+              setDecompressMetrics(bytes, start);
             }
 
             final DataPageV1 decompressedPage;
@@ -234,8 +241,10 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
                     - dataPageV2.getRepetitionLevels().size());
                 ByteBuffer decompressedBuffer =
                     options.getAllocator().allocate(uncompressedSize);
+                long start = System.nanoTime();
                 decompressor.decompress(
                     byteBuffer, (int) compressedSize, decompressedBuffer, uncompressedSize);
+                setDecompressMetrics(pageBytes, start);
 
                 // HACKY: sometimes we need to do `flip` because the position of output bytebuffer is
                 // not reset.
@@ -255,7 +264,9 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
                 int uncompressedSize = Math.toIntExact(dataPageV2.getUncompressedSize()
                     - dataPageV2.getDefinitionLevels().size()
                     - dataPageV2.getRepetitionLevels().size());
+                long start = System.nanoTime();
                 pageBytes = decompressor.decompress(pageBytes, uncompressedSize);
+                setDecompressMetrics(pageBytes, start);
               }
             }
           } catch (IOException e) {
@@ -293,6 +304,22 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
       });
     }
 
+    private void setDecompressMetrics(BytesInput bytes, long start) {
+      if (metricsCallback != null) {
+        long time = System.nanoTime() - start;
+        long len = bytes.size();
+        double throughput = ((double) len / time) * ((double) 1000_000_000L) / (1024 * 1024);
+        LOG.debug(
+            "Decompress block: Length: {} MB, Time: {} msecs, throughput: {} MB/s",
+            len / (1024 * 1024),
+            time / 1000_000L,
+            throughput);
+        metricsCallback.setValueLong(ParquetFileReaderMetrics.DecompressTime.name(), time);
+        metricsCallback.setValueLong(ParquetFileReaderMetrics.DecompressSize.name(), len);
+        metricsCallback.setValueDouble(ParquetFileReaderMetrics.DecompressThroughput.name(), throughput);
+      }
+    }
+
     @Override
     public DictionaryPage readDictionaryPage() {
       if (compressedDictionaryPage == null) {
@@ -303,6 +330,10 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
         if (null != blockDecryptor) {
           bytes = BytesInput.from(blockDecryptor.decrypt(bytes.toByteArray(), dictionaryPageAAD));
         }
+        long start = System.nanoTime();
+        BytesInput decompressed =
+            decompressor.decompress(bytes, compressedDictionaryPage.getUncompressedSize());
+        setDecompressMetrics(bytes, start);
         DictionaryPage decompressedPage = new DictionaryPage(
             decompressor.decompress(bytes, compressedDictionaryPage.getUncompressedSize()),
             compressedDictionaryPage.getDictionarySize(),
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -77,7 +77,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
     private final BlockCipher.Decryptor blockDecryptor;
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
-    private final ParquetMetricsCallback metricsCallback;
+//     private final ParquetMetricsCallback metricsCallback;
 
     ColumnChunkPageReader(
         BytesInputDecompressor decompressor,
@@ -89,8 +89,7 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
         byte[] fileAAD,
         int rowGroupOrdinal,
         int columnOrdinal,
-        ParquetReadOptions options,
-        ParquetMetricsCallback callback) {
+        ParquetReadOptions options) {
       this.decompressor = decompressor;
       this.compressedPages = new ArrayDeque<DataPage>(compressedPages);
       this.compressedDictionaryPage = compressedDictionaryPage;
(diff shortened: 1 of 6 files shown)


parthchandra:
Note that this can be used to compare read performance of the existing file reader with the the hadoop vector io reader https://github.com/apache/parquet-mr/pull/1139

parthchandra:
@shangxinli could you take a look at this PR?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -237,21 +246,22 @@ public DataPage visit(DataPageV2 dataPageV2) {
               }
             } else {
               if (null != blockDecryptor) {
-                pageBytes = BytesInput.from(
-                    blockDecryptor.decrypt(pageBytes.toByteArray(), dataPageAAD));
+                  pageBytes = BytesInput.from(
```
wgtmac, line 249: ditto

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -237,21 +246,22 @@ public DataPage visit(DataPageV2 dataPageV2) {
...
+                      blockDecryptor.decrypt(pageBytes.toByteArray(), dataPageAAD));
               }
               if (dataPageV2.isCompressed()) {
                 int uncompressedSize = Math.toIntExact(
                     dataPageV2.getUncompressedSize()
                         - dataPageV2.getDefinitionLevels().size()
                         - dataPageV2.getRepetitionLevels().size());
+                long start = System.nanoTime();
                 pageBytes = decompressor.decompress(pageBytes, uncompressedSize);
+                setDecompressMetrics(pageBytes, start);
```
wgtmac, line 259: Do we need to include decompression of dictionary page?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1870,6 +1886,23 @@ public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOEx
...
+      if (metricsCallback != null) {
+        long totalFileReadTimeNs = System.nanoTime() - startNs;
+        double sizeInMb = ((double) length) / (1024 * 1024);
+        double timeInSec = ((double) totalFileReadTimeNs) / 1000_0000_0000L;
+        double throughput = sizeInMb / timeInSec;
+        LOG.debug(
+          "Parquet: File Read stats:  Length: {} MB, Time: {} secs, throughput: {} MB/sec ",
+          sizeInMb,
+          timeInSec,
+          throughput);
```
wgtmac, line 1899:
```suggestion
        LOG.debug(
          "Parquet: File Read stats:  Length: {} MB, Time: {} secs, throughput: {} MB/sec ",
          sizeInMb, timeInSec, throughput);
```

wgtmac (changes requested):
Thanks for adding this! I have left a few comments.

parthchandra (commented):
Thank you for reviewing @wgtmac !
Addressed your comments. There was some weird reformatting that occurred. I've fixed those in places you pointed out as well as some others.

wgtmac (approved):
Thanks @parthchandra! LGTM

parthchandra:
Thank you @wgtmac !

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -125,10 +125,20 @@ public class ParquetFileReader implements Closeable {
...
+
   private final ParquetMetadataConverter converter;
 
   private final CRC32 crc;
 
+
+  /**
+   * set a callback to send back metrics info
+   */
+  public synchronized void initMetrics(ParquetMetricsCallback callback) {
```
ConeyLiu, line 138: why the `synchronized` is needed?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -125,10 +125,20 @@ public class ParquetFileReader implements Closeable {
 
   public static String PARQUET_READ_PARALLELISM = "parquet.metadata.read.parallelism";
 
+  public ParquetMetricsCallback metricsCallback;
```
ConeyLiu, line 128: do we need this field as `public`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -80,10 +80,12 @@ static final class ColumnChunkPageReader implements PageReader {
     private final byte[] dataPageAAD;
     private final byte[] dictionaryPageAAD;
 
+    ParquetMetricsCallback metricsCallback;
```
ConeyLiu, line 83: It should be `private final`

steveloughran:
it'd be really nice if somehow there was a way to push hadoop stream IOStats here, especially the counters, min, max and mean maps: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/iostatistics.html

and its really interesting for s3, azure and gcs clients, where we collect stream specific stuff, including things like: bytes discarded in seek, time for GET, whether we did a HEAD first, and more. These are collected in a thread level, but also include stats from helper threads such as those in async stream draining, vector IO...

It'd take a move to hadoop 3.3.1+ to embrace the API, but if there was a way for something to publish stats to your metric collector, then maybe something could be done

Tip: you can enable a dump of a filesystem's aggregate stats in process shutdown for azure and s3a
```
fs.iostatistics.logging.level=info
```

```
2023-11-17 18:30:28,634 [shutdown-hook-0] INFO  statistics.IOStatisticsLogging (IOStatisticsLogging.java:logIOStatisticsAtLevel(269)) - IOStatistics: counters=((action_http_head_request=3)
(audit_request_execution=15)
(audit_span_creation=12)
(object_list_request=12)
(object_metadata_request=3)
(op_get_file_status=1)
(op_glob_status=1)
(op_list_status=9)
(store_io_request=15));

gauges=();

minimums=((action_http_head_request.min=22)
(object_list_request.min=25)
(op_get_file_status.min=1)
(op_glob_status.min=9)
(op_list_status.min=25));

maximums=((action_http_head_request.max=41)
(object_list_request.max=398)
(op_get_file_status.max=1)
(op_glob_status.max=9)
(op_list_status.max=408));

means=((action_http_head_request.mean=(samples=3, sum=87, mean=29.0000))
(object_list_request.mean=(samples=12, sum=708, mean=59.0000))
(op_get_file_status.mean=(samples=1, sum=1, mean=1.0000))
(op_glob_status.mean=(samples=1, sum=9, mean=9.0000))
(op_list_status.mean=(samples=9, sum=814, mean=90.4444)));
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1841,8 +1851,12 @@ public void addChunk(ChunkDescriptor descriptor) {
      * @throws IOException if there is an error while reading from the stream
      */
     public void readAll(SeekableInputStream f, ChunkListBuilder builder) throws IOException {
+      long seekStart = System.nanoTime();
```
steveloughran, line 1854: s3a and abfs do lazy seek; no IO until the first read/readFully. best to measure that.

steveloughran (commented):
For the object stores, things to measure are
* time to open() and close() a file
* time for a read after a backwards seek
* time for a read after a forwards seek.
* how many reads actually took place
* for vector IO, whatever gets picked up there
* were errors reported and retried, or throttling events
* number of underlying GET requests

as well as publishing this through the IOStatisticsSource API, we also now collect it at the thread level (IOStatisticsContext) with the goal of having the execution engine collect all stats for a task and then aggregate across the job. you can set the s3a committers up to do this.

If you were for a recent hadoop release only I'd say "publish the stream stats to the thread context for aggregation", but that's not viable here. What would be good if this stats was set up to
* take maps of key-value rather than a fixed enum
* collect those min/mean/max as well as counts.
* somehow provided a plugin point where we could add something to add any of the parquet reader/writer stats to the thread context -trying to collect stats from inside wrapped-many-times-over streams and iterators is way too complex. I know, i have a branch of parquet where I tried that...
*

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
parthchandra: Changed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
parthchandra: No we don't. Made it private

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
parthchandra: My mistake. I initially implemented the metrics callback as a singleton. The method no longer needs to be synchronized.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
parthchandra: Right. Generally, seeks (especially backwards) cause the file system to stop their read-ahead and turn off sequential read optimizations. The seek call itself doesn't take much time.

parthchandra:
@steveloughran I did look into leveraging Hadoop io stats  but my first attempt did not work too well and I thought a simpler initial implementation would be more useful. Once we move to hadoop vector io, I'll take another stab at it. 

> What would be good if this stats was set up to
> 
> take maps of key-value rather than a fixed enum

The fixed enum here is simply the Parquet file reader providing information that these are the values it knows about. This implementation is not really collecting and aggregating anything, it is simply recording the time and counts and passing them on. 
 
> collect those min/mean/max as well as counts.

The implementation of the parquet metrics callback will do that. So if the execution engine is Spark, it can simply get the values and add them to it's own metrics collection subsystem which then computes the min/max/mean.

> somehow provided a plugin point where we could add something to add any of the parquet reader/writer stats to the thread context -trying to collect stats from inside wrapped-many-times-over streams and iterators is way too complex. I know, i have a branch of parquet where I tried that...

Hmm, that will take some work. I wanted to measure streaming decompression time (where the `decompress` call simply returns a stream which is decompressed as it is read), but found it required too many breaking changes to implement. But a standard system like `IOStatistics` where such a stream is a IOStatisticsSource would be perfect.

wgtmac:
> For the object stores, things to measure are
> 
> * time to open() and close() a file
> * time for a read after a backwards seek
> * time for a read after a forwards seek.
> * how many reads actually took place
> * for vector IO, whatever gets picked up there
> * were errors reported and retried, or throttling events
> * number of underlying GET requests
> 

CMIW, it seems that these stats can be collected solely at the input stream level.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
wgtmac: Did you forget to push your local changes? @parthchandra

parthchandra: Yes, I just pushed them now. 🤦🏾

parthchandra:
> > For the object stores, things to measure are
> > 
> > * time to open() and close() a file
> > * time for a read after a backwards seek
> > * time for a read after a forwards seek.
> > * how many reads actually took place
> > * for vector IO, whatever gets picked up there
> > * were errors reported and retried, or throttling events
> > * number of underlying GET requests
> 
> CMIW, it seems that these stats can be collected solely at the input stream level.

Yes, they are best collected by the file system client API. However it would be nice to be able to hook up all these metrics together. Then we could, for instance, show a single Spark scan operator that displays stats for the operator, parquet reader, and the input stream in one place.

wgtmac:
Thanks @parthchandra! Do you have any TODO work item on this (or after vectored I/O is merged)?

parthchandra:
No I don't have any todo for this. If we are going to merge vector IO soon, I can add the metrics for that as well as part of this PR.

wgtmac:
> No I don't have any todo for this. If we are going to merge vector IO soon, I can add the metrics for that as well as part of this PR.

No necessary. We can add them progressively.

@steveloughran @shangxinli @gszadovszky Do you have any concern to merge this?

parthchandra:
@wgtmac do you think we can merge this?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -123,10 +123,19 @@ public class ParquetFileReader implements Closeable {
...
+  private ParquetMetricsCallback metricsCallback;
+
   private final ParquetMetadataConverter converter;
 
   private final CRC32 crc;
 
+  /**
+   * set a callback to send back metrics info
+   */
+  public void initMetrics(ParquetMetricsCallback callback) {
```
ConeyLiu, line 135: Why not set the `callback` in `ParquetReadOptions`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
ConeyLiu: It would be better to mark it as final. I suggest getting `callback` from `ParquetReadOptions` and then we don't need to update the constructor and more reasonable.

shangxinli:
LGTM

Thanks @parthchandra for working on this PR. And thanks @wgtmac and @ConeyLiu for reviewing the changes.

shangxinli (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
parthchandra: That's a good idea, but it may require changes to some of the public constructors which are in use by others. For instance Spark has this https://github.com/apache/spark/blob/15e99cf676d9de02c54ca5ebe9a2bc6a3ce014e5/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/SpecificParquetRecordReaderBase.java#L104
In this case, the constructor is deprecated and it looks like Spark could benefit from a constructor that takes a ParquetReadOptions as a parameter. I can introduce a new constructor but wasn't sure that we would want to do so. Having an additional API to set the metrics callback seemed like the simpler option.
WDYT?

wgtmac: @ConeyLiu Gentle ping, do you have any comment?

ConeyLiu: Sorry for missing this. https://github.com/apache/parquet-mr/blob/afd39dde8fd762bf696fea3dab16d45eae1093c3/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L669
The `open` method passing a `footer` is deprecated. I think we should support it with `ParquetReadOptions`. Then Spark could benefit from this as well. 

I am OK with the current changes to unblock this PR.

ConeyLiu:
@parthchandra one more question, why reverted changes for https://github.com/apache/parquet-mr/pull/1187#discussion_r1396756877?

parthchandra:
> @parthchandra one more question, why reverted changes for [#1187 (comment)](https://github.com/apache/parquet-mr/pull/1187#discussion_r1396756877)?

Hmm. I'm sure I didn't revert that. Let me address this.

Also, I think i will change this to use `ParquetReadOptions`, and add an additional `open` API for use by Spark. So let's hold off on merging this.

steveloughran (commented):
just realised i had a comment but didn't submit it...here you go. Sorry

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetMetricsCallback.java
```diff
@@ -0,0 +1,34 @@
...
+ * under the License.
+ */
+package org.apache.parquet.hadoop;
+
+/**
+ *  a simple interface to pass basic metric values by name to any implementation. Typically an
+ *  implementation of this interface will serve as a bridge to pass metric values on
+ *  to the metrics system of a distributed engine (hadoop, spark, etc).
+ */
+public interface ParquetMetricsCallback {
```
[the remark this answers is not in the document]
parthchandra: > Can you tag as unstable but promise that all new callbacks will have a default implementation for any new metric

Done. 

Re: nanotimer - Learnt something new today. I remember when nanosecond timers were introduced things were unstable, but hopefully things are more reliable nowadays. I usually look at Guava's Stopwatch as the baseline implementation and the nanotimer implementation doesn't have a check for getting a negative diff either.  Anyhow, I added a check for that so we get at least zero.

parthchandra (commented):
@ConeyLiu, @wgtmac, changed the implementation to use ParqutReadOptions and added a new constructor that takes a footer and read options as parameters (for use by Spark).
I checked Iceberg as well and it looks like it doesn't use any deprecated APIs.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -88,7 +89,8 @@ static final class ColumnChunkPageReader implements PageReader {
         byte[] fileAAD,
         int rowGroupOrdinal,
         int columnOrdinal,
-        ParquetReadOptions options) {
+        ParquetReadOptions options,
+        ParquetMetricsCallback callback) {
```
ConeyLiu, line 93: Why not get the `callback` by the `options.getMetricsCallback`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -823,6 +823,38 @@ public ParquetFileReader(Configuration conf, Path file, ParquetMetadata footer)
     this.crc = options.usePageChecksumVerification() ? new CRC32() : null;
   }
 
+  /**
+   * @param conf   the Hadoop Configuration
+   * @param file   Path to a parquet file
+   * @param footer a {@link ParquetMetadata} footer already read from the file
+   * @throws IOException if the file can not be opened
+   * @deprecated will be removed in 2.0.0.
```
ConeyLiu, line 831: Will this eventually be replaced by another constructor or just not recommended?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
parthchandra: Of course. My oversight.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
parthchandra: Oops. Copied the old implementation and did not update the javadoc. Fixed.

parthchandra:
Hi @ConeyLiu, @wgtmac, any additional concerns? If none, should we merge this?

ConeyLiu (approved):
+1, thanks @parthchandra  LGTM

wgtmac:
Sorry for the delay. I just merged this. Thanks @parthchandra!

parthchandra:
Thank you!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetMetricsCallback.java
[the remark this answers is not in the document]
steveloughran: +1
I should revisit state of the art there, and make sure that hadoop duration timers are good at -ve clock changes. [past commentary on 2015 CPUs](jhttps://steveloughran.blogspot.com/2015/09/time-on-multi-core-multi-socket-servers.html)

maybe something like some soft referenced "source of time" per thread which is both (systime, nanotime) and when querying it if current nanotime is "close enough" to previous nanotime, just apply nanotime diff to systime; update systime fields. So it'd be "approximate per cpu core/socket", enough for metrics where things happen in millis

PR #1194, opened 2023-11-15
PARQUET-2382: Remove the deprecated `OriginalType`

Fokko:

For Iceberg we're adding nanosecond timestamps. During my investigation in Parquet, I noticed that there are two ways of declaring logical types:

1. Through the deprecated `OriginalType`
2. Using the new LogicalTypeAnnotation API.

The old API does not support nano's but is still used in downstream projects, such as Parquet, where I'm working on migrating to the new API: https://github.com/apache/iceberg/pull/9063

However, since it was five years ago in [PARQUET-1452](https://issues.apache.org/jira/browse/PARQUET-1452) when it was marked as deprecated, released in Parquet 1.11.0. I would love to remove the old API to make sure that downstream engines migrate to the new API and handle nano's correctly.

### Jira

### Tests

### Commits

### Documentation


parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1715,15 +1729,15 @@ private void buildChildren(Types.GroupBuilder builder,
         childBuilder.as(getLogicalTypeAnnotation(schemaElement.logicalType));
       }
       if (schemaElement.isSetConverted_type()) {
-        OriginalType originalType = getLogicalTypeAnnotation(schemaElement.converted_type, schemaElement).toOriginalType();
-        OriginalType newOriginalType = (schemaElement.isSetLogicalType() && getLogicalTypeAnnotation(schemaElement.logicalType) != null) ?
-           getLogicalTypeAnnotation(schemaElement.logicalType).toOriginalType() : null;
-        if (!originalType.equals(newOriginalType)) {
```
Fokko, line 1721: I think there was a 🐛 hidden here. Because we would only compare on the logical type itself, and not its properties (precision/scale for decimal, or adjust-for-utc for time/timestamp). Tests started failing, therefore I added the `getAdjustToUtc` to retrieve the actual value from the Parquet structure.

wgtmac:
I am fine with it but I'd like to seek advices from @gszadovszky @shangxinli

gszadovszky:
I don't think it is a good idea to remove deprecated API in a minor release. That's why we have japicmp to ensure backward compatibility.
I think, there is no harm for Parquet if they use the old `OriginalType`s. If we enforce significant code changes in minor releases we would also slow down the upgrades.

Fokko:
Makes sense. I think it would be good to remove `OriginalType` at some point. Let's target this PR for 2.0 and leave it for now. I'll create another PR to mark the `getOriginalType()` deprecated (this one was marked as private by Yetus before, but I think it would be best to mark them as deprecated for 1.14.0).

PR #1230, opened 2023-12-04
PARQUET-2405: Include fallback CodecFactory implementation

amousavigourabi:

### Jira

### Tests

### Commits

### Style

### Documentation


wgtmac (approved):

wgtmac:
The implementation looks good. But it seems not used anywhere except the test. Do you want to include its usage in this PR as well?

amousavigourabi:
> The implementation looks good. But it seems not used anywhere except the test. Do you want to include its usage in this PR as well?

Thanks for the review @wgtmac ! The implementation is meant s.t. users can more easily implement and use partial CodecFactories themselves, without worrying about encountering codecs they did not include.

wgtmac:
Thanks for explanation! So users can pass the fallback CodecFactory to the parquet reader/writer now.

parquet-common/src/main/java/org/apache/parquet/compression/FallbackCodecFactory.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+package org.apache.parquet.compression;
+
+import org.apache.parquet.Preconditions;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
+
+/**
+ * {@link CompressionCodecFactory} implementation that delegates to the {@link CompressionCodecFactory} implementations it was constructed with.
+ */
+public class FallbackCodecFactory implements CompressionCodecFactory {
```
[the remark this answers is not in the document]
ConeyLiu: It seems the usage for this is people don't care about which compression codec is used.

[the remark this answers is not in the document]
amousavigourabi: Not necessarily don't care, but who might not have guarantees that any system might be able to run their preferred codecs. If you are to implement one using hardware acceleration there are platforms for which this might not work perfectly. This factory provides an easy and modular way to allow for more stable fallbacks for these scenarios OOTB.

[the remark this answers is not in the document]
wgtmac: I agree with @ConeyLiu. Users can anyway do the same thing all by themselves. It seems sufficient when parquet reader and writer support passing a customized codec factory.

parquet-common/src/main/java/org/apache/parquet/compression/FallbackCodecFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        "FallbackCodecFactory needs to be constructed using at least one CompressionCodecFactory to fall back on.");
+    this.codecFactories = codecFactories;
+  }
+
+  @Override
+  public BytesInputCompressor getCompressor(CompressionCodecName codecName) {
+    for (CompressionCodecFactory codecFactory : codecFactories) {
+      try {
+        return codecFactory.getCompressor(codecName);
+      } catch (CodecNotImplementedException ignored) {
```
wgtmac, line 49: Here it won't catch `CodecNotImplementedException` unless the internal codecFactory is also a FallbackCodecFactory, which sounds weird to me.

parquet-common/src/main/java/org/apache/parquet/compression/FallbackCodecFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        codecFactories != null && codecFactories.length > 0,
+        "FallbackCodecFactory needs to be constructed using at least one CompressionCodecFactory to fall back on.");
+    this.codecFactories = codecFactories;
+  }
+
+  @Override
+  public BytesInputCompressor getCompressor(CompressionCodecName codecName) {
+    for (CompressionCodecFactory codecFactory : codecFactories) {
+      try {
+        return codecFactory.getCompressor(codecName);
```
wgtmac, line 48: What if it returns null? Should it fallback as well?

parquet-common/src/main/java/org/apache/parquet/compression/FallbackCodecFactory.java
amousavigourabi: The exception is a way for these specialized CompressionCodecFactory implementations to signal they fail to resolve a compressor for this request. The current implementations of CompressionCodecFactory we ship OOTB are pretty universal and do not run into any situations where throwing this would be appropriate.

parquet-common/src/main/java/org/apache/parquet/compression/FallbackCodecFactory.java
amousavigourabi: I'm not sure whether that would be a good idea, using the custom exception here seems more idiomatic. Using null returns in codec factories would probably also run into relatively hard-to-diagnose issues when used without such fallback mechanisms, so I'd prefer to discourage null returns as much as possible when (de)compressors cannot be resolved for these specialized user implementations.

wgtmac (changes requested):
I just did another pass. The fallback logic seems to be very customized and should not reside in the parquet-common module. The CodecFactory has a good enough abstraction so it would be trivial for users to implement their own logic like what this PR does. But it may introduce maintenance burden if we want to check in something like this. Sorry but I'm -1 on it now.

amousavigourabi:
@wgtmac thanks for looking at it, I'll close the PR.

PR #1240, opened 2023-12-11
PARQUET-2412: Reduce excessive synchronization in MemoryManager

ravwojdyla:
The surface of the synchronized methods is unnecessarily large which in case of large number of concurrent writers may be suboptimal, threads may starve each other for access to the writerList (if writers are being dynamically closed and created). Please see the profile reports in the Jira issue. The fix in this PR is to use `ConcurrentHashMap` for `MemoryManager`'s `writerList` instead of sync methods.

### Jira

- [x] My PR addresses the following https://issues.apache.org/jira/browse/PARQUET-2412 issues and references
  them in the PR title.

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason: there are already tests for `MemoryManager`.

### Commits

  from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### Style

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,7 +47,7 @@ public class MemoryManager {
 
   private final long totalMemoryPool;
   private final long minMemoryAllocation;
-  private final Map<InternalParquetRecordWriter<?>, Long> writerList = new HashMap<>();
+  private final Map<InternalParquetRecordWriter<?>, Long> writerList = new ConcurrentHashMap<>();
   private final Map<String, Runnable> callBacks = new HashMap<String, Runnable>();
   private double scale = 1.0;
 
@@ -74,7 +75,7 @@ public class MemoryManager {
    * @param writer     the new created writer
    * @param allocation the requested buffer size
    */
-  synchronized void addWriter(InternalParquetRecordWriter<?> writer, Long allocation) {
+  void addWriter(InternalParquetRecordWriter<?> writer, Long allocation) {
     Long oldValue = writerList.get(writer);
     if (oldValue == null) {
       writerList.put(writer, allocation);
@@ -92,7 +93,7 @@ public class MemoryManager {
    *
    * @param writer the writer that has been closed
    */
-  synchronized void removeWriter(InternalParquetRecordWriter<?> writer) {
+  void removeWriter(InternalParquetRecordWriter<?> writer) {
     writerList.remove(writer);
     if (!writerList.isEmpty()) {
       updateAllocation();


ravwojdyla:
To validate this change, I have tried to write partitioned dataset (32 concurrent processes, 200 files per 32 partitions) with and without this patch. Here's lock stats for **without** the patch:

image

And here's stats **with** the patch:

image

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
```diff
@@ -74,7 +75,7 @@ private void checkRatio(float ratio) {
    * @param writer     the new created writer
    * @param allocation the requested buffer size
    */
-  synchronized void addWriter(InternalParquetRecordWriter<?> writer, Long allocation) {
+  void addWriter(InternalParquetRecordWriter<?> writer, Long allocation) {
```
ConeyLiu, line 78: This is not the same. The `synchronized` locks the whole method. While the `ConcurrentHashMap` could only guarantee the safety update here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
```diff
@@ -74,7 +75,7 @@ private void checkRatio(float ratio) {
...
     Long oldValue = writerList.get(writer);
     if (oldValue == null) {
```
ConeyLiu, line 80: I think here could update concurrently. The `get` and `put` are two methods, it is not an atomic update here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
ravwojdyla: @ConeyLiu thank you for a prompt review. Excited to get this fixed. Agree it's not the same, and further `synchronized` is not locking a whole method BUT a whole **instance** which in the case of MemoryManager is the **single/global instance**, so `synchronized` can be very expensive here. The cost is manifested by the lock stats above. I can see the tests are all green. Is such an "excessive" synchronization pattern on the global instance necessary for the semantic of MemoryManager? Which use-cases require this level of locking, and could that be relaxed? What do you recommend as next steps? Thanks in advance!

ravwojdyla: 👋 @ConeyLiu I hope you're doing well. I wanted to gently follow up on this thread. Your input would be really valuable for moving forward. Thank you.

wgtmac: I agree with @ConeyLiu that there is a behavior change. `updateAllocation()` is not thread-safe and may have race condition after this change. I suspect the unit test does not cover the case where multiple writers are added and deleted concurrently.

wgtmac: Unfortunately I do not have too much experience in using this code path. The MemoryManager was introduced long ago to address memory issue for dynamic partition. I'm not sure if Apache Spark or other engines are using it or not. May I seek advise from @dongjoon-hyun @sunchao @parthchandra @shangxinli to see if you have seen this problem before?

parthchandra: @wgtmac` I'm not too familiar with this part of the code path either. However, it does look that the `updateAllocation` call needs to be synchronized and it requires the `writers` map to not be mutated during that time. 
Spark uses the default behaviour of ParquetOutputFormat and so does end up using the MemoryManager. The impact of this change is not clear to me though.
BTW, @ravwojdyla this is a great bit of investigation you did there! Would it be possible to share the jfr file (in the jira, perhaps)? I'd like to see if I can gain any additional insights from the information you have collected.

ravwojdyla: @parthchandra thanks 🙇I unfortunately don't have those specific files anymore, it's been a couple of months actually (June 2023), apologies. If this PR continues, I may need to run another profiling to validation the patch, I would be happy to share the before and after jfr files.

parthchandra: NP. Can I suggest a unit test that adds/removes writers concurrently? Almost sure that the test is likely to fail because of `updateAllocation`. The question remains whether such concurrent access does, in fact, occur in engines like Spark or Hadoop.

ConeyLiu: Does this have a big impact on the end-to-end performance? I think it should only happen when you are writing a file with many small partitions.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
ConeyLiu: why not `computeIfAbsent`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
ravwojdyla: @ConeyLiu yes it does. One of our production Spark job took close to 2 hours with partitioned parquet output and about 45 minutes without. Also please see the stats in JIRA issues:
 * https://issues.apache.org/jira/browse/PARQUET-2412
 * https://issues.apache.org/jira/browse/SPARK-44003

ConeyLiu: Thanks @ravwojdyla for your sharing. I think the changes should keep the safety. It needs more UTs to verify this.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/MemoryManager.java
steveloughran: putIfAbsent() and only update the allocation if it was put in the map? It's probably that allocation code which is holding on to the lock for so long. But: that `updateAllocation()` operation is possibly the slow part -and it now has to be thread safe if the callers are no longer locked

PR #1241, opened 2023-12-18
PARQUET-2413: Support configurable extraMetadata in ParquetWriter

clairemcginty:
https://issues.apache.org/jira/browse/PARQUET-2413

Adds support for Configurable `extraMetadata` in Parquet file footer. This makes it easier for users to migrate from Avro to Parquet (since Avro [supports custom metadata keys](https://github.com/apache/avro/blob/release-1.11.3/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java#L285)).

I chose this approach (parsing values from a preset `Configuration` key prefix) because (a) `Configuration` is already Stringable, so no need to worry about object-to-String conversion, and (b) it doesn't require any API changes (i.e. adding `withExtraMetadata` Builder options to all Parquet writers/implementations).

Alternate approaches:

- Add `withExtraMetadata` Builder method to `ParquetWriter`; in `ParquetWriter#build`, append all to the value of `WriteContext#getExtraMetadata()`.
- Add `extraMetadata` class variables to all WriteSupport _implementations_, and pass to `WriteContext` in `WriteSupport#init` ([i.e. AvroWriteSupport](https://github.com/apache/parquet-mr/blob/apache-parquet-1.13.1/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L131-L152)).

Let me know if either of those approaches are preferable to this one!

### Jira

### Tests

### Commits

### Style
- [z] My contribution adheres to the code style guidelines and Spotless passes.

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -20,6 +20,8 @@ package org.apache.parquet.hadoop;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.column.ParquetProperties;
@@ -403,15 +405,29 @@ public class ParquetWriter<T> implements Closeable {
 
     this.codecFactory = codecFactory;
     CompressionCodecFactory.BytesInputCompressor compressor = codecFactory.getCompressor(compressionCodecName);
+
+    final String extraMetadataConfPrefix = ParquetOutputFormat.EXTRA_WRITE_METADATA + ".";
+    final Map<String, String> extraMetadata = new HashMap<>(writeContext.getExtraMetaData());
+
+    conf.iterator().forEachRemaining(entry -> {
+      if (entry.getKey().startsWith(extraMetadataConfPrefix)) {
+        final String metadataKey = entry.getKey().replaceFirst(extraMetadataConfPrefix, "");
+
+        if (metadataKey.equals(OBJECT_MODEL_NAME_PROP)) {
+          throw new IllegalArgumentException("Cannot overwrite metadata key " + OBJECT_MODEL_NAME_PROP
+              + ". Please use another key name.");
+        }
+
+        if (extraMetadata.put(metadataKey, entry.getValue()) != null) {
+          throw new IllegalArgumentException("Extra metadata key " + metadataKey
+              + " conflicts with reserved metadata keys present in "
+              + writeSupport.getClass().getName() + ". Please use another key name.");
+        }
+      }
+    });
+
     this.writer = new InternalParquetRecordWriter<T>(
-        fileWriter,
-        writeSupport,
-        schema,
-        writeContext.getExtraMetaData(),
-        rowGroupSize,
-        compressor,
-        validating,
-        encodingProps);
+        fileWriter, writeSupport, schema, extraMetadata, rowGroupSize, compressor, validating, encodingProps);
   }
 
   public void write(T object) throws IOException {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -406,25 +406,25 @@ public class ParquetWriter<T> implements Closeable {
     this.codecFactory = codecFactory;
     CompressionCodecFactory.BytesInputCompressor compressor = codecFactory.getCompressor(compressionCodecName);
 
-    final String extraMetadataConfPrefix = ParquetOutputFormat.EXTRA_WRITE_METADATA + ".";
-    final Map<String, String> extraMetadata = new HashMap<>(writeContext.getExtraMetaData());
+    final Map<String, String> extraMetadata;
+    if (encodingProps.getExtraMetaData() == null
+        || encodingProps.getExtraMetaData().isEmpty()) {
+      extraMetadata = writeContext.getExtraMetaData();
+    } else {
+      extraMetadata = new HashMap<>(writeContext.getExtraMetaData());
 
-    conf.iterator().forEachRemaining(entry -> {
-      if (entry.getKey().startsWith(extraMetadataConfPrefix)) {
-        final String metadataKey = entry.getKey().replaceFirst(extraMetadataConfPrefix, "");
-
+      encodingProps.getExtraMetaData().forEach((metadataKey, metadataValue) -> {
         if (metadataKey.equals(OBJECT_MODEL_NAME_PROP)) {
           throw new IllegalArgumentException("Cannot overwrite metadata key " + OBJECT_MODEL_NAME_PROP
               + ". Please use another key name.");
         }
 
-        if (extraMetadata.put(metadataKey, entry.getValue()) != null) {
-          throw new IllegalArgumentException("Extra metadata key " + metadataKey
-              + " conflicts with reserved metadata keys present in "
-              + writeSupport.getClass().getName() + ". Please use another key name.");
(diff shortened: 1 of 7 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -403,15 +405,29 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
 
     this.codecFactory = codecFactory;
     CompressionCodecFactory.BytesInputCompressor compressor = codecFactory.getCompressor(compressionCodecName);
+
+    final String extraMetadataConfPrefix = ParquetOutputFormat.EXTRA_WRITE_METADATA + ".";
+    final Map<String, String> extraMetadata = new HashMap<>(writeContext.getExtraMetaData());
+
+    conf.iterator().forEachRemaining(entry -> {
```
wgtmac, line 412: To support parquet writer without hadoop dependency, it would be good to implement this logic to `ParquetConfiguration` in two more overload constructors as well: https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L341-L379

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -403,15 +405,29 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
...
+      if (entry.getKey().startsWith(extraMetadataConfPrefix)) {
```
wgtmac, line 413: The caveat of this approach is that it wastes the entire loop if no extra metadata is provided, which is a common case.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
clairemcginty: 👍 

Do you think it still makes sense to implement this functionality both as a new `ParquetWriter` constructor arg, *and* as a Configuration option that gets parsed during construction? Like if we add this new constructor overload:

```diff
ParquetWriter(
      OutputFile file,
      ParquetFileWriter.Mode mode,
      WriteSupport<T> writeSupport,
      CompressionCodecName compressionCodecName,
      long rowGroupSize,
      boolean validating,
      ParquetConfiguration conf,
      int maxPaddingSize,
      ParquetProperties encodingProps,
      FileEncryptionProperties encryptionProperties,
+     Map<String, String> extraMetadata)
```
It might be redundant/confusing to also iterate over `conf` for all keys starting with `parquet.write.metadata.extra`. Maybe we should just add it as a constructor arg and drop the Configuration key/parsing altogether. Wdyt?

wgtmac: TBH, I don't think it is a good idea to add an overload of constructor (as we already have a bunch of constructors). What about adding a new method to ParquetWriter to set extra metadata? I have done similar things to parquet cpp: https://github.com/apache/arrow/commit/3bd57e35cc949bce6a04d7fbfed30c6db755b023

clairemcginty: yeah, that might be simplest! I could also add it to [ParquetProperties](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java), since that's already included in the default `ParquetWriter` constructor.

clairemcginty: I think the `ParquetProperties` approach might be cleanest -- I prototyped adding an additional `ParquetWriter#addKeyValueMetadata` method (as in your Arrow PR), but the problem is that we'd have to re-initialize `ParquetWriter::writer` every time it's invoked, since it's constructed with a metadata argument: https://github.com/apache/parquet-mr/blob/238b9a2/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L410

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/ExampleParquetWriter.java
```diff
@@ -113,11 +110,6 @@ public Builder withType(MessageType type) {
       return this;
     }
 
-    public Builder withExtraMetaData(Map<String, String> extraMetaData) {
```
clairemcginty, line 116: This method is now available through the `ParquetWriter.Builder` superclass.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
clairemcginty: Re-implemented it via `ParquetProperties` -- lmk what you think :)

wgtmac:
@Fokko @ConeyLiu Do you have any comment?

wgtmac:
@clairemcginty It seems that we need to add an exclusion to japicmp to make the CI happy.

clairemcginty:
> @clairemcginty It seems that we need to add an exclusion to japicmp to make the CI happy.

Added! The CI on [my fork](https://github.com/clairemcginty/parquet-mr/actions/runs/7506559099)'s GHA seems to be happy now

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -403,15 +405,29 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
...
+
+    final Map<String, String> extraMetadata;
+    if (encodingProps.getExtraMetaData() == null
+        || encodingProps.getExtraMetaData().isEmpty()) {
+      extraMetadata = writeContext.getExtraMetaData();
+    } else {
+      extraMetadata = new HashMap<>(writeContext.getExtraMetaData());
+
+      encodingProps.getExtraMetaData().forEach((metadataKey, metadataValue) -> {
+        if (metadataKey.equals(OBJECT_MODEL_NAME_PROP)) {
```
wgtmac, line 417: Can we avoid specializing any key? IIUC, it can also be caught at line 422 if OBJECT_MODEL_NAME_PROP has been set already.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/ExampleParquetWriter.java
wgtmac: Is it possible to add an overload here to suppress the complaint of japicmp? That's more preferable than an exclusion.

wgtmac (commented):
Thanks for the change! I've left some minor comments.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
clairemcginty: This key is kind of a special case since it's added by the delegated `InternalParqueRecordWriter` at the end of writing: https://github.com/apache/parquet-mr/blob/945836c/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java#L132-L134

So if we remove this extra check and there's a conflicting `OBJECT_MODEL_NAME_PROP` key in `extraMetaData`, `InternalParquetRecordWriter` will silently overwrite it at the end.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/example/ExampleParquetWriter.java
clairemcginty: great idea! updated + removed the exclusion.

wgtmac (approved):
+1

Thanks @clairemcginty and @ConeyLiu!

PR #1245, opened 2024-01-03
PARQUET-2060: Fix infinite loop while reading corrupt files

rathinb-db:
When a parquet file is corrupted, `decompress` can return 0 bytes. This can cause an infinite loop in the JDK `readFully` that calls the NonBlockedDecompressorStream `read`. Input streams are never supposed to return 0, but we do if there's a corrupt file. 

By throwing an error, we break early and exit fast.

### Jira

    - https://issues.apache.org/jira/browse/PARQUET-2060

### Tests

Add existing unit test. Manually tested with a corrupt file.

### Commits

### Style

### Documentation


ConeyLiu:
Could you add a UT for this?

rathinb-db:
> Could you add a UT for this?

cc: @ConeyLiu 
I've added a unit test. Please take a look and suggest any changes if any.

wgtmac:
Thanks for the fix! Could you please make the CI happy?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/codec/TestNonBlockedDecompressorStream.java
```diff
@@ -0,0 +1,54 @@
...
+          .thenReturn(0);
+
+      // Create a NonBlockedDecompressorStream with the mock Decompressor.
+      NonBlockedDecompressorStream decompressorStream =
+          new NonBlockedDecompressorStream(new ByteArrayInputStream(new byte[0]), mockDecompressor, 1024);
+
+      // Attempt to read from the stream, which should trigger the IOException.
+      decompressorStream.read(new byte[1024], 0, 1024);
+
+      // If no exception is thrown, then the test fails.
```
ConeyLiu, line 47: This can be simplified with `Assert.assertThrows`

wgtmac (changes requested):
Rest LGTM

rathinb-db:
Thanks for the inputs! I'd like to replace the mock with an actual corrupt file that can reliable reproduce this. Are there any simple ways to call `NonBlockedDecompressorStream`?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/codec/TestNonBlockedDecompressorStream.java
rathinb-db: Done

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/codec/TestCompressionCodec.java
```diff
@@ -18,9 +18,10 @@
  */
 package org.apache.parquet.hadoop.codec;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
wgtmac, line 21: Please avoid import *.

wgtmac (commented):
Rest LGTM. Thanks! @rathinb-db

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/codec/TestCompressionCodec.java
rathinb-db: Done

ConeyLiu (approved):
+1

PR #1253, opened 2024-01-14
PARQUET-2419: Reduce noisy logging when running test suite

amousavigourabi:

### Jira

    - https://issues.apache.org/jira/browse/PARQUET-2419

### Tests

- Only introduces changes to CI/CD workflows

### Commits

### Style

### Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,8 @@
     <surefire.argLine> </surefire.argLine>
     <surefire.logLevel>INFO</surefire.logLevel>
 
+    <ci-test-profile>true</ci-test-profile>
+
     <!-- Resource intesive tests are enabled by default but disabled in the CI envrionment -->
     <enableResourceIntensiveTests>true</enableResourceIntensiveTests>
 
@@ -716,8 +718,13 @@
     <!-- Profile for CI tests to have less output -->
     <profile>
       <id>ci-test</id>
+      <activation>
+        <property>
+          <name>ci-test-profile</name>
+        </property>
+      </activation>
       <properties>
-        <surefire.logLevel>WARN</surefire.logLevel>
+        <surefire.logLevel>ERROR</surefire.logLevel>
         <surefire.argLine>-XX:MaxJavaStackTraceDepth=10</surefire.argLine>
         <enableResourceIntensiveTests>false</enableResourceIntensiveTests>
       </properties>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -716,8 +718,13 @@
     <!-- Profile for CI tests to have less output -->
     <profile>
       <id>ci-test</id>
+      <activation>
+        <property>
+          <name>ci-test-profile</name>
```
wgtmac, line 723: Why enable it by default? BTW, do the surefire.xx arguments actually work locally?

amousavigourabi: Generally, when you run `mvn test` you do not need the immense amount of low level logs provided now. It makes more sense for them to be opt-in. As for the other question, luckily these settings do work locally.

wgtmac:
cc @Fokko

pom.xml
```diff
@@ -713,13 +713,12 @@
       </build>
     </profile>
 
-    <!-- Profile for CI tests to have less output -->
+    <!-- Profile for tests to have more output -->
     <profile>
-      <id>ci-test</id>
+      <id>verbose-test</id>
```
wgtmac, line 718: What about keeping the `ci-test` profile to be verbose as is, and add a new compact one which is activated by default?

amousavigourabi: As of now our output is verbose by default, with the `ci-test` profile being more compact. I tried looking into enabling `ci-test` by default, but when a profile is `activeByDefault`, it gets deactivated whenever another profile becomes active. The problem then is that `parquet-mr` has a few profiles that get activated automatically depending on the environment Maven is run in, so we cannot rely on it in that way. Because of this, the PR gets rid of the slimmed down `ci-test` profile and replaces it with a verbose one, such that the (currently default) verbose output becomes opt-in, with the more lean output just being the default setting.

wgtmac: Sounds good to me. 

@gszadovszky @shangxinli Do you have different opinions?

gszadovszky (approved):

pom.xml
Fokko: I think `verbose-test` is a better name than `ci-test`

Fokko (approved):
Lookg good, thanks @amousavigourabi for working on this 👍

PR #1273, opened 2024-02-13
[WIP][Proposal] PARQUET-2430: Add parquet joiner

MaxNevermind:
**_This PR is a proposal and Work In Progress._**

## Overview
ParquetJoiner feature is similar to ParquetRewrite class. ParquetRewrite allows to stitch files with the same schema into a single file while ParquetJoiner should enable stitching files with different schemas into a single file. That is possible when: 1) the number of rows in the main and extra files is the same, 2) the ordering of rows in the main and extra files is the same. Main benefit of ParquetJoiner is performance, for the cases when you join/stitch Terabytes/Petabytes of data that seemingly simple low level API can be up to 10x more resource efficient. 

## Implementation details
ParquetJoiner allows to specify the main input parquet file and extra input parquet files. ParquetJoiner will copy the main input as binary data and write extra input files with row groups adjusted to the main input. If main input is much larger than extra inputs then a lot of resources will be saved by working with the main input as binary.

## Use-case examples
A very large Parquet based dataset(dozens or hundreds of fields/Terabytes of data daily/Petabytes of historical partitions). The task is to modify a column or add a new column to it for all the historic data. It is trivial using Spark, but taking into consideration the share scale of a dataset it will take a lot of resources to do that. 

### Side notes
Note that this class of problems could be in theory solved by storing main input and extra inputs in HMS/Iceberg bucketed tables and use a view that joins those tables on the fly into the final version but in practice there is often a requirement to merge parquet files and have a single parquet sources in the file system.

## Use-case implementation details using Apache Spark
You can use Apache Spark to perform the join with ParquetJoiner, read the large main input and prepare the right side of a join in a way that each file on the left have a corresponding file on the right and it preserves records ordering on the right side in the same order as on the left side, that allows the whole input on the left and right to have the same number of files and the same number of records in corresponding files and the same ordering of records in each file pair. Then run ParquetJoiner in parallel for each file pair and perform a join. Example of the code that utilizes this new feature: https://gist.github.com/MaxNevermind/0feaaf380520ca34c2637027ef349a7d.

 

### Jira

- [x] My PR addresses the following [PARQUET-2430: Add ParquetJoiner feature](https://issues.apache.org/jira/browse/PARQUET-2430) 

### Tests

- [x] My PR adds the following unit tests:
  -   testStitchThreeInputsDifferentRowGroupSize

### Commits

### Style

### Documentation


wgtmac:
cc @gszadovszky @ggershinsky @ConeyLiu in case you've missed the ML discussion: https://lists.apache.org/thread/5q7qhkl07of4ypc4t5zx3hyjgv12c7q2

gszadovszky:
I don't have the time to properly review this proposal but I like the idea.
I agree we shall move this joiner approach as close to the rewriter logic as possible. Maybe introducing new interfaces/implementations and use them where the logic is really different.

ConeyLiu:
The idea looks good to me too, actually, we have similar requirements to update history data. Feel free to ping me when this is ready for review.

MaxNevermind:
@wgtmac 
Can I ask do a preliminary review of the code?
So far I've added a single test for a new feature. Wanted to add more tests and polish if the code looks good so far. All tests pass for both old and new code in ParquetRewriterTest atm. Will add more documentation to RewriteOptions if everything else looks good. Just fyi I've added an example of this new feature usage in a Gist here https://gist.github.com/MaxNevermind/0feaaf380520ca34c2637027ef349a7d.

wgtmac:
@MaxNevermind Sure, I will take a look this week.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -39,11 +36,20 @@
...
+ * different schemas.
+ * You can provide the main input file group and multiple right side ones. That is possible when:
+ * 1) the number of rows in the main and extra input groups are the same,
+ * 2) the ordering of rows in the main and extra input groups is the same.
  */
 public class RewriteOptions {
 
   private final ParquetConfiguration conf;
   private final List<InputFile> inputFiles;
+  private final List<List<InputFile>> inputFilesR;
```
wgtmac, line 52: Could we consolidate the internal member variables `inputFiles` and `inputFilesR`. My rough idea is to model these files into `HorizontalInputFile` and `VerticalInputFile`.
- VerticalInputFile: An `InputFile` of a single parquet file which may have partial or full schema.
- HorizontalInputFile: one or more `VerticalInputFile`s with same number rows but different schemas to form a logical wide file. It can choose one `VerticalInputFile` to be the left side and the others to be the right side in your join algorithm. If it contains only one `VerticalInputFile`, this is what we have today.

WDYT?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +108,49 @@ public class ParquetRewriter implements Closeable {
...
     ParquetConfiguration conf = options.getParquetConfiguration();
     OutputFile out = options.getParquetOutputFile();
-    openInputFiles(options.getParquetInputFiles(), conf);
+    inputFiles.addAll(getFileReaders(options.getParquetInputFiles(), conf));
+    List<Queue<TransParquetFileReader>> inputFilesR = options.getParquetInputFilesR()
+        .stream()
+        .map(x -> getFileReaders(x, conf))
+        .collect(Collectors.toList());
+    ensureSameSchema(inputFiles);
+    inputFilesR.forEach(this::ensureSameSchema);
```
wgtmac, line 123: So this imposes strict schema check here. Is it possible to loose the restriction to allow the files on the right side to have different schemas but in total all columns are available? For example: a group of right side files with schema below:
- `a:bigint,b:string`, `c:double,d:binary`
- `a:bigint`, `b:string`, `c:double`, `d:binary`
- `a:bigint,b:string,c:double,d:binary`

I know this would be complex, just asking and not sure if it is an valid real world use case.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +108,49 @@ public class ParquetRewriter implements Closeable {
...
     LOG.info("Start rewriting {} input file(s) {} to {}", inputFiles.size(), options.getParquetInputFiles(), out);
 
-    // Init reader of the first input file
-    initNextReader();
+    extraMetaData.put(
```
wgtmac, line 126: I'm skeptical of the importance of metadata from the right side files. However I don't have a better suggestion here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +108,49 @@ public class ParquetRewriter implements Closeable {
...
+            .map(x -> x.getFooter().getFileMetaData().getCreatedBy())
+            .collect(Collectors.toSet())
+            .stream()
+            .reduce((a, b) -> a + "\n" + b)
+            .orElse("")
+    );
+    Stream.concat(inputFiles.stream(), inputFilesR.stream().flatMap(Collection::stream))
+        .forEach(x -> extraMetaData.putAll(x.getFileMetaData().getKeyValueMetaData()));
+
+    // TODO check that schema on the left and on the right is not identical
```
wgtmac, line 138: nit: the schema check now is complex so it worths a separate method.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -137,16 +160,31 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
...
       schema = pruneColumnsInSchema(schema, prunePaths);
     }
 
-    this.descriptorsMap =
-        schema.getColumns().stream().collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    if (inputFilesR.isEmpty()) { // TODO: find a more suitable solution
+      this.descriptorsMap =
+          schema.getColumns().stream().collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    } else {
+      this.descriptorsMap = schemaL.getColumns().stream()
```
wgtmac, line 175: OK, now I can answer my own question, the complex of loose schema enforcement on the right side may complicate `descriptorsMap` here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -137,16 +160,31 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
...
+        .filter(x -> x.getPath().length == 0 || !fieldNamesR.containsKey(x.getPath()[0]))
+          .collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    }
+
+    long rowCountL = inputFiles.stream().mapToLong(ParquetFileReader::getRecordCount).sum();
+    inputFilesR.stream()
+        .map(x -> x.stream().mapToLong(ParquetFileReader::getRecordCount).sum())
+        .forEach(rowCountR -> {
+          if (rowCountL != rowCountR) {
+            throw new IllegalArgumentException("The number of records on the left and on the right don't match!");
```
wgtmac, line 185: nit: print the row count in the error msg.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -813,7 +841,7 @@ private Type extractField(GroupType candidate, Type targetField) {
       } else {
         Type tempField = extractField(field.asGroupType(), targetField);
         if (tempField != null) {
-          return tempField;
+          return new GroupType(candidate.getRepetition(), candidate.getName(), tempField);
```
wgtmac, line 844: Is it an individual bug fix?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +932,208 @@ public byte[] getDictPageAAD() {
...
+          reader.setStreamPosition(chunk.getStartingPos());
+          BloomFilter bloomFilter = indexCache.getBloomFilter(chunk);
+          ColumnIndex columnIndex = indexCache.getColumnIndex(chunk);
+          OffsetIndex offsetIndex = indexCache.getOffsetIndex(chunk);
+          writer.appendColumnChunk(descriptorL, reader.getStream(), chunk, bloomFilter, columnIndex, offsetIndex);
+        }
+      }
+
+      // Writing the right side
+      for (RightColumnWriter writer: columnWritersR) {
```
wgtmac, line 962: Is it possible to move this into `processBlocksFromReader`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: In theory that is possible but that would make RewriteOptions for the right side more complex which is already not simple, it is List<List<InputFile>> right now, It would need to become List<List<List<InputFile>>> to allow something like that. Also while writing the right side the sate would need to become more complex too so it can track all of those files.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: yeah, that's the issue I mentioned in email communication, it become a problem when I started a using a different for the right side file group

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -110,6 +118,27 @@ public List<Path> getInputFiles() {
         .collect(Collectors.toList());
   }
 
+  /** TODO fix documentation after addition of inputFilesR
+   * Gets the right input {@link Path}s for the rewrite if they exist for all input files,
+   * otherwise throws a {@link RuntimeException}.
+   *
+   * @return a {@link List} of the associated right input {@link Path}s
+   */
+  public List<List<Path>> getInputFilesR() {
```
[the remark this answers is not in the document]
MaxNevermind: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: On the second thought actually the right side might remain List<List<InputFile>>. The requirements are: the order of files for each column need to be expressed somehow to us and number of rows in for each column in total in all files must be the same as on the left side. Having those satisfied we can deduct which file has which column etc. Let me think a little more about it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: Yeah, that would make sense, let me try.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: So the RewriteOptions will contain a holder for List<VerticalInputFile> & List<HorizontalInputFile> or some sort of general List<StackableInputFile> that both VerticalInputFile & HorizontalInputFile need to implement? 


By the way. In my current implementation only the total number of rows of the left and on the right need to match. Individual files don't need the row count matched. It seem you suggest putting the same number of records into each VerticalInputFiles group. I'm not sure that particular requirement is easy to satisfy in practical environment. For example in Spark there are ways to preserve the ordering as I showed in the linked Gist in PR description but controlling the number of rows in each output file in Spark is not trivial as far as I know.

wgtmac: That make sense. Files on the right side do not need to make number of rows to the left side.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: I thought a little more about your suggestion. My notes:
- We can do that but we should not require the same number of rows for VerticalInputFiles in each HorizontalInputFile. In my opinion tools that produce Parquet files often use some threshold size in bytes(e.g.:HDFS block size) as a measure of next file spit. So it will be changeling for user to produce such files with matching row count.
- If we do that the state that tracks current file for each column on the right side will become more complex. Right now the state is easily maintained by a input file queue for each file group.
- I can see that we definitely will provide a more use-case coverage that way but I'm not sure if it will be useful. In my experience parquet files are usually produced using some widespread libraries/framework and those tools usually produce parquet dataset/tables which implies all output files have the same schema. I can see that somebody can produce such parquet files as you described programmatically, but in such cases those files unlikely to be large and the main use-case of my feature is in the context of optimization for very large datasets stitching, if there are no large datasets involved it might be easier to just completely rewrite files using full read & write.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

MaxNevermind:
@wgtmac 
Let me if you want me to implement your VerticalInputFile / HorizontalInputFile idea and what to do you see has left in this PR. 
Let me know what else I should do, I think I should add javadoc I can add more tests. I guess you are aware but just fyi ParquetRewriterTest already takes 3/4 of the total parquet-hadoop testing time, on my machine it is 16 min and 22 min correspondingly.

wgtmac:
Thanks for the update! I will take a look later this week.

> Let me if you want me to implement your VerticalInputFile / HorizontalInputFile idea and what to do you see has left in this PR.

No. The rows on both sides are different, so we don't have to bother with the concepts.

> I guess you are aware but just fyi ParquetRewriterTest already takes 3/4 of the total parquet-hadoop testing time, on my machine it is 16 min and 22 min correspondingly.

Yes, we probably should disable some parameter combinations.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +117,55 @@ public class ParquetRewriter implements Closeable {
...
+    schemaR.stream().flatMap(x -> x.getFields().stream()).forEach(x -> {
+      if (fieldNamesR.containsKey(x.getName())) {
+        throw new IllegalArgumentException(
+            "Found a duplicated field `" + x.getName() + "` in the right side file groups!");
+      }
+      fieldNamesR.put(x.getName(), x);
+    });
+    List<Type> fields = Stream.concat(
+            fieldNamesL.values().stream()
+                .map(x -> fieldNamesR.getOrDefault(
```
wgtmac, line 161: Why do we need to try to get it from fieldNamesR?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -137,16 +175,34 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
...
     }
 
-    this.descriptorsMap =
-        schema.getColumns().stream().collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    if (inputFilesR.isEmpty()) {
+      this.descriptorsMap =
+          schema.getColumns().stream().collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    } else { // TODO: describe in documentation that only top level column can be overwritten
+      this.descriptorsMap = schemaL.getColumns().stream()
+          .filter(x -> x.getPath().length == 0 || !fieldNamesR.containsKey(x.getPath()[0]))
```
wgtmac, line 191: If we restrict both left and right sides do not have duplicate fields, would we make things simpler here and after?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+              chunk.getEncodingStats().usesV2Pages()
+                  ? ParquetProperties.WriterVersion.PARQUET_2_0
+                  : ParquetProperties.WriterVersion.PARQUET_1_0;
+          ParquetProperties props = ParquetProperties.builder()
+              .withWriterVersion(writerVersion)
+              .withBloomFilterEnabled(bloomFilterLength > 0)
+              .build();
+          CodecFactory codecFactory = new CodecFactory(new Configuration(), props.getPageSizeThreshold());
+          CompressionCodecFactory.BytesInputCompressor compressor =
+              codecFactory.getCompressor(chunk.getCodec());
```
wgtmac, line 1053: This uses the same codec in the input file. Should we respect the trans-codec rewriter options?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -137,16 +175,34 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
       getPaths(schema, paths, null);
       for (String col : pruneColumns) {
         if (!paths.contains(col)) {
-          LOG.warn("Input column name {} doesn't show up in the schema of file {}", col, reader.getFile());
+          LOG.warn("Input column name {} doesn't show up in the schema", col);
         }
       }
 
       Set<ColumnPath> prunePaths = convertToColumnPaths(pruneColumns);
       schema = pruneColumnsInSchema(schema, prunePaths);
```
wgtmac, line 183: Right side columns get pruned here, but they are still written in the RightColumnWriter. Do you plan to enable pruning columns on the right side? For whatever reason, we need to state this in the docstring of RewriterOptions.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+      ColumnWriter cWriter = cWriters.get(descriptor);
+      int dMax = descriptor.getMaxDefinitionLevel();
+      Class<?> columnType = descriptor.getPrimitiveType().getPrimitiveTypeName().javaType;
+      ColumnReader reader = colReaders.get(descriptor);
+      for (int i = 0; i < rowsToWrite; i++) {
+        int rlvl = reader.getCurrentRepetitionLevel();
+        int dlvl = reader.getCurrentDefinitionLevel();
+        do {
+          if (dlvl < dMax) {
+            cWriter.writeNull(rlvl, dlvl);
```
wgtmac, line 1098: If a column to mask (only nullifying column is supported) is on the right side, we also need to handle it here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+              codecFactory.getCompressor(chunk.getCodec());
+
+          MessageType columnSchema = parquetRewriter.newSchema(schema, descriptor);
+          ColumnChunkPageWriteStore cPageStore = new ColumnChunkPageWriteStore(
+              compressor,
+              columnSchema,
+              props.getAllocator(),
+              props.getColumnIndexTruncateLength(),
+              props.getPageWriteChecksumEnabled(),
+              writer.getEncryptor(),
```
wgtmac, line 1062: IIUC, this line makes sure that right side column can be encrypted.

wgtmac (changes requested):
I do see that you've taken a lot of effort to consolidate the implementation with original rewriter features. Thank you for doing that! I've left some comments and I think we have made good progress.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -119,6 +130,15 @@ public List<InputFile> getParquetInputFiles() {
     return inputFiles;
   }
 
+  /** TODO fix documentation after addition of inputFilesR
+   * Gets the right {@link InputFile}s for the rewrite.
+   *
+   * @return a {@link List} of the associated right {@link InputFile}s
+   */
+  public List<List<InputFile>> getParquetInputFilesR() {
```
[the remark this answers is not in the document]
MaxNevermind: Any preference?
- InputFilesRight
- InputFilesExtra
- InputFilesToJoin
- InputFilesStitched

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+
+    private void flushWriters() throws IOException {
+      cStores.values().forEach(cStore -> {
+        cStore.flush();
+        cStore.close();
+      });
+      cWriters.values().forEach(ColumnWriter::close);
+      for (ColumnDescriptor descriptor : descriptorsMap.values()) {
+        if (cPageStores.containsKey(descriptor))
+          cPageStores.get(descriptor).flushToFileWriter(writer);
```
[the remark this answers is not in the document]
MaxNevermind: > we may have some issues here since the column flush order does not match the schema order

I didn't know that this might cause problems. I though that schema column order and actual physical order are decoupled, I made my conclusion based on [Parquet File Metadata Schema](https://github.com/apache/parquet-format#metadata) as each column chunk stores its metadata internally: column path and file offset, so it seems there is no need for them to be coupled. Also that is weird that my new test didn't fail as I actually overwrite a column `FloatFraction` which is in the middle of the left schema, and as you can see from `processBlocksFromReader` method right side is always written at the end after all the left side is written. I need to investigate more.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+          ColumnDescriptor descriptor = descriptorsMap.get(chunk.getPath());
+          copyValues(descriptor, writeFromBlock);
+        }
+        rowsToWrite -= writeFromBlock;
+        writtenFromBlock += writeFromBlock;
+        if (rowsToWrite > 0 || (block.getRowCount() == writtenFromBlock)) {
+          rowGroupIdxOut++;
+          if (rowGroupIdxOut == blocks.size()) {
+            inputFiles.poll();
+            rowGroupIdxOut = 0;
```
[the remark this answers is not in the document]
MaxNevermind: You mean `rowGroupIdxOut++`? Yes, it is required. I can rewrite the code like this if it is easier to read:
```
        if (rowsToWrite > 0 || (block.getRowCount() == writtenFromBlock)) {
          if (rowGroupIdxR == blocks.size()) {
            inputFiles.poll();
            rowGroupIdxR = 0;
          } else {
            rowGroupIdxR++;
          }
          ...
        }
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+      this.writer = parquetRewriter.writer;
+      this.schema = inputFiles.peek().getFooter().getFileMetaData().getSchema();
+      this.descriptorsMap = this.schema.getColumns().stream()
+          .collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+      initReaders();
+      initWriters();
+    }
+
+    public void writeRows(int rowGroupIdx, long rowsToWrite) throws IOException {
+      if (rowGroupIdxIn != rowGroupIdx) {
```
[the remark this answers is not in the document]
MaxNevermind: We can but right now that is not supposed to happen and I believe it would brake the right side tests immediately and a problem become visible through that. I can add something like that if you want me to:
```
      if (rowGroupIdxL > rowGroupIdx) {
        throw new IOException("A row group index decrease is determined in RightColumnWriter! Current index: " 
            + rowGroupIdxL + ", new index: " + rowGroupIdx);
      }
```
Let me know

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +953,171 @@ public byte[] getDictPageAAD() {
...
+      this.parquetRewriter = parquetRewriter;
+      this.writer = parquetRewriter.writer;
+      this.schema = inputFiles.peek().getFooter().getFileMetaData().getSchema();
+      this.descriptorsMap = this.schema.getColumns().stream()
+          .collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+      initReaders();
+      initWriters();
+    }
+
+    public void writeRows(int rowGroupIdx, long rowsToWrite) throws IOException {
```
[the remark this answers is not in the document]
MaxNevermind: Done, renamed them to rowGroupIdxL & rowGroupIdxR, hope that improves readability 😄

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -347,8 +390,14 @@ private void processBlocksFromReader(IndexCache indexCache) throws IOException {
...
       }
 
+      // Writing extra columns
+      for (RightColumnWriter writer : columnWritersR) {
+        writer.writeRows(rowGroupIdx, blockMetaData.getRowCount());
+      }
+
       writer.endBlock();
       numBlocksRewritten++;
+      rowGroupIdx++;
```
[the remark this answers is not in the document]
MaxNevermind: You are right. Replaced rowGroupIdx with numBlocksRewritten. We still need to pass it to RightColumnWriter because it is a static class.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -347,8 +390,14 @@ private void processBlocksFromReader(IndexCache indexCache) throws IOException {
         columnId++;
       }
 
+      // Writing extra columns
+      for (RightColumnWriter writer : columnWritersR) {
+        writer.writeRows(rowGroupIdx, blockMetaData.getRowCount());
```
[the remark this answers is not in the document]
MaxNevermind: done

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: I wanted to preserve the original capabilities of ParquetRewriter which allowed duplicated columns as I understand, but for the right side that would create a problem and I added that check. Let me know if you want to support for duplicated columns in general.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: The idea behind is that I wanted to support the use-case when the left files have such a field but it has a different structure on the right.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: Is there an action point here or we can resolve this?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -60,10 +63,7 @@
 import org.apache.parquet.format.DictionaryPageHeader;
 import org.apache.parquet.format.PageHeader;
 import org.apache.parquet.format.converter.ParquetMetadataConverter;
-import org.apache.parquet.hadoop.CodecFactory;
-import org.apache.parquet.hadoop.ColumnChunkPageWriteStore;
-import org.apache.parquet.hadoop.IndexCache;
-import org.apache.parquet.hadoop.ParquetFileWriter;
+import org.apache.parquet.hadoop.*;
```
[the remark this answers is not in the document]
MaxNevermind: done

MaxNevermind:
@wgtmac 
Can you check out the comments I left please?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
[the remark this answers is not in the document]
wgtmac: What about `InputFilesToJoinColumns` or `InputFilesToAddColumns`. This is not a strong opinion. Just want to clarify what to do with these input files.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: I'm a little concerned about the possible explosion of the metadata. Should we add an option to ignore those on the right side? Or at least add a comment to raise the potential issue.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: This implicit behavior may introduce unexpected use. Does it sound better to add a resolve strategy for duplicate column names? By default we can pick it up from the right side.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: IIRC, the original rewriter does not support duplicated columns. However, supporting it while joining files is a good idea. Just as my previous comment, it would be better to support explicit resolve strategy for them if that is not too complicated.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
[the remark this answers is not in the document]
wgtmac: Yes, the specs do not require the column chunk order should be coupled to the schema order. However, I'm not sure if this will break some parquet implementations on the wild.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: That makes sense. I didn't mean we have to implement everything.

wgtmac (commented):
Sorry for the delay. Please see my new comments.

wgtmac:
@ConeyLiu Would you mind taking a look as well?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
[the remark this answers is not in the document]
wgtmac: I misunderstood the code, ignore it pls.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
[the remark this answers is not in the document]
wgtmac: Yes, that sounds good.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +121,55 @@ public class ParquetRewriter implements Closeable {
...
     ParquetConfiguration conf = options.getParquetConfiguration();
     OutputFile out = options.getParquetOutputFile();
-    openInputFiles(options.getParquetInputFiles(), conf);
+    inputFiles.addAll(getFileReaders(options.getParquetInputFiles(), conf));
+    List<Queue<TransParquetFileReader>> inputFilesR = options.getParquetInputFilesR().stream()
+        .map(x -> getFileReaders(x, conf))
+        .collect(Collectors.toList());
+    ensureSameSchema(inputFiles);
+    inputFilesR.forEach(this::ensureSameSchema);
     LOG.info("Start rewriting {} input file(s) {} to {}", inputFiles.size(), options.getParquetInputFiles(), out);
```
ConeyLiu, line 136: This log only contains left input files. It would be better to add the right files as well.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
ConeyLiu: So, the left part columns' order could be changed if overwrite occurred.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +955,175 @@ public byte[] getDictPageAAD() {
...
+    private final ParquetFileWriter writer;
+    private final MessageType schema;
+    private final Map<ColumnPath, ColumnDescriptor> descriptorsMap;
+    private final Map<ColumnDescriptor, ColumnReader> colReaders = new HashMap<>();
+    private final Map<ColumnDescriptor, ColumnChunkPageWriteStore> cPageStores = new HashMap<>();
+    private final Map<ColumnDescriptor, ColumnWriteStore> cStores = new HashMap<>();
+    private final Map<ColumnDescriptor, ColumnWriter> cWriters = new HashMap<>();
+    private int rowGroupIdxL = 0; // index of the rowGroup of the current file on the left
+    private int rowGroupIdxR = 0; // index of the rowGroup of the current file on the right
+    private int writtenFromBlock = 0;
```
ConeyLiu, line 971: I suggest using long type.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +955,175 @@ public byte[] getDictPageAAD() {
...
+        initWriters();
+      }
+      while (rowsToWrite > 0) {
+        List<BlockMetaData> blocks = inputFiles.peek().getFooter().getBlocks();
+        BlockMetaData block = blocks.get(rowGroupIdxR);
+        List<ColumnChunkMetaData> chunks = block.getColumns();
+        long leftInBlock = block.getRowCount() - writtenFromBlock;
+        long writeFromBlock = Math.min(rowsToWrite, leftInBlock);
+        for (ColumnChunkMetaData chunk : chunks) {
+          if (chunk.isEncrypted()) {
```
ConeyLiu, line 1002: Why not check this at `initWriters`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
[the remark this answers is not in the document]
ConeyLiu: +1, `InputFilesR` should be replaced with a better name.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +955,175 @@ public byte[] getDictPageAAD() {
...
+              .findFirst()
+              .orElseThrow(() -> new IllegalStateException(
+                  "Could not find column [" + columnPath.toDotString() + "]."));
+          int bloomFilterLength = chunk.getBloomFilterLength();
+          ParquetProperties.WriterVersion writerVersion =
+              chunk.getEncodingStats().usesV2Pages()
+                  ? ParquetProperties.WriterVersion.PARQUET_2_0
+                  : ParquetProperties.WriterVersion.PARQUET_1_0;
+          ParquetProperties props = ParquetProperties.builder()
+              .withWriterVersion(writerVersion)
```
ConeyLiu, line 1054: The output file could be mixed v1/v2 format if the part of the right files has mixed format. That's right?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -904,4 +955,175 @@ public byte[] getDictPageAAD() {
...
+              .withBloomFilterEnabled(bloomFilterLength > 0)
```
ConeyLiu, line 1055: Does the bloom filter have the same mixed problem?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -740,6 +858,30 @@ private MessageType createSchema() {
...
+        "schema",
+        new GroupType(
+            OPTIONAL,
+            "Links",
+            new PrimitiveType(REPEATED, BINARY, "Backward"),
+            new PrimitiveType(REPEATED, BINARY, "Forward")));
+  }
+
+  private MessageType createSchemaR2() {
+    return new MessageType("schema", new PrimitiveType(REPEATED, FLOAT, "FloatFraction"));
```
ConeyLiu, line 882: Could you add an overwrite column with a different type?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -137,16 +179,34 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
...
+          .collect(Collectors.toMap(x -> ColumnPath.get(x.getPath()), x -> x));
+    }
+
+    long rowCountL =
+        inputFiles.stream().mapToLong(ParquetFileReader::getRecordCount).sum();
+    inputFilesR.stream()
+        .map(x ->
+            x.stream().mapToLong(ParquetFileReader::getRecordCount).sum())
+        .forEach(rowCountR -> {
+          if (rowCountL != rowCountR) {
```
ConeyLiu, line 205: We only check the row count. How do we check the record order? That should be a key requirement as well.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -115,19 +121,55 @@ public class ParquetRewriter implements Closeable {
...
+    List<Type> fields = Stream.concat(
+            fieldNamesL.values().stream()
+                .map(x -> fieldNamesR.getOrDefault(
+                    x.getName(), x)), // take a field on the right if we can
+            fieldNamesR.values().stream()
+                .filter(x -> !fieldNamesL.containsKey(
+                    x.getName())) // takes fields on the right if it was not present on the left
+            )
+        .collect(Collectors.toList());
+    schema = new MessageType(schemaL.getName(), fields);
```
ConeyLiu, line 172: The schema does not match the actual column orders if there is any overwrite. We need to add ut to verify this.

ConeyLiu:
Thanks @MaxNevermind for the great work and thanks @wgtmac for pinging me. I am sorry for the late reply.

Aleis007891 (changes requested):
Yy

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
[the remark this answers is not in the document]
MaxNevermind: renamed to InputFilesToJoinColumns

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: I thought about it and maybe we should make it as a general option for both left and right? In theory it might happen even without the right part as I understand. Let me know. If no, I will add an option for just the right part.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: > resolve strategy for duplicate

Can you please clarify what do you mean by that? You mean if the schema of the column on the left is "compatible" with a new column schema on the right side then we try to "merge" the columns?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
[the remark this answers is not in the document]
MaxNevermind: Should I add a proposed check in a message above or we can resolve this as is?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: No, I mean we should be able to choose how to resolve columns with duplicate names, probable strategies are: pick left, pick right, fail.

Aleis007891 (approved):

Aleis007891 (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: @wgtmac you also mentioned that same problem [in this comment](https://github.com/apache/parquet-mr/pull/1273/files#r1547172865).
I've just spend some time on trying to refactor this part. I tried to align schema column order and actual write order of columns and it is mostly doable but RightColumnWriter state becomes too complex, for example file queue need to be present for each individual column on the right. I think the code will become to difficult to support long term if we do that.

MaxNevermind:
@wgtmac @ConeyLiu 

fyi
Wanted to share an idea: we can use binary copy on the right side, do not do read-write for right columns, that is possible if the number of rows and ordering in row groups on the left and on the right is the same. I believe a related ideas were discussed in this PR. I originally doubt possibility of user being able to produce files in such a fashion, but this week I’ve found a way to do that in Spark 3.3+. I can create a utility class([similar to the one from above](https://gist.github.com/MaxNevermind/0feaaf380520ca34c2637027ef349a7d)) that takes the input files, run transformations for them, and then write the results in a way that it preserves original input files’ names, row groups row count and ordering. The benefit are of course the speed because of binary copy and simplicity as we don't need RightColumnWriter. I will try to validate my idea the next week and get back with the result. If I’m able to do that I suggest to simplify this PR or maybe split the effort into two: one simpler version for simple binary copy and the second one based on current state of this PR.

Aleis007891 (approved):

Aleis007891 (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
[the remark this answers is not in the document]
pyckle: I encountered this same question while writing a parquet driver (https://github.com/Earnix/parquetforge). So I'll share what I dug up.
The parquet-format spec requires the ColumnChunk order in the footer to match the schema order. That's documented here: https://github.com/apache/parquet-format/blob/1dbc814b97c9307687a2e4bee55545ab6a2ef106/src/main/thrift/parquet.thrift#L1002
In the JIRA ticket linked to the commit message (now migrated to GH) - https://github.com/apache/parquet-java/issues/1734 it seems that at least impala accepts columns written in arbitrary orders, and ordering the Dict/DataPages of a column chunk is not required.
With this said, be aware that this can adversely affect performance. See: https://dl.acm.org/doi/10.1145/3035918.3035930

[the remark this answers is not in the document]
wgtmac: IIUC, the order of ColumnChunk metadata should match the schema order. However, their offsets in the Parquet file are not required to be ordered.

PR #1280, opened 2024-02-23
PARQUET-2436: More optimal memory usage in compression codecs

gszadovszky:
Rewritten direct codec implementations to have common memory handling

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2436) issues and references

### Tests

### Commits

### Style

### Documentation


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DirectCodecFactory.java
```diff
@@ -250,75 +341,86 @@ public BytesInput decompress(BytesInput bytes, int uncompressedSize) throws IOEx
...
 
-      outgoing.limit(size);
+    @Override
+    void closeDecompressor() {
+      context.close();
+    }
+  }
+
+  private class ZstdCompressor extends BaseCompressor {
+    private final ZstdCompressCtx context;
```
wgtmac, line 398: I'm not familiar with `ZstdCompressCtx`. Does it provide the same performance as before? IIUC, original impl leverages zstd workers but I didn't see it from the new class.

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DirectZstd.java#L91

gszadovszky: There is a big difference between our "usual" codecs that are using the Hadoop codecs and the ones created by the DirectCodecFactory. The first one has a stream interface behind our interfaces.
At the first attempt of implementing the direct Zstd codec, I was using the stream implementation for the BytesInput and a static method of the Zstd API for the ByteBuffer interface.
Now, I realized that when we use a direct ByteBufferAllocator the BytesInput objects are also mainly represented by direct ByteBuffers. So we can use them without a stream.
The context is used by the static API of the Zstd codec anyway. Based on the class comments it is suggested to use the context instead.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DirectCodecFactory.java
```diff
@@ -186,6 +174,100 @@ public void release() {
     }
   }
 
+  private abstract class BaseDecompressor extends BytesDecompressor {
```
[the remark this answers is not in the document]
gszadovszky: Unfortunately, the CodecFactory's interface still produces BytesInputDecompress/Compressor so I cannot change it here without rewriting the API causing breaking changes.

wgtmac (approved):
LGTM

gszadovszky:
Thanks for the review, @wgtmac!

PR #1291, opened 2024-03-05
PARQUET-2451: Add BYTE_STREAM_SPLIT support for FIXED_LEN_BYTE_ARRAY, INT32 and INT64

pitrou:
Implement the format additions described in PARQUET-2414.

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2451) issues and references
    - https://issues.apache.org/jira/browse/PARQUET-2451

### Tests

### Commits

### Style

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -50,6 +50,7 @@ public class ParquetProperties {
   public static final int DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
   public static final boolean DEFAULT_IS_DICTIONARY_ENABLED = true;
   public static final boolean DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED = false;
+  public static final boolean DEFAULT_IS_EXTENDED_BYTE_STREAM_SPLIT_ENABLED = false;
   public static final WriterVersion DEFAULT_WRITER_VERSION = WriterVersion.PARQUET_1_0;
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
@@ -115,6 +116,7 @@ public class ParquetProperties {
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final boolean enableByteStreamSplit;
+  private final boolean enableExtendedByteStreamSplit;
   private final Map<String, String> extraMetaData;
 
   private ParquetProperties(Builder builder) {
@@ -142,6 +144,7 @@ public class ParquetProperties {
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
     this.enableByteStreamSplit = builder.enableByteStreamSplit;
+    this.enableExtendedByteStreamSplit = builder.enableExtendedByteStreamSplit;
     this.extraMetaData = builder.extraMetaData;
   }
 
@@ -210,6 +213,10 @@ public class ParquetProperties {
 
   public boolean isByteStreamSplitEnabled() {
     return enableByteStreamSplit;
+  }
+
+  public boolean isExtendedByteStreamSplitEnabled() {
+    return enableExtendedByteStreamSplit;
   }
 
   public ByteBufferAllocator getAllocator() {
@@ -350,6 +357,7 @@ public class ParquetProperties {
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
     private boolean pageWriteChecksumEnabled = DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED;
     private boolean enableByteStreamSplit = DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED;
+    private boolean enableExtendedByteStreamSplit = DEFAULT_IS_EXTENDED_BYTE_STREAM_SPLIT_ENABLED;
     private Map<String, String> extraMetaData = new HashMap<>();
 
     private Builder() {
@@ -382,6 +390,7 @@ public class ParquetProperties {
       this.numBloomFilterCandidates = ColumnProperty.<Integer>builder(toCopy.numBloomFilterCandidates);
       this.maxBloomFilterBytes = toCopy.maxBloomFilterBytes;
       this.enableByteStreamSplit = toCopy.enableByteStreamSplit;
+      this.enableExtendedByteStreamSplit = toCopy.enableExtendedByteStreamSplit;
       this.extraMetaData = toCopy.extraMetaData;
     }
 
@@ -420,8 +429,27 @@ public class ParquetProperties {
       return this;
     }
 
-    public Builder withByteStreamSplitEncoding(boolean enableByteStreamSplit) {
-      this.enableByteStreamSplit = enableByteStreamSplit;
+    /**
+     * Enable or disable BYTE_STREAM_SPLIT encoding for FLOAT and DOUBLE columns.
+     *
+     * @param enable whether BYTE_STREAM_SPLIT encoding should be enabled
+     * @return this builder for method chaining.
+     */
+    public Builder withByteStreamSplitEncoding(boolean enable) {
+      this.enableByteStreamSplit = enable;
+      this.enableExtendedByteStreamSplit &= enable;
+      return this;
+    }
+
+    /**
+     * Enable or disable BYTE_STREAM_SPLIT encoding for FLOAT, DOUBLE, INT32, INT64 and FIXED_LEN_BYTE_ARRAY columns.
+     *
+     * @param enable whether BYTE_STREAM_SPLIT encoding should be enabled
+     * @return this builder for method chaining.
+     */
+    public Builder withExtendedByteStreamSplitEncoding(boolean enable) {
+      this.enableByteStreamSplit |= enable;
+      this.enableExtendedByteStreamSplit = enable;
       return this;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -50,7 +50,6 @@ public class ParquetProperties {
   public static final int DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
   public static final boolean DEFAULT_IS_DICTIONARY_ENABLED = true;
   public static final boolean DEFAULT_IS_BYTE_STREAM_SPLIT_ENABLED = false;
-  public static final boolean DEFAULT_IS_EXTENDED_BYTE_STREAM_SPLIT_ENABLED = false;
   public static final WriterVersion DEFAULT_WRITER_VERSION = WriterVersion.PARQUET_1_0;
   public static final boolean DEFAULT_ESTIMATE_ROW_COUNT_FOR_PAGE_SIZE_CHECK = true;
   public static final int DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
@@ -70,6 +69,12 @@ public class ParquetProperties {
   public static final ValuesWriterFactory DEFAULT_VALUES_WRITER_FACTORY = new DefaultValuesWriterFactory();
 
   private static final int MIN_SLAB_SIZE = 64;
+
+  private enum ByteStreamSplitMode {
+    NONE,
+    FLOATING_POINT,
+    EXTENDED
+  }
 
   public enum WriterVersion {
     PARQUET_1_0("v1"),
@@ -115,8 +120,7 @@ public class ParquetProperties {
   private final ColumnProperty<Integer> numBloomFilterCandidates;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
-  private final boolean enableByteStreamSplit;
-  private final boolean enableExtendedByteStreamSplit;
+  private final ColumnProperty<ByteStreamSplitMode> byteStreamSplitEnabled;
   private final Map<String, String> extraMetaData;
 
   private ParquetProperties(Builder builder) {
@@ -143,9 +147,16 @@ public class ParquetProperties {
     this.numBloomFilterCandidates = builder.numBloomFilterCandidates.build();
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
-    this.enableByteStreamSplit = builder.enableByteStreamSplit;
-    this.enableExtendedByteStreamSplit = builder.enableExtendedByteStreamSplit;
+    this.byteStreamSplitEnabled = builder.byteStreamSplitEnabled.build();
     this.extraMetaData = builder.extraMetaData;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static Builder copy(ParquetProperties toCopy) {
+    return new Builder(toCopy);
   }
 
   public ValuesWriter newRepetitionLevelWriter(ColumnDescriptor path) {
@@ -211,12 +222,23 @@ public class ParquetProperties {
     return dictionaryEnabled.getValue(column);
   }
 
+  @Deprecated()
   public boolean isByteStreamSplitEnabled() {
-    return enableByteStreamSplit;
+    return byteStreamSplitEnabled.getDefaultValue() != ByteStreamSplitMode.NONE;
   }
 
-  public boolean isExtendedByteStreamSplitEnabled() {
-    return enableExtendedByteStreamSplit;
+  public boolean isByteStreamSplitEnabled(ColumnDescriptor column) {
+    switch (column.getPrimitiveType().getPrimitiveTypeName()) {
+      case FLOAT:
+      case DOUBLE:
+        return byteStreamSplitEnabled.getValue(column) != ByteStreamSplitMode.NONE;
+      case INT32:
+      case INT64:
+      case FIXED_LEN_BYTE_ARRAY:
+        return byteStreamSplitEnabled.getValue(column) == ByteStreamSplitMode.EXTENDED;
+      default:
+        return false;
+    }
   }
 
   public ByteBufferAllocator getAllocator() {
@@ -308,14 +330,6 @@ public class ParquetProperties {
     return extraMetaData;
   }
 
-  public static Builder builder() {
-    return new Builder();
-  }
-
-  public static Builder copy(ParquetProperties toCopy) {
-    return new Builder(toCopy);
-  }
-
   @Override
   public String toString() {
     return "Parquet page size to " + getPageSizeThreshold() + '\n'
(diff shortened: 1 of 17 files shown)


pitrou:
@wgtmac @etseidl @gszadovszky  This is a draft PR. It works fine on the integration test but still needs unit tests. Since I'm quite new to this codebase (and to Java in general), I welcome any early comments.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInterOpReadByteStreamSplit.java
```diff
@@ -67,7 +72,73 @@ public void testReadFloats() throws IOException {
...
+  public void testReadAllSupportedTypes() throws IOException {
+    Path extendedFile = interop.GetInterOpFile(EXTENDED_FILE, CHANGESET);
+    final int expectRows = 200;
+    compareColumnValues(
+        extendedFile, expectRows, "float_plain", "float_byte_stream_split", ((group, leftCol, rightCol) -> {
+          float leftVal = group.getFloat(leftCol, 0);
+          float rightVal = group.getFloat(rightCol, 0);
+          assertEquals(leftVal, rightVal, 0.0);
+        }));
+    compareColumnValues(
```
pitrou, line 102: I'm curious if there's an easy way to reduce code repetition here. The various typed `get*` methods as well as the differing `assertEquals` signatures seem to get in the way.

wgtmac: I'm curious too. Perhaps we can override GroupValueSource by adding `Object getObject(String field, int index)` which internally dispatches getXXX based on its GroupType?

pitrou: It appeats this triggers a check failure:
```
Error:  Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.18.2:cmp (default) on project parquet-column: There is at least one incompatibility: org.apache.parquet.example.data.Group:METHOD_ABSTRACT_ADDED_IN_SUPERCLASS,org.apache.parquet.example.data.GroupValueSource.getObject(int,int):METHOD_ABSTRACT_ADDED_TO_CLASS -> [Help 1]
```

pitrou: I guess the question is: is `org.apache.parquet.example.data` supposed to be a public API?

pitrou: Or I can simply remove the abstract method from the interface and cast directly to `SimpleGroup`.

pitrou:
This is ready for review now. The only failing test is the interop test, since the [required file](https://github.com/apache/parquet-testing/pull/46) is not in parquet-testing yet.

wgtmac (commented):
I will take a look after I am back from vacation.

@gszadovszky Could you help review this, too?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -420,8 +429,27 @@ public Builder withDictionaryEncoding(String columnPath, boolean enableDictionar
...
+      return this;
+    }
+
+    /**
+     * Enable or disable BYTE_STREAM_SPLIT encoding for FLOAT, DOUBLE, INT32, INT64 and FIXED_LEN_BYTE_ARRAY columns.
+     *
+     * @param enable whether BYTE_STREAM_SPLIT encoding should be enabled
+     * @return this builder for method chaining.
+     */
+    public Builder withExtendedByteStreamSplitEncoding(boolean enable) {
```
gszadovszky, line 450: WDYT about handling it like we do for dictionary encoding. Dictionary can be enabled/disabled globally or per specific columns. It would help experimenting.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForInteger.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import static org.apache.parquet.bytes.BytesUtils.bytesToInt;
+
+public class ByteStreamSplitValuesReaderForInteger extends ByteStreamSplitValuesReader {
+  private static final int width = 4;
```
gszadovszky, line 24: nit: the preferred java format expects to have constants (`static final`) with upper case: `WIDTH`.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForLong.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+import static org.apache.parquet.bytes.BytesUtils.bytesToLong;
+
+public class ByteStreamSplitValuesReaderForLong extends ByteStreamSplitValuesReader {
+  private static final int width = 8;
```
gszadovszky, line 24: upper case...

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java
```diff
@@ -84,10 +92,25 @@ private ValuesWriter getBooleanValuesWriter() {
   }
 
   private ValuesWriter getFixedLenByteArrayValuesWriter(ColumnDescriptor path) {
-    ValuesWriter fallbackWriter = new DeltaByteArrayWriter(
-        parquetProperties.getInitialSlabSize(),
-        parquetProperties.getPageSizeThreshold(),
-        parquetProperties.getAllocator());
+    final ValuesWriter fallbackWriter;
+    // Heuristic: enable BYTE_STREAM_SPLIT for DECIMAL and FLOAT16 columns.
+    final boolean useByteStreamSplit = parquetProperties.isExtendedByteStreamSplitEnabled()
```
gszadovszky, line 97: Why the different heuristics for V1 and V2?

Do we know if `BYTE_STREAM_SPLIT` would be better for `DECIMAL`/`FLOAT16` values than `DELTA_BYTE_ARRAY`? It would be nice to come up with some smart heuristics instead of having the client set it, but if we don't know which one is better in certain cases, I would vote on not having any heuristics.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java
```diff
@@ -102,19 +125,39 @@ private ValuesWriter getBinaryValuesWriter(ColumnDescriptor path) {
   }
 
   private ValuesWriter getInt32ValuesWriter(ColumnDescriptor path) {
-    ValuesWriter fallbackWriter = new DeltaBinaryPackingValuesWriterForInteger(
-        parquetProperties.getInitialSlabSize(),
-        parquetProperties.getPageSizeThreshold(),
-        parquetProperties.getAllocator());
+    final ValuesWriter fallbackWriter;
+    // Ideally we should only enable BYTE_STREAM_SPLIT if compression is enabled for this column.
+    // However, this information is not available here.
```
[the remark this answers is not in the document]
gszadovszky: It is passed down from `ParquetWriter`/`ParquetRecordWriter` so defined per file. I agree it would be nice to have it here but currently I cannot imagine how could we do it.
Feel free to leave this comment here as a TODO, however, since we currently have it explicitly configure by the client, I don't think we need it.

gszadovszky (changes requested):
You probably want to remove `DO NOT MERGE: ` from the title.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFLBA.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public ByteStreamSplitValuesReaderForFLBA(int length) {
+    super(length);
+    valueByteBuffer = new byte[length];
+  }
+
+  @Override
+  public Binary readBytes() {
+    gatherElementDataFromStreams(valueByteBuffer);
+    return Binary.fromConstantByteArray(valueByteBuffer);
```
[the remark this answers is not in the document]
pitrou: Good catch. I should probably also test for this.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java
pitrou: * Footprint-wise:
  1. `DELTA_BYTE_ARRAY` is, at least conceptually, suboptimal for FLBA as it will store both prefix and suffix lengths, even though one can be trivially deduced from the other.
  2. `DELTA_BYTE_ARRAY` is most certainly a bad fit for very short data such as `FLOAT16`.
  3. While https://issues.apache.org/jira/browse/PARQUET-2414 did exercise `BYTE_STREAM_SPLIT` on `DECIMAL` and `FLOAT16` values, we don't know if it can be beneficial or detrimental for other kinds of `FIXED_LEN_BYTE_ARRAY` column.

* CPU-wise, `BYTE_STREAM_SPLIT` should be much faster than `DELTA_BYTE_ARRAY`, though concrete benchmarks would be needed for parquet-mr (we do have benchmarks for Parquet C++).

That said, perhaps we should indeed not add any heuristics and instead let the user configure things column-by-column (combining heuristics + column selection would be a bit delicate, though).

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java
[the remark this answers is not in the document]
pitrou: Pity. I'll just leave the comment then.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForFLBA.java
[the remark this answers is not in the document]
pitrou: Ok, this doesn't change anything unless you also materialize the `Binary` using `Binary::getBytes`, because it's only then that the `byte[]` is actually copied.

I think this begs the question: what is the contract for the `Binary` returned by `ValuesReader::readBytes`? Should its contents survive subsequent calls to `readBytes`?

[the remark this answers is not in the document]
pitrou: An alternative would be to decode all the data eagerly, like `DeltaBinaryPackingValuesReader` does.

parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultV2ValuesWriterFactory.java
gszadovszky: Maybe let the user decide for now and create an additional jira to implement smart heuristics. Then, some additional time can be spent on experimenting.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForDouble.java
```diff
@@ -18,20 +18,15 @@
...
   public ByteStreamSplitValuesReaderForDouble() {
     super(Double.BYTES);
-    valueByteBuffer = new byte[Double.BYTES];
   }
 
   @Override
   public double readDouble() {
-    gatherElementDataFromStreams(valueByteBuffer);
-    return Double.longBitsToDouble(BytesUtils.bytesToLong(valueByteBuffer));
+    return Double.longBitsToDouble(bytesToLong(readElement().getBytesUnsafe()));
```
pitrou, line 30: `getBytesUnsafe` is doing a copy here, but to a very small array. I don't know enough about JVM performance to decide if this is a problem or not. @gszadovszky Do you have any idea or suggestion?

(the underlying problem is that `ByteBuffer` and `Binary` have different limitations in their respective APIs, and whether we want to use one or the other depends on which data type we want to convert to :-))

pitrou: I may just use both a backing `Binary` and a `ByteBuffer` pointing to it.

pitrou: (note to self: https://spotcodereviews.com/articles/optimization/2021/03/15/does-the-jvm-eliminate-allocations-of-temporary-objects.html has an interesting discussion of array allocation elision)

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -76,18 +87,12 @@ public void initFromPage(int valuesCount, ByteBufferInputStream stream)
...
+    // Binary views into the internal decode buffer for FIXED_LEN_BYTE_ARRAY.
     final int totalSizeInBytes = stream.available();
-    byteStreamData = new byte[totalSizeInBytes];
-
-    // Eagerly read the data for each stream.
-    final int numRead = stream.read(byteStreamData, 0, totalSizeInBytes);
-    if (numRead != totalSizeInBytes) {
-      String errorMessage = String.format(
-          "Failed to read requested number of bytes. Expected: %d. Read %d.", totalSizeInBytes, numRead);
-      throw new ParquetDecodingException(errorMessage);
```
wgtmac, line 88: It seems that this check has been changed into assert somewhere else. I'd suggest to throw if unexpected padding or short read are found.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -76,18 +87,12 @@ public void initFromPage(int valuesCount, ByteBufferInputStream stream)
...
-    }
-
+    final ByteBuffer encodedData = stream.slice(totalSizeInBytes).slice(); // possibly zero-copy
+    final byte[] decodedData = decodeData(encodedData, this.valuesCount);
```
wgtmac, line 94: Do we need to hold the reference of `decodedData`?

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForInteger.java
```diff
@@ -0,0 +1,32 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.values.bytestreamsplit;
+
+public class ByteStreamSplitValuesReaderForInteger extends ByteStreamSplitValuesReader {
+  private static final int WIDTH = 4;
```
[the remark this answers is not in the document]
pitrou: I don't know, isn't it more readable by giving the value a name?

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
pitrou: I assume [`ByteBuffer.wrap`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#wrap-byte:A-) stores a reference to it, otherwise those APIs would be marked as unsafe and [`ByteBuffer.array`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#array--) wouldn't possibly work.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -140,4 +141,61 @@ public String memUsageString(String prefix) {
...
+
+    public FixedLenByteArrayByteStreamSplitValuesWriter(
+        int length, int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+      super(length, initialCapacity, pageSize, allocator);
+      this.length = length;
+    }
+
+    @Override
+    public final void writeBytes(Binary v) {
+      if (v.length() != length) {
```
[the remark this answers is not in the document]
pitrou: Good point. I copied this from `FixedLenByteArrayPlainValuesWriter` which does the same runtime check, but other `writeBytes` implementation don't seem to bother.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReaderForInteger.java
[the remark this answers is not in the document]
wgtmac: I just saw the magic number 4/8 in the constructor of IntegerByteStreamSplitValuesWriter . They would be changed to be consistent.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
pitrou: Are there situations where `stream.slice` below would return something smaller than what `stream.available` returned?

pitrou:
For the record, I must still address some of the review comments.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
wgtmac: No, but we do have seen cases where `numRead < totalSizeInBytes` so there is a clarification here: https://github.com/apache/parquet-format/commit/230711fbfd8d3399cce935a4f39d1be7b6ad5ad5

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInterOpReadByteStreamSplit.java
```diff
@@ -20,18 +20,23 @@
...
+import org.apache.parquet.example.data.simple.SimpleGroup;
 import org.apache.parquet.hadoop.example.GroupReadSupport;
 import org.junit.Test;
 
 public class TestInterOpReadByteStreamSplit {
-  private InterOpTester interop = new InterOpTester();
   private static final String FLOATS_FILE = "byte_stream_split.zstd.parquet";
-  private static final String CHANGESET = "4cb3cff";
+  private static final String EXTENDED_FILE = "byte_stream_split_extended.gzip.parquet";
+  // FIXME after https://github.com/apache/parquet-testing/pull/46 is merged
```
wgtmac, line 37: This could be removed now.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
pitrou: Well, I don't understand:
1. The `ByteBufferInputStream` declares that it has `totalSizeInBytes` bytes available
2. You ask the `ByteBufferInputStream` to read  `totalSizeInBytes` bytes
3. You end up with less than  `totalSizeInBytes` bytes read??

It seems unrelated with padding. It's just the physical size of the stream.

However, what could happen is that `totalSizeInBytes` is not a multiple of `elementSizeInBytes` for some reason, and currently this would hit the assertion at the end of `decodeData`. I could instead add a proper exception for this.

pitrou: (I understand that it might theoretically happen if `ByteBufferInputStream` is being read from by another thread, but that sounds like an API violation, doesn't it?)

pitrou: It seems the check was originally useful, but it became trivially true after this change:
https://github.com/apache/parquet-mr/commit/bf38e0fe549460936b310764a762749cb351ae6c#diff-215039b16c580f61a577c71dd878b2acf390d510229b9d190d8738fac43b1a6eR80-R86

wgtmac: Yes, you're right.

wgtmac (approved):
+1 on my side.

wgtmac:
@pitrou Do you want to have this in the 1.14.0 release?

gszadovszky (approved):
Sorry for the late response, @pitrou. Thanks for fixing my comments.

PR #1296, opened 2024-03-08
PARQUET-2448: Fix model conversions for avro <= 1.8 generated classes

RustedBones:
When running with avro <= 1.8, the default data supplier was looking for the root record's class conversions to initialize its model.

In case the record contained a nested class with another logical-type, the conversion will fail
Adapt the code to inspect the whole schema and initialize the model accordingly.


clairemcginty (commented):
Great catch @RustedBones . This change looks good to me  🚢

wgtmac:
Thanks for the fix! Could you please make the CI happy?

wgtmac:
I need to spend some time to get familiar with the code path.

cc avro experts @gszadovszky @Fokko .

doc/dremel_paper/dremel_example.png
gszadovszky, line 1: Something went odd with the formatting. It shouldn't touch binary files.

doc/dremel_paper/schema.png
gszadovszky, line 1: Same as above

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroRecordConverter.java
gszadovszky, line 1: It seems you have rewritten the whole unit test. Was it necessary to do so? I would feel more comfortable if the existing tests would pass as they were.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -215,31 +263,7 @@ static SpecificData getModelForSchema(Schema schema) {
...
-      final Conversion<?>[] conversions;
-      try {
-        conversionsField.setAccessible(true);
-        conversions = (Conversion<?>[]) conversionsField.get(null);
-      } catch (IllegalAccessException e) {
-        LOG.warn(String.format(
-            "Field `conversions` in class %s was inaccessible. Parquet will use default "
-                + "SpecificData model for reading and writing.",
-            clazz));
-        return null;
```
gszadovszky, line 235: At this point we were returning a `null` as the model. The new code does not seem to do so. Is it intentional and correct?

gszadovszky (changes requested):
I am not that experienced in the avro binding either but have some comments anyway. :)

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
RustedBones: Good catch. Update the PR so the previous behavior is kept

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroRecordConverter.java
RustedBones: Previous tests did not covered nested classes. I started to adapt the original code but that lead creating many classes with code duplication.
This refactoring went with the following simplification:
- test valid avro with generated class
- test deprecated avro (1.7 and 1.8 with same pseudo generated code)

gszadovszky: OK, I accept this.
To be on the safe side, could you revert the unit test changes on your env and execute it on the new code?

gszadovszky (approved):

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroRecordConverter.java
RustedBones: I can confirm that the original tests are passing with the changes

gszadovszky: Thank you @RustedBones!

PR #1300, opened 2024-03-14
PARQUET-2450: Fix Avro projection for single-field repeated record types

clairemcginty:
### Jira

  them in the PR title. For example, "PARQUET-2450: My Parquet PR"
    - https://issues.apache.org/jira/browse/PARQUET-2450

### Tests

### Commits

### Style

### Documentation

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -930,6 +930,7 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
   static boolean isElementType(Type repeatedType, Schema elementSchema) {
     if (repeatedType.isPrimitive()
         || repeatedType.asGroupType().getFieldCount() > 1
+        || repeatedType.getName().equals("array")
         || repeatedType.asGroupType().getType(0).isRepetition(REPEATED)) {
       // The repeated type must be the element type because it is an invalid
       // synthetic wrapper. Must be a group with one optional or required field
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.DirectWriterTest;
 import org.apache.parquet.hadoop.ParquetFileReader;
+import org.apache.parquet.io.InvalidRecordException;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.MessageTypeParser;
 import org.junit.Assert;
@@ -1105,7 +1106,7 @@ public class TestArrayCompatibility extends DirectWriterTest {
   }
 
   @Test
-  public void testIsElementTypeRequiredRepeatedRecord() throws Exception {
+  public void testIsElementTypeRequiredRepeatedRecord() {
     // Test `_tuple` style naming
     MessageType parquetSchema = MessageTypeParser.parseMessageType("message SchemaWithList {\n"
         + "  required group list_field (LIST) {\n"
@@ -1136,7 +1137,7 @@ public class TestArrayCompatibility extends DirectWriterTest {
   }
 
   @Test
-  public void testIsElementTypeOptionalRepeatedRecord() throws Exception {
+  public void testIsElementTypeOptionalRepeatedRecord() {
     // Test `_tuple` style naming
     MessageType parquetSchema = MessageTypeParser.parseMessageType("message SchemaWithList {\n"
         + "  optional group list_field (LIST) {\n"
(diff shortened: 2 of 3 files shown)


parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -930,6 +930,7 @@ public void end() {
   static boolean isElementType(Type repeatedType, Schema elementSchema) {
     if (repeatedType.isPrimitive()
         || repeatedType.asGroupType().getFieldCount() > 1
+        || repeatedType.getName().equals("array")
```
clairemcginty, line 933: This now matches the List [logical-type spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules):

> If the repeated field is a group with one field and is named either array or uses the LIST-annotated group's name with _tuple appended then the repeated type is the element type and elements are required.

as well as the equivalent isElementType check in AvroSchemaConverter: https://github.com/apache/parquet-mr/blob/61bdf44/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java#L575-L584

Note: We don't have to add another clause to check for the `_tuple` style naming scheme, as that's already covered by line 934: `repeatedType.asGroupType().getType(0).isRepetition(REPEATED))`.

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
```diff
@@ -603,7 +604,7 @@ public void testAvroCompatOptionalGroupInList() throws Exception {
   public void testAvroCompatOptionalGroupInListWithSchema() throws Exception {
     Path test = writeDirect(
         "message AvroCompatOptionalGroupInListWithSchema {" + "  optional group locations (LIST) {"
-            + "    repeated group array {"
+            + "    repeated group my_list {"
```
clairemcginty, line 607: This test started failing after I made the change to AvroRecordConverter. Looking at it, this schema didn't match the logical type spec for List types -- any repeated group named `array` must contain element type fields directly. So I updated the schema naming.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
wgtmac:
```
1. If the repeated field is not a group, then its type is the element type and elements are required.
2. If the repeated field is a group with multiple fields, then its type is the element type and elements are required.
3. If the repeated field is a group with one field and is named either array or uses the LIST-annotated group's name with _tuple appended then the repeated type is the element type and elements are required.
4. Otherwise, the repeated field's type is the element type with the repeated field's repetition.
```

`repeatedType.getName().equals("array")` is partially for rule 3, and `repeatedType.asGroupType().getType(0).isRepetition(REPEATED))` is for rule 4. But rule 3 also states the _tuple style naming scheme should come with `the repeated field is a group with one field`, isn't it? If the repeated field is a group with multiple fields with `_tuple` appended, is it still a valid element type?

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
wgtmac: I agree with you. Is it better to add a test case with the incorrect schema and check it throws?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
clairemcginty: oh interesting, so a schema like this?

```proto
message MySchema {
  optional group list_field (LIST) {
    repeated group list_field_tuple_1 (LIST) {
      required int32 int_field_1;
    }
    repeated group list_field_tuple_2 (LIST) {
      required int32 int_field_2
    }
  }
}
```

I think luckily, it's impossible to create this in Avro. I tried running it through AvroSchemaConverter, and it threw an exception:

```java
 > MessageType parquetSchema = MessageTypeParser.parseMessageType(
        "message MySchema {\n"
        + "  optional group list_field (LIST) {\n"
        + "    repeated group list_field_tuple_1 (LIST) {\n"
        + "      required int32 int_field_1;\n"
        + "    }\n"
        + "    repeated group list_field_tuple_2 (LIST) {\n"
        + "      required int32 int_field_2;\n"
        + "    }\n"
        + "  }\n"
        + "}");

> new AvroSchemaConverter().convert(parquetSchema));
java.lang.UnsupportedOperationException: Invalid list type optional group list_field (LIST) {
  repeated group list_field_tuple_1 (LIST) {
    required int32 int_field_1;
  }
  repeated group list_field_tuple_2 (LIST) {
    required int32 int_field_2;
  }
}
```

parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java
clairemcginty: sounds good! added.

wgtmac (approved):
Thanks for the detail test!

LGTM

asuresh8:
What's the timeline on getting a version of parquet-mr with this change published to Maven? When will 1.14.0 be release? Can this be backported 1.13?

wgtmac:
Yes, I expect this will be included in the 1.14.0 release which is targeted in the mid 2024.

PR #1304, opened 2024-03-21
PARQUET-2456: Avoid type name conflicts on fixed type

RustedBones:
Fixed types with identical names create name type conflicts

For instance, if a parquet message type contains several repeated fixed types, the following error will be thrown:
```
org.apache.avro.SchemaParseException: Can't redefine: array
```

Use the same conflict resolution trick used for record types.


wgtmac (approved):
+1

Thanks for the fix! Would you mind creating a JIRA issue for this?

PR #1309, opened 2024-03-25
PARQUET-2454: invoking flush before closing the output stream

ahshahid:
In customer deployments, sporadically the parquet generated using spark sql 's "insert overwrite" is found to be corrupted when read back. The writing tasks all commit successfully and no exception of any sort is seen. Given the low frequency of occurrence and deficit of just 248 bytes ( as gleaned from the log analysis), points to the case that footer data might be missing or not written fully, with close getting invoked.
So suggest adding a flush before closing of the output stream. 

### Jira
 My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2454)

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET-2454) issues and references

### Tests

### Commits

### Style

### Documentation


wgtmac (approved):
This seems benign to me. WDYT? @gszadovszky

gszadovszky (commented):
This is a good catch, @ahshahid. Thanks for working on this!

Based on the contracts of `OutputStream.flush()` and `OutputStream.close()`, this call is even necessary. (This was surprising to me. I've thought that `close()` would involve `flush()` based on the implementations I've seen so far.)

Would you mind searching through the code for `OutputStream.close()` calls where the `OutputStream` may come from outside of parquet-mr and add `flush()` as well?

Another thing: if an exception happens during `flush()`, `close()` would never be invoked. (Most probably in this case `close()` would fail as well but this is not 100%.)
You may use `AutoCloseables` to make the code simpler but it would need some additional developments at `AutoCloseables` because of the potential `IOException`.

ahshahid:
@wgtmac @gszadovszky 
Thanks for reviewing and your inputs.
I agree , for buffered streams ,flush is necessary before closing..
Yes, my thinking too is that if flush throws an Exception, close in any case would throw.. But if you believe not 100% guaranteed, can modify the code to attempt close irrespective of flush's outcome.
I have not seen AutoCloseable... let me check.. if it is little invasive, I will avoid using as I am not well versed with the parquet codebase..
After the PR is merged, may be more knowledgeable can refine it.
I will check for OutputStream.close.. Though I assume when you say outside of parquet-mr,  you mean other places in parquet-mr project only right..?

ahshahid:
Hi. Pls allow work flows to be triggered for the PR

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
```diff
@@ -0,0 +1,43 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.util;
+
+import java.io.IOException;
+
+public class AutoCloseableResources implements AutoCloseable {
```
wgtmac, line 24: Why not directly work on `AutoCloseables`?

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
```diff
@@ -108,7 +109,11 @@ public void reset() {
 
   @Override
   public void close() {
-    out.close();
+    try (CapacityByteArrayOutputStream cbaos = out) {
+      cbaos.flush();
```
wgtmac, line 113: This one is used for in-memory encoding only. I suspect similar streams in this module are required to call flush anyway.

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
ahshahid: well I tried to , but AutoCloseables just contain static functions, To me it just seems to ensure that all the passed AutoCloseables are closed. , nothing gets missed.
I cannot figure out how to use it in paradigm of 
try (    AutoCloseable resources) {
    // some operations like say 
   flush
}

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
ahshahid: You want me to remove that ?. For safe side, I am calling flush.. wherever there is buffering possibility

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
wgtmac: Does checking if the object to close is an impl of `Flushable` help?

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
wgtmac: I think this PR aims to a potential I/O issue, which now grows larger than I expected. Those output streams used in compression and encoding are mainly as in-memory buffers. So I suspect we should include them as well.

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
ahshahid: I suppose it can.. but do we want to use that check in AutoCloseables? Coz the suggested paradigm for AutoCloseable is try ( ) { }.  What right now AutoCloseables is doing, not sure why it is requiring AutoCloseable? It could well have been Closeable?

ahshahid:
Hi. pls enable work flows for the PR.

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
ahshahid: Also it may cause issues, and possible messy code, like say we have 3 AutoCloseable Resources, having flush.. And the safest behaviour would be to flush each and then close in the end the three.
If in a single loop ,we put flush and close, it might create problems.  so to be safe, we would may be have to split into two loops, first flushables doing flush, then. closing...  Which would mean handling IOExceptions of the flushable loop too.

ahshahid:
@wgtmac @gszadovszky Are you ok merging these changes?

wgtmac:
I still think this PR has been a little bit distracted like https://github.com/apache/parquet-mr/pull/1309#discussion_r1548839242. I'll wait for advise from @gszadovszky

wgtmac (changes requested):
I still think this PR has been a little bit distracted like https://github.com/apache/parquet-mr/pull/1309#discussion_r1548839242. I'll wait for advise from @gszadovszky

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
ahshahid: @wgtmac Include them .. where. could you pls elaborate more? AFAIK, any outstream which uses in - memory buffers , should be flushed.. When I checked the hierarchy of the Output streams, I found only those which actually needed flush.
For eg: DeltaBinaryPackingValuesWriterForInteger, in close() calls super.close.. And the super.close , has the code for flushing.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesWriter.java
```diff
@@ -126,8 +128,18 @@ public void reset() {
...
-    arrayOut.close();
-    out.close();
+    try (AutoCloseableResources ignored = new AutoCloseableResources(arrayOut, out)) {
+      try {
+        arrayOut.flush();
+      } catch (IOException ignore) {
+      }
+      try {
+        out.flush();
+      } catch (IOException ignore) {
```
shangxinli, line 138: Debug log?

parquet-common/src/main/java/org/apache/parquet/bytes/LittleEndianDataOutputStream.java
```diff
@@ -208,9 +208,9 @@ public final void writeDouble(double v) throws IOException {
   }
 
   public void close() {
-    try {
-      out.close();
-    } catch (IOException e) {
+    try (OutputStream os = this.out) {
+      os.flush();
+    } catch (Exception e) {
       // swallow exception
```
shangxinli, line 214: debug log?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesWriter.java
ahshahid: will add

parquet-common/src/main/java/org/apache/parquet/bytes/LittleEndianDataOutputStream.java
ahshahid: will add .

ahshahid:
pls allow workflows to run

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
wgtmac: For example, `CapacityByteArrayOutputStream` class above only inherits the default flush() method, why should we bother calling it here at all?

parquet-common/src/main/java/org/apache/parquet/io/DelegatingPositionOutputStream.java
```diff
@@ -35,7 +35,11 @@ public OutputStream getStream() {
 
   @Override
   public void close() throws IOException {
-    stream.close();
+    try (OutputStream os = this.stream) {
```
wgtmac, line 38: I also suspect that delegate classes like `LittleEndianDataOutputStream` and `DelegatingPositionOutputStream` should take this complexity.

parquet-common/src/main/java/org/apache/parquet/util/AutoCloseableResources.java
wgtmac: I agree that calling flush in AutoCloseables may introduce more troubles if exception happens. However, adding duplicate code like this may also impose maintenance burden.

parquet-common/src/main/java/org/apache/parquet/io/DelegatingPositionOutputStream.java
ahshahid: Sorry I did not get what you mean.. Are you saying that these classes should not be bothered with flush etc?

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/BitPackingValuesWriter.java
ahshahid: makes sense. it is not needed here.. as of now because as you said CapacityByteArrayOutputStream 's close is sufficient and its flush is a no op.
I can remove it .

ahshahid:
pls allow work flow run

parquet-common/src/main/java/org/apache/parquet/io/DelegatingPositionOutputStream.java
wgtmac: Yes, for example `LittleEndianDataOutputStream` is used by column encoders which is accessed via `ValuesWriter.getBytes()` at https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/values/ValuesWriter.java#L43. We can see that `getBytes()` implementations have guaranteed to call flush() before consuming the encoded bytes. We don't have to call flush() in the close() since the data is not used any more.

ahshahid: I see. well if you all think so, can remove that .. Though the field var is defined as OutputStream, so in theory could contain a buffered output stream or any other custom OutputStream with some buffered data and may or may not have flush implemented or may or may not have flush invoked before close.

ConeyLiu:
@ahshahid which type of exception did you meet when reading the corrupted files? I have met some strange corrupted files recently.

ahshahid:
> @ahshahid which type of exception did you meet when reading the corrupted files? I have met some strange corrupted files recently.

@ConeyLiu .. The actual issue occured when executing Insert/Overwrite sql using Spark And Iceberg ( i.e while writing though the tasks completed successfully)
The issue got revealed during reading .. ( it became clear when reading using impala...)
WARNINGS: Error decompressing data page in file ‘/pathtodatafile/’. Expected 15356 uncompressed bytes but got 15108

parquet-common/src/main/java/org/apache/parquet/io/DelegatingPositionOutputStream.java
wgtmac: This is not a strong opinion.

wgtmac (approved):
LGTM +1

ahshahid:
@wgtmac @gszadovszky  pls approve work flow

ConeyLiu:
Thanks @ahshahid for you sharing.

ahshahid:
@wgtmac @gszadovszky  Thanks for the approval and reviews!.

ahshahid:
I had mistakenly closed it. Pls merge

gszadovszky (approved):
Sorry for the late response. I'm fine merging this.

ahshahid:
Thanks. Kindly merge..I do not have permission..
Regards
Asif

pan3793:
@ahshahid hi, would you mind sharing more about the environment where the issue happened? for example, versions of Spark/Parquet/Hadoop, and what type of storage are you using? thank you in advance.

ahshahid:
Unfortunately, the issue was seen in customer env while I was employed with
Cloudera. I do not have access to details now. May be Steve Loughran can
help you provide specific details..

PR #1310, opened 2024-03-26
PARQUET-1511: Pass the parameters in a configuration object to Schema…

tianchen92:
…Converter

### Jira

related to [https://issues.apache.org/jira/browse/PARQUET-1511](url)

### Tests


wgtmac (changes requested):
IMHO, it is not a good practice to add the hadoop dependency to parquet-arrow. 

cc @Fokko, please correct me if I was wrong.

parquet-arrow/src/main/java/org/apache/parquet/arrow/schema/SchemaConverter.java
```diff
@@ -87,19 +88,21 @@
...
    */
   public SchemaConverter() {
-    this(false);
+    Configuration conf = new Configuration(false);
+    conf.setBoolean(CONVERT_INT96_TO_ARROW_TIMESTAMP, false);
+    this.configuration = conf;
   }
 
-  // TODO(PARQUET-1511): pass the parameters in a configuration object
-  public SchemaConverter(final boolean convertInt96ToArrowTimestamp) {
```
Fokko, line 101: We should also keep this constructor for backward compatibility

Fokko (changes requested):
@wgtmac You're right on this one. We don't want to introduce Hadoop since we're looking into decoupling from Hadoop.

PR #1328, opened 2024-04-23
PARQUET-34: Add #contains FilterPredicate for Array columns

clairemcginty:
Proposal to add a new FilterPredicate, `Contains`, that can be applied to List types, and check if the specified element is present among the repeated values. It can be composed using And or Or:

```java
FilterPredicate predicate = contains(
  eq(binaryColumn("phoneNumbers.phone.kind"), Binary.fromString("cell"))
)

FilterPredicate predicate = or(
  contains(eq(binaryColumn("phoneNumbers.phone.kind"), Binary.fromString("cell"))),
  contains(eq(binaryColumn("phoneNumbers.phone.kind"), Binary.fromString("home")))
)

FilterPredicate predicate = and(
  contains(eq(binaryColumn("phoneNumbers.phone.kind"), Binary.fromString("cell"))),
  contains(eq(binaryColumn("phoneNumbers.phone.kind"), Binary.fromString("home")))
)
```

The filtering logic is largely based on existing Eq predicates to apply filtering at the page/rowgroup level using statistics/dictionaries, with a specialized implementation in `IncrementallyUpdatedFilterPredicateBuilder` to do individual record-level filtering. 

### Jira

    - https://issues.apache.org/jira/browse/PARQUET-34

### Tests

### Commits

### Style

### Documentation


parquet-hadoop/src/main/java/org/apache/parquet/filter2/bloomfilterlevel/BloomFilterImpl.java
```diff
@@ -119,6 +119,16 @@ public <T extends Comparable<T>> Boolean visit(Operators.GtEq<T> gtEq) {
     return BLOCK_MIGHT_MATCH;
   }
 
+  @Override
+  public <T extends Comparable<T>> Boolean visit(Operators.Contains<T> contains) {
+    return BLOCK_MIGHT_MATCH;
```
clairemcginty, line 124: I was testing this out with the BloomFilterVisitor and couldn't produce a test Parquet file with a populated bloom filter for a List field (I tried strings, ints, and longs as element types). I guess this is expected?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
```diff
@@ -257,6 +266,16 @@ public static <T extends Comparable<T>, C extends Column<T> & SupportsEqNotEq> N
     return new NotIn<>(column, values);
   }
 
+  public static <T extends Comparable<T>, C extends Column<T> & SupportsContains> Contains<T> contains(
```
clairemcginty, line 269: Thinking to the future, it could eventually be useful to support Array FilterPredicates like "repeated field X contains an int field greater than Y". We could refactor this API to be written like:

```java
FilterApi.contains(
  FilterApi.eq(
    FilterApi.arrayColumn(FilterApi.intColumn("repeated_int_field")), 
    100
  )
)
```

& we could eventually support `FilterApi.contains(FilterApi.lt(...))`, `FilterApi.contains(FilterApi.gt(...))`, etc...

wgtmac:
Thanks for the new feature! I will try to take a look soon and make it to the 1.14.0 release.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: It's not a standard SQL function, but I've seen it in SQL extension languages such as [BigQuery Standard SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#in_operators), and I've gotten several requests to support this by users of the [Scio](https://github.com/spotify/scio) Parquet library!

that's a good point about making this composable, I think it would be more efficient to do `CONTAINS(a or b)` than `CONTAINS(a) or CONTAINS(b)`. What do you think about supporting `lt/gt` in addition to `eq`-based Contains?  for example, `CONTAINS(eq(a) OR gt(b))` ? It would make this PR a lot more complex but I'm happy to try. We could probably re-use a lot of the existing filter code for `eq`, `lt/gt`, etc...

wgtmac: Yes, I agree that it worths making the function composable. Perhaps we can define all the necessary API (with implementation of the most common ones) in the PR and implementations of all other kinds of inputs can be split into separate PRs. Let's not hurry for the coming release. It still takes time to stabilize.

pom.xml
```diff
@@ -603,6 +603,46 @@
 
               <exclude>org.apache.parquet.io.api.Binary#get2BytesLittleEndian()</exclude>
               <exclude>org.apache.parquet.schema.LogicalTypeAnnotation$Float16LogicalTypeAnnotation#accept(org.apache.parquet.schema.LogicalTypeAnnotation$LogicalTypeAnnotationVisitor)</exclude>
+              <exclude>org.apache.parquet.filter2.predicate.Operators$Contains</exclude>
```
wgtmac, line 606: Is it possible to avoid these exclusions?

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: Sounds great. I would love to get predicates for array size in there, too, but I think that one is going to be a bit more complex (we might need to update Statistics to reflect array sizes?)... that one can wait for a separate PR.

wgtmac:
As [requested on the dev@parquet ML](https://lists.apache.org/thread/cmlmo4nsdhdzq3sfdb18010jmw54z6s2), I'll wait for this PR before starting the releasing process of 1.14.0.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -315,6 +321,66 @@ public <R> R accept(Visitor<R> visitor) {
     }
   }
 
+  public abstract static class ArrayContainsFilterPredicate<T extends Comparable<T>>
```
gszadovszky, line 324: It seems this class does not need to be public. Let's have it package private.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -315,6 +321,66 @@ public <R> R accept(Visitor<R> visitor) {
...
+      String name = getClass().getSimpleName().toLowerCase(Locale.ENGLISH);
+      return name + "(" + elementColumn.getColumnPath().toDotString() + ", " + value + ")";
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      ArrayContainsFilterPredicate<?> that = (ArrayContainsFilterPredicate<?>) o;
+      return elementColumn.equals(that.elementColumn) && value.equals(that.value);
```
gszadovszky, line 353: It seems that `value` can be `null`. In this case it'll throw an NPE.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -487,6 +489,92 @@ public <T extends Comparable<T>> Boolean visit(NotIn<T> notIn) {
...
+  @Override
+  public <T extends Comparable<T>> Boolean visit(DoesNotContain<T> doesNotContain) {
+    T value = doesNotContain.getValue();
+
+    Column<T> filterColumn = doesNotContain.getColumn();
+    ColumnChunkMetaData meta = getColumnChunk(filterColumn.getColumnPath());
+
+    if (value == null) {
+      if (meta == null) {
+        // @Todo ??
```
gszadovszky, line 538: Why the TODO? No metadata means the column is not in the file so every value is null. Therefore returning `BLACK_CANNOT_MATCH` is correct.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
clairemcginty: sounds good! I think I'll have to refactor this API design a bit to make Array filters composable, as @wgtmac suggested above -- will try to push changes late today or Monday 🙏

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: Actually, I think I need some clarification on the expected behavior of And/Or for an array field...

Say we have the following records, each containing a repeated int field with the following values:

```
Record1:
  repeated_int_field: [2, 3, 4, 5]
Record2:
  repeated_int_field: [1, 7]
```

Given the predicate `contains(and(gt(3), lt(6)))`, should we return:

1. Only Record1, because it contains **at least one element that satisfies both predicates** gt(3) and lt(6), or
2. Both Record1 and Record2, because Record2 contains **elements that satisfy each predicate separately** (7 matches `gt(3)` but not `lt(6)`, and 1 satisfies `lt(6)` but not `gt(3)`?

I think I would assume 2 🤔 but not totally sure. 

cc @gszadovszky as well :)

wgtmac: I would expect the option 1. Option 2 looks more like contains(or(gt(3),lt(6))). The case would be more complex if we add NOT expr to it. If there is any confusion like this, what about only supporting eq and ne for now? More complicated expressions need more research and inputs from other experts.

clairemcginty: That sounds good! 👍

parquet-hadoop/src/main/java/org/apache/parquet/filter2/bloomfilterlevel/BloomFilterImpl.java
clairemcginty: that worked! thanks 👍

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: Ok, I've pushed my changes to support composable contains predicates! example usage:

```java
FilterApi.containsOr(
  FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), 5555555555L),
  FilterApi.containsOr(
    FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), -10000000L), // Won't be matched
    FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), 2222222222L)));
```

I left out implementing DoesNotContain/ContainsNotEq for now to keep the PR simpler to parse. Let me know if the API looks ok 👍 Then I can fill out the rest of the implementations/make the unit tests more thorough.

pom.xml
clairemcginty: I ended up removing ArrayColumn and rebasing onto the latest master branch, and the exclusions were no longer needed 👍

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
wgtmac: > It's not a standard SQL function, but I've seen it in SQL extension languages such as [BigQuery Standard SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#in_operators), and I've gotten several requests to support this by users of the [Scio](https://github.com/spotify/scio) Parquet library!
> 
> that's a good point about making this composable, I think it would be more efficient to do `CONTAINS(a or b)` than `CONTAINS(a) or CONTAINS(b)`. What do you think about supporting `lt/gt` in addition to `eq`-based Contains? for example, `CONTAINS(eq(a) OR gt(b))` ? It would make this PR a lot more complex but I'm happy to try. We could probably re-use a lot of the existing filter code for `eq`, `lt/gt`, etc...

I took a second look, it seems that you were mentioning `IN` expression which we already have: https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java#L209-L258

wgtmac: > Ok, I've pushed my changes to support composable contains predicates! example usage:
> 
> ```java
> FilterApi.containsOr(
>   FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), 5555555555L),
>   FilterApi.containsOr(
>     FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), -10000000L),
>     FilterApi.containsEq(longColumn("phoneNumbers.phone.number"), 2222222222L)));
> ```
> 
> I left out implementing DoesNotContain/ContainsNotEq for now to keep the PR simpler to parse. Let me know if the API looks ok 👍 Then I can fill out the rest of the implementations/make the unit tests more thorough.

Sorry I didn't make it clear. I thought we could still leverage [existing composite expression](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java#L293-L326) and combine `CONTAINS(x)` sub-expressions under the same level of `AND` or `OR` expression during rewrite process like https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/filter2/compat/FilterCompat.java#L79. It would be harder to maintain the code if introduce new specialized composite expression like `containsOr`.

clairemcginty: ahh, ok. So the rewriter would inspect the runtime classes of the `left` and `right` predicates of an `And` or `Or`, and if they match the new Contains* expressions, we rewrite it into a single `Contains` predicate containing both clauses?

wgtmac: I would expect so, but it seems requiring a lot more work. We can support the simplest case at this point.

wgtmac:
BTW, should we postpone this feature to 1.15.0 release? We can always release new version as needed and I can volunteer to be the release manager once this is ready.

gszadovszky:
> BTW, should we postpone this feature to 1.15.0 release? We can always release new version as needed and I can volunteer to be the release manager once this is ready.

I agree with @wgtmac. Let's create smaller PRs and make improvements then release when we feel it stable. During the development, we may advertise this feature so others may start experimenting on it and give feedback before actually releasing it. A too early release might make later improvements harder because we need to be backward compatible.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: ok, I think I mostly got this working, supporting just Contains Eq (+ composition using And/Or) to start. hopefully this is closer to what you had in mind!

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -368,6 +369,20 @@ public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(NotIn<T> notIn) {
       return IndexIterator.all(getPageCount());
     }
 
+    @Override
+    public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(Contains<T> contains) {
```
clairemcginty, line 373: I need to think about the behavior here one -- `visit(And and)` and `visit(Or or) ` in `ColumnIndexBuilder` throw an UnsupportedOperationException, so I just copied that behavior for composed Contains predicates. But I'm thinking we can actually support this by taking the union/intersection of the `PrimitiveIterator.OfInt`, kind of similar to how `ColumnIndexFilter` [works](https://github.com/apache/parquet-mr/blob/apache-parquet-1.14.0-rc0/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java#L193-L209).

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
```diff
@@ -181,6 +187,84 @@ public void testInFilter() throws Exception {
     assert (found.size() == 102);
   }
 
+  @Test
```
clairemcginty, line 190: Todo: add tests for a `contains(eq(null))` filter

wgtmac:
Sorry for the delay. I will try to finish another pass by the end of this week.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
wgtmac: CMIW, throwing exception here might prohibit `contains` predicate to be used to filter parquet files with page index available.

wgtmac (commented):
This overall LGTM! Thanks @clairemcginty for working on this and adding exhaustive test!

cc @gszadovszky @julienledem @rdblue Sorry for bothering. I think this PR requires visibility to more reviewers.

clairemcginty:
> This overall LGTM! Thanks @clairemcginty for working on this and adding exhaustive test!

great, I'm glad this implementation looks ok! I have a few more tests that I'd like to add around null handling + behavior of the `Contains` predicate on map types (I think that they should just work, but I haven't tried it out yet...). Will try to add those + address PR comments on Monday or Tuesday next week 👍

clairemcginty:
I tried adding a test case to `TestRecordLevelFilters` to test `contains(eq(null))`. My expectation was that if you have an array schema with an optional element type, this should return true if the array contains one or more null elements. However, I don't think this is possible to make work--I set a debugger on `ValueInspector#update` and `ValueInspector#updateNull`. and `ValueInspector#update` is only invoked for non-null elements, and `ValueInspector#updateNull` is only invoked if the entire array is null, which isn't exactly what we want, either.

So based on my current understanding, I don't think we can support a `contains(eq(null))` predicate, and we can probably add a precondition check to the `Contains` constructor against null elements. Wdyt @wgtmac ?

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
clairemcginty: Makes sense. I'll look into doing an intersection/union operation here. Or otherwise just fall back to not filtering and returning `IndexIterator.all`.

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
```diff
@@ -122,6 +127,11 @@ public void run() throws IOException {
     }
     addVisitEnd();
 
+    addContainsBegin();
+    addVisitBegin("Contains");
+    addContainsCase();
```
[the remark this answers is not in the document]
clairemcginty: Added!

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ContainsRewriter.java
```diff
@@ -0,0 +1,185 @@
...
+      }
+      return ((Contains) left).or(right);
+    } else {
+      return or;
+    }
+  }
+
+  @Override
+  public FilterPredicate visit(Not not) {
+    return not;
```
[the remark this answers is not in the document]
clairemcginty: I think you're right, or at least, all the existing tests passed when I added an exception thrown here 👍

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
clairemcginty: Updated the PR with implementations for each. I implemented them as new `PrimitiveIterator.OfInt`s that zip together the component left & right hand sides, deduplicating/intersecting/merging as needed. Implementing custom iterators is kind of error prone though... if this adds too much complexity I can always save it for a followup PR, or just materialize the iterators into `List`s which are a lot easier to work with 😅

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
```diff
@@ -257,6 +258,10 @@ public static <T extends Comparable<T>, C extends Column<T> & SupportsEqNotEq> N
     return new NotIn<>(column, values);
   }
 
+  public static <T extends Comparable<T>> Contains<T> contains(Eq<T> pred) {
```
clairemcginty, line 261: In a fast-follow-up PR we can just replace this signature with:

```java
public static <T extends Comparable<T>> Contains<T> contains(ColumnFilterPredicate<T> pred) {
   ...
}
```

and remove the `SupportsContains` annotation. I did a few tests of `Contains(lt)` and `Contains(gt)` and everything is working as expected 👍

clairemcginty (commented):
I did some ad-hoc testing of this on larger datasets (1m+ elements with various page/row group distributions) and confirmed that everything looks right! so I think this PR is finally complete/ready for review.

The only outstanding task is to include support for `NotEq/Lt/Gt/Lte/Gte/In/NotIn` with Contains, which is just updating the `FilterApi` signature and adding a few lines to `IncrementallyUpdatedFilterPredicateGenerator`. I think that can be a followup PR 👍

gszadovszky:
@wgtmac, completely agree to have more people reviewing this. Thanks for pinging.
I'll try to take a look this week.

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
```diff
@@ -20,6 +20,7 @@
 
 import static org.apache.parquet.filter2.predicate.FilterApi.and;
 import static org.apache.parquet.filter2.predicate.FilterApi.binaryColumn;
+import static org.apache.parquet.filter2.predicate.FilterApi.contains;
```
gszadovszky, line 23: During checking the code I've found that this record level testing class does not seem to be correct. In several places (just like in your code) the method `PhoneBookWriter.readFile` is used which calls [PhoneBookWriter.createReader](https://github.com/apache/parquet-java/blob/240ab0daa005e9cd9dcc9d3df716cd321761b307/parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/PhoneBookWriter.java#L335-L339) to create the reader instance. The issue is, by default all the filtering (statistics/dictionary/column index) are working. As a result, we cannot be sure that the read results we are getting are really filtered only by the record level filter.
I know this is not your fault but could you please try using a properly set up reader for this test?
```java
    // ...

    return ParquetReader.builder(new GroupReadSupport(), file)
        .withConf(conf)
        .withFilter(filter)
        .withAllocator(allocator)
        .useBloomFilter(false)
        .useDictionaryFilter(false)
        .useStatsFilter(false)
        .useColumnIndexFilter(false)
        .build();
```

If unrelated errors may occur, let's handle this separately, but I want to avoid not catching a potential issue in your code by leaving this test as is.

gszadovszky (changes requested):
Great work, @clairemcginty!
I've added some comments.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -368,6 +369,102 @@ public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(NotIn<T> notIn) {
...
+          // For Contains.And, compute the intersection of the int ranges
+          (l, r) -> new PrimitiveIterator.OfInt() {
+            private Integer next = null;
+
+            @Override
+            public int nextInt() {
+              int result = next;
+              next = null;
+              return result;
+            }
```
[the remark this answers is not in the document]
clairemcginty: makes sense! I had put some of the next-loading logic into `hasNext()` because as far as I could tell, this iterator would be used via the `forEachRemaining()` method, which IIRC does sequentially call `hasNext()/next()`... but it does make this a bit unstable 😅 I can rewrite it and move into static methods in `IndexIterator`.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/FilterApi.java
clairemcginty: understood -- I can commit to working on the follow-up PR right away 👍

parquet-hadoop/src/test/java/org/apache/parquet/filter2/recordlevel/TestRecordLevelFilters.java
clairemcginty: Ha, I did notice that -- I'd been refactoring StatisticsFilter, made a mistake, and was confused why `TestRecordLevelFilter` suddenly started failing.

updated! All tests continued to pass after disabling the other filters.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/IndexIterator.java
```diff
@@ -55,6 +55,105 @@ static PrimitiveIterator.OfInt rangeTranslate(int from, int to, IntUnaryOperator
     return new IndexIterator(from, to + 1, i -> true, translator);
   }
 
+  static PrimitiveIterator.OfInt intersection(PrimitiveIterator.OfInt lhs, PrimitiveIterator.OfInt rhs) {
+    return new PrimitiveIterator.OfInt() {
+      private int next = fetchNext();
+
+      @Override
+      public int nextInt() {
+        int result = next;
```
gszadovszky, line 64: Sorry for bothering with the iterator contract again. It should throw `NoSuchElementException` if there are no more elements (instead of returning `-1`).

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/IndexIterator.java
```diff
@@ -55,6 +55,105 @@ static PrimitiveIterator.OfInt rangeTranslate(int from, int to, IntUnaryOperator
...
+
+  static PrimitiveIterator.OfInt union(PrimitiveIterator.OfInt lhs, PrimitiveIterator.OfInt rhs) {
+    return new PrimitiveIterator.OfInt() {
+      private int peekL = -1;
+      private int peekR = -1;
+      private int next = fetchNext();
+
+      @Override
+      public int nextInt() {
+        int result = next;
```
gszadovszky, line 109: Missing exception, see above.

gszadovszky (changes requested):
I've found an issue missed before.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/IndexIterator.java
clairemcginty: thanks for catching that! Updated both.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/IndexIterator.java
```diff
@@ -55,6 +55,111 @@ static PrimitiveIterator.OfInt rangeTranslate(int from, int to, IntUnaryOperator
...
+        }
+        if (nextL == nextR) {
+          return nextL;
+        }
+        while (nextR < nextL && rhs.hasNext()) {
+          nextR = rhs.next();
+        }
+        if (nextL == nextR) {
+          return nextL;
+        }
```
gszadovszky, line 97: It seems that an outer loop is missing that would run till the end of the shorter iterator. Example:
```
lhs: 1, 3, 5, 7
rhs: 2, 4, 6, 7
```
After these lines at the first invocation of the method `nextL` will be `3` and `nextR` will be `4` and `-1` is about the be returned while `7` is a common element in both.
(It also means that the unit test for this one is not complete...)

gszadovszky (changes requested):
Sorry @clairemcginty, my most important comment got lost from the previous review, somehow.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/IndexIterator.java
clairemcginty: ah I see what you mean and replicated this in a test! I think I need to use a broader while loop. Refactoring now.

gszadovszky (approved):
Thanks, @clairemcginty. It looks good to me. Let's wait for the tests to pass.

@wgtmac, since no one else chimed in this review, I think, we can step forward and merge it (after CI approval). There will be follow up changes anyway and the next release is probably a couple months away that gives time for others to comment on this feature.

wgtmac:
Thanks @gszadovszky for the detail review! I'll take another pass shortly to be familiar with the latest change and then merge it if no concern.

wgtmac (approved):
+1

PR #1335, opened 2024-04-28
PARQUET-2430: Add parquet joiner v2

MaxNevermind:
This is a simplified version of original proposed functionality of a joiner, see description of original idea and simplified design below. 

## Original design

See related original PR: [[WIP][Proposal] PARQUET-2430: Add parquet joiner](https://github.com/apache/parquet-mr/pull/1273)

ParquetJoiner feature is similar to ParquetRewrite class. ParquetRewrite allows to stitch files with the same schema into a single file while ParquetJoiner should enable stitching files with different schemas into a single file. That is possible when: 1) the number of rows in the main and extra files is the same, 2) the ordering of rows in the main and extra files is the same. Main benefit of ParquetJoiner is performance, for the cases when you join/stitch Terabytes/Petabytes of data that seemingly simple low level API can be up to 10x more resource efficient.
Implementation details

ParquetJoiner allows to specify the main input parquet file and extra input parquet files. ParquetJoiner will copy the main input as binary data and write extra input files with row groups adjusted to the main input. If main input is much larger than extra inputs then a lot of resources will be saved by working with the main input as binary.
Use-case examples

A very large Parquet based dataset(dozens or hundreds of fields/Terabytes of data daily/Petabytes of historical partitions). The task is to modify a column or add a new column to it for all the historic data. It is trivial using Spark, but taking into consideration the share scale of a dataset it will take a lot of resources to do that.
Side notes

Note that this class of problems could be in theory solved by storing main input and extra inputs in HMS/Iceberg bucketed tables and use a view that joins those tables on the fly into the final version but in practice there is often a requirement to merge parquet files and have a single parquet sources in the file system.
Use-case implementation details using Apache Spark

You can use Apache Spark to perform the join with ParquetJoiner, read the large main input and prepare the right side of a join in a way that each file on the left have a corresponding file on the right and it preserves records ordering on the right side in the same order as on the left side, that allows the whole input on the left and right to have the same number of files and the same number of records in corresponding files and the same ordering of records in each file pair. Then run ParquetJoiner in parallel for each file pair and perform a join. Example of the code that utilizes this new feature: https://gist.github.com/MaxNevermind/0feaaf380520ca34c2637027ef349a7d.

##  A simplified design(this PR)
- has only one list of `inputFilesToJoin` instead of `List<List<>>` as in original PR
- `inputFilesToJoin` is expected to have the same rowGroups ordering as in `inputFiles`, number of files in `inputFiles` and `inputFilesToJoin` is not necessarily has be the same, but ordering of rowGroups and the rowCount of paired rowGroups must be the same
- `joinColumnsOverwrite` is used if the `inputFilesToJoin` is expected to overwrite column in `inputFiles`
- all the capabilities that available for `inputFiles` like pruning, nullification, binary copy, now should be available for `inputFilesToJoin` too

##  Post PR action points
- ~~Add a part on the ParqeutRewriter to parquet-java's README.md as noted by @wgtmac in [this comment](https://github.com/apache/parquet-java/pull/1335#discussion_r1706403049)~~ Was addressed by adding ParquetRewriter description in JavaDoc
- Add issue / PR for an existing bug: described https://github.com/apache/parquet-java/pull/1335#issuecomment-2323010561


MaxNevermind:
@wgtmac @ConeyLiu

This is the outcome of simplification I mention in a comment here a couple of weeks ago: https://github.com/apache/parquet-mr/pull/1273#issuecomment-2053772198
I’ve limited the set of capabilities, see the description. 
I’ve tired different ideas and it all come out as having too complex of implementation, so I decided to finalize at least something with as simple implementation as possible.
PR is not yet polished. Just wanted to do a quick overview of the new approach. If it looks good, I will polish it.

wgtmac:
Thanks for your effort! I just took a quick glimpse and it does look simpler than the previous patch.

My general question is that now the prerequisite for users to use the joiner is to run a pre-processing like https://gist.github.com/MaxNevermind/0feaaf380520ca34c2637027ef349a7d you've mentioned. The pre-processing also takes time and resource. Does it mean that we have to deal with unaligned blocks anyway if users do not want to pay for the pre-processing task?

MaxNevermind:
> Does it mean that we have to deal with unaligned blocks anyway if users do not want to pay for the pre-processing task?

This new implementation requires blocks to be aligned yes. The gist snippet preparing it need to be updated btw, that one is for the previous version.

I think this version strikes a good balance in terms of features vs implementation complexity, putting all the features as in previous version leads to a very complex implementation imo which I'm not sure is worth pursuing as it is mainly optimization for a pretty niche use-case, for regular use-cases you can just read write the whole thing, considering a niche use-case it is reasonable to assume for users to go extra mile and run that snippet and prepare the data in required way.

wgtmac:
Yes, I agree that we can start from the implementation with the assumption that row groups of files are aligned. One thing that I'm not sure is that some users may not be easy to generate files to join with same row group alignment and any way require the rewrite tool to handle this.

MaxNevermind:
@wgtmac 
I wanted to ask if you see the value behind this version of PR? 

I found a few other potential use-cases and want to try to finish this. I understand that supporters want to see features that are generic or applicable in most use-case which this PR doesn't. I can potentially will work on extension of this so stitched files doesn't need to have the same row groups(feature from original PR). But I want to go incremental and start from something smaller but useful as this PR.

wgtmac:
Thanks for the update! Yes, I agree that this PR provides sufficient value and I don't object to incremental improvement.

wgtmac:
@MaxNevermind Let me know when this is ready for review. Thanks!

MaxNevermind:
@wgtmac 
I think ParquetRewriter is ready to review. More documentation in RewriteOptions and tests in ParquetRewriterTest are needed, but I hope it's okay if I can add it later while we agree and green-light the implementation first.

ConeyLiu:
Let me take a look these days as well, thanks for the great works

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -350,6 +436,28 @@ public Builder indexCacheStrategy(IndexCache.CacheStrategy cacheStrategy) {
       return this;
     }
 
+    /**
+     * Set a flag weather columns from join files need to overwrite columns from input files.
+     *
+     * @param joinColumnsOverwrite
```
ConeyLiu, line 442: Remove this line if no doc is provided

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -350,6 +436,28 @@ public Builder indexCacheStrategy(IndexCache.CacheStrategy cacheStrategy) {
...
+     */
+    public Builder joinColumnsOverwrite(boolean joinColumnsOverwrite) {
+      this.joinColumnsOverwrite = joinColumnsOverwrite;
+      return this;
+    }
+
+    /**
+     * Set a flag weather metadata from join files should be ignored.
+     *
+     * @param ignoreJoinFilesMetadata
```
ConeyLiu, line 453: same as here

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -350,6 +436,28 @@ public Builder indexCacheStrategy(IndexCache.CacheStrategy cacheStrategy) {
...
+     * @param joinColumnsOverwrite
+     * @return self
+     */
+    public Builder joinColumnsOverwrite(boolean joinColumnsOverwrite) {
+      this.joinColumnsOverwrite = joinColumnsOverwrite;
+      return this;
+    }
+
+    /**
+     * Set a flag weather metadata from join files should be ignored.
```
ConeyLiu, line 451: `weather` -> `whether`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -350,6 +436,28 @@ public Builder indexCacheStrategy(IndexCache.CacheStrategy cacheStrategy) {
       return this;
     }
 
+    /**
+     * Set a flag weather columns from join files need to overwrite columns from input files.
```
ConeyLiu, line 440: `weather` -> `whether`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -358,6 +466,9 @@ public Builder indexCacheStrategy(IndexCache.CacheStrategy cacheStrategy) {
     public RewriteOptions build() {
       Preconditions.checkArgument(inputFiles != null && !inputFiles.isEmpty(), "Input file is required");
       Preconditions.checkArgument(outputFile != null, "Output file is required");
+      Preconditions.checkArgument(
+          inputFiles != null && !inputFiles.isEmpty(),
```
ConeyLiu, line 470: Duplicate check?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -245,6 +283,31 @@ public Builder(Configuration conf, List<Path> inputFiles, Path outputFile) {
...
+    }
+
+    public Builder(Configuration conf, List<Path> inputFiles, List<Path> inputFilesToJoin, Path outputFile) {
+      this.conf = new HadoopParquetConfiguration(conf);
+      this.inputFiles = new ArrayList<>(inputFiles.size());
+      for (Path inputFile : inputFiles) {
+        this.inputFiles.add(HadoopInputFile.fromPathUnchecked(inputFile, conf));
+      }
+      this.inputFilesToJoin = new ArrayList<>(inputFilesToJoin.size());
+      for (Path inputFile : inputFilesToJoin) {
```
ConeyLiu, line 297: Do we support both rewrite and join?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
ConeyLiu: What is the overwrite strategy? If the input file joins file A and joins file B. All those files have a column: col, which has the highest priority?

ConeyLiu: Maybe we should doc it here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -106,48 +109,63 @@ public class ParquetRewriter implements Closeable {
...
-      List<String> paths = new ArrayList<>();
-      getPaths(schema, paths, null);
-      for (String col : pruneColumns) {
-        if (!paths.contains(col)) {
-          LOG.warn("Input column name {} doesn't show up in the schema of file {}", col, reader.getFile());
-        }
+    inputFiles.addAll(getFileReaders(options.getParquetInputFiles(), conf));
+    inputFilesToJoin.addAll(getFileReaders(options.getParquetInputFilesToJoin(), conf));
+    ensureSameSchema(inputFiles);
+    ensureSameSchema(inputFilesToJoin);
```
ConeyLiu, line 127: The `inputFilesToJoin` are the right-side files, right? Why do they need to have the same schema?

MaxNevermind: > The inputFilesToJoin are the right-side files, right? 

Yes

> Why do they need to have the same schema?

If they don't have the same schema, then how do we figure out which columns need to be added? I guess we can have a list of columns from the right provided as a parameter form a user or create a union of all present columns but then we again getting into complicated implementation imo.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: You mean over-writing existing column and joining a new column? If so, yes.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: > What is the overwrite strategy? If the input file joins file A and joins file B. All those files have a column: col, which has the highest priority?

All inputFilesToJoin file are currently expected to have the same schema. Additionally there is a flag `overwriteInputWithJoinColumns` which if set to true will prioritize columns from  inputFilesToJoin instead of inputFiles, it is `false` by default so be default inputFiles columns will be preserved.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: You mean input file duplicate? I did it as it was done for `inputFiles`, should I add such a check for both file input lists?

By the way, this check was supposed to be for `inputFilesToJoin` not `inputFiles`, I fixed it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: I thought we can focus on documentation and the tests at the end after we agree on implementation :-)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -106,48 +109,63 @@ public class ParquetRewriter implements Closeable {
...
+        "Start rewriting {} input file(s) {} to {}",
+        inputFiles.size() + inputFilesToJoin.size(),
+        Stream.concat(options.getParquetInputFiles().stream(), options.getParquetInputFilesToJoin().stream())
+            .collect(Collectors.toList()),
+        out);
+
+    this.outSchema = getSchema();
+    this.outSchema = pruneColumnsInSchema(outSchema, options.getPruneColumns());
+
+    List<TransParquetFileReader> allFiles;
```
wgtmac, line 138: What about adding a separate `mergeExtraMetaData()` method?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -106,48 +109,63 @@ public class ParquetRewriter implements Closeable {
...
+        ORIGINAL_CREATED_BY_KEY,
+        allFiles.stream()
+            .map(x -> x.getFooter().getFileMetaData().getCreatedBy())
+            .collect(Collectors.toSet())
+            .stream()
+            .reduce((a, b) -> a + "\n" + b)
+            .orElse(""));
+    allFiles.forEach(x -> extraMetaData.putAll(x.getFileMetaData().getKeyValueMetaData()));
+
+    if (!inputFilesToJoin.isEmpty()) {
```
wgtmac, line 155: What about adding a `checkRowCountToJoin()` method?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -176,22 +192,40 @@ public ParquetRewriter(RewriteOptions options) throws IOException {
...
+      schemaMain.getFields().forEach(x -> fieldNames.put(x.getName(), x));
+      inputFilesToJoin
+          .peek()
+          .getFooter()
+          .getFileMetaData()
+          .getSchema()
+          .getFields()
+          .forEach(x -> {
+            if (!fieldNames.containsKey(x.getName()) || overwriteInputWithJoinColumns) {
+              fieldNames.put(x.getName(), x);
```
wgtmac, line 210: Please add a log to indicate `overwriteInputWithJoinColumns` takes effect for any column

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -267,100 +282,157 @@ public void close() throws IOException {
...
-        // This column has been pruned.
-        if (descriptor == null) {
-          continue;
-        }
-
-        // If a column is encrypted, we simply throw exception.
-        // Later we can add a feature to trans-encrypt it with different keys
-        if (chunk.isEncrypted()) {
-          throw new IOException("Column " + chunk.getPath().toDotString() + " is already encrypted");
+    TransParquetFileReader readerJoin = inputFilesToJoin.peek();
```
wgtmac, line 285:
```suggestion
    TransParquetFileReader readerToJoin = inputFilesToJoin.peek();
```

wgtmac: Same for below variables.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -267,100 +282,157 @@ public void close() throws IOException {
...
+        indexCache.setBlockMetadata(blockMetaData);
+        Map<ColumnPath, ColumnChunkMetaData> pathToChunk =
+            blockMetaData.getColumns().stream().collect(Collectors.toMap(x -> x.getPath(), x -> x));
+
+        if (readerJoin != null
+            && (blockIdxJoin == -1
+                || ++blockIdxJoin
+                    == readerJoin.getFooter().getBlocks().size())) {
+          blockIdxJoin = 0;
+          readerJoin = inputFilesToJoin.poll();
```
wgtmac, line 311: It seems that close() is not called for the reader.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -267,100 +282,157 @@ public void close() throws IOException {
...
-            Type.Repetition repetition =
-                descriptor.getPrimitiveType().getRepetition();
-            if (repetition.equals(Type.Repetition.REQUIRED)) {
-              throw new IOException("Required column ["
-                  + descriptor.getPrimitiveType().getName() + "] cannot be nullified");
+        for (int outColumnIdx = 0; outColumnIdx < outColumns.size(); outColumnIdx++) {
+          ColumnPath colPath =
+              ColumnPath.get(outColumns.get(outColumnIdx).getPath());
+          if (readerJoin != null) {
+            Optional<ColumnChunkMetaData> chunkJoin =
```
wgtmac, line 329:
```suggestion
            Optional<ColumnChunkMetaData> chunkToJoin =
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -39,38 +39,55 @@
 
 /**
  * A set of options to create a ParquetRewriter.
+ *
+ * TODO find a place where to put a proper description of functionality as it is not trivial:
```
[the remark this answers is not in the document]
MaxNevermind: You mean we have to add a part on rewriter in parquet-java's README.md?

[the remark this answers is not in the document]
wgtmac: Not necessarily in this PR. Just my answer to your TODO here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: Change `Join` to `ToJoin` in all variables names here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -325,6 +388,18 @@ public Builder addInputFile(Path path) {
       return this;
     }
 
+    /** TODO fix documentation after addition of InputFilesToJoin
```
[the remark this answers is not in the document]
MaxNevermind: I wanted to polish the documentation after agreeing on implementation details.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
btw, I think it was present for readers before this PR neither 🤷

MaxNevermind:
@wgtmac @ConeyLiu 
I addressed your questions and suggestions. Check it out.

After we finish with implementation I can polish documentation and tests.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
ConeyLiu: I mean the L467 has the same check:
```java
Preconditions.checkArgument(inputFiles != null && !inputFiles.isEmpty(), "Input file is required");
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
ConeyLiu: I thought this works like: file_1<column a, column b> join file_2<column c>, file_3<column d>, then produce file_joined<column a, column b, column c, column d>

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
MaxNevermind: Oh, yeah it was supposed to be for `inputFilesToJoin` not `inputFiles`. I replaced that check with `(inputFilesToJoin != null ? inputFilesToJoin : new ArrayList<>())` bellow. Hope this works.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: Sorry, I didn't get your example. What is the schema for file_2 and file_3? Are file_2 and file_3 both in  `inputFilesToJoin`?


Current logic is:
- `inputFiles` is required (left side), a schema for these file must be the same for all them as it was before this PR
- `inputFilesToJoin` is optional (right side), a schema for these file must be the same for all them
- a final merged schema is produced in a constructor
- in case of an overlapping column that present on the left and on the right its values are extracted from the left or from the right depending on `overwriteInputWithJoinColumns` parameter

ConeyLiu: Sorry,  I have updated the above example. So for the above example, we need to do with two steps: 
1. file_1 join file_2 and get file_joined_1
2. file_joined_ 1 join file_3 and get the final result

ConeyLiu:
+1, I have not more comments.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: okay, I got it
This functionality is not supported by this PR. It was possible in the original version of joiner but In this PR I wanted to start with the simplest implementation possible. Maybe I can add it later.

wgtmac:
Thanks @ConeyLiu for the review! I think it is good time to add more test cases now. 

BTW, the existing `ParquetRewriterTest` is large enough and takes a lot of time to run. Does it make sense to add another file for the joiner?

MaxNevermind:
@wgtmac 

I started to work on the tests but I can't figure out the current approach to ParquetRewriter testing based on already existing tests.  The whole list of features I see:
- data validity after merging
- single / multiple files merging
- column nullification
- column pruning
- column encryption
- codec preservation
- bloom filter preservation
- page index verification
- metadata(CREATED_BY_KEY) preservation

I'm used to approach when features are unit tested independently sequentially. But looking at existing ParquetRewriter tests I can see that some of test tests for multiple things in the same test and I'm not able to figure out the system behind mixing features to tests into a single test.  

So how I should approach it, should I just target covering all the features in one/two big tests or multiple tests while trying to cover all of those at least in one of those?

wgtmac:
Good question. The `ParquetRewriter` is created by the refactoring work to consolidate `ColumnEncryptor`, `ColumnMasker`, `ColumnPruner` and `CompressionConverter`. You can see individual unit test in the `ColumnEncryptorTest`, `ColumnMaskerTest`, `ColumnPrunerTest` and `CompressionConverterTest` respectively. So the main goal of `ParquetRewriterTest` is to cover the combination of these features. I think we mainly need test cases covering the new join features, and prove that it does not break when other features are turned on.

MaxNevermind:
@wgtmac @ConeyLiu 
Can you check out the tests changes? I created a single big test for the new functionality.  The documentation is still in progress.

Code clarifications.
In couple of places in tests I used anonymous nested functions and a local nested class to try to localize/nest method's tightly related logic into a single block. Let me know if that is too weird / too functional.

Found bug.
I found what looks like a bug in a current version of ParquetRewriter. Probably will fill an issue.
When you try to nullify and encrypt different columns it fails. There is a related test but it nullifies and encrypts the same column which doesn't reproduce a bug. The bug can be reproduced by changing a single line `maskColumns.put("DocId", MaskMode.NULLIFY);` to `maskColumns.put("Links.Forward", MaskMode.NULLIFY);` in testNullifyAndEncryptColumn() method. The reason of a failure as I understand is that during the nullification we create a single column schema `MessageType newSchema = newSchema(schema, descriptor)` and later use our main writer's encryption with that schema but that encyptor is expecting our final target schema, not a single column schema.

wgtmac (commented):
Sorry for the delay. The current test cases are already very complicated so the refactoring work on the validation methods makes sense to me.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -206,39 +207,53 @@ private MessageType getSchema() {
           .getSchema()
           .getFields()
           .forEach(x -> {
-            if (!fieldNames.containsKey(x.getName()) || overwriteInputWithJoinColumns) {
+            if (!fieldNames.containsKey(x.getName())) {
+              LOG.info("Column {} is added to the output from inputFilesToJoin side", x.getName());
```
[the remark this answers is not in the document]

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -18,10 +18,7 @@
  */
 package org.apache.parquet.hadoop.rewrite;
 
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.DOUBLE;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FLOAT;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT64;
+import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.*;
```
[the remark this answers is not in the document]
MaxNevermind: done

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -33,18 +30,17 @@
...
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
MaxNevermind: done

MaxNevermind:
@wgtmac 
I addressed small issues you found and added / tried to polish the documentation, check it out.

wgtmac:
> @wgtmac I addressed small issues you found and added / tried to polish the documentation, check it out.

LGTM. Thanks!

MaxNevermind:
@wgtmac 
Is there anything else I'm expected to do in this PR?

wgtmac:
I'm not sure if @ConeyLiu wants to take another look.

BTW, could you fix the PR title and description? It is no longer a WIP.

ConeyLiu:
+1, I have no further comments, thanks for the great work

wgtmac (approved):

wgtmac:
Thanks @MaxNevermind and @ConeyLiu!

PR #1339, opened 2024-05-01
PARQUET-2465: Fall back to HadoopConfig

Fokko:
We see that this causes the 1.14 to be incompatible with the previous releases. 

- Previously it would just create the Hadoop Configuration: https://github.com/apache/parquet-mr/pull/1141/files#diff-1ad34fccaed5421e95285fbcceaf63d9d55b8460b5fb301b65468f8a15603b5fL356
- Now the ParquetConfig will be created and the `getWriteSupport(conf)` is called: https://github.com/apache/parquet-mr/pull/1141/files#diff-1ad34fccaed5421e95285fbcceaf63d9d55b8460b5fb301b65468f8a15603b5fR784
- But since this method is freshly introduced:
```java
    protected WriteSupport<T> getWriteSupport(ParquetConfiguration conf) {
      throw new UnsupportedOperationException(
          "Override ParquetWriter$Builder#getWriteSupport(ParquetConfiguration)");
    }
```
It is not implemented and causes an `UnsupportedOperationException` if you don't supply a config.
- I think this breaks backward compatibility.

### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references
    - https://issues.apache.org/jira/browse/PARQUET-2465

### Tests

### Commits

### Style

### Documentation

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -495,7 +495,9 @@ public class ParquetWriter<T> implements Closeable {
     /**
      * @param conf a configuration
      * @return an appropriate WriteSupport for the object model.
+     * @deprecated Use {@link #getWriteSupport(ParquetConfiguration)} instead
      */
+    @Deprecated
     protected abstract WriteSupport<T> getWriteSupport(Configuration conf);
 
     /**
@@ -503,8 +505,7 @@ public class ParquetWriter<T> implements Closeable {
      * @return an appropriate WriteSupport for the object model.
      */
     protected WriteSupport<T> getWriteSupport(ParquetConfiguration conf) {
-      throw new UnsupportedOperationException(
-          "Override ParquetWriter$Builder#getWriteSupport(ParquetConfiguration)");
+      return getWriteSupport(ConfigurationUtil.createHadoopConfiguration(conf));
     }
 
     /**
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/InitContext.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/api/InitContext.java
@@ -79,7 +79,9 @@ public class InitContext {
 
   /**
    * @return the configuration for this job
+   * @deprecated Use {@link #getParquetConfiguration()} instead
    */
+  @Deprecated
   public Configuration getConfiguration() {
     return ConfigurationUtil.createHadoopConfiguration(configuration);
   }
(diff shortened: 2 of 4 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -503,8 +503,7 @@ protected Builder(OutputFile path) {
      * @return an appropriate WriteSupport for the object model.
      */
     protected WriteSupport<T> getWriteSupport(ParquetConfiguration conf) {
-      throw new UnsupportedOperationException(
```
Fokko, line 506: When people want to decouple from Hadoop, they can just override the methods, otherwise the old behavior is preserved.

Fokko: Thinking of it, I believe we should deprecate the ones where you need to pass in Configuration (from Hadoop), and then we can move to `ParquetConfiguration` where you can also pass in a `HadoopParquetConfiguration`. Thanks for raising this, let me create a commit and let me know what you think!

Fokko:
Good to go. @wgtmac @shangxinli @amousavigourabi @vinooganesh LMKWYT

vinooganesh:
👍 this looks good to me, but do we want to actually mark the hadoop methods as deprecated if we are going to assume that parquet-mr 1.x will always rely on hadoop? Or is there actually a plan to drop the hadoop dependency on future 1.x releases?

wgtmac (approved):
This fix makes sense to me. Thanks!

Fokko:
> 👍 this looks good to me, but do we want to actually mark the hadoop methods as deprecated if we are going to assume that parquet-mr 1.x will always rely on hadoop? Or is there actually a plan to drop the hadoop dependency on future 1.x releases?

You can still use the Hadoop config, but you'll need to wrap it into a `HadoopParquetConfiguration`: https://github.com/apache/parquet-mr/blob/68609198c4fecaa0e8fb1bcaa2c8a353030de962/parquet-hadoop/src/main/java/org/apache/parquet/conf/HadoopParquetConfiguration.java#L42-L44

amousavigourabi:
> 👍 this looks good to me, but do we want to actually mark the hadoop methods as deprecated if we are going to assume that parquet-mr 1.x will always rely on hadoop? Or is there actually a plan to drop the hadoop dependency on future 1.x releases?

I'd like to note that we have other stuff that is deprecated because they will be dropped in 2.0 without any plans to remove them in 1.x releases as well (see: `org.apache.parquet.avro.AvroParquetReader#builder(Path)` for an example of this), so this is consistent with the usage in the rest of the project.

vinooganesh:
Sounds great, thank @Fokko and @amousavigourabi!

PR #1349, opened 2024-05-10
PARQUET-2468: ParquetMetadata must convert to json

RustedBones:
`ParquetMetadata` json serialization failed on `UNENCRYPTED` files.

Add required configuration to the metadata `ObjectMaper`:
- disable `FAIL_ON_EMPTY_BEANS` to serialize LogicalTypeAnnotation (was removed in #1144)
- add Jdk8Module to serialize `Optional` type

Also add custom serializer for `Statistics`

```json
{
  "fileMetaData" : {
    "schema" : {
      "name" : "test",
      "repetition" : "REPEATED",
      "logicalTypeAnnotation" : null,
      "id" : null,
      "fields" : [ {
        "name" : "some_null_field",
        "repetition" : "OPTIONAL",
        "logicalTypeAnnotation" : null,
        "id" : null,
        "primitive" : true,
        "primitiveTypeName" : "BINARY",
        "typeLength" : 0,
        "decimalMetadata" : null,
        "originalType" : null
      } ],
      "columns" : [ {
        "path" : [ "some_null_field" ],
        "type" : "BINARY",
        "primitiveType" : {
          "name" : "some_null_field",
          "repetition" : "OPTIONAL",
          "logicalTypeAnnotation" : null,
          "id" : null,
          "primitive" : true,
          "primitiveTypeName" : "BINARY",
          "typeLength" : 0,
          "decimalMetadata" : null,
          "originalType" : null
        },
        "maxRepetitionLevel" : 0,
        "maxDefinitionLevel" : 1,
        "typeLength" : 0
      } ],
      "paths" : [ [ "some_null_field" ] ],
      "fieldCount" : 1,
      "primitive" : false,
      "originalType" : null
    },
    "keyValueMetaData" : { },
    "createdBy" : null,
    "fileDecryptor" : null,
    "encryptionType" : "UNENCRYPTED"
  },
  "blocks" : [ {
    "columns" : [ {
      "rowGroupOrdinal" : -1,
      "encodingStats" : null,
      "columnIndexReference" : null,
      "offsetIndexReference" : null,
      "bloomFilterOffset" : -1,
      "bloomFilterLength" : -1,
      "dictionaryPageOffset" : 0,
      "valueCount" : 0,
      "totalSize" : 0,
      "totalUncompressedSize" : 0,
      "statistics" : {
        "min" : "\u0000",
        "max" : "\u0000\u0000",
        "null_count" : 0
      },
      "sizeStatistics" : null,
      "firstDataPageOffset" : 0,
      "primitiveType" : {
        "name" : "fake_type",
        "repetition" : "OPTIONAL",
        "logicalTypeAnnotation" : null,
        "id" : null,
        "primitive" : true,
        "primitiveTypeName" : "BINARY",
        "typeLength" : 0,
        "decimalMetadata" : null,
        "originalType" : null
      },
      "startingPos" : 0,
      "codec" : "GZIP",
      "encodings" : [ ],
      "encrypted" : false,
      "type" : "BINARY",
      "path" : [ "foo" ]
    } ],
    "rowCount" : 0,
    "totalByteSize" : 0,
    "path" : null,
    "ordinal" : 0,
    "rowIndexOffset" : -1,
    "compressedSize" : 0,
    "startingPos" : 0
  } ]
}
```


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java
```diff
@@ -50,19 +60,23 @@ public static String toPrettyJSON(ParquetMetadata parquetMetaData) {
...
-            || parquetMetaData.getFileMetaData().getEncryptionType()
-                == FileMetaData.EncryptionType.UNENCRYPTED) {
-          objectToPrint = parquetMetaData;
-        } else {
-          objectToPrint = parquetMetaData.getFileMetaData();
-        }
-        objectMapper.writerWithDefaultPrettyPrinter().writeValue(stringWriter, objectToPrint);
+        writer = objectMapper.writerWithDefaultPrettyPrinter();
       } else {
-        objectMapper.writeValue(stringWriter, parquetMetaData);
```
RustedBones, line 64: If `!isPrettyPrint`, we would try to serialize encrypted metadata, which is probably no desired.

parquet-hadoop/pom.xml
```diff
@@ -118,6 +118,11 @@
       <artifactId>jackson-core</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+    <dependency>
```
wgtmac, line 121: Perhaps add a comment to explain it is required to serialize Optional type?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java
```diff
@@ -32,6 +35,13 @@ public class ParquetMetadata {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  // Enable FAIL_ON_EMPTY_BEANS on objectmapper. Without this feature parquet-casdacing tests fail,
+  // because LogicalTypeAnnotation implementations are classes without any property.
+  static {
+    objectMapper.disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
+    objectMapper.registerModule(new Jdk8Module());
```
wgtmac, line 42: ditto, add comment to explain why is it required?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java
wgtmac: cc @ggershinsky to see if this is fine.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/StatisticsSerializer.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.parquet.hadoop.metadata;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import java.io.IOException;
+import org.apache.parquet.column.statistics.Statistics;
+
+class StatisticsSerializer extends JsonSerializer<Statistics<?>> {
```
wgtmac, line 28: Should we add a test?

wgtmac: BTW, what is the original behavior?

wgtmac (commented):
Sorry for the late review. I have left some minor comments. Mostly LGTM.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/StatisticsSerializer.java
RustedBones: I had 2nd thought on this. Printing the statistics with the corresponding json primitive is not trivial.
As the `sizeStatistics` are not guaranteed to be there, I change to ignore normal stats in the logging

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -109,6 +110,7 @@ public String getCreatedBy() {
     return createdBy;
   }
 
+  @JsonIgnore
   public InternalFileDecryptor getFileDecryptor() {
     return fileDecryptor;
   }
```
RustedBones, line 116: When logging the file metadata, we probably want to skip this
```json
  "fileDecryptor" : {
    "fileAAD" : null,
    "decryptionProperties" : {
      "footerKey" : "AQIDBAUGBwgJCgsMDQ4PEA==",
      "keyRetriever" : null,
      "aadprefix" : null
    }
  },
  ```

File: parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java (hunk 1 above)
Comment (ggershinsky) [reply parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java:64]: SGTM, thanks

Comment (wgtmac) [reply parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ParquetMetadata.java:64]: Thank you!

File: parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java [hunk 6]
```diff
@@ -380,13 +382,15 @@ public PrimitiveType getPrimitiveType() {
   /**
    * @return the stats for this column
    */
+  @JsonIgnore
```
wgtmac, line 385: Does this ignore printing stats for all columns?

RustedBones: yes, with that annotation, all stats are skipped from logging. That's a behavior change.

I did so to align behavior with the newly added `SizeStatistics`. As the default implementation throws an exception, we can't safely serialize this field so I added `JsonIgnore` annotation there too. (I'm actually wondering why `null` is not returned as default).

If we decide to serialize the `SizeStatistics`, I'm also wondering how we should handle [invalid stats](https://github.com/apache/parquet-mr/blob/f12229addfdca512770a0959d8074100b1e42ee5/parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java#L49).

wgtmac: Although it is used in the logging, I'm not sure if ignoring the stats silently will confuse/break developers used to depend on that log to debug the stats.

@gszadovszky @Fokko Do you have any suggestion?

Fokko: I think we're good here. I'm not sure if anyone uses the stats for debugging since it is super verbose.

Fokko (approved):
Thanks for working on this @RustedBones, this makes sense to me 👍

wgtmac:
Thanks for the reply! @Fokko 

It seems that we need a minor release for this. +1 on my side.

wgtmac (approved):

PR #1356, opened 2024-05-21
Update readme to point to 1.14.0 release

vinooganesh:
CC @wgtmac.


wgtmac (approved):
Good catch!

Perhaps we also need to rename `parquet-mr` to `parquet-java` in those links.

PR #1370, opened 2024-06-04
PARQUET-34: Extend Contains support to all ColumnFilterPredicate types

clairemcginty:
Follows up on [this discussion](https://github.com/apache/parquet-java/pull/1328#discussion_r1608679199) from #1328 to fast-follow by expanding `Contains` predicate support from only `Eq<T>` to all `ColumnFilterPredicate<T>` types.

### Jira

    - https://issues.apache.org/jira/browse/PARQUET-34

### Tests

### Commits

### Style

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
@@ -85,8 +85,6 @@ public final class Operators {
   public static interface SupportsLtGt
       extends SupportsEqNotEq {} // marker for columns that can be used with lt(), ltEq(), gt(), gtEq()
 
-  public static interface SupportsContains {}
-
   public static final class IntColumn extends Column<Integer> implements SupportsLtGt {
     IntColumn(ColumnPath columnPath) {
       super(columnPath, Integer.class);
@@ -123,8 +121,13 @@ public final class Operators {
     }
   }
 
+  abstract static class SingleColumnFilterPredicate<T extends Comparable<T>>
+      implements FilterPredicate, Serializable {
+    abstract Column<T> getColumn();
+  }
+
   // base class for Eq, NotEq, Lt, Gt, LtEq, GtEq
-  abstract static class ColumnFilterPredicate<T extends Comparable<T>> implements FilterPredicate, Serializable {
+  abstract static class ColumnFilterPredicate<T extends Comparable<T>> extends SingleColumnFilterPredicate<T> {
     private final Column<T> column;
     private final T value;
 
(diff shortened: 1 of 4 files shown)


parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
```diff
@@ -299,22 +304,23 @@ private void addInNotInCase(TypeInfo info, boolean isEq) throws IOException {
...
     add("            for (" + info.primitiveName + " i : target) {\n");
 
     add("              if(" + compareEquality("value", "i", isEq) + ") {\n");
 
-    add("                 setResult(true);\n");
-
-    add("                 set = true;\n");
-
-    add("                 break;\n");
+    add("                 setResult(true);\n                 return;\n");
```
clairemcginty, line 316: I simplified the existing ValueInspector logic a bit here. Previously it was:

```java
  @Override
  public void update(int value) {
	boolean set = false;
	for (int i : target) {
	  if(comparator.compare(value, i) != 0) {
		 setResult(true);
		 set = true;
		 break;
	   }
	 }
	 if (!set) setResult(false);
   }
 };
}
```

Now it's:
```java
@Override
public void update(int value) {
for (int i : target) {
  if(comparator.compare(value, i) == 0 ) {
	 setResult(true);
	 return;
   }
 }
 setResult(false);
}
```

wgtmac (approved):
+1

Sorry for the delay. cc @gszadovszky

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -415,14 +420,18 @@ public int hashCode() {
...
 
     ContainsColumnPredicate(U underlying) {
       super(underlying.getColumn());
-      if (underlying.getValue() == null) {
-        throw new IllegalArgumentException("Contains predicate does not support null element value");
+      if ((underlying instanceof ColumnFilterPredicate && ((ColumnFilterPredicate) underlying).getValue() == null)
+          || underlying instanceof SetColumnFilterPredicate
+              && ((SetColumnFilterPredicate) underlying)
+                  .getValues()
+                  .contains(null)) {
```
gszadovszky, line 433: nit: I know it means the same logically, but I think it is more readable to wrap in a parenthesis since the `instanceof` and the related cast are tightly related.

gszadovszky (approved):
Looks good to me! I have just one nit comment.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
clairemcginty: makes sense! updated

gszadovszky (approved):

PR #1379, opened 2024-06-19
PARQUET-2471: Add support for geometry logical type

wgtmac:
This is the PoC of https://github.com/apache/parquet-format/pull/240

### Jira

- [ ] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references

### Tests

### Commits

### Style

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/GeometryStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/GeometryStatistics.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.statistics;
+
+import org.apache.parquet.Preconditions;
+import org.apache.parquet.column.statistics.geometry.BoundingBox;
+import org.apache.parquet.column.statistics.geometry.Covering;
+import org.apache.parquet.column.statistics.geometry.EnvelopeCovering;
+import org.apache.parquet.column.statistics.geometry.GeometryTypes;
+import org.apache.parquet.io.api.Binary;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.io.ParseException;
+import org.locationtech.jts.io.WKBReader;
+
+public class GeometryStatistics {
+
+  private final BoundingBox boundingBox;
+  private final Covering covering;
+  private final GeometryTypes geometryTypes;
+  private final WKBReader reader = new WKBReader();
+
+  public GeometryStatistics(BoundingBox boundingBox, Covering covering, GeometryTypes geometryTypes) {
+    this.boundingBox = boundingBox;
+    this.covering = covering;
+    this.geometryTypes = geometryTypes;
+  }
+
+  public GeometryStatistics() {
+    this(new BoundingBox(), new EnvelopeCovering(), new GeometryTypes());
+  }
+
+  public BoundingBox getBoundingBox() {
+    return boundingBox;
+  }
+
+  public Covering getCovering() {
+    return covering;
+  }
+
+  public GeometryTypes getGeometryTypes() {
+    return geometryTypes;
+  }
+
+  public void update(Binary value) {
+    if (value == null) {
+      return;
+    }
+    try {
+      Geometry geom = reader.read(value.getBytes());
+      update(geom);
+    } catch (ParseException e) {
+      abort();
+    }
+  }
+
+  public void update(Geometry geom) {
+    boundingBox.update(geom);
+    covering.update(geom);
+    geometryTypes.update(geom);
+  }
+
+  public void merge(GeometryStatistics other) {
+    Preconditions.checkArgument(other != null, "Cannot merge with null GeometryStatistics");
+    boundingBox.merge(other.boundingBox);
+    covering.merge(other.covering);
+    geometryTypes.merge(other.geometryTypes);
+  }
+
+  public void reset() {
+    boundingBox.reset();
+    covering.reset();
+    geometryTypes.reset();
+  }
+
+  public void abort() {
+    boundingBox.abort();
+    covering.abort();
+    geometryTypes.abort();
+  }
+
+  public GeometryStatistics copy() {
+    return new GeometryStatistics(boundingBox.copy(), covering.copy(), geometryTypes.copy());
+  }
+
+  @Override
+  public String toString() {
+    return "GeometryStatistics{" + "boundingBox="
+        + boundingBox + ", covering="
+        + covering + ", geometryTypes="
+        + geometryTypes + '}';
+  }
+}
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/BoundingBox.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/BoundingBox.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.column.statistics.geometry;
+
+import org.apache.parquet.Preconditions;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Geometry;
+
+public class BoundingBox {
+
+  private double xMin = Double.MAX_VALUE;
+  private double xMax = Double.MIN_VALUE;
+  private double yMin = Double.MAX_VALUE;
+  private double yMax = Double.MIN_VALUE;
+  private double zMin = Double.MAX_VALUE;
+  private double zMax = Double.MIN_VALUE;
+  private double mMin = Double.MAX_VALUE;
+  private double mMax = Double.MIN_VALUE;
+
+  public BoundingBox(
+      double xMin, double xMax, double yMin, double yMax, double zMin, double zMax, double mMin, double mMax) {
+    this.xMin = xMin;
+    this.xMax = xMax;
+    this.yMin = yMin;
+    this.yMax = yMax;
+    this.zMin = zMin;
+    this.zMax = zMax;
+    this.mMin = mMin;
+    this.mMax = mMax;
+  }
+
+  public BoundingBox() {}
+
+  public double getXMin() {
+    return xMin;
+  }
+
+  public double getXMax() {
+    return xMax;
+  }
+
+  public double getYMin() {
+    return yMin;
+  }
+
+  public double getYMax() {
+    return yMax;
+  }
+
+  public double getZMin() {
+    return zMin;
+  }
+
+  public double getZMax() {
+    return zMax;
+  }
+
+  public double getMMin() {
+    return mMin;
+  }
+
+  public double getMMax() {
+    return mMax;
+  }
+
+  public void update(Geometry geom) {
+    if (geom == null || geom.isEmpty()) {
+      return;
+    }
+    Coordinate[] coordinates = geom.getCoordinates();
+    for (Coordinate coordinate : coordinates) {
+      update(coordinate.getX(), coordinate.getY(), coordinate.getZ(), coordinate.getM());
+    }
+  }
+
+  public void update(double x, double y, double z, double m) {
+    xMin = Math.min(xMin, x);
+    xMax = Math.max(xMax, x);
+    yMin = Math.min(yMin, y);
+    yMax = Math.max(yMax, y);
+    zMin = Math.min(zMin, z);
+    zMax = Math.max(zMax, z);
+    mMin = Math.min(mMin, m);
+    mMax = Math.max(mMax, m);
+  }
+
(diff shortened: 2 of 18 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestGeometryTypeRoundTrip.java
```diff
@@ -0,0 +1,114 @@
...
+      System.out.println("Statistics");
+      System.out.println(columnChunkMetaData.getStatistics().toString());
+      System.out.println();
+
+      GeometryStatistics geometryStatistics =
+          ((BinaryStatistics) columnChunkMetaData.getStatistics()).getGeometryStatistics();
+      Assert.assertNotNull(geometryStatistics);
+      System.out.println("GeometryStatistics");
+      System.out.println(geometryStatistics);
+      System.out.println();
```
wgtmac, line 111: I have just finished the PoC to read/write geometry values in the form of WKB and verify the `GeometryStatistics` have been collected as expected. This is the output of this minimal test.

```
Footer
ParquetMetaData{FileMetaData{schema: message msg {
  required binary col_geom (GEOMETRY(WKB,PLANAR,EPSG:4326));
}
, metadata: {writer.model.name=example}}, blocks: [BlockMetaData{2, 73, rowIndexOffset = 0 [ColumnMetaData{UNCOMPRESSED [col_geom] required binary col_geom (GEOMETRY(WKB,PLANAR,EPSG:4326))  [BIT_PACKED, PLAIN], 4}]}]}

Statistics
min: POINT (1 1), max: POINT (2 2), num_nulls: 0

GeometryStatistics
GeometryStatistics{boundingBox=BoundingBox{xMin=1.0, xMax=2.0, yMin=1.0, yMax=2.0, zMin=NaN, zMax=NaN, mMin=NaN, mMax=NaN}, covering=Covering{geometry=POLYGON ((1 1, 1 2, 2 2, 2 1, 1 1)), edges=PLANAR}, geometryTypes=GeometryTypes{types=[1]}}
```

@szehon-ho @jiayuasu @paleolimbot @emkornfield 

Please let me know what do you think. Thanks!

wgtmac: Could you help take an initial review to see if this looks promising? @gszadovszky @julienledem

parquet-column/src/main/java/org/apache/parquet/column/statistics/GeometryStatistics.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.parquet.column.statistics.geometry.EnvelopeCovering;
+import org.apache.parquet.column.statistics.geometry.GeometryTypes;
+import org.apache.parquet.io.api.Binary;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.io.ParseException;
+import org.locationtech.jts.io.WKBReader;
+
+public class GeometryStatistics {
+
+  private final BoundingBox boundingBox;
```
wgtmac, line 33: This is something that I can hardly find a way to fit into an `extension` type. Please let me know if you have any better idea. @pitrou

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestGeometryTypeRoundTrip.java
gszadovszky: @wgtmac, I am not really familiar with this new geometry logical type. So, my comments are more general.

I can see a TODO for the comparator but I cannot see any specification in the format PR about the total ordering. (BTW, I cannot see a proper spec in LogicalTypes.md. It is for the other PR, though.)

Could you please double-check the licensing of `org.locationtech.jts` whether we are able to distribute `parquet-java` as is? I am not expert of this, though.

I think, it is not a good idea to include external objects (`org.locationtech.jts.geom.Geometry`) into our public API. In our public API we should stick to the primitive type (`Binary`).

Otherwise, it looks good to me.

wgtmac: Thanks @gszadovszky for a quick review! Please see my inline comments below.

> I can see a TODO for the comparator but I cannot see any specification in the format PR about the total ordering.

I thought we'd better make the `ColumnOrder` of geometry logical type as undefined and do not populate min/max values. Here is the related line in the format PR: https://github.com/apache/parquet-format/pull/240/files#diff-834c5a8d91719350b20995ad99d1cb6d8d68332b9ac35694f40e375bdb2d3e7cR1083


> Could you please double-check the licensing of org.locationtech.jts whether we are able to distribute parquet-java as is? 

It has dual licenses:
- [Eclipse Public License 2.0](https://www.eclipse.org/legal/epl-v20.html)
- [Eclipse Distribution License 1.0](http://www.eclipse.org/org/documents/edl-v10.php) (a BSD Style License). 

I have also checked Apache Sedona which includes `org.locationtech.jts/jts-core` in its dependency: https://github.com/apache/sedona/blob/119682321f5301e6ef776315a8f2cd53a8ff63b0/pom.xml#L168-L188. So I assume we are safe here. 

@jiayuasu may provide more context as the domain expert. 

> I think, it is not a good idea to include external objects (org.locationtech.jts.geom.Geometry) into our public API. In our public API we should stick to the primitive type (Binary).

I agree that it is not a good idea to expose `org.locationtech.jts.geom.Geometry` as a public API. Let me switch it to internal method for now. If downstream projects like Apache Iceberg (which has customized column writer implementation) could benefit from directly passing a `org.locationtech.jts.geom.Geometry` to build stats without unnecessary conversion, we can revisit this later.

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/EnvelopeCovering.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Override
+  void update(Geometry geom) {
+    if (geometry == null) {
+      return;
+    }
+    try {
+      if (geometry != EMPTY) {
+        Geometry envelope = reader.read(geometry.array());
+        geometry = ByteBuffer.wrap(writer.write(envelope.union(geom).getEnvelope()));
```
paleolimbot, line 46: I am a little unclear as to the utility of the `EnvelopeCovering`...it seems like the optimal strategy would be to always merge `BoundingBox`es and build a geometry at the end. A union in a loop is not usually efficient and should perhaps be more difficult to accidentally do.

paleolimbot (commented):
I took a look through for geometry things...this is very cool!

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestGeometryTypeRoundTrip.java
paleolimbot: > Please let me know what do you think

This looks good to me! (I left some geometry-specific notes on the implementation)

> I have also checked Apache Sedona which includes org.locationtech.jts/jts-core in its dependency

I believe Calcite also uses JTS for its spatial component and it may be worth checking there to see to what extent those types are used in various APIs and/or if it's ever caused any issues.

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/EnvelopeCovering.java
wgtmac: I agree. Perhaps I can call `Envelope.expandToInclude(Envelope)` to maintain a running envelope.

https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/Envelope.html#expandToInclude-org.locationtech.jts.geom.Envelope-

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -765,9 +812,51 @@ public static Statistics toParquetStatistics(
...
+
+    if (stats.getCovering().getGeometry() != null) {
+      Covering formatCovering = new Covering();
+      formatCovering.setGeometry(stats.getCovering().getGeometry());
+      formatCovering.setEdges(convertEdges(stats.getCovering().getEdges()));
+      formatStats.setCovering(formatCovering);
+    }
+
+    formatStats.setGeometry_types(
+        new ArrayList<Integer>(stats.getGeometryTypes().getTypes()));
```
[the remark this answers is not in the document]
wgtmac: Make sense. Fixed.

wgtmac:
Thanks @paleolimbot for the quick review! I've addressed your feedback and also added page index support for geometry stats. BTW, it is a convention that we need PoC on another Parquet implementation to move forward the proposal. I plan to add similar things to parquet-cpp. Do you happen to know any JTS-parity on the C++ side? Or do you have a plan to implement the PoC? I guess there should be a lot of existing code on the GeoParquet side.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestGeometryTypeRoundTrip.java
jiayuasu: I am pretty confident that JTS is fully compatible with ASF license: https://www.apache.org/legal/resolved.html

This is also why Sedona added it years ago

jiayuasu:
@wgtmac The JTS-parity on C++ side is GEOS, which is a C++ port of JTS: https://github.com/libgeos/geos

Note that: GEOS is LGPL 2.1, which is NOT compatible with ASF

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/BoundingBox.java
```diff
@@ -0,0 +1,153 @@
...
+
+  public double getMMin() {
+    return mMin;
+  }
+
+  public double getMMax() {
+    return mMax;
+  }
+
+  void update(Geometry geom) {
```
jiayuasu, line 82: This is not the correct way to update Geometry in JTS. You need to use CoordinateSequenceFilter. Sedona has many examples here: https://github.com/apache/sedona/blob/0a1db3d35fd7b5721301aa3ce5d39aa8cd828661/common/src/main/java/org/apache/sedona/common/Functions.java#L294

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/BoundingBox.java
```diff
@@ -0,0 +1,153 @@
...
+package org.apache.parquet.column.statistics.geometry;
+
+import org.apache.parquet.Preconditions;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Geometry;
+
+public class BoundingBox {
+
+  private double xMin = Double.MAX_VALUE;
+  private double xMax = Double.MIN_VALUE;
```
jiayuasu, line 28: For the init value of xMax, yMax, ..., please use `-Double.MAX_VALUE` because Double.MIN_VALUE is not guaranteed to be negative. See https://stackoverflow.com/questions/3884793/why-is-double-min-value-in-not-negative

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/EnvelopeCovering.java
```diff
@@ -0,0 +1,97 @@
...
+        Envelope envelope = reader.read(geometry.array()).getEnvelopeInternal();
+        envelope.expandToInclude(geom.getEnvelopeInternal());
+        Geometry polygon = factory.createPolygon(new Coordinate[] {
+          new Coordinate(envelope.getMinX(), envelope.getMinY()),
+          new Coordinate(envelope.getMinX(), envelope.getMaxY()),
+          new Coordinate(envelope.getMaxX(), envelope.getMaxY()),
+          new Coordinate(envelope.getMaxX(), envelope.getMinY()),
+          new Coordinate(envelope.getMinX(), envelope.getMinY())
+        });
+        geometry = ByteBuffer.wrap(writer.write(polygon));
```
jiayuasu, line 58: Pay attention to a trap here: JTS WKB writer has a bug that does not write M value. There is a PR on JTS to fix this but it hasn't been merged: https://github.com/locationtech/jts/pull/734.

I think we also need to be a bit more explicit here about which WKB standard we are supporting.

Standard WKB format does not allow Z and M dimension. ISO WKB allow Z and M dimension. JTS follows ISO WKB but its writer has the aforementioned bug. See this article from GEOS: https://libgeos.org/specifications/wkb/

jiayuasu (changes requested):

jiayuasu:
Core sedona contributors @Kontinuation and @zhangfengcdt , can you also review this?

@wgtmac Thanks for putting together this PR. I guess you might need some help from the Sedona community regarding the Java geometry implementation together with JTS. We can work together to implement this POC.

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/EnvelopeCovering.java
wgtmac: Thanks for the advise! BTW, it seems that JTS envelope has only X and Y axises. I'm not sure if this is another issue here.

wgtmac:
Thanks @jiayuasu for your valuable feedback! Yes, it would be great if we can get direct help from geospatial experts!

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1091,6 +1112,104 @@ public int hashCode() {
...
+      sb.append(",");
+      sb.append(edges);
+      if (crs != null && !crs.isEmpty()) {
+        sb.append(",");
+        sb.append(crs);
+      }
+      if (metadata != null) {
+        sb.append(",");
+        sb.append(metadata);
+      }
```
Kontinuation, line 1169: There's a high probability that `crs` itself contains comma, so this may introduce ambiguity to the generated type parameters.

`metadata` will be appended as something like `java.nio.HeapByteBuffer[pos=_ lim=_ cap=_]` to the type parameter. Does this have any bad implications?

parquet-column/src/main/java/org/apache/parquet/column/statistics/geometry/EnvelopeCovering.java
```diff
@@ -0,0 +1,97 @@
...
+      if (geometry != EMPTY) {
+        Envelope envelope = reader.read(geometry.array()).getEnvelopeInternal();
+        envelope.expandToInclude(geom.getEnvelopeInternal());
+        Geometry polygon = factory.createPolygon(new Coordinate[] {
+          new Coordinate(envelope.getMinX(), envelope.getMinY()),
+          new Coordinate(envelope.getMinX(), envelope.getMaxY()),
+          new Coordinate(envelope.getMaxX(), envelope.getMaxY()),
+          new Coordinate(envelope.getMaxX(), envelope.getMinY()),
+          new Coordinate(envelope.getMinX(), envelope.getMinY())
+        });
```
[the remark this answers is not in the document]
wgtmac: `EnvelopeCovering` is just a possible implementation of the `Covering`. The idea behind putting `Covering` and `BoundingBox` at the same time originates from the suggestion of https://github.com/apache/parquet-format/pull/240#discussion_r1615980422 and https://github.com/apache/parquet-format/pull/240#discussion_r1607000607

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
wgtmac: This is definitely something that I need to improve :)

wgtmac:
Close this in favor of https://github.com/apache/parquet-java/pull/2971

Issue #2930, opened 2024-06-24
Introduce issue templates


rok:
Now that we moved issue tracking to GitHub we should introduce issue templates to enforce some structure.

PR #2931, opened 2024-06-24
GH-2930: Add issue templates

rok:
This introduces [Arrow-like issue templates](https://github.com/apache/arrow/tree/main/.github/ISSUE_TEMPLATE) to structure and categorize issues somewhat as they are reported.

See #2930

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -33,8 +33,8 @@ body:
       label: Component(s)
       multiple: true
       options:
-        - Release
-        - Continuous Integration
+        - Core
+        - Build
         - Arrow
         - Avro
         - Pig
@@ -43,6 +43,5 @@ body:
         - Thrift
         - CLI
         - Benchmarking
-        - Other
     validations:
-      required: true
+      required: false
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -42,6 +42,6 @@ body:
         - Scala
         - Thrift
         - CLI
-        - Benchmarking
+        - Benchmark
     validations:
       required: false
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,44 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Bug Report
+description: File a bug report
+labels: ["Type: bug"]
+assignees: []
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Describe the bug, including details regarding any error messages, version, and platform.
+      description: Please include what you expected.
+    validations:
+      required: true
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component(s)
+      multiple: true
+      options:
+        - Java
+        - Release
+        - Benchmarking
+        - Documentation
+        - Continuous Integration
+        - Packaging
+        - Other
+    validations:
+      required: true
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -33,12 +33,16 @@ body:
       label: Component(s)
       multiple: true
       options:
-        - Java
         - Release
-        - Benchmarking
-        - Documentation
         - Continuous Integration
-        - Packaging
+        - Arrow
+        - Avro
+        - Pig
+        - Protobuf
+        - Scala
+        - Thrift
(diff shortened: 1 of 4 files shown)


rok:
@wgtmac

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,44 @@
...
+      label: Describe the bug, including details regarding any error messages, version, and platform.
+      description: Please include what you expected.
+    validations:
+      required: true
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component(s)
+      multiple: true
+      options:
```
wgtmac, line 35: Do we actually need these `options`? Or should it be consistent with modules like `parquet-avro`, `parquet-cli`, etc.? Same question for other yamls.

wgtmac (commented):
Thanks for adding this!

cc @gszadovszky @julienledem @Fokko

.github/ISSUE_TEMPLATE/bug_report.yaml
rok: Also please note I introduced as many new labels as I could during migration. We now might want to reduce the current set by deleting them [here](https://github.com/apache/parquet-java/labels) (delete is irreversible). Same goes for [milestones](https://github.com/apache/parquet-java/milestones).

rok: Thanks for the feedback @gszadovszky. I've change the list like so:
```diff
      options:
-       - Java
        - Release
-       - Documentation
        - Continuous Integration
-       - Packaging
+       - Arrow
+       - Avro
+       - Pig
+       - Protobuf
+       - Scala
+       - Thrift
+       - CLI
        - Benchmarking
        - Other
```

> I think it makes more sense to somehow sum up Release+CI+Packaging.

Agreed, but I don't have a good suggestion how. For now I just removed packaging.

gszadovszky: Thank you, @rok!

It would be nice to see somehow how many issues we have related to `CI` and `Release`. Maybe `Other` is enough but not sure. I am fine as is, though.

One thing is clearly missing. This is the core functionality that most PRs have updates in. These are the modules `parquet-column`, `parquet-common`, `parquet-encoding`, `parquet-format-structures`, `parquet-generator`, `parquet-hadoop`. I am fine calling it `Core` if there are no better suggestion.

`parquet-hadoop` is definitely part of core currently, even though we want to get rid of the hadoop dependencies. After this work would be done, the separate component `Hadoop` would make sense, but currently I don't think we need it.

rok: To see distribution of component in Jira you can check the current (imported) label counts [here](https://github.com/apache/parquet-java/labels) - these are based on jira components, so CI and Release wont be present. For a quick idea of how much CI issues there are I made a quick dump of titles and bodies of issues: [issues_dump.csv](https://github.com/user-attachments/files/15985203/issues_dump.csv)

gszadovszky: Some quick searches in the csv suggests we have enough release and CI related issues to have them separated. So, I'm fine with the list as is, only the `Core` label is missing.

Fokko: Thanks @rok for working on this 🙌  I would be inclined to remove `Other` and make it optional. WDYT?

rok: Thanks for the feedback! :D I've pushed a change and here's the current list:
```markdown
- Core
- Release
- Continuous Integration
- Arrow
- Avro
- Pig
- Protobuf
- Scala
- Thrift
- CLI
- Benchmarking
```
I'm glad to defer to your suggestions as I'm quite unfamiliar with `parquet-java` (except for the migration).

Fokko: One more thing, I was looking at Iceberg regarding @gszadovszky's comment.

> I think it makes more sense to somehow sum up Release+CI+Packaging.

At Iceberg we've packaged these into a `Build` label.

rok: Seems like a good idea. Will do.

rok: Done.

Fokko (approved):
LGTM, thanks @rok for working on this!

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - Core
+        - Build
+        - Arrow
+        - Avro
+        - Pig
+        - Protobuf
+        - Scala
+        - Thrift
+        - CLI
+        - Benchmarking
```
wgtmac, line 45:
```suggestion
        - Benchmark
```

I would prefer `benchmark` here.

wgtmac (approved):
+1. Thanks @rok!

.github/ISSUE_TEMPLATE/bug_report.yaml
rok: Sure!

gszadovszky (approved):
Thanks, @rok!

wgtmac:
I will merge this after next 24 hours if no objection.

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - Core
+        - Build
+        - Arrow
+        - Avro
+        - Pig
+        - Protobuf
+        - Scala
+        - Thrift
+        - CLI
+        - Benchmarking
```
wgtmac, line 48:
```suggestion
        - Benchmark
```

.github/ISSUE_TEMPLATE/usage_question.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - Core
+        - Build
+        - Arrow
+        - Avro
+        - Pig
+        - Protobuf
+        - Scala
+        - Thrift
+        - CLI
+        - Benchmarking
```
wgtmac, line 68:
```suggestion
        - Benchmark
```

PR #1234, opened 2023-12-07
PARQUET-2410: Use row count instead of value count to get row count from OffsetIndex

ConeyLiu:
Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Parquet Jira](https://issues.apache.org/jira/browse/PARQUET/) issues and references
  them in the PR title. For example, "PARQUET-1234: My Parquet PR"
    - https://issues.apache.org/jira/browse/PARQUET-XXX
    - In case you are adding a dependency, check if the license complies with
      the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines
  from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
    1. Subject is separated from body by a blank line
    1. Subject is limited to 50 characters (not including Jira issue reference)
    1. Subject does not end with a period
    1. Subject uses the imperative mood ("add", not "adding")
    1. Body wraps at 72 characters
    1. Body explains "what" and "why", not "how"

### Style
- [x] My contribution adheres to the code style guidelines and Spotless passes.
    - To apply the necessary changes, run `mvn spotless:apply -Pvector-plugins`

### Documentation

- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain Javadoc that explain what it does

We should use row count instead of value count to get row count from OffsetIndex in rewriter. Even this only influences the last row count and does not influence the result. It would be better to correct it.

PR #2941, opened 2024-07-04
PARQUET-34: implement not() for Contains predicate

clairemcginty:
This should be the final subtask for the Contains() predicate work. It adds support for logical inversion -- `not(contains(...))`.

Also, it fixes a small bug with the composition of mixed Contains predicates.

### Issue

- [x] My PR addresses the following [GitHub issues](https://github.com/apache/parquet-java/issues) and references
  them in the PR title. For example, "GH-1234: My Parquet PR"
    - https://issues.apache.org/jira/browse/PARQUET-34 (should I make a new Github issue for this?)

### Tests

### Commits

- [x] My commits all reference GitHub issues in their subject lines. In addition, my commits follow the guidelines
    1. Subject is limited to 50 characters (not including GitHub issue reference)

### Style

### Documentation

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ContainsRewriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ContainsRewriter.java
@@ -107,7 +107,7 @@ public final class ContainsRewriter implements Visitor<FilterPredicate> {
     } else if (and.getLeft() instanceof Contains) {
       left = and.getLeft();
     } else {
-      return and;
+      left = and.getLeft();
     }
 
     final FilterPredicate right;
@@ -118,15 +118,18 @@ public final class ContainsRewriter implements Visitor<FilterPredicate> {
     } else if (and.getRight() instanceof Contains) {
       right = and.getRight();
     } else {
-      return and;
+      right = and.getRight();
     }
 
-    if (left instanceof Contains) {
-      if (!(right instanceof Contains)) {
-        throw new UnsupportedOperationException(
-            "Contains predicates cannot be composed with non-Contains predicates");
-      }
+    // If two Contains predicates refer to the same column, optimize by combining into a single predicate
+    if ((left instanceof Contains && right instanceof Contains)
+        && ((Contains<?>) left)
+            .getColumn()
+            .getColumnPath()
+            .equals(((Contains<?>) right).getColumn().getColumnPath())) {
       return ((Contains) left).and(right);
+    } else if (left != and.getLeft() || right != and.getRight()) {
+      return new And(left, right);
     } else {
       return and;
     }
@@ -142,7 +145,7 @@ public final class ContainsRewriter implements Visitor<FilterPredicate> {
     } else if (or.getLeft() instanceof Contains) {
       left = or.getLeft();
     } else {
-      return or;
+      left = or.getLeft();
     }
 
     final FilterPredicate right;
@@ -153,15 +156,18 @@ public final class ContainsRewriter implements Visitor<FilterPredicate> {
     } else if (or.getRight() instanceof Contains) {
       right = or.getRight();
     } else {
-      return or;
+      right = or.getRight();
     }
 
-    if (left instanceof Contains) {
-      if (!(right instanceof Contains)) {
-        throw new UnsupportedOperationException(
-            "Contains predicates cannot be composed with non-Contains predicates");
-      }
+    // If two Contains predicates refer to the same column, optimize by combining into a single predicate
+    if ((left instanceof Contains && right instanceof Contains)
+        && ((Contains<?>) left)
+            .getColumn()
+            .getColumnPath()
+            .equals(((Contains<?>) right).getColumn().getColumnPath())) {
       return ((Contains) left).or(right);
+    } else if (left != or.getLeft() || right != or.getRight()) {
+      return new Or(left, right);
     } else {
       return or;
     }
--- a/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
+++ b/parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
@@ -324,16 +324,12 @@ public final class Operators {
     }
   }
 
-  static class DoesNotContain<T extends Comparable<T>> extends Contains<T> {
+  private static class DoesNotContain<T extends Comparable<T>> extends Contains<T> {
     Contains<T> underlying;
 
     protected DoesNotContain(Contains<T> underlying) {
       super(underlying.getColumn());
       this.underlying = underlying;
-    }
-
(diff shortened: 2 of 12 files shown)


parquet-column/src/main/java/org/apache/parquet/filter2/predicate/ContainsRewriter.java
```diff
@@ -118,15 +118,18 @@ public FilterPredicate visit(And and) {
...
-      return and;
+      right = and.getRight();
     }
 
-    if (left instanceof Contains) {
-      if (!(right instanceof Contains)) {
-        throw new UnsupportedOperationException(
-            "Contains predicates cannot be composed with non-Contains predicates");
-      }
+    // If two Contains predicates refer to the same column, optimize by combining into a single predicate
```
clairemcginty, line 124: This fixes a bug with the way Contains predicates were composed -- previously it would throw an error if you tried to do something like `and(contains(binaryColumn("foo", ...), contains(binaryColumn("bar"), ...))` because the columns didn't match. however, that is still a valid predicate -- it's just that they should not be attempted to be combined into a single `Contains` predicate.

parquet-column/src/main/java/org/apache/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
```diff
@@ -145,6 +146,83 @@ public boolean accept(Visitor visitor) {
     }
   }
 
+  /**
+   * A ValueInspector implementation that keeps state for one or more delegate inspectors.
+   */
+  abstract static class DelegatingValueInspector extends ValueInspector {
```
clairemcginty, line 152: I created this because I thought the generated `IncrementallyUpdatedFilterPredicateBuilder` was getting complex in terms of LOC -- all the `Contains` ValueInspectors were overriding `public void update(int|float|binary|double|... value)` as well as `reset()`, and it was getting hard to reason about. However.... this is probably a bit less efficient, since the delegate ValueInspectors are accessed via iterator loops rather than directly invoking `left`/`right`. Let me know what you think, I'm happy to change it back to the way it was

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -371,7 +371,11 @@ public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(NotIn<T> notIn) {
 
     @Override
     public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(Contains<T> contains) {
-      return contains.filter(this, IndexIterator::intersection, IndexIterator::union);
+      return contains.filter(
```
clairemcginty, line 374: My initial idea was to implement `IndexIterator::subtract` for `not(contains())` -- for example, `not(contains(eq("bar")))` would return something like `IndexIterator.all(getPageCount()).subtract(visit(eq("bar"))`. But we can't rely on the individual predicate implementations to return exact row ranges -- for example, `NotIn` [always returns IndexIterator.all](https://github.com/apache/parquet-java/blob/c26624fd8fc196f3d65edb0bcb0460a03cd163fd/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java#L369). So I don't think it's possible to safely subtract any row ranges here.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -214,7 +214,7 @@ public <T extends Comparable<T>> Boolean visit(NotIn<T> notIn) {
 
   @Override
   public <T extends Comparable<T>> Boolean visit(Contains<T> contains) {
-    return contains.filter(this, (l, r) -> l || r, (l, r) -> l && r);
+    return contains.filter(this, (l, r) -> l || r, (l, r) -> l && r, b -> BLOCK_MIGHT_MATCH);
```
wgtmac, line 217:
```suggestion
    return contains.filter(this, (l, r) -> l || r, (l, r) -> l && r, v -> BLOCK_MIGHT_MATCH);
```

Just want to be consistent with `BloomFilterImpl` and `DictionaryFilter`.

parquet-column/src/main/java/org/apache/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
wgtmac: `IncrementallyUpdatedFilterPredicateGenerator.java` is pretty difficult to read and review. Since this is only applied to `ContainsPredicate` and does not affect others, I'm fine to make this change.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
wgtmac: Yes, we cannot make such subtraction unless it is a single-value page which has only `bar`.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -345,11 +374,14 @@ public <R> R accept(Visitor<R> visitor) {
...
-     * and finally delegating to the underlying ColumnFilterPredicate.
+     * Applies a filtering Visitor to the Contains predicate, traversing any composed And or Or clauses,
+     * and finally delegating to the underlying column predicate.
      */
     public abstract <R> R filter(
-        Visitor<R> visitor, BiFunction<R, R, R> andBehavior, BiFunction<R, R, R> orBehavior);
+        Visitor<R> visitor,
+        BiFunction<R, R, R> andBehavior,
+        BiFunction<R, R, R> orBehavior,
+        Function<R, R> notBehavior);
```
wgtmac, line 384: IIUC, all kinds of filters return all rows if `notBehavior` is found, right?

wgtmac (commented):
Thanks for the change! I just took an initial pass and has some questions overall.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
clairemcginty: Yes, for example `NotIn` 👍

wgtmac (approved):
+1

wgtmac:
@gszadovszky Do you want to take a look?

gszadovszky (approved):

wgtmac:
Merged. Thanks @clairemcginty and @gszadovszky!

PR #2951, opened 2024-07-13
 GH-2935: Avoid double close of ParquetFileWriter

### LINKED ISSUE #2935: Double close of ParquetFileWriter in ParquetWriter
ParquetWriter.close() invokes InternalParquetRecordWriter.close() with following logic:
```
  public void close() throws IOException, InterruptedException {
    if (!closed) {
      try {
        if (aborted) {
          return;
        }
        flushRowGroupToStore();
        FinalizedWriteContext finalWriteContext = writeSupport.finalizeWrite();
        Map<String, String> finalMetadata = new HashMap<String, String>(extraMetaData);
        String modelName = writeSupport.getName();
        if (modelName != null) {
          finalMetadata.put(ParquetWriter.OBJECT_MODEL_NAME_PROP, modelName);
        }
        finalMetadata.putAll(finalWriteContext.getExtraMetaData());
        parquetFileWriter.end(finalMetadata);
      } finally {
        AutoCloseables.uncheckedClose(columnStore, pageStore, bloomFilterWriteStore, parquetFileWriter);
        closed = true;
      }
    }
  }
```

Apparently parquetFileWriter is closed twice here, first time by
**parquetFileWriter.end(finalMetadata)**, which eventually calls parquetFileWriter.close()

second time by
AutoCloseables.uncheckedClose(columnStore, pageStore, bloomFilterWriteStore, **parquetFileWriter**);

This causes the underlying PositionOutputStream in ParquetFileWriter to be flushed again after it's closed, which may raise exception depending on the underlying stream implementation.

```
    try (PositionOutputStream temp = out) {
      temp.flush();
      if (crcAllocator != null) {
        crcAllocator.close();
      }
    }
  }
```

sample exception:

Caused by: org.apache.parquet.util.AutoCloseables$ParquetCloseResourceException: Unable to close resource
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:85)
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:94)
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.close(InternalParquetRecordWriter.java:144)
	at org.apache.parquet.hadoop.ParquetWriter.close(ParquetWriter.java:437)
	... 70 more
Caused by: java.io.IOException: stream is already closed
	(-------- specific stream implementation ----------------)
	at java.io.FilterOutputStream.flush(FilterOutputStream.java:140)
	at java.io.DataOutputStream.flush(DataOutputStream.java:123)
	at org.apache.parquet.hadoop.util.HadoopPositionOutputStream.flush(HadoopPositionOutputStream.java:59)
	at org.apache.parquet.hadoop.ParquetFileWriter.close(ParquetFileWriter.java:1659)
	at org.apache.parquet.util.AutoCloseables.close(AutoCloseables.java:49)
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:83)

This issue is observed since 1.14.0, and I suspect PARQUET-2496 is caused by this similar issue.

hellishfire:
### Rationale for this change:
Refer to #2935 

Prevent double close of ParquetFileWriter

### Are these changes tested?
Yes

No

Closes #2935

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -1667,6 +1667,7 @@ public class ParquetFileWriter implements AutoCloseable {
       if (crcAllocator != null) {
         crcAllocator.close();
       }
+    } finally {
       closed = true;
     }
   }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -170,6 +170,7 @@ public class ParquetFileWriter implements AutoCloseable {
 
   // set when end is called
   private ParquetMetadata footer = null;
+  private boolean closed;
 
   private final CRC32 crc;
   private final ReusingByteBufferAllocator crcAllocator;
@@ -1658,11 +1659,15 @@ public class ParquetFileWriter implements AutoCloseable {
 
   @Override
   public void close() throws IOException {
+    if (closed) {
+      return;
+    }
     try (PositionOutputStream temp = out) {
       temp.flush();
       if (crcAllocator != null) {
         crcAllocator.close();
       }
+      closed = true;
     }
   }
 
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
@@ -286,6 +286,9 @@ public class TestParquetFileWriter {
     w.endColumn();
     w.endBlock();
     w.end(new HashMap<String, String>());
+    // Although writer is already closed in previous end(),
+    // explicitly close it again to verify double close behavior.
+    w.close();
 
     ParquetMetadata readFooter = ParquetFileReader.readFooter(configuration, path);
     assertEquals("footer: " + readFooter, 2, readFooter.getBlocks().size());


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1658,11 +1659,15 @@ public void end(Map<String, String> extraMetaData) throws IOException {
...
   public void close() throws IOException {
+    if (closed) {
+      return;
+    }
     try (PositionOutputStream temp = out) {
       temp.flush();
       if (crcAllocator != null) {
         crcAllocator.close();
       }
+      closed = true;
```
wgtmac, line 1670: Should we put this in the finally block just in case of any exception?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -286,6 +286,9 @@ public void testWriteRead() throws Exception {
     w.endColumn();
     w.endBlock();
     w.end(new HashMap<String, String>());
+    // Although writer is already closed in previous end(),
+    // explicitly close it again to verify double close behavior.
+    w.close();
```
Fokko, line 291: Should we give a best practice in the test, and turn this into a try-with-resource? The `ParquetFileWriter` is auto-closable.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
hellishfire: fixed

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
hellishfire: I didn't do this for two reasons:

1 The verification codes following this write operation use multiple local variables declared during the write (e.g. c1Starts c1p1Starts c1Ends). It's a pain to pre declare all these variables before the try block.

2 Each test in this test file is coded like this, so it's a pain to refactor all usages of ParquetFileWriter here.

doki23: Maybe we can just turn this one into a try-with-resource style.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
doki23: Putting it in finally block makes it be not retryable.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
Fokko: Ah yes, let's keep it like this for now.

Fokko (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
hellishfire: > Putting it in finally block makes it be not retryable.

That was also my initial thought. I was actually torn between these two choices, but I suppose it's rare for people to actually retry failed close operation. Feedback is welcome.

doki23: > it's rare for people to actually retry failed close operation

I agree with you, it's rare.
But if someone retries, it will not work as expected -- file is not closed as expected. This is not user friendly, right?

doki23: I notice that `try {} finally { closed = true; }` is common in parquet writer code base, so it' ok to follow up.

wgtmac (approved):

hellishfire:
@wgtmac Is there any plan to release another 1.14 minor version? 1.14 is unusable with some stream implementations caused by this bug.

wgtmac:
Could you please point me to the downstream issue? We do not have a planned release for 1.14.2 but we can do it at any time if it breaks many downstream cases. @hellishfire 

BTW, I saw that Apache Iceberg is unable to pick parquet-java 1.14.1. Is there anything we can do on the parquet side for a quick release? @Fokko

hellishfire:
> Could you please point me to the downstream issue? We do not have a planned release for 1.14.2 but we can do it at any time if it breaks many downstream cases. @hellishfire
> 
> BTW, I saw that Apache Iceberg is unable to pick parquet-java 1.14.1. Is there anything we can do on the parquet side for a quick release? @Fokko

Ok...We encountered this issue in a private code base, but I suppose this bug can break many filesystem implementations if they don't allow flush after close, which should be a very reasonable behavior.

wgtmac:
@hellishfire What about sticking to 1.13.1 for a while? We have a planned release of 1.15.0 in this October and there are not many defect reports received for 1.14.1.

Fokko:
I'm happy to run a quick release, as another PR has been raised https://github.com/apache/parquet-java/pull/2958

wgtmac:
> It has issues with the latest Jackson release, which carries JDK17/21 specific files and is not supported by the Gradle Shadow plugin we're using.

Does it make sense to downgrade Jackson on our side? I know it was tied to some other fixes so it might be difficult.

> I'm happy to run a quick release, as another PR has been raised #2958

That's awesome! Let me know if I can help anything.

Fokko:
> Does it make sense to downgrade Jackson on our side? I know it was tied to some other fixes so it might be difficult.

I would be hesitant to downgrade for _just_ Iceberg. When more folks are running into it, then it makes sense, but it looks like Iceberg is a bit isolated because of Gradle.

hellishfire:
> > Does it make sense to downgrade Jackson on our side? I know it was tied to some other fixes so it might be difficult.
> 
> I would be hesitant to downgrade for _just_ Iceberg. When more folks are running into it, then it makes sense, but it looks like Iceberg is a bit isolated because of Gradle.

We used maven shade plugin to skip these JDK17/21 specific files from Jackson.
But anyway, we should merge this pull request.

Fokko:
Thanks for fixing this @hellishfire and thanks @wgtmac, @doki23 for the review 🚀

Issue #2952, opened 2024-07-13
Add maven wrapper


nastra:
### Describe the enhancement requested

By adding a maven wrapper user don't need to have `mvn` installed on their environment to run the build

### Component(s)

Build

PR #2953, opened 2024-07-13
GH-2952: Add maven wrapper

nastra:
### Rationale for this change
Having a maven wrapper in the project removes the burden for users to have maven installed on their system.
The wrapper was added by running `mvn wrapper:wrapper`

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+wrapperVersion=3.3.2
+distributionType=only-script
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.8/apache-maven-3.9.8-bin.zip
--- a/mvnw
+++ b/mvnw
@@ -1,0 +1,259 @@
+#!/bin/sh
+# ----------------------------------------------------------------------------
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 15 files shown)


nastra:
As a next step (please let me know if that should be within this PR) I was planning to update all `mvn` references in the codebase to `./mvnw`

wgtmac:
This looks nice. However, my concern is the maintenance burden of the new script.

@julienledem @gszadovszky WDYT?

joyCurry30:
Missing mvnw.cmd

.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+wrapperVersion=3.3.2
+distributionType=only-script
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.8/apache-maven-3.9.8-bin.zip
```
joyCurry30, line 19: This version is too new and may cause problems.

Fokko: The CI runs 3.8.8. Did you test this locally @nastra?

nastra: yes I tested this locally. If there are concerns with that Maven version then I can switch that to a version that is used by CI. Or I could update CI scripts to use the wrapper to make sure that particular maven version runs without issues

nastra:
> This looks nice. However, my concern is the maintenance burden of the new script.
> 
> @julienledem @gszadovszky WDYT?

@wgtmac typically this script doesn't require any maintenance and will be updated automatically when updating the underlying maven version through the wrapper. That doesn't mean that every maven version bump requires an updated script, just that the wrapper task will update the script if necessary

.mvn/wrapper/maven-wrapper.properties
nastra: I've updated CI to run through the wrapper to see if there would be any issues with that version

Fokko: Nice, that would be my preference. Let's see what others think.

nastra: @joyCurry30 @Fokko CI seems to be running fine with that Maven version

Fokko (approved):
I think this is nice, but I'll leave this open for others to chime in as well 👍

joyCurry30 (approved):

mvnw
```diff
@@ -0,0 +1,259 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# ----------------------------------------------------------------------------
+
+# ----------------------------------------------------------------------------
+# Apache Maven Wrapper startup batch script, version 3.3.2
```
wgtmac, line 22: Is it copied from somewhere else? It would be good if a link is provided so we can keep it updated.

nastra: no, it's being generated and will be automatically updated (if necessary) when updating the underlying maven version of the wrapper

nastra: if you run `mvn wrapper:wrapper` against `master` you'll get the same scripts generated

wgtmac (approved):
LGTM

Thanks @nastra!

Issue #2956, opened 2024-07-16
parquet-avro 1.14 not compatible with avro 1.8


RustedBones:
### Describe the bug, including details regarding any error messages, version, and platform.

Some changes made to `parquet-avro` [here](https://github.com/apache/parquet-java/pull/1262#discussion_r1678977866) uses API introduced in avro 1.9.

It should be possible to keep `parquet-avro` compatibility with older avro version by using the schema builder api instead

PR #2957, opened 2024-07-16
GH-2956: Use avro SchemaBuilder API to convert record

RustedBones:
The avro schema builder API is cleaner and more stable. It decreases chance of using newly introduced avro API in case user run with legacy avro version

As OPTIONAL converted fields sets null as default, increase consistency by using [] as default for REPEATED converted fields.

Fix: #2956


Fokko (approved):
Looks good @RustedBones Thanks for working on this

Can you run spotless?

wgtmac (approved):

PR #2993, opened 2024-08-19
GH-2992: Gate LocalTimestamp references in AvroSchemaConverter

### LINKED ISSUE #2992: Parquet 1.14.2-RC1 breaks compat with Avro 1.7, 1.8, 1.9

AvroSchemaConverter references Avro logical type converters that aren't available in Avro versions below 1.10: See more here: https://github.com/apache/parquet-java/pull/1115/files#r1721966099

For reference, these are the local TIMESTAMP converters available in Avro 1.10+: https://github.com/apache/avro/blob/release-1.10.0/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L172-L194

Whereas in 1.9.2, only `timestampMillis`/`timestampMicros` area available: https://github.com/apache/avro/blob/release-1.9.2/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L158-L168

### Component(s)

Avro

clairemcginty:
### Rationale for this change

See linked [issue](https://github.com/apache/parquet-java/issues/2992); AvroSchemaConverter references Avro classes and methods that don't exist before Avro 1.10

Performs an Avro version check to silo references to local timestamp classes.

### Are these changes tested?

I've tested them locally on one of my projects using Avro 1.8... unfortunately there isn't a great way to unit-test this :/

This should preserve the existing behavior for Avro 1.8 users while unlocking the new LocalTimestamp types for 1.10+ users.

Closes #2992

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -22,6 +22,7 @@ import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static org.apache.parquet.avro.AvroReadSupport.READ_INT96_AS_FIXED;
 import static org.apache.parquet.avro.AvroReadSupport.READ_INT96_AS_FIXED_DEFAULT;
+import static org.apache.parquet.avro.AvroRecordConverter.getRuntimeAvroVersion;
 import static org.apache.parquet.avro.AvroWriteSupport.WRITE_FIXED_AS_INT96;
 import static org.apache.parquet.avro.AvroWriteSupport.WRITE_OLD_LIST_STRUCTURE;
 import static org.apache.parquet.avro.AvroWriteSupport.WRITE_OLD_LIST_STRUCTURE_DEFAULT;
@@ -488,15 +489,20 @@ public class AvroSchemaConverter {
       return timeType(true, MICROS);
     } else if (logicalType instanceof LogicalTypes.TimestampMillis) {
       return timestampType(true, MILLIS);
-    } else if (logicalType instanceof LogicalTypes.LocalTimestampMillis) {
-      return timestampType(false, MILLIS);
     } else if (logicalType instanceof LogicalTypes.TimestampMicros) {
       return timestampType(true, MICROS);
-    } else if (logicalType instanceof LogicalTypes.LocalTimestampMicros) {
-      return timestampType(false, MICROS);
     } else if (logicalType.getName().equals(LogicalTypes.uuid().getName()) && writeParquetUUID) {
       return uuidType();
     }
+
+    if (avroVersionSupportsLocalTimestampTypes()) {
+      if (logicalType instanceof LogicalTypes.LocalTimestampMillis) {
+        return timestampType(false, MILLIS);
+      } else if (logicalType instanceof LogicalTypes.LocalTimestampMicros) {
+        return timestampType(false, MICROS);
+      }
+    }
+
     return null;
   }
 
@@ -538,7 +544,7 @@ public class AvroSchemaConverter {
             LogicalTypeAnnotation.TimeUnit unit = timestampLogicalType.getUnit();
             boolean isAdjustedToUTC = timestampLogicalType.isAdjustedToUTC();
 
-            if (isAdjustedToUTC) {
+            if (isAdjustedToUTC || !avroVersionSupportsLocalTimestampTypes()) {
               switch (unit) {
                 case MILLIS:
                   return of(LogicalTypes.timestampMillis());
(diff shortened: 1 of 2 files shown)


clairemcginty:
cc @RustedBones , can you take a look at this fix as well?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -605,4 +611,14 @@ private static String namespace(String name, Map<String, Integer> names) {
     Integer nameCount = names.merge(name, 1, (oldValue, value) -> oldValue + 1);
     return nameCount > 1 ? name + nameCount : null;
   }
+
+  /* Avro <= 1.9 does not support conversions to LocalTimestamp{Micros, Millis} classes */
+  private static boolean avroVersionSupportsLocalTimestampTypes() {
```
Fokko, line 616: What do you think of mocking this method to capture the behavior of supporting and not-supporting local timestamps.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -538,7 +544,7 @@ public Optional<LogicalType> visit(
             LogicalTypeAnnotation.TimeUnit unit = timestampLogicalType.getUnit();
             boolean isAdjustedToUTC = timestampLogicalType.isAdjustedToUTC();
 
-            if (isAdjustedToUTC) {
+            if (isAdjustedToUTC || !avroVersionSupportsLocalTimestampTypes()) {
```
Fokko, line 547: If local timestamps aren't supported, isn't it more safe to default to `empty()`? Otherwise, we're mixing up types.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
clairemcginty: I'll give it a try with Powermock!

I might start a thread on the mailing list this week discussing Avro version support more broadly, to formalize which Avro versions are currently supported/any planned EOL dates? I've been thinking it would also be good to have some automated testing of parquet-avro on different Avro versions... it would just be a little complicated to set up since we'd need either a separate Maven module per supported Avro version, or some very clever classloading :)

Fokko: Hey @clairemcginty, that's a good idea!

What do we consider the lower-bound? :D Avro 1.8.2 is May 2017, so that's pretty old, but I'm aware of some breaking API changes because we exposed some Jackson classes in the public API.

To do cross-testing, we can do something similar to we do with Hadoop 2, you can get some inspiration there: https://github.com/apache/parquet-java/blob/d4384d3f2e7703fab6363fd9cd80a001e9561db2/pom.xml#L638-L643

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
clairemcginty: sorry for the confusion! changes to both methods should be needed; I've added test cases to `TestAvroSchemaConverter` that demonstrate the issue and fail without my changes to `AvroSchemaConverter` applied. Basically, we don't want to try to assign a logical type that doesn't exist in the user's Avro version.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
clairemcginty: I think 1.8 is a reasonable lower bound! for context on our use case, we use the [Beam](https://github.com/apache/beam) library, which until very recently was tightly coupled with 1.8. Things are now more flexible, but there's a lot of inertia surrounding 1.8 -- the breaking changes surrounding logical types make it a non-trivial upgrade 😅 

I'll take a look at the hadoop setup! Avro is a bit tricky because we need to both set the Avro-compiler version, and load the Avro core library, for whatever version of Avro we want to test...

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
Fokko: No problem, thanks for adding the test 👍

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
Fokko: Thanks for the context, that helps. I see that 1.8.2 is [still in there](https://github.com/apache/beam/blob/b3a874f476646676bf7eb909cae738a696f36303/sdks/java/extensions/avro/build.gradle#L40), so that sounds reasonable to me 👍

Fokko (approved):

PR #2999, opened 2024-08-24
MINOR: Don't run all the tests on a release

Fokko:
### Rationale for this change

We don't want to rerun all the tests during a release since it takes a lot of time. We can rely on the CI instead.


wgtmac (approved):
+1

julienledem (approved):
SGTM

nastra (approved):

PR #3015, opened 2024-09-18
GH-3016: Do not use dependency-reduced pom for attached shaded artifact

RustedBones:
Since 0.14.0, `parquet-cli` does not contain transitive compile scope dependencies. Check [0.14.0](https://central.sonatype.com/artifact/org.apache.parquet/parquet-cli/1.14.0/dependencies) vs [0.13.1](https://central.sonatype.com/artifact/org.apache.parquet/parquet-cli/1.13.1/dependencies).

this is an issue from the `maven-shade-plugin`. When `shadedArtifactAttached` is set, `createDependencyReducedPom` should be `false`.

See https://issues.apache.org/jira/projects/MSHADE/issues/MSHADE-419

fix: #3016

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Parquet-Java uses Maven to build and depends on the thrift compiler (protoc is n
 To build and install the thrift compiler, run:
 
 ```
-wget -nv http://archive.apache.org/dist/thrift/0.20.0/thrift-0.20.0.tar.gz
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

wgtmac:
cc @gszadovszky

gszadovszky (approved):

Issue #3035, opened 2024-10-21
ParquetRewriter: add column renaming feature


MaxNevermind:
### Describe the enhancement requested

ParquetRewriter can benefit from addition of column renaming functionality.

### Component(s)

Core

PR #3036, opened 2024-10-21
GH-3035: ParquetRewriter: Add a column renaming feature

MaxNevermind:
### Rationale for this change
This feature extension is based on a real use-case when a input parquet dataset need to transformed to a new one using a set of basic schema transformations.  `ParquetRewriter` already supports some of transformations: pruning, masking, encrypting, changing a codec. This PR add one of missing transformations - renaming.

- Add a new option `renameColumns` to `RewriteOptions` class which is options builder for `ParquetRewriter`
- Add support of to `renameColumns` to `ParquetRewriter`

### Are these changes tested?
Yes.

Yes.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
@@ -624,6 +624,29 @@ class IntColumnChunkMetaData extends ColumnChunkMetaData {
   public SizeStatistics getSizeStatistics() {
     return sizeStatistics;
   }
+
+  /**
+   * Copies this ColumnChunkMetaData with chunk's path and type changed to provided ones.
+   *
+   * @param path ColumnPath of resulting chunk
+   * @param type PrimitiveType of resulting chunk
+   * @return resulting chunk
+   */
+  public ColumnChunkMetaData copy(ColumnPath path, PrimitiveType type) {
+    return ColumnChunkMetaData.get(
+        path,
+        type,
+        this.getCodec(),
+        this.getEncodingStats(),
+        this.getEncodings(),
+        this.getStatistics(),
+        this.getFirstDataPageOffset(),
+        this.getDictionaryPageOffset(),
+        this.getValueCount(),
+        this.getTotalSize(),
+        this.getTotalUncompressedSize(),
+        this.getSizeStatistics());
+  }
 }
 
 class LongColumnChunkMetaData extends ColumnChunkMetaData {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
@@ -504,29 +504,6 @@ public abstract class ColumnChunkMetaData {
   public boolean isEncrypted() {
     return false;
   }
-
-  /**
-   * Copies this ColumnChunkMetaData with path and type changed to provided ones.
-   *
-   * @param path a new ColumnPath of a chunk
-   * @param type a new PrimitiveType of a chunk
-   * @return resulting chunk
-   */
-  public ColumnChunkMetaData copy(ColumnPath path, PrimitiveType type) {
-    return ColumnChunkMetaData.get(
-        path,
-        type,
-        this.getCodec(),
-        this.getEncodingStats(),
-        this.getEncodings(),
-        this.getStatistics(),
-        this.getFirstDataPageOffset(),
-        this.getDictionaryPageOffset(),
-        this.getValueCount(),
-        this.getTotalSize(),
-        this.getTotalUncompressedSize(),
-        this.getSizeStatistics());
-  }
 }
 
 class IntColumnChunkMetaData extends ColumnChunkMetaData {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
@@ -504,6 +504,29 @@ public abstract class ColumnChunkMetaData {
   public boolean isEncrypted() {
     return false;
   }
+
+  /**
+   * Copies this ColumnChunkMetaData with path and type changed to provided ones.
+   *
+   * @param path a new ColumnPath of a chunk
+   * @param type a new PrimitiveType of a chunk
+   * @return resulting chunk
+   */
+  public ColumnChunkMetaData copy(ColumnPath path, PrimitiveType type) {
+    return ColumnChunkMetaData.get(
+        path,
+        type,
+        this.getCodec(),
+        this.getEncodingStats(),
+        this.getEncodings(),
+        this.getStatistics(),
+        this.getFirstDataPageOffset(),
+        this.getDictionaryPageOffset(),
+        this.getValueCount(),
+        this.getTotalSize(),
+        this.getTotalUncompressedSize(),
+        this.getSizeStatistics());
+  }
 }
 
 class IntColumnChunkMetaData extends ColumnChunkMetaData {
@@ -623,29 +646,6 @@ class IntColumnChunkMetaData extends ColumnChunkMetaData {
   @Override
   public SizeStatistics getSizeStatistics() {
     return sizeStatistics;
-  }
-
-  /**
-   * Copies this ColumnChunkMetaData with chunk's path and type changed to provided ones.
-   *
-   * @param path ColumnPath of resulting chunk
-   * @param type PrimitiveType of resulting chunk
-   * @return resulting chunk
-   */
-  public ColumnChunkMetaData copy(ColumnPath path, PrimitiveType type) {
-    return ColumnChunkMetaData.get(
-        path,
-        type,
-        this.getCodec(),
-        this.getEncodingStats(),
-        this.getEncodings(),
-        this.getStatistics(),
-        this.getFirstDataPageOffset(),
-        this.getDictionaryPageOffset(),
-        this.getValueCount(),
-        this.getTotalSize(),
-        this.getTotalUncompressedSize(),
-        this.getSizeStatistics());
   }
 }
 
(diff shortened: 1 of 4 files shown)


MaxNevermind:
@wgtmac 
Can you check this out? I need your opinion - If you see any problems with such a feature conceptually. Then I can polish the implementation.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -431,7 +480,27 @@ private void processBlock(
     if (chunk.isEncrypted()) {
       throw new IOException("Column " + chunk.getPath().toDotString() + " is already encrypted");
     }
-    ColumnDescriptor descriptor = outSchema.getColumns().get(outColumnIdx);
+
+    ColumnChunkMetaData chunkColumnsRenamed = chunk;
+    if (renamedColumns != null && !renamedColumns.isEmpty()) {
+      chunkColumnsRenamed = ColumnChunkMetaData.get(
```
wgtmac, line 486: If we add a field to ColumnChunkMetaData in the future, the current `ColumnChunkMetaData.get` method gets deprecated and we can easily get out of sync here. I'm not sure if it is a good idea to introduce a copy|clone method to it with only the name change.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -145,15 +145,18 @@ public class ParquetRewriter implements Closeable {
   private final Queue<TransParquetFileReader> inputFiles = new LinkedList<>();
   private final Queue<TransParquetFileReader> inputFilesToJoin = new LinkedList<>();
   private final MessageType outSchema;
+  private final MessageType outSchemaWithRenamedColumns;
```
wgtmac, line 148: Is it possible not to add a new `outSchemaWithRenamedColumns` and stick to `outSchema` with field name replaced?

wgtmac (commented):
Thanks @MaxNevermind for the proposal! I just took an initial pass and have the following questions:
- What is your use case? Do you need reordering fields in the future?
- How does renaming work with other features, including join, mask, and encrypt?

cc @ConeyLiu

MaxNevermind:
> What is your use case? 

We have a large base dataset which is split into multiple versions at the end of the flow. Those final datasets have majority of columns overlapping but some columns are dropped & some columns are renamed.  Similar thing can be achieved by multiple HMS views on the top of unified base HMS table but it is not always possible, for example if different users are supposed to have access just to a single version of a dataset, it is hard to achieve that with HMS views without giving access to users to that underlying unified table.

> Do you need reordering fields in the future?

In our case we don't care about column ordering.

> How does renaming work with other features, including join, mask, and encrypt?

mask - it is not supported as it seems meaningless to rename a column that is dropped
join - I planned to support it, so the column could be joined and then renamed if needed
encrypt - I planned to support it, so the column can be encrypted and renamed if needed

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: I tried that originally but it just doesn't work. There are places where we need to have old schema. For example when we read the data we need to provide an old path in original schema, if we use a new schema, renamed column just won't be able found and won't be able to read as it is just not there. Maybe we can have some sort of abstraction like a some schema holder that allows to extract both old and rename columns through a single entity but I don't it is possible to avoid the necessity of having both somewhere in the code.

wgtmac:
Thanks for the explanation! I asked for `column ordering` because renaming and reordering may happen together in the case of schema evolution. We can support them in the future when needed.

> does it make sense to allow nullification + renaming

I am also skeptical of this case. But it seems to be valid if one wants to nullify a renamed column which has sensitive data?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
wgtmac: Do we support the case where some input files have renamed columns but others don't?

MaxNevermind: As I understand that behavior is not supported by current master branch version of ParquetRewriter and I didn't plan to change it in this PR, it expects the schema for all input files to be the same, both main and join input(if provided), so there are two schemas at max: one schema(if no join files provided), two schemas(if join files are provided). If some of the files are missing a column then I think `ensureSameSchema()` will fail.

Renaming feature doesn't change that, a renamed column can be in the main input or it can be in a join input. In case of join column name overlap with main input, the result of a join depends on a `overwriteInputWithJoinColumns` option, but it doesn't change the output of renaming, if that column is to be renamed it should not change the source of input data being main or join files.

MaxNevermind:
> I asked for `column ordering` because renaming and reordering may happen together in the case of schema evolution. We can support them in the future when needed.

In case of column reordering there is a need to provide a new order of column, correct? If I would be implementing it I guess I would create a RewriteOptions's option like `Map<SrcColName, Map.Entry<DistColName, DestColIdx>> columnsReordered` which would have merged capabilities of both renaming and reordering. I think current implementation can be extended to something like that in the future, for example if `columnsReordered` is used we should prohibit `renameColumns` option usage, wdyt?

> I am also skeptical of this case. But it seems to be valid if one wants to nullify a renamed column which has sensitive data?

I've just checked `MaskMode` enum, it has planned hashing feature as I understand. Renaming hashed column would definitely should be beneficial to support. Okay, I think we should try to support renaming of masked columns.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/RewriteOptions.java
```diff
@@ -561,13 +582,29 @@ public RewriteOptions build() {
...
+        if (renameColumns != null) {
+          for (Map.Entry<String, String> entry : renameColumns.entrySet()) {
+            Preconditions.checkArgument(
+                !encryptColumns.contains(entry.getKey()), "Cannot prune and rename same column");
+          }
+        }
+      }
+
+      if (renameColumns != null && !renameColumns.isEmpty()) {
+        for (Map.Entry<String, String> entry : renameColumns.entrySet()) {
```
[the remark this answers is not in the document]
MaxNevermind: done
check ParquetRewriter's added method `ensureRenamedColumnsCorrectness()`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -431,7 +480,27 @@ private void processBlock(
     if (chunk.isEncrypted()) {
       throw new IOException("Column " + chunk.getPath().toDotString() + " is already encrypted");
     }
-    ColumnDescriptor descriptor = outSchema.getColumns().get(outColumnIdx);
+
+    ColumnChunkMetaData chunkColumnsRenamed = chunk;
```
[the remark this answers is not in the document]

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: Not sure if I understood correctly, I added `ColumnChunkMetaData.copy()`, let me know if that works.

MaxNevermind:
@wgtmac 
In general do you think this feature looks worth adding, no hard no-go traits in this feature? If yes, I will work on extending functionality to those overlapping cases with masking and will start polishing it.

wgtmac:
Yes, I think this is a fair use case, provided that the code logic does not change a lot.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
MaxNevermind: > Is it possible not to add a new outSchemaWithRenamedColumns and stick to outSchema with field name replaced?

I replaced renamed schema variable with dynamic extraction of schema with renamed fields where needed, check it out. Didn't want to do it because of potential overhead but it should be very small.

wgtmac: Yes, I think this looks better now. Thanks for making this change!

MaxNevermind:
@wgtmac 
I think I'm done with implementation, can you take a look?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -149,27 +149,29 @@ public class ParquetRewriter implements Closeable {
...
     LOG.info(
         "Start rewriting {} input file(s) {} to {}",
         inputFiles.size() + inputFilesToJoin.size(),
         Stream.concat(options.getParquetInputFiles().stream(), options.getParquetInputFilesToJoin().stream())
             .collect(Collectors.toList()),
-        out);
-
-    this.outSchema = pruneColumnsInSchema(getSchema(), options.getPruneColumns());
-    this.extraMetaData = getExtraMetadata(options);
+        options.getParquetOutputFile());
```
wgtmac, line 174:
```suggestion
        out);
```

wgtmac (commented):
The implementation looks good. Some nits to fix. Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/ColumnChunkMetaData.java
```diff
@@ -504,6 +504,29 @@ public boolean hasDictionaryPage() {
...
   }
+
+  /**
+   * Copies this ColumnChunkMetaData with path and type changed to provided ones.
+   *
+   * @param path a new ColumnPath of a chunk
+   * @param type a new PrimitiveType of a chunk
+   * @return resulting chunk
+   */
+  public ColumnChunkMetaData copy(ColumnPath path, PrimitiveType type) {
```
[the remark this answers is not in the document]
MaxNevermind: Done, reverted back and added a comment

wgtmac (approved):
LGTM. Thanks!

PR #3038, opened 2024-10-30
GH-3029: Fix EncryptionPropertiesHelper not to use java.nio.file.Path

### LINKED ISSUE #3029: Path parsing error in ParquetWriter on Windows

Hello,

I've found problem with parsing Windows paths in ParquestWriter since version 1.14.2. Given following code:
```
Path file = new Path(new File("out.parquet").toURI());
Configuration conf = new Configuration(false);
GroupWriteSupport gws = new GroupWriteSupport();
GroupWriteSupport.setSchema(new MessageType("name"), conf);
new ParquetWriter(file, conf, gws);
```
the exception is thrown:
```
java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:/Z:/out.parquet
        at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
        at java.base/java.nio.file.Path.of(Path.java:147)
        at java.base/java.nio.file.Paths.get(Paths.java:69)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:392)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:357)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:332)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:273)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:228)
        at org.apache.parquet.hadoop.ParquetWriter.<init>(ParquetWriter.java:307)
```
The problem is caused by change in [ParquetWriter:392](https://github.com/apache/parquet-java/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L392). Here used to be `new Path(path)` where Path is from org.apache.hadoop.fs (see [here](https://github.com/apache/parquet-java/blob/apache-parquet-1.13.1/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L286)) but now there is `Paths.get(path)` from java.nio.file. The problem is that path is in fact URI, i.e. something like `file:/...`, but java.nio.file.Path does not expect URI. It probably does not work correctly also on Linux but there is no exception since ":" is OK in path there.

Please note that the example code does not fully work. The code is part of larger application with complex initialization so right now I don't have the correct schema for the writer. Fortunately that exception is thrown before bad schema error.

### Component(s)

Core

wgtmac:
### Rationale for this change

EncryptionPropertiesHelper has been changed to use java.nio.file.Path but the `file:` URI will fail on Windows.

See #3029 for detail.

Change EncryptionPropertiesHelper not to use java.nio.file.Path.

### Are these changes tested?

Pass CI.

No

Closes #3029

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/EncryptionPropertiesHelper.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/EncryptionPropertiesHelper.java
@@ -18,27 +18,25 @@
  */
 package org.apache.parquet.hadoop;
 
-import java.net.URI;
-import java.nio.file.Path;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.conf.ParquetConfiguration;
 import org.apache.parquet.crypto.EncryptionPropertiesFactory;
 import org.apache.parquet.crypto.FileEncryptionProperties;
 import org.apache.parquet.hadoop.api.WriteSupport;
 import org.apache.parquet.hadoop.util.ConfigurationUtil;
+import org.apache.parquet.io.OutputFile;
 
 final class EncryptionPropertiesHelper {
   static FileEncryptionProperties createEncryptionProperties(
-      ParquetConfiguration fileParquetConfig, Path tempFilePath, WriteSupport.WriteContext fileWriteContext) {
+      ParquetConfiguration fileParquetConfig, OutputFile file, WriteSupport.WriteContext fileWriteContext) {
     EncryptionPropertiesFactory cryptoFactory = EncryptionPropertiesFactory.loadFactory(fileParquetConfig);
     if (null == cryptoFactory) {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


wgtmac:
@gszadovszky Could you help review this? Thanks!

gszadovszky (approved):

PR #1278, opened 2024-02-21
PARQUET-2432: Use ByteBufferAllocator over hardcoded heap allocation

gszadovszky:
* Updated BytesInput implementations to rely on a ByteBufferAllocator instance for allocating/releasing ByteBuffer objects.
* Extend the usage of a ByteBufferAllocator instead of the hardcoded usage of heap (e.g. byte[], ByteBuffer.allocate etc.)
* parquet-cli related code parts including ParquetRewriter and tests are not changed in this effort

Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [PARQUET-2432](https://issues.apache.org/jira/browse/PARQUET-2432) issues and references
  them in the PR title. For example, "PARQUET-1234: My Parquet PR"
    - https://issues.apache.org/jira/browse/PARQUET-XXX
    - In case you are adding a dependency, check if the license complies with
      the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines
  from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":
    1. Subject is separated from body by a blank line
    1. Subject is limited to 50 characters (not including Jira issue reference)
    1. Subject does not end with a period
    1. Subject uses the imperative mood ("add", not "adding")
    1. Body wraps at 72 characters
    1. Body explains "what" and "why", not "how"

### Style
- [x] My contribution adheres to the code style guidelines and Spotless passes.
    - To apply the necessary changes, run `mvn spotless:apply -Pvector-plugins`

### Documentation

- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #3041, opened 2024-11-03
GH-3040: DictionaryFilter.canDrop may return false positive result when dict size exceeds 8k

### LINKED ISSUE #3040: DictionaryFilter.canDrop may return false positive result when dict size exceeds 8k
### Background

I get some data loss reports after upgrading the internal Spark's Parquet from 1.13.1 to 1.14.3, after some experiments, I believe this should be a bug on the Parquet side, and it could be worked around by disabling `spark.sql.parquet.filterPushdown`.

### Analysis

With some debugging, I think the issue was introduced by PARQUET-2432(https://github.com/apache/parquet-java/pull/1278).

The issue is, during the evaluation of `DictionaryFilter.canDrop`(this happens when reading a column that has `PLAIN_DICTIONARY` with pushed predications), when dict size exceeds 8k, only the head 8k was copied

https://github.com/apache/parquet-java/blob/274dc51bc9e5cc880ba3c77c3db826d2a4943965/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DictionaryPageReader.java#L113

the correct data
```
...
|00001fd0| 34 32 35 34 30 39 0a 00 00 00 37 32 39 33 39 38 |425409....729398|
|00001fe0| 39 35 33 30 0a 00 00 00 39 38 32 30 31 36 39 36 |9530....98201696|
|00001ff0| 34 34 0a 00 00 00 37 33 39 31 32 34 37 38 30 36 |44....7391247806|
|00002000| 0b 00 00 00 31 32 35 32 35 31 36 31 34 31 36 0a |....12525161416.|
|00002010| 00 00 00 38 37 38 35 35 34 34 36 34 35 0b 00 00 |...8785544645...|
|00002020| 00 31 32 32 38 30 38 37 35 39 30 32 0b 00 00 00 |.12280875902....|
...
```

the copied data
```
...
|00001fd0| 34 32 35 34 30 39 0a 00 00 00 37 32 39 33 39 38 |425409....729398|
|00001fe0| 39 35 33 30 0a 00 00 00 39 38 32 30 31 36 39 36 |9530....98201696|
|00001ff0| 34 34 0a 00 00 00 37 33 39 31 32 34 37 38 30 36 |44....7391247806|
|00002000| 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 |................|
|00002010| 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 |................|
|00002020| 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 |................|
...
```

the root cause is 
https://github.com/apache/parquet-java/blob/274dc51bc9e5cc880ba3c77c3db826d2a4943965/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java#L379

may not read fully if the underlying `InputStream`'s `available` method always returns 0

```
    private static class ReadableByteChannelImpl
        extends AbstractInterruptibleChannel    // Not really interruptible
        implements ReadableByteChannel
    {
     ...
        private static final int TRANSFER_SIZE = 8192;

        public int read(ByteBuffer dst) throws IOException {
            if (!isOpen()) {
                throw new ClosedChannelException();
            }

            int len = dst.remaining();
            int totalRead = 0;
            int bytesRead = 0;
            synchronized (readLock) {
                while (totalRead < len) {
                    int bytesToRead = Math.min((len - totalRead),
                                               TRANSFER_SIZE);
                    if (buf.length < bytesToRead)
                        buf = new byte[bytesToRead];
                    if ((totalRead > 0) && !(in.available() > 0))
                        break; // block at most once
                    try {
                        begin();
                        bytesRead = in.read(buf, 0, bytesToRead);
                    } finally {
                        end(bytesRead > 0);
                    }
                    if (bytesRead < 0)
                        break;
                    else
                        totalRead += bytesRead;
                    dst.put(buf, 0, bytesRead);
                }
                if ((bytesRead < 0) && (totalRead == 0))
                    return -1;

                return totalRead;
            }
        }
        ...
    }
```

### Component(s)

Core

pan3793:
### Rationale for this change

Fixes the data loss issue that reported in #3040

Ensure that `StreamBytesInput#writeInto(ByteBuffer buffer)` copies data fully, even if the underlying `InputStream` does not report available correctly.

### Are these changes tested?

UTs are added, I also tested it with an internal production data loss case.

Yes, this fixes some data loss cases, and I acknowledge that the bug affects Spark 4.0.0 preview2 which ships Parquet 1.14.2.

Closes #3040

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
@@ -18,7 +18,12 @@
  */
 package org.apache.parquet.bytes;
 
-import java.io.*;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.Channels;


parquet-common/src/test/java/org/apache/parquet/bytes/AvailableAgnosticInputStream.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.ByteArrayInputStream;
+
+public class AvailableAgnosticInputStream extends ByteArrayInputStream {
+
+  public AvailableAgnosticInputStream(byte[] buf) {
+    super(buf);
+  }
+
+  // In practice, there are some implementations always return 0 even if they has more data
```
pan3793, line 30: in my case, the underlying `IntputStream` is `H1SeekableInputStream`

pan3793:
cc @gszadovszky @wgtmac @Fokko

parquet-common/src/test/java/org/apache/parquet/bytes/TestBytesInput.java
```diff
@@ -140,6 +140,20 @@ public void testFromInputStream() throws IOException {
...
+  @Test
+  public void testFromLargeAvailableAgnosticInputStream() throws IOException {
+    // allocate a bytes that large than
+    // java.nio.channel.Channels.ReadableByteChannelImpl.TRANSFER_SIZE = 8192
+    byte[] data = new byte[9 * 1024];
+    RANDOM.nextBytes(data);
+    byte[] input = new byte[data.length + 10];
+    RANDOM.nextBytes(input);
+    System.arraycopy(data, 0, input, 0, data.length);
+    Supplier<BytesInput> factory = () -> BytesInput.from(new AvailableAgnosticInputStream(input), 9 * 1024);
```
wgtmac, line 152: What about using an anonymous class here instead of adding a new file?

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -376,7 +377,11 @@ void writeInto(ByteBuffer buffer) {
         ByteBuffer workBuf = buffer.duplicate();
         int pos = buffer.position();
         workBuf.limit(pos + byteCount);
-        Channels.newChannel(in).read(workBuf);
+        ReadableByteChannel channel = Channels.newChannel(in);
+        int remaining = byteCount;
+        while (remaining > 0) {
+          remaining -= channel.read(workBuf);
```
wgtmac, line 383: Is `remaining` reliable? Should we check the return value of `channel.read(workBuf)`?

pan3793: added a check to detect EOF case

wgtmac (approved):
+1. Thanks @pan3793!

parquet-common/src/test/java/org/apache/parquet/bytes/AvailableAgnosticInputStream.java
Fokko: Out of curiosity. Why are you using `H1SeekableInputStream`? This one is related to Hadoop 1.

pan3793: My test code is
(sorry, the file contains private data so I can not share)
```
public class MyDictionaryFilterTest {

  private static final Configuration conf = new Configuration();
  List<ColumnChunkMetaData> ccmd;
  ParquetFileReader reader;
  DictionaryPageReadStore dictionaries;
  private Path file = new Path("/Users/chengpan/Temp/part-2bb8404a-f6e5-4e9f-9161-f749c4bf46d0-2-2222");

  @Before
  public void setUp() throws Exception {
    reader = ParquetFileReader.open(conf, file);
    ParquetMetadata meta = reader.getFooter();
    ccmd = meta.getBlocks().get(0).getColumns();
    dictionaries = reader.getDictionaryReader(meta.getBlocks().get(0));
  }

  @After
  public void tearDown() throws Exception {
    reader.close();
  }

  @Test
  public void testEqBinary() throws Exception {
    BinaryColumn b = binaryColumn("source_id");
    FilterPredicate pred = eq(b, Binary.fromString("5059661515"));

    assertFalse(canDrop(pred, ccmd, dictionaries));
  }
}
```

Fokko: Thanks, this is very helpful!

gszadovszky (approved):

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -376,7 +378,15 @@ void writeInto(ByteBuffer buffer) {
         ByteBuffer workBuf = buffer.duplicate();
         int pos = buffer.position();
         workBuf.limit(pos + byteCount);
-        Channels.newChannel(in).read(workBuf);
```
ConeyLiu, line 379: Is there any other place that is used like this?

pan3793: I went through the [original PR](https://github.com/apache/parquet-java/pull/1278) and found nothing else, it would be great if others have a double check.

Fokko:
Thanks @pan3793 for finding and fixing this, and thanks @wgtmac @ConeyLiu and @gszadovszky for the review 🙌

PR #3061, opened 2024-11-14
GH-2943: Remove hadoop-2 support

### LINKED ISSUE #2943: Remove support for Hadoop <3.3
### Remove Hadoop 2 support

There is fallback logic in case it needs to seek within a file.

### Component(s)

_No response_

steveloughran:
### Rationale for this change

As discussed on the mailing list; hadoop 3.3.0 is the minimum
support hadoop version for 1.15

pom changes
- remove the hadoop-2 profile
- set base hadoop 3 version to 3.3.0

### Are these changes tested?

locally, yes.

The changed minimum hadoop version needs to go into the release notes

Closes #2943


Fokko:
@steveloughran I think we want to include the Hadoop 2 Github Actions as well

steveloughran:
> I think we want to include the Hadoop 2 Github Actions as well

I wanted to see what what happened there. Let me work on that.

steveloughran:
Looking at this I note that thift is download from apache archives and built every time. Apache infra might be unhappy about this; archive isn't replicated the way others are and they have "expressed concerns" about it being used elsewhere.

If he CI target OS/cpu is the same across builds, what about just sticking a precompiled thrift binary into dev/thift; if the cpu varies then it could be split for x86-64 and arm with the path set up appropriately.

steveloughran:
hadoop 3 builds are all happy; no hadoop 2 build any more

steveloughran:
no, vector failed. draft again

Fokko (approved):

steveloughran:
@Fokko thanks; merge when ready

PR #3072, opened 2024-11-22
GH-3070: Add Variant logical type annotation to parquet-java

### LINKED ISSUE #3070: Add Variant Logical Type

The Variant spec is added to Parquet project in https://github.com/apache/parquet-format/pull/456 and the Variant logical type thrift change is added in https://github.com/apache/parquet-format/pull/460 . 

This is to add the actual Variant logical type in Parquet-java.

### Component(s)

Core

aihuaxu:
### Rationale for this change
This is to add Variant logical type in parquet-java to be used by dependent projects.

The `Variant` logical type has been added to LogicalTypeAnnotation. For variant columns, the corresponding Parquet group is annotated as VARIANT(<version>), indicating that the variant data may be encoded according to the specified or lower version. Readers can use this version information to validate compatibility and fail early if the version is not supported.

### Are these changes tested?
Yes. 

Yes. Variant logical type is available.

Closes #3070


aihuaxu:
@Fokko, @wgtmac  Can you help check if this is right implementation? Thanks.

wgtmac:
Do we need to support writing and reading variant data?

aihuaxu:
The Variant reading and writing are getting implemented in Iceberg and/or the engines themselves. I think later we can think of pulling the implementation to Parquet if needed.

emkornfield:
> The Variant reading and writing are getting implemented in Iceberg and/or the engines themselves. I think later we can think of pulling the implementation to Parquet if needed.

I think this is problematic if the spec lives in parquet and doesn't have a complete implementation per  previously [agreed upon guidelines for new parquet features](https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#additionschanges-to-the-format).  This probably warrants a discussion on the mailing list.  CC @julienledem @rdblue @RussellSpitzer

wgtmac:
Usually we need two reference implementations for spec changes like this. I'm not sure if there is any chance to have another implementation ready in a timely manner. IMO, at least parquet-java should support basic roundtrip read and write.

aihuaxu:
I see. Per guideline, we need to have the implementation in parquet-java and then another one. Do we usually include the implementation with this annotation change or should be separate?   

> Completeness: The goal of this phase is to ensure the feature is viable, there is no ambiguity in its specification by demonstrating compatibility between implementations. Once a change has lazy consensus, two implementations of the feature demonstrating interopability must also be provided. One implementation MUST be [parquet-java](http://github.com/apache/parquet-java). It is preferred that the second implementation be [parquet-cpp](https://github.com/apache/arrow) or [parquet-rs](https://github.com/apache/arrow-rs),

wgtmac:
I think it should be in one change. The parquet-format cannot be released without concrete PoC implementation in parquet-java. Without that release, separate changes may break CI and thus cannot be merged.

aihuaxu:
@wgtmac  With https://github.com/apache/parquet-java/pull/3117/files implementing encoding/decoding, should we consider merging this separately?

wgtmac:
I think at least it needs the conversion from/to thrift definition of the variant type. So we need to wait for the release of parquet-format 2.11.0.

parquet-column/src/test/java/org/apache/parquet/parser/TestParquetParser.java
```diff
@@ -447,4 +449,30 @@ public void testEmbeddedAnnotations() {
...
+        .requiredGroup()
+        .as(LogicalTypeAnnotation.variantType())
+        .required(BINARY)
+        .named("metadata")
+        .required(BINARY)
+        .named("value")
+        .named("aVariant")
+        .named("Message");
+
+      MessageType parsed = parseMessageType(message);
```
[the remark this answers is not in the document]
aihuaxu: I added Variant to OriginalType to make the parsing work.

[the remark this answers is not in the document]
rdblue: Variant shouldn't be added to `OriginalType`. The type annotations replace `OriginalType` and we aren't adding anything else to it.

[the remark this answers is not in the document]
aihuaxu: I updated to parse group annotation rather than updating `OriginalType`.

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -1414,6 +1415,26 @@ public void testTimestampLogicalTypeWithUTCParameter() {
     Assert.assertEquals(nonUtcMicrosExpected, nonUtcMicrosActual);
   }
 
+  @Test
+  public void testVariantLogicalType() {
```
rdblue, line 1419: Can you test a shredded example as well?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -516,14 +517,19 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.Float16LogicalTypeAnnot
     }
 
     @Override
-    public Optional<LogicalType> visit(LogicalTypeAnnotation.UnknownLogicalTypeAnnotation intervalLogicalType) {
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.UnknownLogicalTypeAnnotation unknownLogicalType) {
```
rdblue, line 520: Can you move this to a separate PR?

parquet-format-structures/src/main/java/org/apache/parquet/format/LogicalTypes.java
```diff
@@ -53,4 +53,5 @@ public static LogicalType DECIMAL(int scale, int precision) {
   public static final LogicalType JSON = LogicalType.JSON(new JsonType());
   public static final LogicalType BSON = LogicalType.BSON(new BsonType());
   public static final LogicalType FLOAT16 = LogicalType.FLOAT16(new Float16Type());
+  public static final LogicalType VARIANT = LogicalType.VARIANT(new VariantType());
```
rdblue, line 56: Didn't we add a version number to the format? I think that we need to write version=1 and verify that the version is 1 when reading.

aihuaxu: Yeah. We added the version.

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
aihuaxu: I added a simple test below. Initially I added with shredded object and array but for the test here, I think there is no need to have such complexity so I just add the simple test instead.

parquet-format-structures/src/test/java/org/apache/parquet/format/TestLogicalTypes.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class TestLogicalTypes {
+  @Test
+  public void testVariantLogicalTypeVersion() {
+    LogicalType variant = LogicalTypes.VARIANT();
+    assertEquals(1, (variant.getVARIANT().getSpecification_version()));
```
rdblue, line 29: I don't really see the value of this test. This is testing that a constant has some expected value.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -524,6 +525,11 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.UnknownLogicalTypeAnnot
     public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnnotation intervalLogicalType) {
       return of(LogicalType.UNKNOWN(new NullType()));
     }
+
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
+      return of(LogicalTypes.VARIANT());
```
rdblue, line 531: The other cases create new instances of the logical type, but I agree with using the `LogicalTypes` constant. This just doesn't need to be a method call. It would also be nice to follow up with a PR that converts all of these to the correct constant rather than creating new instances.

aihuaxu: Will do.

parquet-format-structures/src/test/java/org/apache/parquet/format/TestLogicalTypes.java
aihuaxu: I was trying to make sure the version is as expected, but otherwise, it doesn't do much. Let me remove for now.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
aihuaxu: It has been addressed in https://github.com/apache/parquet-java/pull/3186.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
aihuaxu: It has been addressed in https://github.com/apache/parquet-java/pull/3186.

parquet-format-structures/src/main/java/org/apache/parquet/format/LogicalTypes.java
```diff
@@ -54,4 +54,11 @@ public static LogicalType DECIMAL(int scale, int precision) {
   public static final LogicalType BSON = LogicalType.BSON(new BsonType());
   public static final LogicalType FLOAT16 = LogicalType.FLOAT16(new Float16Type());
   public static final LogicalType UUID = LogicalType.UUID(new UUIDType());
+  public static final LogicalType VARIANT = LogicalType.VARIANT(variant());
+
+  private static final VariantType variant() {
```
rdblue, line 59: There is no need for this method to be `final`.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1183,6 +1183,8 @@ LogicalTypeAnnotation getLogicalTypeAnnotation(LogicalType type) {
         return LogicalTypeAnnotation.uuidType();
       case FLOAT16:
         return LogicalTypeAnnotation.float16Type();
+      case VARIANT:
+        return LogicalTypeAnnotation.variantType();
```
rdblue, line 1186: This needs to handle `specification_version`.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1589,6 +1590,24 @@ public void testMapConvertedTypeReadWrite() throws Exception {
     verifyMapMessageType(messageType, "map");
   }
 
+  @Test
+  public void testVariantLogicalType() {
```
rdblue, line 1594: This needs to test that `specification_version` is passed correctly.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
aihuaxu: @rdblue, @emkornfield I have some basic questions on this specification_version. We already the version (that should be this specification_version?) in the variant header.  

I can see that we define a constant of `VARIANT_SPECIFICATION_VERSION` and the writer will write that into the header. Then the reader should be able to fetch the version and parse the variant if it's supported.

Variant logical type may not  be associated with a version. Let me know if I misunderstand anything.

aihuaxu: Synced offline. The spec specifies the supported latest version while the data can be of different versions. 

With that, variant logical type will be like: `variant(1)` with 1 to be the version.

emkornfield: Sorry for the delay, this sounds right to me.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1128,6 +1140,49 @@ public int hashCode() {
     }
   }
 
+  public static class VariantLogicalTypeAnnotation extends LogicalTypeAnnotation {
+    private byte specificationVersion;
```
rdblue, line 1144: Nit: I'd prefer the shorter `specVersion`.

rdblue (approved):

Issue #3074, opened 2024-11-23
read footer using 1 call readFully(byte[8]) instead of 5 calls ( 4 x read() for footer length + 1 x read(byte[4]) for magic marker )


Arnaud-Nauwynck:
### Describe the enhancement requested

This is a minor performance improvement, but worth when reading many files.
read footer using 1 call readFully(byte[8]) instead of 5 calls ( 4 x read() for footer length + 1 x read(byte[4]) for magic marker )

in summary the patch is for file ParquetFileReader.java,  method "readFooter()" :

> -    byte[] magic = new byte[MAGIC.length];
> -    long fileMetadataLengthIndex = fileLen - magic.length - FOOTER_LENGTH_SIZE;
> +    long fileMetadataLengthIndex = fileLen - MAGIC.length - FOOTER_LENGTH_SIZE;
>      LOG.debug("reading footer index at {}", fileMetadataLengthIndex);
>      f.seek(fileMetadataLengthIndex);
> -    int fileMetadataLength = readIntLittleEndian(f);
> -    f.readFully(magic);
> +    byte[] magicAndLengthBytes = new byte[FOOTER_LENGTH_SIZE + MAGIC.length];
> +    f.readFully(magicAndLengthBytes);
> +    int fileMetadataLength = readIntLittleEndian(magicAndLengthBytes, 0);
> 

### Component(s)

Core

PR #3075, opened 2024-11-23
GH-3074: read footer using 1 call readFully(byte[8]) instead of 5 calls

Arnaud-Nauwynck:
GH-3074: read footer using 1 call readFully(byte[8]) instead of 5 calls

### Rationale for this change
performance

only method FilePargetReader.readFooter()

no


wgtmac (commented):
Thanks for the improvement! Please make the CI happy.

wgtmac:
Could you try to rebase it? I think the CI failure is due to based on a problematic commit.

steveloughran (commented):
LGTM, at least for now.

Parquet 3.0 is going to add some bytes between the metadata length and MAGIC; I don't know if there is anything extra which could be done here.

I think we could start with some test files which already have this new marker and use parsing that as regression testing

PR #3079, opened 2024-11-26
GH-3078: Use Hadoop FileSystem.openFile() to open files

### LINKED ISSUE #3078: Use Hadoop FileSystem.openFile() to open files

Use the hadoop `openFile()` builder API to open a file, passing down
* the existing FileStatus (saves a HEAD on cloud stores)
* a read policy appropriate for parquet files. This ensures that the object stores know what kind of GET/prefect/cache to do.

### Component(s)

Core

steveloughran:
### Rationale for this change

If parquet passes down the `FileStatus` it previously collected, S3A and ABFS clients can skip their own HEAD probe. This saves time, money and IO capacity.

When the read policy is passed in, the clients know immediately what kind of read pattern to expect, rather than infer it as reads take place or from the file suffix.

* Open files with FileSystem.openFile(), passing in file status
* And read policy of "parquet, vector, random, adaptive"

### Are these changes tested?

Through parquet-hadoop.

no.

Closes #3078

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
@@ -110,7 +110,7 @@ public class HadoopInputFile implements InputFile {
     FSDataInputStream stream;
     try {
       // this method is async so that implementations may do async HEAD head
-      // requests. Not done in S3A/ABFS when a file status passed down (as is done here)
+      // requests, such as S3A/ABFS when a file status is passed down.
       final CompletableFuture<FSDataInputStream> future = fs.openFile(stat.getPath())
           .withFileStatus(stat)
           .opt(OPENFILE_READ_POLICY_KEY, PARQUET_READ_POLICY)
@@ -121,7 +121,14 @@ public class HadoopInputFile implements InputFile {
       // equal the path in the FileStatus; Hive virtual FS could create this condition.
       // As the path to open is derived from stat.getPath(), this condition seems
       // near-impossible to create -but is handled here for due diligence.
-      stream = fs.open(stat.getPath());
+      try {
+        stream = fs.open(stat.getPath());
+      } catch (IOException | RuntimeException ex) {
+        // failure on this attempt attaches the failure of the openFile() call
+        // so the stack trace is preserved.
+        ex.addSuppressed(e);
+        throw ex;
+      }
     }
 
     return HadoopStreams.wrap(stream);
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
@@ -19,8 +19,13 @@
 
 package org.apache.parquet.hadoop.util;
 
+import static org.apache.parquet.hadoop.util.wrapped.io.FutureIO.awaitFuture;
+
 import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.util.concurrent.CompletableFuture;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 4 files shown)


wgtmac:
@Fokko @gszadovszky Could you help review this? Thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/wrapped/io/FutureIO.java
```diff
@@ -70,6 +70,29 @@ public static <T> T awaitFuture(final Future<T> future, final long timeout, fina
...
+   * @throws InterruptedIOException future was interrupted
+   * @throws IOException if something went wrong
+   * @throws RuntimeException any nested RTE thrown
+   */
+  public static <T> T awaitFuture(final Future<T> future)
+      throws InterruptedIOException, IOException, RuntimeException {
+    try {
+      return future.get();
+    } catch (InterruptedException e) {
+      throw (InterruptedIOException) new InterruptedIOException(e.toString()).initCause(e);
```
gszadovszky, line 91: nit: Why the cast?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -70,9 +93,38 @@ public long getLength() {
...
+          .withFileStatus(stat)
+          .opt(OPENFILE_READ_POLICY_KEY, PARQUET_READ_POLICY)
+          .build();
+      stream = awaitFuture(future);
+    } catch (RuntimeException e) {
+      // S3A < 3.3.5 would raise illegal path exception if the openFile path didn't
+      // equal the path in the FileStatus; Hive virtual FS could create this condition.
+      // As the path to open is derived from stat.getPath(), this condition seems
+      // near-impossible to create -but is handled here for due diligence.
+      stream = fs.open(stat.getPath());
```
gszadovszky, line 124: Shouldn't we at least log the original exception?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/wrapped/io/FutureIO.java
steveloughran: `initCause()` returns a throwable; you need to case it back to what the actual exception is

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
steveloughran: good point. at debug? I'm never a fan of excessively noisy libraries as they ruin the lives of people downstream, especially if messages for developers end up being printed in user logs far too often. In fact, I'm not above dynamically patching log levels to shut those libraries up (HADOOP-19272)...

stack trace then, debug.
error text itself? Probably the same unless there's a desire to log exactly once during the life of a process, which could be done trivially

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/wrapped/io/FutureIO.java
gszadovszky: Oh, missed that, thanks!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
gszadovszky: I'm fine with any solutions just did not want to loose the exception may occur.

Another idea I've had is to catch the potential exception may come out of this line (let's call it `e2`) and add the previously caught `e` as suppressed to `e2`, then re-throw. But it might be an overkill since this code path "near-impossible" to be taken.

gszadovszky: Sounds good!

steveloughran:
I'm going to propose the following tests to simulate failure and validate handling.

### openFile() raises IllegalArgumentException, open() works.

open() succeeds so stream returned.


### openFile() raises IllegalArgumentException, open() raises IOE

IOE thrown with IllegalArgumentException as suppressed

### openFile() returns a Future, whose get() raises an UncheckedIOE

unchecked IOE unwrapped and thrown.


### openFile() returns a stream, open() raises IOE

stream returned, showing open() wasn't invoked.

Does that cover the codepaths?

gszadovszky:
At the last scenario, the "open() raises IOE" doesn't matter, right? I mean, it can be a test coding detail to throw an exception there so we are sure open() is not invoked, but this is not part of the actual scenario. (We might use Mockito to check if the method was invoked or not, no need for an exception.)

Otherwise, yes, I believe, these are all the scenarios might happen in this code.

steveloughran:
> We might use Mockito to check if the method was invoked or not, no need for an exception

Good point, and yes same codepath.

My feelings about mockito are _mixed_ fwiw; tests can be very brittle and often have high maintenance costs. As this is public API stuff, not only shouldn't it be an issue -if it does fail, that's a regression.

gszadovszky (approved):
Looks nice and clear. Thanks for the PR!

Fokko (approved):
LGTM, appreciate the Javadoc/comments, really helps to add Hadoop context

steveloughran:
shaves a HEAD request!

for s3a it tells things to seek properly rather than having to guess afterwards.

FWIW there's a "whole-file" read policy, we use this in hadoop itself for stuff like distcp...now I need to get the sequential policy into avro so it knows that prefetching is good, rather than a waste of IO capacity

Issue #3083, opened 2024-11-27
Make DELTA_LENGTH_BYTE_ARRAY default encoding for binary values


raunaqmorarka:
### Describe the enhancement requested

The current default for V1 pages is PLAIN encoding. This encoding mixes string length with string data. This is inefficient for for skipping N values, as the encoding does not allow random access. It's also slow to decode as the interleaving of lengths with data does not allow efficient batched implementations and forces most implementations to make copies of the data to fit the usual representation of separate offsets and data for strings.

DELTA_LENGTH_BYTE_ARRAY has none of the above problems as it separates offsets and data. The parquet-format spec also seems to recommend this
https://github.com/apache/parquet-format/blob/c70281359087dfaee8bd43bed9748675f4aabe11/Encodings.md?plain=1#L299
```
### Delta-length byte array: (DELTA_LENGTH_BYTE_ARRAY = 6)

Supported Types: BYTE_ARRAY

This encoding is always preferred over PLAIN for byte array columns.
```

V2 pages use DELTA_BYTE_ARRAY as the default encoding, this is an improvement over PLAIN but adds complexity which makes it slower to decode than DELTA_LENGTH_BYTE_ARRAY with the potential benefit of lower storage requirements.

JMH benchmarks in Trino's parquet reader at `io.trino.parquet.reader.BenchmarkBinaryColumnReader` showed that DELTA_LENGTH_BYTE_ARRAY can be decoded at over 5X speed and DELTA_BYTE_ARRAY at over 2X the speed of decoding PLAIN encoding.
Given the above recommendation of parquet-format spec and significant performance difference, the reference implementation here should be updated to use DELTA_LENGTH_BYTE_ARRAY by default.

### Component(s)

Core

PR #3085, opened 2024-11-28
GH-3083: Make DELTA_LENGTH_BYTE_ARRAY default encoding for binary

raunaqmorarka:
### Rationale for this change
The current default for V1 pages is PLAIN encoding. This encoding mixes string length with string data. This is inefficient for for skipping N values, as the encoding does not allow random access. It's also slow to decode as the interleaving of lengths with data does not allow efficient batched implementations and forces most implementations to make copies of the data to fit the usual representation of separate offsets and data for strings.

DELTA_LENGTH_BYTE_ARRAY has none of the above problems as it separates offsets and data. The parquet-format spec also seems to recommend this https://github.com/apache/parquet-format/blob/c70281359087dfaee8bd43bed9748675f4aabe11/Encodings.md?plain=1#L299

- Changes the default fallback encoding for BINARY columns in the V1 writer from PLAIN to DELTA_LENGTH_BYTE_ARRAY.
- Adds a new configuration property `parquet.enable.delta-length-byte-array-for-binary` (default: `true`) to allow disabling the new default and reverting to PLAIN encoding.
- The toggle is wired through `ParquetProperties.Builder.withDeltaLengthByteArrayForBinary(boolean)`, `ParquetWriter.Builder`, and `ParquetOutputFormat` (Hadoop configuration).

### Are these changes tested?
Yes. Existing tests updated to reflect the new default, and two new test cases added for the disabled configuration path (`testBinary_NoDeltaLengthByteArray` and `testBinary_NoDict_NoDeltaLengthByteArray`).

Yes. BINARY columns written with `PARQUET_1_0` writer version now use DELTA_LENGTH_BYTE_ARRAY encoding by default instead of PLAIN when dictionary encoding falls back. Users can set `parquet.enable.delta-length-byte-array-for-binary=false` to revert to the previous behavior.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -67,6 +67,7 @@ public class ParquetProperties {
   public static final int DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER = 5;
   public static final boolean DEFAULT_STATISTICS_ENABLED = true;
   public static final boolean DEFAULT_SIZE_STATISTICS_ENABLED = true;
+  public static final boolean DEFAULT_DELTA_LENGTH_BYTE_ARRAY_FOR_BINARY = true;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
@@ -133,6 +134,7 @@ public class ParquetProperties {
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
   private final ColumnProperty<ByteStreamSplitMode> byteStreamSplitEnabled;
+  private final boolean deltaLengthByteArrayForBinary;
   private final Map<String, String> extraMetaData;
   private final ColumnProperty<Boolean> statistics;
   private final ColumnProperty<Boolean> sizeStatistics;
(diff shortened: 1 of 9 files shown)


raunaqmorarka:
> Hey @raunaqmorarka thanks for raising this. I think we want to [discuss on the devlist](https://lists.apache.org/list.html?dev@parquet.apache.org) first if we want to change behavior. Would you be interested to raise this?

I'm not sure how to start a discussion on the devlist, I don't have credentials to login there.
It would be nice to discuss on the GH issue https://github.com/apache/parquet-java/issues/3083 if that's possible

wgtmac:
@raunaqmorarka You can send an email to dev-subscribe@parquet.apache.org to subscribe. If you don't want to subscribe, you may directly send an email to dev@parquet.apache.org. You can see https://lists.apache.org/list.html?dev@parquet.apache.org for reference.

jhorstmann:
In my opinion the default should not be changed, but it would be really useful to allow users to configure the encoding per column, similar to how `BYTE_STREAM_SPLIT` is now handled in [`ParquetProperties`](https://github.com/apache/parquet-java/blob/master/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java#L500).

There might still be parquet readers with incomplete support for delta encodings, a simple version update could otherwise lead to problems which are only noticed much later.

raunaqmorarka:
@Fokko @wgtmac @jhorstmann Trino has recently made changes to start writing DELTA_LENGTH_BYTE_ARRAY by default https://github.com/trinodb/trino/pull/29246 based on the benchmark results there.
I have updated this PR to include a flag that will allow disabling DELTA_LENGTH_BYTE_ARRAY by default for any old readers that may still not support this. I believe all the major parquet projects are supporting reading this encoding by now.

PR #3096, opened 2024-12-03
GH-3080: HadoopStreams to support ByteBufferPositionedReadable

### LINKED ISSUE #3080: HadoopStreams to support ByteBufferPositionedReadable input streams

If a stream declares in its StreamCapabilities that it supports
ByteBufferPositionedReadable, then use it for `readFully(ByteBuffer)`
All streams in Hadoop 3.0.0 + do declare this.

+ use StreamCapabilities to look for `ByteBufferReadable`.

For detecting ByteBufferReadable, use this probe falling back to the recursive scan.
All streams in the hadoop codebase will report this via StreamCapabilities, but there
may be some third-party streams which do not. 

### Component(s)

_No response_

steveloughran:
### Rationale for this change

If a stream declares in its StreamCapabilities that it supports
`ByteBufferPositionedReadable`, then use that API for
`readFully(ByteBuffer)`

```
ByteBufferPositionedReadable.readFully(long position, ByteBuffer buf)
```

Adding support for Hadoop `ByteBufferPositionedReadable` streams may improve performance
by pushing retry/recovery logic into the filesystem client library.

This interface is implemented by the HDFS input stream; we are considering adding
it elsewhere.

* New SeekableInputStream implementation: `H3ByteBufferInputStream`
* Instantiated in `HadoopStreams` if the `FSDataInputStream` is considered suitable.
* Tests for the new behavior and that no regressions are caused.

#### Class `H3ByteBufferInputStream`

The reading is done in a new class, `H3ByteBufferInputStream`, which subclasses `H2ByteBufferInputStream`. This reduces the amount of duplicate code, it just makes it a bit unclean.

The purist way to do it would be to create an abstract superclass `HadoopInputStream` to hold all commonality between the the three input streams.

I'm happy to do this, just didn't want to doing some larger refactoring without (a) showing the core design worked and (b) getting permission to do this. Should I do this?

#### `HadoopStreams` changes

Selection of the new input stream is done if and only if the stream declares the capability `in:preadbytebuffer`.
There is no equivalent of `isWrappedStreamByteBufferReadable()` which recurses through
a chain of wrapped streams looking for the API.
If a stream doesn't declare its support for the API, it won't get picked up.
This is done knowing that the sole production implemenation which currently exists, 
the HDFS input stream, does declare this capability.

### Are these changes tested?

There is new test suite, for new behavior and ensuring that the integration with
HadoopStreams still retains the correct behavior for existing streams.
Suite is parameterized on heap and direct buffers.

No

Closes GH-3080

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
@@ -100,6 +100,10 @@ class H2SeekableInputStream extends DelegatingSeekableInputStream {
     VectorIoBridge.instance().readVectoredRanges(stream, ranges, allocator);
   }
 
+  protected Reader getReader() {
+    return reader;
+  }
+
   public static void readFully(Reader reader, ByteBuffer buf) throws IOException {
     // unfortunately the Hadoop 2 APIs do not have a 'readFully' equivalent for the byteBuffer read
     // calls. The read(ByteBuffer) call might read fewer than byteBuffer.hasRemaining() bytes. Thus we
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
@@ -59,19 +59,23 @@ class H3ByteBufferInputStream extends H2SeekableInputStream {
 
   /**
    * Read the buffer fully through use of {@code ByteBufferPositionedReadable.readFully()}
-   * at the current location.
+   * from the current location.
+   * That is it reads from stream[pos] to stream[pos + buf.remaining() -1]
    *
    * @param buf a byte buffer to fill with data from the stream
+   * @return number of bytes read.
+   *
    * @throws EOFException the buffer length is greater than the file length
    * @throws IOException other IO problems.
    */
   // Visible for testing
-  static void performRead(final FSDataInputStream stream, final ByteBuffer buf) throws IOException {
+  static int performRead(final FSDataInputStream stream, final ByteBuffer buf) throws IOException {
     // remember the current position
     final long pos = stream.getPos();
     final int size = buf.remaining();
     stream.readFully(pos, buf);
     // then move read position on afterwards.
     stream.seek(pos + size);
+    return size;
   }
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
@@ -54,8 +54,19 @@ class H3ByteBufferInputStream extends H2SeekableInputStream {
    */
   @Override
   public void readFully(final ByteBuffer buf) throws EOFException, IOException {
-    // use ByteBufferPositionedReadable.readFully()
-    final FSDataInputStream stream = getStream();
+    performRead(getStream(), buf);
+  }
+
+  /**
+   * Read the buffer fully through use of {@code ByteBufferPositionedReadable.readFully()}
+   * at the current location.
+   *
+   * @param buf a byte buffer to fill with data from the stream
+   * @throws EOFException the buffer length is greater than the file length
+   * @throws IOException other IO problems.
+   */
+  // Visible for testing
+  static void performRead(final FSDataInputStream stream, final ByteBuffer buf) throws IOException {
     // remember the current position
     final long pos = stream.getPos();
     final int size = buf.remaining();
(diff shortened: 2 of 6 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -111,14 +79,8 @@ private static Function<FSDataInputStream, SeekableInputStream> unwrapByteBuffer
...
   private static Boolean isWrappedStreamByteBufferReadable(FSDataInputStream stream) {
-    if (hasCapabilitiesMethod.isNoop()) {
-      // When the method is not available, just return a null
-      return null;
-    }
-
-    boolean isByteBufferReadable = hasCapabilitiesMethod.invoke(stream, "in:readbytebuffer");
 
-    if (isByteBufferReadable) {
+    if (stream.hasCapability(READBYTEBUFFER)) {
```
wgtmac, line 83: We can avoid any reflection here because of the Hadoop version bump?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopStreams.java
```diff
@@ -53,42 +49,14 @@ public class HadoopStreams {
...
-   * Is the inner stream byte buffer readable?
-   * The test is 'the stream is not FSDataInputStream
-   * and implements ByteBufferReadable'
-   * <p>
-   * This logic is only used for Hadoop <2.9.x, and <3.x.x
-   *
-   * @param stream stream to probe
-   * @return A H2SeekableInputStream to access, or H1SeekableInputStream if the stream is not seekable
-   */
-  private static Function<FSDataInputStream, SeekableInputStream> unwrapByteBufferReadableLegacy(
```
wgtmac, line 81: Is there any behavior change of a wrapped stream after removing `unwrapByteBufferReadableLegacy`?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H2SeekableInputStream.java
```diff
@@ -100,6 +100,10 @@ public void readVectored(List<ParquetFileRange> ranges, ByteBufferAllocator allo
     VectorIoBridge.instance().readVectoredRanges(stream, ranges, allocator);
   }
 
+  protected Reader getReader() {
```
wgtmac, line 103: Why adding this but not used elsewhere?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Read the buffer fully through use of {@code ByteBufferPositionedReadable.readFully()}
+   * from the current location.
+   * That is it reads from stream[pos] to stream[pos + buf.remaining() -1]
+   *
+   * @param buf a byte buffer to fill with data from the stream
+   * @return number of bytes read.
+   *
+   * @throws EOFException the buffer length is greater than the file length
```
wgtmac, line 68: Add `EOFException` to the method signature?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/H3ByteBufferInputStream.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import org.apache.hadoop.fs.FSDataInputStream;
+
+/**
+ * Class which implements {@link #readFully(ByteBuffer)} through
+ * {@code ByteBufferPositionedReadable.readFully()}.
+ * <p>This is implemented by HDFS and possibly other clients,
+ */
+class H3ByteBufferInputStream extends H2SeekableInputStream {
```
wgtmac, line 31: I'm fine with the inheritance to reduce code duplication.

wgtmac (commented):
Thanks for adding this! I've left some comments.

steveloughran:
I'm away until 2025; will reply to comments then. Thanks for the review.

steveloughran:
I'm back, don't think I've forgotten this. In fact I've been actually setting up a test-only-loop for hadoop for regression testing parquet support through the cloud connectors.
https://github.com/apache/hadoop/pull/7285

steveloughran:
ah, did neglect this didn't I? will revisit.

Fokko:
@steveloughran Is this something we want to get in 1.17.0? See https://lists.apache.org/thread/g1cngnkzhjt86yt4dfl078yrplfmzcf5

steveloughran:
forgotten about this; too many other things. I don't see myself having the time to work on this in the next few weeks i'm afraid

steveloughran:
ok, did get to this!

- rebase
- move all tests to java 5 + assertj

Fokko (approved):

wgtmac (approved):

wgtmac:
Thanks @steveloughran for adding this and @Fokko for the review! Long time to merge this.

Issue #1452, opened 2014-07-24
Add support for repeated columns in the filter2 API


asfimport:
They currently are not supported. They would need their own set of operators, like contains() and size() etc.

**Reporter**: [Alex Levenson](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=alexlevenson) / @isnotinvain
#### PRs and other links:
- [GitHub Pull Request #1328](https://github.com/apache/parquet-java/pull/1328)
- [GitHub Pull Request #1370](https://github.com/apache/parquet-java/pull/1370)
- [GitHub Pull Request #1328](https://github.com/apache/parquet-mr/pull/1328)

<sub>**Note**: *This issue was originally created as [PARQUET-34](https://issues.apache.org/jira/browse/PARQUET-34). Please see the [migration documentation](https://issues.apache.org/jira/browse/PARQUET-2502) for further details.*</sub>

asfimport:
[Viktor Szathmáry](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14143934) / @phraktle:
Any plans on supporting this?

asfimport:
[Alex Levenson](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14144019) / @isnotinvain:
We'd like to, contains() at least shouldn't be too hard, it may not even require any changes except removing the check that prevents it.

That said we don't have any concrete plans to do it right away – pull requests always welcome and we're happy to help!

asfimport:
[Flavio Pompermaier](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14627779):
Repeated columns (protobuf) and lists (thrift) support would be VERY useful..

asfimport:
[Stefano Bortoli](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14627810):
I agree with Flavio. +1 for supporting multi-valued columns.

asfimport:
[Nicola](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14738942):
+1

asfimport:
[Julien Le Dem](https://issues.apache.org/jira/browse/PARQUET-34?#comment-14992586) / @julienledem:
@alexlevenson did you have any plan for this?

asfimport:
[Flavio Pompermaier](https://issues.apache.org/jira/browse/PARQUET-34?#comment-15000213):
If someone could give me some advice I could try to implement it by myself (I never looked into Parquet code...)

[...]

PR #3098, opened 2024-12-05
GH-1452: implement Size() filter for repeated columns

clairemcginty:
### Rationale for this change

this PR continues the work outlined in #1452. It implements a `size()` predicate for filtering on # of elements in repeated fields:

```java
FilterPredicate hasThreeElements = size(intColumn("my_list_field"), Operators.Size.Operator.EQ, 3)
```

`Size()` and `not(size())` implemented for all list fields with **`required` element type**. Attempting to filter on a list of optional elements will throw an exception in the schema validator. This is because the existing record-level filtering setup (`IncrementallyUpdatedFilterPredicateEvaluator`) only feeds in non-null values to the `ValueInspectors`. thus if you had an array [1,2, null, 4] it would only count 3 elements. I can file a ticket to support this eventually but I think we'd have to rework  the FilteringRecordMaterializer to be aware of repetition/definition levels.

The list group itself can be `optional` or `required`. Null lists are treated as having size 0. Again, this is due to difficulty disambiguating them at the record-level filtering step. (Would love feedback on both these design decisions!!)

### Are these changes tested?
Unit tests + tested a snapshot build locally with real datasets

New Operators API

Part of #1452


parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -217,6 +219,70 @@ public <T extends Comparable<T>> Boolean visit(Contains<T> contains) {
...
+    // Given the total number of elements and non-null fields, we can compute the max size of any array field
+    long maxArrayElementCount = 1 + (nonNullElementCount - numNonNullRecords);
+    final boolean blockCannotMatch = size.filter(
+        (eq) -> eq > maxArrayElementCount,
+        (lt) -> false,
+        (lte) -> false,
+        (gt) -> gt >= maxArrayElementCount,
+        (gte) -> gte > maxArrayElementCount);
+
+    return blockCannotMatch ? BLOCK_CANNOT_MATCH : BLOCK_MIGHT_MATCH;
```
clairemcginty, line 283: hopefully this is a faithful transcription of the logic outlined here: https://github.com/apache/parquet-java/issues/1452#issuecomment-2271914678

parquet-hadoop/src/test/java/org/apache/parquet/filter2/statisticslevel/TestStatisticsFilter.java
```diff
@@ -435,6 +468,174 @@ public void testOr() {
...
+    assertTrue(canDrop(size(nestedListColumn, Operators.Size.Operator.GTE, 1), columnMeta));
+    assertTrue(canDrop(size(nestedListColumn, Operators.Size.Operator.EQ, 5), columnMeta));
+
+    // These predicates should not be able to filter out the page
+    assertFalse(canDrop(size(nestedListColumn, Operators.Size.Operator.LTE, 1), columnMeta));
+    assertFalse(canDrop(size(nestedListColumn, Operators.Size.Operator.LT, 1), columnMeta));
+    assertFalse(canDrop(size(nestedListColumn, Operators.Size.Operator.EQ, 0), columnMeta));
+  }
+
+  private static SizeStatistics createSizeStatisticsForRepeatedField(
```
clairemcginty, line 588: I'm dynamically generating SizeStatistics for each test case which does add a lot of LOC to the file--I could also just replace it with the computed `SizeStatistics` for each test case if that's simpler. I just wrote it this way originally because I wasn't that confident in my ability to translate the striping algorithm by hand for all these cases 😅

parquet-column/src/main/java/org/apache/parquet/filter2/recordlevel/IncrementallyUpdatedFilterPredicate.java
```diff
@@ -223,6 +224,74 @@ public void reset() {
...
 
+  class CountingValueInspector extends ValueInspector {
+    private long observedValueCount;
+    private final ValueInspector delegate;
+    private final Function<Long, Boolean> shouldUpdateDelegate;
+
+    public CountingValueInspector(ValueInspector delegate, Function<Long, Boolean> shouldUpdateDelegate) {
+      this.observedValueCount = 0;
+      this.delegate = delegate;
+      this.shouldUpdateDelegate = shouldUpdateDelegate;
```
clairemcginty, line 235: note: The "shouldUpdateDelegate" is needed since don't want to terminate prematurely with a false positive. For example if we're filtering on `size(eq(3))` but the input array has 4 elements, we want to prevent the delegated `Eq` from returning true after it hits the third element because it thinks the condition is satisfied.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -378,6 +379,11 @@ public <T extends Comparable<T>> PrimitiveIterator.OfInt visit(Contains<T> conta
           indices -> IndexIterator.all(getPageCount()));
     }
 
+    @Override
+    public PrimitiveIterator.OfInt visit(Size size) {
+      return IndexIterator.all(getPageCount());
```
clairemcginty, line 384: `repetitionLevelHistogram` and `definitionLevelHistogram` are both in scope here, should I repeat the logic from `StatisticsFilter` or is that completely redundant?

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
```diff
@@ -493,6 +494,39 @@ public <T extends Comparable<T>> Boolean visit(Contains<T> contains) {
...
+        return BLOCK_MIGHT_MATCH;
+      }
+      int numDistinctValues = dict.size();
+      final boolean blockCannotMatch = size.filter(
+          (eq) -> eq < numDistinctValues,
+          (lt) -> lt <= numDistinctValues,
+          (lte) -> lte < numDistinctValues,
+          (gt) -> false,
+          (gte) -> false);
+
```
clairemcginty, line 521: actually now that I think about it, this isn't accurate, since we don't know the distribution of values. I guess we could combine it with SizeStatistics to get the number of elements and work out the minimum size from there.

clairemcginty:
> Thanks for adding this! This is a large PR that I need to take some time to review.

thanks, no rush on reviewing it! 👍

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -217,6 +219,70 @@ public <T extends Comparable<T>> Boolean visit(Contains<T> contains) {
...
   }
 
+  @Override
+  public Boolean visit(Size size) {
+    final ColumnChunkMetaData metadata = getColumnChunk(size.getColumn().getColumnPath());
+    if (metadata == null) {
+      // the column isn't in this file, so fail eq/gt/gte targeting size > 0
+      final boolean blockCannotMatch =
+          size.filter((eq) -> eq > 0, (lt) -> false, (lte) -> false, (gt) -> gt >= 0, (gte) -> gte > 0);
+      return blockCannotMatch ? BLOCK_CANNOT_MATCH : BLOCK_MIGHT_MATCH;
```
wgtmac, line 229: Shouldn't it always return `BLOCK_CANNOT_MATCH`? What is the value of `size(null list)`? null or 0?

wgtmac: If the behavior depends on SQL dialect, perhaps we should be conservative to return `BLOCK_MIGHT_MATCH` instead.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
```diff
@@ -217,6 +219,70 @@ public <T extends Comparable<T>> Boolean visit(Contains<T> contains) {
...
+    }
+
+    final SizeStatistics stats = metadata.getSizeStatistics();
```
wgtmac, line 232: Should we check whether the list type can be supported before proceeding? For example, perhaps we don't want to support legacy LIST-annotated 2-level structure and unannotated repeated field according to https://github.com/apache/parquet-format/pull/466.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/dictionarylevel/DictionaryFilter.java
wgtmac: I don't think we can trust `numDistinctValues` because a row group might contain only a subset of entries from the dictionary.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
wgtmac: Repeating the same logic is necessary if you want to support page-level filtering.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
wgtmac: IIUC, these are true for only un-nested list type, right?

wgtmac (commented):
Thanks for the effort! I just took an initial pass on it and left a couple of questions.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -505,6 +505,82 @@ public <R> R filter(
     }
   }
 
+  public static final class Size implements FilterPredicate, Serializable {
+    public enum Operator {
```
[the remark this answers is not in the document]
Fokko: I was thinking the same for the `LogicalInverter` above

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -505,6 +505,82 @@ public <R> R filter(
...
+      EQ,
+      LT,
+      LTE,
+      GT,
+      GTE
```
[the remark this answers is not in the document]
Fokko: I think every project uses a slightly different abbreviation: https://github.com/apache/iceberg/blob/c0bd4bfbceeaf3cb6e4ead675fcb47232361af3c/api/src/main/java/org/apache/iceberg/transforms/ProjectionUtil.java#L42-L61 👯

clairemcginty:
> Thanks for the effort! I just took an initial pass on it and left a couple of questions.

Thanks for the review!! I should have time to address everything early next week at the latest 👍

wgtmac:
BTW, the level histogram might not be available when max_level is 0 because there is only single level (i.e. 0) and its count can be deduced from `num_values` of the column chunk or page. It will complicate the size filter here.

parquet-column/src/main/java/org/apache/parquet/filter2/predicate/Operators.java
```diff
@@ -505,6 +505,82 @@ public <R> R filter(
...
+    }
+
+    private final Column<?> column;
+    private final Operator operator;
+    private final long value;
```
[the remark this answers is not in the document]
clairemcginty: yes good point! updated

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
wgtmac: Yes, that is how I tried to print the per-page histogram: https://github.com/apache/parquet-java/pull/3126/files#diff-5469df3d4b1fc51a21a1c341964c046789df8a9e7a487926a10a1c26eb4915b6R185

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
clairemcginty: good question... I updated the unit tests to test out various nesting levels and they all passed: https://github.com/apache/parquet-java/pull/3098/commits/39981e9f8eeec2f8492fbfcb35e997aa299af18d

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
clairemcginty: thanks for the pointer! I took a pass at implementing this for ColumnIndexBuilder. though my logic ended up slightly different, since I think we need to take into account any previous null pages when computing per-page offsets? https://github.com/apache/parquet-java/pull/3098/commits/9586427b780bf28e274949448dfc4ebbe39284cc#diff-5469df3d4b1fc51a21a1c341964c046789df8a9e7a487926a10a1c26eb4915b6R511-R525

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
clairemcginty: yes I think that makes sense. Maybe we can add support for those at a later date... but this PR is already complex enough as-is, I think. I'll add a check!

clairemcginty: We could also offer _only_ record-level filtering (`IncrementallyUpdatedFilterPredicate`) for 2-level and unannotated lists, since that doesn't depend on any particular list implementation. We'd just skip StatisticsFilter/BloomFilter/DictionaryFilter/ColumnIndexFilter. That way `size()` would still work for any kind of list--it would just be less less efficient.

parquet-hadoop/src/main/java/org/apache/parquet/filter2/statisticslevel/StatisticsFilter.java
clairemcginty: hey @wgtmac , quick (maybe?) question about testing nested lists...

`ExampleParquetWriter` doesn't support non-primitive `element` fields so I switched to writing directly with `RecordConsumer`, similar to how [this test](https://github.com/apache/parquet-java/blob/apache-parquet-1.15.0/parquet-thrift/src/test/java/org/apache/parquet/hadoop/thrift/TestArrayCompatibility.java#L152-L185) is set up. However... switching from `ExampleParquetWriter` to `RecordConsumer` produces a different repetition histogram for the same element values 🤔

If you have time, could you take a quick look at this repro and let me know if my test setup looks ok? From what I understand, the repetition histogram produced by `ExampleParquetWriter` looks correct and the `RecordConsumer` looks wrong: https://github.com/apache/parquet-java/pull/3098/commits/3b46d8f7534660dd97e970039c7a2ea27c114b4e

PR #3101, opened 2024-12-09
GH-3091: Add verification guide and .rat-excludes.txt for release

### LINKED ISSUE #3091: Add documentation about how to verify the release

I wasn't able to find notes on how to verify the release and followed the example and some steps provided on the Apache Doris wiki: https://github.com/apache/doris/wiki/How-to-verify-Apache-Release

I might have missed where this is documented but if it's missing, it might be interesting to add some documentation. I am happy to add some info about it. In any case would be good to add that link to the voting thread email to encourage people to vote.

### Component(s)

_No response_

raulcd:
### Rationale for this change

There is currently no guide on how to verify. Having a guider will encourage people to verify the release.
This guide should be linked on the Voting thread.

Added documentation about how to verify and add `.rat-excludes.txt` file with the files that do not contain a current license but we are fine with those. Those files are mainly testing `.avsc`, `.parquet` , `.gitignore` and `PULL_REQUEST_TEMPLATE.md` files.

### Are these changes tested?

I've validated all steps localy

No, just more docs.

Closes #3091

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -91,3 +91,57 @@ Merge hash: 485658a5
 Would you like to pick 485658a5 into another branch? (y/n):
 ```
 For now just say n as we have 1 branch
+
+# Release Verification
+
+The Apache Arrow Release Approval process follows the guidelines defined at the
+`Apache Software Foundation Release Approval <https://www.apache.org/legal/release-policy.html#release-approval>`_.
+
+For a release vote to pass, a minimum of three positive binding votes and more
+positive binding votes than negative binding votes MUST be cast.
+Releases may not be vetoed. Votes cast by PMC members are binding, however,
+non-binding votes are greatly encouraged and a sign of a healthy project.
+
+In order to cast a vote individuals are expected to follow the following steps.
+
+## Download source package, signature file, hash file and KEYS
+
+The Release candidate will be present at `https://dist.apache.org/repos/dist/dev/parquet/`.
+The RC folder will depend on the version and the release candidate id. See the following example files for
+Apache Parquet 1.15.0 RC 1:
+```
+wget https://dist.apache.org/repos/dist/dev/parquet/apache-parquet-1.15.0-rc1/apache-parquet-1.15.0.tar.gz
+wget https://dist.apache.org/repos/dist/dev/parquet/apache-parquet-1.15.0-rc1/apache-parquet-1.15.0.tar.gz.asc
+wget https://dist.apache.org/repos/dist/dev/parquet/apache-parquet-1.15.0-rc1/apache-parquet-1.15.0.tar.gz.sha512
+wget https://dist.apache.org/repos/dist/release/parquet/KEYS
+```
+
+## Verify signature and hash
+
+GnuPG is recommended, which can install by yum install gnupg or apt-get install gnupg.
+
+```
+gpg --import KEYS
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -498,6 +498,7 @@
           <consoleOutput>true</consoleOutput>
           <excludes>
```
raulcd, line 499: I am pretty sure we can change `excludes` for `excludesFile` (https://creadur.apache.org/rat/apache-rat-plugin/rat-mojo.html#excludesFile) but I am unsure why some of the regex on the individual excludes don't seem to work with excludesFile when I ran:
`java  -jar apache-rat-0.16.1/apache-rat-0.16.1.jar -a -d apache-parquet-1.15.0.tar.gz -E $PARQUET_SRC_FOLDER/.rat-excludes.txt`
I'll investigate how to consolidate those two lists

dev/README.md
```diff
@@ -91,3 +91,61 @@ Merge hash: 485658a5
 Would you like to pick 485658a5 into another branch? (y/n):
 ```
 For now just say n as we have 1 branch
+
+# Release Verification
```
wgtmac, line 95: Thanks for adding this! Is it better to add this to the parquet site: https://github.com/apache/parquet-site/blob/production/content/en/docs/Contribution%20Guidelines/releasing.md?

wgtmac:
cc @Fokko @gszadovszky

dev/README.md
gszadovszky: I agree the "release verification" should be moved to the `parquet-site` repo instead. Then, we even can have a link to this section in the VOTE email template.

I'm not sure why we need to check for the license headers separately in the tarball. It is already in the build process so we shall not have license header issues in the repo. What I usually do instead is comparing the content of the tarball with a freshly cloned repo set to the release RC tag. There should be no differences.

raulcd: > I'm not sure why we need to check for the license headers separately in the tarball. It is already in the build process so we shall not have license header issues in the repo.

I see, we can probably remove explicitly checking license headers. This is something that in general I've seen all projects do as part of their release verification process and something that I would say falls under the "verify that they meet all requirements of ASF policy on releases as described below" point on the [ASF release guide](https://www.apache.org/legal/release-policy.html#release-approval). But it is true that as soon as there hasn't been any change as those are already done feels unnecessary.

How do you perform the comparison between the content of the tarball with a freshly cloned repo set to the release RC tag? Do we want to add that as a step?

I will move the PR to the parquet-site one, I might take a couple of days as I am slightly busy at the moment.

gszadovszky: > How do you perform the comparison between the content of the tarball with a freshly cloned repo set to the release RC tag? Do we want to add that as a step?

I use meld as diff tool but I don't think it should be added. Probably GNU `diff` can be configured to work on directory trees.

> I will move the PR to the parquet-site one, I might take a couple of days as I am slightly busy at the moment.

I don't think we need to hurry. Please refer the `parquet-site` PR here so anyone call follow up.

Thanks a lot for working on this!

Fokko: I agree that the `parquet-site` is more suitable for these steps.

Regarding the license headers. It is part of the verification; having the `rat` check is just one way. All code must have an ASv2 license header. It would also be good to do manual checks when a new version is being released, as the RAT check might also miss something.

raulcd:
Finally moved to the arrow-site repo:
https://github.com/apache/parquet-site/pull/98

I am closing this.

PR #3105, opened 2024-12-19
parquet1.10.0 support zstd

singer-bin:


Fokko:
@singer-bin Parquet 1.10 is pretty old (Jan 2019), and no new releases will be done for this version. Thanks for raising this PR, but I'm going to close it. I would highly recommend upgrading to a later version of Parquet-Java.

PR #3111, opened 2025-01-02
GH-3114: Fix LogicalType conversions for nested records on Avro <= 1.8

### LINKED ISSUE #3114: Built-in conversions for nested logical types fail on Avro <= 1.8

When an Avro record contains nested logical types but no top-level logical types, i.e. 

```json
{
  "type": "record",
  "name": "NestedLogicalTypes",
  "fields": [
    {
      "name": "innerType",
      "type": {
        "type": "record",
        "name": "InnerType",
        "fields": [
          {
            "name": "ts",
            "type": {
              "type": "long",
              "logicalType": "timestamp-millis"
            }
          }
        ]
      }
    }
  ]
}
```

Writes fails to convert the logical types from in-memory representation to underlying primitive:

```
Cause: java.lang.ClassCastException: class org.joda.time.DateTime cannot be cast to class java.lang.Number (org.joda.time.DateTime is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
at org.apache.parquet.avro.AvroWriteSupport.writeValueWithoutConversion(AvroWriteSupport.java:345)
at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:282)
at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:202)
at org.apache.parquet.avro.AvroWriteSupport.writeRecord(AvroWriteSupport.java:185)
at org.apache.parquet.avro.AvroWriteSupport.writeValueWithoutConversion(AvroWriteSupport.java:371)
at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:285)
at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:202)
at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:178)
at org.apache.parquet.hadoop.InternalParquetRecordWriter.write(InternalParquetRecordWriter.java:152)
at org.apache.parquet.hadoop.ParquetWriter.write(ParquetWriter.java:428)
```

### Component(s)

Avro

clairemcginty:
### Rationale for this change

Fixes logical type conversions for nested records on Avro <= 1.8. This was also addressed in https://github.com/apache/parquet-java/pull/1296/files but was missing the test case where _all_ logical types are in nested classes and there are zero top-level logical types. In this case, invoking `clazz.getDeclaredField("conversions")` in AvroRecordConverter throws a `NoSuchFieldException` and terminates the loop before attempting to traverse the record's inner fields.

Fixes #3114 

Fixes built-in logical type conversions for Avro records that contain only nested logical type fields on <= 1.8

### Are these changes tested?

Included a test for the case described above

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -196,10 +196,10 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
             }
           } catch (NoSuchFieldException e) {
             // Avro classes without logical types (denoted by the "conversions" field)
-          } finally {
-            for (Schema.Field field : schema.getFields()) {
-              addLogicalTypeConversion(model, field.schema(), seenSchemas);
-            }
+          }
+
+          for (Schema.Field field : schema.getFields()) {
+            addLogicalTypeConversion(model, field.schema(), seenSchemas);
           }
         }
         break;
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -194,12 +194,12 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
                 model.addLogicalTypeConversion(conversion);
               }
             }
-
+          } catch (NoSuchFieldException e) {
+            // Avro classes without logical types (denoted by the "conversions" field)
+          } finally {
             for (Schema.Field field : schema.getFields()) {
               addLogicalTypeConversion(model, field.schema(), seenSchemas);
             }
-          } catch (NoSuchFieldException e) {
-            // Avro classes without logical types (denoted by the "conversions" field)
           }
         }
         break;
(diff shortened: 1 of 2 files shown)


clairemcginty:
cc @RustedBones can you take a look when you have a chance?

wgtmac:
Thanks for the fix! Could you please create a Github issue for this and associate with it in the title?

cc @Fokko @gszadovszky as Avro experts.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -194,12 +194,12 @@ private static void addLogicalTypeConversion(SpecificData model, Schema schema,
                 model.addLogicalTypeConversion(conversion);
               }
             }
-
+          } catch (NoSuchFieldException e) {
+            // Avro classes without logical types (denoted by the "conversions" field)
+          } finally {
```
RustedBones, line 199: I don't think this should be in the finally block. We should not execute that in case of other exception

clairemcginty: hm what should we do instead? Just because the `conversions` class is null (meaning no top-level logical types) doesn't mean we shouldn't traverse the inner fields for any record types with conversions... at least in the generated classes I looked at

RustedBones: just have it after the try/catch without enclosing finally

clairemcginty: 👍 done

RustedBones (approved):

Fokko (approved):
This makes sense to me @clairemcginty, thanks for fixing this!

PR #3034, opened 2024-10-13
MINOR: Support int value for timeOfDay in ProtoTimeConverter

mwong38:
### Rationale for this change

TimeOfDay can sometimes be encoded as INT32 rather than INT64, especially if saved from Kafka Connect. 

### What changes are included in this PR?

Add support for Int as well as Long.

### Are these changes tested?

Yes

### Are there any user-facing changes?

No

PR #3113, opened 2025-01-06
GH-3112: Fix reading of proto Uint32Value

### LINKED ISSUE #3112: Fail to read google.protobuf.UInt32Value from parquet

TLDR: the parquet protobuf reader doesn't work for UInt32Value

I have protobuf using wrapped unsigned and signed integer:

```
syntax = "proto3";

package org.apache.parquet.test;

import "google/protobuf/wrappers.proto";

message MyTestMessage {
    google.protobuf.UInt32Value uint32_field = 11;
    google.protobuf.Int32Value int32_field = 12;
}
```

I then generate a parquet file for that data:

```
import pyarrow as pa
import pyarrow.parquet as pq

pq.write_table(
    pa.table(
        {
            "uint32_field": pa.array([None, None, 28], pa.uint32()),
            "int32_field": pa.array([None, 28, 28], pa.int32()),
        }
    ),
    "/tmp/my_test_messages.parquet",
)
```

And I try to read it using parquet-java (in kotlin, but it doesn't matter):

```kotlin
package org.apache.parquet.test

import org.apache.parquet.test.MyTestMessage
import com.google.protobuf.Int32Value
import io.kotest.matchers.shouldBe
import org.apache.hadoop.fs.Path
import org.apache.parquet.proto.ProtoConstants
import org.apache.parquet.proto.ProtoParquetReader
import org.apache.parquet.proto.ProtoReadSupport
import org.junit.jupiter.api.Test

class TestUInt32Value {
  @Test
  fun `test can not load UInt32Value`() {
    val reader =
      ProtoParquetReader.builder<MyTestMessage.Builder>(
          Path("file:///tmp/my_test_messages.parquet")
        )
        .set(ProtoReadSupport.PB_CLASS, MyTestMessage::class.java.canonicalName)
        .set(ProtoConstants.CONFIG_IGNORE_UNKNOWN_FIELDS, "true")
        .build()
    val firstMessage = reader.read().build()
    firstMessage shouldBe MyTestMessage.getDefaultInstance()

    val secondMessage = reader.read().build()
    secondMessage shouldBe
      MyTestMessage.newBuilder().setInt32Field(Int32Value.of(28)).build()


    val thirdMessage = reader.read()
  }
}
```

I get this error when reading the third message:

```
    org.apache.parquet.io.ParquetDecodingException: Can not read value at 3 in block 0 in file file:/tmp/my_test_messages.parquet
        at app//org.apache.parquet.hadoop.InternalParquetRecordReader.nextKeyValue(InternalParquetRecordReader.java:280)
        at app//org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:136)
        at app//org.apache.parquet.test.TestUInt32Value.test can load bad not nested plain(TestUInt32Value.kt:29)
        Caused by:
        java.lang.UnsupportedOperationException: org.apache.parquet.proto.ProtoMessageConverter$ProtoUInt32ValueConverter
            at org.apache.parquet.io.api.PrimitiveConverter.addInt(PrimitiveConverter.java:101)
            at org.apache.parquet.column.impl.ColumnReaderBase$2$3.writeValue(ColumnReaderBase.java:321)
            at org.apache.parquet.column.impl.ColumnReaderBase.writeCurrentValueToConverter(ColumnReaderBase.java:486)
            at org.apache.parquet.column.impl.ColumnReaderImpl.writeCurrentValueToConverter(ColumnReaderImpl.java:30)
            at org.apache.parquet.io.RecordReaderImplementation.read(RecordReaderImplementation.java:425)
            at org.apache.parquet.hadoop.InternalParquetRecordReader.nextKeyValue(InternalParquetRecordReader.java:249)
            ... 2 more
```

A few thing to note:
- this works for the second message, which means it is implemented correctly for (signed) Int32Value
- It works if you generate the data using the JVM.
   But this is because when you do so the parquet table has got a different structure (each message is a nested struct `{"value": 28}` 
```kotlin
  @Test
  fun `test jvm round trip`() {

    val path = Path("file:///tmp/my_test_messages_jvm.parquet")

    ProtoParquetWriter.builder<MyTestMessage>(path)
      .withMessage(MyTestMessage::class.java)
      .build()
      .use {
        it.write(MyTestMessage.getDefaultInstance())
        it.write(MyTestMessage.newBuilder().setInt32Field(Int32Value.of(28)).build())
        it.write(
          MyTestMessage.newBuilder()
            .setInt32Field(Int32Value.of(1))
            .setUint32Field(UInt32Value.of(32))
            .build()
        )
        it.close()
      }

    val reader =
      ProtoParquetReader.builder<MyTestMessage.Builder>(path)
        .set(ProtoReadSupport.PB_CLASS, MyTestMessage::class.java.canonicalName)
        .set(ProtoConstants.CONFIG_IGNORE_UNKNOWN_FIELDS, "true")
        .build()
    val firstMessage = reader.read().build()
    firstMessage shouldBe MyTestMessage.getDefaultInstance()

    val secondMessage = reader.read().build()
    secondMessage shouldBe MyTestMessage.newBuilder().setInt32Field(Int32Value.of(28)).build()

    val thirdMessage = reader.read().build()
    thirdMessage shouldBe MyTestMessage.newBuilder()
      .setInt32Field(Int32Value.of(1))
      .setUint32Field(UInt32Value.of(32))
      .build()
  }
```

This is basically generating a table that looks like this:

```python
import pyarrow as pa

pa.table(
    {
        "uint32_field": pa.array(
            [None, None, {"value": 28}], pa.struct([("value", pa.uint32())])
        ),
        "int32_field": pa.array(
            [None, {"value": 28}, {"value": 28}], pa.struct([("value", pa.int32())])
        ),
    }
)

```

### Component(s)

Protobuf

0x26res:
### Rationale for this change

Fix for https://github.com/apache/parquet-java/issues/3112

- Uint32Value primitive / physical type switched to INT32 (from INT64)
- Reading value for UInt32Value now uses int instead of long. In order to be backward compatible with previous parquet files (that were using long), I've left the `addLong` implementation of the  `ProtoUInt32ValueConverter`

### Are these changes tested?

There was extensive test coverage of this feature for round trip reading. The main change is the physical / underlying type change from int32 to int64 which is tested.

There are however no test for backward compatibility.

The parquet file generated from protobuf using Uint32Value will now use int32 instead of int64.

Related MR:
- https://github.com/apache/parquet-java/pull/900
- https://github.com/apache/parquet-java/pull/3034

Closes #3112


parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
```diff
@@ -260,7 +260,7 @@ private <T> Builder<? extends Builder<?, GroupBuilder<T>>, GroupBuilder<T>> addF
           return builder.primitive(INT32, getRepetition(descriptor));
         }
         if (messageType.equals(UInt32Value.getDescriptor())) {
-          return builder.primitive(INT64, getRepetition(descriptor));
+          return builder.primitive(INT32, getRepetition(descriptor));
```
Fokko, line 263: An unsigned 32-bit integer might overflow a signed 32-bit integer, right? It would be good to test this.

0x26res: I've just added a test using the min and max value of Integer and doing a round trip to make sure we get it back. It looks a bit weird because uint32 are represented as signed 32 bits integer in java, but it works.

Fokko: I don't think the test really test the out of bounds case. The `Integer.MAX_VALUE` is `2^31-1 = 2147483647`, but the unsigned integer could go up to `2^32-1 = 4294967295`. Maybe the original author @mwong38 has an opinion on this (I'm not a protobuf expert myself).

0x26res: Here's my understanding.

`uint32` is not a java concept. Therefore a protobuf field of type `uint32`, in the java generated code, will be treated as signed 32 bit integer. What java calls an `int`. 

Something like:

```proto
message TestMessage {
  uint32 value = 1;
}
```

Will generate setters and getters that use an `int`:
```java
 public Builder setValue(int value) // ....
```

However if I set the value to a negative value, for example `builder.setValue(-1)` and call `toString`, it will correctly interpret it as 4294967295.

Again what's important to note is that the underlying representation of a uint32 protobuf type in java is (32bit) int. Say I am to store, from a language that support unsigned integers (eg python), the value of a `uint32` to `4294967295`. Then if I read it in java I would get `-1`. But if I serialize the message again and pass it back to C or python, I would get back `4294967295` which is what I want.

I've extended the test to illustrate the behaviour better (adding calls to `toString`). Also I've added `-1` on top of `Integer.MIN_VALUE` and `Integer.MAX_VALUE` to correctly test the full spectrum of unsigned integer.

Fokko: I see, thanks for the excellent example. But why wouldn't we just handle this as an INT64 in Java, sacrificing some more memory, but see `4294967295` instead of `-1`?

0x26res: @Fokko I'd love to get this merged, could you have another look please.

mwong38 (approved):

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
0x26res: @Fokko again, could you please have another look at the MR, (or if busy suggest another reviewer). Thanks.

mwong38: @Fokko any luck on this? Thanks.

Fokko (approved):
Sorry for the late reply here, looking a bit deeper, I think this is correct since we handle int64 in the same way 👍

parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
Fokko: Certainly, sorry for letting this linger

Fokko:
Thanks @0x26res for the fix, and thanks @mwong38 for the review. This is scheduled to be included in 1.17.1 which should go out any day

Issue #3115, opened 2025-01-07
Fail to read INT96Timestamp from a list


pratyush84:
### Describe the bug, including details regarding any error messages, version, and platform.

Can be reproduced through the following code - 
```
  public void testIsElementTypeInt96Element(){
    Configuration configuration = new Configuration();
    configuration.setBoolean(READ_INT96_AS_FIXED, true);
    MessageType parquetSchema = MessageTypeParser.parseMessageType( "message SchemaWithInt96 {\n"
        + "  optional group list (LIST) {\n"
        + "    repeated group list {\n"
        + "      optional int96 a_timestamp;\n"
        + "    }\n"
        + "  }\n"
        + "}");
    Schema avroSchema = new AvroSchemaConverter(configuration).convert(parquetSchema);
    Assert.assertFalse(AvroRecordConverter.isElementType(
        parquetSchema.getType("list").asGroupType().getType("list"),
        AvroSchemaConverter.getNonNull(avroSchema.getFields().get(0).schema()).getElementType()
    ));
  }
```
When trying to read a List of INT96 timestamps, it gives the following error-  
`Exception in thread "main" java.lang.IllegalArgumentException: INT96 is deprecated. As interim enable READ_INT96_AS_FIXED flag to read as byte array.`

The issue is because in the [AvroRecordConverter.java](https://github.com/apache/parquet-java/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java#L942) class, AvroSchemaConverter is instantiated without the `ParquetConfiguration`.
`private static final AvroSchemaConverter CONVERTER = new AvroSchemaConverter(true); `

### Component(s)

Avro

PR #3118, opened 2025-01-10
GH-3115-Fix int96 read issue in complex type

pratyush-sharma-2025:
### Rationale for this change
There is an issue with reading of INT96 timestamp when they are present in a complex data type such as List<INT96>.
It gives the following error -
`Exception in thread "main" java.lang.IllegalArgumentException: INT96 is deprecated. As interim enable READ_INT96_AS_FIXED flag to read as byte array.`

More details are present [here](https://github.com/apache/parquet-java/issues/3115).

* Use `AvroSchemaConverter(boolean assumeRepeatedIsListElement, ParquetConfiguration conf)` while checking if the given type is element type of a list in `AvroRecordConverter#isElementType()`

### Are these changes tested?
Added a new test case `TestArrayCompatibility#testIsElementTypeInt96Element()`

No.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -58,6 +58,7 @@ import org.apache.avro.util.ClassUtils;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.avro.AvroConverters.FieldStringConverter;
 import org.apache.parquet.avro.AvroConverters.FieldStringableConverter;
+import org.apache.parquet.conf.ParquetConfiguration;
 import org.apache.parquet.io.InvalidRecordException;
 import org.apache.parquet.io.api.Converter;
 import org.apache.parquet.io.api.GroupConverter;
@@ -91,9 +92,12 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
 
   private final GenericData model;
   private final Map<Schema.Field, Object> recordDefaults = new HashMap<Schema.Field, Object>();
+  private final AvroSchemaConverter avroSchemaConverter;
+  private final ParquetConfiguration conf;
 
-  public AvroRecordConverter(MessageType parquetSchema, Schema avroSchema, GenericData baseModel) {
-    this(null, parquetSchema, avroSchema, baseModel);
+  public AvroRecordConverter(
+      MessageType parquetSchema, Schema avroSchema, GenericData baseModel, ParquetConfiguration conf) {
+    this(null, parquetSchema, avroSchema, baseModel, conf);
     LogicalType logicalType = avroSchema.getLogicalType();
     Conversion<?> conversion = baseModel.getConversionFor(logicalType);
     this.rootContainer = ParentValueContainer.getConversionContainer(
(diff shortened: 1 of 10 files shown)


pratyush-sharma-2025:
@rdblue @tomwhite Can you please help review here? Thanks!

pratyush-sharma-2025:
@Fokko @gszadovszky @wgtmac Can you please take a look when you have a chance?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -963,7 +1003,8 @@ static boolean isElementType(Type repeatedType, Schema elementSchema) {
       // synthetic wrapper. Must be a group with one optional or required field
       return true;
     } else if (elementSchema != null && elementSchema.getType() == Schema.Type.RECORD) {
-      Schema schemaFromRepeated = CONVERTER.convert(repeatedType.asGroupType());
+      AvroSchemaConverter converter = new AvroSchemaConverter(true, conf);
```
gszadovszky, line 1006: As the description of `CONVERTER` states, it is only for testing whether the generated schema contains 2-level lists. I think, it would be safe to simply hard-code for that `AvroSchemaConverter` constructor to enable reading `INT96` values to not to throw an exception. As a result, you don't have to push down the conf all the way down.

gszadovszky (changes requested):
Thanks for working on this, @pratyush-sharma-2025. I think you can simplify your code greatly. See my comment in the code.

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
```diff
@@ -587,16 +587,12 @@ public void testParquetInt96AsFixed12AvroType() throws Exception {
   }
 
   @Test
-  public void testParquetInt96DefaultFail() throws Exception {
+  public void testParquetInt96DefaultPass() throws Exception {
```
gszadovszky, line 590: The expected behavior is to fail INT96 reading by default. We should not change this test.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -103,7 +103,7 @@ public AvroSchemaConverter() {
     this.assumeRepeatedIsListElement = assumeRepeatedIsListElement;
     this.writeOldListStructure = WRITE_OLD_LIST_STRUCTURE_DEFAULT;
     this.writeParquetUUID = WRITE_PARQUET_UUID_DEFAULT;
-    this.readInt96AsFixed = READ_INT96_AS_FIXED_DEFAULT;
+    this.readInt96AsFixed = true;
```
gszadovszky, line 106: Oh, I've missed the fact that this constructor is invoked from the parameterless one, which is public. We shall keep the original default behavior of the int96 reading for all cases but the one where we initialize `AvroRecordConverter.CONVERTER`.

gszadovszky (changes requested):

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
pratyush-sharma-2025: @gszadovszky Added a new param to the AvroRecordConverter constructor for reading int96 as fixed. Please see if all ok now.

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroSchemaConverter.java
pratyush-sharma-2025: Reverted back this change.

gszadovszky (approved):
Thanks, @pratyush-sharma-2025. Everything looks good.
Let me wait for an additional day before committing so others have the chance to object.

Issue #3123, opened 2025-01-17
Omit level histogram for some max levels without loss of precision


wgtmac:
### Describe the enhancement requested

According the spec of SizeStatistics, we can omit level histogram without loss of precision when max_repetition_level is 0 or max_definition_level is 0 or 1. See https://github.com/apache/arrow/pull/45285

### Component(s)

Core

PR #3124, opened 2025-01-17
GH-3123: Omit level histogram for some max levels

wgtmac:
### Rationale for this change

The level histogram of size statistics can be omitted without loss of precision if its max_definition_level is 1 or 0, or max_repetition_level is 0.

Omit level histogram without loss of precision if possible.

### Are these changes tested?

Yes, add test case for the omitted case.

No.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
@@ -67,8 +67,16 @@ public class SizeStatistics {
     private Builder(PrimitiveType type, int maxRepetitionLevel, int maxDefinitionLevel) {
       this.type = type;
       this.unencodedByteArrayDataBytes = 0L;
-      repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
-      definitionLevelHistogram = new long[maxDefinitionLevel + 1];
+      if (maxRepetitionLevel > 0) {
+        repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
+      } else {
+        repetitionLevelHistogram = new long[0]; // omitted
+      }
+      if (maxDefinitionLevel > 1) {
+        definitionLevelHistogram = new long[maxDefinitionLevel + 1];
+      } else {
+        definitionLevelHistogram = new long[0]; // omitted
+      }
     }
 
     /**
@@ -79,8 +87,12 @@ public class SizeStatistics {
      * @param definitionLevel definition level of the value
      */
     public void add(int repetitionLevel, int definitionLevel) {
-      repetitionLevelHistogram[repetitionLevel]++;
-      definitionLevelHistogram[definitionLevel]++;
+      if (repetitionLevelHistogram.length > 0) {
+        repetitionLevelHistogram[repetitionLevel]++;
+      }
+      if (definitionLevelHistogram.length > 0) {
+        definitionLevelHistogram[definitionLevel]++;
+      }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wgtmac:
@emkornfield Could you please take a look?

cc @etseidl

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -67,8 +67,16 @@ public static class Builder {
...
       this.type = type;
       this.unencodedByteArrayDataBytes = 0L;
-      repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
-      definitionLevelHistogram = new long[maxDefinitionLevel + 1];
+      if (maxRepetitionLevel > 0) {
+        repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
+      } else {
+        repetitionLevelHistogram = new long[0]; // omitted
+      }
+      if (maxDefinitionLevel > 1) {
```
etseidl, line 75: I'm curious why you've chosen 1 here, when other implementations use 0. Is it so parquet-java will implement the spec exactly or are you thinking to minimize metadata size? Not a criticism, just genuinely curious.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -67,8 +67,16 @@ public static class Builder {
     private Builder(PrimitiveType type, int maxRepetitionLevel, int maxDefinitionLevel) {
       this.type = type;
       this.unencodedByteArrayDataBytes = 0L;
-      repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
-      definitionLevelHistogram = new long[maxDefinitionLevel + 1];
+      if (maxRepetitionLevel > 0) {
+        repetitionLevelHistogram = new long[maxRepetitionLevel + 1];
+      } else {
+        repetitionLevelHistogram = new long[0]; // omitted
```
emkornfield, line 73: not sure if there is a style for preferring null vs 0 length array (or does this flow through more easily for consuption)?

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
```diff
@@ -79,8 +87,12 @@ private Builder(PrimitiveType type, int maxRepetitionLevel, int maxDefinitionLev
      * @param definitionLevel definition level of the value
      */
     public void add(int repetitionLevel, int definitionLevel) {
-      repetitionLevelHistogram[repetitionLevel]++;
-      definitionLevelHistogram[definitionLevel]++;
+      if (repetitionLevelHistogram.length > 0) {
```
emkornfield, line 90: do we have benchmarks for this class?  I'm curious if there is a trade-off of always incrementing here and avoiding writing at the end (I suppose JVM might be smart enough to remove the conditionals so I'm fine keeping it as is).

emkornfield (approved):
Not an expert in java but the changes seem reasonable to me, asked a few questions mostly for curiousity.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
wgtmac: 0 length array makes it easier for other methods (e.g. `copy()`) to avoid `NullPointerException`.

parquet-column/src/main/java/org/apache/parquet/column/statistics/SizeStatistics.java
wgtmac: I'm not a Java expert so I prefer to keep the current impl just in case.

gszadovszky (approved):
LGTM

wgtmac:
I just merged it. Thanks @etseidl @emkornfield @gszadovszky!

PR #3132, opened 2025-01-19
MINOR: Remove Joda as a direct dependency

Fokko:
I saw https://github.com/apache/parquet-java/pull/3130 and I couldn't find any direct references to the Joda library in the Parquet code.

Instead of bumping it, it would be better to just remove it.


wgtmac (approved):

wgtmac:
Good catch!

PR #3134, opened 2025-01-21
GH-3133: Fix SizeStatistics to handle omitted histogram

wgtmac:
### Rationale for this change

If SizeStatistics has omitted level histogram, creating an `SizeStatistics` will result in a null object which leads to NullPointerException when calling getter or copy methods.

### What changes are included in this PR?

Use a zero-length list for omitted histogram.

### Are these changes tested?

Yes.

### Are there any user-facing changes?

No.

Closes #3133

Issue #3133, opened 2025-01-21
Fix SizeStatistics to handle omitted histogram


wgtmac:
### Describe the bug, including details regarding any error messages, version, and platform.

See https://github.com/apache/parquet-java/pull/3126/files#r1921884951

### Component(s)

Core

PR #3135, opened 2025-01-21
GH-3133: Fix SizeStatistics to handle omitted histogram

wgtmac:
This is the backport of #3134


parquet-plugins/parquet-encoding-vector/pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <groupId>org.apache.parquet</groupId>
     <artifactId>parquet</artifactId>
-    <version>1.15.0-SNAPSHOT</version>
+    <version>1.15.1-SNAPSHOT</version>
```
ConeyLiu, line 25: Why is this changed? Is it because this was overlooked before?

wgtmac: The release process does not automatically update this...

ConeyLiu: OK, maybe we should fix it later.

ConeyLiu (approved):
+1

PR #3143, opened 2025-02-05
MINOR: Improve exception message in InternalFileDecryptor

zhongyujiang:
### Rationale for this change
When reading encrypted columns, if the provided key is NULL, the thrown exception message might be confusing: "Column [foo] is encrypted with NULL column key".  However NULL key is certainly not permitted for encryption, the occurrence of such an exception invariably indicates an issue with the key provisioning process. So this PR explicitly states this in the exception message.

Exception msg change.

### Are these changes tested?
No need.

No.

@wgtmac @ggershinsky can you please take a look? Thanks!


wgtmac (approved):
+1 Thanks @zhongyujiang!

cc @ggershinsky

ggershinsky (approved):
Thanks @zhongyujiang

ggershinsky:
good catch, these messages sounded wrong in a decryptor.

PR #3147, opened 2025-02-06
[GH-3146] Optimize the binaryToDecimal function

### LINKED ISSUE #3146: Optimize the binaryToDecimal function in the DecimalUtils class

https://github.com/apache/parquet-java/blob/master/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
```java
  public static BigDecimal binaryToDecimal(Binary value, int precision, int scale) {
    /*
     * Precision <= 18 checks for the max number of digits for an unscaled long,
     * else treat with big integer conversion
     */
    if (precision <= 18) {
      ByteBuffer buffer = value.toByteBuffer();
      byte[] bytes = buffer.array();
      int start = buffer.arrayOffset() + buffer.position();
      int end = buffer.arrayOffset() + buffer.limit();
      long unscaled = 0L;
      int i = start;
      while (i < end) {
        unscaled = (unscaled << 8 | bytes[i] & 0xff);
        i++;
      }
      int bits = 8 * (end - start);
      long unscaledNew = (unscaled << (64 - bits)) >> (64 - bits);
      if (unscaledNew <= -pow(10, 18) || unscaledNew >= pow(10, 18)) {
        return new BigDecimal(unscaledNew);
      } else {
        return BigDecimal.valueOf(unscaledNew / pow(10, scale));
      }
    } else {
      return new BigDecimal(new BigInteger(value.getBytes()), scale);
    }
  }
```

If precision is less than 18, the condition `unscaledNew <= -pow(10, 18) || unscaledNew >= pow(10, 18)` can not be true, so we can remove the judgment logic here. Additionally, using `BigDecimal.valueOf(unscaledNew, scale)` is preferable over using `BigDecimal.valueOf(unscaledNew / pow(10, scale))`, as it does not convert the unscaled value to double.

### Component(s)

_No response_

qian0817:
### Rationale for this change
#3146 
If precision is less than 18, the condition unscaledNew <= -pow(10, 18) || unscaledNew >= pow(10, 18) can not be true, so we can remove the judgment logic here. Additionally, using BigDecimal.valueOf(unscaledNew, scale) is preferable over using BigDecimal.valueOf(unscaledNew / pow(10, scale)), as it does not convert the unscaled value to double.

Optimize the binaryToDecimal function

### Are these changes tested?
pass unit test.

No

Closes #3146

### DIFF
--- a/parquet-pig/src/test/java/org/apache/parquet/pig/TestDecimalUtils.java
+++ b/parquet-pig/src/test/java/org/apache/parquet/pig/TestDecimalUtils.java
@@ -60,12 +60,12 @@ public class TestDecimalUtils {
     // Test LONG
     testDecimalConversion(Long.MAX_VALUE, 19, 0, "9223372036854775807");
     testDecimalConversion(Long.MIN_VALUE, 19, 0, "-9223372036854775808");
-    testDecimalConversion(0L, 0, 0, "0.0");
+    testDecimalConversion(0L, 0, 0, "0");
 
     // Test INTEGER
     testDecimalConversion(Integer.MAX_VALUE, 10, 0, "2147483647");
     testDecimalConversion(Integer.MIN_VALUE, 10, 0, "-2147483648");
-    testDecimalConversion(0, 0, 0, "0.0");
+    testDecimalConversion(0, 0, 0, "0");
 
     // Test DOUBLE
     testDecimalConversion(12345678912345678d, 17, 0, "12345678912345678");
--- a/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
+++ b/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
@@ -19,8 +19,6 @@
 
 package org.apache.parquet.pig.convert;
 
-import static java.lang.Math.pow;
-
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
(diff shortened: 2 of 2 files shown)


parquet-pig/src/test/java/org/apache/parquet/pig/TestDecimalUtils.java
```diff
@@ -60,12 +60,12 @@ public void testBinaryToDecimal() throws Exception {
     // Test LONG
     testDecimalConversion(Long.MAX_VALUE, 19, 0, "9223372036854775807");
     testDecimalConversion(Long.MIN_VALUE, 19, 0, "-9223372036854775808");
-    testDecimalConversion(0L, 0, 0, "0.0");
+    testDecimalConversion(0L, 0, 0, "0");
```
wgtmac, line 63: Why do these two lines need change?

wgtmac (commented):
parquet-pig has been discussed to be removed: https://lists.apache.org/thread/vh1twzdbvm4fr4sl2wt8swqgq92k8369

Is it actually used in your case @qian0817?

cc @Fokko

qian0817:
> parquet-pig has been discussed to be removed: https://lists.apache.org/thread/vh1twzdbvm4fr4sl2wt8swqgq92k8369
> 
> Is it actually used in your case @qian0817?
> 
> cc @Fokko

I did not directly use the parquet-pig module; while writing my own parquet converter, I referenced some code from parquet-pig and then discovered the optimization points here.

parquet-pig/src/test/java/org/apache/parquet/pig/TestDecimalUtils.java
qian0817: For this use case, 0 is the correct value. Using double to construct BigDecimal previously may lead to potential errors.

Fokko:
Thanks for pinging me here @wgtmac. I've just raised a PR to remove Pig: https://github.com/apache/parquet-java/pull/3153 

@qian0817 Can I ask why you went through the trouble of writing your own converter?

qian0817:
> Thanks for pinging me here @wgtmac. I've just raised a PR to remove Pig: #3153
> 
> @qian0817 Can I ask why you went through the trouble of writing your own converter?

@Fokko I need to read the Parquet file and convert it to our internal system's special format.

PR #3148, opened 2025-02-06
GH-3122: Correct V2 page header compression fields for zero-size data pages

### LINKED ISSUE #3122: Parquet-java sometimes produces 0-size compressed data in data page v2
### Describe the bug

See: 
1. https://github.com/apache/arrow/pull/45252#discussion_r1916253275
2. File: https://github.com/user-attachments/files/18292070/snappy_bug.parquet.gz

V2 data pages do not compress def/rep levels, only the values. In the (not uncommon) case where all data is null, Parquet-java may write 0 compressed bytes for a 0-size decompressed data. However, a 0-size compressed buffer is not a valid input for compressors. The data page then fails decompressing in the C++ and Rust implementations.

### Component(s)

Core

ConeyLiu:
### Rationale for this change

Fixes #3122 

Set `is_compressed` to false and `compressed_page_size` to 0 for the V2 page with no data size.

### Are these changes tested?

Yes, new UTs.

No.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
@@ -55,7 +55,11 @@ public class AesCtrDecryptor extends AesCipher implements BlockCipher.Decryptor 
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength < 1) {
+    if (plainTextLength == 0) {
+      return new byte[0];
+    }
+
+    if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
 
@@ -91,7 +95,11 @@ public class AesCtrDecryptor extends AesCipher implements BlockCipher.Decryptor 
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength < 1) {
+    if (plainTextLength == 0) {
+      return ByteBuffer.allocate(0);
+    }
+
+    if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
@@ -55,10 +55,6 @@ public class AesCtrDecryptor extends AesCipher implements BlockCipher.Decryptor 
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength == 0) {
-      return new byte[0];
-    }
-
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
@@ -95,10 +91,6 @@ public class AesCtrDecryptor extends AesCipher implements BlockCipher.Decryptor 
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength == 0) {
-      return ByteBuffer.allocate(0);
-    }
-
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
@@ -51,9 +51,6 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-    if (plainTextLength == 0) {
-      return new byte[0];
-    }
 
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
@@ -85,10 +82,6 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
     int cipherTextOffset = SIZE_LENGTH;
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-
-    if (plainTextLength == 0) {
-      return ByteBuffer.allocate(0);
-    }
 
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
@@ -51,7 +51,6 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
@@ -82,7 +81,6 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
     int cipherTextOffset = SIZE_LENGTH;
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-
     if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
@@ -51,7 +51,11 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-    if (plainTextLength < 1) {
+    if (plainTextLength == 0) {
+      return new byte[0];
+    }
+
+    if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
 
@@ -81,7 +85,12 @@ public class AesGcmDecryptor extends AesCipher implements BlockCipher.Decryptor 
     int cipherTextOffset = SIZE_LENGTH;
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-    if (plainTextLength < 1) {
+
+    if (plainTextLength == 0) {
+      return ByteBuffer.allocate(0);
+    }
+
+    if (plainTextLength < 0) {
       throw new ParquetCryptoRuntimeException("Wrong input length " + plainTextLength);
     }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -1981,7 +1981,8 @@ public class ParquetMetadataConverter {
             rowCount,
             dataEncoding,
             rlByteLength,
-            dlByteLength),
+            dlByteLength,
+            true /* compressed by default */),
         to);
   }
 
@@ -2059,6 +2060,10 @@ public class ParquetMetadataConverter {
         pageHeaderAAD);
   }
 
+  /**
+   * @deprecated will be removed in 2.0.0. Use {@link ParquetMetadataConverter#writeDataPageV2Header(int, int, int, int, int, org.apache.parquet.column.Encoding, int, int, boolean, OutputStream)} instead
+   */
+  @Deprecated
   public void writeDataPageV2Header(
       int uncompressedSize,
       int compressedSize,
@@ -2079,11 +2084,16 @@ public class ParquetMetadataConverter {
         dataEncoding,
         rlByteLength,
         dlByteLength,
+        true, /* compressed by default */
         to,
         null,
         null);
   }
 
+  /**
+   * @deprecated will be removed in 2.0.0. Use {@link ParquetMetadataConverter#writeDataPageV2Header(int, int, int, int, int, org.apache.parquet.column.Encoding, int, int, boolean, OutputStream, BlockCipher.Encryptor, byte[])} instead
+   */
+  @Deprecated
   public void writeDataPageV2Header(
       int uncompressedSize,
       int compressedSize,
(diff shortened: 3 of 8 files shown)


ConeyLiu:
@mapleFU @pitrou @wgtmac pls take a look. Thanks a lot.

parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV2.java
```diff
@@ -163,6 +170,33 @@ public DataPageV2(
...
+      Encoding dataEncoding,
+      BytesInput data,
+      int compressedSize,
+      int uncompressedSize,
+      Statistics<?> statistics,
+      boolean isCompressed) {
+    super(compressedSize, uncompressedSize, valueCount);
+    if (!isCompressed && compressedSize != 0) {
+      throw new IllegalArgumentException("compressedSize must be 0 if page is not compressed");
+    }
```
pitrou, line 188: In Parquet C++, compressed_size is always written out.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2123,6 +2123,9 @@ private PageHeader newDataPageV2Header(
       int dlByteLength) {
     DataPageHeaderV2 dataPageHeaderV2 = new DataPageHeaderV2(
         valueCount, nullCount, rowCount, getEncoding(dataEncoding), dlByteLength, rlByteLength);
+    if (compressedSize == 0) {
+      dataPageHeaderV2.setIs_compressed(false);
+    }
```
pitrou, line 2128: I think you should take an explicit `bool isCompressed` parameter instead.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -295,14 +295,19 @@ public void writePageV2(
...
+        // TODO: decide if we compress
+        compressedData = compressor.compress(data);
+        if (null != pageBlockEncryptor) {
+          AesCipher.quickUpdatePageAAD(dataPageAAD, pageOrdinal);
+          compressedData =
+              BytesInput.from(pageBlockEncryptor.encrypt(compressedData.toByteArray(), dataPageAAD));
+        }
+        compressedSize =
+            toIntWithCheck(compressedData.size() + repetitionLevels.size() + definitionLevels.size());
       }
```
pitrou, line 310: So encryption wouldn't be handled properly if the uncompressed size is 0?

Also, `compressedSize` should always contain the total size of the data page (excluding header, but including encryption).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1969,6 +1969,7 @@ public ColumnChunkPageReader readAllPages(
                 definitionLevels,
                 converter.getEncoding(dataHeaderV2.getEncoding()),
                 values,
+                dataHeaderV2.isIs_compressed() ? compressedPageSize : 0,
```
pitrou, line 1972: Why this condition?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1150,8 +1150,11 @@ public void writeDataPageV2(
     int rlByteLength = toIntWithCheck(repetitionLevels.size(), "page repetition levels");
     int dlByteLength = toIntWithCheck(definitionLevels.size(), "page definition levels");
 
-    int compressedSize =
-        toIntWithCheck(compressedData.size() + repetitionLevels.size() + definitionLevels.size(), "page");
+    int compressedSize = 0;
+    if (compressedData.size() > 0) {
+      compressedSize =
+          toIntWithCheck(compressedData.size() + repetitionLevels.size() + definitionLevels.size(), "page");
+    }
```
pitrou, line 1157: So this class is duplicating code from `ColumnChunkPageWriteStore.java` above? Do you know why that is?

parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV2.java
ConeyLiu: So we only should set the `is_compressed` as false and don't care about the `compressed_size` for this case?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
ConeyLiu: OK

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: `ParquetFileWriter` supports writing out pages directly. So there are some duplicate codes. I plan to reduce the duplication for better maintenance.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2123,6 +2123,9 @@ private PageHeader newDataPageV2Header(
       int dlByteLength) {
     DataPageHeaderV2 dataPageHeaderV2 = new DataPageHeaderV2(
         valueCount, nullCount, rowCount, getEncoding(dataEncoding), dlByteLength, rlByteLength);
+    if (compressedSize == 0) {
+      dataPageHeaderV2.setIs_compressed(false);
```
wgtmac, line 2127: I'm surprised that `dataPageHeaderV2.setIs_compressed()` has never been called before.

wgtmac: BTW, I think the description in the spec needs to be improved since it does not consider the case when `comressed_page_size` is 0: https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L665-L668

>  Which means the section of the page between definition_levels_byte_length + repetition_levels_byte_length + 1 and compressed_page_size (included) is compressed with the compression_codec.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
wgtmac: Agreed. Data page v2 was designed to adaptively fall back to uncompressed data when compression is not promising (though we don't implement it yet). Using an explicit parameter makes sense.

parquet-column/src/main/java/org/apache/parquet/column/page/Page.java
```diff
@@ -34,6 +34,9 @@ public abstract class Page {
     this.uncompressedSize = uncompressedSize;
   }
 
+  /**
+   * @return the compressed size of the page when the bytes are compressed, otherwise return 0
```
wgtmac, line 38: I would rather return uncompressed size if it is not compressed.

parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV2.java
wgtmac: IMO, we should assume `compressedSize` == `uncompressedSize` when it is uncompressed. `UNCOMPRESSED` codec is still a valid compression codec type, otherwise, `RowGroup.total_compressed_size` may have problems.

wgtmac (changes requested):
Thanks for fixing this! I've left some comments.

parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV2.java
pitrou: > IMO, we should assume `compressedSize` == `uncompressedSize` when it is uncompressed. 

No, because `compressedSize` can also include the overhead of encryption.

`uncompressedSize` is sizeof(levels + uncompressed data).
`compressedSize` is sizeof(encrypted(levels + compressed data))

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
pitrou: Well, I don't think `compressed_page_size` can be 0, except if you have 0 levels and 0 data (is that possible?).

wgtmac: My bad. Fixed my comment.

parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV2.java
wgtmac: Ah yes, I forgot encryption.

ConeyLiu: I reverted here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
ConeyLiu: no needed, reverted it

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
```diff
@@ -55,7 +55,11 @@ public byte[] decrypt(byte[] lengthAndCiphertext, byte[] AAD) {
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength < 1) {
+    if (plainTextLength == 0) {
```
ConeyLiu, line 58: decryptor doesn't support decrypting zero-byte data, fixes here.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2152,12 +2207,43 @@ public void writeDataPageV2Header(
...
+  public void writeDataPageV2Header(
+      int uncompressedSize,
+      int compressedSize,
+      int valueCount,
+      int nullCount,
+      int rowCount,
+      org.apache.parquet.column.Encoding dataEncoding,
+      int rlByteLength,
+      int dlByteLength,
+      boolean compressed,
```
ConeyLiu, line 2225: @pitrou @wgtmac changed to an explicit parameter.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -737,6 +737,7 @@ private void processChunk(
               encryptColumn,
               dataEncryptor,
               dataPageAAD);
+          boolean compressed = compressor != null;
```
wgtmac, line 740: Shouldn't we get it from `headerV2` to keep it unchanged?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1131,7 +1227,9 @@ public void writeDataPageV2(
    * @param pageHeaderAAD pageHeader AAD
    * @param sizeStatistics size statistics for the page
    * @throws IOException if any I/O error occurs during writing the file
+   * @deprecated will be removed in 2.0.0. Use {@link ParquetFileWriter#writeDataPageV2(int, int, int, BytesInput, BytesInput, Encoding, BytesInput, boolean, int, Statistics, BlockCipher.Encryptor, byte[], SizeStatistics)} instead
    */
+  @Deprecated
```
wgtmac, line 1232: Not related to this PR: we have plenty of deprecated public methods with a bunch of parameters like this. It takes pain for the downstream to migrate after removing them in the 2.0.0 and still likely to be deprecated if we need a new parameter next time. Should we change these into a `Builder` or a class with many parameters so we won't break anything in the future? @gszadovszky @Fokko

wgtmac (commented):
Generally LGTM. I've left a few comments.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
```diff
@@ -51,7 +51,11 @@ public byte[] decrypt(byte[] lengthAndCiphertext, byte[] AAD) {
   public byte[] decrypt(byte[] ciphertext, int cipherTextOffset, int cipherTextLength, byte[] AAD) {
 
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-    if (plainTextLength < 1) {
```
ggershinsky, line 54: AES GCM decryption is able to handle an empty plaintext. It will also use the key to check the IV/tag integrity.
So the fix can be just replacement of  `if (plainTextLength < 1)` { with `if (plainTextLength < 0) {`

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
```diff
@@ -81,7 +85,12 @@ public ByteBuffer decrypt(ByteBuffer ciphertext, byte[] AAD) {
     int cipherTextOffset = SIZE_LENGTH;
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
     int plainTextLength = cipherTextLength - GCM_TAG_LENGTH - NONCE_LENGTH;
-    if (plainTextLength < 1) {
```
ggershinsky, line 84: same here

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesGcmDecryptor.java
ConeyLiu: Thanks for the suggestions. Updated.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ConeyLiu: Separating those methods from `ParquetFileWriter` class and reducing the duplication between `ColumnChunkPageWriteStore` could help the maintenance.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
ConeyLiu: Changed `headerV2.is_compressed` since `UNCOMPRESSED` is still a compression codec.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
pitrou: @ggershinsky Does this look ok to you?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1146,12 +1244,60 @@ public void writeDataPageV2(
...
+      byte[] pageHeaderAAD,
+      SizeStatistics sizeStatistics)
+      throws IOException {
     state = state.write();
     int rlByteLength = toIntWithCheck(repetitionLevels.size(), "page repetition levels");
     int dlByteLength = toIntWithCheck(definitionLevels.size(), "page definition levels");
 
-    int compressedSize =
-        toIntWithCheck(compressedData.size() + repetitionLevels.size() + definitionLevels.size(), "page");
+    int compressedSize = toIntWithCheck(bytes.size() + repetitionLevels.size() + definitionLevels.size(), "page");
```
pitrou, line 1300: Isn't `compressedSize` also supposed to include the size of `pageHeaderAAD`? @ggershinsky

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
ggershinsky: yep, technically this should work ok - unlike GCM, the CTR mode doesn't guarantee/check the integrity, so we don't need to protect against page replacement attacks (we can't anyway).

In terms of implementation, it might be a bit cleaner if we use cipher decryption of zero-sized arrays; fewer lines of code. But I'll verify first if this is supported in CTR, will get back on that.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
ggershinsky: The AAD suffix parameters are calculated in-memory, they are not stored in the file.

The `compressedSize` does include the encryption IV/tag (28 bytes for GCM), but this was already accounted for in the old `compressedData.size()`. I presume the new `bytes.size()` is the same.

ConeyLiu: Yes, here just renamed the `compressedData` to `bytes` since the data maybe not be compressed.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
```diff
@@ -91,7 +95,11 @@ public ByteBuffer decrypt(ByteBuffer ciphertext, byte[] AAD) {
     int cipherTextLength = ciphertext.limit() - ciphertext.position() - SIZE_LENGTH;
 
     int plainTextLength = cipherTextLength - NONCE_LENGTH;
-    if (plainTextLength < 1) {
```
ggershinsky, line 94: same here

parquet-hadoop/src/main/java/org/apache/parquet/crypto/AesCtrDecryptor.java
ConeyLiu: Updated

ConeyLiu:
hi, @pitrou @ggershinsky. Do you have more comments about this?

ggershinsky (approved):
Nope, LGTM

wgtmac:
I just merged it. Thanks all!

ConeyLiu:
Thanks @wgtmac @pitrou @ggershinsky

PR #3154, opened 2025-02-13
Add logical type annotation for `UnknownType`

Fokko:
### Rationale for this change

Implements https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#unknown-always-null

This was missing from the Java codebase, and I think it would be good to add this afterward since this is also implemented in other languages: https://github.com/apache/arrow-rs/blob/9c92a50b6d190ca9d0c74c3ccc69e348393d9246/parquet/src/format.rs#L1852


wgtmac (approved):
+1

I am surprised that we don't have this yet

BTW, should we create an issue for this? Or should we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) if it is boring to create an issue every time?

Fokko:
@wgtmac Thanks. I'm not a big fan of the issues (I keep forgetting to create one), but I do think it would be good to add more context to the PR itself.

Issue #3163, opened 2025-02-26
ParquetRewriterTests use up a lot of memory and take very long to run


rahulketch:
Currently the ParquetRewriterTests take >12GB of memory and >750s to run. This is because they create multiple files with 100000 records where each record id a few kb. I propose reducing the number of records to 10000.

PR #3164, opened 2025-02-26
GH-3163: Reduce memory and time overhead of ParquetRewriterTests

rahulketch:
### Rationale for this change
Reduce the memory overhead and the time taken to run ParquetRewriterTests

Reducing number of records from 100000 to 10000

No

Closes #GH-3163

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
@@ -122,7 +122,9 @@ public class ParquetRewriterTest {
   private final EncryptionTestFile gzipEncryptionTestFileWithoutBloomFilterColumn;
   private final EncryptionTestFile uncompressedEncryptionTestFileWithoutBloomFilterColumn;
 
-  @Parameterized.Parameters(name = "WriterVersion = {0}, IndexCacheStrategy = {1}, UsingHadoop = {2}, numRecord = {3}, rowsPerPage = {4}")
+  @Parameterized.Parameters(
+      name =
+          "WriterVersion = {0}, IndexCacheStrategy = {1}, UsingHadoop = {2}, numRecord = {3}, rowsPerPage = {4}")
   public static Object[][] parameters() {
     final int DefaultNumRecord = 10000;
     final int DefaultRowsPerPage = DefaultNumRecord / 5;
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -107,7 +107,7 @@
 @RunWith(Parameterized.class)
 public class ParquetRewriterTest {
 
-  private final int numRecord = 100000;
+  private final int numRecord = 10000;
```
wgtmac, line 110: This may result in a single page for each column chunk. Could you try following things:
- Make sure more than one page are produced by setting config like `parquet.page.value.count.threshold` and relevant page size check parameters at https://github.com/apache/parquet-java/blob/master/parquet-hadoop/README.md
- Parameterize `numRecord` (and page size check config) and by default use 10000?

rahulketch: @wgtmac : thanks for your comment. Do you happen to know how I can run a single test in the repository after making my changes? I wanted to run only the `ParquetRewriterTest`, but have not figured out a good way to achieve that via mvn.

wgtmac:
```
cd ~/Projects/parquet-java   # replace with your project root directory

cd parquet-hadoop

mvn test -Dtest=org.apache.parquet.hadoop.rewrite.ParquetRewriterTest
```

rahulketch: As far as I understand, the relevant config parameter is `parquet.page.row.count.limit` and I have changed that to be `num_records / 5`. Hope it looks good!

rahulketch:
@wgtmac : Will this line also require a change? 
https://github.com/apache/parquet-java/blob/d20ed14fd80973ca5e2383fef152205c17b5eba9/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java#L928

wgtmac:
> @wgtmac : Will this line also require a change?
> 
> https://github.com/apache/parquet-java/blob/d20ed14fd80973ca5e2383fef152205c17b5eba9/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java#L928

Yes, I think it would be good to test more than one row group.

rahulketch:
> > @wgtmac : Will this line also require a change?
> > https://github.com/apache/parquet-java/blob/d20ed14fd80973ca5e2383fef152205c17b5eba9/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java#L928
> 
> Yes, I think it would be good to test more than one row group.

I verified that currently the test creates more than 1 row group with the current values. Let me know if any more concerns before merging.

wgtmac (approved):
LGTM Thanks!

cc @ConeyLiu

ConeyLiu (approved):
+1, thanks for the contribution.

PR #3180, opened 2025-03-24
[MINOR] Enable jitpack.io repo only when brotli is required

pan3793:
### Rationale for this change

`jitpack.io` Maven repo was added globally but for conditionally `brotli-codec`, my network is very slow to access `jitpack.io` and I'm using an Apple Silicon machine which does not support `brotli-codec`, it causes a fresh `mvn install` very slow due to each jar pull will try `jitpack.io`(I don't understand why Maven does not try Maven Central first ...)

Move `jitpack.io` Maven repo definition from global to profile `non-aarch64`.

### Are these changes tested?

Pass GHA.

No.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,15 +69,6 @@
     </mailingList>
   </mailingLists>
 
-  <repositories>
-    <repository>
-      <id>jitpack.io</id>
-      <url>https://jitpack.io</url>
-      <name>Jitpack.io repository</name>
-      <!-- needed for brotli-codec -->
-    </repository>
-  </repositories>
-
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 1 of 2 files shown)


wgtmac (approved):
Seems reasonable to me. @Fokko WDYT?

pom.xml
```diff
@@ -69,15 +69,6 @@
     </mailingList>
   </mailingLists>
 
-  <repositories>
-    <repository>
-      <id>jitpack.io</id>
-      <url>https://jitpack.io</url>
-      <name>Jitpack.io repository</name>
-      <!-- needed for brotli-codec -->
```
Fokko, line 77: Should we still keep the comment?

Fokko (approved):
Seems fair to me 👍

pom.xml
pan3793: restored

PR #3184, opened 2025-04-08
MINOR: Fix display of logicalTypeAnnotation for parquet cli

pan3793:
### Rationale for this change

I found the partial output of `parquet footer xxx.parquet` is incorrect, for example,
```
"logicalTypeAnnotation" : { }
```
while the value is actually a `DateLogicalTypeAnnotation`

Supply a custom `JsonSerializer` for `LogicalTypeAnnotation`

### Are these changes tested?

Currently, there is no output verification for parquet-cli modules, I manually checked the output of `org.apache.parquet.cli.commands.ShowFooterCommandTest`

```patch
      }, {
        "name" : "date_field",
        "repetition" : "REQUIRED",
-       "logicalTypeAnnotation" : { },
+       "logicalTypeAnnotation" : "DATE",
        "id" : null,
        "primitive" : "INT32",
        "length" : 0,
        "decimalMeta" : null,
        "columnOrder" : {
          "columnOrderName" : "TYPE_DEFINED_ORDER"
        }
      } ],
```

Only affects the output of parquet-cli.

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/util/RawUtils.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/util/RawUtils.java
@@ -103,8 +103,8 @@ public class RawUtils {
   static class LogicalTypeAnnotationJsonSerializer extends JsonSerializer<LogicalTypeAnnotation> {
 
     @Override
-    public void serialize(
-        LogicalTypeAnnotation value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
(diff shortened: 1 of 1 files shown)


pan3793:
cc @wgtmac @Fokko

Fokko (approved):
Thanks for fixing this @pan3793! 🙌 This makes sense to me. Would it be possible to throw in a test? Just to avoid that we break this in the future

pan3793:
@Fokko it has non-trivial work to add test to parquet-cli module to verify the output of commands, see https://github.com/apache/parquet-java/pull/3187

PR #3186, opened 2025-04-10
Minor: Use logicaltypes constants in ParquetMetadataConverter

aihuaxu:
In ParquetMetadataConverter, we can reuse defined constant types rather than recreating a new instance. The existing tests should have the coverage.


wgtmac (approved):

Fokko:
Good one @aihuaxu, thanks for cleaning this up 👍

PR #3187, opened 2025-04-10
[DO-NOT-REVIEW][TEST] Verify outputs for parquet cli

pan3793:

### DIFF
--- a/parquet-cli/src/test/resources/cli-output/show_footer_command.txt
+++ b/parquet-cli/src/test/resources/cli-output/show_footer_command.txt
@@ -1,0 +1,331 @@
+{
+  "fileMetaData" : {
+    "schema" : {
+      "name" : "schema",
+      "repetition" : "REPEATED",
+      "logicalTypeAnnotation" : null,
+      "id" : null,
+      "fields" : [ {
+        "name" : "int32_field",
+        "repetition" : "REQUIRED",
+        "logicalTypeAnnotation" : null,
+        "id" : null,
+        "primitive" : "INT32",
+        "length" : 0,
+        "decimalMeta" : null,
+        "columnOrder" : {
+          "columnOrderName" : "TYPE_DEFINED_ORDER"
+        }
+      }, {
+        "name" : "int64_field",
+        "repetition" : "REQUIRED",
+        "logicalTypeAnnotation" : null,
+        "id" : null,
+        "primitive" : "INT64",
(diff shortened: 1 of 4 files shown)


parquet-cli/src/test/resources/cli-output/show_footer_command.txt
```diff
@@ -0,0 +1,331 @@
...
+          "RLE_DICTIONARY" : 1
+        },
+        "usesV2Pages" : true
+      },
+      "properties" : {
+        "codec" : "UNCOMPRESSED",
+        "path" : {
+          "p" : [ "binary_field" ]
+        },
+        "encodings" : [ "PLAIN", "RLE_DICTIONARY" ]
```
pan3793, line 243: it's a HashSet, the output is unstable, see https://github.com/FasterXML/jackson-databind/issues/3166

parquet-cli/src/test/resources/cli-output/show_footer_command_raw.txt
```diff
@@ -0,0 +1,280 @@
...
+          "encoding" : 8,
+          "page_type" : 3
+        } ],
+        "encodings" : [ 8, 0 ],
+        "num_values" : 10,
+        "path_in_schema" : [ "flba_field" ],
+        "size_statistics" : { },
+        "statistics" : {
+          "max_value" : "�\u0014y\u001DZ���6��m",
+          "min_value" : "\u0013�Ш]�.\u000F�Y>",
```
pan3793, line 197: it also requires a custom serializer

parquet-cli/src/test/resources/cli-output/show_footer_command.txt
```diff
@@ -0,0 +1,331 @@
...
+      "rowGroupOrdinal" : 0,
+      "encodingStats" : {
+        "dictStats" : { },
+        "dataStats" : {
+          "DELTA_BINARY_PACKED" : 1
+        },
+        "usesV2Pages" : true
+      },
+      "properties" : {
+        "codec" : "UNCOMPRESSED",
```
pan3793, line 144: this test uses UNCOMPRESSED thus the size of data SHOULD be platform-agnostic, while if it uses codec like zstd, gzip, the compressed size might be different in different platforms, let's say I use macOS M1 machine to produce the golden file, which might be different from the output of Linux x64 machine.

PR #3189, opened 2025-04-11
GH-3188: Set the global configured column stats enable flag to default

### LINKED ISSUE #3188: Set the global configured column stats enable flag to default

When user configures a default for columns statistics enabled, it does not take effect as the hardcoded default true is always returned.

The following line is never reached.
https://github.com/apache/parquet-java/blob/master/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java#L350

### Component(s)

Core

huaxiangsun:

Closes #3188


huaxiangsun:
> LGTM. 
Thanks for quick review, @wgtmac !

huaxingao (approved):
LGTM

huaxiangsun:
> LGTM

Thanks @huaxingao

Issue #3193, opened 2025-04-15
Parquet writing to HDFS may deadlock


wangyum:
### Describe the bug, including details regarding any error messages, version, and platform.

Parquet version: 1.15.1
Hadoop version: 3.3.3
Spark version: 3.5.0

Stack trace:
```
Found one Java-level deadlock:
=============================
"Executor 566 task launch worker for task 202024534, task 19644.1 in stage 13967543.0 of app application_1736396393732_100191":
  waiting to lock monitor 0x00007f9435525aa0 (object 0x00007f9575000a70, a java.lang.Object),
  which is held by "Task reaper-9"

"Task reaper-9":
  waiting to lock monitor 0x00007fa06b315500 (object 0x00007f963d0af788, a org.apache.hadoop.hdfs.DFSOutputStream),
  which is held by "Executor 566 task launch worker for task 202024534, task 19644.1 in stage 13967543.0 of app application_1736396393732_100191"

Java stack information for the threads listed above:
===================================================
"Executor 566 task launch worker for task 202024534, task 19644.1 in stage 13967543.0 of app application_1736396393732_100191":
	at org.apache.spark.util.UninterruptibleThread.interrupt(UninterruptibleThread.scala:96)
	- waiting to lock <0x00007f9575000a70> (a java.lang.Object)
	at org.apache.hadoop.hdfs.DataStreamer.waitAndQueuePacket(DataStreamer.java:989)
	- locked <0x00007f963d0af760> (a java.util.LinkedList)
	at org.apache.hadoop.hdfs.DFSOutputStream.enqueueCurrentPacket(DFSOutputStream.java:496)
	at org.apache.hadoop.hdfs.DFSOutputStream.enqueueCurrentPacketFull(DFSOutputStream.java:505)
	- locked <0x00007f963d0af788> (a org.apache.hadoop.hdfs.DFSOutputStream)
	at org.apache.hadoop.hdfs.DFSOutputStream.writeChunk(DFSOutputStream.java:445)
	- locked <0x00007f963d0af788> (a org.apache.hadoop.hdfs.DFSOutputStream)
	at org.apache.hadoop.fs.FSOutputSummer.writeChecksumChunks(FSOutputSummer.java:218)
	at org.apache.hadoop.fs.FSOutputSummer.flushBuffer(FSOutputSummer.java:165)
	- eliminated <0x00007f963d0af788> (a org.apache.hadoop.hdfs.DFSOutputStream)
[...]

PR #3196, opened 2025-04-18
[GH-3193] Avoid deadlock in writeAllTo by removing WritableByteChannel usage

wangyum:
### Rationale for this change

The reason changing the original implementation (using `Channels.newChannel(out)`) to directly writing via OutputStream resolves the deadlock issue is as follows:

In the original implementation, using `Channels.newChannel(out)` introduces an internal lock (`WritableByteChannelImpl`) that interacts with the underlying `OutputStream`. When Spark's task interruption mechanism (`Task reaper` thread) attempts to interrupt or close the channel, it acquires locks in a different order compared to the executor thread writing data. Specifically:

- The executor thread holds the `DFSOutputStream` lock and waits for the internal lock of `WritableByteChannelImpl`.

- The `Task reaper` thread holds the internal lock of `WritableByteChannelImpl` and waits for the `DFSOutputStream` lock (during `hflush()`).
This conflicting lock acquisition order results in a deadlock.

By directly writing to the `OutputStream` without using `Channels.newChannel`, the intermediate locking introduced by `WritableByteChannelImpl` is eliminated. This removes the conflicting lock order scenario, thus resolving the deadlock.

- Removed the usage of `Channels.newChannel(out)` and `WritableByteChannel`.
- Updated the `writeAllTo` method to write data directly to the `OutputStream` using `ByteBuffer` operations.

### Are these changes tested?

Manual test.
There is no zombie task(deadlock) after applying this patch:
image

No.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
@@ -23,8 +23,6 @@ import static java.lang.String.format;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
-import java.nio.channels.Channels;
-import java.nio.channels.WritableByteChannel;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -72,9 +70,15 @@ public class ConcatenatingByteBufferCollector extends BytesInput implements Auto
 
   @Override
   public void writeAllTo(OutputStream out) throws IOException {
-    WritableByteChannel channel = Channels.newChannel(out);
     for (ByteBuffer buffer : slabs) {
-      channel.write(buffer.duplicate());
+      ByteBuffer dup = buffer.duplicate();
+      if (dup.hasArray()) {
+        out.write(dup.array(), dup.arrayOffset() + dup.position(), dup.remaining());
+      } else {
+        byte[] bytes = new byte[dup.remaining()];
+        dup.get(bytes);
+        out.write(bytes);
+      }
     }
   }
 


wangyum:
@gszadovszky Would you please take a look at this patch?

vrozov:
The new implementation is uninterruptible compared to the previous implementation that uses `WritableByteChannel`.

gszadovszky:
@wangyum, not sure I get the purpose of this fix. We have already closed the related Parquet issue saying that the bug is in Spark. See [here](https://github.com/apache/parquet-java/issues/3193#issuecomment-2810010827). Am I misinterpreting something?

vrozov:
@wangyum There is a [fix](https://github.com/apache/spark/pull/50594) on the Spark side and if it is critical CVE, IMO it will be better to ask Spark community to backport the fix to 3.5.x so existing Spark users can benefit from Parquet upgrade. Can you please try to cherry pick and test it.

wgtmac:
I agree with @vrozov and @gszadovszky that Spark not Parquet should fix this. Since the root cause is not from Parquet, other user dependency may also trigger this bug and workaround on the Parquet side does not help either.

pan3793:
@wgtmac possible to trigger patch releases for older Parquet branches? e.g. 1.13 (used by Spark 3.5) and 1.14

vrozov:
> What should users do if they are still using Spark 3.0 - Spark 3.4? The Spark community no longer maintains these branches.

@wangyum  Those that use Spark 3.0 - 3.4 should upgrade to Spark 3.5.x so they can continue to receive CVE and other fixes from the Spark community. Additionally from the discussion on the Apache Spark, the Parquet [CVE-2025-30065](https://github.com/advisories/GHSA-2c59-37c4-qrx5) does not affect Spark, so Spark community even considered not upgrading parquet-java to 1.5.1.

I also think that avoiding using `WritableByteChannel` is **not** the correct generic fix. While it may work for Spark, it may break other consumers of the parquet-java as it converts interruptible write to uninterruptible.

As the last resort, both Spark and Parquet are open source and it is possible to recompile both using your own fork.

wangyum:
> it converts interruptible write to uninterruptible

Before [Parquet 1.14.0](https://github.com/apache/parquet-java/commit/274dc51bc9e5cc880ba3c77c3db826d2a4943965). It also uninterruptible write:
https://github.com/apache/parquet-java/blob/274dc51bc9e5cc880ba3c77c3db826d2a4943965/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteArrayCollector.java#L51-L56

vrozov:
It is interruptible starting with 1.14 and going back to uninterruptible is not a good option, IMO.  That will even impact Spark as it will now have to write all data instead of I/O (write) being interrupted and aborted. If Spark needs this task to run without being interrupted it should use `runUninterruptibly`.

PR #3197, opened 2025-04-18
GH-3116: Implement the decoding of Variant values

### LINKED ISSUE #3116: Implement the Variant binary encoding

Implement the encoding and decoding of the [Variant binary encoding](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md)

### Component(s)

_No response_

gene-db:
### Rationale for this change

This is a reference implementation for decoding Variant binary encoded values, according to the specification.

A new module for decoding the Variant binary format.

### Are these changes tested?

Added unit tests.

No

Closes #3116


parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  Variant(byte[] value, int valuePos, byte[] metadata, int metadataPos) {
+    this(
+        ByteBuffer.wrap(value, valuePos, value.length - valuePos),
+        ByteBuffer.wrap(metadata, metadataPos, metadata.length - metadataPos));
+  }
+
+  Variant(ByteBuffer value, ByteBuffer metadata) {
+    this.value = value.asReadOnlyBuffer();
```
rdblue, line 53: Do you need to verify endianness or are you only reading single bytes from the buffer?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,288 @@
...
+    this.value.mark();
+
+    this.metadata = metadata.asReadOnlyBuffer();
+    this.metadata.mark();
+
+    // There is currently only one allowed version.
+    if ((metadata.get(metadata.position()) & VariantUtil.VERSION_MASK) != VariantUtil.VERSION) {
+      throw new UnsupportedOperationException(String.format(
+          "Unsupported variant metadata version: %02X",
```
rdblue, line 62: The only allowed version is 1 and only 4 bits are used? Why encode the unsupported version as 2 hex digits?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+   * nanoseconds from the Unix epoch.
+   * @param value The Variant value
+   * @return The long value
+   */
+  public static long getLong(ByteBuffer value) {
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    String exceptionMessage =
+        "Expect type to be one of: BYTE, SHORT, INT, LONG, TIMESTAMP, TIMESTAMP_NTZ, TIME, TIMESTAMP_NANOS, TIMESTAMP_NANOS_NTZ";
```
rdblue, line 502: There's no need for a local variable. This can either be in the `IllegalStateException` or should be a static constant.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+    long signedByteValue = buffer.get(pos + numBytes - 1);
+    result |= signedByteValue << (8 * (numBytes - 1));
+    return result;
+  }
+
+  /**
+   * Read a little-endian unsigned int value from `bytes[pos, pos + numBytes)`. The value must fit
+   * into a non-negative int (`[0, Integer.MAX_VALUE]`).
+   */
+  static int readUnsigned(byte[] bytes, int pos, int numBytes) {
```
rdblue, line 263: This isn't used. Can you remove it?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+
+  /**
+   * Write the least significant `numBytes` bytes in `value` into `bytes[pos, pos + numBytes)` in
+   * little endian.
+   * @param bytes The byte array to write into
+   * @param pos The starting index of the byte array to write into
+   * @param value The value to write
+   * @param numBytes The number of bytes to write
+   */
+  public static void writeLong(byte[] bytes, int pos, long value, int numBytes) {
```
rdblue, line 198: This isn't used.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,288 @@
...
+   */
+  public float getFloat() {
+    return VariantUtil.getFloat(value);
+  }
+
+  /**
+   * @return the binary value
+   */
+  public byte[] getBinary() {
+    return VariantUtil.getBinary(value);
```
rdblue, line 139: Should this also return ByteBuffer? If not this requires a copy. If you used ByteBuffer then whether to copy is a decision made by the caller.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+    int idStartOffset = 1 + sizeBytes;
+    int offsetStartOffset = idStartOffset + numElements * idSize;
+    int dataStartOffset = offsetStartOffset + (numElements + 1) * offsetSize;
+    return new ObjectInfo(numElements, idSize, offsetSize, idStartOffset, offsetStartOffset, dataStartOffset);
+  }
+
+  /**
+   * A helper class representing the details of a Variant array, used for `ArrayHandler`.
+   */
+  public static class ArrayInfo {
```
rdblue, line 716: Why is this public? It should only be used internally right?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+    value.position(start);
+    ByteBuffer newSlice = value.slice();
+    value.position(oldPos);
+    return newSlice;
+  }
+
+  /**
+   * A helper class representing the details of a Variant object, used for `ObjectHandler`.
+   */
+  public static class ObjectInfo {
```
rdblue, line 657: Similar to the comment below, I think we should start with this as package-private and open it up later.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+  /**
+   * Slices the `value` buffer starting from `start` index.
+   * @param value The ByteBuffer to slice
+   * @param start The starting index of the slice
+   * @return The sliced ByteBuffer
+   */
+  public static ByteBuffer slice(ByteBuffer value, int start) {
+    int oldPos = value.position();
+    value.position(start);
+    ByteBuffer newSlice = value.slice();
```
rdblue, line 649: I think this shouldn't modify the `value` buffer and modify it back to the original state. Instead, you could create a duplicate and alter just the copy.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+    return new java.util.UUID(bb.getLong(), bb.getLong());
+  }
+
+  /**
+   * Slices the `value` buffer starting from `start` index.
+   * @param value The ByteBuffer to slice
+   * @param start The starting index of the slice
+   * @return The sliced ByteBuffer
+   */
+  public static ByteBuffer slice(ByteBuffer value, int start) {
```
rdblue, line 646: Can this be private?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+  public static java.util.UUID getUUID(ByteBuffer value) {
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE || typeInfo != UUID) {
+      throw unexpectedType(Type.UUID);
+    }
+    int start = value.position() + 1;
+    checkIndex(start + UUID_SIZE - 1, value.limit());
+    ByteBuffer bb = VariantUtil.slice(value, start).order(ByteOrder.BIG_ENDIAN);
```
rdblue, line 636: Usually you would omit the `VariantUtil` prefix when using methods from this class.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+ * - Version: 1-byte unsigned integer. The only acceptable value is 1 currently.
+ * - Dictionary size: 4-byte little-endian unsigned integer. The number of keys in the
+ *                    dictionary.
+ * - Offsets: (size + 1) * 4-byte little-endian unsigned integers. `offsets[i]` represents the
+ * starting position of string i, counting starting from the address of `offsets[0]`. Strings
+ * must be stored contiguously, so we don’t need to store the string size, instead, we compute it
+ * with `offset[i + 1] - offset[i]`.
+ * - UTF-8 string data.
+ */
+public class VariantUtil {
```
rdblue, line 45: I don't think that most of this class needs to be public since it has helper methods for the `Variant` API. I think this class should be package-private and anything that needs to be public should move to `Variant` or should be a top-level class. I think the only thing that is needed externally is the type enum?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,288 @@
...
+
+  Variant(byte[] value, int valuePos, byte[] metadata, int metadataPos) {
+    this(
+        ByteBuffer.wrap(value, valuePos, value.length - valuePos),
+        ByteBuffer.wrap(metadata, metadataPos, metadata.length - metadataPos));
+  }
+
+  Variant(ByteBuffer value, ByteBuffer metadata) {
+    this.value = value.asReadOnlyBuffer();
+    this.value.mark();
```
rdblue, line 54: No need to mark if reset is never called.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,288 @@
...
+   * an object. This is a performance optimization to avoid the overhead of binary search for a
+   * short list.
+   */
+  static final int BINARY_SEARCH_THRESHOLD = 32;
+
+  public Variant(byte[] value, byte[] metadata) {
+    this(value, 0, metadata, 0);
+  }
+
+  Variant(byte[] value, int valuePos, byte[] metadata, int metadataPos) {
```
rdblue, line 46: For this constructor, I'd normally expect a length for value and metadata as well.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+          case TIME:
+            return Type.TIME;
+          case TIMESTAMP_NANOS:
+            return Type.TIMESTAMP_NANOS;
+          case TIMESTAMP_NANOS_NTZ:
+            return Type.TIMESTAMP_NANOS_NTZ;
+          case UUID:
+            return Type.UUID;
+          default:
+            throw new UnknownVariantTypeException(typeInfo);
```
rdblue, line 387: I'd probably avoid adding the exception type and just throw UnsupportedOperationException.

rdblue (changes requested):
This is really close. The tests all look great. I think the main thing is to clean up the unused methods and restrict the visibility of most of the `VariantUtil` class that doesn't need to be exposed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+   * Date value. Content is 4-byte little-endian signed integer that represents the
+   * number of days from the Unix epoch.
+   */
+  public static final int DATE = 11;
+  /**
+   * Timestamp value. Content is 8-byte little-endian signed integer that represents the number of
+   * microseconds elapsed since the Unix epoch, 1970-01-01 00:00:00 UTC. It is displayed to users in
+   * their local time zones and may be displayed differently depending on the execution environment.
+   */
+  public static final int TIMESTAMP = 12;
```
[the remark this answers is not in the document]
gene-db: Renamed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Removed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+    SHORT,
+    INT,
+    LONG,
+    STRING,
+    DOUBLE,
+    DECIMAL4,
+    DECIMAL8,
+    DECIMAL16,
+    DATE,
+    TIMESTAMP,
```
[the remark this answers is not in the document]
gene-db: Renamed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Updated.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+  }
+
+  /**
+   * Computes the actual size (in bytes) of the Variant value at `value[pos...]`.
+   * `value.length - pos` is an upper bound of the size, but the actual size may be smaller.
+   * @param value The Variant value
+   * @param pos The starting index of the Variant value
+   * @return The actual size of the Variant value
+   */
+  public static int valueSize(byte[] value, int pos) {
```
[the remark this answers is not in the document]
gene-db: Removed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Done.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Yeah, that make sense. Updated.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: `Variant` uses this also, so made it package private.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,795 @@
...
+          String.format("Invalid dictionary id: %d. dictionary size: %d", id, dictSize));
+    }
+    // The offset list after the header byte, and a `dictSize` with `offsetSize` bytes.
+    int offsetListPos = metadata.position() + 1 + offsetSize;
+    // The data starts after the offset list, and `(dictSize + 1)` offset values.
+    int dataPos = offsetListPos + (dictSize + 1) * offsetSize;
+    int offset = readUnsigned(metadata, offsetListPos + (id) * offsetSize, offsetSize);
+    int nextOffset = readUnsigned(metadata, offsetListPos + (id + 1) * offsetSize, offsetSize);
+    if (offset > nextOffset) {
+      throw new MalformedVariantException(
```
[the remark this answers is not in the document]
gene-db: Done.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Updated to package private.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Updated to package private.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
gene-db: Yeah, that make sense. I made this all package private, and moved the type enum over to `Variant`.

gene-db (commented):
@rdblue Thanks! I updated the PR.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+  public Variant(byte[] value, byte[] metadata) {
+    this(value, 0, value.length, metadata, 0, metadata.length);
+  }
+
+  public Variant(byte[] value, int valuePos, int valueLength, byte[] metadata, int metadataPos, int metadataLength) {
+    this(ByteBuffer.wrap(value, valuePos, valueLength), ByteBuffer.wrap(metadata, metadataPos, metadataLength));
+  }
+
+  public Variant(ByteBuffer value, ByteBuffer metadata) {
+    // THe buffers are read single-byte at a time, so the endianness of the input buffers
```
RussellSpitzer, line 51:
```suggestion
    // The buffers are read a single-byte at a time, so the endianness of the input buffers 
```

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+    // are not important.
```
RussellSpitzer, line 52:
```suggestion
    // is not important.
```

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+
+  /**
+   * @return the type of the variant value
+   */
+  public Type getType() {
+    return VariantUtil.getType(value);
+  }
+
+  /**
+   * @return the number of object fields in the variant. `getType()` must be `Type.OBJECT`.
```
RussellSpitzer, line 189: Is the note here on getType because this is unchecked?

parquet-variant/src/main/java/org/apache/parquet/variant/MalformedVariantException.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.parquet.variant;
+
+/**
+ * An exception indicating that the Variant is malformed.
+ */
+public class MalformedVariantException extends RuntimeException {
```
rdblue, line 22: This isn't used. Can you remove it?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
rdblue: What is the behavior if the variant is not an object?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+      }
+    }
+    return null;
+  }
+
+  /**
+   * A field in a Variant object.
+   */
+  public static final class ObjectField {
```
rdblue, line 253: I don't think this needs to be public because it isn't returned through the public methods.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE) {
+      throw new IllegalStateException(
+          "Expect type to be one of: BYTE, SHORT, INT, LONG, TIMESTAMP, TIMESTAMP_NTZ, TIME, TIMESTAMP_NANOS, TIMESTAMP_NANOS_NTZ");
+    }
+    switch (typeInfo) {
+      case INT8:
+        return readLong(value, value.position() + 1, 1);
```
rdblue, line 374: Not a big deal, but it's a little odd that the public `getShort` and similar methods end up calling this, which checks the type to see how many bytes to read but then returns a long. Then that value is checked to ensure it fits in a short and cast. It seems easier to just have a `getShort` method in `VariantUtil` that produces a short directly.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+    }
+    throw unexpectedType(Variant.Type.STRING);
+  }
+
+  static java.util.UUID getUUID(ByteBuffer value) {
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE || typeInfo != UUID) {
+      throw unexpectedType(Variant.Type.UUID);
```
rdblue, line 499: Minor: it would be nice to include what the type actual was.

Is there a guarantee that the type is always checked when reading through these util methods?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+            return Variant.Type.UUID;
+          default:
+            throw new UnsupportedOperationException(
+                String.format("Unknown type in Variant. primitive type: %d", typeInfo));
+        }
+    }
+  }
+
+  private static IllegalArgumentException unexpectedType(Variant.Type type) {
+    return new IllegalArgumentException("Expected type to be " + type);
```
rdblue, line 333: I think this may be phrased incorrectly. Isn't the problem that a method like `getUUID` was called for a value that is not a UUID? In that case do we trust the user called the right method and the expectation is valid? Instead, I would say something more direct that doesn't make that assumption: `"Cannot read %s value as a %s", actualType, expectedType`

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+    DECIMAL4,
+    DECIMAL8,
+    DECIMAL16,
+    DATE,
+    TIMESTAMP_TZ,
+    TIMESTAMP_NTZ,
+    FLOAT,
+    BINARY,
+    TIME,
+    TIMESTAMP_NANOS,
```
rdblue, line 176: This should also have a `_TZ` suffix.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+   * @param value The Variant value
+   * @return The long value
+   */
+  static long getLong(ByteBuffer value) {
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE) {
+      throw new IllegalStateException(
+          "Expect type to be one of: BYTE, SHORT, INT, LONG, TIMESTAMP, TIMESTAMP_NTZ, TIME, TIMESTAMP_NANOS, TIMESTAMP_NANOS_NTZ");
```
rdblue, line 370: References to `TIMESTAMP` in this method and javadoc should also include a suffix. We want to avoid any ambiguity because Spark's `timestamp` differs from the SQL standard.

rdblue:
Thanks, @gene-db! There are a few minor things to fix, like some of the lingering `TIMESTAMP` references and the unused class. Overall I think this is good to go once the minor changes are done (and when @RussellSpitzer's comments are addressed).

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -0,0 +1,313 @@
...
+  /**
+   * @return the number of object fields in the variant. `getType()` must be `Type.OBJECT`.
+   */
+  public int numObjectElements() {
+    return VariantUtil.getObjectInfo(value).numElements;
+  }
+
+  /**
+   * Returns the object field Variant value whose key is equal to `key`.
+   * Return null if the key is not found. `getType()` must be `Type.OBJECT`.
```
RussellSpitzer, line 197: Same question about what is the "notObject" behavior

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE || (typeInfo != TRUE && typeInfo != FALSE)) {
+      throw unexpectedType(Variant.Type.BOOLEAN);
+    }
+    return typeInfo == TRUE;
+  }
+
+  /**
+   * Returns a long value from Variant value `value[pos...]`.
+   * It is only legal to call it if `getType` returns one of Type.BYTE, SHORT, INT, LONG,
```
RussellSpitzer, line 352: This is also unchecked correct? So it's up to the caller to verify this?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,659 @@
...
+        return readLong(value, value.position() + 1, 4);
+      case INT64:
+      case TIMESTAMP_TZ:
+      case TIMESTAMP_NTZ:
+      case TIME:
+      case TIMESTAMP_NANOS:
+      case TIMESTAMP_NANOS_NTZ:
+        return readLong(value, value.position() + 1, 8);
+      default:
+        throw new IllegalStateException(
```
RussellSpitzer, line 388: Should this be unexpectedType(...)?

RussellSpitzer (approved):
I had a few nits on the exceptions being thrown, it looks like we aren't using one of the 'unexepectedType" methods which seems to generate a very similar exception. We also throw both IllegalArg and IllegalState but I think the intent was to throw the IllegalArg all the time? 

The other suggestion I would have is to add some more tests for these bad cases. I think we currently only have tests for the good path but none for the exception handling behavior.

I tried my best to go through the bit/shifting logic but there is a bit too much for me to go through so I relied on the tests to make sure there is good coverage there as well as the fact that this impl is also in use in the Spark repo.

gene-db (commented):
@rdblue Thanks for the reviews! I updated the PR and added some "invalid type" tests.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,773 @@
...
+  static final int DECIMAL8 = 9;
+  /** 16-byte decimal. Content is 1-byte scale + 16-byte little-endian signed integer. */
+  static final int DECIMAL16 = 10;
+  /**
+   * Date value. Content is 4-byte little-endian signed integer that represents the
+   * number of days from the Unix epoch.
+   */
+  static final int DATE = 11;
+  /**
+   * Timestamp value. Content is 8-byte little-endian signed integer that represents the number of
```
rdblue, line 123: Should have a _tz suffix.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,773 @@
...
+   */
+  static final int LONG_STR = 16;
+  /**
+   * Time value. Values can be from 00:00:00 to 23:59:59.999999.
+   * Content is 8-byte little-endian unsigned integer that represents the number of microseconds
+   * since midnight.
+   */
+  static final int TIME = 17;
+  /**
+   * Timestamp nanos value. Similar to `TIMESTAMP_TZ`, but represents the number of nanoseconds
```
rdblue, line 152: Missing _tz suffix.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,773 @@
...
+   * elapsed since the Unix epoch, 1970-01-01 00:00:00 UTC.
+   */
+  static final int TIMESTAMP_NANOS_TZ = 18;
+  /**
+   * Timestamp nanos (without timestamp) value. It has the same content as `TIMESTAMP_NANOS_TZ` but
```
rdblue, line 157: without time zone?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
rdblue: I've gone through and I think all of the methods do check for compatible types now.

rdblue (approved):
I think this is ready to go. There are a few nits on the comments and an error message, but nothing major. I'll merge this soon to unblock the next commit.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -0,0 +1,773 @@
...
+    checkIndex(value.position(), value.limit());
+    int basicType = value.get(value.position()) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(value.position()) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    if (basicType != PRIMITIVE || typeInfo != BINARY) {
+      throw unexpectedType(Variant.Type.BINARY, value);
+    }
+    int start = value.position() + 1 + U32_SIZE;
+    int length = readUnsigned(value, value.position() + 1, U32_SIZE);
+    checkIndex(start + length - 1, value.limit());
+    return slice(value, start);
```
gszadovszky, line 577: @gene-db, sorry for the late comment, but shouldn't the slice also deal with `length`? This way we pass back the entire remaining buffer as the binary value, don't we?

PR #3199, opened 2025-04-23
GH-3198: Allow specifying trusted classes by class name

### LINKED ISSUE #3198: Update the way serializable packages can be specified

Currently serializable packages can only be specified by a system property. Add new ways for easy to use.

### Component(s)

Avro

gszadovszky:
### Rationale for this change
Trusted classes are only able to be specified by their packages (and super packages) which is not precise enough.

Allow to set the trusted classes by their names directly. Also, remove the default packages so the user must specify the classes they want to allow to be loaded.

### Are these changes tested?
Covered by TestStringBehavior.test*ValidationFail

Introduced the new Parquet/Hadoop property `parquet.avro.serializable.classes`

Closes #3198


gszadovszky:
@wgtmac, could you check this, plz?

parquet-avro/README.md
```diff
@@ -26,21 +26,22 @@ Apache Avro integration
...
-| `parquet.avro.writeFixedAsInt96`    | `String` | Comma separated list of paths pointing to Avro schema elements which are to be converted to `INT96` Parquet types.<br/>The path is a `'.'` separated list of field names and does not contain the name of the schema nor the namespace. The type of the referenced schema elements must be `fixed` with the size of 12 bytes.<br/>**NOTE: The `INT96` Parquet type is deprecated. This option is only to support old data.** |
+| Name                                    | Type      | Description                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
wgtmac, line 40: Could we avoid these irrelevant changes? Does IntelliJ IDEA format it automatically?

parquet-avro/src/main/java/org/apache/parquet/avro/ReflectClassValidator.java
```diff
@@ -0,0 +1,94 @@
...
+
+  static class PackageValidator extends ReflectClassValidator {
+
+    private final List<String> trustedPackagePrefixes = Stream.of(AvroConverters.SERIALIZABLE_PACKAGES)
+        .map(p -> p.endsWith(".") ? p : p + ".")
+        .collect(Collectors.toList());
+    private final boolean trustAllPackages =
+        trustedPackagePrefixes.size() == 1 && "*".equals(trustedPackagePrefixes.get(0));
+    // The primitive "class names" based on Class.isPrimitive()
+    private static final Set<String> PRIMITIVES = new HashSet<>(Arrays.asList(
```
wgtmac, line 40: Why do we need this?

wgtmac (commented):
Thanks! It generally looks good to me. However, I'm not an Avro expert so it would be good if @rdblue @Fokko can sign off.

parquet-avro/src/main/java/org/apache/parquet/avro/ReflectClassValidator.java
gszadovszky: You mean primitives? I'm just trying to mirror the original behavior without actually loading the class (checking the name only).

wgtmac (approved):

parquet-avro/src/main/java/org/apache/parquet/avro/ReflectClassValidator.java
wgtmac: Just wonder what will happen if we remove this.

gszadovszky: I'm not a expert in Avro either. Not sure if we can hit this with a primitive, but that's how it was originally implemented here and how it is currently implemented in Avro.

PR #3200, opened 2025-04-24
PARQUET-2417: Add `geometry` and `geography` logical type annotations

zhangfengcdt:
### Rationale for this change
This is to add geometry and geography logical types in parquet-java according to the new parquet-format spec changes to add the geo types - https://github.com/apache/parquet-format/blob/master/Geospatial.md.

The Geometry and Geography logical type have been added to LogicalTypeAnnotation. For geo columns, the corresponding Parquet group is annotated as GEOMETRY / GEOGRAPHY, with optimal parameters of crs and edge algorithm. Unit tests to test the new logical types are also added.

### Are these changes tested?
Yes, TestTypeBuilders

Yes, Geometry and Geography types are now available.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
@@ -41,6 +41,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
 import org.apache.parquet.Preconditions;
+import org.apache.parquet.format.EdgeInterpolationAlgorithm;
 
 public abstract class LogicalTypeAnnotation {
   enum LogicalTypeToken {
@@ -153,6 +154,40 @@ public abstract class LogicalTypeAnnotation {
       @Override
       protected LogicalTypeAnnotation fromString(List<String> params) {
         return float16Type();
+      }
+    },
+    GEOMETRY {
+      @Override
+      protected LogicalTypeAnnotation fromString(List<String> params) {
+        if (params.size() > 1) {
+          throw new RuntimeException("Expecting only crs for geometry logical type, got " + params.size());
+        }
+        if (params.size() == 1) {
+          String crs = params.get(0);
+          return geometryType(crs);
+        } else {
+          return geometryType();
+        }
+      }
+    },
+    GEOGRAPHY {
+      @Override
+      protected LogicalTypeAnnotation fromString(List<String> params) {
+        if (params.size() > 2) {
+          throw new RuntimeException(
+              "Expecting at most 2 parameters for geography logical type (crs and edgeAlgorithm), got "
+                  + params.size());
+        }
+        if (params.size() == 1) {
+          String crs = params.get(0);
+          return geographyType(crs, null);
+        }
+        if (params.size() == 2) {
+          String crs = params.get(0);
+          String edgeAlgorithm = params.get(1);
+          return geographyType(crs, EdgeInterpolationAlgorithm.valueOf(edgeAlgorithm));
+        }
+        return geographyType();
       }
     },
     UNKNOWN {
@@ -332,6 +367,22 @@ public abstract class LogicalTypeAnnotation {
 
   public static Float16LogicalTypeAnnotation float16Type() {
     return Float16LogicalTypeAnnotation.INSTANCE;
+  }
+
+  public static GeometryLogicalTypeAnnotation geometryType(String crs) {
+    return new GeometryLogicalTypeAnnotation(crs);
+  }
+
+  public static GeometryLogicalTypeAnnotation geometryType() {
+    return new GeometryLogicalTypeAnnotation("OGC:CRS84");
+  }
+
+  public static GeographyLogicalTypeAnnotation geographyType(String crs, EdgeInterpolationAlgorithm edgeAlgorithm) {
+    return new GeographyLogicalTypeAnnotation(crs, edgeAlgorithm);
+  }
+
+  public static GeographyLogicalTypeAnnotation geographyType() {
+    return new GeographyLogicalTypeAnnotation("OGC:CRS84", null);
   }
 
   public static UnknownLogicalTypeAnnotation unknownType() {
@@ -1183,6 +1234,128 @@ public abstract class LogicalTypeAnnotation {
     }
   }
 
+  public static class GeometryLogicalTypeAnnotation extends LogicalTypeAnnotation {
+    private final String crs;
+
+    private GeometryLogicalTypeAnnotation(String crs) {
+      this.crs = crs;
+    }
+
+    @Override
+    @Deprecated
+    public OriginalType toOriginalType() {
+      return null;
+    }
+
+    @Override
+    public <T> Optional<T> accept(LogicalTypeAnnotationVisitor<T> logicalTypeAnnotationVisitor) {
+      return logicalTypeAnnotationVisitor.visit(this);
+    }
+
+    @Override
+    LogicalTypeToken getType() {
+      return LogicalTypeToken.GEOMETRY;
+    }
+
+    @Override
+    protected String typeParametersAsString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("(").append(crs != null && !crs.isEmpty() ? crs : "").append(")");
+      return sb.toString();
+    }
+
+    public String getCrs() {
+      return crs;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (!(obj instanceof GeometryLogicalTypeAnnotation)) {
+        return false;
+      }
+      GeometryLogicalTypeAnnotation other = (GeometryLogicalTypeAnnotation) obj;
+      return Objects.equals(crs, other.crs);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(crs);
+    }
+
+    @Override
+    PrimitiveStringifier valueStringifier(PrimitiveType primitiveType) {
+      return PrimitiveStringifier.WKB_STRINGIFIER;
+    }
+  }
+
+  public static class GeographyLogicalTypeAnnotation extends LogicalTypeAnnotation {
+    private final String crs;
+    private final EdgeInterpolationAlgorithm edgeAlgorithm;
+
+    private GeographyLogicalTypeAnnotation(String crs, EdgeInterpolationAlgorithm edgeAlgorithm) {
+      this.crs = crs;
+      this.edgeAlgorithm = edgeAlgorithm;
+    }
+
+    @Override
+    @Deprecated
+    public OriginalType toOriginalType() {
+      return null;
+    }
+
+    @Override
+    public <T> Optional<T> accept(LogicalTypeAnnotationVisitor<T> logicalTypeAnnotationVisitor) {
+      return logicalTypeAnnotationVisitor.visit(this);
+    }
+
+    @Override
+    LogicalTypeToken getType() {
+      return LogicalTypeToken.GEOGRAPHY;
+    }
+
+    @Override
+    protected String typeParametersAsString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("(");
+      if (crs != null && !crs.isEmpty()) {
+        sb.append(crs);
+      }
+      if (edgeAlgorithm != null) {
+        if (crs != null && !crs.isEmpty()) sb.append(",");
+        sb.append(edgeAlgorithm);
+      }
+      sb.append(")");
+      return sb.toString();
+    }
+
+    public String getCrs() {
+      return crs;
+    }
+
+    public EdgeInterpolationAlgorithm getEdgeAlgorithm() {
+      return edgeAlgorithm;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (!(obj instanceof GeographyLogicalTypeAnnotation)) {
+        return false;
+      }
+      GeographyLogicalTypeAnnotation other = (GeographyLogicalTypeAnnotation) obj;
+      return Objects.equals(crs, other.crs) && Objects.equals(edgeAlgorithm, other.edgeAlgorithm);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(crs, edgeAlgorithm);
+    }
+
+    @Override
+    PrimitiveStringifier valueStringifier(PrimitiveType primitiveType) {
+      return PrimitiveStringifier.WKB_STRINGIFIER;
+    }
+  }
+
   /**
    * Implement this interface to visit a logical type annotation in the schema.
    * The default implementation for each logical type specific visitor method is empty.
@@ -1256,6 +1429,14 @@ public abstract class LogicalTypeAnnotation {
     }
 
     default Optional<T> visit(Float16LogicalTypeAnnotation float16LogicalType) {
+      return empty();
+    }
+
+    default Optional<T> visit(GeometryLogicalTypeAnnotation geometryLogicalType) {
+      return empty();
+    }
+
+    default Optional<T> visit(GeographyLogicalTypeAnnotation geographyLogicalType) {
       return empty();
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
@@ -41,7 +41,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
 import org.apache.parquet.Preconditions;
-import org.apache.parquet.format.EdgeInterpolationAlgorithm;
+import org.apache.parquet.column.schema.EdgeInterpolationAlgorithm;
 
 public abstract class LogicalTypeAnnotation {
   enum LogicalTypeToken {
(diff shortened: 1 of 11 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1183,6 +1234,136 @@ public boolean equals(Object obj) {
...
+    @Override
+    LogicalTypeToken getType() {
+      return LogicalTypeToken.GEOMETRY;
+    }
+
+    @Override
+    protected String typeParametersAsString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("(").append(crs != null && !crs.isEmpty() ? crs : "").append(")");
+      return sb.toString();
```
wgtmac, line 1264:
```suggestion
      if (crs == null || crs.isEmpty()) {
        return "";
      }
      return String.format("(%s)", crs);
```

This is more readable.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1183,6 +1211,12 @@ LogicalTypeAnnotation getLogicalTypeAnnotation(LogicalType type) {
         return LogicalTypeAnnotation.uuidType();
       case FLOAT16:
         return LogicalTypeAnnotation.float16Type();
+      case GEOMETRY:
+        GeometryType geometry = type.getGEOMETRY();
+        return LogicalTypeAnnotation.geometryType(geometry.getCrs());
+      case GEOGRAPHY:
+        GeographyType geography = type.getGEOGRAPHY();
+        return LogicalTypeAnnotation.geographyType(geography.getCrs(), null);
```
wgtmac, line 1219: Why not getting the edge algo?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,31 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.GeographyLogicalTypeAnnotation geographyLogicalType) {
+      GeographyType geographyType = new GeographyType();
+      if (geographyLogicalType.getCrs() != null) {
+        geographyType.setCrs(geographyLogicalType.getCrs());
+      }
+      if (geographyLogicalType.getEdgeAlgorithm() != null) {
+        EdgeInterpolationAlgorithm algorithm =
+            EdgeInterpolationAlgorithm.valueOf(String.valueOf(geographyLogicalType.getEdgeAlgorithm()));
+        if (algorithm != null) {
```
wgtmac, line 545: It would be good to throw for an unknown value. Otherwise we may silently write an a future new algorithm to null which incorrectly defaults to the spherical algorithm.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -334,6 +369,22 @@ public static Float16LogicalTypeAnnotation float16Type() {
     return Float16LogicalTypeAnnotation.INSTANCE;
   }
 
+  public static GeometryLogicalTypeAnnotation geometryType(String crs) {
+    return new GeometryLogicalTypeAnnotation(crs);
+  }
+
+  public static GeometryLogicalTypeAnnotation geometryType() {
+    return new GeometryLogicalTypeAnnotation("OGC:CRS84");
+  }
```
wgtmac, line 378:
```suggestion
  public static GeometryLogicalTypeAnnotation geometryType() {
    return new GeometryLogicalTypeAnnotation(null);
  }
```

Setting the default to null may save some bytes in the thrift message because we don't need to serialize crs. We have to write the string if the default is `"OGC:CRS84"`.

wgtmac: Maybe we don't have to provide `geometryType()` at all?

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1183,6 +1234,136 @@ public boolean equals(Object obj) {
...
+
+    @Override
+    PrimitiveStringifier valueStringifier(PrimitiveType primitiveType) {
+      return PrimitiveStringifier.WKB_STRINGIFIER;
+    }
+  }
+
+  public static class GeographyLogicalTypeAnnotation extends LogicalTypeAnnotation {
+    private final String crs;
+    private final EdgeInterpolationAlgorithm edgeAlgorithm;
```
wgtmac, line 1293:
```suggestion
    private final EdgeInterpolationAlgorithm algorithm;
```

What about making it shorter and consistent with the spec? I think its meaning is clear in this context.

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1183,6 +1234,136 @@ public boolean equals(Object obj) {
...
+      }
+      sb.append(")");
+      return sb.toString();
+    }
+
+    public String getCrs() {
+      return crs;
+    }
+
+    public EdgeInterpolationAlgorithm getEdgeAlgorithm() {
```
wgtmac, line 1343:
```suggestion
    public EdgeInterpolationAlgorithm getAlgorithm() {
```

parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -1183,6 +1234,136 @@ public boolean equals(Object obj) {
...
+      sb.append("(");
+      if (hasCrs) {
+        sb.append(crs);
+      }
+      if (hasEdgeAlgorithm) {
+        if (hasCrs) sb.append(",");
+        sb.append(edgeAlgorithm);
+      }
+      sb.append(")");
+      return sb.toString();
```
wgtmac, line 1336:
```suggestion
      boolean hasCrs = crs != null && !crs.isEmpty();
      boolean hasAlgo = algorithm != null;
      if (!hasCrs && !hasAlgo) {
        return "";
      }
      return String.format("(%s,%s)", hasCrs ? crs : DEFAULT_CRS, hasAlgo ? algorithm : DEFAULT_ALGO);
```

Let's be more readable. This requires us to define following variables somewhere at the top:

```
  public static final String DEFAULT_CRS = "OGC:CRS84";
  public static final EdgeInterpolationAlgorithm DEFAULT_ALGO = EdgeInterpolationAlgorithm.SPHERICAL;
```

wgtmac (changes requested):
Thanks for splitting the large PR! I've left some comments.

cc @rdblue @gszadovszky

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,35 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
     public Optional<LogicalType> visit(LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
       return of(LogicalTypes.VARIANT(variantLogicalType.getSpecVersion()));
     }
+
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.GeometryLogicalTypeAnnotation geometryLogicalType) {
```
wgtmac, line 528: Could you please add some test cases in TestParquetMetadataConverter.java to verify that these geospatial types are well preserved from/to thrift message, especially when some type parameters are missing?

wgtmac (commented):
Thanks for the update! It overall looks good! I have left a comment for adding some test cases.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
zhangfengcdt: @wgtmac Thanks for the review comments and I have added some tests to validate the handling of Geometry and Geography logical types in Parquet:
- testGeometryLogicalType - Verifies that Geometry types with CRS convert correctly through schema serialization/deserialization
- testGeographyLogicalType - Ensures Geography types with CRS and interpolation algorithm are preserved during schema conversion
- testGeometryLogicalTypeWithMissingCrs - Confirms that Geometry types without specified CRS are handled properly (resulting in null CRS)
- testGeographyLogicalTypeWithMissingParameters - Checks that Geography types with missing CRS and algorithm are correctly processed with null values
- testGeographyLogicalTypeWithAlgorithmButNoCrs - Validates that Geography types with algorithm but no CRS are handled properly, preserving the algorithm while CRS remains null

Please let me know if you have other cases that you would like you add.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,35 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+      GeometryType geometryType = new GeometryType();
+      if (geometryLogicalType.getCrs() != null) {
```
wgtmac, line 530:
```suggestion
      if (geometryLogicalType.getCrs() != null && !geometryLogicalType.getCrs().isEmpty()) {
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,35 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+        geometryType.setCrs(geometryLogicalType.getCrs());
+      }
+      return of(LogicalType.GEOMETRY(geometryType));
+    }
+
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.GeographyLogicalTypeAnnotation geographyLogicalType) {
+      GeographyType geographyType = new GeographyType();
+      if (geographyLogicalType.getCrs() != null) {
```
wgtmac, line 539:
```suggestion
      if (geographyLogicalType.getCrs() != null && !geographyLogicalType.getCrs().isEmpty()) {
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,35 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+        geographyType.setCrs(geographyLogicalType.getCrs());
+      }
+      if (geographyLogicalType.getAlgorithm() != null) {
+        try {
+          // Convert from schema.EdgeInterpolationAlgorithm to format.EdgeInterpolationAlgorithm
+          EdgeInterpolationAlgorithm algorithm = EdgeInterpolationAlgorithm.valueOf(
+              geographyLogicalType.getAlgorithm().name());
+          geographyType.setAlgorithm(algorithm);
+        } catch (IllegalArgumentException e) {
```
wgtmac, line 548: Why not simply removing the catch since you throw the same exception here?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,35 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.GeographyLogicalTypeAnnotation geographyLogicalType) {
+      GeographyType geographyType = new GeographyType();
+      if (geographyLogicalType.getCrs() != null) {
+        geographyType.setCrs(geographyLogicalType.getCrs());
+      }
+      if (geographyLogicalType.getAlgorithm() != null) {
+        try {
+          // Convert from schema.EdgeInterpolationAlgorithm to format.EdgeInterpolationAlgorithm
+          EdgeInterpolationAlgorithm algorithm = EdgeInterpolationAlgorithm.valueOf(
```
wgtmac, line 545: Can we add a direct conversion method between thrift and parquet EdgeInterpolationAlgorithm in this file? String conversion looks inefficient.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1183,6 +1215,24 @@ LogicalTypeAnnotation getLogicalTypeAnnotation(LogicalType type) {
...
         return LogicalTypeAnnotation.float16Type();
+      case GEOMETRY:
+        GeometryType geometry = type.getGEOMETRY();
+        return LogicalTypeAnnotation.geometryType(geometry.getCrs());
+      case GEOGRAPHY:
+        GeographyType geography = type.getGEOGRAPHY();
+        // Handle when either algorithm or CRS is null
+        if (geography == null) {
+          return LogicalTypeAnnotation.geographyType(null, null);
+        }
```
wgtmac, line 1226:
```suggestion
```

They looks strange here.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1183,6 +1215,24 @@ LogicalTypeAnnotation getLogicalTypeAnnotation(LogicalType type) {
...
+
+        EdgeInterpolationAlgorithm algorithm = geography.getAlgorithm();
+        org.apache.parquet.column.schema.EdgeInterpolationAlgorithm parquetAlgorithm = null;
+        if (algorithm != null) {
+          parquetAlgorithm =
+              org.apache.parquet.column.schema.EdgeInterpolationAlgorithm.valueOf(algorithm.name());
```
wgtmac, line 1232: Same as above. We'd better implement a direct conversion.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
wgtmac: Thanks for adding the test!

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -520,6 +523,32 @@ public Optional<LogicalType> visit(LogicalTypeAnnotation.IntervalLogicalTypeAnno
...
+    }
+
+    @Override
+    public Optional<LogicalType> visit(LogicalTypeAnnotation.GeographyLogicalTypeAnnotation geographyLogicalType) {
+      GeographyType geographyType = new GeographyType();
+      if (geographyLogicalType.getCrs() != null) {
+        geographyType.setCrs(geographyLogicalType.getCrs());
+      }
+      if (geographyLogicalType.getAlgorithm() != null
+          && !geographyLogicalType.getCrs().isEmpty()) {
```
wgtmac, line 544:
```suggestion
      if (geographyLogicalType.getAlgorithm() != null) {
```

wgtmac: I believe this is the wrong line to add the check

wgtmac (approved):
I've pushed some minor changes. Thanks for working on this!

rdblue:
Thanks, @zhangfengcdt! Great to see this go in

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -442,6 +445,21 @@ private void appendHex(byte[] array, int offset, int length, StringBuilder build
     }
   };
 
+  static final PrimitiveStringifier WKB_STRINGIFIER = new BinaryStringifierBase("WKB_STRINGIFIER") {
```
rdblue, line 448: @wgtmac and @zhangfengcdt, what do you think about updating this to use WKT only if JTS is available, rather than adding JTS to the classpath for all downstream clients? This could just produce a generic string if a geo library isn't available.

jiayuasu: I believe it’s reasonable to add JTS as a dependency. Its license is compatible with the ASF, and it’s the most widely used geospatial library in the industry. The API is also very stable. [JTS has been around for over 20 years](https://en.wikipedia.org/wiki/JTS_Topology_Suite).

An alternative would be to implement a standalone WKB parser and maintain it within `parquet-java`, but that could introduce significant long-term maintenance overhead for the community.

For context, the Parquet Geo C++ PR included a standalone WKB parser because (1) there isn’t a clean, well-maintained WKB parser in C++, and (2) Dewey had already implemented WKB parsing in C++ multiple times before.

PR #3202, opened 2025-04-25
GH-3201: Implement a Variant builder to create Variant values

### LINKED ISSUE #3201: Implement a Variant builder

Implement a builder which create Variant values according to the [Variant encoding spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md).

### Component(s)

_No response_

gene-db:
### Rationale for this change

We need a way to build/encode Variant values. This builder allows users to create Variant values according to the [Variant binary encoding spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md).

A builder class for building Variant values.

### Are these changes tested?

Yes, added unit tests.

A new public class to build Variant vlues.

Closes #3201

### DIFF
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
@@ -42,6 +42,7 @@ public class VariantArrayBuilder extends VariantBuilder {
       throw new IllegalStateException(String.format(
           "Number of offsets (%d) do not match the number of values (%d).", offsets.size(), numValues));
     }
+    checkMultipleNested("Cannot call endArray() while a nested object/array is still open.");
     return offsets;
   }
 
@@ -49,16 +50,13 @@ public class VariantArrayBuilder extends VariantBuilder {
   protected void onAppend() {
     checkAppendWhileNested();
     offsets.add(writePos);
+    numValues++;
   }
 
   @Override
   protected void onStartNested() {
-    checkMultipleNested();
+    checkMultipleNested("Cannot call startObject()/startArray() without calling endObject()/endArray() first.");
     offsets.add(writePos);
-  }
-
-  @Override
-  protected void incrementNumValues() {
     numValues++;
   }
 
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
@@ -22,6 +22,13 @@ import java.util.ArrayList;
  * Builder for creating Variant arrays, used by VariantBuilder.
  */
 public class VariantArrayBuilder {
+  /** The underlying VariantBuilder. */
+  private final VariantBuilder builder;
+  /** The saved builder.writPos() for the start of this array. */
+  private final int startPos;
+  /** The offsets of the elements in this array. */
+  private final ArrayList<Integer> offsets;
+
   VariantArrayBuilder(VariantBuilder builder) {
     this.builder = builder;
     this.startPos = builder.writePos();
@@ -39,11 +46,4 @@ public class VariantArrayBuilder {
   ArrayList<Integer> offsets() {
     return offsets;
   }
-
-  /** The underlying VariantBuilder. */
-  private final VariantBuilder builder;
-  /** The saved builder.writPos() for the start of this array. */
-  private final int startPos;
-  /** The offsets of the elements in this array. */
-  private final ArrayList<Integer> offsets;
 }
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
@@ -21,29 +21,43 @@ import java.util.ArrayList;
 /**
  * Builder for creating Variant arrays, used by VariantBuilder.
  */
-public class VariantArrayBuilder {
-  /** The underlying VariantBuilder. */
-  private final VariantBuilder builder;
-  /** The saved builder.writPos() for the start of this array. */
-  private final int startPos;
+public class VariantArrayBuilder extends VariantBuilder {
+  /** The parent VariantBuilder. */
+  private final VariantBuilder parent;
   /** The offsets of the elements in this array. */
   private final ArrayList<Integer> offsets;
 
-  VariantArrayBuilder(VariantBuilder builder) {
-    this.builder = builder;
-    this.startPos = builder.writePos();
+  VariantArrayBuilder(VariantBuilder parent) {
+    this.parent = parent;
     this.offsets = new ArrayList<>();
   }
 
-  void startElement() {
-    offsets.add(builder.writePos() - startPos);
-  }
-
-  int startPos() {
-    return startPos;
-  }
-
-  ArrayList<Integer> offsets() {
+  /**
+   * @return the list of element offsets in this array
+   */
+  ArrayList<Integer> validateAndGetOffsets() {
+    if (offsets.size() != numValues) {
+      throw new IllegalStateException(String.format(
+          "Number of offsets (%d) do not match the number of values (%d).", offsets.size(), numValues));
+    }
     return offsets;
+  }
+
+  @Override
+  protected void onAppend() {
+    checkAppendWhileNested();
+    offsets.add(writePos);
+  }
+
+  @Override
+  protected void onStartNested() {
+    checkMultipleNested();
+    offsets.add(writePos);
+  }
+
+  @Override
+  int addDictionaryKey(String key) {
+    // Add to the parent dictionary.
+    return parent.addDictionaryKey(key);
   }
 }
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
@@ -26,6 +26,8 @@ public class VariantArrayBuilder extends VariantBuilder {
   private final VariantBuilder parent;
   /** The offsets of the elements in this array. */
   private final ArrayList<Integer> offsets;
+  /** The number of values appended to this array. */
+  protected long numValues = 0;
 
   VariantArrayBuilder(VariantBuilder parent) {
     this.parent = parent;
@@ -53,6 +55,11 @@ public class VariantArrayBuilder extends VariantBuilder {
   protected void onStartNested() {
     checkMultipleNested();
     offsets.add(writePos);
+  }
+
+  @Override
+  protected void incrementNumValues() {
+    numValues++;
   }
 
   @Override
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.parquet.variant;
+
+import java.util.ArrayList;
+
+/**
+ * Builder for creating Variant arrays, used by VariantBuilder.
+ */
+public class VariantArrayBuilder {
+  VariantArrayBuilder(VariantBuilder builder) {
+    this.builder = builder;
+    this.startPos = builder.writePos();
+    this.offsets = new ArrayList<>();
+  }
+
+  void startElement() {
+    offsets.add(builder.writePos() - startPos);
+  }
+
+  int startPos() {
+    return startPos;
+  }
+
+  ArrayList<Integer> offsets() {
+    return offsets;
+  }
+
+  /** The underlying VariantBuilder. */
+  private final VariantBuilder builder;
+  /** The saved builder.writPos() for the start of this array. */
+  private final int startPos;
+  /** The offsets of the elements in this array. */
+  private final ArrayList<Integer> offsets;
+}
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
@@ -24,26 +24,39 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 
 /**
  * Builder for creating Variant value and metadata.
  */
 public class VariantBuilder {
+  /** The buffer for building the Variant value. The first `writePos` bytes have been written. */
+  protected byte[] writeBuffer = new byte[1024];
+
+  protected int writePos = 0;
+  /** The dictionary for mapping keys to monotonically increasing ids. */
+  private final HashMap<String, Integer> dictionary = new HashMap<>();
+  /** The keys in the dictionary, in id order. */
+  private final ArrayList<byte[]> dictionaryKeys = new ArrayList<>();
+
+  /** The number of values appended to this builder. Must be updated after each append(). */
+  protected long numValues = 0;
+
+  protected VariantObjectBuilder objectBuilder = null;
+
   /**
    * Creates a VariantBuilder.
-   * @param allowDuplicateObjectKeys if true, only the last occurrence of a duplicate object key
-   *                                 will be kept. Otherwise, an exception will be thrown.
    */
-  public VariantBuilder(boolean allowDuplicateObjectKeys) {
-    this.allowDuplicateObjectKeys = allowDuplicateObjectKeys;
-  }
+  public VariantBuilder() {}
 
   /**
    * @return the Variant value
    */
   public Variant build() {
+    if (objectBuilder != null) {
+      throw new IllegalStateException(
+          "Cannot call build() while an object is being built. Must call endObject() first.");
+    }
     int numKeys = dictionaryKeys.size();
     // Use long to avoid overflow in accumulating lengths.
     long dictionaryTotalDataSize = 0;
@@ -85,11 +98,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendString(String str) {
+    checkAppendState();
     byte[] data = str.getBytes(StandardCharsets.UTF_8);
     boolean longStr = data.length > VariantUtil.MAX_SHORT_STR_SIZE;
     checkCapacity((longStr ? 1 + VariantUtil.U32_SIZE : 1) + data.length);
     if (longStr) {
-      writeBuffer[writePos] = VariantUtil.HDR_LONG_STRING;
+      writeBuffer[writePos] = VariantUtil.HEADER_LONG_STRING;
       writePos += 1;
       VariantUtil.writeLong(writeBuffer, writePos, data.length, VariantUtil.U32_SIZE);
       writePos += VariantUtil.U32_SIZE;
@@ -99,6 +113,7 @@ public class VariantBuilder {
     }
     System.arraycopy(data, 0, writeBuffer, writePos, data.length);
     writePos += data.length;
+    numValues++;
     return this;
   }
 
@@ -107,9 +122,11 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendNull() {
+    checkAppendState();
     checkCapacity(1);
-    writeBuffer[writePos] = VariantUtil.HDR_NULL;
+    writeBuffer[writePos] = VariantUtil.HEADER_NULL;
     writePos += 1;
+    numValues++;
     return this;
   }
 
@@ -119,9 +136,11 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendBoolean(boolean b) {
+    checkAppendState();
     checkCapacity(1);
-    writeBuffer[writePos] = b ? VariantUtil.HDR_TRUE : VariantUtil.HDR_FALSE;
+    writeBuffer[writePos] = b ? VariantUtil.HEADER_TRUE : VariantUtil.HEADER_FALSE;
     writePos += 1;
+    numValues++;
     return this;
   }
 
@@ -131,10 +150,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendLong(long l) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 8);
-    writeBuffer[writePos] = VariantUtil.HDR_INT64;
+    writeBuffer[writePos] = VariantUtil.HEADER_INT64;
     VariantUtil.writeLong(writeBuffer, writePos + 1, l, 8);
     writePos += 9;
+    numValues++;
     return this;
   }
 
@@ -144,10 +165,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendInt(int i) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 4);
-    writeBuffer[writePos] = VariantUtil.HDR_INT32;
+    writeBuffer[writePos] = VariantUtil.HEADER_INT32;
     VariantUtil.writeLong(writeBuffer, writePos + 1, i, 4);
     writePos += 5;
+    numValues++;
     return this;
   }
 
@@ -157,10 +180,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendShort(short s) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 2);
-    writeBuffer[writePos] = VariantUtil.HDR_INT16;
+    writeBuffer[writePos] = VariantUtil.HEADER_INT16;
     VariantUtil.writeLong(writeBuffer, writePos + 1, s, 2);
     writePos += 3;
+    numValues++;
     return this;
   }
 
@@ -170,10 +195,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendByte(byte b) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 1);
-    writeBuffer[writePos] = VariantUtil.HDR_INT8;
+    writeBuffer[writePos] = VariantUtil.HEADER_INT8;
     VariantUtil.writeLong(writeBuffer, writePos + 1, b, 1);
     writePos += 2;
+    numValues++;
     return this;
   }
 
@@ -183,10 +210,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendDouble(double d) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 8);
-    writeBuffer[writePos] = VariantUtil.HDR_DOUBLE;
+    writeBuffer[writePos] = VariantUtil.HEADER_DOUBLE;
     VariantUtil.writeLong(writeBuffer, writePos + 1, Double.doubleToLongBits(d), 8);
     writePos += 9;
+    numValues++;
     return this;
   }
 
@@ -197,17 +226,18 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendDecimal(BigDecimal d) {
+    checkAppendState();
     BigInteger unscaled = d.unscaledValue();
     if (d.scale() <= VariantUtil.MAX_DECIMAL4_PRECISION && d.precision() <= VariantUtil.MAX_DECIMAL4_PRECISION) {
       checkCapacity(2 /* header and scale size */ + 4);
-      writeBuffer[writePos] = VariantUtil.HDR_DECIMAL4;
+      writeBuffer[writePos] = VariantUtil.HEADER_DECIMAL4;
       writeBuffer[writePos + 1] = (byte) d.scale();
       VariantUtil.writeLong(writeBuffer, writePos + 2, unscaled.intValueExact(), 4);
       writePos += 6;
     } else if (d.scale() <= VariantUtil.MAX_DECIMAL8_PRECISION
         && d.precision() <= VariantUtil.MAX_DECIMAL8_PRECISION) {
       checkCapacity(2 /* header and scale size */ + 8);
-      writeBuffer[writePos] = VariantUtil.HDR_DECIMAL8;
+      writeBuffer[writePos] = VariantUtil.HEADER_DECIMAL8;
       writeBuffer[writePos + 1] = (byte) d.scale();
       VariantUtil.writeLong(writeBuffer, writePos + 2, unscaled.longValueExact(), 8);
       writePos += 10;
@@ -215,7 +245,7 @@ public class VariantBuilder {
       assert d.scale() <= VariantUtil.MAX_DECIMAL16_PRECISION
           && d.precision() <= VariantUtil.MAX_DECIMAL16_PRECISION;
       checkCapacity(2 /* header and scale size */ + 16);
-      writeBuffer[writePos] = VariantUtil.HDR_DECIMAL16;
+      writeBuffer[writePos] = VariantUtil.HEADER_DECIMAL16;
       writeBuffer[writePos + 1] = (byte) d.scale();
       writePos += 2;
       // `toByteArray` returns a big-endian representation. We need to copy it reversely and sign
@@ -230,6 +260,7 @@ public class VariantBuilder {
       }
       writePos += 16;
     }
+    numValues++;
     return this;
   }
 
@@ -240,10 +271,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendDate(int daysSinceEpoch) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 4);
-    writeBuffer[writePos] = VariantUtil.HDR_DATE;
+    writeBuffer[writePos] = VariantUtil.HEADER_DATE;
     VariantUtil.writeLong(writeBuffer, writePos + 1, daysSinceEpoch, 4);
     writePos += 5;
+    numValues++;
     return this;
   }
 
@@ -254,10 +287,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendTimestampTz(long microsSinceEpoch) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 8);
-    writeBuffer[writePos] = VariantUtil.HDR_TIMESTAMP_TZ;
+    writeBuffer[writePos] = VariantUtil.HEADER_TIMESTAMP_TZ;
     VariantUtil.writeLong(writeBuffer, writePos + 1, microsSinceEpoch, 8);
     writePos += 9;
+    numValues++;
     return this;
   }
 
@@ -268,10 +303,12 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendTimestampNtz(long microsSinceEpoch) {
+    checkAppendState();
     checkCapacity(1 /* header size */ + 8);
-    writeBuffer[writePos] = VariantUtil.HDR_TIMESTAMP_NTZ;
+    writeBuffer[writePos] = VariantUtil.HEADER_TIMESTAMP_NTZ;
     VariantUtil.writeLong(writeBuffer, writePos + 1, microsSinceEpoch, 8);
     writePos += 9;
+    numValues++;
     return this;
   }
 
@@ -282,10 +319,16 @@ public class VariantBuilder {
    * @return this builder
    */
   public VariantBuilder appendTime(long microsSinceMidnight) {
+    if (microsSinceMidnight < 0) {
+      throw new IllegalArgumentException(
+          String.format("Time value (%d) cannot be negative.", microsSinceMidnight));
+    }
+    checkAppendState();
     checkCapacity(1 /* header size */ + 8);
-    writeBuffer[writePos] = VariantUtil.HDR_TIME;
+    writeBuffer[writePos] = VariantUtil.HEADER_TIME;
     VariantUtil.writeLong(writeBuffer, writePos + 1, microsSinceMidnight, 8);
     writePos += 9;
+    numValues++;
     return this;
   }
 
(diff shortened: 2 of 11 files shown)


parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+
+/**
+ * Builder for creating Variant value and metadata.
+ */
+public class VariantBuilder {
+  /**
+   * Creates a VariantBuilder.
+   * @param allowDuplicateKeys if true, only the last occurrence of a duplicate key will be kept.
```
rdblue, line 36: Duplicate object keys? This could be more clear.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+    this.allowDuplicateKeys = allowDuplicateKeys;
+  }
+
+  /**
+   * @return the Variant value
+   */
+  public Variant build() {
+    int numKeys = dictionaryKeys.size();
+    // Use long to avoid overflow in accumulating lengths.
+    long dictionaryStringSize = 0;
```
rdblue, line 49: The variable name here doesn't quite make sense to me. What about `totalDataSize` or `totalContentSize`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+      dictionaryStringSize += key.length;
+    }
+    // Determine the number of bytes required per offset entry.
+    // The largest offset is the one-past-the-end value, which is total string size. It's very
+    // unlikely that the number of keys could be larger, but incorporate that into the calculation
+    // in case of pathological data.
+    long maxSize = Math.max(dictionaryStringSize, numKeys);
+    int offsetSize = getMinIntegerSize((int) maxSize);
+
+    int offsetStart = 1 + offsetSize;
```
rdblue, line 60: I think this could also be more clear as `offsetListOffset`. This is the header plus the dictionary size.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+    int stringStart = offsetStart + (numKeys + 1) * offsetSize;
+    long metadataSize = stringStart + dictionaryStringSize;
+
+    byte[] metadata = new byte[(int) metadataSize];
+    int headerByte = VariantUtil.VERSION | ((offsetSize - 1) << 6);
```
rdblue, line 65: Always unsorted?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+    VariantUtil.writeLong(metadata, 0, headerByte, 1);
+    VariantUtil.writeLong(metadata, 1, numKeys, offsetSize);
+    int currentOffset = 0;
+    for (int i = 0; i < numKeys; ++i) {
```
rdblue, line 69: Loop looks correct.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+      VariantUtil.writeLong(metadata, offsetStart + i * offsetSize, currentOffset, offsetSize);
+      byte[] key = dictionaryKeys.get(i);
+      System.arraycopy(key, 0, metadata, stringStart + currentOffset, key.length);
+      currentOffset += key.length;
+    }
+    VariantUtil.writeLong(metadata, offsetStart + numKeys * offsetSize, currentOffset, offsetSize);
+    return new Variant(Arrays.copyOfRange(writeBuffer, 0, writePos), metadata);
```
rdblue, line 76: Why copy here? If the buffer may be reused, then a comment would help clarify the contract.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+  }
+
+  /**
+   * Appends a string value to the Variant builder.
+   * @param str the string value to append
+   * @return this builder
+   */
+  public VariantBuilder appendString(String str) {
+    byte[] text = str.getBytes(StandardCharsets.UTF_8);
```
rdblue, line 85: `text` seems like an odd name, since this is a string encoded as UTF-8 bytes.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+    boolean longStr = text.length > VariantUtil.MAX_SHORT_STR_SIZE;
+    checkCapacity((longStr ? 1 + VariantUtil.U32_SIZE : 1) + text.length);
+    if (longStr) {
+      writeBuffer[writePos++] = VariantUtil.primitiveHeader(VariantUtil.LONG_STR);
```
rdblue, line 89: Could this header be a constant?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+      VariantUtil.writeLong(writeBuffer, writePos, text.length, VariantUtil.U32_SIZE);
+      writePos += VariantUtil.U32_SIZE;
+    } else {
+      writeBuffer[writePos++] = VariantUtil.shortStrHeader(text.length);
```
rdblue, line 93: While it is more succinct, I think it is a bad practice to use the return value of the `++` operator because it forces the reader to remember whether the original value or incremented value is returned along with the context in which that value is used. It is generally more maintainable to use a separate line:

```java
writeBuffer[writePos] = VariantUtil.shortStrHeader(text.length);
writePos += 1;
```

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+
+  /**
+   * Appends a long value to the variant builder. The actual encoded integer type depends on the
+   * value range of the long value.
+   * @param l the long value to append
+   * @return this builder
+   */
+  public VariantBuilder appendLong(long l) {
+    if (l == (byte) l) {
+      checkCapacity(1 + 1);
```
rdblue, line 129: Without an inline comment for context, this is a bit odd. I think it would make sense to use either `1 /* header size */ + 1` or just switch to constants. That is, without a comment, I don't think this is more clear than `checkCapacity(2)`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+      writeBuffer[writePos++] = VariantUtil.primitiveHeader(VariantUtil.INT8);
+      VariantUtil.writeLong(writeBuffer, writePos, l, 1);
+      writePos += 1;
+    } else if (l == (short) l) {
+      checkCapacity(1 + 2);
+      writeBuffer[writePos++] = VariantUtil.primitiveHeader(VariantUtil.INT16);
+      VariantUtil.writeLong(writeBuffer, writePos, l, 2);
```
rdblue, line 136: Wouldn't it be better to update `writePos` just once instead? This could write to `writePos + 1` and then increment by 3 on the next line.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: The original reason to copy out was to only keep the required length and not the full capacity.

What do you think? Should we just keep the buffer in the builder, and point to it with a `ByteBuffer`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,590 @@
...
+    return this;
+  }
+
+  /**
+   * Appends a long value to the variant builder. The actual encoded integer type depends on the
+   * value range of the long value.
+   * @param l the long value to append
+   * @return this builder
+   */
+  public VariantBuilder appendLong(long l) {
```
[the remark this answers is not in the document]
gene-db: Updated this and introduced `appendInt`, `appendShort`, `appendByte`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+
+  int startPos() {
+    return startPos;
+  }
+
+  ArrayList<Integer> offsets() {
+    return offsets;
+  }
+
+  /** The underlying VariantBuilder. */
```
rdblue, line 43: It's more common to list instance fields at the top of a class.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,621 @@
...
+   * values have been added to the object.
+   * @param objBuilder the VariantObjectBuilder to use
+   * @return this builder
+   */
+  public VariantBuilder endObject(VariantObjectBuilder objBuilder) {
+    ArrayList<FieldEntry> fields = objBuilder.fields();
+    int size = fields.size();
+    Collections.sort(fields);
+    int maxId = size == 0 ? 0 : fields.get(0).id;
+    if (allowDuplicateObjectKeys) {
```
rdblue, line 406: I find `allowDuplicateObjectKeys` confusing now that I see how it is used. This doesn't actually allow keys to have multiple values like a multi-map, it just enables a mode where the builder will replace existing keys with the last value that is set.

I think that you could simplify this by always allowing duplicate object keys. There isn't much of a benefit gained by now allowing duplicate keys because duplicates cause a failure although they could be handled. And the code path to deduplicate the keys is nearly identical in the case where there are no duplicates: the first branch comparing the ID is never taken so the block that copies values to the right offset will not run.

That means that the extra cost when there are no duplicate keys is the `fields.set` call that doesn't move the value. And the API would simply handle duplicate keys instead of failing with an exception.

I think this should remove `allowDuplicateObjectKeys` and always handle duplicates.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,621 @@
...
+   * be used for future calls to addObjectKey() and endObject(). To add a (key, value) to the
+   * Variant object, call appendObjectKey() to add the key name, and then append the value.
+   * endObject() must be called to finish writing the Variant object.
+   *
+   * Example usage:
+   * VariantBuilder builder = new VariantBuilder();
+   * VariantObjectBuilder objBuilder = builder.startObject();
+   * builder.appendObjectKey(objBuilder, "key1");
+   * builder.appendString("value1");
+   * builder.endObject(objBuilder);
```
rdblue, line 378: This API is confusing until you know what is happening in `endObject` because the values are written, but then the data section is relocated to make room for the header (and sometimes values are copied to deduplicate keys). I would expect this to use the object builder for both key and value.

It looks like the purpose of this structure is to avoid buffer allocation, so that values are always written to the `writeBuffer` directly and moved when needed, instead of staging encoded values in buffers and copying between them. These choices lead to an odd API, as well as a few passes of sorting fields and copying values around. The implementation in `endObject` is hard to understand and will be hard to maintain.

There are a couple of solutions that would help.

One solution is to have a separate value buffer in each `VariantObjectBuilder` and to call `appendString` or `appendDate` etc. on the object builder -- it would implement the full builder API and write field values directly to its buffer. That would have a benefit of being able to get the size of the last value written when a new field is started, and would allow better validation that the number of keys matches the number of fields. When the object is ended, the `endObject` method would simply sort the keys and copy each value between buffers. Then the object builder (and its buffer) could be reused the next time `startObject` is called by its parent builder.

Another option is to keep keys and values in an in-memory map. This would have an API similar to the first option. When a value is added, you'd add it to the map using the last field key. This would require wrapping the value to keep track of its type (like Iceberg's [`PrimitiveWrapper`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/variants/PrimitiveWrapper.java)). Serialization would happen when `endObject` is called.

I think both of these would be better options than the current complexity of `endObject`, and would also allow the builder API to provide better checks that it is being used correctly because the `VariantObjectBuilder` would be used for both keys and values so the parent builder would be able to fail if any `append` method is called before `endObject`. This would also avoid the need to pass the object builder back into the variant builder API.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,621 @@
...
+    arrayBuilder.startElement();
+  }
+
+  /**
+   * Ends appending an array to this variant builder. This method must be called after all elements
+   * have been added to the array.
+   * @param arrayBuilder the VariantArrayBuilder to use
+   * @return this builder
+   */
+  public VariantBuilder endArray(VariantArrayBuilder arrayBuilder) {
```
rdblue, line 513: I think that once the object builder is updated, the same design should be used to update the array builder.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,621 @@
...
+      return VariantUtil.U16_SIZE;
+    }
+    if (value <= VariantUtil.U24_MAX) {
+      return VariantUtil.U24_SIZE;
+    }
+    return VariantUtil.U32_SIZE;
+  }
+
+  /** The buffer for building the Variant value. The first `writePos` bytes have been written. */
+  private byte[] writeBuffer = new byte[128];
```
rdblue, line 612: If this is trying to avoid allocations, then I think this should start at a larger size. It would make sense to me to start a 1k or 2k, considering this is probably reused and we don't want to reallocate very often.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,621 @@
...
+
+  private int writePos = 0;
```
rdblue, line 614: As with the other builders, this would be better at the top.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -187,6 +187,29 @@ class VariantUtil {
   // The size (in bytes) of a UUID.
   static final int UUID_SIZE = 16;
 
+  // header bytes
+  static final byte HDR_NULL = primitiveHeader(NULL);
```
rdblue, line 191: Nit: Omitting vowels doesn't help much. The code is less readable and it's not actually easier to type hdr than it is to type header.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -329,6 +366,75 @@ static Variant.Type getType(ByteBuffer value) {
     }
   }
 
+  /**
+   * Computes the actual size (in bytes) of the Variant value at `value[pos...]`
+   * @param value The Variant value
+   * @param pos The starting index of the Variant value
+   * @return The size (in bytes) of the Variant value
+   */
+  public static int valueSize(ByteBuffer value, int pos) {
```
rdblue, line 375: I think this can be avoided by a different serialization strategy for arrays and objects because you could use the actual bytes written for the length of elements or field values.

parquet-variant/src/test/java/org/apache/parquet/variant/VariantTestUtil.java
```diff
@@ -0,0 +1,116 @@
...
+  private static SecureRandom random = new SecureRandom(new byte[] {1, 2, 3, 4, 5});
+
+  static final ByteBuffer EMPTY_METADATA = ByteBuffer.wrap(new byte[] {0b1});
+
+  static void checkType(Variant v, int expectedBasicType, Variant.Type expectedType) {
+    Assert.assertEquals(expectedBasicType, v.value.get(v.value.position()) & VariantUtil.BASIC_TYPE_MASK);
+    Assert.assertEquals(expectedType, v.getType());
+  }
+
+  static long microsSinceEpoch(Instant instant) {
```
rdblue, line 47: Why is this interacting with `Instant`? There is no need to have a conversion here because the write methods are called with `long` values not `Instant`. Tests only need to validate that the value passed in matches the value read back and that the type is preserved. Conversion from a date/time representation is the responsibility of engines, not Parquet.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
```diff
@@ -250,22 +212,61 @@ public void testMixedObject() {
...
       Assert.assertEquals(
           LocalDate.parse("2025-04-17"),
           LocalDate.ofEpochDay(nestedV.getFieldByKey("a").getInt()));
-      checkType(nestedV.getFieldByKey("c"), VariantUtil.PRIMITIVE, Variant.Type.NULL);
+      VariantTestUtil.checkType(nestedV.getFieldByKey("c"), VariantUtil.PRIMITIVE, Variant.Type.NULL);
+    });
+  }
+
+  @Test
+  public void testMixedObjectBuilder() {
```
rdblue, line 235: I would prefer to have builder tests in a `TestVariantBuilder` or `TestVariantObjectBuilder` suite to keep them separate from the decoding tests. These tests validate the `Variant` decoding and that is used to validate the encoding/builder.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
```diff
@@ -250,22 +212,61 @@ public void testMixedObject() {
...
+      b.appendString("variant");
+      b.endObject(nestedBuilder);
+    }
+    b.endObject(objBuilder);
+
+    VariantTestUtil.testVariant(b.build(), v -> {
+      VariantTestUtil.checkType(v, VariantUtil.OBJECT, Variant.Type.OBJECT);
+      Assert.assertEquals(3, v.numObjectElements());
+      Assert.assertTrue(v.getFieldByKey("outer 1").getBoolean());
+      Assert.assertEquals(1234567890, v.getFieldByKey("outer 2").getLong());
```
rdblue, line 262: I think this assertion needs to verify that the field was encoded as a long because `appendLong` was used. Can you add assertions for the types as well?

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalar.java
```diff
@@ -137,136 +118,248 @@ public void testNegativeLong() {
...
+          vb2.appendShort(s);
+          VariantTestUtil.testVariant(vb2.build(), v -> {
+            VariantTestUtil.checkType(v, VariantUtil.PRIMITIVE, Variant.Type.SHORT);
+            Assert.assertEquals((short) s, v.getShort());
+          });
+        });
+  }
+
+  @Test
+  public void testByteBuilder() {
```
rdblue, line 251: Good testing here!

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalar.java
```diff
@@ -137,136 +118,248 @@ public void testNegativeLong() {
...
+        VariantTestUtil.EMPTY_METADATA);
+    VariantTestUtil.testVariant(value, v -> {
+      VariantTestUtil.checkType(v, VariantUtil.PRIMITIVE, Variant.Type.FLOAT);
       Assert.assertEquals(-0.0F, v.getFloat(), 0);
     });
   }
 
+  @Test
+  public void testFloatBuilder() {
+    Arrays.asList(Float.MIN_VALUE, 0f, Float.MAX_VALUE).forEach(f -> {
```
rdblue, line 286: -0f?

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalar.java
```diff
@@ -415,23 +574,43 @@ public void testNegativeTimestamp() {
...
       Assert.assertEquals(
           Instant.parse("1969-12-31T23:59:59.999999Z"), Instant.EPOCH.plus(v.getLong(), ChronoUnit.MICROS));
     });
   }
 
+  @Test
+  public void testTimestampTzBuilder() {
+    DateTimeFormatter dtf = DateTimeFormatter.ISO_DATE_TIME;
+    VariantBuilder vb = new VariantBuilder(false);
+    long micros = VariantTestUtil.microsSinceEpoch(Instant.from(dtf.parse("2024-12-16T10:23:45.321456-08:00")));
```
rdblue, line 589: It doesn't look like converting the parsed date brings much value so I'd probably remove the conversion. It isn't worth the maintenance.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalar.java
```diff
@@ -451,52 +630,97 @@ public void testNegativeTimestampNtz() {
...
+      VariantTestUtil.checkType(v, VariantUtil.PRIMITIVE, Variant.Type.BINARY);
       Assert.assertEquals(ByteBuffer.wrap(new byte[] {'a', 'b', 'c', 'd', 'e'}), v.getBinary());
     });
   }
 
+  @Test
+  public void testBinaryBuilder() {
+    VariantBuilder vb = new VariantBuilder(false);
+    byte[] binary = new byte[] {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+    vb.appendBinary(binary);
```
rdblue, line 670: Shouldn't this append the ByteBuffer instead? Why force the caller to copy into a byte array starting at position 0?

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalar.java
```diff
@@ -506,32 +730,54 @@ public void testTime() {
...
   }
 
   @Test
-  public void testTimestampNanos() {
+  public void testTimeBuilder() {
+    for (String timeStr : Arrays.asList(
+        "00:00:00.000000", "00:00:00.000120", "12:00:00.000000", "12:00:00.002300", "23:59:59.999999")) {
+      VariantBuilder vb = new VariantBuilder(false);
+      long micros = LocalTime.parse(timeStr).toNanoOfDay() / 1_000;
+      vb.appendTime(micros);
```
rdblue, line 746: Should this allow or reject negative values?

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArray.java
```diff
@@ -20,104 +20,32 @@
...
-  private static final byte[] VALUE_NULL = new byte[] {primitiveHeader(0)};
-  private static final byte[] VALUE_BOOL = new byte[] {primitiveHeader(1)};
-  private static final byte[] VALUE_INT = new byte[] {primitiveHeader(5), (byte) 0xD2, 0x02, (byte) 0x96, 0x49};
+  private static final byte[] VALUE_NULL = new byte[] {VariantTestUtil.primitiveHeader(0)};
+  private static final byte[] VALUE_BOOL = new byte[] {VariantTestUtil.primitiveHeader(1)};
+  private static final byte[] VALUE_INT =
+      new byte[] {VariantTestUtil.primitiveHeader(5), (byte) 0xD2, 0x02, (byte) 0x96, 0x49};
   private static final byte[] VALUE_STRING =
-      new byte[] {primitiveHeader(16), 0x07, 0x00, 0x00, 0x00, 'v', 'a', 'r', 'i', 'a', 'n', 't'};
+      new byte[] {VariantTestUtil.primitiveHeader(16), 0x07, 0x00, 0x00, 0x00, 'v', 'a', 'r', 'i', 'a', 'n', 't'};
```
rdblue, line 37: Overall, I think the tests are covering relevant encoding cases. I'd prefer encoding and decoding tests in separate suites to be able to maintain them better.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
```diff
@@ -205,22 +137,52 @@ private static ByteBuffer constructMetadata(Boolean isSorted, List<String> field
...
 
+  @Test
+  public void testLargeObjectBuilder() {
+    VariantBuilder b = new VariantBuilder(true);
+    VariantObjectBuilder o = b.startObject();
+    for (int i = 0; i < 1234; i++) {
+      b.appendObjectKey(o, "a" + i);
+      b.appendLong(i);
+    }
+    b.endObject(o);
```
aihuaxu, line 176: I also did the similar implementation before to have endObject() and endArray(), but the caller needs to be careful to call in certain sequence.

We probably can have something similar tohttps://docs.oracle.com/javaee/7/api/javax/json/JsonObjectBuilder.html and JsonArrayBuilder.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Updated the array builder api.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Updated the API so that the object builder writes to its own buffer, and will do the key/value validation.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.parquet.variant;
+
+import java.util.ArrayList;
+
+/**
+ * Builder for creating Variant arrays, used by VariantBuilder.
```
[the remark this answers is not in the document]
gene-db: We decided to do that separately.

gene-db (commented):
Thanks @rdblue @aihuaxu ! I updated the PR.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
rdblue: This is a note to myself that I think this is implemented correctly so I don't need to review it again.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+   */
+  public VariantObjectBuilder startObject() {
+    onStartNested();
+    if (objectBuilder != null) {
+      throw new IllegalStateException("Cannot call startObject() without calling endObject() first.");
+    }
+    if (arrayBuilder != null) {
+      throw new IllegalStateException("Cannot call startObject() without calling endArray() first.");
+    }
+    objectBuilder = new VariantObjectBuilder(this);
```
rdblue, line 418: When setting instance fields, it helps to add the prefix `this.` so the reader can easily see it isn't a local variable.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+  protected byte[] writeBuffer = new byte[1024];
+
+  protected int writePos = 0;
+  /** The dictionary for mapping keys to monotonically increasing ids. */
+  private final HashMap<String, Integer> dictionary = new HashMap<>();
+  /** The keys in the dictionary, in id order. */
+  private final ArrayList<byte[]> dictionaryKeys = new ArrayList<>();
+
+  /** The number of values appended to this builder. */
+  protected long numValues = 0;
```
rdblue, line 43: Why would this builder support more than one value?

Only object and array builders should support multiple values so I think this builder doesn't need to keep track of this. It can be incremented in the `append` methods.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+    }
+    VariantUtil.writeLong(writeBuffer, offsetStart + numElements * offsetSize, dataSize, offsetSize);
+    writePos += headerSize + dataSize;
+    numValues++;
+    arrayBuilder = null;
+  }
+
+  protected void onAppend() {
+    checkAppendWhileNested();
+    if (numValues > 0) {
```
rdblue, line 553: This could be checked using `writePos` instead, right?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+    int dataSize = arrayBuilder.writePos;
+    boolean largeSize = numElements > VariantUtil.U8_MAX;
+    int sizeBytes = largeSize ? VariantUtil.U32_SIZE : 1;
+    int offsetSize = getMinIntegerSize(dataSize);
+    // The space for header byte, object size, and offset list.
+    int headerSize = 1 + sizeBytes + (numElements + 1) * offsetSize;
+    checkCapacity(headerSize + dataSize);
+
+    // Copy all the element data to the write buffer.
+    System.arraycopy(arrayBuilder.writeBuffer, 0, writeBuffer, writePos + headerSize, dataSize);
```
rdblue, line 536: Minor: isn't the header just the first byte? I would change the name of `headerSize` to `dataOffset`.

rdblue: Yeah, this calls `arrayHeader` in the next statement, which produces a single byte.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,654 @@
...
+    Collections.sort(fields);
+    int maxId = numFields == 0 ? 0 : fields.get(0).id;
+    int dataSize = numFields == 0 ? 0 : fields.get(0).valueSize;
+
+    int distinctPos = 0;
+    // Maintain a list of distinct keys in-place.
+    for (int i = 1; i < numFields; ++i) {
+      maxId = Math.max(maxId, fields.get(i).id);
+      if (fields.get(i).id == fields.get(i - 1).id) {
+        // Found a duplicate key. Keep the field with the greater offset.
```
rdblue, line 441: I think this comment is missing "because it was written last"

rdblue (changes requested):
Mostly looks good, but there are some minor comments. Thanks, @gene-db!

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Added the `this.` prefix.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Added.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Yes, updated.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Renamed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
gene-db: Removed this.

gene-db (commented):
Thanks @rdblue ! I updated the PR.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Override
+  protected void onStartNested() {
+    checkMultipleNested();
+    offsets.add(writePos);
+  }
+
+  @Override
+  protected void incrementNumValues() {
```
rdblue, line 61: Why does this require a protected method rather than putting this in `onAppend`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantObjectBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+    this.fields = new ArrayList<>();
+  }
+
+  /**
+   * Appends an object key to this object. This method must be called before appending any value.
+   * @param key the key to append
+   */
+  void appendKey(String key) {
+    if (fields.size() > numValues) {
+      throw new IllegalStateException("Cannot call appendKey() before appending an object value.");
```
rdblue, line 43: I think this is technically "before appending a value for the previous key" right?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
rdblue: Right now, if the builder isn't going to be reused then there's no point in copying. It would be better to use a buffer since we already know the length of the variant value in the array (`writePos`).

If you wanted to reuse the builder then I think this would change a little. We want a reused builder to keep track of its underlying buffer so that it doesn't need to grow each time it creates a new value (assuming that we have somewhat uniform value sizes). In that case, I think you'd want to copy the value out of the reused buffer that is owned by the builder.

Builder reuse is something we can work on later.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,658 @@
...
+      writePos += 1;
+      VariantUtil.writeLong(writeBuffer, writePos, data.length, VariantUtil.U32_SIZE);
+      writePos += VariantUtil.U32_SIZE;
+    } else {
+      writeBuffer[writePos] = VariantUtil.shortStrHeader(data.length);
+      writePos += 1;
+    }
+    System.arraycopy(data, 0, writeBuffer, writePos, data.length);
+    writePos += data.length;
+    incrementNumValues();
```
rdblue, line 123: As noted above, if we can do this in `onAppend` then we probably should.

rdblue: After looking into `endObject` and `endArray`, I think the rationale for incrementing values at the end of each append is to make the consistency check in `endObject` catch cases where a nested object is partially written. So if you have a nested object, the parent won't have a matching number of keys and values until the inner `endObject` is called.

I think we need a stronger validation that doesn't rely on when `incrementNumValues` is called. First, I don't think that check would similarly work when an object in an array is unfinished. So the `validate` methods for both arrays and objects need to call `checkMultipleNested` to verify that there are no unfinished objects or arrays in the child builder.

gene-db (commented):
@rdblue Thanks for the reviews! I updated the PR.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArrayBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    try {
+      b.build();
+      Assert.fail("Expected Exception when calling build() without endArray()");
+    } catch (Exception e) {
+      // expected
+    }
+  }
+
+  @Test
+  public void testMissingStartArray() {
```
rdblue, line 191: For testing exception cases, you can use `TestUtils.assertThrows` in `parquet-common`. That's a more uniform assertion that also checks the exception message.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArrayBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+    }
+  }
+
+  @Test
+  public void testInvalidAppendDuringArray() {
+    VariantBuilder b = new VariantBuilder();
+    b.startArray();
+    try {
+      b.appendInt(1);
+      Assert.fail("Expected Exception when calling append() before endArray()");
```
rdblue, line 207: I don't think this exception message is helpful because the API is a bit confusing. Some similar builders (like Jackson) would change the state of the current builder, rather than forcing the caller to use a new builder that is returned. It's okay to make this choice about how the builders are related, but the error message needs to tell the caller what to do. In this case, the caller should be informed that `append` methods must be called on the builder returned by `startArray` and `endArray` is called on the original builder when those are complete.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArrayBuilder.java
```diff
@@ -0,0 +1,293 @@
...
+      VariantTestUtil.checkType(nested.getElementAtIndex(0), VariantUtil.ARRAY, Variant.Type.ARRAY);
+      Assert.assertEquals(0, nested.getElementAtIndex(0).numArrayElements());
+      VariantTestUtil.checkType(nested.getElementAtIndex(1), VariantUtil.SHORT_STR, Variant.Type.STRING);
+      Assert.assertEquals("variant", nested.getElementAtIndex(1).getString());
+      VariantTestUtil.checkType(nested.getElementAtIndex(2), VariantUtil.OBJECT, Variant.Type.OBJECT);
+      Assert.assertEquals(0, nested.getElementAtIndex(2).numObjectElements());
+    });
+  }
+
+  private void buildNested(int i, VariantArrayBuilder obj) {
```
rdblue, line 110: I find this way too hard to reason about for a test. Can you rewrite it to a for loop in `testNestedBuilder`?

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObjectBuilder.java
```diff
@@ -0,0 +1,450 @@
...
+    }
+  }
+
+  @Test
+  public void testMultipleAppendKey() {
+    VariantBuilder b = new VariantBuilder();
+    VariantObjectBuilder obj = b.startObject();
+    obj.appendKey("a");
+    try {
+      obj.appendKey("a");
```
rdblue, line 332: Can you use a different key to ensure that it isn't a duplicate key problem?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -0,0 +1,631 @@
...
+  protected void checkMultipleNested(String message) {
+    if (objectBuilder != null || arrayBuilder != null) {
+      throw new IllegalStateException(message);
+    }
+  }
+
+  protected void checkAppendWhileNested() {
+    if (objectBuilder != null) {
+      throw new IllegalStateException(
+          "Cannot call append() methods while an object is being built. Must call endObject() first.");
```
rdblue, line 552: I don't think this exception message is very helpful because the API is a bit confusing. Some similar builders (like Jackson) would change the state of the current builder, rather than forcing the caller to use a new builder that is returned. It's okay to make this choice about how the builders are related, but the error message needs to tell the caller what to do.

In this case, the caller should be informed that `append` methods must be called on the builder returned by `startArray` and `endArray` is called on the original builder when those are complete. If you don't know that you should be using a different builder, then "cannot call append methods while an object is being built" doesn't make sense. "Append object fields using appendKey and append methods on the builder returned by startObject and call endObject on this builder" Or something like that?

rdblue (approved):

rdblue:
There are a couple of minor comments, but I think overall this is correct and ready -- we can fix exception messages in follow ups and merge this to unblock shredding work.

PR #3204, opened 2025-04-25
 GH-3203: HadoopPositionOutputStream.close() to call FSDataOutputStream.flush()

### LINKED ISSUE #3203: HadoopPositionOutputStream calls FSDataOutputStream.hflush() in close()

HadoopPositionOutputStream calls FSDataOutputStream.hflush() in close()
 
#1309  added an `hflush()` when closing a stream. This is needless and somewhat troublesome
* hdfs: extra overhead of a blocking write to all datanodes participating in the write (it doesn't know that close() is about to follow, after all)
* abfs: overhead of another write
* s3a: prints a message warning user that the Syncable api is unsupported. (once per process...we don't want log spam)
If anyone has a deployment with the setting `fs.s3a.downgrade.syncable.exceptions=false` then s3a will actually fail. This is the "fail fast if anyone tries to run algorithms expecting Syncable to persist data". it's a great way to find code which does exactly this.

And it won't be the cause of the problem. A simple flush() should suffice, if that is the underlying problem anyway.

### Component(s)

parquet-hadoop

steveloughran:
Downgrade flush in HadoopPositionOutputStream.close() from Syncable.hflush() to a basic flush() operation.

### Rationale for this change

calling Syncable.sync() is time-consuming overkill just before a close() operation, and trouble on other stores (s3a will log at warn, when not configured to fail fast)

Switches to classic OutputStream.flush()

### Are these changes tested?

Expecting test runner to do this.

No

Closes #3203

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
@@ -62,7 +62,7 @@ public class HadoopPositionOutputStream extends PositionOutputStream {
   @Override
   public void close() throws IOException {
     try (FSDataOutputStream fdos = wrapped) {
-      fdos.hflush();
+      fdos.flush();
     }
   }
 }


wgtmac:
Is there any behavioral change? cc @gszadovszky

steveloughran:
* the stream flush() ensures that all buffered data is pushed into whatever the final upload buffers of the client are, so in close() is guaranteed to have all its data.

* hflush() pushes it out all HDFS data nodes. Which is done immediately after in close().
* s3afs tells people off for hflush(), as its part of the Syncable interface, whose hsync() method is absolutely unsupported -code trying to use that (hbase, some spark streaming committers) do not get the semantics they required.

In https://github.com/apache/hadoop/pull/7662 I downgrade S3ABlockOutputStream.hflush() to a just increment stats and log@debug (it's not the critical method), but that'll be 3.4.2+ only

wgtmac (approved):
LGTM. Thanks for the explanation!

PR #3206, opened 2025-04-29
GH-3205: Make HadoopPositionOutputStream.close() safe to call even if closed

### LINKED ISSUE #3205: HadoopPositionOutputStream.close() is not idempotent (i.e. calling it more than once may fail)
If HadoopPositionOutputStream.close() is called more than once, an exception may be thrown. Since HadoopPositionOutputStream implements Closable, its close method should have no effect if already closed.

Example exception:

```
Caused by: org.apache.parquet.util.AutoCloseables$ParquetCloseResourceException: Unable to close resource
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:85) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:94) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.close(InternalParquetRecordWriter.java:144) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.hadoop.ParquetRecordWriter.close(ParquetRecordWriter.java:223) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.hadoop.hive.ql.io.parquet.write.ParquetRecordWriterWrapper.close(ParquetRecordWriterWrapper.java:137) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
...
Caused by: java.nio.channels.ClosedChannelException
	at org.apache.hadoop.hdfs.ExceptionLastSeen.throwException4Close(ExceptionLastSeen.java:73) ~[hadoop-hdfs-client-3.4.1.jar:?]
	at org.apache.hadoop.hdfs.DFSOutputStream.checkClosed(DFSOutputStream.java:158) ~[hadoop-hdfs-client-3.4.1.jar:?]
	at org.apache.hadoop.hdfs.DFSOutputStream.flushOrSync(DFSOutputStream.java:639) ~[hadoop-hdfs-client-3.4.1.jar:?]
	at org.apache.hadoop.hdfs.DFSOutputStream.hflush(DFSOutputStream.java:594) ~[hadoop-hdfs-client-3.4.1.jar:?]
	at org.apache.hadoop.fs.FSDataOutputStream.hflush(FSDataOutputStream.java:136) ~[hadoop-common-3.4.1.jar:?]
	at org.apache.parquet.hadoop.util.HadoopPositionOutputStream.close(HadoopPositionOutputStream.java:65) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.hadoop.ParquetFileWriter.close(ParquetFileWriter.java:1663) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.util.AutoCloseables.close(AutoCloseables.java:49) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:83) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.util.AutoCloseables.uncheckedClose(AutoCloseables.java:94) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.close(InternalParquetRecordWriter.java:144) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.parquet.hadoop.ParquetRecordWriter.close(ParquetRecordWriter.java:223) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
	at org.apache.hadoop.hive.ql.io.parquet.write.ParquetRecordWriterWrapper.close(ParquetRecordWriterWrapper.java:137) ~[hive-exec-4.0.0-20-IBM.jar:4.0.0]
```

### Component(s)
parquet-hadoop

### Component(s)

_No response_

dominicso:
### Rationale for this change
HadoopPositionOutputStream implements Closable, so its close method should have no effect if already closed.

Add code to HadoopPositionOutputStream.close() to check if already closed. If yes, do nothing.

### Are these changes tested?
Expecting test runner to do this.

No

 Closes #3205

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
@@ -25,6 +25,7 @@ import org.apache.parquet.io.PositionOutputStream;
 
 public class HadoopPositionOutputStream extends PositionOutputStream {
   private final FSDataOutputStream wrapped;
+  private boolean closed;
 
   HadoopPositionOutputStream(FSDataOutputStream wrapped) {
     this.wrapped = wrapped;
@@ -61,8 +62,13 @@ public class HadoopPositionOutputStream extends PositionOutputStream {
 
   @Override
   public void close() throws IOException {
+    if (closed) {
+      return;
+    }
     try (FSDataOutputStream fdos = wrapped) {
       fdos.flush();
+    } finally {
+      closed = true;
     }
   }
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
@@ -20,12 +20,13 @@
 package org.apache.parquet.hadoop.util;
 
 import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.parquet.io.PositionOutputStream;
 
 public class HadoopPositionOutputStream extends PositionOutputStream {
   private final FSDataOutputStream wrapped;
-  private boolean closed;
+  private final AtomicBoolean closed = new AtomicBoolean(false);
 
   HadoopPositionOutputStream(FSDataOutputStream wrapped) {
     this.wrapped = wrapped;
@@ -62,13 +63,11 @@ public class HadoopPositionOutputStream extends PositionOutputStream {
 
   @Override
   public void close() throws IOException {
-    if (closed) {
+    if (closed.getAndSet(true)) {
       return;
     }
     try (FSDataOutputStream fdos = wrapped) {
       fdos.flush();
-    } finally {
-      closed = true;
     }
   }
 }


wgtmac:
If https://github.com/apache/parquet-java/pull/3204 is merged, do we still need this fix?

cc @steveloughran

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopPositionOutputStream.java
```diff
@@ -61,8 +62,13 @@ public void flush() throws IOException {
 
   @Override
   public void close() throws IOException {
+    if (closed) {
```
steveloughran, line 65: safest to use an atomic bool here
```
if (closed.getAndSet(true) { 
  return;
}
```

streams are nominally not thread safe, but cross stream use happens. Making close() itself thread safe is low cost and avoids problems on cleanup

steveloughran (changes requested):
hadoop streams are meant to have close() idempotent, it is hflush() which is failing here as that is the strict call. Even with a switch to flush() things may have problems.


I think this needs a test, even just for local file output, calling close() twice, ideally on separate threads.

Is there one?

steveloughran:
just looked at flush() implementations

s3a: tells you off
```
  @Override
  public synchronized void flush() throws IOException {
    try {
      checkOpen();
    } catch (ClosedIOException e) {
      LOG.warn("Stream closed: {}", e.getMessage());
      return;
    }
    S3ADataBlocks.DataBlock dataBlock = getActiveBlock();
    if (dataBlock != null) {
      dataBlock.flush();
    }
  }
```

abfs: probably raises `IllegalStateException`
file: whatever `java.io.FileOutputStream` does
hdfs: does seem to queue stuff

overall then: dangerous to call flush() on a closed stream. Wrapping it is safer

wgtmac:
Thanks @dominicso @steveloughran! https://github.com/apache/parquet-java/pull/3204 is merged. Let's rebase this PR.

dominicso:
Rebased and changed to use AtomicBoolean.

> I think this needs a test, even just for local file output, calling close() twice, ideally on separate threads.

@steveloughran If a test is added to use local file output, wouldn't it use `LocalPositionOutputStream` instead of `HadoopPositionOutputStream`? Perhaps a unit test should be added just to call `HadoopPositionOutputStream.close(`)?

Looking at `LocalPositionOutputStream.close()`, it looks like it also needs similar changes to make repeat calls to its close safe. Let me know if this should be done as part of this PR as well.

wgtmac (approved):
+1

steveloughran (approved):
LGTM

wgtmac:
Merged. Thanks @dominicso and @steveloughran!

PR #3208, opened 2025-04-30
GH-3207: ParquetFileReader supports detachFileInputStream

### LINKED ISSUE #3207: ParquetFileReader should allow closing resources except for file input stream

This allows the caller to reuse the `SeekableInputStream` but avoid leaking other resources held by `ParquetFileReader`, for example, in Spark Parquet vectorized reading code path, it opens two times for each Parquet file:

1. the first time opens and reads the footer, do some pruning and push down stuff.
2. the second time opens and reads the row groups

### Component(s)

_No response_

pan3793:
### Rationale for this change

This allows the caller to reuse the `SeekableInputStream` across `ParquetFileReader`s, for example, in Spark Parquet vectorized reading code path, it opens two times for each Parquet file:

1. the first time opens and reads the footer, do some pruning and push down stuff.
2. the second time opens and reads the row groups

it's possible to reuse the `SeekableInputStream` to reduce the NameNode RPC

Providing a `detachFileInputStream ` method for `ParquetFileReader` to allow detaching file input stream before closing the `ParquetFileReader`

### Are these changes tested?

N/A

No.

Closes #3207

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1743,8 +1743,12 @@ public class ParquetFileReader implements Closeable {
 
   @Override
   public void close() throws IOException {
+    close(true);
+  }
+
+  public void close(boolean closeFileInputStream) throws IOException {
     try {
-      if (f != null) {
+      if (closeFileInputStream && f != null) {
         f.close();
       }
     } finally {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -729,7 +729,7 @@ public class ParquetFileReader implements Closeable {
     return new ParquetFileReader(file, options, f);
   }
 
-  protected final SeekableInputStream f;
+  protected SeekableInputStream f;
   private final InputFile file;
   private final ParquetReadOptions options;
   private final Map<ColumnPath, ColumnDescriptor> paths = new HashMap<>();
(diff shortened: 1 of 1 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1743,8 +1743,12 @@ public OffsetIndex readOffsetIndex(ColumnChunkMetaData column) throws IOExceptio
 
   @Override
   public void close() throws IOException {
+    close(true);
+  }
+
+  public void close(boolean closeFileInputStream) throws IOException {
```
wgtmac, line 1749: Is it better to explicitly add a `detachFileInputStream` method to set `f` to null so we don't need to hack the `close` here?

cc @gszadovszky

pan3793: thanks for your suggestion, but are you comfortable removing the `final` modifier for `f`?

```
protected final SeekableInputStream f;
```

gszadovszky: +1

pan3793: updated

wgtmac: Thanks for the update! It would be good to reflect the change in the PR title.

pan3793: PR title is updated too

wgtmac: Merged. Thanks @pan3793 and @gszadovszky!

PR #3212, opened 2025-05-07
GH-3211: Implement Variant parquet reader

### LINKED ISSUE #3211: Implement Variant reader

Add support to read Variant values from Parquet based on the spec in https://github.com/apache/parquet-format/blob/master/VariantEncoding.md and https://github.com/apache/parquet-format/blob/master/VariantShredding.md.

### Component(s)

_No response_

cashmand:
### Rationale for this change

Provide a reference implementation for reading Variant values from Parquet based on the Variant shredding spec in https://github.com/apache/parquet-format/blob/master/VariantShredding.md.

Adds VariantConverter classes to convert shredded Variant values in Parquet into a single metadata/value pair, representing the full value.

The PR adds a few new methods to the existing Variant and VariantBuilder classes to simplify the shredding reader, including a couple that are public to simplify construction of the Avro tests.

The PR also adds support in parquet-avro for using this interface: when a parquet schema element is annotated with the Variant logical type, the Avro reader will produce a record containing only the reconstructed metadata/value pair, rather than the full shredded result.

### Are these changes tested?

Added unit tests, mostly based on similar tests in https://github.com/apache/iceberg/blob/main/parquet/src/test/java/org/apache/iceberg/parquet/TestVariantReaders.java.

Connectors can now read Variant values, and the parquet-avro connector will infer a Variant schema if the logical type annotation is present.

Closes #3211


parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -101,6 +132,32 @@ public Variant build() {
...
+  // This is used in the shredded reader to check if anything has been written to the Variant since the last call.
+  // Tracking in the reader code is a bit awkward, but maybe better than polluting this interface.
+  int getWritePos() {
+    return writePos;
+  }
+
+  /**
+   * @return the constructed Variant value binary, without metadata.
+   */
+  public byte[] valueWithoutMetadata() {
```
rdblue, line 144: I wonder if we should make metadata part of this API or if we should pass in a metadata instance that does the lookup to resolve a name. Then metadata would be a separate builder that implements that API.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -101,6 +132,32 @@ public Variant build() {
...
+    return Arrays.copyOfRange(writeBuffer, 0, writePos);
+  }
+
+  /**
+   * Directly append a Variant value. Its keys must already be in the metadata
+   * dictionary.
+   */
+  void shallowAppendVariant(Binary value) {
```
rdblue, line 152: I think this should use `ByteBuffer` rather than `Binary`.

Should this also check that fixed metadata is used?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -373,6 +430,21 @@ public void appendUUID(java.util.UUID uuid) {
     writePos += VariantUtil.UUID_SIZE;
   }
 
+
+  /**
+   * Append raw bytes in the form stored in Variant.
+   * @param bytes a 16-byte value.
+   */
+  void appendUUIDBytes(byte[] bytes) {
```
rdblue, line 438: ByteBuffer?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
cashmand: We could check if fixed metadata is used, although it is a bit messy because the nested object/array builders would need to bubble the question up to the root VariantBuilder every time.

I could make it a bit cleaner by changing the builder to store a pointer to the root builder (possibly itself). I think this is better than the current approach of storing the parent in the object/array builders, since those builder actually always want the root builder, not the parent.

cashmand: I switched to storing a `rootBuilder` field in VariantBuilder instead of a parent in each child, so that I can add this check. I think it's a mild improvement regardless of this check, but let me know what you think: https://github.com/apache/parquet-java/pull/3212/commits/dffa73c2114b75a873031b319e2202ad6900a3d9. cc @gene-db

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -61,6 +61,14 @@ public Variant(ByteBuffer value, ByteBuffer metadata) {
     }
   }
 
+  public ByteBuffer getValueRawBytes() {
```
rdblue, line 64: I think this should have a different name. It is odd to return `ByteBuffer` from a method referring to "bytes". What about `getValueBuffer` and `getMetadataBuffer`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -101,6 +135,34 @@ public Variant build() {
...
+   */
+  public byte[] valueWithoutMetadata() {
+    return Arrays.copyOfRange(writeBuffer, 0, writePos);
+  }
+
+  /**
+   * Directly append a Variant value. Its keys must already be in the metadata
+   * dictionary.
+   */
+  void shallowAppendVariant(ByteBuffer value) {
```
rdblue, line 155: I'm not sure that it is obvious what "shallow" means here. If the variant is a nested type (object or array) then what happens? Instead I think this should be `appendEncodedValue`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    @Override
+    public void addInt(int value) {
+      BigDecimal decimal = BigDecimal.valueOf(value, scale);
+      builder.builder.appendDecimal(decimal);
+    }
+  }
+
+  static class VariantUuidConverter extends VariantScalarConverter {
```
rdblue, line 448: UUID is usually a special case where we don't downcase in method or class names. I think this should be `VariantUUIDConverter` instead of `VariantUuidConverter`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
```diff
@@ -22,15 +22,13 @@
...
-  /** The parent VariantBuilder. */
-  private final VariantBuilder parent;
   /** The offsets of the elements in this array. */
   private final ArrayList<Integer> offsets;
   /** The number of values appended to this array. */
   protected long numValues = 0;
 
-  VariantArrayBuilder(VariantBuilder parent) {
-    this.parent = parent;
+  VariantArrayBuilder(VariantBuilder rootBuilder) {
```
rdblue, line 30: As long as we are changing the structure, would it make sense to separate out the metadata?

Right now, we assume that the builder can add keys. But that's a very simple interface with just `int addDictionaryKey(String)`. And we also have to add a boolean for fixed metadata and a new `setFixedMetadata` method. It seems like it may be easier to change the builder so that it accepts some `MetadataBuilder`:

```java
interface Metadata {
  int resolve(String key); // maybe a better name?
  ByteBuffer getEncodedBuffer();
}
```

Then you could pass fixed metadata using a simple wrapper class, `EncodedMetadata` that throws an exception if you're resolving a string that is missing. And when building metadata from scratch, `resolve` would assign the new ID and produce encoded metadata when `getEncodedBuffer` is called.

I think that would be simpler than having so much metadata-related state in the builder itself.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -53,10 +62,35 @@ public class VariantBuilder {
...
+  public void setFixedMetadata(ByteBuffer metadata) {
+    setFixedMetadata(VariantUtil.getMetadataMap(metadata));
+  }
+
   /**
    * @return the Variant value
    */
   public Variant build() {
+    if (fixedMetadata) {
+      throw new IllegalStateException("Cannot reconstruct metadata when using fixed metadata");
```
rdblue, line 92: If you passed in metadata, then this would not be needed. And it would separate the concerns of building metadata and building values.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -395,15 +471,15 @@ public VariantObjectBuilder startObject() {
...
+    this.objectBuilder = new VariantObjectBuilder(this.rootBuilder);
     return objectBuilder;
   }
 
   /**
    * Finishes appending the object to this builder. This method must be called after startObject(),
    * before other append*() methods can be called on this builder.
    */
-  protected void endObject() {
+  public void endObject() {
```
rdblue, line 482: :+1:

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -466,6 +466,21 @@ public Optional<Schema> visit(
...
+                String name = parquetGroupType.getName();
+                SchemaBuilder.FieldAssembler<Schema> builder = SchemaBuilder.builder(
+                        namespace(name, names))
+                    .record(name)
+                    .fields();
+                builder.name("metadata")
+                    .type(Schema.create(Schema.Type.BYTES))
+                    .noDefault();
+                builder.name("value").type().optional().type(Schema.create(Schema.Type.BYTES));
+                return of(builder.endRecord());
```
rdblue, line 482: I think we should also add a `variant` logical type name to the record.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -466,6 +466,21 @@ public Optional<Schema> visit(
                   LogicalTypeAnnotation.EnumLogicalTypeAnnotation enumLogicalType) {
                 return of(Schema.create(Schema.Type.STRING));
               }
+
+              @Override
+              public Optional<Schema> visit(
+                  LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
+                String name = parquetGroupType.getName();
```
rdblue, line 473: I wouldn't recommend using the schema builder. It has proven to be a confusing API because you can't really understand exactly what gets created. I think the normal API is much more clear and succinct:

```java
                List<Schema.Field> fields = new ArrayList<>();
                fields.add(new Schema.Field("metadata", Schema.create(Schema.Type.BYTES)));
                fields.add(new Schema.Field("value", Schema.create(Schema.Type.BYTES)));
                return of(Schema.createRecord(parquetGroupType.getName(), null, namespace(name, names), false, fields));
```

That corresponds to what we do in Iceberg: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java#L199-L209

parquet-avro/src/main/java/org/apache/parquet/avro/AvroConverters.java
```diff
@@ -339,4 +341,26 @@ public String convert(Binary binary) {
       return stringifier.stringify(binary);
     }
   }
+
+  static final class FieldVariantConverter<T> extends VariantColumnConverter {
+    protected final ParentValueContainer parent;
```
rdblue, line 346: This can be private

parquet-avro/src/main/java/org/apache/parquet/avro/AvroConverters.java
```diff
@@ -339,4 +341,26 @@ public String convert(Binary binary) {
...
+    public FieldVariantConverter(
+        ParentValueContainer parent, GroupType schema, Schema avroSchema, GenericData model) {
+      super(schema);
+      this.avroSchema = avroSchema;
+      this.model = model;
+      this.parent = parent;
+    }
+
+    @Override
+    public void addVariant(Binary value, Binary metadata) {
```
rdblue, line 359: Looks like this class needs to suppress the warning for an unchecked cast:

```java
  @Override
  @SuppressWarnings("unchecked")
  public void addVariant(...)
```

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
rdblue: Yeah, I think that adding the logical type here would allow the caller to convert this record into a `Variant` representation using Avro logical type converters. Not a high priority, but it would be a good addition.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantColumnConverter.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.List;
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.Type;
+
+/**
+ * Converter for shredded Variant values. Connectors should implement the addVariant method, similar to
+ * the add* methods on PrimitiveConverter.
+ */
+public abstract class VariantColumnConverter extends VariantElementConverter {
```
rdblue, line 32: It appears that this is the top-level Variant. I think that should be called out more clearly than just "VariantColumn". I see why that makes sense, but "column" is overloaded so we should have a name that signals to the reader a little more clearly. Using `VariantConverter` would make sense to me if it weren't already used for the interface. Maybe `VariantValueConverter` could be used for the interface instead since those are dealing specifically with values and this is both metadata and value?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantColumnConverter.java
```diff
@@ -0,0 +1,81 @@
...
+      if (fieldName.equals("metadata")) {
+        this.topLevelMetadataIdx = i;
+        if (!field.isPrimitive() || field.asPrimitiveType().getPrimitiveTypeName() != BINARY) {
+          throw new IllegalArgumentException("Metadata must be a binary value");
+        }
+      }
+    }
+    if (topLevelMetadataIdx < 0) {
+      throw new IllegalArgumentException("Metadata missing from schema");
+    }
```
rdblue, line 52: It looks like the loop and the check above this line could be replaced with:

```java
    this.topLevelMetadataIdx = variantSchema.getFieldIndex("metadata");
```

That will throw an exception if `metadata` isn't found.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+        builder.appendNull();
+      }
+    }
+  }
+
+  /**
+   * Converter for the metadata column. It sets the current metadata in the parent converter,
+   * so that it can be used by the typed_value converter on the same row.
+   */
+  static class VariantMetadataConverter extends PrimitiveConverter implements VariantConverter {
```
rdblue, line 227: Why is this an inner class of `VariantElementConverter` when it doesn't extend `VariantElementConverter`? This is independent and seems like it should be completely separate.

Also, why not create/extend a `BinaryConverter` interface to avoid duplicating the dictionary code? That would make sense to me.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+    private VariantBuilderTopLevelHolder builder;
+    Binary[] dict;
+
+    public VariantMetadataConverter() {
+      dict = null;
+    }
+
+    @Override
+    public void init(VariantBuilderHolder builderHolder) {
```
rdblue, line 236: Why create this converter and then pass in the holder?

To me, it seems simpler to make the "holder" the parent, so you always pass in the parent. This converter sets metadata on the parent. Other converters can use the parent (which would have a holder-like type) to get metadata and to set the result value.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+ * class.
+ *
+ * All converters for a Variant column append their results to a VariantBuilder as values are read from Parquet.
+ *
+ * Values in `typed_value` are appended by the child converter. Values in `value` are stored by a
+ * child converter, but only appended when completing this group. Additionally, object fields are
+ * appended by the `typed_value` converter, but because residual values are stored in `value`, this
+ * converter is responsible for finalizing the object.
+ */
+class VariantElementConverter extends GroupConverter implements VariantConverter {
```
rdblue, line 60: The nesting for the converters is odd. Rather than choosing one "main" converter and putting unrelated converters inside it, the more common Java pattern is to use a class with a private constructor and no methods to hold all of the converters. Something like this:

```java
class VariantConverters {
  // do not allow instantiating this class
  private VariantConverters() {}

  static class ShreddedVariantConverter extends GroupConverter implements VariantConverter { ... }
}
```

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+      } else if (fieldName.equals("typed_value")) {
+        this.typedValueIdx = i;
+      }
+    }
+
+    if (valueIdx >= 0) {
+      converters[valueIdx] = new VariantValueConverter(this);
+    }
+
+    if (typedValueIdx >= 0) {
```
rdblue, line 117: I think that this class is doing too much. I would rather have a separate class for each shredded case: primitive, object, and array. That would make the code easier to read and would avoid cases where you're keeping important state in unexpected places, like the branches in `end` that handle object or array based on whether `objectFieldName` is null.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+  // The following are only used if this is an object field.
+  private String objectFieldName = null;
+  private int objectFieldId = -1;
+  private VariantObjectConverter parent = null;
+
+  // Only used if typedValueIsObject is true.
+  private Set<String> shreddedObjectKeys;
+
+  @Override
+  public void init(VariantBuilderHolder holder) {
```
rdblue, line 80: Constructors should be before instance methods.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+        typedValueIsObject = true;
+        shreddedObjectKeys = new HashSet<>();
+        for (Type f : typedValue.getFields()) {
+          shreddedObjectKeys.add(f.getName());
+        }
+      } else {
+        typedConverter = VariantScalarConverter.create(field.asPrimitiveType());
+      }
+
+      assert (typedConverter != null);
```
rdblue, line 135: We don't use `assert` in Java. Instead, don't initialize `typedConverter` so that the compiler detects cases where it has not been initialized. Here's an example of the error message, which happens when I change the else case so that there branches where `typedConverter` is not assigned.

image

parquet-variant/src/main/java/org/apache/parquet/variant/VariantElementConverter.java
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    @Override
+    public void addValueFromDictionary(int dictionaryId) {
+      currentValue = dict[dictionaryId];
+    }
+  }
+
+  // Base class for converting primitive typed_value fields.
+  static class VariantScalarConverter extends PrimitiveConverter implements VariantConverter {
```
rdblue, line 313: `VariantScalar` doesn't necessarily signal that this is a shredded field How about `ShreddedScalarConverter`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantColumnConverter.java
cashmand: I already use `VariantValueConverter` for the converter for the `value` field :(. Maybe `VariantConverterInterface`? Naming is the hardest thing.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantArrayBuilder.java
cashmand: Good idea, I refactored along the lines you suggested, with VariantBuilder using `MetadataBuilder` by default, but `ImmutableMetadata` when created from the shredding reader.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
cashmand: Done.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
```diff
@@ -0,0 +1,783 @@
...
+    public VariantValueConverter(VariantElementConverter parent) {
+      super(parent);
+      this.currentValue = null;
+    }
+
+    void reset() {
+      currentValue = null;
+    }
+
+    Binary getValue() {
```
cashmand, line 324: I'll try to replace this with a `setBinaryValue()` method to push the value to the parent, to be more consistent with how `setMetadata()` is being called.

parquet-variant/src/main/java/org/apache/parquet/variant/ImmutableMetadata.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+
+/**
+ * Metadata that is set up at construction time, and never modified.
+ *
+ */
+public class ImmutableMetadata implements Metadata {
+  private HashMap<String, Integer> dictionary = null;
+  private ByteBuffer metadataBuffer;
```
rdblue, line 28: These can (and should) both be `final`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
```diff
@@ -0,0 +1,776 @@
...
+   * so that it can be used by the typed_value converter on the same row.
+   */
+  static class VariantMetadataConverter extends BinaryConverter implements VariantConverter {
+
+    public VariantMetadataConverter(VariantColumnConverter parent) {
+      super(parent);
+    }
+
+    @Override
+    public void addBinary(Binary value) {
```
rdblue, line 302: It would probably make sense to refactor these methods into the base `BinaryConverter` and add a `convert` method that does the implementation-specific part:

```java
abstract class BinaryConverter ... {
  protected abstract void convert(Binary value);

  public void addBinary(Binary value) {
    convert(value);
  }

  public void addValueFromDictionary(int dictionaryId) {
    convert(dict[dictionaryId]);
  }
}

class VariantMetadataConverter extends BinaryConverter {
  protected void convert(Binary value) {
    getParent().setMetadata(value);
  }
}
```

Also, why use `Binary` instead of `ByteBuffer` for the internal API?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
```diff
@@ -0,0 +1,776 @@
...
+    }
+
+    @Override
+    public VariantConverter getParent() {
+      return parent;
+    }
+
+    @Override
+    public VariantBuilder getBuilder() {
+      return null;
```
rdblue, line 275: Does this need to be a `VariantConverter`? I don't think anything here is specific to Variant. I'm not sure that the children need to be either, or at least that the metadata converter doesn't need to be. That could be passed a `Consumer<ByteBuffer>` instead of a parent and pass the buffer to it.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
cashmand: Sounds good. There was no reason to use `Binary`, it was just what the `addBinary` method in `PrimitiveConverter` provides. I can convert it to `ByteBuffer` immediately.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
```diff
@@ -0,0 +1,776 @@
...
+    }
+
+    @Override
+    public void addBinary(Binary value) {
+      ((VariantColumnConverter) getParent()).setMetadata(value);
+    }
+
+    @Override
+    public void addValueFromDictionary(int dictionaryId) {
+      ((VariantColumnConverter) getParent()).setMetadata(dict[dictionaryId]);
```
rdblue, line 308: As I mentioned above, changing this so that `setMetadata` is passed as a `Consumer<ByteBuffer>`:

```java
  abstract static class BinaryConverter extends PrimitiveConverter {
    Binary[] dict;

    BinaryConverter() {
      dict = null;
    }

    protected abstract void handleBinary(Binary value);

    @Override
    public boolean hasDictionarySupport() {
      return true;
    }

    @Override
    public void setDictionary(Dictionary dictionary) {
      dict = new Binary[dictionary.getMaxId() + 1];
      for (int i = 0; i <= dictionary.getMaxId(); i++) {
        dict[i] = dictionary.decodeToBinary(i);
      }
    }

    @Override
    public void addBinary(Binary value) {
      handleBinary(value);
    }

    @Override
    public void addValueFromDictionary(int dictionaryId) {
      handleBinary(dict[dictionaryId]);
    }
  }

  static class VariantMetadataConverter extends BinaryConverter {
    private final Consumer<ByteBuffer> parent;

    public VariantMetadataConverter(Consumer<ByteBuffer> parent) {
      super();
      this.parent = parent;
    }

    @Override
    protected void handleBinary(Binary value) {
      parent.accept(value.toByteBuffer());
    }
  }
```

Then that converter is instantiated like this:

```java
      converters[topLevelMetadataIdx] = new VariantConverters.VariantMetadataConverter(this::setMetadata);
```

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
cashmand: Oh, actually, I think there was a reason to use `Binary`. When I set the metadata, I check for reference equality with the previous value, so that if the metadata was the same dictionary-encoded value in every row, we don't need to rebuild it. If I immediately convert it to a ByteBuffer with `toByteBuffer`, I don't think it would compare equal to the previous ByteBuffer.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
```diff
@@ -0,0 +1,776 @@
...
+        Type field = fields.get(i);
+        String fieldName = field.getName();
+        if (fieldName.equals("value")) {
+          this.valueIdx = i;
+          if (!field.isPrimitive() || field.asPrimitiveType().getPrimitiveTypeName() != BINARY) {
+            throw new IllegalArgumentException("Value must be a binary value");
+          }
+        } else if (fieldName.equals("typed_value")) {
+          this.typedValueIdx = i;
+        }
```
rdblue, line 119: If there is no `else` case, then I think this is going to fail at read time when Parquet tries to call methods on a `null` converter. To ignore a field we would need to add a noop converter. For now, I'd recommend just failing with a good exception message, like `throw new UnsupportedOperationException("Cannot read variant with unexpected field: " + fieldName);`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantConverters.java
cashmand: Good point, it doesn't.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -807,4 +808,44 @@ static String getMetadataKey(ByteBuffer metadata, int id) {
...
   }
+
+  /**
+   * Returns a map from each string to its ID in the Variant metadata.
+   * @param metadata The Variant metadata
+   * @return A map from metadata key to its position.
+   */
+  static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
+    int pos = metadata.position();
+    checkIndex(pos, metadata.limit());
```
rdblue, line 819: Doesn't `ByteBuffer` guarantee that its position and limit are valid?

cashmand: I believe it's possible for position to equal limit (i.e. no byte can be read).

rdblue:
This looks good to me. I'll merge when tests are passing.

rdblue:
Thanks for fixing the tests, @cashmand! I merged this.

PR #3221, opened 2025-05-20
GH-3223: Implement Variant parquet writer

### LINKED ISSUE #3223: Implement Variant writer

Add support to write shredded Variant values to Parquet based on the spec in https://github.com/apache/parquet-format/blob/master/VariantEncoding.md and https://github.com/apache/parquet-format/blob/master/VariantShredding.md

### Component(s)

_No response_

cashmand:
### Rationale for this change

Provide a reference implementation for writing Variant values from Parquet based on the Variant shredding spec in https://github.com/apache/parquet-format/blob/master/VariantShredding.md.

Adds a `VariantValueWriter` to write Variant values to a Parquet file with a shredding schema.

The PR also adds support in parquet-avro to identify Variant fields in a Parquet schema, and write using the shredding schema.

### Are these changes tested?

Uni tests included in the PR, based on similar tests in https://github.com/apache/iceberg/blob/main/parquet/src/test/java/org/apache/iceberg/parquet/TestVariantWriters.java.

Connectors can now write to a shredded Variant table, and the parquet-avro connector will do this if a shredding schema is provided.

Closes #3223


parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,310 @@
...
+import org.apache.parquet.schema.Type;
+
+/**
+ * Class to write Variant values to a shredded schema.
+ */
+public class VariantValueWriter {
+  // We store a reference to the metdata so that we can lazily build metadataMap if needed.
+  // It could be relatively expensive to build, and if everything shreds perfectly, we won't need it.
+  private ByteBuffer metadata;
+  private HashMap<String, Integer> metadataMap = null;
```
cashmand, line 37: TODO: replace with an ImmutableMetadata. I think we still want to create is lazily.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,517 @@
...
+        new TestWriterBuilder(path).withFileType(testSchema.writeSchema).build()) {
+      for (GenericRecord record : records) {
+        writer.write(record);
+      }
+    }
+
+    Configuration conf = new Configuration();
+    // We need to set an explicit read schema because Avro wrote the shredding schema as the Avro
+    // schema in the write, and it will use that by default. If we write using a proper shredding
+    // writer, the Avro schema should just contain a <metadata, value> record, and we won't need this.
```
rdblue, line 356: Is this out of date for this test suite?

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,517 @@
...
+
+  private static MessageType parquetSchema(GroupType variantGroup) {
+    return Types.buildMessage()
+        .required(INT32)
+        .named("id")
+        .addField(variantGroup)
+        .named("table");
+  }
+
+  private static MessageType readSchema = parquetSchema(variantGroup);
```
rdblue, line 98: This can be final and since it's static it should use `ALL_CAPS`.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,517 @@
...
+
+  private static ByteBuffer variant(long val) {
+    return variant(b -> b.appendLong(val));
+  }
+
+  private static ByteBuffer variant(String s) {
+    return variant(b -> b.appendString(s));
+  }
+
+  private static GroupType variantGroup = Types.buildGroup(Type.Repetition.REQUIRED)
```
rdblue, line 82: Should this be `UNSHREDDED_SCHEMA` or `UNSHREDDED_GROUP`?

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,517 @@
...
+    for (Variant v : VARIANTS) {
+      GenericRecord record = createRecord(1, v);
+      MessageType writeSchema = shreddingSchema(v);
+      TestSchema testSchema = new TestSchema(writeSchema, readSchema);
+
+      GenericRecord actual = writeAndRead(testSchema, record);
+      assertEquals(record.get(0), actual.get(0));
+      assertEquals(((GenericRecord) record.get(1)).get(0), ((GenericRecord) actual.get(1)).get(0));
+      // assertEquals(((GenericRecord) record.get(1)).get(1), ((GenericRecord) actual.get(1)).get(1));
+      if (!((GenericRecord) record.get(1)).get(1).equals(((GenericRecord) actual.get(1)).get(1))) {
```
rdblue, line 274: This seems odd. What's going on here?

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,517 @@
...
+      List<GenericRecord> expected = new ArrayList<>();
+      for (int i = 0; i < VARIANTS.length; i++) {
+        expected.add(createRecord(i, VARIANTS[i]));
+      }
+
+      MessageType writeSchema = shreddingSchema(v);
+      TestSchema testSchema = new TestSchema(writeSchema, readSchema);
+
+      List<GenericRecord> actual = writeAndRead(testSchema, expected);
+      // TODO: CHECK RESULTS
```
rdblue, line 292: Looks like this is needed.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -61,6 +61,14 @@ public Variant(ByteBuffer value, ByteBuffer metadata) {
     }
   }
 
+  public ByteBuffer getValueRawBytes() {
```
rdblue, line 64: "raw" isn't very clear. What about `encodedValue` and `encodedMetadata`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,310 @@
...
+import org.apache.parquet.schema.LogicalTypeAnnotation;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+
+/**
+ * Class to write Variant values to a shredded schema.
+ */
+public class VariantValueWriter {
+  // We store a reference to the metdata so that we can lazily build metadataMap if needed.
+  // It could be relatively expensive to build, and if everything shreds perfectly, we won't need it.
```
rdblue, line 35: Is this true? I think that metadata should be the same regardless of shredding so that it does not need to be rebuilt and re-encoded. That way we also don't need to change IDs.

cashmand: I'm not referring here to building metadata, but using the existing metadata to build a map from keys to metadata IDs. We need it in the case where we need to build a residual object in `value` if an object contains keys that aren't in `typed_value`. In any case, with the latest version of the reader PR, this happens in the `ImmutableMetadata` constructor, so the code will just defer creating `ImmutableMetadata` until it's actually needed.

The truth is, I don't think we strictly need any metadata at all for the writer: it's only necessary because we're using the `VariantObjectBuilder` interface to construct the leftover objects in `value`, and the object builder requires the key String to be provided by name, not ID, so that it can sort objects by key when it finalizes an object. If we assume the input is valid (i.e. already sorted), we could create a short-circuiting version of `VariantObjectBuilder` that operates entirely in terms of the IDs in the variant, in which case we'd never need to reference the metadata. Let me know if you think it's worth extending the object builder with that change. I think it's just a performance benefit, and maybe we don't care that much about that if this is mainly a reference implementation.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
cashmand: Oops, the commented out line is the right assert. I think I was doing something in the debugger and forgot to restore it.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
cashmand: Yes, deleted.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
cashmand: Changed to UNSHREDDED_GROUP.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -96,7 +96,7 @@ public ByteBuffer encodedValue() {
    */
   public void appendEncodedValue(ByteBuffer value) {
     onAppend();
-    int size = value.remaining();
+    int size = VariantUtil.valueSize(value);
```
cashmand, line 99: Without this change, we would write the entire `ByteBuffer`, which could go to the end of the buffer. The written value would technically be correct (since a reader would ignore the remaining bytes), but obviously very inefficient. An alternative to limiting the size here is to do it in the `getValueBuffer()` method. But depending on what the caller plans to do with the buffer, the extra bytes might not matter, and we'd be parsing the value unnecessarily.

parquet-avro/src/test/java/org/apache/parquet/avro/AvroTestUtil.java
```diff
@@ -129,4 +130,36 @@ public static Configuration conf(String name, boolean value) {
     conf.setBoolean(name, value);
     return conf;
   }
+
+  /**
+   * Assert that to Variant values are logically equivalent.
+   * E.g. object fields may be ordered differently in the binary.
```
rdblue: Do you mean values? The fields themselves must have the same order because the ID list must be in lexicographical order with respect to the keys.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
rdblue: Yeah, I think this is a good change if we are not going to guarantee that the ByteBuffer is limited to the current value. That is an assumption that we make in Iceberg, but not here.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -848,4 +848,73 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
   }
+
+  /**
+   * Computes the actual size (in bytes) of the Variant value at `value[pos...]`
+   * @param value The Variant value
+   * @return The size (in bytes) of the Variant value
+   */
+  public static int valueSize(ByteBuffer value) {
+    int pos = value.position();
+    checkIndex(pos, value.limit());
```
rdblue: This is guaranteed by `Buffer`, the parent of `ByteBuffer`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -848,4 +848,73 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
     }
     return result;
   }
+
+  /**
+   * Computes the actual size (in bytes) of the Variant value at `value[pos...]`
```
rdblue: Is this still relevant? It looks like it was copied from an instance method that expected `value` and `pos` fields.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -848,4 +848,73 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+   * @param value The Variant value
+   * @return The size (in bytes) of the Variant value
+   */
+  public static int valueSize(ByteBuffer value) {
+    int pos = value.position();
+    checkIndex(pos, value.limit());
+    int basicType = value.get(pos) & BASIC_TYPE_MASK;
+    int typeInfo = (value.get(pos) >> BASIC_TYPE_BITS) & PRIMITIVE_TYPE_MASK;
+    switch (basicType) {
```
rdblue, line 860: The logic in this function look correct to me. It may be helpful to note in a comment that the size includes the header byte.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -848,4 +848,73 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+      case SHORT_STR:
+        return 1 + typeInfo;
```
rdblue: This made me go check the spec. It may be easier to maintain/read if you just duplicated the logic.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -181,9 +184,58 @@ public void write(T record) {
...
+      } else {
+        binarySchema = false;
+        break;
+      }
+    }
+
+    if (binarySchema) {
+      VariantValueWriter.write(recordConsumer, schema, new Variant(value, metadata));
+    } else {
+      // If the schema was something other than value and metaadata, treat the value as a non-variant record.
```
rdblue: Typo: `metaa`

rdblue: I think that this should be a failure. `writeVariantFields` is called when there is a variant logical type annotation. If the record is not actually a variant then it is better to fail quickly instead of trying to recover by treating it like a record and then failing for some unknown reason later.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -181,9 +184,58 @@ public void write(T record) {
...
+      if (fieldType.getName() == "value") {
+        Object valueObj = model.getField(record, avroField.name(), index);
+        if (valueObj instanceof byte[]) {
+          value = ByteBuffer.wrap((byte[]) valueObj);
+        } else {
+          value = (ByteBuffer) valueObj;
+        }
+      } else if (fieldType.getName() == "metadata") {
+        Object metadataObj = model.getField(record, avroField.name(), index);
+        if (metadataObj instanceof byte[]) {
```
rdblue: Why is the type passed in here not known? I think that the representation used for metadata and value should use (and validate) an expected type. Probably `ByteBuffer`.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+import org.apache.parquet.variant.ImmutableMetadata;
+import org.apache.parquet.variant.Variant;
+import org.apache.parquet.variant.VariantArrayBuilder;
+import org.apache.parquet.variant.VariantBuilder;
+import org.apache.parquet.variant.VariantObjectBuilder;
+import org.junit.Test;
+
+public class TestWriteVariant extends DirectWriterTest {
+
+  // Construct a variant, and return the value binary, dropping metadata.
```
rdblue: This doesn't really discard metadata. The `Variant` returned has both the metadata and value buffers.

parquet-avro/src/test/java/org/apache/parquet/avro/AvroTestUtil.java
cashmand: Yes, that's what I meant. Is `values in an object may be ordered differently` better?

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+      VariantObjectBuilder ob = b.startObject();
+      ob.appendKey("a");
+      ob.appendInt(123456789);
+      ob.appendKey("c");
+      ob.appendString("string");
+      b.endObject();
+    });
+
+    // The first array element defines the schema.
+    TEST_ARRAY = variant(TEST_METADATA, b -> {
```
rdblue: Looks like this deviates from the Iceberg array tests, which use an array of strings, an array of strings and an int, an array of arrays of strings, and then some nested objects. Why does this skip the simpler array cases?

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+      fullVariant(b -> b.appendTimestampNanosTz(-1234567890L)),
+      fullVariant(b -> b.appendTimestampNanosNtz(1234567890L)),
+      fullVariant(b -> b.appendTimestampNanosNtz(-1234567890L)),
+      fullVariant(b -> b.appendUUID(UUID.fromString("f24f9b64-81fa-49d1-b74e-8c09a6e31c56")))
+    };
+  }
+
+  /**
+   * Create a record containing a Variant value using the standard schema.
+   * @return
```
rdblue: Omit `@return` if there isn't relevant context to add.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+      List<GenericRecord> expected = new ArrayList<>();
+      for (int i = 0; i < VARIANTS.length; i++) {
+        expected.add(createRecord(i, VARIANTS[i]));
+      }
+
+      MessageType writeSchema = shreddingSchema(valueForSchema);
+      TestSchema testSchema = new TestSchema(writeSchema, READ_SCHEMA);
+
+      List<GenericRecord> actual = writeAndRead(testSchema, expected);
+      assertEquals(actual.size(), VARIANTS.length);
```
rdblue: I think it would be more clear if this looped over `expected`.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+      Variant actualV = new Variant((ByteBuffer) ((GenericRecord) actual.get(1)).get(1), (ByteBuffer)
+          ((GenericRecord) actual.get(1)).get(0));
+      AvroTestUtil.assertEquivalent(v, actualV);
+    }
+  }
+
+  @Test
+  public void testMixedShredding() throws IOException {
+    for (Variant valueForSchema : VARIANTS) {
+      List<GenericRecord> expected = new ArrayList<>();
```
rdblue: This can be moved outside the `valueForSchema` loop because it doesn't change.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+
+    @Override
+    protected WriteSupport<GenericRecord> getWriteSupport(Configuration conf) {
+      return new AvroWriteSupport<>(schema, new AvroSchemaConverter().convert(schema), GenericData.get());
+    }
+  }
+
+  GenericRecord writeAndRead(TestSchema testSchema, GenericRecord record) throws IOException {
+    List<GenericRecord> result = writeAndRead(testSchema, Arrays.asList(record));
+    assert (result.size() == 1);
```
rdblue: This isn't the same as a test assertion. In tests, prefer using `assertEquals(1, result.size())`.

In fact, `assert` is not used in most Java projects in our space.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
cashmand: Sounds good. My thought was that someone might want to write data that has already been shredded - i.e. annotate the Parquet type as Variant, but directly write using an Avro record that already matches the shredded schema. But if there's a real use case for that, and no other workaround, it can always be added later.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
cashmand: I was just copying the code used for writing `BYTES`: https://github.com/apache/parquet-java/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L357. But I can enforce that it's always `ByteBuffer`.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
```diff
@@ -181,9 +184,58 @@ public void write(T record) {
...
+    // Extract the value and metadata binary.
+    for (int index = 0; index < avroFields.size(); index++) {
+      Schema.Field avroField = avroFields.get(index);
+      Schema fieldSchema = AvroSchemaConverter.getNonNull(avroField.schema());
+      if (!fieldSchema.getType().equals(Schema.Type.BYTES)) {
+        binarySchema = false;
+        break;
+      }
+      Type fieldType = fields.get(index);
+      if (fieldType.getName() == "value") {
```
rdblue: Strings can't be compared with `==` because it checks identity. For strings you need to use `"value".equals(fieldType.getName())`.

rdblue: Looks like this is working because the strings used by tests are also interned constants.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
cashmand: Oops, yes, this comment had been on the `variant()` functions below, and ended up in the wrong place.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+
+import static org.apache.parquet.avro.AvroTestUtil.*;
+import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.*;
+import static org.junit.Assert.*;
+
+import java.io.File;
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.nio.ByteBuffer;
+import java.util.*;
```
rdblue: Please remove wildcard imports.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
```diff
@@ -0,0 +1,518 @@
...
+import org.apache.parquet.schema.LogicalTypeAnnotation.TimeUnit;
+import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
+import org.apache.parquet.variant.ImmutableMetadata;
+import org.apache.parquet.variant.Variant;
+import org.apache.parquet.variant.VariantArrayBuilder;
+import org.apache.parquet.variant.VariantBuilder;
+import org.apache.parquet.variant.VariantObjectBuilder;
+import org.junit.Test;
+
+public class TestWriteVariant extends DirectWriterTest {
```
rdblue, line 54: I've verified that the tests are actually testing what is intended. I'd prefer to have the same array cases as Iceberg, but otherwise this looks mostly good.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+      LogicalTypeAnnotation logicalType = primitiveType.getLogicalTypeAnnotation();
+      PrimitiveType.PrimitiveTypeName primitiveTypeName = primitiveType.getPrimitiveTypeName();
+
+      switch (variantType) {
+        case BOOLEAN:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.BOOLEAN;
+        case BYTE:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.INT32
+              && logicalType instanceof LogicalTypeAnnotation.IntLogicalTypeAnnotation
+              && ((LogicalTypeAnnotation.IntLogicalTypeAnnotation) logicalType).getBitWidth() == 8;
```
rdblue, line 141: This needs to also check that the logical type is signed.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        recordConsumer.addBinary(Binary.fromReusedByteBuffer(variant.getBinary()));
+        break;
+      default:
+        throw new IllegalArgumentException("Unsupported scalar type: " + variant.getType());
+    }
+  }
+
+  private void writeArrayValue(RecordConsumer recordConsumer, Variant variant, GroupType arrayType) {
+    if (variant.getType() != Variant.Type.ARRAY) {
+      throw new IllegalArgumentException("Expected array type but got: " + variant.getType());
```
rdblue: Use `Preconditions.checkArgument`.

Also, error messages are typically more direct, rather than using "but got". A better message is `Cannot write variant type=%s as array`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    // Get the element type from the array schema
+    GroupType listType = arrayType.getType(0).asGroupType();
```
rdblue: Do you also want to assert that the structure is a 3-level list rather than just accessing fields?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+    Type elementType = listType.getType(0);
+
+    recordConsumer.startGroup();
```
rdblue, line 304: Comments would help here. The Avro list writer has `// group wrapper (original type LIST)`, which would probably be `// list field, annotated as LIST` now.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+    recordConsumer.startField(arrayType.getFieldName(0), 0);
```
rdblue: It would be a bit better to store the repeated type and access its name here: `repeatedType.getName()`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+    // Write each array element
+    for (int i = 0; i < variant.numArrayElements(); i++) {
+      recordConsumer.startGroup();
```
rdblue: `// repeated group`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+      recordConsumer.startField("element", 0);
```
rdblue: Could you use a constant like in `AvroWriteSupport`?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+
+      recordConsumer.startGroup();
```
rdblue: `// element group`

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        case FLOAT:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.FLOAT;
+        case DOUBLE:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.DOUBLE;
+        case DECIMAL4:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.INT32
+              && logicalType instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation;
+        case DECIMAL8:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.INT64
+              && logicalType instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation;
```
rdblue: Doesn't this need to check the decimal scale?

rdblue: I think we need a couple of different decimals in our tests to catch this case.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * Write a Variant value to a shredded schema.
+   */
+  public static void write(RecordConsumer recordConsumer, GroupType schema, Variant value) {
+    recordConsumer.startGroup();
+    int metadataIndex = schema.getFieldIndex("metadata");
+    recordConsumer.startField("metadata", metadataIndex);
+    recordConsumer.addBinary(Binary.fromConstantByteBuffer(value.getMetadataBuffer()));
+    recordConsumer.endField("metadata", metadataIndex);
+    VariantValueWriter writer = new VariantValueWriter(recordConsumer, value.getMetadataBuffer());
```
rdblue, line 66: This is not a strong opinion, but it looks like the only reason why there is a writer instance is to carry the metadata buffer. Since all of these are internal methods that aren't exposed, I'd rather just use method args to pass the record consumer and metadata.

I guess this also lazily instantiates an `ImmutableMetadata`, but that doesn't seem like a high cost.

parquet-avro/src/test/java/org/apache/parquet/avro/TestWriteVariant.java
cashmand: When I originally copied the tests, Iceberg didn't have array support, so I just added my own tests. I'll add the cases from the Iceberg tests.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
cashmand: I guess I could, but I don't really understand the benefit. I like that the current version makes it clear that the entire write shares a record consumer and metadata.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
cashmand: Yikes, good catch! I added more decimal test cases, which catches this issue.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
rdblue: If I understand correctly, you're saying that this is avoiding _deserializing_ the object's metadata. The metadata may need to be read into a map in order to re-encode the residual object. I think that makes sense.

The residual object is a strict subset of the original object's fields (just the remaining unshredded ones) so the residual's fields will be in the same order as the original and will use the same IDs because the metadata doesn't change. But the API for actually building the residual object uses the field names rather than static IDs. So you can either build the residual with the normal API or you could have a special case in the builder where the caller is responsible for supplying the IDs instead of names. I wouldn't bother with that optimization in this version.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
rdblue: Looks like this was only done for metadata and not for value. Since it matches bytes, I don't really mind allowing both, but it does make this longer than necessary.

I'd also recommend using `Preconditions.checkArgument` rather than using a separate `if` that throws a generic `RuntimeException`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
rdblue: Sounds reasonable to me.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+                      && ((LogicalTypeAnnotation.IntLogicalTypeAnnotation) logicalType).isSigned()
+                      && ((LogicalTypeAnnotation.IntLogicalTypeAnnotation) logicalType)
+                              .getBitWidth()
+                          == 32));
+        case LONG:
+          return primitiveTypeName == PrimitiveType.PrimitiveTypeName.INT64
+              && (logicalType == null
+                  || (logicalType instanceof LogicalTypeAnnotation.IntLogicalTypeAnnotation
+                      && ((LogicalTypeAnnotation.IntLogicalTypeAnnotation) logicalType)
+                          .isSigned()));
```
rdblue: Doesn't this also need to check that the width is 64?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+
+  /**
+   * Write an object to typed_value
+   *
+   * @return the residual value that must be written to the value column, or null if all values were written
+   *         to typed_value.
+   */
+  private ByteBuffer writeObjectValue(RecordConsumer recordConsumer, Variant variant, GroupType objectType) {
+    if (variant.getType() != Variant.Type.OBJECT) {
+      throw new IllegalArgumentException("Expected object type but got: " + variant.getType());
```
rdblue: Looks like the array case was updated but this one was not.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+      int valueIdx = schema.getFieldIndex("value");
+      recordConsumer.startField("value", valueIdx);
+      recordConsumer.addBinary(Binary.fromReusedByteBuffer(value.getValueBuffer()));
+      recordConsumer.endField("value", valueIdx);
+    }
+  }
+
+  // Return true if the logical type is a decimal with the same scale as the provided value, with enough
+  // precision to hold the value. The provided value must be a decimal.
+  private boolean compatibleDecimalType(Variant value, LogicalTypeAnnotation logicalType) {
```
rdblue: This can be static, right?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+      return false;
+    }
+    LogicalTypeAnnotation.DecimalLogicalTypeAnnotation decimalType =
+        (LogicalTypeAnnotation.DecimalLogicalTypeAnnotation) logicalType;
+
+    BigDecimal decimal = value.getDecimal();
+    return decimal.scale() == decimalType.getScale() && decimal.precision() <= decimalType.getPrecision();
+  }
+
+  private boolean isTypeCompatible(Variant.Type variantType, Type typedValueField, Variant value) {
```
rdblue: Can be static.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+      }
+    } else if (typedValueField.getLogicalTypeAnnotation()
+        instanceof LogicalTypeAnnotation.ListLogicalTypeAnnotation) {
+      return variantType == Variant.Type.ARRAY;
+    } else {
+      return variantType == Variant.Type.OBJECT;
+    }
+  }
+
+  private void writeScalarValue(RecordConsumer recordConsumer, Variant variant, PrimitiveType type) {
```
rdblue: Type is not used.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
rdblue: I think it's fine to pass `recordConsumer` as an instance field, but right now this is mixed. `writeScalarValue` accepts a `recordConsumer` but is not a static method, so its argument has the same name as the instance field. I think this should fix that issue and either make methods static and pass in `recordConsumer` or should use the instance field. Methods that don't use instance fields should be `static`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.LogicalTypeAnnotation;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+
+/**
+ * Class to write Variant values to a shredded schema.
+ */
+public class VariantValueWriter {
+  private ByteBuffer metadataBuffer;
```
rdblue: This can be final.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
```diff
@@ -0,0 +1,372 @@
...
+  // We defer initializing the ImmutableMetata until it's needed. It has some construction cost, and if all
+  // object fields are shredded into typed_value, it will never be used.
+  private ImmutableMetadata metadata = null;
+  private RecordConsumer recordConsumer;
```
rdblue: This can be final.

Also, I like to keep final instance fields located together before others.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
cashmand: I don't think the [spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md) allows any width other than 64 to be annotated onto an INT64, but I can add it for consistency/readability.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantValueWriter.java
cashmand: Good point, I'll remove `recordConsumer` from the argument list of the non-static methods.

rdblue (approved):

rdblue:
Thanks for the updates, @cashmand! I think this looks good to go. Running the tests again.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java
cashmand: Hi @rdblue, I realized that this was needed in order to allow the tests in TestReadVariant.java to work, because those tests are doing what I described in this comment: writing already-shredded data with the logical annotation on the top-level struct.

I pushed a commit to explicitly check for the cases of having exactly three fields in the schema, or having two fields with `typed_value`, which is also valid, but otherwise leave in the stricter checking that was added based on this comment. Let me know if you think this is okay, or if you prefer a different approach.

rdblue:
Running tests

rdblue:
Merged! Thanks @cashmand for getting this in. Nice work!

PR #3222, opened 2025-05-21
MINOR: Update BoundingBox for Empty and Antimeridian Handling

zhangfengcdt:
### Rationale for this change

The current implementation of BoundingBox.updateBounds() incorrectly handles wraparound conditions (where geometries cross the antimeridian at ±180° longitude). JTS (Java Topology Suite) operates in a 2D Cartesian space and doesn't natively support wraparound geometries. This change fixes how bounding boxes are updated to properly handle these edge cases, ensuring correct spatial calculations when working with geometries that cross the international date line.

This is also to make the behavior consistent with the C++ implementation of the new parquet geometry type:
https://github.com/apache/arrow/pull/45459/files

- Modified the updateBounds() method in BoundingBox.java to properly handle coordinates that cross the antimeridian
- Added logic to correctly calculate bounds when coordinates wrap around the globe
- Added comprehensive JavaDoc comments explaining JTS's limitations and how the new implementation addresses them
- Updated coordinate calculations to maintain proper spatial relationships for geometries crossing the ±180° longitude

### Are these changes tested?

Yes, tests have been added or updated to verify:

- Geometries that cross the antimeridian are properly bounded
- Existing functionality for non-wraparound cases continues to work as expected

Yes. Applications working with geospatial data near or crossing the antimeridian (±180° longitude) will now have properly calculated bounding boxes. This fixes potential issues with spatial queries, contains checks, and other geometric operations when working with global datasets that span the international date line.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
@@ -314,17 +314,14 @@ public class BoundingBox {
    */
   private void updateBounds(double minX, double maxX, double minY, double maxY) {
     if (!Double.isNaN(minX) && !Double.isNaN(maxX)) {
-      double newXMin = Math.min(xMin, minX);
-      double newXMax = Math.max(xMax, maxX);
-
       // Check if the update would create a wraparound condition
       // This should never happen with standard JTS geometry operations
-      if (isWraparound(newXMin, newXMax)) {
-        throw new ShouldNeverHappenException("Wraparound X is not supported by BoundingBox.update()");
+      if (isWraparound(minX, maxX) || isWraparound(xMin, xMax)) {
+        throw new ShouldNeverHappenException("Wraparound bounding boxes are not yet supported");
       }
 
-      xMin = newXMin;
-      xMax = newXMax;
+      xMin = Math.min(xMin, minX);
+      xMax = Math.max(xMax, maxX);
     }
 
     if (!Double.isNaN(minY) && !Double.isNaN(maxY)) {
@@ -353,7 +350,7 @@ public class BoundingBox {
    * The Parquet specification allows X bounds to be "wraparound" to allow for
    * more compact bounding boxes when a geometry happens to include components
    * on both sides of the antimeridian (e.g., the nation of Fiji). This function
-   * checks for that case (see GeoStatistics::lower_bound/upper_bound for more details).
+   * checks for that case.
    */
   public static boolean isWraparound(double xmin, double xmax) {
     return !Double.isInfinite(xmin - xmax) && xmin > xmax;
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.column.statistics.geospatial;
 
+import org.apache.parquet.ShouldNeverHappenException;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
@@ -169,7 +170,7 @@ public class BoundingBox {
    * @return true if the X dimension is empty, false otherwise.
    */
   public boolean isXEmpty() {
-    return Double.isInfinite(xMin) && Double.isInfinite(xMax);
+    return Double.isInfinite(xMin - xMax);
   }
 
   /**
@@ -178,7 +179,7 @@ public class BoundingBox {
    * @return true if the Y dimension is empty, false otherwise.
    */
   public boolean isYEmpty() {
-    return Double.isInfinite(yMin) && Double.isInfinite(yMax);
+    return Double.isInfinite(yMin - yMax);
   }
 
   /**
@@ -187,7 +188,7 @@ public class BoundingBox {
    * @return true if the Z dimension is empty, false otherwise.
    */
   public boolean isZEmpty() {
-    return Double.isInfinite(zMin) && Double.isInfinite(zMax);
+    return Double.isInfinite(zMin - zMax);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


zhangfengcdt:
@wgtmac @paleolimbot @jiayuasu This is a small fix to include the wraparound handling of bounding box statistics. Could you please take a look?

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
```diff
@@ -272,12 +298,33 @@ public void update(Geometry geometry) {
...
-      xMin = Math.min(xMin, minX);
-      xMax = Math.max(xMax, maxX);
+      double newXMin = Math.min(xMin, minX);
+      double newXMax = Math.max(xMax, maxX);
+
+      // Check if the update would create a wraparound condition
+      // This should never happen with standard JTS geometry operations
+      if (isWraparound(newXMin, newXMax)) {
+        throw new ShouldNeverHappenException("Wraparound X is not supported by BoundingBox.update()");
+      }
```
paleolimbot, line 324: Would this check be more clear like:

```java
if (isWraparound(minX, maxX) || isWraparound(xMin, xMax)) {
  SomeAppropriateException("Wraparound bounding boxes are not yet supported")
}
```

(i.e., you can't update bounds with wraparound bounds, nor can you update wraparound bounds with anything?)

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
```diff
@@ -302,6 +349,16 @@ public void reset() {
     valid = true;
   }
 
+  /**
+   * The Parquet specification allows X bounds to be "wraparound" to allow for
+   * more compact bounding boxes when a geometry happens to include components
+   * on both sides of the antimeridian (e.g., the nation of Fiji). This function
+   * checks for that case (see GeoStatistics::lower_bound/upper_bound for more details).
```
paleolimbot, line 356:
```suggestion
   * checks for that case.
```

paleolimbot (commented):
Thanks! I left a few comments here but generally looking good!

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.column.statistics.geospatial;
 
+import org.apache.parquet.ShouldNeverHappenException;
```
wgtmac, line 21: Ah, I didn't know this before this PR

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
```diff
@@ -218,16 +233,27 @@ public void merge(BoundingBox other) {
...
     }
 
-    this.xMin = Math.min(this.xMin, other.xMin);
-    this.xMax = Math.max(this.xMax, other.xMax);
+    // We don't yet support merging wraparound bounds.
+    // Rather than throw, we mark the X bounds as invalid.
+    if (isXWraparound() || other.isXWraparound()) {
+      // Mark X dimension as invalid by setting to NaN
+      xMin = Double.NaN;
+      xMax = Double.NaN;
```
wgtmac, line 241:
```suggestion
      valid = false;
      resetBBox();
```

Can we simply do this?

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
```diff
@@ -218,16 +233,27 @@ public void merge(BoundingBox other) {
       return;
     }
 
-    this.xMin = Math.min(this.xMin, other.xMin);
-    this.xMax = Math.max(this.xMax, other.xMax);
+    // We don't yet support merging wraparound bounds.
+    // Rather than throw, we mark the X bounds as invalid.
```
wgtmac, line 237:
```suggestion
    // We don't yet support merging wraparound bounds.
```

If you accept the below suggestion, this line should also be removed.

wgtmac (commented):
Thanks for the quick fix! It looks good to me if all comments from @paleolimbot have been addressed.

parquet-column/src/test/java/org/apache/parquet/column/statistics/geospatial/TestBoundingBox.java
```diff
@@ -681,4 +681,103 @@ public void testMergingRowGroupBoundingBoxes() {
...
+    BoundingBox wraparoundBox = new BoundingBox(170, 20, 10, 20, 0, 0, 0, 0);
+    Assert.assertTrue(wraparoundBox.isXWraparound());
+
+    // Edge case: equal bounds
+    BoundingBox equalBoundsBox = new BoundingBox(10, 10, 20, 20, 0, 0, 0, 0);
+    Assert.assertFalse(equalBoundsBox.isXWraparound());
+
+    // Test static method directly
+    Assert.assertTrue(BoundingBox.isWraparound(180, -180));
+    Assert.assertFalse(BoundingBox.isWraparound(-180, 180));
```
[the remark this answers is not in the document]
zhangfengcdt: I added the check on Infinity, -Infinity below.

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
zhangfengcdt: Thanks for the review @wgtmac ! 

For the reseting the bbox, my read from the C++ implementation below is that this wrap_x only impacts the x dimension, and it does not reset the whole bbox. This way, other dimensions are still namely "valid". I am good if we want to change this in both implementations. CC: @paleolimbot for opinion on this.
```
    // We don't yet support updating wraparound bounds. Rather than throw,
    // we just mark the X bounds as invalid such that they are not used.
    auto other_bounds = other.bounder_.Bounds();
    if (is_wraparound_x() || other.is_wraparound_x()) {
      other_bounds.min[0] = kNaN;
      other_bounds.max[0] = kNaN;
    }
```

The valid flag is set below by calling "valid = isXYValid();", so I think we are good here.

paleolimbot: I like the current code here...eventually the inside of that `if()` will actually update the bounds to finite values. I think either is correct, though (feel free to correct me if I missed something!).

paleolimbot (approved):
All good from my end as long as @wgtmac is comfortable with the NaN vs. invalid thread! Thanks!

parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/BoundingBox.java
wgtmac: I'm good with the current one because they are equivalent because setting x to NaN will finally discard the bbox from the file.

PR #3227, opened 2025-05-25
GH-3226: Improve performance of InternalParquetRecordReader (1%)

### LINKED ISSUE #3226: Improve performance of InternalParquetRecordReader (1%)

Profiling the load of a Parquet file with Java Mission Control, I've noticed that `InternalParquetRecordReader`  [LongStream](https://github.com/apache/parquet-java/blob/1f1e07bbf750fba228851c2d63470c3da5726831/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java#L323) consumes relevant amount of time.

This `LongStream` can be replaced with a simpler Long Iterator that iterates from 0 to `pages.getRowCount()`.

To measure the overhead I've created a test project that overwrites `InternalParquetRecordReader` implementation with a Long Iterator: https://github.com/jerolba/parquet-rowindexiterator 

The execution time is sensitive to the context of the JVM, but running the benchmark multiple times shows that LongStream is slower than LongIterator, between 1% and 4% depending on the run.

### Component(s)

_No response_

jerolba:
Avoid LongStream reading files and use an ad-hoc Long Iterator

### Rationale for this change

Profiling the load of a Parquet file with Java Mission Control, I've noticed that InternalParquetRecordReader [LongStream](https://github.com/apache/parquet-java/blob/1f1e07bbf750fba228851c2d63470c3da5726831/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java#L323) consumes relevant amount of time:

image

This LongStream can be replaced with a simpler Long Iterator that iterates from 0 to pages.getRowCount().

To measure the overhead I've created a test [project](https://github.com/jerolba/parquet-rowindexiterator) that overwrites `InternalParquetRecordReader` class using a Long Iterator (the same change than proposed in the PR)

The execution time is sensitive to the context of the JVM, but running the benchmark multiple times shows that LongStream is slower than LongIterator, between 1% and 4% depending on the run.

A new `LongIterator` that implements `PrimitiveIterator.OfLong` and replaces a `LongStream.range(0, pages.getRowCount()).iterator()`

### Are these changes tested?

Not directly, but it's covered by existing tests

No

Closes #3226

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordReader.java
@@ -30,7 +30,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.PrimitiveIterator;
 import java.util.Set;
-import java.util.stream.LongStream;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.ParquetReadOptions;
 import org.apache.parquet.column.page.PageReadStore;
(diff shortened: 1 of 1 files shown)


pan3793:
I ran the benchmark using different JDKs, it looks like your assertion is almost true in the lower versions of JDK (I tested 8, 17), but in new versions(I tested 21), long-stream is a little bit faster than long-iterator. So I believe that new JDKs have some optimization for cases you mentioned, if so, we should keep the code as-is for future-proofing.

PR #3234, opened 2025-06-03
GH-3233: Parquet CLI supports version command

### LINKED ISSUE #3233: Parquet CLI supports version command

Something like
```
$ parquet version
parquet-mr version 1.15.2 (build 859eac165b08f927fa14590c33bc5f476405fb68)
```

### Component(s)

CLI

pan3793:
### Rationale for this change

Follow the convention for the CLI tools to add a `version` command.

As title.

### Are these changes tested?

UT added, also tested manually

```
$ java -cp 'target/parquet-cli-1.16.0-SNAPSHOT.jar:target/dependency/*' org.apache.parquet.cli.Main version     
parquet-mr version 1.16.0-SNAPSHOT (build eb65987333a6bb2ae729b88eef0d6f941a7c1867)
```

Yes.

Closes #3233

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.cli;
+
+import com.beust.jcommander.*;
+import java.util.List;
+import org.apache.parquet.Version;
+import org.slf4j.Logger;
+
+@Parameters(commandDescription = "Print version of the Parquet CLI tool")
+public class ShowVersionCommand implements Command {
+
+  private final Logger console;
+
+  public ShowVersionCommand(Logger console) {
+    this.console = console;
+  }
+
+  @Override
+  public int run() {
+    console.info(Version.FULL_VERSION);
+    return 0;
+  }
+
+  @Override
+  public List<String> getExamples() {
+    return null;
+  }
+}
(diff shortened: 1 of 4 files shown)


wgtmac:
Isn't it already displayed by the `Created by` line from running the `meta` command?

pan3793:
@wgtmac it's a different thing, `Created by` represents the version of Parquet lib used to write the Parquet file, while this command prints the Parquet CLI tool's version.

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public int run() {
+    console.info(Version.FULL_VERSION);
+    return 0;
+  }
+
+  @Override
+  public List<String> getExamples() {
+    return null;
```
nandorKollar: nit: probably doesn't matter much, but other commands without examples return empty array list instead of null (`Lists.newArrayList()` or maybe even better `Collections.EMPTY_LIST`)

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.cli;
+
+import com.beust.jcommander.*;
```
nandorKollar: nit: I think we usually avoid * imports

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+import org.apache.parquet.Version;
+import org.slf4j.Logger;
+
+@Parameters(commandDescription = "Print version of the Parquet CLI tool")
+public class ShowVersionCommand implements Command {
```
nandorKollar: Why didn't you inherit from BaseCommand? I believe console is already present there.

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
pan3793: Thanks for taking a look, I copied it from `Help#getExamples`. Anyway, avoiding NULL propagation is generally a good idea, let me update both.

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
pan3793: because the "version" is quite a simple command that does not require accessing a parquet file, so inheriting from BaseCommand does not help, but I can change that for consistency

parquet-cli/src/main/java/org/apache/parquet/cli/ShowVersionCommand.java
pan3793: addressed

wgtmac (approved):

PR #3236, opened 2025-06-04
GH-3235: Row count limit for each row group

### LINKED ISSUE #3235: Row count limit for each row group

In some compute engines, like Spark, the row group is the minimally splittable unit for scanning tasks. Currently, only the row group size is configurable (via `parquet.block.size`) when writing Parquet files. In some cases, especially for fewer columns with many duplicated values, one row group may have tons of records, thus causes extremely bad performance on downstream Spark queries.

Image

I propose to make the row count limit for each row group configurable. [ORC-1172](https://issues.apache.org/jira/browse/ORC-1172) also has a similar configuration.

### Component(s)

Core

pan3793:
### Rationale for this change

Similar to ORC-1172, see background at #3235

A new configuration `parquet.block.row.count.limit` is added.

### Are these changes tested?

UT is added.

Also verified by integrating with Spark, for example, to set the row count threshold to 100000000 for each row group in Spark
```
spark-sql --conf spark.hadoop.parquet.block.row.count.limit=100000000
```

Yes, a new conf is added.

Closes #3235

### DIFF
--- a/parquet-hadoop/README.md
+++ b/parquet-hadoop/README.md
@@ -41,68 +41,68 @@ ParquetOutputFormat.setBlockSize(writeJob, 1024);
 
 ## Class: ParquetOutputFormat
 
-**Property:** `parquet.summary.metadata.level`  
-**Description:** Write summary files in the same directory as parquet files.  
-If this property is set to `all`, write both summary file with row group info to _metadata and summary file without row group info to _common_metadata.  
-If it is `common_only`, write only the summary file without the row group info to _common_metadata.  
-If it is `none`, don't write summary files.  
+**Property:** `parquet.summary.metadata.level`
+**Description:** Write summary files in the same directory as parquet files.
+If this property is set to `all`, write both summary file with row group info to _metadata and summary file without row group info to _common_metadata.
+If it is `common_only`, write only the summary file without the row group info to _common_metadata.
+If it is `none`, don't write summary files.
 **Default value:** `all`
 
 ---
 
-**Property:** `parquet.enable.summary-metadata`  
-**Description:** This property is deprecated, use `parquet.summary.metadata.level` instead.  
-If it is `true`, it similar to `parquet.summary.metadata.level` with `all`. If it is `false`, it is similar to `NONE`.   
-**Default value:** `true`  
+**Property:** `parquet.enable.summary-metadata`
+**Description:** This property is deprecated, use `parquet.summary.metadata.level` instead.
+If it is `true`, it similar to `parquet.summary.metadata.level` with `all`. If it is `false`, it is similar to `NONE`.
+**Default value:** `true`
 
 ---
 
-**Property:** `parquet.block.size`  
+**Property:** `parquet.block.size`
 **Description:** The block size in bytes. This property depends on the file system:
 
-- If the file system (FS) used supports blocks like HDFS, the block size will be the maximum between the default block size of FS and this property. And the row group size will be equal to this property.  
-  - block_size = max(default_fs_block_size, parquet.block.size)  
+- If the file system (FS) used supports blocks like HDFS, the block size will be the maximum between the default block size of FS and this property. And the row group size will be equal to this property.
+  - block_size = max(default_fs_block_size, parquet.block.size)
   - row_group_size = parquet.block.size`
 
 - If the file system used doesn't support blocks, then this property will define the row group size.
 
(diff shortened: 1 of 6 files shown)


parquet-hadoop/README.md
```diff
@@ -41,68 +41,68 @@ ParquetOutputFormat.setBlockSize(writeJob, 1024);
 
 ## Class: ParquetOutputFormat
 
-**Property:** `parquet.summary.metadata.level`  
-**Description:** Write summary files in the same directory as parquet files.  
-If this property is set to `all`, write both summary file with row group info to _metadata and summary file without row group info to _common_metadata.  
-If it is `common_only`, write only the summary file without the row group info to _common_metadata.  
-If it is `none`, don't write summary files.  
+**Property:** `parquet.summary.metadata.level`
```
pan3793: auto reformated by IDEA, let me revert that ...

parquet-hadoop/README.md
```diff
@@ -266,12 +266,16 @@ conf.set("parquet.bloom.filter.fpp#column.path", 0.02)
...
 
-
 **Property:** `parquet.decrypt.off-heap.buffer.enabled`  
 **Description:** Whether to use direct buffers to decrypt encrypted files. This should be set to 
 true if the reader is using a `DirectByteBufferAllocator`
 **Default value:** `false`
 
+---
+
+**Property:** `parquet.block.row.count.limit`
```
pan3793, line 276: since there is an existing `parquet.block.size`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -48,8 +48,8 @@ class InternalParquetRecordWriter<T> {
   private final WriteSupport<T> writeSupport;
   private final MessageType schema;
   private final Map<String, String> extraMetaData;
-  private final long rowGroupSize;
```
pan3793, line 51: unused field

pan3793:
cc @wgtmac and @nandorKollar

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -166,9 +166,16 @@ public long getDataSize() {
   }
 
   private void checkBlockSizeReached() throws IOException {
-    if (recordCount
-        >= recordCountForNextMemCheck) { // checking the memory size is relatively expensive, so let's not do it
-      // for every record.
+    if (recordCount >= rowGroupRecordCountThreshold) {
+      LOG.debug("record count reaches threshold: flushing {} records to disk.", recordCount);
```
pan3793, line 170: Maybe better to escalate the log level to INFO? It should not be noisy.

pan3793:
Velox Parquet write also has such a configuration

https://github.com/facebookincubator/velox/blob/d02351911f5f8ca2b00242d6010859baaf513813/velox/dwio/parquet/writer/Writer.h#L47-L51

pan3793:
kindly ping @wgtmac @gszadovszky, could you please take a look? thank you in advance

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
gszadovszky: Maybe it would not be noisy to raise the level to INFO but what purpose does it serve? If we want to answer why a Parquet file looks like how it does (the number/size of row groups etc.), the logs of the file creation are probably long gone.

pan3793: For a typical Hadoop YARN cluster, which serves Spark workloads, application logs are preserved for a few days by collecting and aggregating to HDFS. Anyway, it's not a big deal, we can always use parquet-cli to analyze the suspicious Parquet files :)

gszadovszky: Yeah, what I was thinking of, you write a Parquet file, and then, you read it somewhere else months/years later. You won't have the related logs for sure.

It is a very separate topic, but it would be a good idea to write these properties directly into the footer so if we have any issues with a file, at least we would know, how it was created.

PR #3243, opened 2025-06-12
GH-3242: Emit and Read min/max statistics for int96 timestamp columns

### LINKED ISSUE #3242: Emit and Read min/max statistics for int96 timestamp columns

Parquet-java does not emit or read stats for int96 timestamp columns. Since int96 is used as the default timestamp in spark, this limits a lot of optimization opportunities. Engines like [Photon](https://www.databricks.com/product/photon) populate the statistics for the int96 timestamps correctly. So parquet-java coud also emit the statistics, and also allow reading these statistics from known good writers. 

### Component(s)

_No response_

rahulketch:
### Rationale for this change

Parquet-java does not emit or read stats for int96 timestamp columns. Since int96 is used as the default timestamp in spark, this limits a lot of optimization opportunities. Engines like [Photon](https://www.databricks.com/product/photon) populate the statistics for the int96 timestamps correctly. So parquet-java can also emit the statistics, and also allow reading these statistics from known good writers.

1. Defining column ordering for int96
2. Implementation of comparator for int96
3. A flag `parquet.read.int96stats.enabled` to control if the stats are read. It is defaulted to true
4. `ValidInt96Stats`: Reads stats from known good writers. Currently including:
 a. `parquet-mr 1.15.0+`
 b. `photon`
5. Fixing tests which fail due to this change

### Are these changes tested?
1. Added more unit tests for the new behaviour

1. `toParquetStatistics` and `fromParquetStatistics` are no longer static functions. It doesn't look like there is a good reason for these functions to be static. 
2. Signatures of some of the constructors of `ParquetMetadataConverter` is now different due to an added parameter.  

 Closes #3242


emkornfield:
It looks like Alkis started a discussion on the ML, but we probably want to come to a consensus and update https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L1079 first before merging this.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -206,6 +206,31 @@ public String toString() {
...
+  static final PrimitiveComparator<Binary> BINARY_AS_INT_96_COMPARATOR = new BinaryComparator() {
+    @Override
+    int compareBinary(Binary b1, Binary b2) {
+      ByteBuffer bb1 = b1.toByteBuffer();
+      ByteBuffer bb2 = b2.toByteBuffer();
+      bb1.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+      bb2.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+      int jd1 = bb1.getInt(8);
+      int jd2 = bb2.getInt(8);
+      if (jd1 != jd2) return Integer.compareUnsigned(jd1, jd2) < 0 ? -1 : 1;
```
emkornfield, line 224: shouldn't this be signed comparison?  Year 0001-01-01 is in the wild, and I think that would be a negative value for days since julien epoch?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -206,6 +206,31 @@ public String toString() {
...
+  /*
+   * This comparator is for comparing two timestamps represented as int96 binary.
+   */
+  static final PrimitiveComparator<Binary> BINARY_AS_INT_96_COMPARATOR = new BinaryComparator() {
+    @Override
+    int compareBinary(Binary b1, Binary b2) {
+      ByteBuffer bb1 = b1.toByteBuffer();
+      ByteBuffer bb2 = b2.toByteBuffer();
+      bb1.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+      bb2.order(java.nio.ByteOrder.LITTLE_ENDIAN);
```
emkornfield, line 221: this logic should probably be commented.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
emkornfield: Never mind, I think this is actually counting from something pre-year 1.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -206,6 +206,31 @@ public String toString() {
         }
       };
 
+  /*
+   * This comparator is for comparing two timestamps represented as int96 binary.
+   */
+  static final PrimitiveComparator<Binary> BINARY_AS_INT_96_COMPARATOR = new BinaryComparator() {
```
rdblue: Shouldn't this be `AS_INT96_TIMESTAMP`? This is not a comparison that would work for int96.

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
```diff
@@ -0,0 +1,75 @@
...
+  public static boolean hasValidInt96Stats(String createdBy) {
+    if (Strings.isNullOrEmpty(createdBy)) {
+      warnOnce("Cannot verify INT96 statistics because created_by is null or empty");
+      return false;
+    }
+
+    try {
+      ParsedVersion version = VersionParser.parse(createdBy);
+      if ("parquet-mr".equals(version.application)) {
+        return true;
```
rdblue: Wouldn't this cause all previous stats to be valid? Do we know that is the case?

rahulketch: Previously parquet-mr does not emit stats for int96 columns. After this PR is merged int96 stats will be emitted so they're safe to read.

rdblue: Were int96 stats ever emitted? I don't remember.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
alkis: INT96 is a physical type and there was never a converted/logical type for it. Ergo it is implicitly only a timestamp.

Adding `AS_INT96_TIMESTAMP` is redundant but perhaps leaves some context for the reader on what we are comparing here.

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
alkis: Not from parquet-mr.

rahulketch: I added a check for parquet-mr version > 1.15.0

rahulketch:
After this change, `toParquetStatistics` and `fromParquetStatistics` in `ParquetMetadataConverter` are no longer static functions. This breaks a test on compatibility (see [here](https://github.com/databricks-eng/runtime/pull/151256#discussion_r2177379031)). @emkornfield @rdblue: Do you have advice on how to move forward with those changes?

rahulketch:
@emkornfield @rdblue Given the latest [message](https://lists.apache.org/thread/6t9fr6v602zwt0tw22bqwg81f1ny9ncj) in the mailing thread, can we merge this? If so, could you help me with addressing this [issue](https://github.com/apache/parquet-java/pull/3243#issuecomment-3023677407)

rahulketch:
> > After this change, `toParquetStatistics` and `fromParquetStatistics` in `ParquetMetadataConverter` are no longer static functions. This breaks a test on compatibility (see [here](https://github.com/databricks-eng/runtime/pull/151256#discussion_r2177379031)). @emkornfield @rdblue: Do you have advice on how to move forward with those changes?
> 
> @wgtmac : could you help with this?

@wgtmac : Could you help with this, or suggest who could help out? Thanks!

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -206,6 +206,34 @@ public String toString() {
         }
       };
 
+  /*
+   * This comparator is for comparing two timestamps represented as int96 binary.
+   * It is a two level comparison.
+   * Days (last 4 bytes compared as unsigned Little endian int32),
```
wgtmac:
```suggestion
   * Days (last 4 bytes compared as unsigned little endian int32),
```

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -206,6 +206,34 @@ public String toString() {
...
+   * Nanoseconds (first 8 bytes compared as unsigned little endian int64)
+   */
+  static final PrimitiveComparator<Binary> BINARY_AS_INT96_TIMESTAMP_COMPARATOR = new BinaryComparator() {
+    @Override
+    int compareBinary(Binary b1, Binary b2) {
+      ByteBuffer bb1 = b1.toByteBuffer().slice();
```
wgtmac, line 218: Should we check if their lengths are exactly 12 before anything? I recall that `BINARY_AS_FLOAT16_COMPARATOR` did this.

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
```diff
@@ -0,0 +1,75 @@
...
+  public static boolean hasValidInt96Stats(String createdBy) {
+    if (Strings.isNullOrEmpty(createdBy)) {
+      warnOnce("Cannot verify INT96 statistics because created_by is null or empty");
+      return false;
+    }
+
+    try {
+      ParsedVersion version = VersionParser.parse(createdBy);
+      if ("parquet-mr".equals(version.application)) {
+        return version.version != null && version.version.compareTo("1.15.0") > 0;
```
wgtmac:
```suggestion
        return version.version != null && version.version.compareTo("1.16.0") > 0;
```

1.16 has been released.

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public static boolean hasValidInt96Stats(String createdBy) {
+    if (Strings.isNullOrEmpty(createdBy)) {
+      warnOnce("Cannot verify INT96 statistics because created_by is null or empty");
+      return false;
+    }
+
+    try {
+      ParsedVersion version = VersionParser.parse(createdBy);
+      if ("parquet-mr".equals(version.application)) {
```
wgtmac, line 52: I recall that @rdblue has an opinion in maintaining an allow-list here.

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Decides if the statistics from a file created by createdBy (the created_by field from parquet format)
+   * should be trusted for INT96 columns.
+   *
+   * @param createdBy  the created-by string from a file footer
+   * @return true if the statistics are valid and can be trusted, false otherwise
+   */
+  public static boolean hasValidInt96Stats(String createdBy) {
+    if (Strings.isNullOrEmpty(createdBy)) {
+      warnOnce("Cannot verify INT96 statistics because created_by is null or empty");
```
wgtmac, line 46: What is the convention for it? `created_by`, `createdBy` or `created-by`?

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
```diff
@@ -0,0 +1,75 @@
...
+        return true;
+      }
+    } catch (RuntimeException | VersionParseException e) {
+      warnParseErrorOnce(createdBy, e);
+    }
+    return false;
+  }
+
+  private static void warnParseErrorOnce(String createdBy, Throwable e) {
+    if (!alreadyLogged.getAndSet(true)) {
```
wgtmac, line 65: If it is running in a service, we are unable to log warnings for different jobs.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -762,12 +772,12 @@ public List<PageEncodingStats> convertEncodingStats(EncodingStats stats) {
     return formatStats;
   }
 
-  public static Statistics toParquetStatistics(org.apache.parquet.column.statistics.Statistics stats) {
-    return toParquetStatistics(stats, ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH);
+  public Statistics toParquetStatistics(String createdBy, org.apache.parquet.column.statistics.Statistics stats) {
```
wgtmac, line 775: Isn't it a breaking change?

rahulketch: I am not sure of a better way to achieve this. We need to:
1. Know created-by in order to have an allow-list for valid readers
2. Make the function non-static so we can respect the flag which disables reading the stats.

Any ideas for this?

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
rahulketch: The [spec](https://parquet.apache.org/docs/file-format/metadata/) says `created_by`, so I went with that

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
rahulketch: I took the logic from [CorruptStatistics.java](https://github.com/apache/parquet-java/blob/master/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java#L106-L110). Would you recommend one of the following instead?
1. Always log
2. Never log
I am happy to change it.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
wgtmac: Should we just follow the existing pattern of `toParquetStatistics` to introduce yet another overload?

```
      org.apache.parquet.column.statistics.Statistics stats, int truncateLength, bool convertInt96Stats)
```

@gszadovszky WDYT?

parquet-column/src/main/java/org/apache/parquet/ValidInt96Stats.java
wgtmac: I'm fine to keep this as-is. It is better than nothing or a lot of logs.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
gszadovszky: I agree with @wgtmac.

Or, we can deprecate the static methods (marking them for removal in 2.0) and create non-static ones to be used from now on.

Issue #3249, opened 2025-07-11
Data issue after the optimization of bloom_filter_length


wangyum:
### Describe the bug, including details regarding any error messages, version, and platform.

Parquet version: From 1.14.0 to 1.15.2

There is a data issue after this improvement. The result is empty after this patch. My query is:
```sql
CREATE TEMPORARY VIEW parquetTable
USING org.apache.spark.sql.parquet
OPTIONS (
  path "viewfs://cluster/path/to/file/00000-62-eff62af9-bea8-4eb6-b9c9-c3ee29fd795e-0-00001.parquet"
);

SELECT *
FROM parquetTable
WHERE JNL_DEDUP_ID = 'PG.PG_CHARGE_ID.25030465176.EBAY_MM_PPFMM_PP_CAPTURE'
AND ACCT_ID = 4920012120302 AND SEQ_NUM = 734343586;
```
We build bloom filter on the column of JNL_DEDUP_ID using Parquet 1.13.1(Read data using Parquet 1.15.2), and it's data type is string type.

### Component(s)

Core

PR #3250, opened 2025-07-11
GH-3249: Fix incorrect Bloom filter data when reading from ByteArrayInputStream by using readFully()

wangyum:
### Rationale for this change

When reading Bloom filter data from files with older versions(< Parquet 1.13), the code uses `in.read(bitset)` to read the bitset data. However, `InputStream.read(byte[])` doesn't guarantee reading all requested bytes in a single call - it may read fewer bytes than the buffer size and the remaining portion of the buffer stays uninitialized.

This can lead to incorrect Bloom filter behavior as parts of the bitset might be missing or contain zeros instead of the actual data.

This PR modifies the logic to properly ensure all bytes are read from the input stream:

For older file versions (negative bloomFilterLength), we continue using f.readFully(bitset)
For newer file versions (positive bloomFilterLength), we still use `in.read(bitset)`.

### Are these changes tested?

Manual testing.

No.


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1667,7 +1667,14 @@ public BloomFilter readBloomFilter(ColumnChunkMetaData meta) throws IOException
     byte[] bitset;
     if (null == bloomFilterDecryptor) {
       bitset = new byte[numBytes];
-      in.read(bitset);
+      // For negative bloomFilterLength (files from older versions), use readFully() instead of read().
+      // readFully() guarantees reading exactly numBytes bytes, while read() may read fewer bytes in a single
+      // call. This ensures the entire bitset is properly loaded.
+      if (bloomFilterLength < 0) {
```
wgtmac, line 1673: Why not using `readFully` for all cases? Isn't it a bug if a positive bloomFilterLength is not respected?

wangyum: We already consume some data if bloomFilterLength > 0:
https://github.com/apache/parquet-java/blob/0e6eefba0889132c8554cec0ed437dcd3efaaade/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L1636-L1640

So we can not use readFully here, and it will throw `java.io.EOFException` if using `readFully` for all cases:
https://github.com/apache/parquet-java/blob/eface4db2e961a0174d4d1cd809134262d8c4c76/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInteropBloomFilter.java#L131

wgtmac: Thanks for the clarification!

wgtmac:
The community is planning the next release shortly :)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1667,7 +1667,14 @@ public BloomFilter readBloomFilter(ColumnChunkMetaData meta) throws IOException
     byte[] bitset;
     if (null == bloomFilterDecryptor) {
       bitset = new byte[numBytes];
-      in.read(bitset);
+      // For negative bloomFilterLength (files from older versions), use readFully() instead of read().
```
pan3793, line 1670: > files from older versions

could this be explained in more accurate words?

wgtmac:
It is 1.16.0. @manuzhang

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
wangyum: <= 1.13.

PR #3258, opened 2025-07-27
Test Variant implementation with external test cases

aihuaxu:
### Rationale for this change
This is to test out variant implementation in Parquet java against the test cases generated in https://github.com/apache/parquet-testing/pull/91 (test cases from Iceberg) and https://github.com/apache/parquet-testing/pull/94 (test cases from GO language). 

Variant Implementation Compatibility

Overall, the results demonstrate that the Variant implementation is compatible with Iceberg.

Notable observations:

- Missing logical type annotation in Parquet files:
Since Parquet files do not yet include the variant logical type annotation (pending this test for release of Parquet-java with the annotation), a temporary workaround was added to the code.

- Decimal type inconsistency in Iceberg:
Encountered a known issue with decimal types in Iceberg ([iceberg#13692](https://github.com/apache/iceberg/pull/13692)). Regenerated test data accordingly. The Variant implementation in Parquet-java handles decimal encoding correctly.

- Error message discrepancies:
While error cases throw different messages compared to Iceberg, they fail as expected with appropriate exceptions.

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -395,9 +395,7 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
         }
         return newStringConverter(schema, model, parent, validator);
       case RECORD:
-        if (type.getName().equals("var")
-            || type.getLogicalTypeAnnotation()
-                instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
+        if (type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
           return new AvroVariantConverter(parent, type.asGroupType(), schema, model);
         } else {
           return new AvroRecordConverter(parent, type.asGroupType(), schema, model, validator);
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -395,7 +395,7 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
         }
         return newStringConverter(schema, model, parent, validator);
       case RECORD:
-        if (type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
+        if (type.getName().equals("var") || type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
           return new AvroVariantConverter(parent, type.asGroupType(), schema, model);
         } else {
           return new AvroRecordConverter(parent, type.asGroupType(), schema, model, validator);
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
@@ -395,7 +395,9 @@ class AvroRecordConverter<T> extends AvroConverters.AvroGroupConverter {
         }
         return newStringConverter(schema, model, parent, validator);
       case RECORD:
-        if (type.getName().equals("var") || type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
+        if (type.getName().equals("var")
+            || type.getLogicalTypeAnnotation()
+                instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
           return new AvroVariantConverter(parent, type.asGroupType(), schema, model);
         } else {
           return new AvroRecordConverter(parent, type.asGroupType(), schema, model, validator);
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -470,8 +470,11 @@ public class AvroSchemaConverter {
               @Override
               public Optional<Schema> visit(
                   LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
-                return of(
-                    convertFields(parquetGroupType.getName(), parquetGroupType.getFields(), names));
+                String name = parquetGroupType.getName();
+                List<Schema.Field> fields = new ArrayList<>();
+                fields.add(new Schema.Field("metadata", Schema.create(Schema.Type.BYTES)));
+                fields.add(new Schema.Field("value", Schema.create(Schema.Type.BYTES)));
+                return of(Schema.createRecord(name, null, namespace(name, names), false, fields));
               }
             })
             .orElseThrow(
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -470,11 +470,8 @@ public class AvroSchemaConverter {
               @Override
               public Optional<Schema> visit(
                   LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
-                String name = parquetGroupType.getName();
-                List<Schema.Field> fields = new ArrayList<>();
-                fields.add(new Schema.Field("metadata", Schema.create(Schema.Type.BYTES)));
-                fields.add(new Schema.Field("value", Schema.create(Schema.Type.BYTES)));
-                return of(Schema.createRecord(name, null, namespace(name, names), false, fields));
+                return of(
+                    convertFields(parquetGroupType.getName(), parquetGroupType.getFields(), names));
               }
             })
             .orElseThrow(
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
@@ -21,7 +21,6 @@ package org.apache.parquet.avro;
 import java.nio.ByteBuffer;
 import java.util.function.Consumer;
 import org.apache.avro.Schema;
-import org.apache.avro.SchemaBuilder;
 import org.apache.avro.generic.GenericData;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.io.api.Converter;
@@ -35,20 +34,11 @@ import org.apache.parquet.variant.VariantConverters;
  * Converter for Variant values.
  */
 class AvroVariantConverter extends GroupConverter implements VariantConverters.ParentConverter<VariantBuilder> {
-  private static final Schema VARIANT_SCHEMA = SchemaBuilder.record("VariantRecord")
-      .fields()
-      .name("metadata")
-      .type()
-      .bytesType()
-      .noDefault()
-      .name("value")
-      .type()
-      .bytesType()
-      .noDefault()
-      .endRecord();
-
   private final ParentValueContainer parent;
+  private final Schema avroSchema;
   private final GenericData model;
+  private final int metadataPos;
+  private final int valuePos;
   private final GroupConverter wrappedConverter;
 
   private VariantBuilder builder = null;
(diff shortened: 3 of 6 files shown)


parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
```diff
@@ -77,10 +78,12 @@ public void end() {
 
     builder.appendNullIfEmpty();
 
-    Object record = model.newRecord(null, avroSchema);
-    model.setField(record, "metadata", metadataPos, metadata.getEncodedBuffer());
-    model.setField(record, "value", valuePos, builder.encodedValue());
-    parent.add(record);
+    Variant variant = builder.build();
+    parent.add(variant);
```
aihuaxu: @cashmand  Would this convert to Variant object here?

cashmand: I'm not sure I understand the question. Are you saying that we should be setting the Variant object directly here, rather than a `(metadata, value)` record? That might be reasonable, my understanding of Avro, and how it's meant to be used, is pretty weak. cc @rdblue

aihuaxu: That's right. Seems we should return Variant rather than `(metadata, value)` record.

aihuaxu: I want to confirm with you and I will can make the change.

cashmand: It sounds fine to me, I'd suggest getting @rdblue to approve. Maybe we want to do something similar on the write side, where I think it also currently expects a `metadata, value` pair from Avro rather than a Variant object.

aihuaxu: Confirmed with @rdblue  offline and it should return a variant. Let me work on the fix and also address write side as well.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -395,7 +395,7 @@ private static Converter newConverter(
         }
         return newStringConverter(schema, model, parent, validator);
       case RECORD:
-        if (type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
+        if (type.getName().equals("var") || type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
```
rdblue: I don't think that this PR can be merged until this is no longer required.

aihuaxu: That's right. I workaround to test the compatibility for now.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
aihuaxu: By reading the mapping in https://iceberg.apache.org/spec/#avro, actually we are expecting a record of metadata and value in Avro for variant. 

Looks like that we just need to update the test rather than making the code. @rdblue  and @cashmand 

image

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
```diff
@@ -77,9 +84,9 @@ public void end() {
 
     builder.appendNullIfEmpty();
 
-    Object record = model.newRecord(null, avroSchema);
-    model.setField(record, "metadata", metadataPos, metadata.getEncodedBuffer());
-    model.setField(record, "value", valuePos, builder.encodedValue());
+    Object record = model.newRecord(null, VARIANT_SCHEMA);
```
aihuaxu: I think we need this change that we will produce a fixed schema of value and metadata rather than reading the original avro schema.

Also, we also fix the issue that value field may be missing from the schema since that is allowed, i.e., typed_value exists but value doesn't. We don't read value field for the position and the output schema should be fixed.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -470,11 +470,8 @@ public Optional<Schema> visit(
               @Override
               public Optional<Schema> visit(
                   LogicalTypeAnnotation.VariantLogicalTypeAnnotation variantLogicalType) {
-                String name = parquetGroupType.getName();
-                List<Schema.Field> fields = new ArrayList<>();
-                fields.add(new Schema.Field("metadata", Schema.create(Schema.Type.BYTES)));
-                fields.add(new Schema.Field("value", Schema.create(Schema.Type.BYTES)));
-                return of(Schema.createRecord(name, null, namespace(name, names), false, fields));
+                return of(
```
aihuaxu: @rdblue  and @cashmand Here is another issue that I have noticed:

If the messageType is as follows, do we generate the schema including everything like typed_value or we just include metadata and value? In Iceberg code, right now we are expecting full schema, e.g., we will set the value to `typed_value` field.

```
GroupType variantType =
        Types.buildGroup(Type.Repetition.REQUIRED)
            .as(LogicalTypeAnnotation.variantType((byte)1))
            .id(2)
            .required(PrimitiveTypeName.BINARY)
            .named("metadata")
            .addField(objectFields)
            .named("var");
    MessageType parquetSchema = parquetSchema(variantType);
    GenericRecord recordA = record(fieldA, Map.of("value", serialize(Variants.of((short) 1234))));
    GenericRecord recordB = record(fieldB, Map.of("typed_value", "iceberg"));
```


```
message table {
  required int32 id;
  optional group var (VARIANT(1)) {
    required binary metadata;
    optional binary value;
    optional group typed_value {
      required group a {
        optional binary value;
        optional group typed_value (LIST) {
          repeated group list {
            required group element {
              optional binary value;
              optional binary typed_value (STRING);
            }
          }
        }
      }
    }
  }
}
```

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
```diff
@@ -34,21 +35,27 @@
...
+      .type()
+      .bytesType()
+      .noDefault()
+      .endRecord();
+
   private final ParentValueContainer parent;
-  private final Schema avroSchema;
   private final GenericData model;
-  private final int metadataPos;
-  private final int valuePos;
```
rdblue: Why are all of these changes needed? I think this should continue to use the schema that was passed in rather than replacing it with a different one.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
rdblue: We can't change the expected Avro schema. We can reject it, but the contract is to use the schema that was passed in. I think the earlier code is correct.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
rdblue: Is that the behavior in production code or test code?

parquet-avro/src/test/java/org/apache/parquet/variant/TestVariantReadsFromFile.java
```diff
@@ -0,0 +1,278 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Stream;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.parquet.Preconditions;
+import org.apache.parquet.avro.AvroParquetReader;
+import org.apache.parquet.io.InputFile;
+import org.apache.parquet.io.LocalInputFile;
+import org.assertj.core.api.Assertions;
```
rdblue, line 47: This uses JUnit 5. Does that work in Parquet?

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
aihuaxu: It's in the test code that we use AvroSchemaConverter to create the schema in https://github.com/apache/iceberg/blob/main/parquet/src/test/java/org/apache/iceberg/parquet/TestVariantReaders.java#L1597. With the current logic, if we pass in a shredded variant type like 

```
GroupType variantType =
        Types.buildGroup(Type.Repetition.REQUIRED)
            .as(LogicalTypeAnnotation.variantType((byte)1))
            .id(2)
            .required(PrimitiveTypeName.BINARY)
            .named("metadata")
            .addField(objectFields)
            .named("var");
```

We are converted to a record of `value` and `metadata`, then  we cannot set the value for `typed_value`.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
aihuaxu: Here what we are passing in the shredded schema which may/may not have `value` or `typed_value` field. 

But we want to generate a variant which maps to a record of `value` and `metadata` fields, right?

image

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
aihuaxu: We are passing in a schema which may not have `value` or `typed_value` fields. As I understand, we want to return a record of `value` and `metadata` fields for a variant in Avro.

parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java
cashmand: I think the intent was the parquet schema should be the shredding schema, but the Avro schema provided for Variant should always be a record of `(value, metadata)`, even if the parquet schema doesn't contain `value`. So as @rdblue said, we should reject the schema if that isn't what was provided for a Variant column.

aihuaxu: @cashmand  I followed up with Ryan and I got more context around Avro.

For some reader path, seems we are passing AvroSchema as a shredded schema which should not happen. Let me checkout more how that happened. The AvroSchema here should always be a record of (value, metadata).

parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java
```diff
@@ -396,7 +408,7 @@ private static Converter newConverter(
         return newStringConverter(schema, model, parent, validator);
       case RECORD:
         if (type.getLogicalTypeAnnotation() instanceof LogicalTypeAnnotation.VariantLogicalTypeAnnotation) {
-          return new AvroVariantConverter(parent, type.asGroupType(), schema, model);
+          return new AvroVariantConverter(parent, type.asGroupType(), VARIANT_SCHEMA, model);
```
aihuaxu: We are reading avro schema from the `parquet.avro.schema` of parquet file . If this schema doesn't have value field, then in [AvroVariantConverter.java#L51](https://github.com/apache/parquet-java/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroVariantConverter.java#L51),  it will error out.  
The schema in `parquet.avro.schema` seems to be correct to represent the write schema and we  we need to make the change to pass in record of value + metadata for variant logical type.
```
aixu@K7YJWY4PK6 variant % parquet2 meta case-111.parquet
File path: case-111.parquet
Created by: parquet-mr version 1.16.0-SNAPSHOT (build ee34713e4d906d61f95d2b09145945638b2e2296)
Properties:
 parquet.avro.schema: {"type":"record","name":"table","fields":[{"name":"id","type":"int"},{"name":"var","type":["null",{"type":"record","name":"var","fields":[{"name":"metadata","type":"bytes"},{"name":"value","type":["null","bytes"],"default":null},{"name":"typed_value","type":["null","string"],"default":null}]}],"default":null}]}
  writer.model.name: avro
Schema:
message table {
 required int32 id = 1;
 optional group var (VARIANT(1)) = 2 {
  required binary metadata;
  optional binary value;
  optional binary typed_value (STRING);
 }
}
```

Let me know if I misunderstand `parquet.avro.schema`.

parquet-avro/src/test/java/org/apache/parquet/variant/TestVariantReadsFromFile.java
aihuaxu: I have added the dependency in pom.xml and seems to be working.

PR #3259, opened 2025-07-29
HG-3239: Improve ByteBufferReadable detection in HadoopStream

wangyum:
This PR improves the `ByteBufferReadable` detection in `HadoopStreams` by checking for the "in:readbytebuffer" capability using the `hasCapability` method instead of `hasCapabilities ` in `FSDataInputStream`.


wgtmac (approved):
Looks good to me as we are already on Hadoop 3.3.0.

Do you have any concern? @gszadovszky @Fokko

nandorKollar (approved):

steveloughran (approved):
always good to get rid of needless reflection.

I think I noted in some other jira that the reflection code was broken too, it was looking for hasCapabilities when it needed hasCapability , so it'd never get found. This PR fixes that with compile-time checking

wgtmac:
Merged. Thanks everyone!

PR #3260, opened 2025-07-31
GH-3253: Apply ServicesResourceTransformer to parquet-jackson

### LINKED ISSUE #3253: jackson-parquet doesn't shade ServiceLoader discovery files

If you peek inside `parquet-jackson` jar you can easily find under `META-INF/services/` a couple of services like `META-INF/services/com.fasterxml.jackson.core.JsonFactory`. Both the offered service name and implementing classes inside aren't shaded.

This doesn't seem to break the code itself, but does raise a couple of concerns:
1. If there is a need in offering services in shaded Jackson, those should be fixed.
2. If there is no such need those can be skipped not to interfere with non-shaded Jackson (*however the effect here is minimal, besides just duplicating an entry*).

For example, AWS SDK for Java v2 rewrites their services file from Jackson core.

Checked on 1.13.1 and 1.15.2.

### Component(s)

Build

pan3793:
### Rationale for this change

Fix #3253

As the title.

### Are these changes tested?

Before
image

After
Xnip2025-07-31_17-08-35

No.

Closes #3253

### DIFF
--- a/parquet-jackson/pom.xml
+++ b/parquet-jackson/pom.xml
@@ -122,6 +122,9 @@
                   <shadedPattern>META-INF.versions.22.${shade.prefix}.${jackson.package}</shadedPattern>
                 </relocation>
               </relocations>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+              </transformers>
             </configuration>
           </execution>
         </executions>


pan3793:
cc @LDVSOFT who reports this issue

wgtmac:
I'm not familiar with this. Could you help review this? @gszadovszky

pan3793:
> parquet-java does not use them

@gszadovszky Probably right, because I haven't seen user reports issues related to that. But this can not be proven by UT, because currently, Maven UT always runs against vanilla Jackson libs, class relocation only happens on packaging.

Anyway, we should either correctly transform those service files or purge them.

Generally, applying `ServicesResourceTransformer` is preferred to keep the functionality completeness of the shaded libs, for example, we generally keep all classes of the shaded lib instead of keeping only the used classes.

gszadovszky:
@pan3793, can the shaded services interfere with a potential official Jackson library on the classpath? If not, I'm good with this. If yes, we should rather purge the services that we are not using.

PR #3165, opened 2025-02-28
GH-3141: Add constructor to ParquetFileReader to allow passing in parquet footer

yuzhu:
### Rationale for this change
HadoopFile is getting deprecated, it is useful to be able to pass in a parsed footer instead of reading the footer every time

### What changes are included in this PR?
a new constructor

### Are these changes tested?
Yes

### Are there any user-facing changes?
No

Some code duplication is there, because we cant call another constructor and catch the IOException to close the stream. Let me know if you would like this handled in another way. Thanks. 

Closes #3141

PR #3262, opened 2025-08-06
GH-3141: Add constructor to `ParquetFileReader` to allow passing in parquet footer and expose setRequestedSchema that accepts `List<ColumnDescriptor>`

### LINKED ISSUE #3141: Add more constructors to ParquetFileReader

It seems that we are moving towards InputFile instead of HadoopConf and path (indicated by the deprecation notice), but the constructors using InputFile are missing important ones, such as 
public ParquetFileReader(InputFile file, ParquetMetadata footer,ParquetReadOptions options, SeekableInputStream f) 
where an external footer can be passed in. 

If it seems ok, I would like to add some additional constructors to bring the InputFile-based ones to parity with HadoopConf-based ones. 

### Component(s)

_No response_

pan3793:
Co-authored-by: David Zhu <david@alluxio.com>

### Rationale for this change

This PR takes over #3165 as the author is silent for a while.

The proposed API allows the caller to reuse Footer/InputStream to save HDFS NameNode RPCs, https://github.com/apache/spark/pull/50765 demonstrates how this PR benefits Spark.

- Add constructor to `ParquetFileReader` to allow passing in parquet footer
- Add new method `setRequestedSchema(List<ColumnDescriptor>)` to `ParquetFileReader`
- UT is modified to use the newly added API.

### Are these changes tested?

Yes, UT is tuned.

It exposes new public APIs for `ParquetFileReader`.

Close #3141
Close #3165

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -729,6 +729,21 @@ public class ParquetFileReader implements Closeable {
     return new ParquetFileReader(file, options, f);
   }
 
+  /**
+   * Open a {@link InputFile file} with {@link ParquetMetadata footer} and {@link ParquetReadOptions options}.
+   *
+   * @param file    an input file
+   * @param footer  a {@link ParquetMetadata} footer already read from the file
+   * @param options parquet read options
+   * @param f       the input stream for the file
+   * @return an open ParquetFileReader
+   * @throws IOException if there is an error while opening the file
+   */
+  public static ParquetFileReader open(InputFile file, ParquetMetadata footer, ParquetReadOptions options, SeekableInputStream f)
+      throws IOException {
+    return new ParquetFileReader(file, footer, options, f);
+  }
+
   protected SeekableInputStream f;
   private final InputFile file;
   private final ParquetReadOptions options;
@@ -930,6 +945,12 @@ public class ParquetFileReader implements Closeable {
     this(file, options, file.newStream());
   }
 
+  /**
+   * @param file    Path to a parquet file
+   * @param options {@link ParquetReadOptions}
+   * @param f       a {@link SeekableInputStream} for the parquet file
+   * @throws IOException if the file can not be opened
+   */
   public ParquetFileReader(InputFile file, ParquetReadOptions options, SeekableInputStream f) throws IOException {
     this.converter = new ParquetMetadataConverter(options);
     this.file = file;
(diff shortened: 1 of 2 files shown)


wangyum (approved):
LGTM

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1054,13 +1121,17 @@ public List<BlockMetaData> getRowGroups() {
     return blocks;
   }
 
-  public void setRequestedSchema(MessageType projection) {
+  public void setRequestedSchema(List<ColumnDescriptor> columns) {
```
wgtmac, line 1100: Why do we need this change? Does it exist in the original PR?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageChecksums.java
```diff
@@ -688,8 +690,13 @@ private int getDataOffset(Page page) {
...
       throws IOException {
-    ParquetMetadata footer = ParquetFileReader.readFooter(conf, path);
-    return new ParquetFileReader(conf, footer.getFileMetaData(), path, footer.getBlocks(), columns);
+    HadoopInputFile inputFile = HadoopInputFile.fromPath(path, conf);
+    SeekableInputStream inputStream = inputFile.newStream();
+    ParquetReadOptions readOptions = HadoopReadOptions.builder(conf).build();
+    ParquetMetadata footer = ParquetFileReader.readFooter(inputFile, readOptions, inputStream);
+    ParquetFileReader reader = ParquetFileReader.open(inputFile, footer, readOptions, inputStream);
+    reader.setRequestedSchema(columns);
+    return reader;
```
wgtmac, line 699: It would be better to explicitly add a new test case rather than silently changing old ones.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
pan3793: the diff here is not smart, this actually adds a new method

`public void setRequestedSchema(List<ColumnDescriptor> columns)`

there are many ctors of `ParquetFileReader` were marked as deprecated but without providing the equivalent replacement, for example,

```
  @Deprecated
  public ParquetFileReader(
      Configuration configuration,
      FileMetaData fileMetaData,
      Path filePath,
      List<BlockMetaData> blocks,
      List<ColumnDescriptor> columns)
```

this API is proposed to restore the ability to set read columns via `List<ColumnDescriptor>`, instead of `MessageType`

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDataPageChecksums.java
pan3793: I understand that, but it seems there is no dedicated test suite for `ParquetFileReader` constructor. The replaced API is deprecated, and the replacement does not reduce the test coverage because the deprecated API is used many times in other test suites.

wgtmac (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -943,6 +965,51 @@ public ParquetFileReader(InputFile file, ParquetReadOptions options, SeekableInp
       f.close();
       throw e;
     }
+
+    this.fileMetaData = footer.getFileMetaData();
```
wgtmac: Ah, I found these lines duplicate a lot with the below ctor. Can we create a new method instead?

pan3793: @wgtmac I'm afraid we can not do that

image

I also tried to make `ParquetMetadata footer` nullable, and read from the file when `footer` is NULL, but this does not work either, because some places pass NULL explicitly.

To completely fix the code duplication issue of `ParquetFileReader`, I think we should introduce a `Builder` for it, but this is out of the scope of this PR. Given that this is a public API used widely by downstream projects, deprecating current APIs and suggesting migrating to new APIs is another burden ...

wgtmac: Can we use a private delegate constructor?

pan3793: Seems infeasible before Java 23 ([JEP 482](https://openjdk.org/jeps/482))

image

wgtmac: Can't we simply write this:

```java
  public ParquetFileReader(InputFile file, ParquetReadOptions options, SeekableInputStream f) throws IOException {
    this(file, readFooter(file, options, f, /*closeStreamOnFailure=*/true), options, f);
  }
```

To keep the same behavior, we also need to modify readFooter to close the stream once failed:

```java
  public static final ParquetMetadata readFooter(InputFile file, ParquetReadOptions options, SeekableInputStream f, boolean closeStreamOnFailure)
      throws IOException {
    ParquetMetadataConverter converter = new ParquetMetadataConverter(options);
    try {
      return readFooter(file, options, f, converter);
    } catch (Exception e) {
      // In case that readFooter throws an exception in the constructor, the new stream
      // should be closed. Otherwise, there's no way to close this outside.
      if (closeStreamOnFailure) {
        f.close();
      }
      throw e;
    }
  }
```

pan3793: ~`readFooter` is a public API, I don't think we should change the semantics silently~

oh, I see, you add a new parameter `closeStreamOnFailure`

pan3793: @wgtmac thank you for the suggestion, it's feasible, updated.

wgtmac: Thanks a lot for your effort!

wgtmac (approved):

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -567,8 +567,23 @@ public static final ParquetMetadata readFooter(InputFile file, MetadataFilter fi
 
   public static final ParquetMetadata readFooter(InputFile file, ParquetReadOptions options, SeekableInputStream f)
       throws IOException {
+    return readFooter(file, options, f, /*closeStreamOnFailure*/ false);
+  }
+
+  public static final ParquetMetadata readFooter(
```
wgtmac: nit: can we make it private?

pan3793:
```suggestion
  private static final ParquetMetadata readFooter(
```

pan3793: updated

wgtmac:
Thanks @pan3793 and @wangyum!

PR #3264, opened 2025-08-06
GH-3263: Add DictionaryPage.decode to allow dictionary reuse in the ColumnReaderBase ctor

### LINKED ISSUE #3263: Add and Expose c-tor to ColumnReaderBase to allow Dictionary reuse

I wrote a driver for reading/writing flat parquet files column by column. ColumnReaderBase is used to read column values. In some use cases, multiple threads read the same ColumnChunk concurrently. In this scenario I want to avoid needless duplication of the immutable Dictionary object which in many cases copies the DictionaryPage into a primitive array store.

In order to accomplish without upstream code changes, I abused Java reflection to set this field in a very roundabout way. See:
https://github.com/Earnix/parquetforge/blob/74073acb4642a9dc9afd114e7151ae4f8288fdcf/parquetforge-base/src/main/java/com/earnix/parquet/columnar/reader/chunk/internal/HackyParquetExtendedColumnReader.java#L15

With a minor visibility change to the base class c-tor, and Dictionary as a c-tor parameter, this can be avoided

### Component(s)

Core

pyckle:
### Rationale for this change
I wrote a driver for reading/writing flat parquet files column by column. ColumnReaderBase is used to read column values. In some use cases, multiple threads read the same ColumnChunk concurrently. In this scenario I want to avoid needless duplication of the immutable Dictionary object which in many cases copies the DictionaryPage into a primitive array store.

In order to accomplish without upstream code changes, I abused Java reflection to set this field in a very roundabout way. See:
https://github.com/Earnix/parquetforge/blob/74073acb4642a9dc9afd114e7151ae4f8288fdcf/parquetforge-base/src/main/java/com/earnix/parquet/columnar/reader/chunk/internal/HackyParquetExtendedColumnReader.java#L15

With a minor visibility change to the base class c-tor, and Dictionary as a c-tor parameter, this can be avoided

ColumnReaderBase c-tor is changed to protected to support this integration. An additional c-tor is added that receives the Dictionary object.

### Are these changes tested?
Yes. Existing tests cover these flows already.

No.

Closes #3263

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
@@ -453,7 +453,7 @@ abstract class ColumnReaderBase implements ColumnReader {
     this.writerVersion = writerVersion;
     this.maxDefinitionLevel = path.getMaxDefinitionLevel();
     DictionaryPage dictionaryPage = pageReader.readDictionaryPage();
-    this.dictionary = dictionaryPage == null ? null : dictionaryPage.getDictionary(path);
+    this.dictionary = dictionaryPage == null ? null : dictionaryPage.decode(path);
     if (dictionary != null && converter.hasDictionarySupport()) {
       converter.setDictionary(dictionary);
     }
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
@@ -37,7 +37,6 @@ import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.page.DataPage;
 import org.apache.parquet.column.page.DataPageV1;
 import org.apache.parquet.column.page.DataPageV2;
-import org.apache.parquet.column.page.DictionaryPage;
 import org.apache.parquet.column.page.PageReader;
 import org.apache.parquet.column.values.RequiresPreviousReader;
 import org.apache.parquet.column.values.ValuesReader;
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
@@ -445,39 +445,18 @@ abstract class ColumnReaderBase implements ColumnReader {
    * @param converter     a converter that materializes the values in this column in the current record
    * @param writerVersion writer version string from the Parquet file being read
    */
-  protected ColumnReaderBase(
+  ColumnReaderBase(
       ColumnDescriptor path, PageReader pageReader, PrimitiveConverter converter, ParsedVersion writerVersion) {
-    this(path, readDict(path, pageReader), pageReader, converter, writerVersion);
-  }
-
-  protected ColumnReaderBase(
-        ColumnDescriptor path, Dictionary dictionary, PageReader dataPageReader, PrimitiveConverter converter, ParsedVersion writerVersion) {
     this.path = Objects.requireNonNull(path, "path cannot be null");
-    this.pageReader = Objects.requireNonNull(dataPageReader, "pageReader cannot be null");
+    this.pageReader = Objects.requireNonNull(pageReader, "pageReader cannot be null");
     this.converter = Objects.requireNonNull(converter, "converter cannot be null");
     this.writerVersion = writerVersion;
     this.maxDefinitionLevel = path.getMaxDefinitionLevel();
-    this.dictionary = dictionary;
-    if (dictionary != null && converter.hasDictionarySupport()) {
-      converter.setDictionary(dictionary);
-    }
+    this.dictionary = pageReader.getDictionary(path);
     this.totalValueCount = pageReader.getTotalValueCount();
     if (totalValueCount <= 0) {
       throw new ParquetDecodingException("totalValueCount '" + totalValueCount + "' <= 0");
     }
-  }
-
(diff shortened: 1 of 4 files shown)


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
```diff
@@ -445,30 +445,39 @@ void writeValue() {
...
+
+  protected ColumnReaderBase(
+        ColumnDescriptor path, Dictionary dictionary, PageReader dataPageReader, PrimitiveConverter converter, ParsedVersion writerVersion) {
     this.path = Objects.requireNonNull(path, "path cannot be null");
-    this.pageReader = Objects.requireNonNull(pageReader, "pageReader cannot be null");
+    this.pageReader = Objects.requireNonNull(dataPageReader, "pageReader cannot be null");
     this.converter = Objects.requireNonNull(converter, "converter cannot be null");
     this.writerVersion = writerVersion;
     this.maxDefinitionLevel = path.getMaxDefinitionLevel();
-    DictionaryPage dictionaryPage = pageReader.readDictionaryPage();
```
wgtmac, line 455: Is it better to wrap the `PageReader` to directly return the cached `Dictionary` instead of hacking the public api like this?

pyckle: The PageReader returns a DictionaryPage which is then decoded in the constructor:
`dictionary = dictionaryPage.getEncoding().initDictionary(path, dictionaryPage);`
As `getEncoding()` returns an enum, there's no way to override the method of `initDictionary()` to return a cached instance.
Adding a `getDictionary()` function to the `PageReader` interface would therefore be necessary. In order not to break any downstream `PageReader` implementations, a default method must be added, which seems unclean to me, but maybe it's OK.
I implemented this so you can see what it looks like. If you prefer this approach, I can add the caching logic to the `ColumnChunkPageReadStore` too.
Another option could be adding a function `decodeDictionary()` to `DictionaryPage`
Let me know what you think is cleanest.

wgtmac: I think the current implementation is much cleaner. The original solution does not prohibit passing any kind of dictionary into the column reader which may cause other issues.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
```diff
@@ -452,19 +452,7 @@ void writeValue() {
...
-    if (dictionaryPage != null) {
-      try {
-        this.dictionary = dictionaryPage.getEncoding().initDictionary(path, dictionaryPage);
-        if (converter.hasDictionarySupport()) {
```
wgtmac, line 459: The converter logic should be preserved, right?

pyckle: Whoops, good catch. Fixed.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
pyckle: I see your point with the cleanliness. This prevents a dictionary with the wrong encoding from being passed in by mistake.
Looking at `ColumnChunkPageReadStore.ColumnChunkPageReader`, I am no longer certain that caching the decoded Dictionary is a safe change, as it will increase memory pressure if `PageReaders` are stored in memory with no active reader. This is due to the fact that `ColumnChunkPageReader` stores the pages with compression, and decompresses when it needs to be read.
As such, I think it's safer to merge in its current state.

wgtmac: Yes, that makes sense.

Since this is a change related to public api, I'd like to hear advice from @gszadovszky @Fokko

gszadovszky: I think I would prefer the `DictionaryPage.decode` approach.

pyckle: @wgtmac  @gszadovszky - implemented. Let me know what you think

parquet-column/src/main/java/org/apache/parquet/column/page/DictionaryPage.java
```diff
@@ -74,6 +77,17 @@ public DictionaryPage copy() throws IOException {
     return new DictionaryPage(BytesInput.copy(bytes), getUncompressedSize(), dictionarySize, encoding);
   }
 
+  /**
+   * @return the decoded dictionary
+   */
+  public Dictionary getDictionary(ColumnDescriptor path) {
```
gszadovszky: I think something like `decode` would be a better naming.

gszadovszky (approved):

parquet-column/src/main/java/org/apache/parquet/column/page/DictionaryPage.java
pyckle: Changed the name of the method to decode.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnReaderBase.java
wgtmac: Thanks! Could you please add a test case?

pyckle: Added.

Issue #3267, opened 2025-08-15
Fix Todos/Tests throughout the codebase


ArnavBalyan:
### Describe the enhancement requested

Improve tests and fix todos in the codebase

PR #3268, opened 2025-08-15
GH-3267: Add comprehensive assertions to TestMemPageStore

ArnavBalyan:
- Resolves the TODO comment in TestMemPageStore.java by adding proper assertions for page storage/retrieval
 - Added some more assertions to verify

### DIFF
--- a/parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
@@ -22,6 +22,7 @@ import static org.apache.parquet.column.Encoding.BIT_PACKED;
 import static org.apache.parquet.column.Encoding.PLAIN;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import org.apache.parquet.bytes.BytesInput;
@@ -71,6 +72,7 @@ public class TestMemPageStore {
       // Assert page has expected value count
       assertEquals("Each page should have 209 values", 209, readPage.getValueCount());
       // Assert encodings when the implementation is DataPageV1
+      assertTrue("Page should be an instance of DataPageV1", readPage instanceof DataPageV1);
       if (readPage instanceof DataPageV1) {
         DataPageV1 v1 = (DataPageV1) readPage;
         assertEquals("Page repetition level encoding should be BIT_PACKED", BIT_PACKED, v1.getRlEncoding());
@@ -80,7 +82,6 @@ public class TestMemPageStore {
 
       total += readPage.getValueCount();
       pageCount++;
-      LOG.info("Page " + pageCount + ": " + readPage.toString());
     } while (total < totalValueCount);
 
     // Assert we read exactly the expected number of pages and values


parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
```diff
@@ -46,19 +49,42 @@ public void test() throws IOException {
...
+    LOG.info("Total value count: " + totalValueCount);
+
+    assertEquals("Expected total value count to be 836 (4 pages * 209 values)", 836, totalValueCount);
+
     int total = 0;
+    int pageCount = 0;
     do {
       DataPage readPage = pageReader.readPage();
+
+      // Assert page was successfully read
```
wgtmac, line 70: It seems that TestMemPageStore previously tests nothing but merely to prove that methods of MemPageStore are working.

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
```diff
@@ -46,19 +49,42 @@ public void test() throws IOException {
...
+        assertEquals("Page repetition level encoding should be BIT_PACKED", BIT_PACKED, v1.getRlEncoding());
+        assertEquals("Page definition level encoding should be BIT_PACKED", BIT_PACKED, v1.getDlEncoding());
+        assertEquals("Page value encoding should be PLAIN", PLAIN, v1.getValueEncoding());
+      }
+
       total += readPage.getValueCount();
-      LOG.info(readPage.toString());
-      // TODO: assert
+      pageCount++;
+      LOG.info("Page " + pageCount + ": " + readPage.toString());
```
wgtmac: Should we totally remove the LOG? It is useless at this moment.

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
```diff
@@ -46,19 +49,42 @@ public void test() throws IOException {
...
+    int pageCount = 0;
     do {
       DataPage readPage = pageReader.readPage();
+
+      // Assert page was successfully read
+      assertNotNull("Page should not be null", readPage);
+      // Assert page has expected value count
+      assertEquals("Each page should have 209 values", 209, readPage.getValueCount());
+      // Assert encodings when the implementation is DataPageV1
+      if (readPage instanceof DataPageV1) {
```
wgtmac, line 76: I'd prefer asserting that this is a DataPageV1

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
ArnavBalyan: Sure will remove

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
ArnavBalyan: Added the assertion thanks

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
ArnavBalyan: Removed the redundant log

parquet-column/src/test/java/org/apache/parquet/column/mem/TestMemPageStore.java
ArnavBalyan: Yes seems like previously`MemPageStore` was just supposed to test the MemPageStore upper level functionality. Seems that `TestMemColumn` does the actual data and lower level checks

wgtmac (approved):

wgtmac:
Thanks @ArnavBalyan for polishing it! (TBH I think we should delete this file since it actually tests nothing meaningful)

PR #3270, opened 2025-08-18
GH-3246: Clarify the behaviour of big decimal for BYTE_ARRAY

### LINKED ISSUE #3246: Is big Decimal in Parquet big endian?
### Describe the usage question you have. Please include as many useful details as  possible.

It's not clear from description https://parquet.apache.org/docs/file-format/data-pages/encodings/ how are big Decimals (precision >= 18) placed. As I could understand they can be encoded as BYTE_ARRAY or FIXED_LEN_BYTE_ARRAY. But there're two ways to place "the bytes contained in the array" from memory: big endian and little endian. I.e. they are in LE in int128 in memory. Could I place the bytes in FIXED_LEN_BYTE_ARRAY as is?

It's also not clean if there's some encoding over these decimal's bytes, should the data be swapped to correct endianness before of after the encoding?

### Component(s)

_No response_

ArnavBalyan:
- The behaviour of BYTE_ARRAY or FIXED_LEN_BYTE_ARRAY for big decimals is not clear from the docs.
 - Added code comments at the read/writer to make the behaviour more clear for the reader.
 - Fixes #3246

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
@@ -27,6 +27,14 @@ import org.apache.parquet.io.api.Binary;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * ValuesReader for variable-length {@code BYTE_ARRAY} columns.
+ *
+ * <p>When the column carries the logical type {@code DECIMAL}, the bytes read
+ * here are the big-endian two's-complement form of the un-scaled integer.
+ * It slices the requested number of bytes without flipping and
+ * returns them as a {@link org.apache.parquet.io.api.Binary}.
+ */
 public class BinaryPlainValuesReader extends ValuesReader {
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -27,6 +27,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * ValuesReader for variable-length {@code BYTE_ARRAY} columns.
```
wgtmac, line 31: I don't think this is the good place to add the comment since this class is for a specific encoding of binary physical type.

As I have replied in the issue, the spec is clear on this so we don't even need the explaination.

ArnavBalyan: Thanks for the review, yes on retrospect code comment may not be the best, I'll check if the readme can be made more clear since the user pointed to the readme asking for clarity. Will iterate and get back thanks!

Issue #2961, opened 2024-07-18
Infinite recursion in AvroSchemaConverter


gauravkhatri05:
### Describe the bug, including details regarding any error messages, version, and platform.

**Issue:**  I am trying to write Parquet file using S3SinkConnector and when it tries to convert avro schema using a method `AvroSchemaConverter.convert(Schema avroSchema)` it enters into the infinite recursion.

Please refer to link below for the problematic code:
https://github.com/apache/parquet-java/blob/1de2af8b8b7dc86af120da7c0ce3e43453ea98dc/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java#L147C3-L163C4

**Details**
Please find the schema which has nested recursive reference (ObjXX.struct_add_list.fld_list.ref_val.fld_list.ref_val...) that is being parsed:
`
{
  "type": "record",
  "name": "ObjXX",
  "fields": [
    {
      "name": "id",
      "type": [
        "null",
        "long"
      ],
      "default": null
    },
    {
      "name": "struct_add_list",
      "type": [
        "null",
        {
          "type": "array",
          "items": [
            "null",
            {
              "type": "record",
              "name": "ObjStructAdd",
              "fields": [
                {
                  "name": "name",
                  "type": [
                    "null",
                    "string"
                  ],
                  "default": null
                },
                {
                  "name": "fld_list",
                  "type": [
                    "null",
                    {
                      "type": "array",
                      "items": [
                        "null",
                        {
                          "type": "record",
                          "name": "ObjStructAddFld",
                          "fields": [
                            {
                              "name": "name",
                              "type": [
                                "null",
                                "string"
                              ],
                              "default": null
                            },
                            {
                              "name": "ref_val",
                              "type": [
                                "null",
                                "ObjStructAdd"
                              ],
                              "default": null
                            }
                          ]
                        }
                      ]
                    }
                  ],
                  "default": null
                }
              ]
            }
          ]
        }
      ],
      "default": null
    },
    {
      "name": "kafka_offset",
      "type": [
        "null",
        "long"
      ],
      "default": null
    },
    {
      "name": "kafka_partition",
      "type": [
        "null",
        "int"
      ],
      "default": null
    },
    {
      "name": "kafka_timestamp",
      "type": {
        "type": "long",
        "logicalType": "timestamp-millis"
      }
    },
    {
      "name": "kafka_key",
      "type": [
        "null",
        "string"
      ],
      "default": null
    },
    {
[...]

PR #3272, opened 2025-08-20
GH-2961: Cycle detection in AvroSchemaConverter to prevent infinite recursion

ArnavBalyan:
Summary: 
 - Currently recursive objects in avro can cause infinite recursion and stackoverflow error.
 - Added a cycle detection to prevent it and gracefully through error at a certain max depth, can be controlled through: `parquet.avro.max-recursion`
 - Fixes: #2961

Testing:
 - Added UTs

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -51,6 +51,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -151,14 +152,14 @@ public class AvroSchemaConverter {
       throw new IllegalArgumentException("Avro schema must be a record.");
     }
     return new MessageType(
-        avroSchema.getFullName(), convertFields(avroSchema.getFields(), "", new HashSet<Schema>()));
+        avroSchema.getFullName(), convertFields(avroSchema.getFields(), "", new IdentityHashMap<Schema, Void>()));
   }
 
   private List<Type> convertFields(List<Schema.Field> fields, String schemaPath) {
-    return convertFields(fields, schemaPath, new HashSet<Schema>());
+    return convertFields(fields, schemaPath, new IdentityHashMap<Schema, Void>());
   }
 
-  private List<Type> convertFields(List<Schema.Field> fields, String schemaPath, Set<Schema> seenSchemas) {
+  private List<Type> convertFields(List<Schema.Field> fields, String schemaPath, IdentityHashMap<Schema, Void> seenSchemas) {
     List<Type> types = new ArrayList<Type>();
     for (Schema.Field field : fields) {
       if (field.schema().getType().equals(Schema.Type.NULL)) {
@@ -173,29 +174,29 @@ public class AvroSchemaConverter {
     return convertField(fieldName, schema, Type.Repetition.REQUIRED, schemaPath);
   }
 
-  private Type convertField(String fieldName, Schema schema, String schemaPath, Set<Schema> seenSchemas) {
+  private Type convertField(String fieldName, Schema schema, String schemaPath, IdentityHashMap<Schema, Void> seenSchemas) {
     return convertField(fieldName, schema, Type.Repetition.REQUIRED, schemaPath, seenSchemas);
   }
 
   @SuppressWarnings("deprecation")
   private Type convertField(String fieldName, Schema schema, Type.Repetition repetition, String schemaPath) {
-    return convertField(fieldName, schema, repetition, schemaPath, new HashSet<Schema>());
+    return convertField(fieldName, schema, repetition, schemaPath, new IdentityHashMap<Schema, Void>());
   }
 
   @SuppressWarnings("deprecation")
   private Type convertField(
-      String fieldName, Schema schema, Type.Repetition repetition, String schemaPath, Set<Schema> seenSchemas) {
+      String fieldName, Schema schema, Type.Repetition repetition, String schemaPath, IdentityHashMap<Schema, Void> seenSchemas) {
     Schema.Type type = schema.getType();
     LogicalType logicalType = schema.getLogicalType();
 
     if (type.equals(Schema.Type.RECORD) || type.equals(Schema.Type.ENUM) || type.equals(Schema.Type.FIXED)) {
       // If this schema has already been seen in the current branch, we have a recursion loop
-      if (seenSchemas.contains(schema)) {
+      if (seenSchemas.containsKey(schema)) {
         throw new UnsupportedOperationException(
             "Recursive Avro schemas are not supported by parquet-avro: " + schema.getFullName());
       }
-      seenSchemas = new HashSet<>(seenSchemas);
-      seenSchemas.add(schema);
+      seenSchemas = new IdentityHashMap<>(seenSchemas);
+      seenSchemas.put(schema, null);
     }
 
     Types.PrimitiveBuilder<PrimitiveType> builder;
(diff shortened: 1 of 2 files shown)


parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -150,16 +151,23 @@ public MessageType convert(Schema avroSchema) {
     if (!avroSchema.getType().equals(Schema.Type.RECORD)) {
       throw new IllegalArgumentException("Avro schema must be a record.");
     }
-    return new MessageType(avroSchema.getFullName(), convertFields(avroSchema.getFields(), ""));
+    return new MessageType(
+        avroSchema.getFullName(),
+        convertFields(avroSchema.getFields(), "", new IdentityHashMap<Schema, Integer>()));
```
gszadovszky: You don't need a map, only a set since you only add the same element once. (Maybe use `HashSet`?)

Isn't it enough to use the type name only instead of the whole `Schema` object?

gszadovszky (changes requested):

parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
ArnavBalyan: updated thanks! I believe the name type may repeat but the schema reference would always be unique. Would you suggest updating to type name?

gszadovszky: I'm sorry I was wrong. Your choice of using `IdentityHashMap` was the best approach. (Unfortunately, there is no such thing in standard Java as `IdentityHashSet`.)

Could you change it back using `IdentityHashMap`? (Maybe with the value type `Void` and setting `null` values to be clear we don't actually use them.)

ArnavBalyan: Updated thanks!

gauravkhatri05:
Hi @gszadovszky,

Thanks for your follow-up! I want to clarify the key distinction between the current approach and our patch:

### What’s happening in current patch

With the `IdentityHashMap` check in `convertField(...)`:

```java
if (seenSchemas.containsKey(schema)) {
  throw new UnsupportedOperationException(
      "Recursive Avro schemas are not supported by parquet-avro: " + schema.getFullName());
}
```

As soon as the converter encounters a recursive reference (e.g., `Employee -> Employee`), it **throws immediately**, preventing even common, benign recursive structures from working.

### What our patch achieves (max-depth)

We don’t block recursion outright. Instead, with a **configurable max-depth** (e.g., default = 10), we allow traversal to continue until that limit is reached. Only then do we prevent further descent. This allows recursive schemas to function *within a bounded, safe depth*.

A typical use case is an `Employee` record with a `List<Employee>` for subordinates—common in organizational hierarchies.

Here’s an example payload (max depth = 4) that works with our patch but fails with the current implementation (IdentityHashMap):

```
A (CEO)
|
|--------------- B (CTO)
|                        |
|                        |---------------- B1 (Mgr 1)
|                        |                         |---------------- B1-1 (Architect)
|                        |                         |---------------- B1-2 (Team Lead)
|                        |                                                   |---------------- B1-2-1 (SDE)
|                        |
|                        |---------------- B2 (Product Owner)
|
|--------------- C (HR)
```

* **Current patch (IdentityHashMap):** Rejects the recursive reference at the first encounter.
* **Max-depth patch:** Supports the structure up to the safe limit, enabling real-world recursive models like trees or org charts.

### For reference: Similar approach advocated elsewhere

This [Medium](https://medium.com/hadoop-noob/recursive-avro-schema-for-parquet-a87c84b6aca1?utm_source=chatgpt.com) post explicitly suggests the same solution—stop schema generation when a specified **max\_depth** is reached to handle recursive Avro schemas for Parquet gracefully.

### Here’s the essential snippet from our max-depth patch:

```java
private final int maxDepth = Integer.parseInt(
    getApplicationProps().getProperty("parquet.schema.parser.recursion.max.depth", "10")
);

// MAX-DEPTH PATCH START
public MessageType convert(Schema avroSchema) {
  if (!avroSchema.getType().equals(Schema.Type.RECORD)) {
    throw new IllegalArgumentException("Avro schema must be a record.");
  }
  return new MessageType(
    avroSchema.getFullName(),
    convertFields(avroSchema.getFields(), "", maxDepth)
  );
}

private List<Type> convertFields(List<Schema.Field> fields, String schemaPath, int depth) {
  var types = new ArrayList<Type>();
  --depth;
  for (Schema.Field field : fields) {
    if (field.schema().getType().equals(NULL)) {
      continue;
    }
    if (depth > 0 || !isRecord(field)) {
      types.add(convertField(field, appendPath(schemaPath, field.name()), depth));
    }
  }
  return types;
}

private boolean isRecord(Schema.Field field) {
    var fieldSchema = field.schema();
    if (UNION == fieldSchema.getType()) {
      var optionalNonNullUnionElementSchema = fieldSchema.getTypes().stream()
        .filter(schema -> NULL != schema.getType())
        .findFirst();
      if (optionalNonNullUnionElementSchema.isPresent()) {
        var nonNullUnionElementSchema = optionalNonNullUnionElementSchema.get();
        if (ARRAY == nonNullUnionElementSchema.getType()) {
          return checkArrayForRecord(nonNullUnionElementSchema);
        } else if (MAP == nonNullUnionElementSchema.getType()) {
          return checkMapForRecord(nonNullUnionElementSchema);
        }
        return RECORD == nonNullUnionElementSchema.getType();
      }
      return false;
    } else if (ARRAY == fieldSchema.getType()) {
      return checkArrayForRecord(fieldSchema);
    } else if (MAP == fieldSchema.getType()) {
      return checkMapForRecord(fieldSchema);
    }
    return fieldSchema.getType() == RECORD;
  }

  private boolean checkMapForRecord(Schema mapSchema) {
    var valueSchema = mapSchema.getValueType();
    if (UNION == valueSchema.getType()) {
      return valueSchema.getTypes().stream().anyMatch(
        schema -> RECORD == schema.getType()
      );
    } else {
      return RECORD == valueSchema.getType();
    }
  }

  private boolean checkArrayForRecord(Schema arraySchema) {
    if (UNION == arraySchema.getElementType().getType()) {
      return arraySchema.getElementType().getTypes().stream().anyMatch(
        schema -> RECORD == schema.getType()
      );
    } else {
      return RECORD == arraySchema.getElementType().getType();
    }
  }

// ... rest of convertField, convertUnion, etc., updated similarly ...
// MAX-DEPTH PATCH END
```

This enhancement allows recursive schemas to be **usable and safe**, not just blocked. The depth is configurable so it can be tuned based on workload and performance needs—default = 10 provides stability; dropping below 7 has shown issues in our tests.

Let me know if my explanation makes sense or if you’d like me to clarify further.

gszadovszky:
Thanks @gauravkhatri05 for the detailed clarification.

So, we are trying to translate a recursive schema to a type system where recursive schemas are not supported. By using a max depth you translate the recursive Avro schema to a Parquet schema where the recursive types are simply repeated again. So the Avro schema A -> B -> C -> D -> A gets translated (with the default 10 max depth) to A -> B -> C -> D -> A -> B -> C -> D -> A -> B. What happens with the data having more depth? Do we fail or silently ignore the rest? Neither seems to be the best approach for systems where no issues are expected after a proper schema conversion.

It seems this topic has some history: [PARQUET-129](https://issues.apache.org/jira/browse/PARQUET-129)

Maybe bring up this topic on the dev list. Let's see what the community thinks.

gauravkhatri05:
> Thanks for raising it @gauravkhatri05. +1 on the dev list since we are discussing adding partial support for an unsupported feature. In the meanwhile let me check if there is anything else that can be done do to unblock this usecase. (formally supporting recursive schemas in parquet has likely already been discussed will check if anything can be done)

Thanks @gszadovszky & @ArnavBalyan for the response.

While thinking further, I realized a limitation in my current approach:

* If the self-reference is immediate, then max-depth works as expected, allowing recursive traversal of the same schema until the depth is exhausted.
* But if the self-reference occurs deeper in the path, then *every* child schema along that path also consumes part of the max-depth before we even reach the recursive schema.

For example, with max-depth = 10:

* Input Avro schema: `A -> B -> C -> D -> A`
* Output schema today: `A -> B -> C -> D -> A -> B -> C -> D -> A -> B`
* Expected behavior: `A -> B -> C -> D` repeated 10 times (instead of prematurely exhausting depth).

To make this more optimal, I think we need to track **seen schemas within the same traversal path**. That way, only encountering the *same schema in the same path* decrement the depth count. This is essentially a tree traversal problem where we need to track nodes along the current branch.

What do you think about this refinement?

As you suggested, I’ll also bring this up on the dev list so we can get broader community feedback. Could you please share the correct DL address / process to raise this discussion?

ddeath:
@gszadovszky I would like to point out that we already have max depth in parquet-proto. It was implemented couple years ago.

https://github.com/apache/parquet-java/pull/995

It would be nice to support this if possible. I came here because confluent s3 sink connector is failing on this :D we have field of type `google.protobuf.Struct` in our message but that Struct is self referencing and is causing StackOverflow on older version.

Do you think we could take similar approach?

Issue #3273, opened 2025-08-21
Size stats produces unbounded output allow user to scope output


ArnavBalyan:
### Describe the enhancement requested

The parquet CLI command size-stats currently prints chunk-level size statistics for every row group and every column in the file. On wide schemas and/or large files, this can generate massive, output, making the command hard to use.

PR #3274, opened 2025-08-21
GH-3273: Add scoped chunk level statistics to avoid unbounded output

ArnavBalyan:
- Size stats currently prints chunk-level size statistics for every row group and every column in the file. On wide schemas and/or large files, this can generate massive, output, making the command hard to use.
 - Add filtering flags to size-stats aligning with column-index:
      -c, --column: filter by column dot paths (e.g., a.b.c)
      -r, --row-group: filter by row-group indexes (0-based)
 - Fixes: #3273


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowSizeStatisticsCommand.java
```diff
@@ -60,9 +72,13 @@ public int run() throws IOException {
 
       console.info("\nFile path: {}", source);
 
-      List<BlockMetaData> rowGroups = footer.getBlocks();
-      for (int index = 0, n = rowGroups.size(); index < n; index++) {
-        printRowGroupSizeStats(console, index, rowGroups.get(index), schema);
+      List<BlockMetaData> blocks = footer.getBlocks();
+      Set<Integer> allowedRowGroups = rowGroups == null ? null : new HashSet<>(rowGroups);
```
gszadovszky: nit: If you would create an empty Set instead of initializing with a null, you wouldn't have to check for null in the loop.

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowSizeStatisticsCommand.java
```diff
@@ -84,7 +100,16 @@ private void printRowGroupSizeStats(Logger console, int index, BlockMetaData row
     console.info(
         String.format(formatString, "column", "unencoded bytes", "rep level histogram", "def level histogram"));
 
+    Set<String> allowedColumns = null;
+    if (columns != null && !columns.isEmpty()) {
+      allowedColumns = new HashSet<>(columns);
+    }
```
gszadovszky: See previous comment

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowSizeStatisticsCommand.java
ArnavBalyan: Previous implementation was using map since we were tracking depth, updated to set thanks!

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowSizeStatisticsCommand.java
ArnavBalyan: updated

gszadovszky (approved):

wgtmac (approved):
Thanks @ArnavBalyan!

Issue #2914, opened 2024-06-05
Add usage documentation for the Java library


asfimport:
The Java parquet library has no usage documentation besides the sparse information available in the README. The only thing I could find were a few old (10yr) 3rd party tutorials scattered on the internet using the hadoop module. I spent a work day sifting through the API docs and searching on the internet to try to piece together something. Ultimately, I decided to give up on doing Parquet files using Java because there are alternative file formats that are better documented, and I felt trying to use parquet-mr would be a huge hassle to maintain in the future. This library seems reasonably maintained and comprehensive, but there is just a huge barrier to using the library which I think turns off a lot of developers like me.

I kindly request usage documentation be written to cover all the major aspects of the library, and for the more nitty gritty use cases, pointers to what API classes/methods could be looked at further.

I may be misunderstanding the purpose of this library, and if so, is there a different Java Parquet library that is recommended for higher level parquet file IO?

**Reporter**: [Isaac Nygaard](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=azmisov)

<sub>**Note**: *This issue was originally created as [PARQUET-2490](https://issues.apache.org/jira/browse/PARQUET-2490). Please see the [migration documentation](https://issues.apache.org/jira/browse/PARQUET-2502) for further details.*</sub>

VarshaUN:
hi @asfimport ,can u brief about what should be added to readme and if the issue is open can u assign it to me so that i can work on it!!

wgtmac:
@VarshaUN I just assigned it to you. Please feel free to add any documentation as you see fit. Thanks for your interest!

VarshaUN:
> @VarshaUN I just assigned it to you. Please feel free to add any documentation as you see fit. Thanks for your interest!

het thanks for assigning.but are there any prerequistes for contributing to this or just clone it and add doc??

wgtmac:
There is no prerequisite. I'm not sure if your proposal is too wide to complete. In my mind it may be some code examples like what we have in Apache Arrow: https://arrow.apache.org/cookbook/java/. Parquet-java contains a lot of modules, in which I think `parquet-avro` and `parquet-hadoop` are the most important ones from the perspective of end users.

VarshaUN:
[...]

PR #3275, opened 2025-08-22
GH-2914: Add documentation for the Java library

ArnavBalyan:
- Addresses the issues raised in #2914
 - Added detailed java examples for various usecases under examples.
 - Developers can add more usage examples under `examples` for different usecases in the future.


wgtmac (approved):

wgtmac (changes requested):
Sorry for approving it by accident. I need to leave a comment to retract it.

examples/src/main/java/org/apache/parquet/examples/AdvancedFeaturesExample.java
```diff
@@ -0,0 +1,169 @@
...
+
+        testPredicatePushdown(filename);
+
+        testPerformanceOptimization(filename);
+    }
+
+    private static void writeSalesData(String filename, MessageType schema) throws IOException {
+        Path file = new Path(filename);
+
+        try (ParquetWriter<Group> writer = ExampleParquetWriter.builder(file)
```
wgtmac: I am just hesitant to use `ExampleParquetWriter` as examples which is not for production purpose. Adding an example module also incurs more maintenance burden so I don't think this is a good idea TBH.

ArnavBalyan: I see, we can remove the sub-module and just produce it as reference only example, it should also be able to resolve the documentation concerns raised in the issue, wdyt?

ArnavBalyan: removed the dependency from ExampleParquetWriter and removed the pom to eliminate maintenance overhead

Issue #3277, opened 2025-08-24
Persist write configuration to parquet files


ArnavBalyan:
### Describe the enhancement requested

The writer configuration influences the nature of the parquet file and often required for debugging and posterity.
Application logs are temporary and wiped after a few days. 
Support configurable behaviour which can pass the writer configuration to the parquet footer optionally.

PR #3278, opened 2025-08-24
GH-3277: Support passing write configurations to footer optionally

ArnavBalyan:
- Write configurations are often needed for debugging and posterity however application logs are lost in a few days.
 - This change adds an optional flag, which when enabled passes the write configurations to the file footer.
 - Default flag is false, and can be enabled by users to pass this additional metadata to their parquet files.


wgtmac:
IMHO, this is merely a customized logic which can be handled pretty well by specific applications. We don't want to take the maintenance overhead.

Issue #3279, opened 2025-08-25
Decouple Hadoop/Native I/O implementations


ArnavBalyan:
### Describe the enhancement requested

Provide a clean interface to use hadoop/native io classes. Users should be able to use parquet without hadoop on the classpath. This issue tracks the effort

PR #3280, opened 2025-08-25
GH-3279: Introduce StorageProvider to de-couple Hadoop vs NIO constructs

ArnavBalyan:
- Add a minimal StorageProvider abstraction and selector that routes to hadoop vs non-hadoop classes.
 - Make Hadoop I/O resolve FileSystem per path to correctly hit the right connector.
 - This isolates local I/O from Hadoop today and sets up a clean interface to pull the correct concrete implementation at runtime.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/storage/StorageProvider.java
+++ b/parquet-common/src/main/java/org/apache/parquet/storage/StorageProvider.java
@@ -1,0 +1,69 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.storage;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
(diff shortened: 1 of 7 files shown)


parquet-common/src/main/java/org/apache/parquet/storage/StorageProvider.java
```diff
@@ -0,0 +1,69 @@
...
+ * can be referenced by any other module without creating additional coupling
+ * between them.</p>
+ */
+public interface StorageProvider {
+
+  /**
+   * Opens the given path for reading.
+   *
+   * @param path fully-qualified file path (implementation specific semantics)
+   * @return an InputStream that must be closed by the caller
```
shangxinli, line 46: The JavaDoc should specify what specific IOException
  subclasses might be thrown (e.g., FileNotFoundException,
  AccessDeniedException) to help implementers and users handle
  errors appropriately.

parquet-common/src/main/java/org/apache/parquet/storage/StorageProvider.java
```diff
@@ -0,0 +1,69 @@
...
+  InputStream openForRead(String path) throws IOException;
+
+  /**
+   * Opens the given path for writing. If the file already exists the behaviour
+   * depends on the overwrite flag.
+   *
+   * @param path fully-qualified file path
+   * @param overwrite  whether an existing file should be replaced
+   * @return an OutputStream that must be closed by the caller
+   */
```
shangxinli, line 57: The interface should clarify stream ownership and closing
   responsibilities. Consider returning AutoCloseable wrappers or
  documenting that callers must use try-with-resources.

parquet-common/src/main/java/org/apache/parquet/storage/StorageProvider.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public interface StorageProvider {
+
+  /**
+   * Opens the given path for reading.
+   *
+   * @param path fully-qualified file path (implementation specific semantics)
+   * @return an InputStream that must be closed by the caller
+   */
+  InputStream openForRead(String path) throws IOException;
```
jerolba, line 48: In the future if we want to use this abstraction to read parquet files, the code requires to create a [SeekableInputStream](
https://github.com/apache/parquet-java/blob/master/parquet-common/src/main/java/org/apache/parquet/io/InputFile.java#L42)
Can be more useful to use [SeekableInputStream](https://github.com/apache/parquet-java/blob/master/parquet-common/src/main/java/org/apache/parquet/io/SeekableInputStream.java) that already extends InputStream?

PR #3281, opened 2025-08-25
GH-3282: Add encryption info CLI support for Parquet file encryption metadata

ArnavBalyan:
- Since Parquet 1.12, encryption has become a first class citizen, with support for footer and column level encryption.
 - However, users have no clear way to check encryption metadata, mode, or whether footer/file is encrypted.
 - This PR adds a simple, dedicated CLI command: `parquet-cli encryption-info <file>`
 - The command reports the following:
     - File-level encryption type: PLAINTEXT_FOOTER or ENCRYPTED_FOOTER.
     - Summary of column encryption, per-column details and their encryption status.


parquet-cli/src/main/java/org/apache/parquet/cli/commands/EncryptionInfoCommand.java
```diff
@@ -0,0 +1,204 @@
...
+
+  @Override
+  public int run() throws IOException {
+    Preconditions.checkArgument(targets.size() == 1, "Exactly one Parquet file must be specified.");
+    String source = targets.get(0);
+
+    ParquetMetadata footer =
+        ParquetFileReader.readFooter(getConf(), qualifiedPath(source), ParquetMetadataConverter.NO_FILTER);
+
+    FileMetaData meta = footer.getFileMetaData();
```
nandorKollar, line 71: I think it would be nice to also print out details about the encryption algorithm, wouldn't it?

wgtmac:
cc @ggershinsky @shangxinli for experts on encryption

parquet-cli/src/main/java/org/apache/parquet/cli/commands/EncryptionInfoCommand.java
ArnavBalyan: yeah that's a great point will add support

ggershinsky:
Some other details worth printing -
- is a column encrypted with the footer key or with a column-specific key?
- if all columns are encrypted with the footer key, then the file is in "uniform encryption" mode; can print this (so the user knows one key only is used in a file and can open every column)
- explicit info on the footer encryption mode - encrypted or plaintext
- optional (via a flag) printing of the key metadata of the footer key and (if available) of the column keys - can be useful for debugging key retrieval. This is binary, but maybe something similar to "hexdump -C" can be performed where some effort is made to find/print ASCII text chunks (often, key metadata has text/json parts)
- advanced debugging: print the AAD-related fields

Issue #2967, opened 2024-07-22
Providing parquet-avro configuration to parquet-cli


andrewthad:
### Describe the usage question you have. Please include as many useful details as  possible.

The parquet-avro library supports using some more recent parquet features with configuration settings like `parquet.avro.write-parquet-uuid` (needs to be true) and `parquet.avro.write-old-list-structure` (needs to be false). However, parquet-cli does not have a documented way to set these. Poking around with strace, I've found a `convert` command does result in an `openat` syscall targetting `logging.properties` to look for logging-related configuration, but I've not found any other attempts to open any other files with the `properties` extensions. To work around this, I have been manually changing the values of `WRITE_PARQUET_UUID_DEFAULT` and friends in the source and rebuilding the project.

This issue can be resolved in one of two ways:

1. Document in the readme file for parquet-cli that it does not support configuring any `AvroWriteSupport` settings
2. Document in the readme file for parquet-cli that it does support configuring `AvroWriteSupport` settings, and provide an example of what CLI option is used to do this.

PR #3283, opened 2025-08-27
GH-2967: Support unified config options for convert parquet-cli

ArnavBalyan:
- Added --conf parameter that accepts user's key=value configuration pairs
 - users can now configure any Avro/Parquet setting via CLI.
 - The conf can be passed any number of times with the user defined values and will be appended to the internally constructed config.
 - Fixes: #2967

### DIFF
--- a/parquet-cli/README.md
+++ b/parquet-cli/README.md
@@ -136,23 +136,15 @@ Usage: parquet [options] [command] [command options]
 
 ### Configuration Options
 
-The `convert` and `convert-csv` commands support a generic configuration mechanism:
-
 - `--conf` or `--property`: Set any configuration property in format `key=value`. Can be specified multiple times.
-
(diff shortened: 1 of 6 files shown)


gszadovszky (commented):
This is a great improvement. Thanks a lot for working on it!

I have two ideas to improve it:
* What do you think about adding the conf support centrally for every commands by setting it in [Main](https://github.com/apache/parquet-java/blob/4aa2ea91863274aebb1eded243ce275912c16010/parquet-cli/src/main/java/org/apache/parquet/cli/Main.java#L167-L167)?
* Maybe adding an additional option that supports reading a config file (and the config pairs at command line can override these values)?

gszadovszky:
Thanks for the ping, @ArnavBalyan. I usually don't like merging just when I'm approving to give some time for other comments. But 3 days are more than enough...

Issue #3286, opened 2025-08-28
Parquet-Protobuf does not work with Parquet-cli


ArnavBalyan:
### Describe the bug, including details regarding any error messages, version, and platform.

 - Currently parquet-cli breaks while operating on parquet files generated through parquet protobuf
 - This is because the CLI currently uses AvroReadSupport and AvroRecrodConverter which breaks for protobuf since underlying schema and data is different.
 - Fix the issue to allow users to be able to view files generated through protobuf.

Issue #2708, opened 2022-08-31
parquet-cli fails at supporting parquet-protobuf generated files


asfimport:
i generated a parquet file using a protobuf with this proto definition:

```java

message IndexPath {
  // Index of item in path.
  repeated int32 index = 1;
}

message SomeEvent {
  // truncated/obfuscated wrapper
  optional IndexPath client_position = 1;
}
```

this gets translated to the following parquet schema using the new compliant schema for lists:

```java

message SomeEvent {
  optional group client_position = 1 {
    optional group index (LIST) = 1 {
      repeated group list {
        required int32 element;
      }
    }
  }
}
```

this causes parquet-cli cat to barf on a file containing these events:
> java.lang.RuntimeException: Failed on record 0
> at org.apache.parquet.cli.commands.CatCommand.run(CatCommand.java:86)
> at org.apache.parquet.cli.Main.run(Main.java:157)
> at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
> at org.apache.parquet.cli.Main.main(Main.java:187)
> Caused by: java.lang.ClassCastException: required int32 element is not a group
> at org.apache.parquet.schema.Type.asGroupType(Type.java:248)
> at org.apache.parquet.avro.AvroRecordConverter.newConverter(AvroRecordConverter.java:284)
> at org.apache.parquet.avro.AvroRecordConverter.newConverter(AvroRecordConverter.java:228)
> at org.apache.parquet.avro.AvroRecordConverter.access$100(AvroRecordConverter.java:74)
> at org.apache.parquet.avro.AvroRecordConverter$AvroCollectionConverter$ElementConverter.<init>(AvroRecordConverter.java:539)
> at org.apache.parquet.avro.AvroRecordConverter$AvroCollectionConverter.<init>(AvroRecordConverter.java:489)
> at org.apache.parquet.avro.AvroRecordConverter.newConverter(AvroRecordConverter.java:293)
> at org.apache.parquet.avro.AvroRecordConverter.<init>(AvroRecordConverter.java:137)
> at org.apache.parquet.avro.AvroRecordConverter.newConverter(AvroRecordConverter.java:284)
> at org.apache.parquet.avro.AvroRecordConverter.<init>(AvroRecordConverter.java:137)
> at org.apache.parquet.avro.AvroRecordConverter.<init>(AvroRecordConverter.java:91)
[...]

PR #3287, opened 2025-08-28
GH-3286: Add support for Parquet-Protobuf in Parquet-cli

ArnavBalyan:
### Summary
 - Currently parquet-cli breaks while operating on parquet files generated through parquet protobuf
 - This is because the CLI currently uses AvroReadSupport and AvroRecrodConverter which breaks for protobuf since underlying schema/data is different.
 - We now support reading proto files through CLI reader which routes the request to simple group factory for protobuf parquet files.
 - Related to #2708 and others.

#### - Before:
```
Time elapsed: 1.351 s <<< ERROR!
java.lang.RuntimeException: Failed on record 0 in file /tmp/junit149783857212573183/proto_someevent.parquet
        at org.apache.parquet.cli.commands.CatCommand.run(CatCommand.java:89)
        at org.apache.parquet.cli.commands.CatCommandTest.testCatCommandProtoParquetAutoDetected(CatCommandTest.java:82)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
Caused by: org.apache.parquet.io.ParquetDecodingException: Can not read value at 0 in block -1 in file file:/tmp/junit149783857212573183/proto_someevent.parquet
        at org.apache.parquet.hadoop.InternalParquetRecordReader.nextKeyValue(InternalParquetRecordReader.java:280)
        at org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:136)
        at org.apache.parquet.hadoop.ParquetReader.read(ParquetReader.java:140)
        at org.apache.parquet.cli.BaseCommand$2$1.advance(BaseCommand.java:407)
        at org.apache.parquet.cli.BaseCommand$2$1.<init>(BaseCommand.java:388)
        at org.apache.parquet.cli.BaseCommand$2.iterator(BaseCommand.java:386)
        at org.apache.parquet.cli.commands.CatCommand.run(CatCommand.java:76)
        at org.apache.parquet.cli.commands.CatCommandTest.testCatCommandProtoParquetAutoDetected(CatCommandTest.java:82)
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   CatCommandTest.testCatCommandProtoParquetAutoDetected:82 » Runtime Failed on record 0 in file /tmp/junit149783857212573183/proto_someevent.parquet
```

#### - After:
```
[INFO] Running org.apache.parquet.cli.commands...
repeatedInt: 1
repeatedInt: 2
repeatedInt: 3
```
(Succesful read)

### DIFF
--- a/parquet-cli/pom.xml
+++ b/parquet-cli/pom.xml
@@ -85,6 +85,8 @@
       <artifactId>parquet-avro</artifactId>
       <version>${project.version}</version>
     </dependency>
+
+    <!-- Protobuf dependencies for CLI Tests -->
     <dependency>
       <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-protobuf</artifactId>
@@ -92,19 +94,19 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
-    <!-- CatCommandTest (for TestProtobuf) -->
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
       <version>3.25.6</version>
-      <scope>${deps.scope}</scope>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-protobuf</artifactId>
       <version>${project.version}</version>
-      <scope>${deps.scope}</scope>
+      <scope>test</scope>
     </dependency>
+
     <dependency>
       <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-format-structures</artifactId>
--- a/parquet-cli/src/test/java/org/apache/parquet/cli/commands/CatCommandTest.java
+++ b/parquet-cli/src/test/java/org/apache/parquet/cli/commands/CatCommandTest.java
@@ -82,13 +82,15 @@ public class CatCommandTest extends ParquetFileTest {
   }
 
   @Test
-  public void testCatCommandProtoParquetSucceedsWithAutoDetection() throws Exception {
-    File protoFile = new File(getTempFolder(), "proto_someevent.parquet");
-    writeProtoParquet(protoFile);
+  public void testCatCommandWithSimpleReaderConfig() throws Exception {
+    File regularFile = parquetFile();
 
-    CatCommand cmd = new CatCommand(createLogger(), 0);
-    cmd.sourceFiles = Arrays.asList(protoFile.getAbsolutePath());
-    cmd.setConf(new Configuration());
+    Configuration conf = new Configuration();
+    conf.setBoolean("parquet.enable.simple-reader", true);
+
+    CatCommand cmd = new CatCommand(createLogger(), 5);
+    cmd.sourceFiles = Arrays.asList(regularFile.getAbsolutePath());
+    cmd.setConf(conf);
 
     int result = cmd.run();
     Assert.assertEquals(0, result);
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/util/Schemas.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/util/Schemas.java
@@ -82,7 +82,8 @@ public class Schemas {
     if (schemaString != null) {
       return new Schema.Parser().parse(schemaString);
     } else {
-      return new AvroSchemaConverter().convert(footer.getFileMetaData().getSchema());
+      return new AvroSchemaConverter(conf)
+          .convert(footer.getFileMetaData().getSchema());
     }
   }
 


parquet-cli/pom.xml
```diff
@@ -85,6 +85,26 @@
...
+      <artifactId>protobuf-java</artifactId>
+      <version>3.25.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-protobuf</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
```
gszadovszky, line 108: By adding these dependencies in test scope only, wouldn't cause them missing at the command line execution?
There are two ways we can use the cli. One is containing the "normal" scoped dependencies for the Hadoop env, and the other is containing the "provided" scope as well for standalone. I don't think these deps will be added to either one.

parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
```diff
@@ -55,14 +56,56 @@
...
 import org.slf4j.Logger;
 
 public abstract class BaseCommand implements Command, Configurable {
 
   private static final String RESOURCE_URI_SCHEME = "resource";
   private static final String STDIN_AS_SOURCE = "stdin";
 
+  /**
+   * Note for dev: Due to legancy reasons, parquet-cli used the avro schema reader which
+   * breaks for files generated through proto. This logic is in place to auto-detect such cases
```
gszadovszky, line 73: I'm wondering if this problem is only related to the files created with the proto binding. I am fine with auto-detecting the proto related files and use the example binding to read the files automatically, but maybe, it would be a good idea, to also provide this option for any files.

gszadovszky (changes requested):

parquet-cli/pom.xml
ArnavBalyan: Thats a great catch thanks! moved to ${deps.scope} which can be enabled via the profile being used

parquet-cli/src/main/java/org/apache/parquet/cli/BaseCommand.java
ArnavBalyan: Agreed, have added a config to allow user side enablement

parquet-cli/src/test/java/org/apache/parquet/cli/commands/CatCommandTest.java
```diff
@@ -63,4 +67,42 @@ public void testCatCommandWithInvalidColumn() throws IOException {
...
+    CatCommand cmd = new CatCommand(createLogger(), 0);
+    cmd.sourceFiles = Arrays.asList(protoFile.getAbsolutePath());
+    cmd.setConf(new Configuration());
+
+    int result = cmd.run();
+    Assert.assertEquals(0, result);
+  }
+
+  @Test
+  public void testCatCommandProtoParquetSucceedsWithAutoDetection() throws Exception {
```
gszadovszky: Aren't the two tests the same? Maybe we should test the new config as well. (AFAIK the avro reader fails with INT96 values by default.)

parquet-cli/pom.xml
```diff
@@ -85,6 +85,26 @@
...
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-protobuf</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
+    <!-- CatCommandTest (for TestProtobuf) -->
```
gszadovszky: I think I was wrong with my previous comment. We do not actually use protobuf in prod, right? We check only for the related key in the footer and use the example binding to get the values. So `test` scope seems legit.
(If you want to keep this comment, maybe let it be for all three dependencies with some separation?)

parquet-cli/src/test/java/org/apache/parquet/cli/commands/CatCommandTest.java
ArnavBalyan: Updated, added the test with the config, it seems the behaviour is same for the INT96 values. Added a test to set the config, and verified simple reader being used on the other side. This can be hopefully be better tested with the e2e test where debug statements can be logged and verified

parquet-cli/pom.xml
ArnavBalyan: updated, yeah just noticed protobuf is only pulled to write the test file

gszadovszky (approved):

PR #3288, opened 2025-08-29
GH-3290: Restore Snapshot versions for vector/benchmark modules

### LINKED ISSUE #3290: CI broke after release process

Investigate CI/Build failure on latest main.
Certain module versions were not reverted to snapshot version causing build failures

### Component(s)

_No response_

ArnavBalyan:
- 1.16 release process did not revert back the snapshot versions correctly.
 - This caused the main to have inconsistent dependencies/ possible build failures.
 - Since the versions are listed as additional profiles and not in the module list by default, we should probably configure a maven release plugin which can include them through custom rules. Will add support as a followup.
 - Fixes #3290


gszadovszky (approved):
Thanks @ArnavBalyan for fixing this

PR #3295, opened 2025-08-30
Add comparator for `UnknownLogicalType` (#3292)

wgtmac:
This was missing, which caused exceptions. This PR adds them including a test.

It takes the comparator of the underlying physical type, but it should always compare two null values:

https://github.com/apache/parquet-format/blob/471d33a3c9d46e05b8a189437a9364f959f93f09/LogicalTypes.md?plain=1#L887-L892


wgtmac:
Since the release script does not affect the vector plugin, leaving them to use 1.16.0 will just coincidentally make them 1.16.0 in the release tarball.

Issue #3289, opened 2025-08-29
End to end tests for Parquet CLI


ArnavBalyan:
### Describe the enhancement requested

The CLI currently does not have end to end tests to check if argument passed to the command made it to the application logic. This can be solved in various ways, this issue tracks adding support for better tests in parquet CLI

PR #3296, opened 2025-08-30
GH-3289: Introduce Parquet E2E test harness to support CLI testing

ArnavBalyan:
- Currently parquet does not have a good mechanism to have e2e testing for CLI
 - Developers have to manually call methods on the concrete class instead of the actual CLI route, this makes testing hard and prone to errors. Further the method return is tested instead of the actual console output.
 <br>

 - This change introduces a test harness which is a wrapper over main CLI class and provides utilities to implement CLI tests easily.
 - Users can simply test with single line of code such as: cli("command").ok().matchOutputFromFile("txt"). The harness provides some other utilities like `linecount`, `partial match` etc.
 - The CLI output can be added to a txt to ensure readability and test the actual console output


parquet-cli/src/test/java/org/apache/parquet/cli/ShowSizeStatisticsCliTest.java
```diff
@@ -0,0 +1,48 @@
...
+    cli("help size-stats").ok().matchOutputFromFile("src/test/resources/cli-outputs/size-stats-help.txt");
+  }
+
+  @Test
+  public void showsSchemaOutput() throws Exception {
+    File file = parquetFile();
+    cli("schema " + file.getAbsolutePath())
+        .ok()
+        .matchOutputFromFile("src/test/resources/cli-outputs/size-stats-column.txt");
+  }
```
gszadovszky: I don't get why it is related to size statistics. Shouldn't it be put into a separate test for `schema`?

parquet-cli/src/test/resources/cli-outputs/size-stats.txt
```diff
@@ -0,0 +1,10 @@
+Row group 0
+--------------------------------------------------------------------------------
+column         unencoded bytes rep level histogram                      def level histogram                     
+[int32_field]  -               -                                        -
```
gszadovszky: Shouldn't we choose a file that actually has size statistics for some columns?

parquet-cli/src/test/java/org/apache/parquet/cli/commands/FileTest.java
```diff
@@ -33,13 +33,13 @@
...
-  static final String BINARY_FIELD = "binary_field";
-  static final String FIXED_LEN_BYTE_ARRAY_FIELD = "flba_field";
-  static final String DATE_FIELD = "date_field";
+  public static final String INT32_FIELD = "int32_field";
+  public static final String INT64_FIELD = "int64_field";
+  public static final String FLOAT_FIELD = "float_field";
+  public static final String DOUBLE_FIELD = "double_field";
+  public static final String BINARY_FIELD = "binary_field";
+  public static final String FIXED_LEN_BYTE_ARRAY_FIELD = "flba_field";
+  public static final String DATE_FIELD = "date_field";
```
gszadovszky: At the end you haven't used these constants, have you?

parquet-cli/src/main/java/org/apache/parquet/cli/Main.java
```diff
@@ -82,7 +82,8 @@ public class Main extends Configured implements Tool {
   @VisibleForTesting
   final JCommander jc;
 
-  Main(Logger console) {
+  @VisibleForTesting
+  public Main(Logger console) {
```
gszadovszky: nit: If you would create the cli harness fw into the same package, you wouldn't have to make it public.

gszadovszky (changes requested):
Thank you @ArnavBalyan for picking this up. It is a great improvement to improve the test coverage of the cli!

parquet-cli/src/test/java/org/apache/parquet/cli/ShowSizeStatisticsCliTest.java
ArnavBalyan: Ah yes, I wanted to show some examples, will move it to a different test 👍

ArnavBalyan: Updated

parquet-cli/src/test/resources/cli-outputs/size-stats.txt
ArnavBalyan: Updated the test to generate more data which produces some meaningful stats

parquet-cli/src/test/java/org/apache/parquet/cli/commands/FileTest.java
ArnavBalyan: updated

parquet-cli/src/main/java/org/apache/parquet/cli/Main.java
ArnavBalyan: Updated

gszadovszky:
@ArnavBalyan, could you check the failures?

gszadovszky (approved):
Thanks, @ArnavBalyan.

Fokko:
@ArnavBalyan Gentle ping! Did you find some time to fix the test?

Issue #3294, opened 2025-08-29
Include optional profiles for releases


ArnavBalyan:
### Describe the enhancement requested

Investigate why the optional profile was not included in the release causing #3290 to occur

PR #3297, opened 2025-08-30
GH-3294: Include optional profiles for release process

ArnavBalyan:
- Currently the optional profile (vector-plugins) in parquet includes 2 additional modules which are not included in the reactor list by default.
 - This can cause them to be missed during the release/revert process.
 - Including the same in the release process.
 - Fixes: #3294, related to #3288

### DIFF
--- a/dev/prepare-release.sh
+++ b/dev/prepare-release.sh
@@ -39,6 +39,6 @@ new_development_version="$release_version-SNAPSHOT"
 tag="apache-parquet-$release_version-rc$2"
 
 ./mvnw release:clean
-./mvnw release:prepare -DskipTests -Darguments=-DskipTests -Dtag="$tag" "-DreleaseVersion=$release_version" -DdevelopmentVersion="$new_development_version"
+./mvnw release:prepare -Pvector-plugins -DskipTests -Darguments=-DskipTests -Dtag="$tag" "-DreleaseVersion=$release_version" -DdevelopmentVersion="$new_development_version"
 
 echo "Finish staging binary artifacts by running: ./mvnw release:perform"
--- a/dev/finalize-release
+++ b/dev/finalize-release
@@ -34,7 +34,7 @@ rc_tag="$release_tag-rc$2"
 new_development_version="$3-SNAPSHOT"
 
 git tag -am "Release Apache Parquet $release_version" "$release_tag" "$rc_tag"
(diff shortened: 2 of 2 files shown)


wgtmac (approved):

dev/prepare-release.sh
```diff
@@ -39,6 +39,6 @@ new_development_version="$release_version-SNAPSHOT"
 tag="apache-parquet-$release_version-rc$2"
 
 ./mvnw release:clean
-./mvnw release:prepare -DskipTests -Darguments=-DskipTests -Dtag="$tag" "-DreleaseVersion=$release_version" -DdevelopmentVersion="$new_development_version"
+./mvnw release:prepare -Pvector-plugins -DskipTests -Darguments=-DskipTests -Dtag="$tag" "-DreleaseVersion=$release_version" -DdevelopmentVersion="$new_development_version"
```
pan3793, line 42: does this command perform compile and upload jars? `vector-plugins` requires JDK 17 to build

wgtmac: Ah, that's a good catch!

pan3793: there is an alternative command that applies the version change to all modules (include optional modules)
```
./mvnw versions:set -DgenerateBackupPoms=false -DnewVersion=$release_version
```

pan3793: Or, maybe you can consider releasing Parquet using a higher JDK version, with `-release 8` to ensure Java 8 compatibility.

ArnavBalyan: Thanks thats a great find @pan3793!

I think the solution may depend on whether we want to include `vector-plugins` in our releases. It seems the current behaviour is not to include vector-plugins. Reverting this should continue maintaining this behaviour (not sure how it got included this time for 1.16 as vector-plugins is not included by default) 

Optionally if we want to include vector-plugins in the releases, release script can be updated to include it conditionally on the JDK being used. Maybe have an automatic CI job for releases to avoid dev overhead. 
@wgtmac @Fokko what do you think? Can take next steps accordingly thanks!

wgtmac: It is a separate discussion to include vector-plugins in the release. Perhaps we can go with `./mvnw versions:set -DgenerateBackupPoms=false -DnewVersion=$release_version` proposed by @pan3793

ArnavBalyan: 👍 Will raise the fix

PR #3301, opened 2025-09-01
GH-3300: add ParquetWriter and ParquetReader builders constructor without params

### LINKED ISSUE #3300: Allow creating ParquetWriter and ParquetReader builders without an initial file

**Background**

I am the developer of [Carpet](https://github.com/jerolba/parquet-carpet), a library that builds upon the Parquet Java project. 

Carpet reuses the primitives provided by Parquet Java library: to instantiate `ParquetWriter` and `ParquetReader` classes, I use and extend existing builders. To support all Parquet Java features transparently to users, Carpet exposes the same configuration options as the underlying library, and transitively reuse all logic implemented by `ParquetWriter`, `InternalParquetRecordWriter`, `ParquetReader`  and `InternalParquetRecordReader`.

My goal is to implement data partitioning and parallel read/write capabilities in Carpet. I want to continue reusing existing builders and logic, but with additional abstractions for partitioning and parallelism. To support this, I need to build multiple instances of ParquetWriter/ParquetReader based on a single builder configuration, only changing the target file to write/read before calling the build method.

With the [existing](https://github.com/apache/parquet-java/blob/8be0dadaea9ded29d61fb10afb6dfe7d516ee316/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L479) [builders](https://github.com/apache/parquet-java/blob/8be0dadaea9ded29d61fb10afb6dfe7d516ee316/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java#L217), it's not possible to implement this feature because an Input/OutputFile is a required argument in the static builder-creation methods. This forces the file to be defined at the very beginning of the configuration process, preventing the builder instance from being reused for different files.

**Suggested change**

I propose to add empty constructor builders for `ParquetWriter` and `ParquetReader` classes.

The builder pattern usually allows configuring all options in different stages and order, but guarantees that all required parameters are set and consistent for the instances to be created. Why force setting the file at the beginning instead of allowing it to be set later?

[This example code](https://github.com/apache/parquet-java/blob/4b6fbc1fb636f5553416b6bfd9ce7767ed058bbb/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java#L170):

```java
ParquetWriter<Group> writer = ExampleParquetWriter.builder(outputFile)
            .withAllocator(allocator)
            .withCompressionCodec(UNCOMPRESSED)
            .withRowGroupSize(1024)
            .withPageSize(1024)
            .withDictionaryPageSize(512)
            .enableDictionaryEncoding()
            .withValidation(false)
            .withWriterVersion(version)
            .withConf(conf)
            .build();
```

Could also be valid as:

```java
ExampleParquetWriter.Builder<Group> builder = ExampleParquetWriter.builder()
            .withAllocator(allocator)
            .withCompressionCodec(UNCOMPRESSED)
            .withRowGroupSize(1024)
            .withPageSize(1024)
            .withDictionaryPageSize(512)
            .enableDictionaryEncoding()
            .withValidation(false)
            .withWriterVersion(version)
            .withConf(conf);
ParquetWriter<Group> writer1 = builder.withFile(outputFile1).build();
...
ParquetWriter<Group> writer2 = builder.withFile(outputFile2).build();
```

My proposal is to:

* Introduce no-argument builder() static methods for the ParquetWriter and ParquetReader builders
* Allow setting the file later in the builder configuration process (only for InputFile and OutputFile types, leaving apart the deprecated Hadoop Path)
* Ensure validation is performed within the `build` and `withFile` methods to throw an exception if the file has not been set, preserving the builder's safety guarantees.

These changes would be fully backward-compatible, as the existing `builder(file)` methods would remain untouched.

This would allow for more flexible usage patterns and better support for advanced use cases like partitioning and parallel processing in Carpet or even as part of Parquet Java.

### Component(s)

Core

jerolba:
### Rationale for this change

Support the creation of multiple `ParquetWriter` and `ParquetReader` instances with different files, which part from a single builder.

More explanations in the issue GH-3300

- Create no-argument builder() static methods for the `ParquetWriter` and `ParquetReader` builders
- Make `file` non final to support late configuration
- Add `withFile` methods to configure the Output/InputFile to use after calling build. The code validates that there is no `path` already configured
- Add validations in `build` methods to guarantee the consistency of configured file.

### Are these changes tested?

`TestParquetReader` and `TestParquetWriter` classes are updated to include tests that verify all cases:
- Null configuration
- No configuration of file calling to build method
- Duplicate configuration via `path` and `file` options
- A file can be read or written

No

Closes #3300

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
@@ -22,6 +22,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.bytes.ByteBufferAllocator;
@@ -476,6 +477,9 @@ public class ParquetWriter<T> implements Closeable {
     private boolean enableValidation = DEFAULT_IS_VALIDATING_ENABLED;
     private ParquetProperties.Builder encodingPropsBuilder = ParquetProperties.builder();
 
+    protected Builder() {}
+
+    @Deprecated
     protected Builder(Path path) {
       this.path = path;
     }
@@ -524,6 +528,20 @@ public class ParquetWriter<T> implements Closeable {
      */
     public SELF withConf(ParquetConfiguration conf) {
       this.conf = conf;
+      return self();
+    }
+
+    /**
+     * Set the {@link OutputFile} to be written by the constructed writer.
+     *
+     * @param file a {@code OutputFile}
+     * @return this builder for method chaining.
+     */
+    public SELF withFile(OutputFile file) {
+      this.file = Objects.requireNonNull(file, "file cannot be null");
+      if (this.path != null) {
+        throw new IllegalStateException("Cannot set both path and file");
+      }
       return self();
     }
 
@@ -960,6 +978,9 @@ public class ParquetWriter<T> implements Closeable {
      * @throws IOException if there is an error while creating the writer
      */
     public ParquetWriter<T> build() throws IOException {
+      if (file == null && path == null) {
+        throw new IllegalStateException("File or Path must be set");
+      }
       if (conf == null) {
         conf = new HadoopParquetConfiguration();
       }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java
@@ -190,9 +190,13 @@ public class ParquetReader<T> implements Closeable {
     return new Builder<>(readSupport, path);
   }
 
+  public static <T> Builder<T> builder() {
+    return new Builder<T>();
+  }
+
   public static class Builder<T> {
     private final ReadSupport<T> readSupport;
-    private final InputFile file;
+    private InputFile file;
     private final Path path;
     private Filter filter = null;
     private ByteBufferAllocator allocator = new HeapByteBufferAllocator();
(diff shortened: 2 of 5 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -476,6 +477,9 @@ public abstract static class Builder<T, SELF extends Builder<T, SELF>> {
     private boolean enableValidation = DEFAULT_IS_VALIDATING_ENABLED;
     private ParquetProperties.Builder encodingPropsBuilder = ParquetProperties.builder();
 
+    protected Builder() {}
+
+    @Deprecated
```
gszadovszky, line 482: I don't think we should deprecate this. It is completely fine using the builder as before.

jerolba: I've added the @Deprecated annotation to make it consistent with `ParquetReader` [builder](https://github.com/apache/parquet-java/blob/4f577a6b64f491d62a2b4bc2acf93d82342f525c/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetReader.java#L217) where  `protected Builder(Path path)` constructor is also deprecated.

My understanding is that the roadmap includes removing `Path` usage in Parquet to avoid Hadoop dependencies.

As you say, the change is not related with the objective of the PR and I can remove it and open an issue to discuss the topic.

gszadovszky: Sorry @jerolba, you're understanding is correct. I've completely overlooked that you're introducing the usage of `InputFile` and deprecating `Path`. All good.

gszadovszky (approved):

PR #3304, opened 2025-09-02
GH-3298: Support unified file based configurations for CLI

### LINKED ISSUE #3298: Add support for unified file based configurations for parquet CLI

Currently users have no way of passing configurations to cli.
We have recently added this support.
Since there can be multiple configs passed to the CLI, allow users to cleanly pass a file instead to provide the config payload

### Component(s)

_No response_

ArnavBalyan:
- Minor patch to allow setting configurations through a single properties file available locally for the CLI
 - User can simply pass a path to config instead of hardcoding the whole config bag.
 - Followup to #3283, closes #3298
 - Supports hadoop style xml and properties files

### DIFF
--- a/parquet-cli/README.md
+++ b/parquet-cli/README.md
@@ -137,6 +137,7 @@ Usage: parquet [options] [command] [command options]
 ### Configuration Options
 
 - `--conf` or `--property`: Set any configuration property in format `key=value`. Can be specified multiple times.
+- `--config-file`: Path to a properties configuration file containing key=value pairs.
 
 Examples:
 ```bash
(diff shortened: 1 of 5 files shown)


gszadovszky:
I think adding the possibility to set the configs via a properties file is great. It would help users to collect their generally used configs.
Meanwhile, since we use Hadoop Configuration, it would be nice to also support the Hadoop xml style config as well. See [here](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/conf/Configuration.html#addResource-org.apache.hadoop.fs.Path-) for details.

PR #3305, opened 2025-09-03
MINOR: Post release of 1.16.0

wgtmac:
### Rationale for this change

This is the post release process of 1.16.0.

- Set previous version to 1.16.0 for japicmp
- Remove optional profile from release script

### Are these changes tested?

No

No

### DIFF
--- a/dev/finalize-release
+++ b/dev/finalize-release
@@ -34,7 +34,7 @@ rc_tag="$release_tag-rc$2"
 new_development_version="$3-SNAPSHOT"
 
 git tag -am "Release Apache Parquet $release_version" "$release_tag" "$rc_tag"
-./mvnw --batch-mode -Pvector-plugins release:update-versions -DdevelopmentVersion="$new_development_version"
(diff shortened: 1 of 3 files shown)


dev/finalize-release
```diff
@@ -34,7 +34,7 @@ rc_tag="$release_tag-rc$2"
 new_development_version="$3-SNAPSHOT"
 
 git tag -am "Release Apache Parquet $release_version" "$release_tag" "$rc_tag"
-./mvnw --batch-mode -Pvector-plugins release:update-versions -DdevelopmentVersion="$new_development_version"
```
Fokko, line 37: Why this change, shouldn't that profile be part of the release?

wgtmac: No, we haven't officially released it yet and it required Java 17 support. Perhaps we can include it once your proposal on the dev ML to bump to Java 17 has reached consensus?

Fokko: Ah, got it. Yes, that would be nice. Less maven profiles is more :)

Fokko (approved):

Issue #3224, opened 2025-05-24
ParquetProperties.valuesWriterFactory is not thread-safe


juja0:
### Describe the bug, including details regarding any error messages, version, and platform.

ParquetProperties.valuesWriterFactory defaults to a singleton instance of DefaultValuesWriterFactory.

ParquetOutputFormat.getRecordWriter provides no way of setting/overriding "valuesWriterFactory" in ParquetProperties.Builder so the same singleton instance of DefaultValuesWriterFactory is used every time.

DefaultValuesWriterFactory.initialize dynamically sets the instance field "delegateFactory" to either a v1 or v2 factory.

When writing to multiple parquet files concurrently, with some threads using v1 and others using v2, the actual writer used becomes unpredictable and files get written with an incorrect format.

Potential fix: ParquetProperties.Builder.valuesWriterFactory should not be instantiated to a singleton instance of DefaultValuesWriterFactory. Instead, it should be instantiated to a new instance of DefaultValuesWriterFactory in the constructor of ParquetProperties.Builder

PR #3308, opened 2025-09-08
GH-3224: Make ParquetProperties.valuesWriterFactory thread safe

ArnavBalyan:
### Rationale for this change
 - Move away from shared singleton DEFAULT_VALUES_WRITER_FACTORY to new instance per builder to prevent concurrent threads from corrupting each other's delegateFactory state.
 - Currently there is a bug where multiple builders can corrupt each others delegateFactory that can cause files to get written with wrong format.
 - Since valuesWriterFactory.initialize() uses the singleton state which can be altered by another writer thread. 
 - Fixes: #3224
 - Added UT to ensure multiple builders dont share the same ValuesWriterFactory state.

### Are these changes tested?
- Yes

- Partially

Closes: #3224


gszadovszky (approved):
> Did you also mean to remove the change altogether and just mark it deprecated with a comment?

No, the change looks good to me now.

Issue #3312, opened 2025-09-16
Support UUID read on parquet thrift reader


ArnavBalyan:
### Describe the enhancement requested

Add uuid read supported on parquet thrift

PR #3313, opened 2025-09-16
GH-3312: Support uuid read converter for parquet thrift

ArnavBalyan:
### Rationale for this change
 - Currently uuid in parquet-thrift is not supported even though it's a formal data type in the schema.
 - The reader throws an error if uuid is attempted to be read for thrift.
 - Add read converter support for reading binary data as uuid in ThriftRecordConverter.
 - Add UT to demonstrate the failure and the fix.

### Are these changes tested?
 - Yes

 - Yes

Fixes: #3312


Fokko (approved):
Looks reasonable to me 👍

wgtmac (approved):

PR #3314, opened 2025-09-16
Bump to Java 11

Fokko:
### Rationale for this change

See dev-list: https://lists.apache.org/thread/cblz22jgllyvw2t63xs3fbzxcvhy5y42

### DIFF
--- a/parquet-plugins/parquet-encoding-vector/pom.xml
+++ b/parquet-plugins/parquet-encoding-vector/pom.xml
@@ -35,7 +35,7 @@
   <url>https://parquet.apache.org</url>
 
   <properties>
-    <!-- Those properties prevent java 8 to try and compile this code -->
+    <!-- Those properties prevent Java 11 to try and compile this code -->
     <maven.compiler.source>17</maven.compiler.source>
     <maven.compiler.target>17</maven.compiler.target>
   </properties>
@@ -74,7 +74,7 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
           <compilerArgs combine.children="append">
-            <compilerArg>${extraJavaVectorArgs}</compilerArg>
+            <compilerArg>--add-modules=jdk.incubator.vector</compilerArg>
           </compilerArgs>
         </configuration>
       </plugin>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -629,17 +629,6 @@
...
 
   <profiles>
-    <profile>
-      <id>jdk9+</id>
-      <activation>
-        <jdk>[9,)</jdk>
-      </activation>
-      <properties>
-        <!-- release takes precedence over source/target if java version is 9 or higher -->
-        <maven.compiler.release>8</maven.compiler.release>
```
gszadovszky, line 639: By removing this profile wouldn't we lack of using 'maven.compiler.release'?

pom.xml
```diff
@@ -69,8 +69,8 @@
   </mailingLists>
 
   <properties>
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
gszadovszky: I think we should start using `maven.compiler.release` instead.

parquet-plugins/parquet-encoding-vector/pom.xml
```diff
@@ -35,7 +35,7 @@
   <url>https://parquet.apache.org</url>
 
   <properties>
-    <!-- Those properties prevent java 8 to try and compile this code -->
+    <!-- Those properties prevent Java 11 to try and compile this code -->
     <maven.compiler.source>17</maven.compiler.source>
     <maven.compiler.target>17</maven.compiler.target>
```
gszadovszky: See the comment in the root pom

pom.xml
Fokko: Good call, updated 👍

gszadovszky (approved):

Fokko:
Result of the thread on the dev-list: https://lists.apache.org/thread/cblz22jgllyvw2t63xs3fbzxcvhy5y42

PR #3323, opened 2025-09-22
MINOR: Remove unused parquet-thrift dependencies

dossett:
### Rationale for this change
There are some unnecessary dependencies in parquet-thrift.  Comments for a couple of them suggest that they're still needed for pig integration, but my local builds worked fine without them.

NO


Fokko (approved):
Less is more 👍

PR #3325, opened 2025-09-23
GH-3320: Ensure parquet reader does not fail due to incorrect statistics

### LINKED ISSUE #3320: Parquet read fails in certain scenarios for files written by parquet-go

Parquet-go can write extra entries in the column index for 1 column chunk. This causes Parquet java to fail with java.lang.ArrayIndexOutOfBoundsException when using column indexes.
 - Parquet java expects the same number of entries in column index and offset index (one entry per data page).
 - When they don't match, the reader seems to be trying to read a non existent page (based on the column stats) and fails. (stacktrace below)

Structure of indexes which lead to the failure:
```
DEBUG: COLUMN1 ColumnIndex=5 stats
  Entry 0: nullPage=false, nullCount=0, min='', max=''
  Entry 1: nullPage=false, nullCount=0, min='xyz', max='abc'
  Entry 2: nullPage=false, nullCount=0, min='xyz', max='abc'
  Entry 3: nullPage=false, nullCount=0, min='xyz', max='abc'
  Entry 4: nullPage=false, nullCount=0, min='xyz', max='abc'

DEBUG: COLUMN1 OffsetIndex=4 pages
  Entry 0: offset=..., compressedSize=..., firstRowIndex=0
  Entry 1: offset=..., compressedSize=..., firstRowIndex=30
  Entry 2: offset=..., compressedSize=..., firstRowIndex=60
  Entry 3: offset=..., compressedSize=..., firstRowIndex=90

(Here we have 1 additional column index for the same data page)
```

There could be potentially 2 ways to fix this:
1. Detect inconsistent metadata and do not rely on column index for unsuitable/incomplete metadata. Log a WARN and continue reading the file without index, which will correctly read the file.
2. Clearly throw an exception indicating the missing/unhandled metadata (in this case, the mismatch in number of column/offset index), so user can disable column index or correct upstream.

Stacktrace: 
```
        at org.apache.parquet.internal.column.columnindex.OffsetIndexBuilder$OffsetIndexImpl.getFirstRowIndex(OffsetIndexBuilder.java:66)
        at org.apache.parquet.internal.filter2.columnindex.RowRanges.create(RowRanges.java:144) 
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter.applyPredicate(ColumnIndexFilter.java:189)
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter.visit(ColumnIndexFilter.java:126)
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter.visit(ColumnIndexFilter.java:57) 
        at org.apache.parquet.filter2.predicate.Operators$NotEq.accept(Operators.java:192)
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter$1.visit(ColumnIndexFilter.java:87) 
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter$1.visit(ColumnIndexFilter.java:82) 
        at org.apache.parquet.filter2.compat.FilterCompat$FilterPredicateCompat.accept(FilterCompat.java:149) 
        at org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter.calculateRowRanges(ColumnIndexFilter.java:82)
        at org.apache.parquet.hadoop.ParquetFileReader.getRowRanges(ParquetFileReader.java:1219) 
        at org.apache.parquet.hadoop.ParquetFileReader.getFilteredRecordCount(ParquetFileReader.java:875) 
...
```

### Component(s)

_No response_

ArnavBalyan:
### Rationale for this change
 - Parquet reader fails if upstream writes wrong statistics in column index vs offset index. 
 - Ensure parquet java can still read the file and ignore corrupt metadata in such cases.
 - Added UT to ensure existing behaviour is preserved.
 - Fixes #3320

### Are these changes tested?
 - Yes

 - Yes

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
@@ -192,7 +192,12 @@ public class ColumnIndexFilter implements Visitor<RowRanges> {
       return allRows();
     }
 
-    return RowRanges.create(rowCount, func.apply(ci), oi);
+    if (!isValidMetadata(ci, oi, columnPath)) {
+      return allRows();
+    }
+
+    PrimitiveIterator.OfInt pageIndexes = func.apply(ci);
+    return RowRanges.create(rowCount, pageIndexes, oi);
   }
 
   @Override
@@ -219,5 +224,36 @@ public class ColumnIndexFilter implements Visitor<RowRanges> {
   public RowRanges visit(Not not) {
     throw new IllegalArgumentException(
         "Predicates containing a NOT must be run through LogicalInverseRewriter. " + not);
+  }
+
+  /**
+   * Validates that column index and offset index metadata are consistent and can be used safely.
+   *
+   * @param columnIndex the column index to validate
+   * @param offsetIndex the offset index to validate
+   * @param columnPath the column path for error reporting
+   * @return true if metadata is valid and safe to use, false if corrupt and should be ignored
+   */
+  private static boolean isValidMetadata(ColumnIndex columnIndex, OffsetIndex offsetIndex, ColumnPath columnPath) {
+    if (columnIndex == null || offsetIndex == null) {
+      return true;
+    }
+
+    int columnIndexSize = columnIndex.getMinValues().size();
+    int offsetIndexSize = offsetIndex.getPageCount();
+
+    if (columnIndexSize != offsetIndexSize) {
+      LOGGER.warn(
+          "Column index and offset index size mismatch for column {}: "
+              + "column index has {} entries but offset index has {} pages. "
+              + "This indicates corrupted metadata from the writer. "
+              + "Ignoring column index for filtering to avoid errors.",
+          columnPath,
+          columnIndexSize,
+          offsetIndexSize);
+      return false;
+    }
+
+    return true;
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
@@ -192,7 +192,7 @@ public class ColumnIndexFilter implements Visitor<RowRanges> {
       return allRows();
     }
 
-    if (!isValidMetadata(ci, oi, columnPath)) {
+    if (!isValidIndexSize(ci, oi, columnPath)) {
       return allRows();
     }
 
@@ -234,10 +234,7 @@ public class ColumnIndexFilter implements Visitor<RowRanges> {
    * @param columnPath the column path for error reporting
    * @return true if metadata is valid and safe to use, false if corrupt and should be ignored
    */
-  private static boolean isValidMetadata(ColumnIndex columnIndex, OffsetIndex offsetIndex, ColumnPath columnPath) {
-    if (columnIndex == null || offsetIndex == null) {
-      return true;
-    }
+  private static boolean isValidIndexSize(ColumnIndex columnIndex, OffsetIndex offsetIndex, ColumnPath columnPath) {
 
     int columnIndexSize = columnIndex.getMinValues().size();
     int offsetIndexSize = offsetIndex.getPageCount();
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
```diff
@@ -220,4 +225,35 @@ public RowRanges visit(Not not) {
...
+   *
+   * @param columnIndex the column index to validate
+   * @param offsetIndex the offset index to validate
+   * @param columnPath the column path for error reporting
+   * @return true if metadata is valid and safe to use, false if corrupt and should be ignored
+   */
+  private static boolean isValidMetadata(ColumnIndex columnIndex, OffsetIndex offsetIndex, ColumnPath columnPath) {
+    if (columnIndex == null || offsetIndex == null) {
+      return true;
+    }
```
gszadovszky: Based on the code it seems they cannot be null at this stage.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnIndexFiltering.java
```diff
@@ -650,4 +660,75 @@ public void testFilteringWithProjection() throws IOException {
             false,
             true));
   }
+
+  @Test
+  public void testValidMetadata() throws Exception {
```
wgtmac: Shouldn't we test invalid page index?

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
```diff
@@ -220,4 +225,35 @@ public RowRanges visit(Not not) {
...
+
+  /**
+   * Validates that column index and offset index metadata are consistent and can be used safely.
+   *
+   * @param columnIndex the column index to validate
+   * @param offsetIndex the offset index to validate
+   * @param columnPath the column path for error reporting
+   * @return true if metadata is valid and safe to use, false if corrupt and should be ignored
+   */
+  private static boolean isValidMetadata(ColumnIndex columnIndex, OffsetIndex offsetIndex, ColumnPath columnPath) {
```
wgtmac: It might be misleading to readers that this function does not validate everything but just `columnIndex.getMinValues().size()` vs `offsetIndex.getPageCount()`. But I'm not sure if we want to pay for the cost of validating everything.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
ArnavBalyan: updated thanks

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/ColumnIndexFilter.java
ArnavBalyan: updated to `isValidIndexSize`, was initially planned to be a general pre-check which can be enhanced in the future. For now restricting to index size.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnIndexFiltering.java
ArnavBalyan: we need the buggy parquet-go lib to create the invalid index, current check is in the column read flow, will only fail if the upstream wrote invalid data, is there a better way to test it?

gszadovszky (approved):
LGTM.
@wgtmac, could you check if you're OK with the answers/changes?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnIndexFiltering.java
wgtmac: If this test does not cover the new case, perhaps we can simply revert all changes in this file?

A better way would be uploading the corrupted data produced by parquet-go to https://github.com/apache/parquet-testing/tree/master/bad_data. But I do not think it worths the effort...

ArnavBalyan: updated thanks

PR #3329, opened 2025-09-24
MINOR: parquet-avro tests should not debug to stderr

dossett:
### Rationale for this change
Logging debug information to stderr creates noise, especially when trying to run tests with maven's quite mode (`-q`).  stderr is sometime appealing because stdout of tests can be suppressed by default.  I was initially going to change these statements to print to stdout but I think they can be removed entirely because any assertion failures would show the schemas. 

### Are these changes tested?
N/A, but tests will still pass

NO


Fokko (approved):
Thanks for cleaning this up @dossett 🙌

Issue #3338, opened 2025-09-29
Support encrypted files for show bloom filter command


ArnavBalyan:
### Describe the enhancement requested

 - Currently show bloom filter does not support encrypted parquet files.
 - Support reading and showing bloom filter metadata for encrypted files

PR #3339, opened 2025-09-29
GH-3338: Support encrypted files for Parquet CLI commands

ArnavBalyan:
### Rationale for this change
 - Currently show bloom filter does not support encrypted parquet files.
 - Support reading and showing bloom filter metadata for encrypted files.
 - Added support to read encrypted files within base command and supported for main parquet CLI commands: ShowBloomFilter, ShowPages, ParquetMetadata. 

### Are these changes tested?
 - Yes

 - Yes

Closes: #3338

### DIFF
--- a/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
+++ b/parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
@@ -34,8 +34,6 @@ import org.apache.parquet.hadoop.BloomFilterReader;
 import org.apache.parquet.hadoop.ParquetFileReader;
 import org.apache.parquet.hadoop.metadata.BlockMetaData;
 import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
-import org.apache.parquet.hadoop.util.HadoopInputFile;
-import org.apache.parquet.io.InputFile;
 import org.apache.parquet.io.api.Binary;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.PrimitiveType;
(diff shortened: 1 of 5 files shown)


parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
```diff
@@ -63,14 +67,31 @@ public ShowBloomFilterCommand(Logger console) {
...
+  @Parameter(
+      names = {"--footer-key"},
+      description = "Footer key for encrypted files (hex string, 16/24/32 bytes)")
+  String footerKeyHex;
+
+  @Parameter(
+      names = {"--column-key"},
+      description = "Column key for encrypted files (hex string, 16/24/32 bytes)")
+  String columnKeyHex;
+
```
gszadovszky: @ArnavBalyan, is it only bloom filters? Shouldn't we add this at a more common level, so other commands would also benefit from it?

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
```diff
@@ -63,14 +67,31 @@ public ShowBloomFilterCommand(Logger console) {
...
 
+  @Parameter(
+      names = {"--footer-key"},
+      description = "Footer key for encrypted files (hex string, 16/24/32 bytes)")
+  String footerKeyHex;
+
+  @Parameter(
+      names = {"--column-key"},
+      description = "Column key for encrypted files (hex string, 16/24/32 bytes)")
+  String columnKeyHex;
```
gszadovszky: Cannot the column key be different for each column?

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
ArnavBalyan: Moved this to base, any CLI command can leverage this. I tested and supported 3 commands for now. 
Will add UTs and support for the remaining CLI commands shortly. Each command should now be able to read without duplication

parquet-cli/src/main/java/org/apache/parquet/cli/commands/ShowBloomFilterCommand.java
ArnavBalyan: Ah good point, updated to follow the existing key format within PropertiesDrivenCryptoFactory

gszadovszky (approved):
This is a great improvement! Thanks, @ArnavBalyan.

gszadovszky:
> cc @gszadovszky could you please merge if all looks good thanks!

Thanks for pinging, sorry for the delay.

Issue #2815, opened 2023-09-19
Create Configuration key for enabling Byte Stream Split Encoding in ParquetWRiter


asfimport:
All of the properties in [ParquetWriter](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java) have an associated Configuration key (for example, [ParquetOutputFormat.DICTIONARY_PAGE_SIZE](https://github.com/apache/parquet-mr/blob/910bcc4edc2d707670e02e9ceadd98dacd9f08d2/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java#L140) corresponds to ParquetWriter#withDictionaryPageSize), except for `ParquetWriter#withByteStreamSplitEncoding`.

 

Can we add a Configuration key for this? Happy to make a PR, given some input on naming convention (`parquet.encoding.bytestreamsplit.enabled` maybe?)

 

 

 

**Reporter**: [Claire McGinty](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=clairemcginty) / @clairemcginty

<sub>**Note**: *This issue was originally created as [PARQUET-2350](https://issues.apache.org/jira/browse/PARQUET-2350). Please see the [migration documentation](https://issues.apache.org/jira/browse/PARQUET-2502) for further details.*</sub>

PR #3340, opened 2025-09-30
GH-2815: Allow bytestreamsplit available via Hadoop Configuration

ArnavBalyan:
### Rationale for this change
 - `ParquetWriter#withByteStreamSplitEncoding` is available programatically but no config option exists to enable/disable it.
 - Added config to allow users to configure it through properties similar to other confs.
 - Closes: #2815 

### Are these changes tested?
 - Yes

 - Yes


gszadovszky (approved):
The title is a bit misleading. ByteStreamSplit is in fact already configurable through ParquetProperties. What this PR is about is to make the feature available via Hadoop Configuration. Otherwise, LGTM.

Issue #2607, opened 2021-04-20
Make "null decryptor" exception more informative


asfimport:
Forgetting to pass decryption properties when reading an encrypted column in files with plaintext footer, results in a "null decryptor" exception thrown in the ColumnChunkMetaData class. The exception text can/should be updated to point to the possible reason.

**Reporter**: [Gidon Gershinsky](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gershinsky) / @ggershinsky
**Assignee**: [Gidon Gershinsky](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gershinsky) / @ggershinsky

<sub>**Note**: *This issue was originally created as [PARQUET-2033](https://issues.apache.org/jira/browse/PARQUET-2033). Please see the [migration documentation](https://issues.apache.org/jira/browse/PARQUET-2502) for further details.*</sub>

PR #3341, opened 2025-09-30
GH-2607: Improve "null decryptor" exception in column chunk metadata

ArnavBalyan:
### Rationale for this change
 - Closes: #2607
 - Users get cryptic "null decryptor" exception when decryption properties are not available/decryptor could not be created.
 - Improve exception and add more actionable error message for reader.

### Are these changes tested?
 - Yes / CI

 - Yes


gszadovszky (changes requested):
I think the error message is too detailed related to how the decryption properties can be set. I think there are more options to do this.
I see the point of providing actionable error messages but it is more the job of the application and not the library. Let's keep only the actual message that no decryption properties were provided for the encrypted column.

PR #3351, opened 2025-10-23
GH-3350: Avoid flushing data to cloud when exception is thrown

### LINKED ISSUE #3350: Avoid flushing data to cloud when exception is thrown in parquet writer close

Inside the InternalParquetRecordWriter::Close finally block, we call close on parquetFileWriter, which may cause incomplete data to be flushed to the cloud if an exception is thrown during the close.

This bug is introduced to parquet writer by this [commit](https://github.com/apache/parquet-java/commit/274dc51bc9e5cc880ba3c77c3db826d2a4943965) and was backported up to 1.14.0.

### Component(s)

_No response_

Jiayi-Wang-db:
### Rationale for this change
Inside the `InternalParquetRecordWriter::Close` finally block, we call close on parquetFileWriter, which may cause incomplete data to be flushed to the cloud if an exception is thrown during the close .

Remove `parquetFileWriter.close` out of finally block and added a unit test.
### Are these changes tested?

Yes.

Users wouldn't get incomplet parquet files because of torn writes.

 Closes #3350

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -654,15 +654,16 @@ public class ParquetFileWriter implements AutoCloseable {
       currentChunkFirstDataPage = -1;
       compressedLength = 0;
       uncompressedLength = 0;
-      // The statistics will be copied from the first one added at writeDataPage(s) so we have the correct typed one
+      // The statistics will be copied from the first one added at writeDataPage(s) so we have the correct typed
+      // one
       currentStatistics = null;
       currentSizeStatistics = SizeStatistics.newBuilder(
               descriptor.getPrimitiveType(),
               descriptor.getMaxRepetitionLevel(),
               descriptor.getMaxDefinitionLevel())
           .build();
-      currentGeospatialStatistics =
-          GeospatialStatistics.newBuilder(descriptor.getPrimitiveType()).build();
+      currentGeospatialStatistics = GeospatialStatistics.newBuilder(descriptor.getPrimitiveType())
+          .build();
 
       columnIndexBuilder = ColumnIndexBuilder.getBuilder(currentChunkType, columnIndexTruncateLength);
       offsetIndexBuilder = OffsetIndexBuilder.getBuilder();
@@ -715,7 +716,9 @@ public class ParquetFileWriter implements AutoCloseable {
       this.uncompressedLength += uncompressedSize + headerSize;
       this.compressedLength += compressedPageSize + headerSize;
       LOG.debug("{}: write dictionary page content {}", out.getPos(), compressedPageSize);
-      dictionaryPage.getBytes().writeAllTo(out); // for encrypted column, dictionary page bytes are already encrypted
+      dictionaryPage
+          .getBytes()
+          .writeAllTo(out); // for encrypted column, dictionary page bytes are already encrypted
       encodingStatsBuilder.addDictEncoding(dictionaryPage.getEncoding());
       currentEncodings.add(dictionaryPage.getEncoding());
     });
@@ -1345,7 +1348,8 @@ public class ParquetFileWriter implements AutoCloseable {
       int rlByteLength = toIntWithCheck(repetitionLevels.size(), "page repetition levels");
       int dlByteLength = toIntWithCheck(definitionLevels.size(), "page definition levels");
 
-      int compressedSize = toIntWithCheck(bytes.size() + repetitionLevels.size() + definitionLevels.size(), "page");
+      int compressedSize =
+          toIntWithCheck(bytes.size() + repetitionLevels.size() + definitionLevels.size(), "page");
 
       int uncompressedSize =
           toIntWithCheck(uncompressedDataSize + repetitionLevels.size() + definitionLevels.size(), "page");
@@ -1519,7 +1523,8 @@ public class ParquetFileWriter implements AutoCloseable {
         // write bloom filter if one of data pages is not dictionary encoded
         boolean isWriteBloomFilter = false;
         for (Encoding encoding : dataEncodings) {
-          // dictionary encoding: `PLAIN_DICTIONARY` is used in parquet v1, `RLE_DICTIONARY` is used in parquet v2
+          // dictionary encoding: `PLAIN_DICTIONARY` is used in parquet v1, `RLE_DICTIONARY` is used in
+          // parquet v2
           if (encoding != Encoding.PLAIN_DICTIONARY && encoding != Encoding.RLE_DICTIONARY) {
             isWriteBloomFilter = true;
             break;
@@ -1739,7 +1744,8 @@ public class ParquetFileWriter implements AutoCloseable {
         }
         length += chunk.getTotalSize();
 
-        if ((i + 1) == columnsInOrder.size() || columnsInOrder.get(i + 1).getStartingPos() != (start + length)) {
+        if ((i + 1) == columnsInOrder.size()
+            || columnsInOrder.get(i + 1).getStartingPos() != (start + length)) {
           // not contiguous. do the copy now.
           copy(from, out, start, length);
           // reset to start at the next column chunk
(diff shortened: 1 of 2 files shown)


wgtmac:
This looks reasonable to me. WDYT? @gszadovszky @Fokko

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -140,8 +140,9 @@ public void close() throws IOException, InterruptedException {
         }
         finalMetadata.putAll(finalWriteContext.getExtraMetaData());
         parquetFileWriter.end(finalMetadata);
+        AutoCloseables.uncheckedClose(parquetFileWriter);
```
gszadovszky: `parquetFileWriter.end(Map<String, String>)` invokes `parquetFileWriter.close()`, so this call is not necessary. Why we added the `close()` call to `finally` to ensure the related buffers are freed (required for direct memory handling).
Maybe introduce a similar logic of having an `aborted` flag in `ParquetFileWriter` just like we have done it for `InternalParquetRecordWriter`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1812,6 +1813,8 @@ public void end(Map<String, String> extraMetaData) throws IOException {
       LOG.debug("{}: end", out.getPos());
       this.footer = new ParquetMetadata(new FileMetaData(schema, extraMetaData, Version.FULL_VERSION), blocks);
       serializeFooter(footer, out, fileEncryptor, metadataConverter);
+    } catch (Exception e) {
+      aborted = true;
```
gszadovszky: We do not want to swallow the exception, just set the flag and re-throw.

gszadovszky: We should probably do the same pattern for every public method that may throw an exception.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -140,8 +140,9 @@ public void close() throws IOException, InterruptedException {
...
       } finally {
-        AutoCloseables.uncheckedClose(columnStore, pageStore, bloomFilterWriteStore, parquetFileWriter);
+        AutoCloseables.uncheckedClose(columnStore, pageStore, bloomFilterWriteStore);
```
gszadovszky: Now that we have `ParquetFileWriter` to handle the "aborted" state, this change can be reverted.

Jiayi-Wang-db: Yes, I haven't finish my change.

gszadovszky: Sorry, then. I was too fast. :smile:
Ping me when you're ready.

Jiayi-Wang-db: Thanks for the quick review!

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
Jiayi-Wang-db: I’m not familiar with the direct buffer change, but in `InternalParquetRecordWriter`, there’s only one place where aborted is marked. Is that the only place that could cause an aborted write? If so, we don’t need to apply the same pattern to every public method in `ParquetFileWriter`.

Jiayi-Wang-db: It does look that way. The `write` function in `InternalParquetRecordWriter` is the only public function that can throw an exception (except close). So after we mark it as aborted there and abort the file write in the close call, we should cover all cases.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1812,22 +1813,35 @@ public void end(Map<String, String> extraMetaData) throws IOException {
...
       serializeFooter(footer, out, fileEncryptor, metadataConverter);
+    } catch (IOException e) {
+      abort();
+      throw e;
     } finally {
       close();
     }
   }
 
+  /* Mark the writer as aborted to avoid flushing incomplete data to the cloud. */
```
gszadovszky: nit: "to the cloud" is not required. That is only one use-case.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -1812,22 +1813,35 @@ public void end(Map<String, String> extraMetaData) throws IOException {
...
     }
-    try (PositionOutputStream temp = out) {
-      temp.flush();
+
+    try {
+      if (!aborted && out != null) {
+        out.flush();
+      }
+    } catch (IOException e) {
+      throw e;
```
gszadovszky: This is not required.

Jiayi-Wang-db:
> For a more complete fix it would be nicer to handle this case as well.

It ends up a bit messy and ugly. Let me know if you have better advice to cover all the cases. @gszadovszky

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -335,6 +336,34 @@ public ParquetFileWriter(OutputFile file, MessageType schema, Mode mode, long ro
...
+
+  private void withAbortOnFailure(IORunnable action) throws IOException {
+    try {
+      action.run();
+    } catch (IOException e) {
+      aborted = true;
+      throw e;
+    }
+  }
+
```
gszadovszky, line 366: Maybe, you could move these to a specific utility. At least the functional interfaces. But, with an additional "abort action" in the parameter, even the withAbort functions as well.

gszadovszky (changes requested):
You also have some code format issues.

I don't think your solution is ugly. I think you've made the best out of it. @wgtmac, do you have a better idea?

Meanwhile, this change only prevents additional writes after an exception happens. If we write e.g. multiple row groups and no I/O issue happens during the flushing of the first one, we would still have garbage on the target if the writing of the second row group fails, right? (We practically cannot prevent that because we don't want to keep the whole file in memory.)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
Jiayi-Wang-db: Can we leave them as they are to make it simple. And maybe extract them later if we want to use the same pattern in other files?I

gszadovszky: Sure, that makes sense

gszadovszky:
> > If we write e.g. multiple row groups and no I/O issue happens during the flushing of the first one, we would still have garbage on the target if the writing of the second row group fails, right? (We practically cannot prevent that because we don't want to keep the whole file in memory.)
> 
> @gszadovszky Digging into this further, I think the statement is not correct. The function you mentioned, `InternalParquetRecordWriter::flushRowGroupToStore`, is called when the in-memory records are about to exceed the limit. However, the data is written into the stream, but flush/close is not called, so the data is not visible to other readers. Only the final `close()` call in `ParquetFileWriter` triggers a `FSOutputFileStream` close, which finalizes the file. Therefore, our fix addresses all of the potential incomplete parquet files.

Thanks for the clarification, @Jiayi-Wang-db. It makes sense. I'll approve this but will wait for next week for other potential feedback.

gszadovszky:
@Jiayi-Wang-db, since it is a bug, it makes sense to backport the fix to the last two release branches: `parquet-1.16.x` and `parquet-1.15.x`. I do not feel the urgency though, to actually make releases only for this one. But it'll be included when we make them.

Issue #3358, opened 2025-11-19
Add Configurable Thrift Max Message Size for Parquet Metadata Reading


cravani:
### Describe the enhancement requested

When reading Parquet files with large metadata (e.g., files with thousands of columns), the default Thrift message size limit can be insufficient, causing TTransportException: Message size exceeds limit errors. Currently, the Thrift protocol configuration uses default max message size (100MB), preventing users from reading files with exceptionally large metadata footers.

There is similar fix for PyArrow: https://issues.apache.org/jira/browse/ARROW-16546

Exception while reading large footer message
```
Caused by: java.io.IOException: can not read FileMetaData: MaxMessageSize reached
	at org.apache.parquet.format.Util.readFileMetaData(Util.java:396)
	at org.apache.parquet.format.Util.readFileMetaData(Util.java:208)
	at org.apache.parquet.format.converter.ParquetMetadataConverter$3.visit(ParquetMetadataConverter.java:1581)
	at org.apache.parquet.format.converter.ParquetMetadataConverter$3.visit(ParquetMetadataConverter.java:1567)
	at org.apache.parquet.format.converter.ParquetMetadataConverter$SkipMetadataFilter.accept(ParquetMetadataConverter.java:1285)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.readParquetMetadata(ParquetMetadataConverter.java:1567)
	at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:629)
	at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:934)
	at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:925)
	at org.apache.parquet.hadoop.ParquetFileReader.open(ParquetFileReader.java:710)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetFooterReader.readFooter(ParquetFooterReader.java:85)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetFooterReader.readFooter(ParquetFooterReader.java:76)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat$.$anonfun$readParquetFootersInParallel$1(ParquetFileFormat.scala:451)
	at org.apache.spark.util.ThreadUtils$.$anonfun$parmap$2(ThreadUtils.scala:416)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:687)
[...]

PR #3359, opened 2025-11-19
GH-3358: Add Configurable Thrift Max Message Size for Parquet Metadata Reading

cravani:
### Rationale for this change
When reading Parquet files with large metadata (e.g., files with thousands of columns), the default Thrift message size limit can be insufficient, causing TTransportException: Message size exceeds limit errors. Currently, the Thrift protocol configuration uses default max message size (100MB), preventing users from reading files with exceptionally large metadata footers.

Add a new configuration key: parquet.thrift.string.size.limit
Default value: 100 MB (104857600 bytes)
Allow users to override this via Configuration

### Are these changes tested?
Yes and added a Test case [TestParquetFileReaderMaxMessageSize.java](https://github.com/apache/parquet-java/blob/e8de6b55842b63a00d35b520811600dc407250cd/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileReaderMaxMessageSize.java)

Not by default, user can set config parquet.thrift.string.size.limit=<limit> to increase it based on need.

Closes #GH-3358

### DIFF
--- a/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
+++ b/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
@@ -45,6 +45,7 @@ import org.apache.parquet.format.event.TypedConsumer.I32Consumer;
 import org.apache.parquet.format.event.TypedConsumer.I64Consumer;
 import org.apache.parquet.format.event.TypedConsumer.StringConsumer;
 import org.apache.thrift.TBase;
+import org.apache.thrift.TConfiguration;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TCompactProtocol;
 import org.apache.thrift.protocol.TProtocol;
@@ -59,6 +60,7 @@ import org.apache.thrift.transport.TTransportException;
 public class Util {
 
   private static final int INIT_MEM_ALLOC_ENCR_BUFFER = 100;
+  private static final int DEFAULT_MAX_MESSAGE_SIZE = 104857600; // 100 MB
 
   public static void writeColumnIndex(ColumnIndex columnIndex, OutputStream to) throws IOException {
     writeColumnIndex(columnIndex, to, null, null);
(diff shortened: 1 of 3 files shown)


parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -358,26 +391,56 @@ public void consume(RowGroup rowGroup) {
...
-  private static InterningProtocol protocol(TIOStreamTransport t) {
+  private static InterningProtocol protocol(TIOStreamTransport t, int configuredMaxMessageSize)
+      throws TTransportException, NumberFormatException {
+
+    int maxMessageSize = 104857600; // Default to 100 MB
+    if (configuredMaxMessageSize > 0) {
+      maxMessageSize = configuredMaxMessageSize;
+    } else if (configuredMaxMessageSize == -1) {
+      // Set to default 100 MB
+      maxMessageSize = 104857600;
```
Fokko: I think we set the default here twice. Can we break out the default to a static constant?

cravani: Updated the code to use Static default value and remove duplicate assignment.

parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
```diff
@@ -358,26 +392,54 @@ public void consume(RowGroup rowGroup) {
...
-  private static InterningProtocol protocol(TIOStreamTransport t) {
+  private static InterningProtocol protocol(TIOStreamTransport t, int configuredMaxMessageSize)
+      throws TTransportException, NumberFormatException {
+    int maxMessageSize = configuredMaxMessageSize;
+    if (configuredMaxMessageSize == -1) {
+      // Set to default 100 MB
+      maxMessageSize = DEFAULT_MAX_MESSAGE_SIZE;
+    }
+    if (configuredMaxMessageSize < -1 || configuredMaxMessageSize == 0) {
+      throw new NumberFormatException("Invalid max message size: " + configuredMaxMessageSize);
```
Fokko:
```suggestion
    if (configuredMaxMessageSize <= 0) {
      throw new NumberFormatException("Max message size must be positive: " + configuredMaxMessageSize);
```

PR #3368, opened 2025-12-08
GH-3344: Adaptive compression for v2 page

### LINKED ISSUE #3344: Make compression adaptive with V2 data pages

When writing a V2 data page, it seems that compression is always unconditionally enabled even when compression doesn't actually yield any benefits:
https://github.com/apache/parquet-java/blob/0fea3e1e22fffb0a25193e3efb9a5d090899458a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L305-L311

It would be relatively easy to use a hardcoded threshold above which compression is disabled, which makes reading faster.

### Component(s)

Core

ConeyLiu:
### Rationale for this change

Closes #3344

### Are these changes tested?

Yes, added UT.

Yes, add new configuration. No breaking changes.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -849,12 +849,18 @@ public class ParquetProperties {
      *
      * <p>When the compression ratio (compressed size / uncompressed size) exceeds this threshold,
      * the uncompressed data will be used instead. For example, with a threshold of 0.98, if
-     * compression only saves 2% of space, the data will not be compressed.
+     * compression saves less than 2% of space, the data will not be compressed.
+     * A value of 0 disables data page compression and a value of 1 only skips compression when it
+     * increases the data size.
      *
      * @param threshold the compression ratio threshold, default is {@value #DEFAULT_PAGE_COMPRESS_THRESHOLD}
      * @return this builder for method chaining
      */
     public Builder withPageCompressThreshold(double threshold) {
+      Preconditions.checkArgument(
+          Double.isFinite(threshold) && threshold >= 0.0 && threshold <= 1.0,
+          "Invalid page compression threshold: %s. It must be a finite value between 0.0 and 1.0.",
+          threshold);
       this.pageCompressThreshold = threshold;
       return this;
     }
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -69,6 +69,7 @@ public class ParquetProperties {
   public static final boolean DEFAULT_SIZE_STATISTICS_ENABLED = true;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
+  public static final double DEFAULT_PAGE_COMPRESS_THRESHOLD = 0.98;
 
   /**
    * @deprecated This shared instance can cause thread safety issues when used by multiple builders concurrently.
@@ -121,6 +122,7 @@ public class ParquetProperties {
   private final int statisticsTruncateLength;
   private final boolean statisticsEnabled;
   private final boolean sizeStatisticsEnabled;
+  private final double pageCompressThreshold;
 
   // The expected NDV (number of distinct values) for each columns
   private final ColumnProperty<Long> bloomFilterNDVs;
@@ -157,6 +159,8 @@ public class ParquetProperties {
     this.statisticsTruncateLength = builder.statisticsTruncateLength;
     this.statisticsEnabled = builder.statisticsEnabled;
     this.sizeStatisticsEnabled = builder.sizeStatisticsEnabled;
+    this.pageCompressThreshold = builder.pageCompressThreshold;
+
     this.bloomFilterNDVs = builder.bloomFilterNDVs.build();
     this.bloomFilterFPPs = builder.bloomFilterFPPs.build();
     this.bloomFilterEnabled = builder.bloomFilterEnabled.build();
@@ -327,6 +331,10 @@ public class ParquetProperties {
     return pageWriteChecksumEnabled;
   }
 
+  public double pageCompressThreshold() {
+    return pageCompressThreshold;
+  }
+
   public OptionalLong getBloomFilterNDV(ColumnDescriptor column) {
     Long ndv = bloomFilterNDVs.getValue(column);
     return ndv == null ? OptionalLong.empty() : OptionalLong.of(ndv);
@@ -415,7 +423,8 @@ public class ParquetProperties {
         + "Page row count limit to " + getPageRowCountLimit() + '\n'
         + "Writing page checksums is: " + (getPageWriteChecksumEnabled() ? "on" : "off") + '\n'
         + "Statistics enabled: " + statisticsEnabled + '\n'
-        + "Size statistics enabled: " + sizeStatisticsEnabled;
+        + "Size statistics enabled: " + sizeStatisticsEnabled + '\n'
+        + "Page compress threshold: " + pageCompressThreshold;
     String perColumn = "";
     if (!columnCodecs.toString().equals(Objects.toString(columnCodecs.getDefaultValue()))) {
       perColumn = "Per-column codecs: " + columnCodecs;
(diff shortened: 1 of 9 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java
```diff
@@ -316,7 +316,130 @@ public void testColumnOrderV1() throws IOException {
...
+        schema,
+        col,
+        rowCount,
+        nullCount,
+        valueCount,
+        repetitionLevels,
+        definitionLevels,
+        data,
+        statistics,
+        10);
```
pitrou: It does not make sense to have a threshold greater than 1.0, does it?
In any case, the data of a column full of zeros is extremely compressible (you might want to disable dictionary encoding?), so the default value should work IMHO.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java
```diff
@@ -316,7 +316,130 @@ public void testColumnOrderV1() throws IOException {
...
+    ParquetMetadata footer = ParquetFileReader.readFooter(conf, file, NO_FILTER);
+    ColumnChunkMetaData columnMeta = footer.getBlocks().get(0).getColumns().get(0);
+
+    long uncompressedSize = columnMeta.getTotalUncompressedSize();
+    long compressedSize = columnMeta.getTotalSize();
+
+    assertEquals(
+        "Data should be stored uncompressed when compression ratio exceeds threshold",
+        uncompressedSize,
+        compressedSize);
```
pitrou: Perhaps you want to read the data back and check it's as expected?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -756,6 +767,21 @@ public Builder withSizeStatisticsEnabled(String columnPath, boolean enabled) {
...
+     * Sets the compression threshold for V2 data pages.
+     *
+     * <p>When the compression ratio (compressed size / uncompressed size) exceeds this threshold,
+     * the uncompressed data will be used instead. For example, with a threshold of 0.98, if
+     * compression only saves 2% of space, the data will not be compressed.
+     *
+     * @param threshold the compression ratio threshold, default is {@value #DEFAULT_V2_PAGE_COMPRESS_THRESHOLD}
+     * @return this builder for method chaining
+     */
+    public Builder withV2PageCompressThreshold(double threshold) {
```
pitrou: I'm not sure it makes sense to put "V2" in the API name? For users it doesn't seem be relevant (just keep it mentioned in the docstrings).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java
ConeyLiu: Added

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageWriteStore.java
ConeyLiu: Changed to 1.0

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
ConeyLiu: Good point, updated

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/rewrite/ParquetRewriter.java
```diff
@@ -763,7 +769,7 @@ private void processChunk(
               dlLevels,
               converter.getEncoding(headerV2.getEncoding()),
               BytesInput.from(pageLoad),
-              headerV2.is_compressed,
+              compressed,
```
ConeyLiu, line 772: Fixed a potential bug

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -68,6 +68,7 @@ public class ParquetProperties {
   public static final boolean DEFAULT_SIZE_STATISTICS_ENABLED = true;
 
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
+  public static final double DEFAULT_PAGE_COMPRESS_THRESHOLD = 0.98;
```
wgtmac, line 72: Why is this magic number? Is it better to use a smaller number like 0.9 or 0.85?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -694,4 +731,63 @@ public void flushToFileWriter(ParquetFileWriter writer) throws IOException {
       pageWriter.writeToFileWriter(writer);
     }
   }
+
+  public static Builder build(BytesInputCompressor compressor, MessageType schema, ByteBufferAllocator allocator) {
```
wgtmac: Is it better to use a separate issue to implement this, though I'm fine to keep it as is.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
ConeyLiu: Per the suggestion linked here: https://github.com/apache/parquet-java/issues/3344#issuecomment-3431666491

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
ConeyLiu: Now, the ColumnChunkPageWriteStore implements a similar builder pattern. Just add a new option method.

ConeyLiu:
Sorry this PR hasn’t been updated for so long due to work. Please review it when you get a chance. @wgtmac

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -313,9 +316,14 @@ public void writePageV2(
       boolean compressed = false;
       BytesInput compressedData = BytesInput.empty();
       if (data.size() > 0) {
-        // TODO: decide if we compress
         compressedData = compressor.compress(data);
         compressed = true;
+        double compressionRatio = (double) compressedData.size() / data.size();
+        if (compressor.getCodecName() != CompressionCodecName.UNCOMPRESSED
+            && compressionRatio > pageCompressThreshold) {
```
wgtmac: This fallback can break for one-shot `BytesInput` implementations. `compressor.compress(data)` may consume `data` (for example stream-backed input), and then the fallback assigns the already-consumed input back to `compressedData` for CRC/writing. Please materialize/copy the original page bytes before compression, or otherwise make the fallback use a replayable copy. (Reviewed by Codex)

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -833,6 +844,21 @@ public Builder withCompressionLevel(String columnPath, Integer level) {
...
+     * Sets the compression threshold for data pages, only effect for V2 pages.
+     *
+     * <p>When the compression ratio (compressed size / uncompressed size) exceeds this threshold,
+     * the uncompressed data will be used instead. For example, with a threshold of 0.98, if
+     * compression only saves 2% of space, the data will not be compressed.
+     *
+     * @param threshold the compression ratio threshold, default is {@value #DEFAULT_PAGE_COMPRESS_THRESHOLD}
+     * @return this builder for method chaining
+     */
+    public Builder withPageCompressThreshold(double threshold) {
```
wgtmac, line 859: Please validate the threshold or document the allowed range here. As written, values like `NaN`, negative numbers, or `100` are accepted even though the API describes this as a compressed/uncompressed ratio threshold. I would either enforce a finite range (for example `0.0 <= threshold <= 1.0`) or explicitly document the special meanings of out-of-range values. (Reviewed by Codex)

ConeyLiu: Added validation

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
ConeyLiu: Materialized data before compression.

PR #3370, opened 2025-12-17
Allow reading dictionary encoded boolean

Fokko:
### Rationale for this change

I've observed some Parquet files in the wild that contain dictionary-encoded boolean values, which is also wild.

I don't think we want to allow producing this, but I think it would be good to allow reading this. We don't judge.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/PlainValuesDictionary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/PlainValuesDictionary.java
@@ -331,7 +331,7 @@ public abstract class PlainValuesDictionary extends Dictionary {
 
     @Override
     public String toString() {
-      StringBuilder sb = new StringBuilder("PlainIntegerDictionary {\n");
+      StringBuilder sb = new StringBuilder("PlainBooleanDictionary {\n");
       for (int i = 0; i < boolDictionaryContent.length; i++) {
         sb.append(i).append(" => ").append(boolDictionaryContent[i]).append("\n");
       }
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/PlainValuesDictionary.java
```diff
@@ -300,4 +301,46 @@ public int getMaxId() {
...
+    }
+
+    @Override
+    public boolean decodeToBoolean(int id) {
+      return boolDictionaryContent[id];
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder("PlainIntegerDictionary {\n");
```
wgtmac: Is `PlainIntegerDictionary` correct?

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/TestDictionary.java
```diff
@@ -24,7 +24,7 @@
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.DOUBLE;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FLOAT;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT32;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
wgtmac: Let's avoid wildcard import.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/PlainValuesDictionary.java
Fokko: Copy paste, thanks 👍 😁

parquet-column/src/test/java/org/apache/parquet/column/values/dictionary/TestDictionary.java
Fokko: I agree, this was the work of my editor.

PR #3378, opened 2025-12-29
[parquet-thrift] move on from java 8

dossett:
Now that java 8 is no longer supported, we can use language features introduced in later versions and get ahead of other changes for subsequent upgrades.  I'm trying this in one module as a test of automating this with codex.  Of the four points below the first seems like it should definitely be addressed. The others are potentially useful, but not pressing.

Codex generated summary:

  - Replaced deprecated reflection APIs with getDeclaredConstructor().newInstance() and unified exception handling, enabled by Java 9+ deprecations now acceptable with Java 11 baseline.
  - Adopted List.of and List.copyOf for immutable collections, removing verbose Arrays.asList/manual wrappers and leveraging Java 9/10 collection factories.
  - Simplified null-defaulting with Objects.requireNonNullElse and removed redundant null checks made unnecessary by Objects.requireNonNull, both available in Java 9+.
  - Modernized small collection usages (e.g., ArrayDeque for queues, diamond operator cleanup), consistent with Java 11 APIs and performance best practices.


parquet-thrift/src/main/java/org/apache/parquet/thrift/struct/ThriftType.java
```diff
@@ -502,12 +502,7 @@ public EnumType(@JsonProperty("values") List<EnumValue> values) {
     }
 
     public Iterable<EnumValue> getValues() {
-      return new Iterable<EnumValue>() {
-        @Override
-        public Iterator<EnumValue> iterator() {
-          return values.iterator();
-        }
-      };
+      return List.copyOf(values);
```
Fokko: Would this impose any regression on performance? Returning an iterator (view) looks cheaper than taking a copy of the values.

dossett: @Fokko That's probably true, but in practice are EnumTypes ever large enough to matter?  I don't know.

What about returning `Collections.unmodifiableList(values)` ? I believe that would still return a view but be more defensive about preventing modification of the underlying data.

dossett: We could also make `private final List<EnumValue> values;` a `unmodifiableList` from the start and just directly return that in getValues().  Although maybe that's technically a breaking change if [someone was relying](https://www.hyrumslaw.com/) on being able to indirectly update the list.

Or the simplest change to keep current behavior but express it in a Java 11 idiom would be `return () -> values.iterator();`

I'm ok with any of these options!

dossett: Update this to just `return values::iterator` since that's in keeping with the spirit of the PR (java modernization). Other changes could be discussed separately.  Thanks @Fokko

PR #3379, opened 2025-12-29
[minor] Move on from java8 for three core modules

dossett:
Move on from Java 8 in `parquet-{common,column,hadoop}`. This PR was created with codex and this is how it describes the changes:

  What changed (Java 11 cleanups enabled by dropping Java 8):

  - Replaced Arrays.asList(...) and other collection boilerplate with List.of(...) where nulls aren’t used, reducing allocations and improving clarity.
  - Used Set.of(...) for small constant sets to simplify initialization.
  - Tightened helper/collection usages in tests to lean on Java 9+ convenience factories.
  - Kept Arrays.asList(...) where nulls are required to preserve behavior.


parquet-column/src/main/java/org/apache/parquet/schema/LogicalTypeAnnotation.java
```diff
@@ -822,8 +819,7 @@ PrimitiveStringifier valueStringifier(PrimitiveType primitiveType) {
   }
 
   public static class IntLogicalTypeAnnotation extends LogicalTypeAnnotation {
-    private static final Set<Integer> VALID_BIT_WIDTH =
-        Collections.unmodifiableSet(new HashSet<>(asList(8, 16, 32, 64)));
+    private static final Set<Integer> VALID_BIT_WIDTH = Set.of(8, 16, 32, 64);
```
Fokko, line 822: `Set.of` is also immutable 👍

PR #3381, opened 2025-12-29
Minor: `ci-before_install.sh` should show wget errors and retry

dossett:
The `wget` fetch of the thrift distribution is failing sometimes (due to rate limiting?) but the exact failure is completely suppressed with the `-q` flag.  Adding `-nv` will show the errors but keep the successes quiet.  Setting pipefail will allow the script to fail fast at this point since the failure is in a pipelined command.

If we see that rate limiting is in fact the problem, we can add retries here.

### DIFF
--- a/dev/ci-before_install.sh
+++ b/dev/ci-before_install.sh
@@ -31,7 +31,16 @@ sudo apt-get install -qq --no-install-recommends build-essential pv autoconf aut
    libevent-dev automake libtool flex bison pkg-config g++ libssl-dev xmlstarlet
 date
 pwd
-wget -nv -O- https://archive.apache.org/dist/thrift/$THRIFT_VERSION/thrift-$THRIFT_VERSION.tar.gz | tar zxf -
+for attempt in 1 2 3; do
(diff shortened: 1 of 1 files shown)


dev/ci-before_install.sh
```diff
@@ -23,14 +23,15 @@
 export THRIFT_VERSION=0.22.0
 
 set -e
+set -o pipefail
 date
 sudo apt-get update -qq
 sudo apt-get install -qq --no-install-recommends build-essential pv autoconf automake libtool curl make \
-   g++ unzip libboost-dev libboost-test-dev libboost-program-options-dev \
+   g++ unzip libboost-dev libboost-test-dev libboost-program-options-dev wget \
```
dossett, line 30: Trying to ensure an up to date version of wget

dossett:
I actually saw some failures on the workflows kicked off from this PR:

```
failed: Connection timed out.
failed: Network is unreachable.
```

I pushed an additional change to try the wget command up to three times.  Hopefully this makes the actions less flaky.

PR #3384, opened 2026-01-01
GH-3383: Modernize mockito

### LINKED ISSUE #3383: mockito should be upgrade to 5.x

Since Java 8 is no longer supported, we can make a big leap forward in mockito.  This will also allow to drop the dependency on powermock and use new functionality in mockito since powermock is not actively developed.

Hopefully a modern mocking platform will encourage more/better use of mocks when writing tests

### Component(s)

_No response_

dossett:
Closes #3383 

- Substantially upgrades mockito
- drop use of powermock in favor of new mockito functionality
- replace a single use of `easymock` with mockito functionality

### DIFF
--- a/parquet-thrift/src/test/java/org/apache/parquet/thrift/projection/TestStrictFieldProjectionFilter.java
+++ b/parquet-thrift/src/test/java/org/apache/parquet/thrift/projection/TestStrictFieldProjectionFilter.java
@@ -113,18 +113,20 @@ public class TestStrictFieldProjectionFilter {
 
   @Test
   public void testWarnWhenMultiplePatternsMatch() {
-    StrictFieldProjectionFilter filter = spy(new StrictFieldProjectionFilter(
-        Arrays.asList("a.b.c.{x_average,z_average}", "a.*_average")));
+    StrictFieldProjectionFilter filter =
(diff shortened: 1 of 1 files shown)


parquet-column/pom.xml
```diff
@@ -100,6 +100,12 @@
       <version>${commons-lang3.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-core</artifactId>
```
dossett, line 105: For reasons I don't understand this started getting flagged as an undeclared dependency.  This seems correct, but I don't know why it wasn't getting flagged before.

parquet-thrift/pom.xml
```diff
@@ -77,6 +77,12 @@
       <version>${guava.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <version>5.6.0</version>
+      <scope>test</scope>
+    </dependency>
```
dossett: I moved the easymock dependency here from the top-level pom since this is the only nodule using it.  Not sure if there's a project standard for that or not.

PR #3390, opened 2026-01-22
Add ALP (Adaptive Lossless floating-Point) encoding support

julienledem:
- related to https://github.com/apache/parquet-format/pull/548

**Following Micah's suggestion yesterday, I took a stab at using Claude to produce a java implementation of ALP based on Prateek's spec and c++ implementation.
Bear in mind that I haven't closely reviewed it yet, it is fairly experimental but it seems promising.**

Implements ALP encoding for FLOAT and DOUBLE types, which converts floating-point values to integers using decimal scaling, then applies Frame of Reference (FOR) encoding and bit-packing for compression.

New files:
- AlpConstants.java: Constants for ALP encoding
- AlpEncoderDecoder.java: Core encoding/decoding logic
- AlpValuesWriter.java: Writer implementation
- AlpValuesReaderForFloat/Double.java: Reader implementations

Includes comprehensive unit tests and interop test infrastructure.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -750,8 +750,7 @@ public class ParquetMetadataConverter {
   public Encoding getEncoding(org.apache.parquet.column.Encoding encoding) {
     // ALP encoding is not yet part of the parquet-format specification
     if (encoding == org.apache.parquet.column.Encoding.ALP) {
-      throw new IllegalArgumentException(
-          "ALP encoding is not yet supported in the parquet-format specification");
+      throw new IllegalArgumentException("ALP encoding is not yet supported in the parquet-format specification");
     }
     return Encoding.valueOf(encoding.name());
   }


parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpConstants.java
```diff
@@ -0,0 +1,136 @@
+/*
```
emkornfield, line 1: there was a recent comment on AI generated code on Arrow, if we aren't doing a lot of edits I'm not sure how it should be licensed.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpConstants.java
```diff
@@ -0,0 +1,136 @@
...
+  /** FOR encoding for integers (0 = FOR) */
+  public static final int ALP_INTEGER_ENCODING_FOR = 0;
+
+  /** Size of the ALP page header in bytes */
+  public static final int ALP_HEADER_SIZE = 8;
+
+  // ========== Vector Constants ==========
+
+  /** Default number of elements per compressed vector (2^10 = 1024) */
+  public static final int ALP_VECTOR_SIZE = 1024;
```
emkornfield, line 55: nit: this is default but it should be configurable.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpConstants.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /** Log2 of the default vector size */
+  public static final int ALP_VECTOR_SIZE_LOG = 10;
```
emkornfield, line 58: same comment on configurable.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpConstants.java
```diff
@@ -0,0 +1,136 @@
...
+  /** Number of (exponent, factor) combinations for float: sum(1..11) = 66 */
+  public static final int FLOAT_COMBINATIONS = 66;
+
+  /** Number of (exponent, factor) combinations for double: sum(1..19) = 190 */
+  public static final int DOUBLE_COMBINATIONS = 190;
+
+  // ========== Sampling Constants ==========
+
+  /** Number of values sampled per vector */
+  public static final int SAMPLER_SAMPLES_PER_VECTOR = 256;
```
emkornfield, line 77: wonder if these should be configurable somehow?  Probably OK if not.

can these be package private?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+   * @return true if the value is an exception for this encoding
+   */
+  public static boolean isFloatException(float value, int exponent, int factor) {
+    if (isFloatException(value)) {
+      return true;
+    }
+
+    // Try encoding and check for round-trip failure
+    float multiplier = FLOAT_POW10[exponent];
+    if (factor > 0) {
```
emkornfield, line 87: Not sure I understand the > 0 check, need to examine encoding logic in more detail, is this just an optimization to avoid dividing by 1?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+   * @return true if the value is an exception for this encoding
+   */
+  public static boolean isDoubleException(double value, int exponent, int factor) {
+    if (isDoubleException(value)) {
+      return true;
+    }
+
+    // Try encoding and check for round-trip failure
+    double multiplier = DOUBLE_POW10[exponent];
+    if (factor > 0) {
```
emkornfield, line 195: I think we can just call encoding here?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+      multiplier /= DOUBLE_POW10[factor];
+    }
+
+    double scaled = value * multiplier;
```
emkornfield, line 199: maybe just call decode.  I imaging JIT inlining should be able to do the optimizations for lookup.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+   * <p>Formula: encoded = round(value * 10^(exponent - factor))
+   *
+   * @param value    the double value to encode
+   * @param exponent the decimal exponent (0-18)
+   * @param factor   the decimal factor (0 &lt;= factor &lt;= exponent)
+   * @return the encoded long value
+   */
+  public static long encodeDouble(double value, int exponent, int factor) {
+    double multiplier = DOUBLE_POW10[exponent];
+    if (factor > 0) {
```
emkornfield, line 226: micro-optimization: would it be faster to just have 1.0 in the array at index 0 (also just to check I believe this is numerically stable but not 100% sure).

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+  }
+
+  // ========== Bit Width Calculation ==========
+
+  /**
+   * Calculate the bit width needed to store unsigned values up to maxDelta.
+   *
+   * @param maxDelta the maximum delta value (unsigned)
+   * @return the number of bits needed (0-32 for int, 0-64 for long)
```
emkornfield, line 270: long is not covered here.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpConstants.java
emkornfield: I think this is less important to make configurable then vector size as we are specifically calling it out as configurable in the spec we should make sure we generate some data at different bit-widths.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+   */
+  public static int bitWidth(int maxDelta) {
+    if (maxDelta == 0) {
+      return 0;
+    }
+    return 32 - Integer.numberOfLeadingZeros(maxDelta);
```
emkornfield, line 276: nit: Integer.SIZE?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+   * Calculate the bit width needed to store unsigned values up to maxDelta.
+   *
+   * @param maxDelta the maximum delta value (unsigned)
+   * @return the number of bits needed (0-64)
+   */
+  public static int bitWidth(long maxDelta) {
+    if (maxDelta == 0) {
+      return 0;
+    }
+    return 64 - Long.numberOfLeadingZeros(maxDelta);
```
emkornfield, line 289: nit: Long.SIZE

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+
+  // ========== Bit Width Calculation ==========
+
+  /**
+   * Calculate the bit width needed to store unsigned values up to maxDelta.
+   *
+   * @param maxDelta the maximum delta value (unsigned)
+   * @return the number of bits needed (0-32 for int, 0-64 for long)
+   */
+  public static int bitWidth(int maxDelta) {
```
emkornfield, line 272: I worry about accidental type coercion that might call this method instead of the Long method.  Maybe have an explicit name here

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+
+  // ========== Float Encoding/Decoding ==========
+
+  /**
+   * Check if a float value is an exception (cannot be losslessly encoded).
+   *
+   * @param value the float value to check
+   * @return true if the value is an exception
+   */
+  public static boolean isFloatException(float value) {
```
emkornfield, line 56: please audit the visibily for these methods, it seems most could be at least package private?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpEncoderDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+          if (isFloatException(values[offset + i], e, f)) {
+            exceptions++;
+          }
+        }
+        if (exceptions < bestExceptions) {
+          bestExponent = e;
+          bestFactor = f;
+          bestExceptions = exceptions;
+          // Early exit if we found perfect encoding
+          if (bestExceptions == 0) {
```
emkornfield, line 338: need to read the paper, but does the order we are exploring the exceptions and factors make sense (is there any benefit to exploring from largest to smallest).

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+    if (version != ALP_VERSION) {
+      throw new ParquetDecodingException("Unsupported ALP version: " + version + ", expected " + ALP_VERSION);
+    }
+    if (compressionMode != ALP_COMPRESSION_MODE) {
+      throw new ParquetDecodingException("Unsupported ALP compression mode: " + compressionMode);
+    }
+    if (integerEncoding != ALP_INTEGER_ENCODING_FOR) {
+      throw new ParquetDecodingException("Unsupported ALP integer encoding: " + integerEncoding);
+    }
+
```
emkornfield, line 80: lets check ` 0  <=  logVectorScale <= 16`

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+      throws ParquetDecodingException, IOException {
+    LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
+
+    // Read and validate header
+    ByteBuffer headerBuf = stream.slice(ALP_HEADER_SIZE).order(ByteOrder.LITTLE_ENDIAN);
+    int version = headerBuf.get() & 0xFF;
+    int compressionMode = headerBuf.get() & 0xFF;
+    int integerEncoding = headerBuf.get() & 0xFF;
+    int logVectorSize = headerBuf.get() & 0xFF;
+    int numElements = headerBuf.getInt();
```
emkornfield, line 69: check numElements <= valuesCount ?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+    this.totalCount = 0;
+  }
+
+  @Override
+  public void initFromPage(int valuesCount, ByteBufferInputStream stream)
+      throws ParquetDecodingException, IOException {
+    LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
+
+    // Read and validate header
+    ByteBuffer headerBuf = stream.slice(ALP_HEADER_SIZE).order(ByteOrder.LITTLE_ENDIAN);
```
emkornfield, line 64: it looks like this call does a copy?  i wonder if we are better off using DataInputStream and reversing the bytes for numElements?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+
+    int vectorSize = 1 << logVectorSize;
+    int numVectors = (numElements + vectorSize - 1) / vectorSize;
+
+    this.totalCount = numElements;
+    this.decodedValues = new double[numElements];
+    this.currentIndex = 0;
+
+    // Read AlpInfo array
+    ByteBuffer alpInfoBuf = stream.slice(ALP_INFO_SIZE * numVectors).order(ByteOrder.LITTLE_ENDIAN);
```
emkornfield, line 89: same comment as above on avoiding the slice.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+    int[] numExceptions = new int[numVectors];
+
+    for (int v = 0; v < numVectors; v++) {
+      exponents[v] = alpInfoBuf.get() & 0xFF;
+      factors[v] = alpInfoBuf.get() & 0xFF;
+      numExceptions[v] = alpInfoBuf.getShort() & 0xFFFF;
+    }
+
+    // Read ForInfo array
+    ByteBuffer forInfoBuf = stream.slice(DOUBLE_FOR_INFO_SIZE * numVectors).order(ByteOrder.LITTLE_ENDIAN);
```
emkornfield, line 101: we might as well slice this together with AlpInfo we keep the slices.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+
+      // Read and unpack values
+      long[] deltas = new long[vectorLen];
+      if (bitWidths[v] > 0) {
+        ByteBuffer packedBuf = stream.slice(packedBytes).order(ByteOrder.LITTLE_ENDIAN);
+        unpackLongs(packedBuf, deltas, vectorLen, bitWidths[v]);
+      }
+      // else: all deltas are 0 (bitWidth == 0)
+
+      // Reverse FOR and decode
```
emkornfield, line 127: nit: spell out FOR?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+    }
+  }
+
+  /**
+   * Unpack longs from a byte buffer using the specified bit width.
+   * Note: This method reads from the buffer's current position, using absolute indexing
+   * relative to that position.
+   */
+  private void unpackLongs(ByteBuffer packed, long[] values, int count, int bitWidth) {
```
emkornfield, line 153: shouldn't there already be a method someplace that does this?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+    ByteBuffer forInfoBuf = stream.slice(DOUBLE_FOR_INFO_SIZE * numVectors).order(ByteOrder.LITTLE_ENDIAN);
+    long[] frameOfReference = new long[numVectors];
+    int[] bitWidths = new int[numVectors];
+
+    for (int v = 0; v < numVectors; v++) {
+      frameOfReference[v] = forInfoBuf.getLong();
+      bitWidths[v] = forInfoBuf.get() & 0xFF;
+    }
+
+    // Decode each vector
```
emkornfield, line 110: Do we really want do this all up-front?  Don't we only want to decode one vector at a time?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForFloat.java
```diff
@@ -0,0 +1,201 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
emkornfield, line 2: not reviewing this directly but all comments on double apply here I think.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesWriter.java
```diff
@@ -0,0 +1,525 @@
...
+   * Float-specific ALP values writer.
+   */
+  public static class FloatAlpValuesWriter extends AlpValuesWriter {
+    private float[] buffer;
+    private int count;
+
+    public FloatAlpValuesWriter(int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+      super(initialCapacity, pageSize, allocator);
+      // Initial buffer size - will grow as needed
+      this.buffer = new float[Math.max(ALP_VECTOR_SIZE, initialCapacity / Float.BYTES)];
```
emkornfield, line 80: I don't think I understand this logic.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesReaderForDouble.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  @Override
+  public void skip() {
+    skip(1);
+  }
+
+  @Override
+  public void skip(int n) {
+    if (n < 0 || currentIndex + n > totalCount) {
```
emkornfield, line 195: When we move to incremental decoding this should skip by forwarding avoiding decoding vector data for intermediate results.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesWriter.java
emkornfield: This should just be pageSize?

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesWriter.java
```diff
@@ -0,0 +1,525 @@
...
+        float[] newBuffer = new float[buffer.length * 2];
+        System.arraycopy(buffer, 0, newBuffer, 0, count);
+        buffer = newBuffer;
+      }
+      buffer[count++] = v;
+    }
+
+    @Override
+    public long getBufferedSize() {
+      // Estimate: each float value contributes roughly 2-4 bytes after compression
```
emkornfield, line 97: I think we should be flushing incrementally to byte buffers for the encoded data to get a more accurate value.  For unflushed data the size heuristic is maybe OK for an initial version.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesWriter.java
```diff
@@ -0,0 +1,525 @@
...
+    public FloatAlpValuesWriter(int initialCapacity, int pageSize, ByteBufferAllocator allocator) {
+      super(initialCapacity, pageSize, allocator);
+      // Initial buffer size - will grow as needed
+      this.buffer = new float[Math.max(ALP_VECTOR_SIZE, initialCapacity / Float.BYTES)];
+      this.count = 0;
+    }
+
+    @Override
+    public void writeFloat(float v) {
+      if (count >= buffer.length) {
```
emkornfield, line 86: I don't think we should be growing here to all floats.  Once we reach vector-size we should be encoding the page.

parquet-column/src/main/java/org/apache/parquet/column/values/alp/AlpValuesWriter.java
```diff
@@ -0,0 +1,525 @@
...
+      float[][] exceptionValues = new float[numVectors][];
+
+      // Process each vector
+      for (int v = 0; v < numVectors; v++) {
+        int vectorStart = v * ALP_VECTOR_SIZE;
+        int vectorEnd = Math.min(vectorStart + ALP_VECTOR_SIZE, count);
+        int vectorLen = vectorEnd - vectorStart;
+
+        // Find best encoding parameters
+        AlpEncoderDecoder.EncodingParams params =
```
emkornfield, line 132: Don't we want to cache encodings and use those first (and stop searching after a while)?

PR #3396, opened 2026-02-16
Implement per column compression

rahil-c:
### Rationale for this change

Issue Raised here: https://github.com/apache/parquet-format/issues/553

The Parquet spec already supports per-column compression, each column chunk stores its own CompressionCodecName in the footer metadata. However, the parquet-java writer API currently forces a single compression codec for all columns in a file. This PR address that gap by exposing per-column compression configuration through the existing ColumnProperty<T> infrastructure.

- ParquetProperties: Added ColumnProperty<CompressionCodecName> following the same pattern used for dictionary encoding, bloom filters.
 - ColumnChunkPageWriteStore: Added a new constructor that accepts CompressionCodecFactory + ParquetProperties 
 - InternalParquetRecordWriter: Added a new constructor accepting CompressionCodecFactory instead of a single BytesInputCompressor. 
 - ParquetWriter: Added withCompressionCodec(String,CompressionCodecName) builder method. Updated the core constructor to pass the CompressionCodecFactory through to the writer stack.
 - ParquetOutputFormat: Added ColumnConfigParser entry so per-column compression can be configured via Hadoop config keys (parquet.compression#<column>=CODEC).
 - ParquetRecordWriter: Updated to pass CompressionCodecFactory to InternalParquetRecordWriter.

### Are these changes tested?
* Added test within this pr

Two new public APIs are introduced:
```
  ParquetWriter.builder(path)
      .withCompressionCodec(CompressionCodecName.SNAPPY)
         // default for all columns
      .withCompressionCodec("embeddings",
  CompressionCodecName.UNCOMPRESSED)  // per-column override
      .build();

  Hadoop configuration (new key pattern):
  parquet.compression#<column_path>=<CODEC_NAME>
```

cc @julienledem @emkornfield

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -39,6 +39,7 @@ import org.apache.parquet.column.values.factory.DefaultValuesWriterFactory;
 import org.apache.parquet.column.values.factory.ValuesWriterFactory;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWriter;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.schema.MessageType;
 
 /**
@@ -135,6 +136,7 @@ public class ParquetProperties {
   private final Map<String, String> extraMetaData;
   private final ColumnProperty<Boolean> statistics;
   private final ColumnProperty<Boolean> sizeStatistics;
+  private final ColumnProperty<CompressionCodecName> compressionCodec;
 
   private ParquetProperties(Builder builder) {
     this.pageSizeThreshold = builder.pageSize;
@@ -167,6 +169,7 @@ public class ParquetProperties {
     this.extraMetaData = builder.extraMetaData;
     this.statistics = builder.statistics.build();
     this.sizeStatistics = builder.sizeStatistics.build();
+    this.compressionCodec = builder.compressionCodec.build();
   }
 
   public static Builder builder() {
@@ -348,6 +351,14 @@ public class ParquetProperties {
     return numBloomFilterCandidates.getValue(column);
   }
 
+  public CompressionCodecName getCompressionCodec(ColumnDescriptor column) {
+    return compressionCodec.getValue(column);
+  }
+
+  public CompressionCodecName getDefaultCompressionCodec() {
+    return compressionCodec.getDefaultValue();
+  }
+
   public Map<String, String> getExtraMetaData() {
     return extraMetaData;
   }
@@ -419,6 +430,7 @@ public class ParquetProperties {
     private Map<String, String> extraMetaData = new HashMap<>();
     private final ColumnProperty.Builder<Boolean> statistics;
     private final ColumnProperty.Builder<Boolean> sizeStatistics;
+    private final ColumnProperty.Builder<CompressionCodecName> compressionCodec;
 
     private Builder() {
       enableDict = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_IS_DICTIONARY_ENABLED);
@@ -436,6 +448,8 @@ public class ParquetProperties {
           ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER);
       statistics = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_STATISTICS_ENABLED);
       sizeStatistics = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_SIZE_STATISTICS_ENABLED);
+      compressionCodec =
+          ColumnProperty.<CompressionCodecName>builder().withDefaultValue(CompressionCodecName.UNCOMPRESSED);
     }
 
     private Builder(ParquetProperties toCopy) {
@@ -460,6 +474,7 @@ public class ParquetProperties {
       this.extraMetaData = toCopy.extraMetaData;
       this.statistics = ColumnProperty.builder(toCopy.statistics);
       this.sizeStatistics = ColumnProperty.builder(toCopy.sizeStatistics);
+      this.compressionCodec = ColumnProperty.builder(toCopy.compressionCodec);
     }
 
     /**
@@ -753,6 +768,32 @@ public class ParquetProperties {
      */
     public Builder withSizeStatisticsEnabled(String columnPath, boolean enabled) {
       this.sizeStatistics.withValue(columnPath, enabled);
+      return this;
+    }
+
+    /**
+     * Set the compression codec for the columns not specified by
+     * {@link #withCompressionCodec(String, CompressionCodecName)}.
+     *
+     * @param codecName the compression codec to use by default
+     * @return this builder for method chaining.
+     */
+    public Builder withCompressionCodec(CompressionCodecName codecName) {
+      this.compressionCodec.withDefaultValue(
+          Objects.requireNonNull(codecName, "compressionCodecName cannot be null"));
+      return this;
+    }
+
+    /**
+     * Set the compression codec for the specified column.
+     *
+     * @param columnPath the path of the column (dot-string)
+     * @param codecName  the compression codec to use for the column
+     * @return this builder for method chaining.
+     */
+    public Builder withCompressionCodec(String columnPath, CompressionCodecName codecName) {
+      this.compressionCodec.withValue(
+          columnPath, Objects.requireNonNull(codecName, "compressionCodecName cannot be null"));
       return this;
     }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.zip.CRC32;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.ByteBufferReleaser;
@@ -577,22 +578,7 @@ public class ColumnChunkPageWriteStore implements PageWriteStore, BloomFilterWri
       ByteBufferAllocator allocator,
       int columnIndexTruncateLength,
       boolean pageWriteChecksumEnabled) {
-    this.schema = schema;
-    for (ColumnDescriptor path : schema.getColumns()) {
-      writers.put(
-          path,
-          new ColumnChunkPageWriter(
-              path,
-              compressor,
-              allocator,
-              columnIndexTruncateLength,
-              pageWriteChecksumEnabled,
-              null,
-              null,
-              null,
-              -1,
-              -1));
-    }
+    this(compressor, schema, allocator, columnIndexTruncateLength, pageWriteChecksumEnabled, null, 0);
   }
 
   @Deprecated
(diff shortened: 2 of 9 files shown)


parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -756,6 +771,32 @@ public Builder withSizeStatisticsEnabled(String columnPath, boolean enabled) {
...
     }
 
+    /**
+     * Set the compression codec for the columns not specified by
+     * {@link #withCompressionCodec(String, CompressionCodecName)}.
+     *
+     * @param codecName the compression codec to use by default
+     * @return this builder for method chaining.
+     */
+    public Builder withCompressionCodec(CompressionCodecName codecName) {
```
emkornfield, line 781: There isn't an existing API to set this?  I have to look more closely at the the convention but would `withDefaultCompressionCodec` make sense?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -672,6 +673,83 @@ public ColumnChunkPageWriteStore(
...
+      boolean pageWriteChecksumEnabled,
+      InternalFileEncryptor fileEncryptor,
+      int rowGroupOrdinal) {
+    this.schema = schema;
+    if (null == fileEncryptor) {
+      for (ColumnDescriptor path : schema.getColumns()) {
+        BytesInputCompressor compressor = codecFactory.getCompressor(props.getCompressionCodec(path));
+        writers.put(
+            path,
+            new ColumnChunkPageWriter(
```
emkornfield, line 634: Is this copy and paste from other constructors, I wonder if there is some refactoring that can be done to avoid duplication? (I wonder if we should have a ColumnChunkPageWriterBuilder?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -108,14 +147,27 @@ public ParquetMetadata getFooter() {
...
-    ColumnChunkPageWriteStore columnChunkPageWriteStore = new ColumnChunkPageWriteStore(
-        compressor,
-        schema,
-        props.getAllocator(),
-        props.getColumnIndexTruncateLength(),
-        props.getPageWriteChecksumEnabled(),
-        fileEncryptor,
-        rowGroupOrdinal);
+    ColumnChunkPageWriteStore columnChunkPageWriteStore;
+    if (codecFactory != null) {
```
emkornfield, line 157: Is it possible to create a default codecFactory to avoid the if/else block below?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -858,4 +860,181 @@ public void testNoFlushAfterException() throws Exception {
     FileSystem fs = file.getFileSystem(conf);
     assertTrue(!fs.exists(file) || fs.getFileStatus(file).getLen() == 0);
   }
+
+  @Test
+  public void testPerColumnCompression() throws Exception {
```
emkornfield: does this test anything that the next method does not?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
emkornfield, line 195: It seems like this should be deprecated, and new method without codec passed in should be exposed instead?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
```diff
@@ -201,15 +201,11 @@ public ParquetRecordWriter(
       MemoryManager memoryManager,
       Configuration conf) {
     this.codecFactory = new CodecFactory(conf, props.getPageSizeThreshold());
+    // Ensure the default compression codec from ParquetOutputFormat is set in props
+    ParquetProperties propsWithCodec =
+        ParquetProperties.copy(props).withCompressionCodec(codec).build();
```
emkornfield: does risk overwriting an already set compression codec?

emkornfield (commented):
Took a first pass not very familiar with this code, but I'd also expect potentially more test coverage given the number of classes changed?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
rahil-c: There is an existing api at the Parquet writer builder class: https://github.com/apache/parquet-java/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L567
However there is not an api within this properties class, which we use this class in a couple of places to pass the column's compression.

In terms of the naming I am ok with that suggestion, however I noticed the pattern for the APIs to not prefix with *Default*, even when setting a default value: https://github.com/apache/parquet-java/blob/master/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java#L484

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
rahil-c: It looks like we currently do not have a `ColumnChunkPageWriterBuilder`, if that is the ideal pattern i can look into adding this builder.
Screenshot 2026-02-22 at 8 55 28 AM

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
rahil-c: In my mind I thought the if/else would be simpler, since there are callers that will not be providing/using the codec factory so not sure if having the default codec there would make sense, such as the following:
https://github.com/apache/parquet-java/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java#L106

Let me know though if you think we should still pursue the default code factory approach.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
rahil-c: You are correct the second test offers the same coverage, with additonal compression codecs being tested. I can remove this test.

rahil-c:
> Took a first pass not very familiar with this code, but I'd also expect potentially more test coverage given the number of classes changed?

Will add more coverage

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
rahil-c: I think you are correct, since the callers now will pass a `ParquetProperties` props is using the compression codec builder method `withCompressionCodec`, so I think we can expose a new constructor without `CompressionCodecName` codec itself.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
rahil-c: Actually opted for some other way of reducing duplication let me know if its ok to you, or if the builder is preferable.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
rahil-c: addressed

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
emkornfield: lets keep convention.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
emkornfield: I'm OK wit this.

emkornfield (approved):
At a high level this looks reasonable.  @Fokko, @gszadovszky or @wgtmac would you have a moment to review?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
wgtmac: +1 to add a builder pattern if we don't need to touch a lot of files.

wgtmac (commented):
Thanks for pinging me! I think this overall looks good.

I'm just a little bit concerned that we need to add a new ctor and deprecate the old one every time because a new input parameter is needed.

mengna-lin:
Column level compression level might also be needed. 
For example,  level 22 is a valid value for ZSTD but not GZIP.

emkornfield:
> Column level compression level might also be needed.
For example, level 22 is a valid value for ZSTD but not GZIP.

I think this can probably be added in a follow-up?

rahil-c:
@emkornfield @wgtmac do you think we could get this change in the next parquet java release if there is no other concerns, or did you anything else from my side.

emkornfield:
Sorry for the delay will take another pass through.

emkornfield:
Is this ready for review?  Generally, I think it is good practice to re-request review to make sure reviewers know the ball is in there court (apologies if you did this and I missed it).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
rahil-c: @wgtmac let me know if this https://github.com/apache/parquet-java/pull/3396/changes/08cf03bb1dfd52d3f85252f53ae6d556fae30e36 looks good to you?

rahil-c:
@emkornfield thanks appreciate the help.

In terms of review, I believe the last comment was from @wgtmac on requesting for the builder pattern https://github.com/apache/parquet-java/pull/3396#discussion_r2925354951, so will wait on his approval.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -672,6 +677,88 @@ public ColumnChunkPageWriteStore(
...
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * Builder for {@link ColumnChunkPageWriteStore}. Prefer this over the constructors when new
+   * parameters are needed so that callers do not have to be updated every time a parameter is
+   * added.
+   */
+  public static class Builder {
```
wgtmac, line 689: Does this new builder look good to you? @gszadovszky @Fokko?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -672,6 +677,88 @@ public ColumnChunkPageWriteStore(
...
+      return this;
+    }
+
+    public Builder withRowGroupOrdinal(int rowGroupOrdinal) {
+      this.rowGroupOrdinal = rowGroupOrdinal;
+      return this;
+    }
+
+    public ColumnChunkPageWriteStore build() {
+      if (compressorProvider == null) {
```
wgtmac, line 748: It seems that `schema` and `allocator` are also required arguments so we need to check them here. BTW, we may switch to use `Preconditions.checkArgument` to be consistent with the repo.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -672,6 +677,88 @@ public ColumnChunkPageWriteStore(
...
+    public Builder withCompressor(BytesInputCompressor compressor) {
+      this.compressorProvider = path -> compressor;
+      return this;
+    }
+
+    /**
+     * Resolve the compressor per column from the given codec factory and properties, allowing
+     * per-column compression codecs.
+     */
+    public Builder withCodecFactory(CompressionCodecFactory codecFactory, ParquetProperties props) {
```
wgtmac, line 712: It looks a little bit weird that `ParquetProperties` is used here. Should we directly use `Function<ColumnDescriptor, BytesInputCompressor>` to be more flexible?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -77,7 +79,9 @@ class InternalParquetRecordWriter<T> {
    * @param extraMetaData     extra meta data to write in the footer of the file
    * @param rowGroupSize      the size of a block in the file (this will be approximate)
    * @param compressor        the codec used to compress
+   * @deprecated Use {@link #InternalParquetRecordWriter(ParquetFileWriter, WriteSupport, MessageType, Map, long, CompressionCodecFactory, boolean, ParquetProperties)} for per-column compression support
    */
+  @Deprecated
```
wgtmac, line 84: Is it better not to deprecate this constructor? `InternalParquetRecordWriter` is not public so we can change the parameter `compressor` to be a function that similar to one used by `ColumnChunkPageWriteStore`.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
```diff
@@ -188,7 +188,10 @@ public ParquetRecordWriter(
    * @param codec         the compression codec used to compress the pages
    * @param validating    if schema validation should be turned on
    * @param props         parquet encoding properties
+   * @deprecated Use {@link #ParquetRecordWriter(ParquetFileWriter, WriteSupport, MessageType, Map, long, boolean, ParquetProperties, MemoryManager, Configuration)}
+   *             and set the compression codec in {@link ParquetProperties} instead
    */
+  @Deprecated
```
wgtmac, line 194: Let's not add `@Deprecated` here because this constructor is not a public one. We can directly modify its signature. We don't need to add a public ctor since this class is only used by `ParquetOutputFormat`.

wgtmac (changes requested):
Sorry for the late reply. I have some concerns about the current API. Let me know what you think. Thanks!

wgtmac:
It seems that https://github.com/apache/parquet-java/pull/3526 also added support for setting compression levels. Due to limited reviewer capacity, should we include level support in this PR as well? (If there are multiple PRs implementing the same functionality, I generally prefer to proceed with the one that was opened first. cc @mengna-lin)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -395,7 +395,6 @@ public ParquetWriter(Path file, Configuration conf, WriteSupport<T> writeSupport
     fileWriter.start();
 
     this.codecFactory = codecFactory;
-    CompressionCodecFactory.BytesInputCompressor compressor = codecFactory.getCompressor(compressionCodecName);
```
wgtmac, line 398: `compressionCodecName` is unused now. Let's remove it?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordWriter.java
```diff
@@ -200,16 +203,42 @@ public ParquetRecordWriter(
...
+    this(
         w,
         writeSupport,
         schema,
         extraMetaData,
         blockSize,
-        codecFactory.getCompressor(codec),
         validating,
-        props);
+        ParquetProperties.copy(props).withCompressionCodec(codec).build(),
```
wgtmac, line 213: The copy constructor of ParquetProperties does not preserve several non-default fields. As a result, old callers can unintentionally reset settings like `pageValueCountThreshold`, `columnIndexTruncateLength`, `statisticsTruncateLength`, and `rowGroupRowCountLimit` when they hit this overload.

PR #3402, opened 2026-02-24
GH-3401: Remove redundant MAP_KEY_VALUE annotation at middle level of logical type map

### LINKED ISSUE #3401: Added redundant `MAP_KEY_VALUE` annotation to parquet's thrift structure

Show in https://github.com/apache/arrow/issues/49114#issuecomment-3839473577.
`list<map<string, string>>` schema will generate this structure, and the `MAP_KEY_VALUE` annotation at middle level is not required in parquet's spec https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps.
```thrift
  required group list_of_maps (LIST) {
    repeated group array (MAP) {
      repeated group key_value (MAP_KEY_VALUE) { // <-- here
        required binary key (STRING);
        required binary value (STRING);
      }
    }
  }
```

### Component(s)

Thrift

HuaHuaY:
### Rationale for this change

Remove redundant `MAP_KEY_VALUE` annotation which is not required in parquet's spec.

`ConversionPatterns` will not pass `LogicalTypeAnnotation.MapKeyValueTypeAnnotation` to `GroupType` function.

### Are these changes tested?

Yes.

Yes. The `MAP_KEY_VALUE` annotation at middle level of thrift structure `map<..., ...>` will be removed.

Closes #3401


Fokko (approved):
I think removing this reduces visual noise, and I like it.


Regarding backward compatibility, could folks use it in their tests as we do here, as shown in this PR?

PR #3407, opened 2026-02-27
GH-3406: fix Float16 statistics handling for NaN and zero values

### LINKED ISSUE #3406: `Float16` treats `Nan`, +0 and -0 differently than `Double`.

It should not create column index when the min or max value is NaN. And it should adjust the min or max value when meets +0 or -0.

### Component(s)

_No response_

HuaHuaY:
### Rationale for this change

1. When creating column index and NaN is the min or max value, column index will not be generated.
2. When creating statistic and the min value is +0, the min value will be changed to -0. When the max value is -0, it will be changed to +0.

- In `Statistics<>.Float16Builder` and `BinaryColumnIndexBuilder`, this PR will adjust value when the min or max value is +0 or -0.
- In `BinaryColumnIndexBuilder`, this PR will mark the column index is invalid when the min or max value is NaN.
- In `Statistics<>.Float16Builder`, statistic will be changed to (+0.0, +0.0) which is the same behavior as `Double`.

### Are these changes tested?

Yes.

Yes. This PR corrects the behavior about `Float16`.

Closes #3406

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestFloat16ReadWriteRoundTrip.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestFloat16ReadWriteRoundTrip.java
@@ -21,6 +21,7 @@ package org.apache.parquet.statistics;
 import static org.apache.parquet.schema.LogicalTypeAnnotation.float16Type;
 import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestFloat16ReadWriteRoundTrip.java
```diff
@@ -123,36 +123,29 @@ public class TestFloat16ReadWriteRoundTrip {
...
-    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x00}), // +0
+    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x80}), // -0
     Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x00})
   }; // +0
 
   private Binary[] valuesAllNegativeZeroMinMax = {
-    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x80}), // -0
-    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x80})
+    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x80}),
+    Binary.fromConstantByteArray(new byte[] {(byte) 0x00, (byte) 0x00}) // +0
```
wgtmac: This comment looks weird

wgtmac:
@gszadovszky Could you help review this as well? I think this fix makes a lot of sense to be consistent with what currently float and double types do.

PR #3415, opened 2026-03-07
 GH-3414: Add parseJson to VariantBuilder for JSON-to-Variant conversion

### LINKED ISSUE #3414: parseJson implementation in VariantBuilder

Every consumer of parquet-variant currently has to implement their own
  JSON-to-Variant parsing (Spark, Kafka Connect, Flink, etc.). This adds
  native parseJson() static methods to VariantBuilder using Jackson
  streaming parser, ported from Apache Spark's production-proven logic.
  - Single-pass streaming (no intermediate JsonNode tree)
  - Smallest integer type selection (byte/short/int/long)
  - Decimal-first float handling, double fallback for scientific notation
  - Overloads for String input and pre-positioned JsonParser

https://github.com/apache/spark/tree/master/common/variant

### Component(s)

Core

gaurav7261:
### Rationale for this change
  Every consumer of `parquet-variant` currently has to independently implement JSON-to-Variant parsing. Apache Spark has one in its `common/variant` module 
  ([source](https://github.com/apache/spark/tree/master/common/variant)), our Kafka Connect S3 sink connector had to write one, and any other project (Flink, Trino, DuckDB-Java, etc.) would need to do the 
  same. Since `VariantBuilder` already provides all the low-level `append*()` primitives, `parseJson()` is the natural completion of that API — a canonical, reusable entry point for the most common use 
  case: converting a JSON string into a Variant.
  - **`parquet-variant/pom.xml`**: Added `jackson-core` (compile) and `parquet-jackson` (runtime) dependencies, following the same pattern as `parquet-hadoop`.
  - **`VariantBuilder.java`**: Added two public static methods:
    - `parseJson(String json)` — convenience method that creates a Jackson streaming parser internally.
    - `parseJson(JsonParser parser)` — for callers who already have a positioned parser (e.g., reading from a stream).
    - Internal helpers ported from Apache Spark's production `VariantBuilder.buildJson`:
      - `buildJson()` — recursive single-pass streaming parser handling OBJECT, ARRAY, STRING, NUMBER_INT, NUMBER_FLOAT, TRUE, FALSE, NULL.
      - `appendSmallestLong()` — selects the smallest integer type (BYTE/SHORT/INT/LONG) based on value range.
      - `tryAppendDecimal()` — decimal-first encoding for floating-point numbers; falls back to double only for scientific notation or values exceeding DECIMAL16 precision (38 digits).
  - **`TestVariantParseJson.java`**: 32 new tests covering all primitive types, objects (empty, simple, nested, null values, sorted keys, duplicate keys), arrays (empty, simple, nested, mixed types), and 
  edge cases (unicode, escaped strings, deeply nested documents, scientific notation, integer overflow to decimal, malformed JSON).
  ### Are these changes tested?
  Yes. 32 new tests in `TestVariantParseJson`.
  Yes. Two new public static methods on `VariantBuilder`:
  - `VariantBuilder.parseJson(String json)` — returns a `Variant`
  - `VariantBuilder.parseJson(JsonParser parser)` — returns a `Variant`
  These are additive API additions with no breaking changes to existing APIs.
  Closes #3414

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,5 @@ target/
 mvn_install.log
 .vscode/*
 .DS_Store
+.memsearch/
 
--- a/parquet-variant/pom.xml
+++ b/parquet-variant/pom.xml
@@ -47,6 +47,17 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>${jackson.groupId}</groupId>
+      <artifactId>jackson-core</artifactId>
+      <version>${jackson.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-jackson</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
@@ -1,0 +1,206 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.parquet.variant;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.core.exc.InputCoercionException;
+import java.io.IOException;
+import java.math.BigDecimal;
+
+/**
+ * Parses JSON into {@link Variant} values using Jackson streaming.
+ *
+ * <p>This class isolates the Jackson dependency from {@link VariantBuilder},
+ * so that core variant construction does not require Jackson on the classpath.
+ *
+ * <p>Ported from Apache Spark's {@code VariantBuilder.parseJson}.
+ */
+public final class VariantJsonParser {
+
+  private static final JsonFactory JSON_FACTORY = JsonFactory.builder()
+      .streamReadConstraints(StreamReadConstraints.builder()
+          .maxNestingDepth(500)
+          .maxStringLength(10_000_000)
+          .maxDocumentLength(50_000_000L)
+          .build())
+      .build();
+
+  private VariantJsonParser() {}
+
+  /**
+   * Parses a JSON string and returns the corresponding {@link Variant}.
+   *
+   * <p>Uses Jackson streaming parser for single-pass conversion
+   * with no intermediate tree. Number handling preserves precision:
+   * integers use the smallest fitting type, floating-point numbers
+   * prefer decimal encoding (no scientific notation) and fall back
+   * to double.
+   *
+   * @param json the JSON string to parse
+   * @return the parsed Variant
+   * @throws IOException if the JSON is malformed or an I/O error occurs
+   */
+  public static Variant parseJson(String json) throws IOException {
+    try (JsonParser parser = JSON_FACTORY.createParser(json)) {
+      parser.nextToken();
+      return parseJson(parser);
+    }
+  }
+
+  /**
+   * Parses a JSON value from an already-positioned {@link JsonParser}
+   * and returns the corresponding {@link Variant}. The parser must
+   * have its current token set (i.e., {@code parser.nextToken()}
+   * or equivalent must have been called).
+   *
+   * @param parser a positioned Jackson JsonParser
+   * @return the parsed Variant
+   * @throws IOException if the JSON is malformed or an I/O error occurs
+   */
+  public static Variant parseJson(JsonParser parser) throws IOException {
+    VariantBuilder builder = new VariantBuilder();
+    buildJson(builder, parser);
+    return builder.build();
+  }
+
+  /**
+   * Recursively builds a Variant value from the current position of a
+   * Jackson streaming parser. Handles objects, arrays, strings, numbers
+   * (int/long/decimal/double), booleans, and null.
+   */
+  private static void buildJson(VariantBuilder builder, JsonParser parser) throws IOException {
+    JsonToken token = parser.currentToken();
+    if (token == null) {
+      throw new JsonParseException(parser, "Unexpected null token");
+    }
+    switch (token) {
+      case START_OBJECT:
+        buildJsonObject(builder, parser);
+        break;
+      case START_ARRAY:
+        buildJsonArray(builder, parser);
+        break;
+      case VALUE_STRING:
+        builder.appendString(parser.getText());
+        break;
+      case VALUE_NUMBER_INT:
+        buildJsonInteger(builder, parser);
+        break;
+      case VALUE_NUMBER_FLOAT:
+        buildJsonFloat(builder, parser);
+        break;
+      case VALUE_TRUE:
+        builder.appendBoolean(true);
+        break;
+      case VALUE_FALSE:
+        builder.appendBoolean(false);
+        break;
+      case VALUE_NULL:
+        builder.appendNull();
+        break;
+      default:
+        throw new JsonParseException(parser, "Unexpected token " + token);
+    }
+  }
+
+  /**
+   * Builds a Variant object from the current JSON object token.
+   *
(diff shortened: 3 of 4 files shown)


NOTICE
```diff
@@ -73,3 +73,14 @@ notice:
 | See the License for the specific language governing permissions and
 | limitations under the License.
 
+--------------------------------------------------------------------------------
+
+This project includes code from Apache Spark with the following copyright
+notice:
+
+  Apache Spark
```
steveloughran: do cross-ASF projects need this credit? What I do think is good is ensure the original authors get credit in the final commit message

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantParseJson.java
```diff
@@ -0,0 +1,307 @@
...
+package org.apache.parquet.variant;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import java.io.IOException;
+import java.math.BigDecimal;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestVariantParseJson {
```
steveloughran, line 28: 1. nested object parsing?
2. what about invalid json?
* empty file
* not a json file
* incomplete
* large json with many values

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -65,6 +73,152 @@ public VariantBuilder(Metadata metadata) {
...
+   * to double.
+   *
+   * <p>Ported from Apache Spark's {@code VariantBuilder.parseJson}.
+   *
+   * @param json the JSON string to parse
+   * @return the parsed Variant
+   * @throws IOException if the JSON is malformed or an I/O error occurs
+   */
+  public static Variant parseJson(String json) throws IOException {
+    try (JsonParser parser = JSON_FACTORY.createParser(json)) {
```
steveloughran: can be quite slow

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -65,6 +73,152 @@ public VariantBuilder(Metadata metadata) {
...
+   * prefer decimal encoding (no scientific notation) and fall back
+   * to double.
+   *
+   * <p>Ported from Apache Spark's {@code VariantBuilder.parseJson}.
+   *
+   * @param json the JSON string to parse
+   * @return the parsed Variant
+   * @throws IOException if the JSON is malformed or an I/O error occurs
+   */
+  public static Variant parseJson(String json) throws IOException {
```
steveloughran: I was to suggest making CharSequence for feeding in from other places (string fields within avro, ...) but it looks like jackson 2 doesn't support that itself.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
```diff
@@ -30,6 +36,8 @@
  */
 public class VariantBuilder {
 
+  private static final JsonFactory JSON_FACTORY = new JsonFactory();
```
steveloughran: factory should be built with some constraints so that giving it a malicious json file should be rejected rather than trigger OOM problems, etc. From the javadocs.

```
JsonFactory f = JsonFactory.builder()
    .streamReadConstraints(
        StreamReadConstraints.builder()
            .maxNestingDepth(500)
            .maxStringLength(10_000_000)
            .maxDocumentLength(5_000_000)
            .build()
    )
    .build();
```

steveloughran (commented):
I think the json parsing should be in a class alongside VariantBuilder, rather than inside it, some `VariantJsonParser` class.

1. ensures that use of jackson classes are isolated; the core variant does not need jackson on the CP
2. if there is anything which can be done to improve single line JSON content performance, then it can be done there as it's life will span the whole file, rather than a row

Once it's split out, a new question surfaces: should it belong in parquet-jackson? Personally, I think it should

parquet-variant/pom.xml
```diff
@@ -46,6 +46,17 @@
       <artifactId>parquet-column</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${jackson.groupId}</groupId>
```
steveloughran, line 50: comes with the parquet-jackson module; no need to reimport at a different scope

steveloughran (commented):
Commented. As I said before, I think this should go into parquet-jackson.

I see you've added jackson-core to the classpath. Is that because the parquet-jackson ones are shaded? If the class was in that module then it'd not be an issue.

of course, adding parquet-variant to the parquet-jackson module dependency list is its own complications, but I don't see any loops in the build.

Why not raise this PR on the dev list and bring up the "where should it go?" question.

parquet-variant/pom.xml
gaurav7261: Hi @steveloughran , email is sent to `dev@parquet.apache.org`, can you please check there

gaurav7261: Hi @steveloughran , saw your reply on email, so all looks good to you in above PR

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
```diff
@@ -0,0 +1,191 @@
...
+        break;
+      case VALUE_NULL:
+        builder.appendNull();
+        break;
+      default:
+        throw new JsonParseException(parser, "Unexpected token " + token);
+    }
+  }
+
+  private static void buildJsonObject(VariantBuilder builder, JsonParser parser) throws IOException {
```
steveloughran, line 132: nit: add javadoc explaining how this will co-recurse into buildJSON

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
```diff
@@ -0,0 +1,191 @@
...
+    VariantObjectBuilder obj = builder.startObject();
+    while (parser.nextToken() != JsonToken.END_OBJECT) {
+      obj.appendKey(parser.currentName());
+      parser.nextToken();
+      buildJson(obj, parser);
+    }
+    builder.endObject();
+  }
+
+  private static void buildJsonArray(VariantBuilder builder, JsonParser parser) throws IOException {
```
steveloughran, line 150: same comment as above: mention co-recursion

steveloughran (approved):
I"m happy. Reviewed code and suggested some javadocs. Co-recursion is rare in production procedural code, so worth highlighting.

Test coverage good

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
gaurav7261: done @steveloughran , please check

steveloughran:
gaurav, you'll need to ask on the mail list for approval to run the workflows. Best to only do incremental changes on a PR

gaurav7261:
@steveloughran i don't have access to email archive, @alamb  created thread for me, @alamb  can you please help here?

gaurav7261:
@steveloughran can you please help here

gaurav7261:
@wgtmac i have fixed mvn spotless, please rerun

gaurav7261:
@wgtmac java version issue in local, now fixed with java 17 spotless,please rerun again, thanks

gaurav7261:
@wgtmac can you please help run workflow

gaurav7261:
@emkornfield  CI passed, can we merge now please

emkornfield:
@gaurav7261 I think we need a committer to take a look.  CC @wgtmac @Fokko @julienledem if you have time, otherwise I can take a look but I'm not super familiar with the java side of things.

.gitignore
```diff
@@ -20,4 +20,5 @@ target/
 mvn_install.log
 .vscode/*
 .DS_Store
+.memsearch/
```
wgtmac, line 23: Do we really need it in this PR?

parquet-variant/pom.xml
```diff
@@ -46,6 +46,17 @@
...
+      <artifactId>jackson-core</artifactId>
```
wgtmac, line 51: Does this dependency change look good to you? @gszadovszky @Fokko

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
```diff
@@ -0,0 +1,206 @@
...
+   * Recursively builds a Variant value from the current position of a
+   * Jackson streaming parser. Handles objects, arrays, strings, numbers
+   * (int/long/decimal/double), booleans, and null.
+   */
+  private static void buildJson(VariantBuilder builder, JsonParser parser) throws IOException {
+    JsonToken token = parser.currentToken();
+    if (token == null) {
+      throw new JsonParseException(parser, "Unexpected null token");
+    }
+    switch (token) {
```
wgtmac, line 94: Is there any chance to parse types like date and timestamp?

wgtmac (approved):
+1 from my side

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
gaurav7261: i think jackson will treat them string only

.gitignore
gaurav7261: not as such @wgtmac , if you need i can remove it

gaurav7261:
@wgtmac  @Fokko can you please review and merge

wgtmac:
Thanks @gaurav7261 for working on this and @steveloughran for the review!

gaurav7261:
thanks @wgtmac, @steveloughran  , i see release 1.17.1 is released 5 hrs ago, just want to know  by when next release will get release so that i can use same in flink parquet and confluent kafka s3 sink connector

wgtmac:
1.17.1 is a patched release with bug fixes only. For this feature, it is targeted in the 1.18.0 which I don't have an ETA yet. Perhaps you may want to reply to the relevant discussion thread in the dev@parquet.apache.org.

PR #3446, opened 2026-03-16
GH-3307: Replace deprecated BIT_PACKED with RLE in DevNullValuesWriter

### LINKED ISSUE #3307: BIT_PACKING is written by default when Definition or Repetition levels are empty
As discussed in https://github.com/apache/arrow/discussions/47113 , this package is writing the deprecated BIT_PACKING encoding to the column metadata and the DataPageHeader. This is only seems to happens when rep/def levels are empty. I think this is where it's happening:

```java
public class DevNullValuesWriter extends ValuesWriter {
  ....
  public Encoding getEncoding() {
    return BIT_PACKED;
  }
}
```

Wouldn't it make sense to just return RLE?

I've seen this in parquet-java 1.15.1 and Overture Maps files using 1.13.1.

LuciferYang:
### Rationale for this change
Fixes #3307

When repetition/definition levels are empty (i.e., `maxLevel == 0` for required, non-repeated fields), `DevNullValuesWriter` is used as a no-op writer that produces zero bytes. However, its `getEncoding()` method returns the deprecated `BIT_PACKED` encoding, which gets written into the `DataPageHeader` metadata (`repetition_level_encoding` / `definition_level_encoding`) and the column chunk encoding list.

Since parquet-java already uses `RLE` as the encoding for levels, the metadata should reflect `RLE` rather than the deprecated `BIT_PACKED`. 

Changed `DevNullValuesWriter.getEncoding()` to return `Encoding.RLE` instead of `Encoding.BIT_PACKED`.

### Are these changes tested?
 Yes. All existing tests in `parquet-column` and `parquet-hadoop` pass without modification.

Newly written Parquet files will report `RLE` instead of `BIT_PACKED` in page header metadata for empty repetition/definition levels. This has no impact on file compatibility — readers do not decode level data when the byte length is zero, regardless of the encoding value in the header.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/DevNullValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/DevNullValuesWriter.java
@@ -18,7 +18,7 @@
  */
 package org.apache.parquet.column.values.bitpacking;
 
-import static org.apache.parquet.column.Encoding.BIT_PACKED;
+import static org.apache.parquet.column.Encoding.RLE;
 
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.column.Encoding;
@@ -72,7 +72,7 @@ public class DevNullValuesWriter extends ValuesWriter {
 
   @Override
   public Encoding getEncoding() {
-    return BIT_PACKED;
+    return RLE;
   }
 
   @Override


dossett:
+1 (non-binding)

parquet-column/src/main/java/org/apache/parquet/column/values/bitpacking/DevNullValuesWriter.java
```diff
@@ -72,7 +72,7 @@ public long getAllocatedSize() {
 
   @Override
   public Encoding getEncoding() {
-    return BIT_PACKED;
+    return RLE;
```
wgtmac, line 75: @gszadovszky @Fokko Is this a minor breaking change? I remember that this discussion has appeared before but I cannot find the source.

PR #3448, opened 2026-03-16
GH-3449: Fix BinaryStats int overflow

### LINKED ISSUE #3449: Fix BinaryStats int overflow

`BinaryStatistics.isSmallerThan()` computes `min.length() + max.length()` as `int + int`. When a binary column contains values >= 2^30 bytes (1 GB), this sum overflows to a negative number, which is then incorrectly evaluated as smaller than `MAX_STATS_SIZE` (4096). This causes the full multi-GB min/max statistics to be serialized into the Thrift footer instead of being dropped, resulting in a corrupted Parquet file whose footer exceeds the 4-byte length field and cannot be read back.

### Component(s)

_No response_

Jiayi-Wang-db:
### Rationale for this change
`BinaryStatistics.isSmallerThan()` computes `min.length() + max.length()` as `int + int`. When a binary column contains values >= 2^30 bytes (1 GB), this sum overflows to a negative number, which is then incorrectly evaluated as smaller than `MAX_STATS_SIZE` (4096). This causes the full multi-GB min/max statistics to be serialized into the Thrift footer instead of being dropped, resulting in a corrupted Parquet file whose footer exceeds the 4-byte length field and cannot be read back.

Promote int arithmetic to long before addition in `BinaryStatistics.isSmallerThan()` and `isSmallerThanWithTruncation()` to prevent integer overflow when binary values are very large.

### Are these changes tested?
Yes. Added a unit test in TestStatistics that creates a 1 GB binary value and asserts isSmallerThan(4096) correctly returns false.

Parquet files with very large binary values (>= 1 GB) will no longer have corrupted footers. The oversized statistics are now correctly dropped, producing a valid and readable file.

Closes #3449

### DIFF
--- a/parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
@@ -930,11 +930,10 @@ public class TestStatistics {
 
   @Test
   public void testBinaryIsSmallerThanNoOverflowForLargeValues() {
-    PrimitiveType type = Types.required(BINARY).named("test_binary");
-    Statistics<?> stats = Statistics.getBuilderForReading(type).build();
-
-    byte[] largeValue = new byte[1_073_741_824]; // 2^30 = 1 GB
-    stats.setMinMaxFromBytes(largeValue, largeValue);
+    BinaryStatistics stats = new BinaryStatistics();
+    // Create a Binary whose length() reports 2^30 without allocating 1 GB
+    Binary fakeLarge = Binary.fromConstantByteArray(new byte[0], 0, 1 << 30);
+    stats.updateStats(fakeLarge);
 
     // min.length() + max.length() = 2^31, must not overflow int to negative
     assertFalse(stats.isSmallerThan(4096));


parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
```diff
@@ -927,4 +927,16 @@ public void testNoopStatistics() {
     assertThrows(UnsupportedOperationException.class, stats::minAsString);
     assertThrows(UnsupportedOperationException.class, () -> stats.isSmallerThan(0));
   }
+
+  @Test
+  public void testBinaryIsSmallerThanNoOverflowForLargeValues() {
+    PrimitiveType type = Types.required(BINARY).named("test_binary");
+    Statistics<?> stats = Statistics.getBuilderForReading(type).build();
+
+    byte[] largeValue = new byte[1_073_741_824]; // 2^30 = 1 GB
```
wgtmac: Do we really need to allocate such large memory in the test?

Jiayi-Wang-db: Thanks, fixed.

PR #3450, opened 2026-03-17
GH-3011: Deny further writes after InternalParquetRecordWriter is aborted

### LINKED ISSUE #3011: Should deny further write if writer is aborted

After [PARQUET-2437](https://github.com/apache/parquet-java/pull/1285), InternalParquetRecordWriter will switch to aborted state if write throws, but still accept further writes even though these writes will not be flushed during close, leaving the file incomplete. Users will not notice the corrupted file until they actually read the file.

Should we check for aborted state and reject further writes to avoid such surprises?
@gszadovszky 

### Component(s)

_No response_

LuciferYang:
### Rationale for this change

When a write fails (e.g. OOM during page flush), `InternalParquetRecordWriter` sets `aborted = true` and re-throws. But if the caller catches the exception and keeps calling `write()`, nothing stops them — all calls go through without error. Then `close()` sees `aborted` and skips the flush, so all those records are just gone. The caller won't know until they read the file back and find data missing.

Granted, writing after an error is the caller's fault. But silently discarding data makes it much harder to track down what went wrong. `java.io.FileOutputStream` handles a similar situation the same way — `write()` after `close()` is also misuse, but it still throws `IOException("Stream Closed")` instead of quietly accepting bytes that will never be persisted.

Check `aborted` at the top of `InternalParquetRecordWriter.write()`. If aborted, throw `IOException` right away.

### Are these changes tested?

Added `testWriteAfterAbortShouldThrow` in `TestParquetWriterError`:

1. Write to an aborted writer → `IOException` with expected message
2. `close()` on an aborted writer → no exception

Existing tests in `parquet-hadoop` still pass.

If you used to catch a write exception and keep writing to the same `ParquetWriter`, you'll now get an `IOException` on the next `write()`. Fix: discard the writer and create a new one.

Closes #3011


wgtmac:
I agree with @Jiayi-Wang-db that once any exception reported by the writer, users should not continue to call `write` any more, otherwise it is a misuse.

LuciferYang:
Hi @wgtmac, agree — calling `write` after an exception is definitely a misuse. This PR just makes that misuse fail fast instead of failing silently.

Right now the writer quietly accepts all writes after abort, and `close()` drops them without a word. The user won't notice until they read the file and find data missing. Throwing an `IOException` right away at least gives them a clear signal.

`java.io.FileOutputStream` does the same thing — `write()` after `close()` is a misuse too, but it still throws `IOException("Stream Closed")` rather than silently discarding bytes.

That said, if you think this isn't worth the change, happy to close this PR and the issue. Let me know.

wgtmac (approved):
Thanks @LuciferYang for the quick response! I think overall the fix in this PR is benign so I'm fine with it.

LuciferYang:
Thank you @wgtmac ~

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriterError.java
```diff
@@ -58,6 +62,48 @@ public class TestParquetWriterError {
...
+      Field internalWriterField = ParquetWriter.class.getDeclaredField("writer");
+      internalWriterField.setAccessible(true);
+      InternalParquetRecordWriter<?> internalWriter =
+          (InternalParquetRecordWriter<?>) internalWriterField.get(writer);
+      Field abortedField = InternalParquetRecordWriter.class.getDeclaredField("aborted");
+      abortedField.setAccessible(true);
+      abortedField.setBoolean(internalWriter, true);
+
+      // Now try to write again - this should throw IOException
+      IOException e = Assert.assertThrows(
```
LuciferYang, line 94: Use `Assert.assertThrows` to better align with the style of other test cases in the project.(like TestIndexCache)

wgtmac (approved):

PR #3452, opened 2026-03-19
GH-3451. Add a JMH benchmark for variants

### LINKED ISSUE #3451: Add a JMH benchmark for variants

similar to what I'm doing for iceberg in https://github.com/apache/iceberg/issues/15628 , add a jmh benchmark to measure variant performance
* build
* serialize
* deserialize
* shed

Some of these operations take microseconds, and while x86 has the `RDTSC` opcode for nS granularity, ARM parts don't seem as accurate, there's a huge variance here. All benchmarks will be designed to take 10s or 100s of uS though multiple iterations, with the slower operations scaled by the same amount to make comparision easier, especially when looking at graphs

### Component(s)

_No response_

steveloughran:
### Rationale for this change

There's no benchmark for variant IO and so there's no knowledge of any problems which exist now, or any way to detect regressions.

* adds parquet-variant to parquet-benchmark dependencies
* new JMH benchmark `VariantBuilderBenchmark` to measure builder cost.
* VariantProjectionBenchmark: selective read of fields from shredded/unshredded files
* Improves VariantConverter of a binary to a string through a new package-private method in VariantBuilder. Benchmark `VariantConverterBenchmark` shows a 20% reduction in time to convert a marshalled string to a java string within a builder. This codepath is used when reconstructing partially shredded variants.

### Are these changes tested?

Apart from a minor change in VariantConverter/VariantBuilder, this is all benchmark

See https://steveloughran.github.io/benchmarking-variants/ for the writeup and the interactive benchmark results of Iceberg and Parquet benchmarks.

No

Closes #3451

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/VariantBenchmark.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/VariantBenchmark.java
@@ -429,17 +429,19 @@ public class VariantBenchmark {
    * the field-name lookup and type dispatch that a query engine performs on every row. Nested
    * objects are recursively traversed so that {@code depth=Nested} incurs the full deserialization
    * cost of sub-objects.
+   * @param blackhole black hole.
    */
   @Benchmark
-  public void deserializeVariant(Blackhole bh) {
+  public void deserializeVariant(Blackhole blackhole) {
     for (int j = 0; j < ITERATIONS; j++) {
-      deserializeAndConsume(preBuiltVariant, bh);
+      deserializeAndConsume(preBuiltVariant, blackhole);
     }
   }
 
   /**
    * Shred the pre-built variant into a fully typed schema. Measures the cost of type dispatch,
    * field matching, and recursive decomposition that {@link VariantValueWriter} perform
+   * @param blackhole black hole.
    */
   @Benchmark
   public void consumeRecordsShredded(Blackhole blackhole) {
@@ -453,6 +455,7 @@ public class VariantBenchmark {
    * Write {@link #FILE_ROWS} rows of the pre-built variant to an in-memory Parquet file using the
    * shredded schema. Measures end-to-end Parquet encoding cost including page/row-group framing.
    * Compare with {@link #consumeRecordsShredded} to quantify the overhead over raw schema traversal.
+   * @param blackhole black hole.
    */
   @Benchmark
   public void writeToMemoryFile(Blackhole blackhole) throws IOException {
@@ -463,6 +466,7 @@ public class VariantBenchmark {
    * Write the pre-built variant to an unshredded schema (metadata + value only).
    * This is the baseline: the entire variant is written as a single binary blob.
    * Compare with {@link #consumeRecordsShredded} to see the cost of shredding.
+   * @param blackhole black hole.
    */
   @Benchmark
   public void consumeRecordsUnshredded(Blackhole blackhole) {
@@ -475,6 +479,7 @@ public class VariantBenchmark {
   /**
    * Write {@link #FILE_ROWS} rows of the pre-built variant to an in-memory Parquet file using the
    * unshredded schema (metadata + value binary blobs only). Baseline for {@link #writeToMemoryFile}.
+   * @param blackhole black hole.
    */
   @Benchmark
   public void writeToMemoryUnshredded(Blackhole blackhole) throws IOException {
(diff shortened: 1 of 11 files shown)


parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/VariantBenchmark.java
```diff
@@ -0,0 +1,551 @@
...
+  private static int counter;
+
+  /**
+   * Get a count value.
+   * @return a new value.
+   */
+  private static int count() {
+    int c = counter++;
+    if (c >= 512) {
+      c = 0;
```
xiaoxuandev: only resets the local copy, counter keeps growing？

steveloughran: good point. will fix.

steveloughran:
build failures are all because java11 javadoc is extra-fussy than the versions either side of it.

steveloughran:
Flame graph profiles so you can compare what is using CPU time when working with variants

[profiles.zip](https://github.com/user-attachments/files/26445694/profiles.zip)

Variant rebuild is expensive, I've got a minor (and will keep package private) tweak to VariantBuilder to allow VariantStringConverter to add a Binary as a string to the variant without converting it to a string en-route, which would save a lot of needless string creation.

Variant.getFieldAtIndex() is pretty expensive too.

steveloughran:
Slides. Reading the PR will make the notion of file vs lean schema clearer.

[2026-04-01-variant reads considered suboptimal](https://github.com/user-attachments/files/26575558/2026-04-01-variant.reads.considered.suboptimal.2.pdf)

Fokko (approved):
This looks great to me, thanks @steveloughran 🙌

steveloughran:
thanks @Fokko! Can you do the merge?

nssalian (approved):
LGTM.

steveloughran:
just doing some feature creep here
* make col4 a long string (and name it as such)
* add a uuid

steveloughran:
latest numbers

before adding uuid read (but with that long column)
```
Benchmark                                           (tableType)  Mode  Cnt     Score    Error  Units
VariantProjectionBenchmark.readAllRecords            Unshredded    ss   10  1645.855 ± 27.618  ms/op
VariantProjectionBenchmark.readAllRecords              Shredded    ss   10  2381.192 ± 41.940  ms/op
VariantProjectionBenchmark.readProjectedFileSchema   Unshredded    ss   10   932.050 ± 37.143  ms/op
VariantProjectionBenchmark.readProjectedFileSchema     Shredded    ss   10  1596.800 ± 50.421  ms/op
VariantProjectionBenchmark.readProjectedLeanSchema   Unshredded    ss   10  1750.998 ± 10.982  ms/op
VariantProjectionBenchmark.readProjectedLeanSchema     Shredded    ss   10   724.603 ± 18.377  ms/op
```
after adding UUID variant field.
```
Benchmark                                           (tableType)  Mode  Cnt     Score     Error  Units
VariantProjectionBenchmark.readAllRecords            Unshredded    ss   10  1913.765 ±  18.896  ms/op
VariantProjectionBenchmark.readAllRecords              Shredded    ss   10  2679.631 ± 150.978  ms/op
VariantProjectionBenchmark.readProjectedFileSchema   Unshredded    ss   10   910.009 ±  33.074  ms/op
VariantProjectionBenchmark.readProjectedFileSchema     Shredded    ss   10  1616.585 ±  57.818  ms/op
VariantProjectionBenchmark.readProjectedLeanSchema   Unshredded    ss   10  1777.288 ±  14.049  ms/op
VariantProjectionBenchmark.readProjectedLeanSchema     Shredded    ss   10   723.679 ±   8.473  ms/op

```

points to note
* full record read is now ~ 15% slower on both unshredded and shredded files just by adding a UUID. Surprisingly Expensive.
* Reading all records on a shredded file is still ~40% slower than on an unshredded one.
* same odd behaviours on a projected schema

steveloughran:
No further changes planned...now it's time for people to look at those numbers and think of improvements. Looking forward!

nssalian:
@steveloughran Let's keep the benchmark code here and move the changes to #3481 (I've already pulled some of the changes there). It's great to verify here, but since the original intent was to add JMH benchmarks, let's scope this to that and get the improvements in another focused PR.

steveloughran:
@nssalian will do.

Fokko:
Thanks @steveloughran for working on this, and thanks @nssalian and @xiaoxuandev for the review 🚀

steveloughran:
@Fokko thanks for merging this!

PR #3458, opened 2026-03-28
GH-3457: Validate column index for null_pages/null_counts contradiction

### LINKED ISSUE #3457: Column index with null_pages=true and null_counts=0 causes silent data loss during filter pushdown
### Describe the bug

When a Parquet file contains a `ColumnIndex` where `null_pages[i]` is `true` and `null_counts[i]` is `0` for the same page, parquet-java's column index filtering silently drops that page from query results. No error or warning is produced.

Per the [Parquet format specification](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift), `null_pages[i]=true` means "a page contains only null values" and `null_counts[i]` is "the number of null values" for each page. These two fields directly contradict each other: a page cannot contain only null values while also having zero null values.

### Why it causes data loss

The column index filtering in `ColumnIndexBase` has two code paths for evaluating predicates, and both exclude pages with this contradiction:

**Non-null predicates** (e.g., `WHERE col = 50`): The `BoundaryOrder` comparators iterate over `pageIndexes` — an internal array that maps min/max array positions to page numbers, omitting pages where `null_pages[i]` is `true`. Pages marked as null have no entry in this array and are never evaluated by the comparators, so their rows are excluded from results.

**Null predicates** (e.g., `WHERE col IS NULL`): `ColumnIndexBase.visit(Eq)` checks `nullCounts[pageIndex] > 0` (line ~299 on master), which returns `false` when `null_counts[i]` is `0`. The page is excluded.

A page with this contradiction is invisible to all predicates. Only unfiltered reads (no `WHERE` clause) return correct results.

### Proposed fix

Add validation in `ColumnIndexBuilder.build(PrimitiveType)` to detect the contradiction and return `null`, following the existing pattern where this method already returns `null` for other kinds of invalid metadata:

```java
if (nullPages.isEmpty()) {
  return null;
}
ColumnIndexBase<?> columnIndex = createColumnIndex(type);
if (columnIndex == null) {
  // Might happen if the specialized builder discovers invalid min/max values
  return null;
}
```

The `null` propagates through the read path (`fromParquetColumnIndex` → `readColumnIndex` → `getColumnIndex` → `ColumnIndexFilter.applyPredicate`), causing the filter to fall back to reading all pages for the affected column. Row-group-level statistics filtering and other columns are unaffected. Performance overhead should be negligible — `build()` already loops over pages multiple times, and this check is a single boolean read per page, short-circuited by `&&`.

### Existing precedent

`ColumnIndexBuilder.build(PrimitiveType)` already returns `null` when the column index has invalid min/max values (`createColumnIndex` returns null). The proposed fix adds one more validation of the same kind — checking a different field (`null_pages` vs `null_counts`) for internal consistency, with the same outcome (`return null`).

More broadly, parquet-java has defensive handling for invalid writer metadata in other areas:

- **`CorruptStatistics`** ([PARQUET-251](https://issues.apache.org/jira/browse/PARQUET-251)): Ignores statistics from writers known to produce invalid binary column statistics.
- **`CorruptDeltaByteArrays`** ([PARQUET-246](https://issues.apache.org/jira/browse/PARQUET-246)): Forces sequential reads for files with broken delta byte array encoding.

### How to reproduce

```java
PrimitiveType type = Types.required(INT32).named("col");

// Pages 1-2 have null_pages=true but null_counts=0 — contradictory
ColumnIndex ci = ColumnIndexBuilder.build(
    type,
    BoundaryOrder.ASCENDING,
    List.of(false, true, true),
    List.of(0L, 0L, 0L),
    List.of(ByteBuffer.allocate(4), ByteBuffer.allocate(0), ByteBuffer.allocate(0)),
    List.of(ByteBuffer.allocate(4), ByteBuffer.allocate(0), ByteBuffer.allocate(0)));

// ci is non-null — the contradiction is not detected.
// Pages 1-2 are silently excluded from all column-index-based filtering.
```

### Additional fix

The static `ColumnIndexBuilder.build()` method (the read-path entry point) does not null-check the return value of `build(PrimitiveType)` before dereferencing it, which could cause an NPE. The fix includes a null-guard for this.

clee704:
Fixes #3457

Add validation in `ColumnIndexBuilder.build(PrimitiveType)` to detect a contradictory column index where `null_pages[i]` is `true` (page consists entirely of null values) but `null_counts[i]` is `0` (page has zero null values). When detected, `build()` returns `null`, following the existing pattern for invalid min/max values.

Also fix a pre-existing NPE in the static `build()` method where `build(type)` could return `null` but was not null-checked before accessing `boundaryOrder`.

Column index filtering silently excludes pages with this contradiction from query results for all predicates:
- **Non-null predicates** (e.g., `WHERE col = 50`): `BoundaryOrder` comparators iterate over `pageIndexes`, which omits pages where `null_pages[i]` is `true`. These pages are never evaluated and their rows are excluded.
- **Null predicates** (e.g., `WHERE col IS NULL`): `ColumnIndexBase.visit(Eq)` checks `nullCounts[pageIndex] > 0`, which returns `false` when `null_counts` is `0`. The page is excluded.

The result is silent data loss with no error or warning. Only unfiltered reads return correct results.

`ColumnIndexBuilder.build(PrimitiveType)` already returns `null` for other kinds of invalid metadata (empty pages, invalid min/max values). This adds one more validation of the same kind.

### How was this patch tested?

3 new test methods (15 assertions) in `TestColumnIndexBuilder`:
- `testBuildReturnsNullForNullPageCountContradiction`: 5 rejection cases — contradiction at first/middle/last page, single page, all pages
- `testBuildPreservesValidColumnIndex`: 6 preservation cases — legitimate null pages, all non-null pages, single pages, boundary null_counts=1
- `testBuildWithoutNullCountsIsNotRejected`: null_counts absent (optional field) is not rejected

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
@@ -616,6 +616,9 @@ public abstract class ColumnIndexBuilder {
 
     builder.fill(nullPages, nullCounts, minValues, maxValues, repLevelHistogram, defLevelHistogram);
     ColumnIndexBase<?> columnIndex = builder.build(type);
+    if (columnIndex == null) {
+      return null;
+    }
     columnIndex.boundaryOrder = requireNonNull(boundaryOrder);
     return columnIndex;
   }
@@ -749,6 +752,19 @@ public abstract class ColumnIndexBuilder {
     // Null counts is optional so keep it null if the builder has no values
     if (!nullCounts.isEmpty()) {
       columnIndex.nullCounts = nullCounts.toLongArray();
+    }
+    // A null_pages entry of true indicates the page consists entirely of null values.
+    // When null_counts is present, a null_counts entry of zero means the page has no
+    // null values at all. These two fields directly contradict each other: a page
+    // cannot both consist entirely of nulls and contain zero nulls. Since null_pages
+    // is the field that causes the reader to skip pages during column index filtering,
+    // the safe response is to discard the column index for this column entirely.
+    if (columnIndex.nullCounts != null) {
+      for (int i = 0; i < columnIndex.nullPages.length; i++) {
+        if (columnIndex.nullPages[i] && columnIndex.nullCounts[i] == 0L) {
+          return null;
+        }
+      }
     }
     columnIndex.pageIndexes = pageIndexes.toIntArray();
     // Repetition and definition level histograms are optional so keep them null if the builder has no values
--- a/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
@@ -761,7 +761,7 @@ public abstract class ColumnIndexBuilder {
     // the safe response is to discard the column index for this column entirely.
     if (columnIndex.nullCounts != null) {
       for (int i = 0; i < columnIndex.nullPages.length; i++) {
-        if (columnIndex.nullPages[i] && columnIndex.nullCounts[i] == 0L) {
+        if (columnIndex.nullPages[i] && columnIndex.nullCounts[i] <= 0L) {
           return null;
         }
       }
--- a/parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBoundaryOrder.java
+++ b/parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBoundaryOrder.java
@@ -250,9 +250,11 @@ public class TestBoundaryOrder {
     RAND_DESCENDING = (ColumnIndexBase<?>) builder.build();
 
     builder = ColumnIndexBuilder.getBuilder(TYPE, Integer.MAX_VALUE);
-    // Add a couple of empty stats so column index will contain null pages only
+    // Add a couple of all-null stats so column index will contain null pages only
     for (int i = 0; i < 10; ++i) {
-      builder.add(Statistics.createStats(TYPE));
+      Statistics<?> nullStats = Statistics.createStats(TYPE);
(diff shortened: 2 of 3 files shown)


parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
```diff
@@ -750,6 +753,19 @@ private ColumnIndexBase<?> build(PrimitiveType type) {
...
     }
+    // A null_pages entry of true indicates the page consists entirely of null values.
+    // When null_counts is present, a null_counts entry of zero means the page has no
+    // null values at all. These two fields directly contradict each other: a page
+    // cannot both consist entirely of nulls and contain zero nulls. Since null_pages
+    // is the field that causes the reader to skip pages during column index filtering,
+    // the safe response is to discard the column index for this column entirely.
+    if (columnIndex.nullCounts != null) {
+      for (int i = 0; i < columnIndex.nullPages.length; i++) {
+        if (columnIndex.nullPages[i] && columnIndex.nullCounts[i] == 0L) {
```
wgtmac:
```suggestion
        if (columnIndex.nullPages[i] && columnIndex.nullCounts[i] <= 0L) {
```

Perhaps this is much better?

wgtmac (commented):
Thanks @clee704 for improving this! While I doubt that if this could really happen, but it would be good for the reader to gracefully deal with malformed stats.

parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/ColumnIndexBuilder.java
clee704: Thanks for the suggestion! Applied

wgtmac:
Could you please double check the test cases to make CI happy?

clee704:
@wgtmac Done. Can you please approve the CI?

wgtmac (approved):

PR #3465, opened 2026-04-06
GH-3464: Improve `DeltaByteArrayWriter.writeBytes` to avoid unnecessary allocation and scalar prefix comparison

### LINKED ISSUE #3464: Improve `DeltaByteArrayWriter.writeBytes` to avoid unnecessary allocation and scalar prefix comparison

`DeltaByteArrayWriter.writeBytes()` has two performance issues in its hot path, executed for every Binary value written with `DELTA_BYTE_ARRAY` encoding (the default for BINARY columns in V2 data pages):

### Problem 1: `getBytes()` allocates a new `byte[]` for every value

[Line 92](https://github.com/apache/parquet-java/blob/4c8f4d4b/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java#L92) calls `v.getBytes()`:

```java
byte[] vb = v.getBytes();
```

`Binary.getBytes()` **always** allocates a new `byte[]` and copies into it — even for `ByteArrayBackedBinary` (which does `Arrays.copyOfRange`) and `ByteBufferBackedBinary` (which allocates `new byte[length]` and copies from the ByteBuffer). The method `getBytesUnsafe()` exists and returns the backing array directly when the Binary owns its bytes (`isBackingBytesReused == false`), avoiding the copy entirely.

The returned `byte[]` (`vb`) is used for two purposes:

1. **Prefix comparison** against `previous` (line 95) — read-only, `getBytesUnsafe()` is sufficient
2. **Assigned to `previous`** for the next iteration (line 99) — must be an owned copy that won't be mutated by the caller

When `isBackingBytesReused` is `false` (common case for values read from Parquet pages via `ColumnReader`), `getBytesUnsafe()` returns a stable array that is safe to retain. When `isBackingBytesReused` is `true`, a defensive copy is needed only for `previous`.

### Problem 2: byte-by-byte prefix comparison loop

[Line 95](https://github.com/apache/parquet-java/blob/4c8f4d4b/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java#L95) finds the common prefix between the current and previous value:

```java
for (i = 0; (i < length) && (previous[i] == vb[i]); i++)
    ;
```

This byte-by-byte loop cannot be vectorized by the JIT because of the per-element early-exit condition. `Arrays.mismatch(byte[], int, int, byte[], int, int)` (available since Java 9) performs the same operation but is [intrinsified by the JVM](https://bugs.openjdk.org/browse/JDK-8033148) to use SIMD vector comparison.

### Component(s)

Core

arouel:
### Rationale for this change

`DeltaByteArrayWriter.writeBytes()` is on the hot path for `DELTA_BYTE_ARRAY` encoding and had avoidable overhead:

1. **Per-value allocation from `getBytes()`.**  
   `getBytes()` always creates a new array. For prefix comparison this copy is unnecessary.

2. **Scalar prefix scan.**  
   The byte-by-byte loop is replaced with `Arrays.mismatch(...)`, which maps to optimized JVM intrinsics.

In profiling (custom JFR benchmark on a large merge workload), this method was a top allocation hotspot before the change.

In `DeltaByteArrayWriter.writeBytes()`:

- `v.getBytes()` -> `v.getBytesUnsafe()` for read-only prefix comparison.
- Manual prefix loop -> `Arrays.mismatch(previous, 0, length, vb, 0, length)`.
- `previous = vb` -> `previous = v.isBackingBytesReused() ? v.getBytes() : vb`  
  (defensive copy only when the backing bytes may be reused by the caller).

This preserves semantics while removing unnecessary allocations in the common case.

### Benchmark signal (custom, directional)

On a custom JFR-profiled merge workload (180M rows, 4 binary columns), this change significantly reduced allocations and lowered CPU attributed to this path.  
(Results are workload/JDK dependent and provided as directional evidence.)

### Are these changes tested?

Yes.

- Existing coverage: `TestDeltaByteArray` round-trip tests (including skip/skipN/reset paths).
- Added regression test: `testReusedBackingArrayRegression` in `TestDeltaByteArray` to verify correctness when the same mutable backing array is reused across writes.

No API or format changes. This is a transparent performance optimization; encoded data remains compatible/interchangeable.

Closes #3464

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
@@ -89,7 +89,7 @@ public class DeltaByteArrayWriter extends ValuesWriter {
 
   @Override
   public void writeBytes(Binary v) {
-    byte[] vb = v.getBytesUnsafe();
+    byte[] vb = v.isBackingBytesReused() ? v.getBytes() : v.getBytesUnsafe();
     int length = Math.min(previous.length, vb.length);
     // Find the number of matching prefix bytes between this value and the previous one.
     // Arrays.mismatch is intrinsified by the JVM to use SIMD instructions.
@@ -99,9 +99,6 @@ public class DeltaByteArrayWriter extends ValuesWriter {
     }
     prefixLengthWriter.writeInteger(i);
     suffixWriter.writeBytes(v.slice(i, vb.length - i));
-    // Retain an owned copy for prefix comparison with the next value.
-    // getBytesUnsafe() may return the backing array directly, so we must copy
-    // if the Binary's backing bytes may be reused by the caller.
-    previous = v.isBackingBytesReused() ? v.getBytes() : vb;
+    previous = vb;
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.column.values.deltastrings;
 
+import java.util.Arrays;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.column.Encoding;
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
```diff
@@ -88,14 +89,19 @@ public String memUsageString(String prefix) {
 
   @Override
   public void writeBytes(Binary v) {
-    int i = 0;
-    byte[] vb = v.getBytes();
-    int length = previous.length < vb.length ? previous.length : vb.length;
-    // find the number of matching prefix bytes between this value and the previous one
-    for (i = 0; (i < length) && (previous[i] == vb[i]); i++)
-      ;
+    byte[] vb = v.getBytesUnsafe();
```
Fokko:
```suggestion
    byte[] vb = v.isBackingBytesReused() ? v.getBytes() : v.getBytesUnsafe();
```

parquet-column/src/main/java/org/apache/parquet/column/values/deltastrings/DeltaByteArrayWriter.java
```diff
@@ -88,14 +89,19 @@ public String memUsageString(String prefix) {
...
+    if (i < 0) {
+      i = length; // all bytes in the common range matched
+    }
     prefixLengthWriter.writeInteger(i);
     suffixWriter.writeBytes(v.slice(i, vb.length - i));
-    previous = vb;
+    // Retain an owned copy for prefix comparison with the next value.
+    // getBytesUnsafe() may return the backing array directly, so we must copy
+    // if the Binary's backing bytes may be reused by the caller.
+    previous = v.isBackingBytesReused() ? v.getBytes() : vb;
```
Fokko:
```suggestion
    previous = vb;
```

Fokko (approved):
Left a minor optimization, but apart from that, this looks great to me 👍

arouel:
@Fokko thank you for the feedback. I applied your suggestions.

Fokko:
Thanks for the follow up @arouel Let's get this in, this looks great! 🙌 Thanks for introducing `Arrays.mismatch` here!

PR #3467, opened 2026-04-06
GH-3466: Improve `RunLengthBitPackingHybridDecoder.readNext` to avoid per-call buffer allocation and `DataInputStream` wrapping

### LINKED ISSUE #3466: Improve `RunLengthBitPackingHybridDecoder.readNext` to avoid per-call buffer allocation and `DataInputStream` wrapping

`RunLengthBitPackingHybridDecoder.readNext()` allocates a new `int[]` and `byte[]` on every PACKED-mode call. In workloads that decode many bit-packed runs (definition levels, repetition levels, RLE-encoded integers), these allocations dominate the read-side allocation profile. The upstream code even acknowledges this with a `// TODO: reuse a buffer` comment.

### Problem 1: per-call buffer allocation

[Lines 94–95](https://github.com/apache/parquet-java/blob/4c8f4d4b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java#L94-L95) allocate fresh arrays on every PACKED-mode `readNext()`:

```java
currentBuffer = new int[currentCount]; // TODO: reuse a buffer
byte[] bytes = new byte[numGroups * bitWidth];
```

`currentCount` is always `numGroups * 8`, and `numGroups` is typically small (1–16 groups = 8–128 values per run). These allocations are individually modest but occur thousands of times per column chunk — once per bit-packed run. In a 180M-row merge with multiple integer/boolean columns, the cumulative allocation is substantial.

Since `currentCount` varies between runs (different `numGroups` values), the fix retains the field-level `int[]` and a new field-level `byte[]`, growing them only when the next run requires a larger buffer.

### Problem 2: per-call DataInputStream wrapping

[Line 98](https://github.com/apache/parquet-java/blob/4c8f4d4b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java#L98) creates a `new DataInputStream(in)` on every PACKED-mode call:

```java
new DataInputStream(in).readFully(bytes, 0, bytesToRead);
```

This allocates a `DataInputStream` wrapper object per call just to access `readFully()`. A private `readFully()` method on the decoder itself eliminates this allocation and the virtual dispatch through the wrapper.

### Component(s)

Core

arouel:
### Rationale for this change

`RunLengthBitPackingHybridDecoder.readNext()` allocates a new `int[]` and `byte[]` on every PACKED-mode call. The existing code even acknowledges this with a `// TODO: reuse a buffer` comment at line 94. In workloads that decode many bit-packed runs (definition levels, repetition levels, RLE-encoded integers), these allocations become a significant source of GC pressure.

There are two issues:

1. **Per-call buffer allocation.** `currentBuffer = new int[currentCount]` and `byte[] bytes = new byte[numGroups * bitWidth]` are allocated fresh on every PACKED-mode `readNext()`. The individual allocations are modest (8–128 ints per run) but occur thousands of times per column chunk. In a custom JFR-profiled benchmark merging 60 Parquet files (180M rows, 10 columns), these two sites were the **number 2 and number 6 allocation hotspots** respectively: 2,402 out of 12,711 total allocation samples (18.9%), accounting for ~7.2 GB of allocation per operation.

2. **Per-call `DataInputStream` wrapping.** `new DataInputStream(in).readFully(bytes, 0, bytesToRead)` creates a wrapper object on every PACKED-mode call just to access `readFully()`.

Three changes to `RunLengthBitPackingHybridDecoder`:

- **`currentBuffer` promoted from local to field**, reused across `readNext()` calls with a grow-only strategy, only reallocated when the next run requires a larger buffer than currently held.
- **`byte[] bytes` promoted to a field `packedBytes`**, same grow-only reuse strategy.
- **`new DataInputStream(in).readFully(...)` replaced with a private `readFully()` method** that reads directly from the underlying `InputStream`, eliminating the per-call wrapper allocation and virtual dispatch.

**Custom JFR-profiled benchmark results** (merge of 60 Parquet files, 180M rows, 10 columns, JDK 26, macOS aarch64):

| Metric | Before | After | Change |
|---|---|---|---|
| Throughput | 44.2 s/op | 42.5 s/op | **-3.6%** |
| Allocation rate | 908.9 MB/s | 793.9 MB/s | **-12.7%** |
| Allocation per op | 42.1 GB | 35.4 GB | **-15.8%** |
| RLE decoder alloc samples | 2,402 (18.9%) | 146 (1.2%) | **-93.9%** |
| RLE decoder alloc bytes | ~7,245 MB | ~417 MB | **-94.2%** |

### Are these changes tested?

Yes, changes are tested. We added a regression test in `TestRunLengthBitPackingHybridEncoder` called `testTruncatedPackedRunAfterFullPackedRunDoesNotReuseStaleBytes` to cover a truncated packed-run edge case after buffer reuse. We also ran the full `TestRunLengthBitPackingHybridEncoder` class and `RunLengthBitPackingHybridIntegrationTest` and all tests passed.

No. This is a transparent performance improvement internal to the RLE decoder. Decoded values are identical. No API changes, no configuration changes, no behavioral changes.

Closes #3466

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
@@ -18,9 +18,9 @@
  */
 package org.apache.parquet.column.values.rle;
 
-import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.bitpacking.BytePacker;
@@ -48,6 +48,8 @@ public class RunLengthBitPackingHybridDecoder {
   private int currentCount;
   private int currentValue;
   private int[] currentBuffer;
+  private int currentBufferLength;
+  private byte[] packedBytes;
 
   public RunLengthBitPackingHybridDecoder(int bitWidth, InputStream in) {
     LOG.debug("decoding bitWidth {}", bitWidth);
@@ -69,7 +71,7 @@ public class RunLengthBitPackingHybridDecoder {
         result = currentValue;
         break;
       case PACKED:
-        result = currentBuffer[currentBuffer.length - 1 - currentCount];
+        result = currentBuffer[currentBufferLength - 1 - currentCount];
         break;
       default:
         throw new ParquetDecodingException("not a valid mode " + mode);
@@ -90,17 +92,23 @@ public class RunLengthBitPackingHybridDecoder {
       case PACKED:
         int numGroups = header >>> 1;
         currentCount = numGroups * 8;
+        currentBufferLength = currentCount;
         LOG.debug("reading {} values BIT PACKED", currentCount);
-        currentBuffer = new int[currentCount]; // TODO: reuse a buffer
-        byte[] bytes = new byte[numGroups * bitWidth];
-        // At the end of the file RLE data though, there might not be that many bytes left.
-        int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
-        bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(bytes, 0, bytesToRead);
+        if (currentBuffer == null || currentBuffer.length < currentCount) {
+          currentBuffer = new int[currentCount];
+        }
+        int bytesNeeded = numGroups * bitWidth;
+        if (packedBytes == null || packedBytes.length < bytesNeeded) {
+          packedBytes = new byte[bytesNeeded];
+        }
+        int bytesRead = in.readNBytes(packedBytes, 0, bytesNeeded);
+        if (bytesRead < bytesNeeded) {
+          Arrays.fill(packedBytes, bytesRead, bytesNeeded, (byte) 0);
+        }
         for (int valueIndex = 0, byteIndex = 0;
             valueIndex < currentCount;
             valueIndex += 8, byteIndex += bitWidth) {
-          packer.unpack8Values(bytes, byteIndex, currentBuffer, valueIndex);
+          packer.unpack8Values(packedBytes, byteIndex, currentBuffer, valueIndex);
         }
         break;
       default:
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
@@ -96,10 +96,14 @@ public class RunLengthBitPackingHybridDecoder {
         LOG.debug("reading {} values BIT PACKED", currentCount);
         if (currentBuffer == null || currentBuffer.length < currentCount) {
           currentBuffer = new int[currentCount];
+        } else {
+          Arrays.fill(currentBuffer, currentCount, currentBuffer.length, 0);
         }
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -90,17 +92,23 @@ private void readNext() throws IOException {
...
-        bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(bytes, 0, bytesToRead);
+        if (currentBuffer == null || currentBuffer.length < currentCount) {
+          currentBuffer = new int[currentCount];
+        }
+        int bytesNeeded = numGroups * bitWidth;
+        if (packedBytes == null || packedBytes.length < bytesNeeded) {
+          packedBytes = new byte[bytesNeeded];
+        }
+        int bytesRead = in.readNBytes(packedBytes, 0, bytesNeeded);
```
Fokko, line 108: Nice, now we're at Java 11 :)

parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -90,17 +92,23 @@ private void readNext() throws IOException {
...
         LOG.debug("reading {} values BIT PACKED", currentCount);
-        currentBuffer = new int[currentCount]; // TODO: reuse a buffer
-        byte[] bytes = new byte[numGroups * bitWidth];
-        // At the end of the file RLE data though, there might not be that many bytes left.
-        int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
-        bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(bytes, 0, bytesToRead);
+        if (currentBuffer == null || currentBuffer.length < currentCount) {
+          currentBuffer = new int[currentCount];
+        }
```
Fokko, line 101:
```suggestion
        } else {
            Arrays.fill(currentBuffer, currentCount, currentBuffer.length, 0);
        }
```

parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -90,17 +92,23 @@ private void readNext() throws IOException {
...
+        int bytesNeeded = numGroups * bitWidth;
+        if (packedBytes == null || packedBytes.length < bytesNeeded) {
+          packedBytes = new byte[bytesNeeded];
+        }
```
Fokko, line 107:
```suggestion
        } else {
          Arrays.fill(packedBytes, bytesNeeded, packedBytes.length, (byte) 0);
        }
```

Fokko (commented):
I've left one fix that addresses a concern is that `currentBuffer` retains stale values from a previous larger packed run in positions beyond `currentCount`. While those positions are never accessed today (the indexing is bounded by `currentBufferLength`), it's a latent risk. Let's add defensive zeroing, consistent with how packedBytes is zeroed on short reads.

arouel:
> I've left one fix that addresses a concern is that `currentBuffer` retains stale values from a previous larger packed run in positions beyond `currentCount`. While those positions are never accessed today (the indexing is bounded by `currentBufferLength`), it's a latent risk. Let's add defensive zeroing, consistent with how packedBytes is zeroed on short reads.

Thank you for the review. I applied the code changes as suggested.

Fokko:
@arouel Could you pull in latest main?

PR #3470, opened 2026-04-08
[Parquet]: GH-563: Make `path_in_schema` optional

etseidl:
### Rationale for this change
Proof of concept for https://github.com/apache/parquet-format/pull/564

Makes changes to `ParquetMetadataConverter` to allow for a missing `path_in_schema` in the Parquet footer. Also adds a writer option controlling whether or not to write the field. The option defaults to `true` to avoid breaking old readers that expect the field to be present.

### Are these changes tested?
Yes, a test has been added, but likely more will be needed.

Adds a new configuration option.


divjotarora:
@etseidl What are the plans for this PR? Can we merge it as-is? If not, can we split out the reader change into a separate PR as it's an orthogonal improvement to stop using `path_in_schema` on the read side?

etseidl:
> @etseidl What are the plans for this PR? Can we merge it as-is? If not, can we split out the reader change into a separate PR as it's an orthogonal improvement to stop using `path_in_schema` on the read side?

@divjotarora this PR is in limbo until the parquet versioning debate is settled. The issue with making only the read side change is that the thrift generated parser will error out until the IDL can be changed to make `path_in_schema` optional. I don't think there's an appetite here to make local changes to parquet.thrift. If I'm wrong, I'm happy to split out the read and write side changes.

PR #3472, opened 2026-04-11
GH-3471: Fix `ByteBuffer` handling in `VariantUtil` and `VariantBuilder`

### LINKED ISSUE #3471: VariantUtil.getMetadataMap reads from wrong position for non-array ByteBuffer; VariantBuilder.appendBinary mutates caller's ByteBuffer

Description:
  
- `getMetadataMap` incorrect position

  In `VariantUtil.getMetadataMap`, the else-branch (read-only or non-array-backed
  `ByteBuffer`) computes the slice position incorrectly:

  ```
  slice(metadata, stringStart + offset)
  ```

  `stringStart` is relative to `pos` (`metadata.position()`), but the array-backed
  branch correctly adds `pos`:

  ```
  metadata.arrayOffset() + pos + stringStart + offset
  ```

  This causes incorrect dictionary parsing or exceptions when the metadata
  `ByteBuffer` has a non-zero position (e.g., when it is a slice of a larger buffer).

  The sibling method `getMetadataKey` does not have this bug because `dataPos` is
  computed as an absolute position.

- `appendBinary` mutates caller's `ByteBuffer`

  `VariantBuilder.appendBinary(ByteBuffer)` calls `.put(binary)` which advances the
  caller's `ByteBuffer` position. The analogous method `appendEncodedValue` uses
  `value.duplicate().get(...)` to avoid this side effect.

### Component(s)

Core

rayokota:
### Rationale for this change

Fix `ByteBuffer` handling in `VariantUtil` and `VariantBuilder`.

- Fix `VariantUtil.getMetadataMap` reading string bytes from the wrong position
  when the metadata `ByteBuffer` has a non-zero `position()`. Changed
  `slice(metadata, stringStart + offset)` to
  `slice(metadata, pos + stringStart + offset`) to match the array-backed branch.
- Fix `VariantBuilder.appendBinary` mutating the caller's `ByteBuffer` by using
  `binary.duplicate()`, consistent with` appendEncodedValue`.

### Are these changes tested?

- Added `TestVariantObject.testMetadataWithNonZeroPositionReadOnly`: constructs a
  metadata `ByteBuffer` with a non-zero position and read-only flag, then verifies
 ` ImmutableMetadata` (which calls `getMetadataMap`) correctly parses the
  dictionary. Fails without the fix.
- Added `TestVariantScalarBuilder.testBinaryBuilderDoesNotMutateCallerBuffer`:
  verifies that appendBinary does not change the caller's` ByteBuffer` position
  or remaining count.

- No

Closes #3471

### DIFF
--- a/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
+++ b/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
@@ -341,4 +341,39 @@ public class TestVariantObject {
       Assert.assertEquals("Cannot read ARRAY value as OBJECT", e.getMessage());
     }
   }
+
+  @Test
+  public void testMetadataWithNonZeroPositionReadOnly() {
+    // Build a variant with object fields to populate the metadata dictionary
+    VariantBuilder vb = new VariantBuilder();
+    VariantObjectBuilder obj = vb.startObject();
+    obj.appendKey("name");
+    obj.appendString("Alice");
+    obj.appendKey("age");
+    obj.appendInt(30);
+    vb.endObject();
+    Variant variant = vb.build();
+
+    // Get the raw metadata bytes
+    ByteBuffer metaBuf = variant.getMetadataBuffer();
+    byte[] metaBytes = new byte[metaBuf.remaining()];
+    metaBuf.duplicate().get(metaBytes);
+
+    // Embed in a larger buffer with a non-zero position, then make read-only
+    // to force the else-branch in getMetadataMap.
+    byte[] padded = new byte[10 + metaBytes.length];
+    System.arraycopy(metaBytes, 0, padded, 10, metaBytes.length);
+    ByteBuffer offsetMetadata = ByteBuffer.wrap(padded);
+    offsetMetadata.position(10);
+    offsetMetadata.limit(10 + metaBytes.length);
+    offsetMetadata = offsetMetadata.asReadOnlyBuffer();
+
+    // ImmutableMetadata calls getMetadataMap, which had the bug.
+    // getMetadataMap builds a key->id dictionary from the metadata buffer.
+    // With a non-zero position and read-only buffer, the else-branch is taken,
+    // which previously used the wrong offset.
+    ImmutableMetadata immutableMetadata = new ImmutableMetadata(offsetMetadata);
+    Assert.assertEquals(0, immutableMetadata.getOrInsert("name"));
+    Assert.assertEquals(1, immutableMetadata.getOrInsert("age"));
+  }
 }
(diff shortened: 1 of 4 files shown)


parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantScalarBuilder.java
```diff
@@ -387,6 +387,17 @@ public void testBinaryBuilder() {
...
   }
 
+  @Test
+  public void testBinaryBuilderDoesNotMutateCallerBuffer() {
+    ByteBuffer buf = ByteBuffer.wrap(new byte[] {0, 1, 2, 3});
+    int positionBefore = buf.position();
+    int remainingBefore = buf.remaining();
+    VariantBuilder vb = new VariantBuilder();
+    vb.appendBinary(buf);
+    Assert.assertEquals(positionBefore, buf.position());
```
steveloughran, line 397: nit add a "position" and "remaining" description for better diagnostics on CI runs

nssalian:
CC: @alamb to help kick off test workflows.

steveloughran (approved):
this is good. does relate to the work in #3481 which needs to pick up the same fix.
getting your patch in first adds the regression tests we need

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObject.java
```diff
@@ -341,4 +341,39 @@ public void testInvalidObject() {
...
+    ByteBuffer offsetMetadata = ByteBuffer.wrap(padded);
+    offsetMetadata.position(10);
+    offsetMetadata.limit(10 + metaBytes.length);
+    offsetMetadata = offsetMetadata.asReadOnlyBuffer();
+
+    // ImmutableMetadata calls getMetadataMap, which had the bug.
+    // getMetadataMap builds a key->id dictionary from the metadata buffer.
+    // With a non-zero position and read-only buffer, the else-branch is taken,
+    // which previously used the wrong offset.
+    ImmutableMetadata immutableMetadata = new ImmutableMetadata(offsetMetadata);
```
nssalian, line 375: nit: The test comment reads more like a bug description than a test description. Consider reframing to describe what's being verified, e.g. "Verify metadata dictionary parsing works correctly when the ByteBuffer has a non-zero position and is read-only."

nssalian (commented):
tests pass and #3481 can follow this PR like @steveloughran mentioned. 
Left a single comment.

Fokko (approved):
Let's get this in to unblock https://github.com/apache/parquet-java/pull/3481

Thanks @rayokota for adding this and thanks @nssalian and @steveloughran for the review 👍

Issue #3475, opened 2026-04-13
Parquet Java vectorized module not working with newer JDKs


iemejia:
### Describe the bug, including details regarding any error messages, version, and platform.

I was building the project with Java 25 (the most recent LTS) and noticed that there was a call to a method of the Vector API that was removed after JDK 17 so it fails at with NoSuchMethodError.

iemejia:
Fixed in https://github.com/apache/parquet-java/pull/3476

PR #3476, opened 2026-04-13
GH-3475: Fix parquet-vector compatiblity with Java > 17

iemejia:
Replace the ByteBuffer-specific vector loads with local helpers that copy the required bytes and then call ByteVector.fromArray. This removes the dependency on JDK-specific ByteVector.fromByteBuffer entry points, which can fail with NoSuchMethodError on newer runtimes.

Assisted-by: OpenCode:gpt-5.4


.gitignore
```diff
@@ -20,4 +21,3 @@ target/
 mvn_install.log
 .vscode/*
 .DS_Store
-
```
wgtmac, line 23: Let's keep this blank line

parquet-plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
```diff
@@ -3175,13 +3171,30 @@ public final void unpackValuesUsingVector(final byte[] in, final int inPos, fina
...
   }
 
+  private static ByteVector fromByteBuffer(VectorSpecies<Byte> species, ByteBuffer input, int inPos) {
+    return ByteVector.fromArray(species, readInputBytes(input, inPos, species.length()), 0);
+  }
+
+  private static ByteVector fromByteBuffer(
+      VectorSpecies<Byte> species, ByteBuffer input, int inPos, VectorMask<Byte> mask) {
+    return ByteVector.fromArray(species, readInputBytes(input, inPos, mask.trueCount()), 0, mask);
+  }
```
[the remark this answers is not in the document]
iemejia: Good catch — this was wrong. `mask.trueCount()` only happens to be safe at the call sites in this file because every mask is built as a contiguous-prefix mask, but `ByteVector.fromArray(species, array, 0, mask)` requires `array.length >= species.length()` for any mask shape, and the masked overload of OpenJDK 17's `ByteVector.fromByteBuffer` actually loads the full window unconditionally and then `blend`s with the mask — not a prefix copy. The fix in dffb6e6f6 is exactly that: always allocate `species.length()` bytes and let `fromArray(..., mask)` apply the mask, matching JDK semantics for arbitrary masks. Verified the AVX-512 unpack tests still pass on JDK 17 / 21 / 25.

parquet-plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
```diff
@@ -3175,13 +3171,30 @@ public final void unpackValuesUsingVector(final byte[] in, final int inPos, fina
...
+
+  private static byte[] readInputBytes(ByteBuffer input, int inPos, int byteCount) {
+    byte[] bytes = new byte[byteCount];
+    ByteBuffer source = input.duplicate();
+    source.position(inPos);
+    source.get(bytes);
```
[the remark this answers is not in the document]
iemejia: Agreed on the regression concern, and the heap-buffer fast path is in dffb6e6f6 — `input.hasArray()` returns `ByteVector.fromArray(species, input.array(), input.arrayOffset() + inPos[, mask])` directly, which HotSpot intrinsifies to the same `vmovdqu64` AVX-512 load as the original `fromByteBuffer`. Since `BytePacker` call sites in parquet-mr use `ByteBuffer.wrap(byte[])` (heap), this fast path covers the dominant production case with no copy and no `duplicate()`.

I'd like to push back on the `ThreadLocal<byte[]>` part:

1. It only eliminates the `byte[]` allocation, not the `ByteBuffer.duplicate()`, which is the larger of the two costs on the direct-buffer path. To eliminate both you'd need a thread-local `ByteBuffer` view as well, at which point the helper is doing more bookkeeping than the load itself.
2. The `byte[]` is 8–64 bytes, TLAB-allocated, never escapes the helper, and is a strong escape-analysis candidate — HotSpot will frequently scalar-replace it on hot paths. `ThreadLocal.get()` (~5 ns) never scalarizes, so on the common JIT-hot case the `ThreadLocal` is a net loss.
3. `ThreadLocal` of a mutable `byte[]` carries the usual operational risks in long-lived JVMs (thread-pool retention, classloader leak surface) for code that is explicitly throwaway: dffb6e6f6 adds a TODO marking `ByteVector.fromMemorySegment` (JEP 454, JDK 22+) as the proper replacement, which removes the copy entirely for both heap and direct buffers via `ScopedMemoryAccess` and is intrinsifiable. Once the project's minimum JDK reaches 22, both helpers and `readInputBytes` go away.

So the change keeps the direct-buffer path as a simple `byte[] + duplicate()` fallback (matching the previous PR's behavior on that path), adds the heap fast path you suggested for the dominant case, and defers further optimization to the `fromMemorySegment` migration tracked in the TODO.

.github/workflows/vector-plugins.yml
```diff
@@ -46,12 +46,19 @@ jobs:
         run: |
           EXTRA_JAVA_TEST_ARGS=$(./mvnw help:evaluate -Dexpression=extraJavaTestArgs -q -DforceStdout)
           export MAVEN_OPTS="$MAVEN_OPTS $EXTRA_JAVA_TEST_ARGS"
-          ./mvnw install --batch-mode -Pvector-plugins -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true -Dmaven.buildNumber.skip=true -Djava.version=${{ matrix.java }} -pl parquet-plugins/parquet-encoding-vector,parquet-plugins/parquet-plugins-benchmarks -am
+          ./mvnw install --batch-mode -Pvector-plugins -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true -Dmaven.buildNumber.skip=true -Dspotless.check.skip=true -Djava.version=${{ matrix.java }} -pl parquet-plugins/parquet-encoding-vector,parquet-plugins/parquet-plugins-benchmarks -am
```
wgtmac, line 49: Is it possible to use similar pattern like `SPOTLESS_ARGS` below?

parquet-plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
```diff
@@ -3175,13 +3171,56 @@ public final void unpackValuesUsingVector(final byte[] in, final int inPos, fina
...
+    }
+    // Off-heap (direct) buffers: fall back to copying species.length() bytes into a scratch
+    // array. The original JDK 17 fast path used ScopedMemoryAccess.loadFromByteBuffer to avoid
+    // the copy, but that intrinsic is not exposed; the alternative ByteVector.fromMemorySegment
+    // is only available on JDK 19+ and cannot be called from --release 17 sources.
+    return ByteVector.fromArray(species, readInputBytes(input, inPos, species.length()), 0);
+  }
+
+  private static ByteVector fromByteBuffer(
+      VectorSpecies<Byte> species, ByteBuffer input, int inPos, VectorMask<Byte> mask) {
```
wgtmac, line 3204: I'm not an expert on this, but Codex complains as below:

The new fromByteBuffer(species, input, inPos, mask) helper changes the contract of the old JDK 17 masked load. Per the JDK 17 ByteVector.fromByteBuffer(..., m) docs, only masked-on lanes must be in bounds, but this fallback always copies species.length() bytes for direct buffers before applying the mask. That breaks packers whose mask intentionally covers fewer bytes than the vector width, such as Packer3 (B128 with a 12-byte mask): a direct ByteBuffer with exactly the required packed bytes now throws BufferUnderflowException, even though the old implementation and the array-backed path both accept it. The fix needs to preserve masked-load bounds semantics instead of unconditionally reading the whole vector window.

.gitignore
iemejia: Done, restored in b5f6cb9fb.

parquet-plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
```diff
@@ -3175,13 +3171,56 @@ public final void unpackValuesUsingVector(final byte[] in, final int inPos, fina
...
+    // Mirror the fast path of OpenJDK 17's ByteVector.fromByteBuffer(species, bb, offset, bo, m):
+    // when the full vector window fits in the buffer, read the entire window unconditionally and
+    // let fromArray(species, array, 0, mask) apply the mask. This matches the JDK semantics for
+    // arbitrary mask shapes (not just contiguous-prefix masks) and keeps array.length equal to
+    // species.length(), which satisfies the bounds-check precondition of fromArray.
+    if (input.hasArray()) {
+      return ByteVector.fromArray(species, input.array(), input.arrayOffset() + inPos, mask);
+    }
+    return ByteVector.fromArray(species, readInputBytes(input, inPos, species.length()), 0, mask);
+  }
```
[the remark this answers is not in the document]
iemejia: Good suggestion. Added `unpackValuesUsingVectorDirectByteBuffer` and `unpackValuesUsingVectorReadOnlyByteBuffer` tests in b5f6cb9fb. They exercise `ByteBuffer.allocateDirect()` and `.asReadOnlyBuffer()` respectively, covering all 32 bit widths.

.github/workflows/vector-plugins.yml
iemejia: Good idea — refactored in b5f6cb9fb. The install step now computes `SPOTLESS_ARGS` the same way as the verify step (skip on non-17 JDKs, run on 17).

parquet-plugins/parquet-encoding-vector/src/main/java/org/apache/parquet/column/values/bitpacking/ByteBitPacking512VectorLE.java
iemejia: Valid concern. Fixed in b5f6cb9fb: the masked direct-buffer path now allocates a `species.length()`-sized array (satisfying `fromArray`'s bounds precondition) but only copies `mask.trueCount()` bytes — preserving JDK 17 semantics where only masked-on lanes must be in bounds. This avoids the `BufferUnderflowException` for packers like Packer3 whose mask covers fewer bytes than the vector width.

iemejia:
@wgtmac @Fokko All review comments have been addressed in b5f6cb9fb — ready for another look when you get a chance.

Fokko:
Thanks @iemejia, eager to get this in. There are some Spotless issues, can you check?

iemejia:
Thanks @Fokko -- fixed in 0574eeae7. The lambda body indentation in `getRangeData()` was off. `spotless:check` and all vector tests pass on JDK 17.

Fokko (approved):

Fokko:
Thanks @iemejia for fixing this 👍 and thanks @wgtmac for the review

PR #3481, opened 2026-04-15
Optimizing Variant read path with lazy caching

nssalian:
### Rationale for this change

Profiling in [3452](https://github.com/apache/parquet-java/pull/3452) identified `Variant.getFieldAtIndex()` and metadata string lookups as hotspots during variant reads. Every call to `getFieldByKey`, `getFieldAtIndex`, and `getElementAtIndex` re-parses headers and re-allocates objects that could be cached.

- Adds lazy caching to `Variant.java` for metadata strings, object headers, and array headers. Field lookups in `getFieldByKey` now defer value construction until a match is found, and child Variants share the parent's metadata cache. Also removes two unused static helper methods. 

- Adds `readUnsignedLittleEndian` to VariantUtil for bulk ByteBuffer reads (adapted from Iceberg), replacing byte-at-a-time readUnsigned calls in the Variant read path. Buffers are set to LITTLE_ENDIAN byte order in the constructor to support this.
                                                                                                                                               
- Concurrency javadoc added; `getMetadataKeyCached` captures all state in locals to ensure thread safety with at worst redundant decoding.

Includes @steveloughran's string converter optimization from #3452: `VariantBuilder.appendAsString(Binary)` and its use in `VariantConverters`. 

### Are these changes tested?

Ran the benchmarks from [3452](https://github.com/apache/parquet-java/pull/3452) locally
`~3.2x variant deserialization speedup`

Command run:
```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar   "VariantBuilderBenchmark.deserializeVariant" -f 1 -wi 3 -i 500 -t 1         
```
Baseline

- Master branch with the above Benchmark code to get the current numbers
```
Benchmark                                   (depth)  (fieldCount)  Mode  Cnt      Score     Error  Units
VariantBuilderBenchmark.deserializeVariant     Flat           200    ss  500   8875.229 ± 116.528  us/op
VariantBuilderBenchmark.deserializeVariant   Nested           200    ss  500  11728.253 ± 107.884  us/op        
```

This PR changes:

```
Benchmark                                   (depth)  (fieldCount)  Mode  Cnt     Score     Error  Units
VariantBuilderBenchmark.deserializeVariant     Flat           200    ss  500  2765.264 ±  97.714  us/op
VariantBuilderBenchmark.deserializeVariant   Nested           200    ss  500  3802.665 ± 102.249  us/op
```

No.


parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -206,22 +246,18 @@ public int numObjectElements() {
...
-            value,
-            metadata,
-            info.idSize,
-            info.offsetSize,
-            value.position() + info.idStartOffset,
-            value.position() + info.offsetStartOffset,
-            value.position() + info.dataStartOffset);
-        if (field.key.equals(key)) {
-          return field.value;
+        int id = VariantUtil.readUnsigned(value, idStart + info.idSize * i, info.idSize);
```
steveloughran, line 256: readUnsigned will need looking at too; the iceberg code looks better. Later PR

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -67,6 +87,26 @@ public Variant(ByteBuffer value, ByteBuffer metadata) {
...
           metadata.get(metadata.position()) & VariantUtil.VERSION_MASK));
     }
+
+    // Pre-compute dictionary size for lazy metadata cache allocation.
+    int pos = this.metadata.position();
+    int metaOffsetSize = ((this.metadata.get(pos) >> 6) & 0x3) + 1;
+    if (this.metadata.remaining() > 1) {
+      this.dictSize = VariantUtil.readUnsigned(this.metadata, pos + 1, metaOffsetSize);
+    } else {
+      this.dictSize = 0;
```
steveloughran, line 97: I didn't benchmark empty variants, did I? Not a case we need to worry much about though

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -324,23 +333,56 @@ public int numArrayElements() {
...
-  private static Variant getElementAtIndex(
-      int index, ByteBuffer value, ByteBuffer metadata, int offsetSize, int offsetStart, int dataStart) {
-    // offsetStart and dataStart are absolute positions in the `value` buffer.
-    int offset = VariantUtil.readUnsigned(value, offsetStart + offsetSize * index, offsetSize);
-    return new Variant(VariantUtil.slice(value, dataStart + offset), metadata);
+  /**
+   * Returns the metadata dictionary string for the given ID, caching the result.
+   */
+  String getMetadataKeyCached(int id) {
+    if (metadataCache == null) {
```
steveloughran: can you add some comments here, e.g

// demand create shared dictionary cache

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -324,23 +333,56 @@ public int numArrayElements() {
...
+      metadataCache = new String[dictSize];
+    }
+    if (id < 0 || id >= dictSize) {
```
steveloughran, line 358: definitely explain this.

steveloughran (commented):
code looks good; made some minor changes.

This should make a very big difference when selectively retrieving multiple fields within a single variant, or within a variant and nested children.

I do worry about concurrency now. The existing Variant didn't have issues here precisely because it recalculated everything.

We have to be confident that even if concurrent access triggers a duplicate cache operation, there's no harm in this. Otherwise cache access will have to be synchronized.

It all looks good to me.

alamb:
I started the workflows

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: Agreed, will look at readUnsigned optimization as a follow-up.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: cautiously added here in case there might be empty metadata. I doubt there's a perf concern here.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: Will do

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: Will add a comment line

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -36,6 +36,26 @@ public final class Variant {
    */
   final ByteBuffer metadata;
 
+  /**
+   * Pre-computed metadata dictionary size
```
steveloughran, line 40: nit. add a "." at the end as some java versions' javadocs really blow up if you don't.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -324,23 +333,64 @@ public int numArrayElements() {
...
+    if (id < 0 || id >= dictSize) {
+      return VariantUtil.getMetadataKey(metadata, id);
+    }
+    // Demand-create shared dictionary cache
+    String[] cache = metadataCache;
+    if (cache == null) {
+      cache = new String[dictSize];
+      metadataCache = cache;
+    }
+    if (cache[id] == null) {
```
steveloughran, line 367: set cache to be metadataCache here, so if there is any race condition the same cache array is updated. If two threads are writing to the same [id], well, one lookup is wasted. The joint cache is still (probably) updated with each value

steveloughran (commented):
reviewing the code again to see if it's possible to get back some of those performance numbers lost with the move to `volatile`.

We're only reading and caching data, so there's no real write conflicts -is the use of volatile everywhere being over-cautious? it's forcing memory reads everywhere. 

And I don't know how common cross-thread reading will actually be in production systems; in spark each worker is its own thread, after all.

Maybe the goal should just be all reviewers being confident that if there are dual writers, the output will always be consistent.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: Fixed this. Now it re-reads metadataCache after assignment so concurrent threads converge on the same array. 
Also, removed the volatile fields. I ran some local concurrent tests and with multiple iterations and confirmed. I don't see any concurrency tests in the codebase but happy to add it to the test suite if it helps build confidence in the approach here.

steveloughran:
I would propose a javadoc on concurrency here.

Till now we've had a mutable builder, the caching makes it mutable. But even in a race condition, as the only change whjch ever takes place is a decode an update of the dictionary, if the dictionary was safe then the worst outcome is one of the lazy evals gets lost.

Lets
-make sure that the thread doing the lazy eval retains the values it needs for the duration of get()
-look at the dictionary impl.

Java HashMap is not thread safe, the very old HashTable is, but it may have its own penalty when used.

That rust variant puts a lot of effort into memory efficiency too. Maybe we should make sure that these changes don't completely explode memory consumption. I know, it's a tradeoff (speed, space, code complexity). And queries like speed. And I think the focus should be "single thread speed and no inconsistency on multithreaded use" as single thread workers is what the query engines do. After all, they shouldn't expect the input streams to be thread-safe, should they? Two threads doing parallel reads of a stream is already making some big assumptions about the underlying layers (*)


(*) hadoop input streams are thread safe precisely because code makes those assumptions, FWIW. Going thread unsafe broke hbase

steveloughran:
@Fokko could you approve the test runs here? thanks

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -232,22 +268,16 @@ public Variant getFieldByKey(String key) {
...
-            mid,
-            value,
-            metadata,
-            info.idSize,
-            info.offsetSize,
-            value.position() + info.idStartOffset,
-            value.position() + info.offsetStartOffset,
-            value.position() + info.dataStartOffset);
-        int cmp = field.key.compareTo(key);
+        int midId = VariantUtil.readUnsigned(value, idStart + info.idSize * mid, info.idSize);
```
steveloughran: see #3529 ; this search only works if the fields are sorted. The `sorted` flag in the header declares whether or not it is. As you are editing the code, might be good to fix

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -299,6 +299,31 @@ static int readUnsigned(ByteBuffer bytes, int pos, int numBytes) {
...
+   * Adapted from Apache Iceberg's VariantUtil.readLittleEndianUnsigned.
+   */
+  static int readUnsignedLittleEndian(ByteBuffer buffer, int pos, int numBytes) {
+    switch (numBytes) {
+      case 1:
+        return buffer.get(pos) & U8_MAX;
+      case 2:
+        return buffer.getShort(pos) & U16_MAX;
+      case 3:
+        return (buffer.getShort(pos) & U16_MAX) | ((buffer.get(pos + 2) & U8_MAX) << 16);
```
steveloughran, line 315: you know, I wonder if you could something radical here like getInt(pos) & 0xfffff, though I suspect signing will ruin everything because java lacks uints, and you need to handle end of buffer situations.
it's just that world aligned LOAD is going to be so, so fast. But then again, if you look into the java code to see handling of unaligned code you end up not knowing what the outcome would be

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -232,22 +272,17 @@ public Variant getFieldByKey(String key) {
...
-            mid,
-            value,
-            metadata,
-            info.idSize,
-            info.offsetSize,
-            value.position() + info.idStartOffset,
-            value.position() + info.offsetStartOffset,
-            value.position() + info.dataStartOffset);
-        int cmp = field.key.compareTo(key);
+        int midId = VariantUtil.readUnsignedLittleEndian(value, idStart + info.idSize * mid, info.idSize);
```
steveloughran, line 283: only works for variants which arrive sorted.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
nssalian: This might be worth fixing in a separate PR - get that fix in prior to the this current PR.

nssalian:
@alamb @emkornfield PTAL

steveloughran:
@nssalian now the variant benchmark is in, can you merge the parquet master branch in to this PR so I can run the benchmarks over it and compare to master itself? thanks

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -56,17 +81,35 @@ public Variant(byte[] value, int valuePos, int valueLength, byte[] metadata, int
...
       throw new UnsupportedOperationException(String.format(
           "Unsupported variant metadata version: %d",
           metadata.get(metadata.position()) & VariantUtil.VERSION_MASK));
     }
+
+    // Pre-compute dictionary size for lazy metadata cache allocation.
+    int pos = this.metadata.position();
+    int metaOffsetSize = ((this.metadata.get(pos) >> 6) & 0x3) + 1;
+    if (this.metadata.remaining() > 1) {
+      this.dictSize = VariantUtil.readUnsignedLittleEndian(this.metadata, pos + 1, metaOffsetSize);
```
steveloughran, line 102: see #3562 ; dictSize must be <= variant metadata size

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -299,6 +299,31 @@ static int readUnsigned(ByteBuffer bytes, int pos, int numBytes) {
...
+      case 4:
+        int v = buffer.getInt(pos);
+        if (v < 0) {
```
steveloughran: maybe use `org.apache.parquet.Preconditions.checkArgument()` here

nssalian:
Thanks @steveloughran. I'll follow up with the changes.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
steveloughran: some tests do create empty variants, it is needed

steveloughran (approved):
this is good, does a lot of the hardening.

+1 (non binding) from me

Fokko (approved):

Fokko:
Thanks @nssalian for working on this, and thanks @steveloughran for the review 👍

steveloughran:
nice!

steveloughran:
Rebased #3562 onto this; resilience tests happy

PR #3488, opened 2026-04-18
GH-3487: Fix `PageReadStore` buffer leak in `ParquetFileReader`

### LINKED ISSUE #3487: `ParquetFileReader` does not close previous `PageReadStore` on `readNextRowGroup()` or on `close()`

`ParquetFileReader` stores a reference to the last returned `ColumnChunkPageReadStore` in `currentRowGroup`, but never closes it:
1. `readNextRowGroup()` (line 1153) overwrites `this.currentRowGroup = rowGroup` without closing the previous instance.
2. `readNextFilteredRowGroup()` (line 1409) does the same.
3. `close()` (line 1816-1827) does not close `currentRowGroup` at all, it only closes the input stream, dictionary reader, and codec factory.

`ColumnChunkPageReadStore.close()` releases the `ByteBufferReleaser` that holds both the compressed file I/O buffers (from `ConsecutivePartList.readAll()`) and any off-heap decompressed page buffers (from the `useOffHeapDecryptBuffer` path). Since `close()` is never called, these buffers are never released.
With the default `HeapByteBufferAllocator` this is masked by GC because `HeapByteBufferAllocator.release()` is a no-op. With a direct `ByteBufferAllocator`, this becomes a hard native memory leak that grows with every row group read.
Note that `InternalParquetRecordReader` works around this by manually calling `currentRowGroup.close()` before each `readNextRowGroup()` (line 134-135) and in its own `close()` (line 164-167). Other direct callers of `ParquetFileReader` (e.g., `ParquetRewriter`, `ColumnIndexValidator`) also close the `PageReadStore` themselves. However, any caller that does not manually close the returned `PageReadStore` will leak buffers.

### Expected behavior

`ParquetFileReader` should close the previous `currentRowGroup` before assigning a new one in `readNextRowGroup()` / `readNextFilteredRowGroup()`, and close the final `currentRowGroup` in its own `close()` method. This matches the lifecycle that `InternalParquetRecordReader` implements manually.

### Error messages

No error is thrown. The buffers silently leak. With a direct allocator, the native memory grows unboundedly until the process is killed or the allocator is explicitly closed.

### Version

1.17.0 (older versions are also affected)

### Component(s)

Core

arouel:
### Rationale for this change

`ParquetFileReader` never closes the `ColumnChunkPageReadStore` it returns from `readNextRowGroup()`. When a subsequent call replaces `currentRowGroup`, the previous instance's `ByteBufferReleaser` is abandoned without releasing the compressed I/O buffers and any off-heap decompressed page buffers it holds. With the default `HeapByteBufferAllocator` this is masked by GC, but with a direct `ByteBufferAllocator` it becomes a hard native memory leak that grows with every row group read. `InternalParquetRecordReader` works around this by manually closing the `PageReadStore` before each read and in its own `close()`, but any direct caller of `ParquetFileReader` that does not replicate this pattern will leak buffers.

A private `closeCurrentRowGroup()` method is added to `ParquetFileReader` that null-safely closes and nulls the `currentRowGroup` field. It is called in `readNextRowGroup()` and `readNextFilteredRowGroup()` before assigning the new row group, and `currentRowGroup` is included in the `AutoCloseables.uncheckedClose()` chain in `close()`. This brings the buffer lifecycle management into `ParquetFileReader` itself so all callers benefit automatically.

### Are these changes tested?

The existing test suites in parquet-hadoop continue to pass. Additional tests got added to verify that `PageReadStore` buffers are properly released.

No API changes. Callers that already close the `PageReadStore` themselves (like `InternalParquetRecordReader`) will see a harmless double-close since `ColumnChunkPageReadStore.close()` is idempotent via `ByteBufferReleaser`. Callers that did not close the `PageReadStore` will now have their buffers released automatically, reducing memory usage.

Closes #3487

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1150,6 +1150,7 @@ public class ParquetFileReader implements Closeable {
     if (rowGroup == null) {
       return null;
     }
+    closeCurrentRowGroup();
     this.currentRowGroup = rowGroup;
     // avoid re-reading bytes the dictionary reader is used after this call
     if (nextDictionaryReader != null) {
@@ -1406,6 +1407,7 @@ public class ParquetFileReader implements Closeable {
       return readNextRowGroup();
     }
 
+    closeCurrentRowGroup();
     this.currentRowGroup = internalReadFilteredRowGroup(block, rowRanges, getColumnIndexStore(currentBlock));
 
     // avoid re-reading bytes the dictionary reader is used after this call
@@ -1819,8 +1821,16 @@ public class ParquetFileReader implements Closeable {
         f.close();
       }
     } finally {
-      AutoCloseables.uncheckedClose(nextDictionaryReader, crcAllocator);
+      AutoCloseables.uncheckedClose(currentRowGroup, nextDictionaryReader, crcAllocator);
+      currentRowGroup = null;
       options.getCodecFactory().release();
+    }
+  }
+
+  private void closeCurrentRowGroup() {
+    if (currentRowGroup != null) {
+      currentRowGroup.close();
+      currentRowGroup = null;
     }
   }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1147,10 +1147,10 @@ public class ParquetFileReader implements Closeable {
       return readNextRowGroup();
     }
 
+    closeCurrentRowGroup();
     if (rowGroup == null) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1150,6 +1150,7 @@ public PageReadStore readNextRowGroup() throws IOException {
     if (rowGroup == null) {
       return null;
     }
+    closeCurrentRowGroup();
```
Fokko: Shouldn't we call this one just above line 1143 similar to the change below in `readNextFilteredRowGroup`?

arouel: Good catch, this ensures the last row group's buffers are freed as soon as iteration is exhausted, not deferred to `close()`.

Fokko (approved):

PR #3490, opened 2026-04-18
GH-3489: Fix Binary.copy() to materialize direct ByteBuffer-backed values to heap

### LINKED ISSUE #3489: `Binary.copy()` does not materialize direct `ByteBuffer`-backed values to heap, causing use-after-free

`Binary.copy()` is intended to produce an owned copy of the value that is safe to store beyond the lifetime of the source buffer. However, the base implementation (line 120-126) only copies when `isBackingBytesReused` is `true`:
```java
public Binary copy() {
    if (isBackingBytesReused) {
        return Binary.fromConstantByteArray(getBytes());
    } else {
        return this;
    }
}
```
For `ByteBufferBackedBinary` instances produced by column readers, `isBackingBytesReused` is `false`, so `copy()` returns `this`, the original object still pointing into the decompressed page buffer.
When the decompressed page buffer is backed by a direct `ByteBuffer` whose memory can be freed independently (e.g., via `Arena.close()` when using a direct `ByteBufferAllocator` with `useOffHeapDecryptBuffer`), any code that called `copy()` expecting to own the data will later hit a use-after-free when the source buffer is released.

In practice this surfaces through `DictionaryValuesWriter.writeBytes(Binary v)`, which calls `v.copy()` before inserting into its dictionary map. The returned object is the same `ByteBufferBackedBinary` still referencing the direct buffer. When the `PageReadStore` is closed (releasing the decompressed page buffers), the dictionary entries become dangling references. A subsequent dictionary fallback triggers access to the freed memory:
```
java.lang.IllegalStateException: Already closed
    at java.base/jdk.internal.foreign.MemorySessionImpl.alreadyClosed(MemorySessionImpl.java:326)
    at java.base/jdk.internal.misc.ScopedMemoryAccess$ScopedAccessError.newRuntimeException(ScopedMemoryAccess.java:114)
    at java.base/jdk.internal.misc.ScopedMemoryAccess.getByte(ScopedMemoryAccess.java:1090)
    at java.base/java.nio.DirectByteBuffer.get(DirectByteBuffer.java:341)
    at java.base/java.nio.ByteBuffer.getArray(ByteBuffer.java:984)
    at java.base/java.nio.ByteBuffer.get(ByteBuffer.java:838)
    at java.base/java.nio.ByteBuffer.get(ByteBuffer.java:865)
    at org.apache.parquet.io.api.Binary$ByteBufferBackedBinary.getBytes(Binary.java:504)
    at org.apache.parquet.io.api.Binary$ByteBufferBackedBinary.getBytesUnsafe(Binary.java:515)
    at org.apache.parquet.column.values.deltastrings.DeltaByteArrayWriter.writeBytes(DeltaByteArrayWriter.java:92)
    at org.apache.parquet.column.values.dictionary.DictionaryValuesWriter$PlainBinaryDictionaryValuesWriter.fallBackDictionaryEncodedData(DictionaryValuesWriter.java:315)
    at org.apache.parquet.column.values.dictionary.DictionaryValuesWriter.fallBackAllValuesTo(DictionaryValuesWriter.java:151)
    at org.apache.parquet.column.values.fallback.FallbackValuesWriter.fallBack(FallbackValuesWriter.java:155)
    at org.apache.parquet.column.values.fallback.FallbackValuesWriter.getBytes(FallbackValuesWriter.java:86)
    at org.apache.parquet.column.impl.ColumnWriterV2.writePage(ColumnWriterV2.java:99)
    at org.apache.parquet.column.impl.ColumnWriterBase.writePage(ColumnWriterBase.java:379)
    at org.apache.parquet.column.impl.ColumnWriteStoreBase.flush(ColumnWriteStoreBase.java:191)
```
The `isBackingBytesReused` flag was designed for the `RecordReader` pattern where a shared `byte[]` backing array might be mutated between calls. It does not account for direct `ByteBuffer`s whose underlying memory can be deallocated externally.

### Expected behavior

`Binary.copy()` on a `ByteBufferBackedBinary` with a direct `ByteBuffer` should always materialize to a heap-backed `Binary`, ensuring the returned value is independent of the source buffer's lifecycle.

### Version

1.17.0 (older versions are also affected)

### Component(s)

Core

arouel:
### Rationale for this change

`Binary.copy()` is expected to return a value that the caller can safely store beyond the lifetime of the source buffer. The base implementation only copies when `isBackingBytesReused` is `true`, otherwise it returns `this`. For `ByteBufferBackedBinary` instances backed by a direct `ByteBuffer`, `isBackingBytesReused` is `false`, so `copy()` is a no-op. The returned object still points into the original direct buffer. When that buffer's memory is freed (e.g., after closing a `PageReadStore` that used the off-heap decompression path), any code that relied on `copy()` for ownership, such as `DictionaryValuesWriter.writeBytes()`, hits `IllegalStateException: Already closed on the next access`.

`ByteBufferBackedBinary.copy()` is overridden to check `value.isDirect()`. When the backing `ByteBuffer` is direct, it unconditionally materializes to a heap-backed `Binary` via `Binary.fromConstantByteArray(getBytes())`. When the backing buffer is heap-based, it delegates to `super.copy()` to preserve the existing behavior and avoid unnecessary copies.

### Are these changes tested?

Tests were added that verify `copy()` on a direct `ByteBuffer`-backed `Binary` returns an independent heap-backed instance whose data remains accessible after the source buffer's is closed. The existing test suites continue to pass.

No API changes. `Binary.copy()` now returns a heap-backed copy instead of this when the source is backed by a direct `ByteBuffer`. This is a correctness fix, callers already expected `copy()` to return an owned value, and the previous behavior silently violated that contract for direct buffers.

Closes #3489

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -498,6 +498,16 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
     }
 
     @Override
+    public Binary copy() {
+      if (value.isDirect()) {
+        // Direct ByteBuffers may be backed by memory that can be freed independently, so always materialize to
+        // a heap-backed copy to avoid use-after-free.
+        return Binary.fromConstantByteArray(getBytes());
+      }
+      return super.copy();
+    }
+
+    @Override
     public int hashCode() {
       if (value.hasArray()) {
         return Binary.hashCode(value.array(), value.arrayOffset() + offset, length);
--- a/parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
+++ b/parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
@@ -20,6 +20,7 @@ package org.apache.parquet.io.api;
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
@@ -106,6 +107,27 @@ public class TestBinary {
     }
   };
 
+  private static final BinaryFactory DIRECT_BUFFER_BF = new BinaryFactory() {
+    @Override
+    public BinaryAndOriginal get(byte[] bytes, boolean reused) throws Exception {
+      ByteBuffer direct = ByteBuffer.allocateDirect(bytes.length);
+      direct.put(bytes);
+      direct.flip();
+      Binary b;
+
+      if (reused) {
+        b = Binary.fromReusedByteBuffer(direct);
+      } else {
+        b = Binary.fromConstantByteBuffer(direct);
+      }
+
+      assertArrayEquals(bytes, b.getBytes());
+      // Return the backing byte[] so tests can mutate it, though for direct buffers
+      // there is no accessible backing array. We return a copy of the original bytes.
+      return new BinaryAndOriginal(b, bytes);
+    }
+  };
+
   private static final BinaryFactory STRING_BF = new BinaryFactory() {
     @Override
     public BinaryAndOriginal get(byte[] bytes, boolean reused) throws Exception {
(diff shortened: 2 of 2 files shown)


Fokko (approved):

Fokko (approved):
Nice one 👍

PR #3496, opened 2026-04-19
GH-3495: Optimize PlainValuesWriter with direct ByteBuffer slab writes (~2.5x encode speedup)

### LINKED ISSUE #3495: Optimize PlainValuesWriter by writing directly to ByteBuffer slabs (up to 2x encode speedup)

`PlainValuesWriter` (used for PLAIN-encoded INT32, INT64, FLOAT, DOUBLE, and BINARY
columns) currently writes each value through two layers of abstraction:

```
PlainValuesWriter -> LittleEndianDataOutputStream -> CapacityByteArrayOutputStream
```

Per `writeInt()`, `LittleEndianDataOutputStream` decomposes the int into 4 bytes
in a temporary `writeBuffer[8]` array and calls `out.write(writeBuffer, 0, 4)`,
which dispatches through the `OutputStream` chain into `CapacityByteArrayOutputStream`.
That path performs:

- 4 byte-shift operations for little-endian decomposition
- 1 intermediate `writeBuffer[8]` array write
- 2 levels of virtual dispatch
- 1 bounds check in `write(byte[], off, len)`
- 1 `System.arraycopy` for 4 bytes

Since `CapacityByteArrayOutputStream` already buffers into `ByteBuffer` slabs
internally, the entire chain can be collapsed into a single `ByteBuffer.putInt()`
call, which is a HotSpot intrinsic that compiles to a single unaligned store on
x86/ARM when the buffer is in `LITTLE_ENDIAN` order.

### Proposal

1. In `CapacityByteArrayOutputStream`:
   - Set `ByteOrder.LITTLE_ENDIAN` on newly allocated slabs in `addSlab()`.
   - Add `writeInt(int)` and `writeLong(long)` methods that call
     `currentSlab.putInt(v)` / `currentSlab.putLong(v)` directly, with a single
     remaining-check that grows the slab if needed.

2. In `PlainValuesWriter`:
   - Remove the `LittleEndianDataOutputStream` field entirely.
   - `writeInteger(v)` -> `arrayOut.writeInt(v)`
   - `writeLong(v)` -> `arrayOut.writeLong(v)`
   - `writeFloat(v)` -> `arrayOut.writeInt(Float.floatToIntBits(v))`
   - `writeDouble(v)` -> `arrayOut.writeLong(Double.doubleToLongBits(v))`
   - `writeBytes(Binary v)` -> `arrayOut.writeInt(v.length()); v.writeTo(arrayOut);`
   - `getBytes()` no longer needs to flush a buffering layer.
   - `close()` no longer closes the defunct stream.

What was eliminated per `writeInt` call:

- 4 byte-shift operations for little-endian decomposition
- 1 intermediate `writeBuffer[8]` array write
- 2 levels of virtual dispatch
- 1 bounds check in `write(byte[], off, len)`
- 1 `System.arraycopy` for 4 bytes

Replaced with:

- 1 remaining-check on the slab `ByteBuffer`
- 1 `ByteBuffer.putInt()` call (single JVM intrinsic, ~1 store instruction on
  little-endian architectures)

### Benchmark results

`IntEncodingBenchmark.encodePlain` (100,000 INT32 values per invocation, JMH
`-wi 3 -i 5 -f 1`):

| Pattern          | Before (ops/s) | After (ops/s) | Improvement |
|------------------|---------------:|--------------:|------------:|
| SEQUENTIAL       |     26,817,451 |    52,953,193 | **+97.5% (2.0x)** |
| RANDOM           |     28,517,312 |    37,774,036 | **+32.5%** |
| LOW_CARDINALITY  |     28,705,158 |    52,819,678 | **+84.0%** |
| HIGH_CARDINALITY |     28,595,519 |    37,862,571 | **+32.4%** |

The improvement varies by pattern: SEQUENTIAL and LOW_CARDINALITY see ~2x because
the slab `putInt()` path has highly predictable branching (slab rarely runs out
for sequential writes). RANDOM and HIGH_CARDINALITY still see a solid +32%
improvement.

The same code path also benefits `writeLong()`, `writeFloat()`, `writeDouble()`,
and the length prefix written by `writeBytes(Binary)`.

Decode round-trip verified: re-reading the encoded data with `PlainValuesReader`
produces identical values at ~1.15B ops/s.

### Validation

All 573 `parquet-column` tests and 308 `parquet-common` tests pass with the
change applied.

### Component(s)

Core

iemejia:
## Summary

Closes #3495. Also subsumes #3518 (closed; PR #3519 closed in favor of folding here).

Three-commit PR: optimize `PlainValuesWriter`, deprecate the now-unused `LittleEndianDataOutputStream` wrapper, and apply the long-standing bulk-write TODO inside that class for the benefit of any remaining external callers.

### Commit 1 — `PlainValuesWriter` direct ByteBuffer writes

Removes the `LittleEndianDataOutputStream` layer between `PlainValuesWriter` and `CapacityByteArrayOutputStream`. Adds `writeInt(int)`/`writeLong(long)` methods on CBOS that write directly to its internal `ByteBuffer` slabs (set to `LITTLE_ENDIAN`), making the value write a single HotSpot intrinsic instead of a 4-byte decomposition through a temp array and an `OutputStream` chain.

**`IntEncodingBenchmark.encodePlain`** (100k INT32 / invocation, JMH `-wi 5 -i 10 -f 2`):

| Pattern           | Before (ops/s) | After (ops/s) | Improvement |
|-------------------|---------------:|--------------:|:-----------:|
| SEQUENTIAL        |     20,944,860 |    53,231,121 | **+154%** (2.55x) |
| RANDOM            |     20,613,242 |    53,419,118 | **+160%** |
| LOW_CARDINALITY   |     20,749,103 |    53,510,247 | **+158%** |
| HIGH_CARDINALITY  |     20,521,786 |    52,825,012 | **+157%** |

The same code path is shared by `writeLong`, `writeFloat`, `writeDouble`, and the length prefix in `writeBytes(Binary)`, so PLAIN-encoded `INT64`/`FLOAT`/`DOUBLE`/`BINARY` columns benefit too. Decode benchmarks (`decodePlain` etc.) are unchanged, as expected.

### Commit 2 — Deprecate `LittleEndianDataOutputStream`, remove last wrapper usages

Pure API cleanup, **no measurable performance impact**. After commit 1, `FixedLenByteArrayPlainValuesWriter` and `DeltaLengthByteArrayValuesWriter` were the last two production usages of `LittleEndianDataOutputStream`. Both wrapped a `CapacityByteArrayOutputStream` only to call `Binary.writeTo(out)`, which goes through `OutputStream.write(byte[], int, int)` — the wrapper added nothing for that call. Removing the wrapper allows marking `LittleEndianDataOutputStream` as `@Deprecated` (kept for binary compatibility, scheduled for removal in a future major).

Benchmarks for the two touched paths (`BinaryEncodingBenchmark`, JMH `-wi 5 -i 10 -f 3`, 30 samples per row) are within ±5% with allocation rates per op unchanged within 2% — consistent with noise rather than a real effect either way. Rationale is code health (one fewer wrapper layer, deprecation of an internal-shaped public class), not performance. Full numbers are in the commit message.

### Commit 3 — Bulk write in `LittleEndianDataOutputStream.writeInt`/`writeShort`

Resolves the long-standing TODO in `writeInt` flagging per-byte writes as a potential improvement. Replaces 4 individual `out.write(int)` calls with a single `out.write(byte[], 0, 4)` using the existing `writeBuffer[]` field — matching the pattern already used by `writeLong`. Same change applied to `writeShort` (4 → 1 call collapses to 2 → 1).

Although commit 2 deprecates the class and no Parquet code path still uses it, this optimization is worth landing for any external Parquet-format producer still on the wrapper until they migrate to direct ByteBuffer writes. The change is ~10 lines and obviously correct (mirrors the existing `writeLong` pattern).

Benchmark (`IntEncodingBenchmark.encodePlain` routed through `LittleEndianDataOutputStream`):

| | Before (ops/s) | After (ops/s) | Improvement |
|---|---:|---:|:---:|
| writeInt path | ~20.9M | ~28.2M | **+35%** |

## Validation

- `parquet-common`: 308 tests pass

## Related

This is the second in a small series of focused performance PRs from work in https://github.com/iemejia/parquet-perf. The first was #3494.

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar 'IntEncodingBenchmark.encodePlain' \
    -wi 5 -i 10 -f 3
```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/deltalengthbytearray/DeltaLengthByteArrayValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/deltalengthbytearray/DeltaLengthByteArrayValuesWriter.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.CapacityByteArrayOutputStream;
-import org.apache.parquet.bytes.LittleEndianDataOutputStream;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.values.ValuesWriter;
 import org.apache.parquet.column.values.delta.DeltaBinaryPackingValuesWriter;
@@ -46,11 +45,9 @@ public class DeltaLengthByteArrayValuesWriter extends ValuesWriter {
 
   private ValuesWriter lengthWriter;
   private CapacityByteArrayOutputStream arrayOut;
-  private LittleEndianDataOutputStream out;
 
   public DeltaLengthByteArrayValuesWriter(int initialSize, int pageSize, ByteBufferAllocator allocator) {
     arrayOut = new CapacityByteArrayOutputStream(initialSize, pageSize, allocator);
-    out = new LittleEndianDataOutputStream(arrayOut);
     lengthWriter = new DeltaBinaryPackingValuesWriterForInteger(
         DeltaBinaryPackingValuesWriter.DEFAULT_NUM_BLOCK_VALUES,
         DeltaBinaryPackingValuesWriter.DEFAULT_NUM_MINIBLOCKS,
@@ -63,7 +60,7 @@ public class DeltaLengthByteArrayValuesWriter extends ValuesWriter {
   public void writeBytes(Binary v) {
     try {
       lengthWriter.writeInteger(v.length());
-      v.writeTo(out);
+      v.writeTo(arrayOut);
     } catch (IOException e) {
       throw new ParquetEncodingException("could not write bytes", e);
     }
@@ -76,11 +73,6 @@ public class DeltaLengthByteArrayValuesWriter extends ValuesWriter {
 
   @Override
   public BytesInput getBytes() {
-    try {
-      out.flush();
-    } catch (IOException e) {
-      throw new ParquetEncodingException("could not write page", e);
-    }
     LOG.debug("writing a buffer of size {}", arrayOut.size());
     return BytesInput.concat(lengthWriter.getBytes(), BytesInput.from(arrayOut));
   }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/FixedLenByteArrayPlainValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/FixedLenByteArrayPlainValuesWriter.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.CapacityByteArrayOutputStream;
-import org.apache.parquet.bytes.LittleEndianDataOutputStream;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.values.ValuesWriter;
 import org.apache.parquet.io.ParquetEncodingException;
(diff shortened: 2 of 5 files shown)


arouel (approved):

Fokko (approved):

iemejia:
Closing in favor of #3565.

PR #3498, opened 2026-04-19
GH-3497: Fix thread-unsafe singleton in `DefaultValuesWriterFactory`

### LINKED ISSUE #3497: `DefaultValuesWriterFactory` uses thread-unsafe static singletons, causing cross-writer allocator contamination

`DefaultValuesWriterFactory` delegates to static singleton instances of `DefaultV1ValuesWriterFactory` / `DefaultV2ValuesWriterFactory`. Both singletons store a mutable `parquetProperties` field that is overwritten on every `initialize()` call. When multiple `ParquetProperties` instances exist concurrently in the same JVM, the last caller to `initialize()` wins, causing other writers to silently use a foreign `ParquetProperties` and its `ByteBufferAllocator`.

### To Reproduce
1. Create two `ParquetProperties` instances, each with a different `ByteBufferAllocator`
2. Use them concurrently from different threads to write Parquet files
3. Close the writer whose allocator "leaked" into the other's factory
4. The other writer crashes on the next buffer write

### Expected behavior
Each `ParquetProperties` instance should use its own allocator exclusively. The `ValuesWriterFactory` should not share mutable state across independent writer instances.

### Version
1.17.0 (also affects all prior versions with this code pattern)

### Component(s)

Core

arouel:
### Rationale for this change

`DefaultValuesWriterFactory` delegates to static singleton instances of `DefaultV1ValuesWriterFactory` and `DefaultV2ValuesWriterFactory`. These singletons store a mutable `parquetProperties` reference via `initialize()`, which gets overwritten by whichever `ParquetProperties` instance calls it last. When multiple Parquet writers run concurrently in the same JVM, the merger's column writers end up using the appender's `ByteBufferAllocator`. When the appender is closed and its allocator released, the merger crashes with `IllegalStateException: Already closed` on direct ByteBuffer access because its encoder slabs were allocated from a now-closed arena.

Replace the static final `DEFAULT_V1_WRITER_FACTORY` and `DEFAULT_V2_WRITER_FACTORY` singletons in `DefaultValuesWriterFactory` with fresh instances created per `initialize()` call. This ensures each `ParquetProperties` instance (and its allocator) is isolated to its own factory, eliminating cross-contamination between concurrent writers sharing the same JVM.

### Are these changes tested?

Yes. A new test `testFactoryIsolation_eachPropertiesUsesOwnAllocator` in `DefaultValuesWriterFactoryTest` verifies that two independently built `ParquetProperties` instances with different allocators produce `ValuesWriter` instances that use their own respective allocators. The test fails against the original code.

No API changes.

Closes #3497

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultValuesWriterFactory.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/factory/DefaultValuesWriterFactory.java
@@ -33,15 +33,12 @@ public class DefaultValuesWriterFactory implements ValuesWriterFactory {
 
   private ValuesWriterFactory delegateFactory;
 
-  private static final ValuesWriterFactory DEFAULT_V1_WRITER_FACTORY = new DefaultV1ValuesWriterFactory();
-  private static final ValuesWriterFactory DEFAULT_V2_WRITER_FACTORY = new DefaultV2ValuesWriterFactory();
-
   @Override
   public void initialize(ParquetProperties properties) {
     if (properties.getWriterVersion() == WriterVersion.PARQUET_1_0) {
-      delegateFactory = DEFAULT_V1_WRITER_FACTORY;
+      delegateFactory = new DefaultV1ValuesWriterFactory();
     } else {
-      delegateFactory = DEFAULT_V2_WRITER_FACTORY;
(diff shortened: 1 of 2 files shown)


Fokko (approved):
LGTM

PR #3500, opened 2026-04-19
GH-3499: Cache hashCode() for non-reused Binary instances (up to 73x dictionary-encode speedup)

### LINKED ISSUE #3499: Cache hashCode() for non-reused Binary instances (huge dictionary-encode speedup)

PLAIN_DICTIONARY encoding of `BINARY` columns repeatedly hashes `Binary` keys during dictionary map lookups (`Object2IntMap.put` / `getInt` calls inside `DictionaryValuesWriter.PlainBinaryDictionaryValuesWriter`). Each call to `Binary.hashCode()` recomputes the FNV-style hash byte-by-byte across the full backing buffer:

```java
private static final int hashCode(byte[] array, int offset, int length) {
  int result = 1;
  for (int i = offset; i < offset + length; i++) {
    byte b = array[i];
    result = 31 * result + b;
  }
  return result;
}
```

For columns with many repeated values, the same `Binary` instance is hashed many times — once per dictionary lookup attempt across the entire page. As string length grows, the cost is dominated by these recomputations.

JMH (`BinaryEncodingBenchmark.encodeDictionary`, 100k values per invocation, JDK 18, JMH -wi 5 -i 10 -f 3) on `master`:

| cardinality | stringLength | ops/s |
|-------------|--------------:|--------:|
| LOW         |  10           |  13.2M |
| LOW         | 100           |   3.0M |
| LOW         | 1000          | 300K |
| HIGH        |  10           | 848K |
| HIGH        | 100           | 418K |
| HIGH        | 1000          |  72.5K |

The 1000-byte LOW case spending 3 ms to encode 100k values is essentially all hashing work.

### Proposal

Cache `hashCode()` per `Binary` instance using the `java.lang.String.hashCode()` idiom — a single `int` field with sentinel `0` meaning "not yet computed":

```java
@Override
public int hashCode() {
  int h = cachedHashCode;
  if (h != 0) {
    return h;
  }
  return cacheHashCode(Binary.hashCode(value, offset, length));
}
```

Properties:

- **Race-safe without `volatile`**: the value computed is a deterministic function of immutable bytes, so any two threads that race on the first call produce the same `int` and either ordering is correct (per `java.lang.String`).
- **Reused (mutable-buffer) Binary instances do not cache**, preserving the current contract that mutating the backing array between calls produces a different hash.
- **Hash that genuinely equals 0 is recomputed every call** — acceptably rare and still semantically correct.

Expected speedup (JMH same configuration, 30 samples per row):

| cardinality | stringLength | Before | After | Δ |
|-------------|--------------:|-------:|------:|---:|
| LOW         |  10           | 13.2M | 20.2M | **+53%** |
| LOW         | 100           |  3.0M | 18.0M | **+511%** |
| LOW         | 1000          | 300K  | 21.9M | **+7193% (72.9x)** |
| HIGH        |  10           | 848K  | 1.34M | **+58%** |
| HIGH        | 100           | 418K  | 1.32M | **+216%** |
| HIGH        | 1000          | 72.5K | 1.30M | **+1687% (17.9x)** |

### Scope

- Single file change to `parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java`.
- Adds two unit tests in `TestBinary` for cache correctness (constant cached, reused not cached).
- No public-API change. Adds one package-private field (`transient int cachedHashCode`) and one package-private helper (`cacheHashCode(int)`).
- No allocation-rate change (cache is one extra `int` field on existing instances).

### Relation

This is part of a small series of focused PRs upstreaming optimizations from work in [parquet-perf](https://github.com/iemejia/parquet-perf). Previous PRs in the series: #3494 (PlainValuesReader), #3496 (PlainValuesWriter).

iemejia:
## Summary

Closes #3499.

Caches `Binary.hashCode()` per instance for non-reused (immutable-backing) `Binary` values. Eliminates repeated full-buffer hash recomputation during `PLAIN_DICTIONARY` encoding, where the same key is hashed many times across a 100k-value page. Reused (mutable-backing) instances skip the cache to preserve their existing semantics.

Uses the `java.lang.String.hashCode()` idiom — a single `int` field with sentinel `0` meaning "not yet computed" — so the cache is **race-safe without `volatile`** (concurrent first calls compute the same deterministic value; either ordering is correct).

## Benchmark

`BinaryEncodingBenchmark.encodeDictionary`, 100k BINARY values per invocation, JDK 18, JMH `-wi 5 -i 10 -f 3` (30 samples per row):

| cardinality | stringLength | Before (ops/s) | After (ops/s) | Improvement |
|-------------|--------------:|---------------:|--------------:|------------:|
| LOW         |  10           |     13,170,110 |    20,203,480 | **+53%**    |
| LOW         | 100           |      2,955,460 |    18,048,610 | **+511%**   |
| LOW         | 1000          |        300,693 |    21,933,470 | **+7193% (72.9x)** |
| HIGH        |  10           |        847,657 |     1,336,238 | **+58%**    |
| HIGH        | 100           |        418,327 |     1,323,284 | **+216%**   |
| HIGH        | 1000          |         72,553 |     1,296,679 | **+1687% (17.9x)** |

The relative gain grows with string length (per-call hash work is O(N), cache lookup is O(1)) and with low cardinality (each unique key is hashed many more times).

**Negative control**: `encodePlain` (writes Binary without dictionary lookups, so doesn't exercise `hashCode`) is unchanged within ±2.5% across all parameter combinations. **Allocation rate per op** (`gc.alloc.rate.norm`) is identical between baseline and optimized — speedup is pure CPU saved.

## Implementation notes

- Single field added: `transient int cachedHashCode` on `Binary` (package-private so the three nested subclasses can read it directly on the hot path; inherited private fields are not accessible from nested subclasses without a method-call indirection).
- The cached value is gated on `!isBackingBytesReused` inside a small package-private `cacheHashCode(int)` helper that runs only on the cache-miss path.
- 2 new tests in `TestBinary`:
  - `testHashCodeCachedForConstantBinary`: constant Binary returns stable `hashCode`, equal across the three impls (`ByteArraySliceBackedBinary`, `ByteArrayBackedBinary`, `ByteBufferBackedBinary`).
  - `testHashCodeNotCachedForReusedBinary`: reused Binary returns the new hash after the backing buffer is replaced.

## Validation

- `parquet-column`: 575 tests pass (was 573; +2 new tests for the cache).
- Built with `-Dspotless.check.skip=true -Drat.skip=true -Djapicmp.skip=true`.

## Related

This is the third in a small series of focused performance PRs from work in https://github.com/iemejia/parquet-perf. Previous: #3494 (PlainValuesReader), #3496 (PlainValuesWriter).

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar 'BinaryEncodingBenchmark.encodeDictionary' \
    -wi 5 -i 10 -f 3
```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -37,6 +37,18 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
 
   protected boolean isBackingBytesReused;
 
+  /**
+   * Cached hash code for non-reused (immutable) Binary instances.
+   * <p>The sentinel value {@code 0} means "not yet computed". This follows the
+   * {@link String#hashCode()} idiom: races between concurrent first calls are benign
+   * because the computation is deterministic, and a hash that genuinely equals {@code 0}
+   * will simply be recomputed on every call (acceptably rare). Reused instances never
+   * cache (their backing bytes can mutate after construction).
+   * <p>Package-private (rather than private) so subclasses can read it directly without
+   * an extra method call on the {@link #hashCode()} hot path.
+   */
+  transient int cachedHashCode;
+
   // this isn't really something others should extend
   private Binary() {}
 
@@ -99,6 +111,18 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
       return equals((Binary) obj);
     }
     return false;
+  }
+
+  /**
+   * Caches {@code hashCode} for non-reused instances and returns it. The cache uses
+   * a single int field with sentinel {@code 0} to remain race-safe without volatile.
+   * If the computed hash is {@code 0}, no caching occurs and the next call recomputes.
+   */
+  final int cacheHashCode(int hashCode) {
+    if (!isBackingBytesReused) {
+      cachedHashCode = hashCode;
+    }
+    return hashCode;
   }
 
   @Override
@@ -180,7 +204,11 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
 
     @Override
     public int hashCode() {
-      return Binary.hashCode(value, offset, length);
+      int h = cachedHashCode;
+      if (h != 0) {
+        return h;
+      }
+      return cacheHashCode(Binary.hashCode(value, offset, length));
     }
 
     @Override
@@ -340,7 +368,11 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
 
     @Override
     public int hashCode() {
-      return Binary.hashCode(value, 0, value.length);
+      int h = cachedHashCode;
+      if (h != 0) {
+        return h;
+      }
+      return cacheHashCode(Binary.hashCode(value, 0, value.length));
     }
 
     @Override
@@ -499,11 +531,18 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
 
     @Override
     public int hashCode() {
-      if (value.hasArray()) {
-        return Binary.hashCode(value.array(), value.arrayOffset() + offset, length);
-      } else {
-        return Binary.hashCode(value, offset, length);
+      int h = cachedHashCode;
+      if (h != 0) {
+        return h;
       }
+
+      int computedHashCode;
+      if (value.hasArray()) {
+        computedHashCode = Binary.hashCode(value.array(), value.arrayOffset() + offset, length);
+      } else {
+        computedHashCode = Binary.hashCode(value, offset, length);
+      }
+      return cacheHashCode(computedHashCode);
     }
 
     @Override
--- a/parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
+++ b/parquet-column/src/test/java/org/apache/parquet/io/api/TestBinary.java
@@ -155,6 +155,51 @@ public class TestBinary {
     assertEquals(bin1, bin2);
   }
 
+  /**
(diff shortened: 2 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -101,6 +113,18 @@ public boolean equals(Object obj) {
     return false;
   }
 
+  /**
+   * Caches {@code hashCode} for non-reused instances and returns it. The cache uses
+   * a single int field with sentinel {@code 0} to remain race-safe without volatile.
+   * If the computed hash is {@code 0}, no caching occurs and the next call recomputes.
+   */
+  final int cacheHashCode(int hashCode) {
```
arouel:
```suggestion
  private int cacheHashCode(int hashCode) {
```

arouel (commented):
I use a similar optimization already in a patched `parquet-column` version on my side and verified the improvement.

arouel (approved):

parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
iemejia: I had to revert this because the classes the inheriting classes were not able to use the method. I think this is the best compromise final and package private

arouel: Ah, that was an oversight by me. Sorry

iemejia:
Closing in favor of #3566.

PR #3502, opened 2026-04-19
GH-3501: Speed up `DictionaryValuesWriter.shouldFallBack()`

### LINKED ISSUE #3501: Avoid per-write virtual dispatch in `DictionaryValuesWriter.shouldFallBack()` by caching the size-exceeded check

`DictionaryValuesWriter.shouldFallBack()` is called by `FallbackValuesWriter.checkFallback()` after every single value write. The current implementation dispatches a virtual call to `getDictionarySize()` on every invocation:
```java
public boolean shouldFallBack() {
  return dictionaryByteSize > maxDictionaryByteSize || getDictionarySize() > MAX_DICTIONARY_ENTRIES;
}
```
`getDictionarySize()` is an abstract method overridden in each typed subclass (Binary, Long, Double, Integer, Float) to return the backing map's `.size()`. Since `shouldFallBack()` is polled after every write, including writes of duplicate values that do not grow the dictionary, the virtual dispatch and map-size query are redundant work for the common case where most values are already in the dictionary.
Both `dictionaryByteSize` and the dictionary entry count can only increase when a new entry is added (inside the `if (id == -1)` branch of each subclass's write method). Therefore the size-exceeded condition can only transition from `false` to `true` at that exact point.

### Proposal
Replace the per-write check with a cached boolean `dictionarySizeExceeded` flag. Introduce a `checkDictionarySizeLimit(int newDictionarySize)` method that subclass write methods call only when a new dictionary entry is actually added. `shouldFallBack()` then returns the cached flag directly, a simple field read with no virtual dispatch.

### Component(s)

Core

arouel:
### Rationale for this change

`DictionaryValuesWriter.shouldFallBack()` is called by `FallbackValuesWriter.checkFallback()` after every single value write. The current implementation dispatches a virtual call to `getDictionarySize()` on every invocation, even for duplicate values that do not grow the dictionary. Both `dictionaryByteSize` and the dictionary entry count can only increase when a new entry is added (inside the `if (id == -1)` branch), so the size-exceeded condition can only transition from false to true at that exact point. The per-write virtual dispatch is redundant work for the common case.

Single file change to DictionaryValuesWriter.java:
- Add a private boolean `dictionarySizeExceeded` field, reset in `resetDictionary()`.
- Add a protected void `checkDictionarySizeLimit(int newDictionarySize)` method that sets the flag when `dictionaryByteSize > maxDictionaryByteSize || newDictionarySize > MAX_DICTIONARY_ENTRIES`.
- Each subclass write method (Binary, FixedLenArray, Long, Double, Integer, Float) calls `checkDictionarySizeLimit(id + 1)` after inserting a new dictionary entry.
- `shouldFallBack()` becomes return `dictionarySizeExceeded`, a simple field read with no virtual dispatch.

### Are these changes tested?

Yes. All existing tests in `TestDictionary` pass without modification, confirming semantic equivalence. The new test `testCheckDictionarySizeLimitExceedsByEntryCount` directly exercises the previously untested `MAX_DICTIONARY_ENTRIES` boundary in `checkDictionarySizeLimit`, verifying correct trigger and reset behavior.

No. This is an internal optimization with no API, behavior, or configuration changes.

Closes #3501

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
@@ -79,6 +79,10 @@ public abstract class DictionaryValuesWriter extends ValuesWriter implements Req
   /* will become true if the dictionary becomes too big */
   protected boolean dictionaryTooBig;
 
+  /* set to true when the dictionary exceeds maxDictionaryByteSize or MAX_DICTIONARY_ENTRIES,
+   * checked by shouldFallBack() to avoid repeated virtual dispatch to getDictionarySize() on every write */
+  private boolean dictionarySizeExceeded;
+
   /* current size in bytes the dictionary will take once serialized */
   protected long dictionaryByteSize;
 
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
```diff
@@ -121,8 +125,20 @@ protected DictionaryPage dictPage(ValuesWriter dictPageWriter) {
...
+  }
+
+  /**
+   * Called by subclass write methods after adding a new dictionary entry to check if the dictionary
+   * has exceeded its size limits. This avoids the per-value virtual dispatch overhead of calling
+   * getDictionarySize() on every write -- the check only runs when a new entry is actually added.
+   *
+   * @param newDictionarySize the current dictionary size after adding the new entry
+   */
+  protected void checkDictionarySizeLimit(int newDictionarySize) {
```
Fokko: Since `dictionarySizeExceeded` is private, shouldn't this method be private as well?

```suggestion
  private void checkDictionarySizeLimit(int newDictionarySize) {
```

arouel: Making the method `private` would break the introduced test case, but I made the method package-private. Would that be sufficient from your perspective?

Fokko:
@arouel Could you pull in latest main?

PR #3508, opened 2026-04-19
GH-3507: Reduce allocation overhead of no-op `GeospatialStatistics.Builder`

### LINKED ISSUE #3507: Reduce allocation overhead of no-op `GeospatialStatistics.Builder`

`GeospatialStatistics.noopBuilder()` is called for every non-geometry column in every row group. Today, each call allocates a new `NoopBuilder` (plus a `WKBReader`, `BoundingBox`, and `GeospatialTypes` that are never used), and each `build()` allocates a new `GeospatialStatistics(null, null)` that is immediately discarded.

Since `NoopBuilder` carries no state and all its methods are no-ops, it can be a singleton. Its `build()` result is equally stateless (null fields, with merge/abort guarded by null checks), so it can be cached too.

### Proposed Changes
- Make `NoopBuilder` a singleton via a static `INSTANCE` field
- Cache `build()` output as a static `EMPTY` instance
- Add a package-private super constructor that skips field initialization so the singleton doesn't allocate unused objects

This eliminates per-column-per-row-group allocations on the common path where geospatial statistics are disabled.

### Component(s)

Core

arouel:
### Rationale for this change

`GeospatialStatistics.noopBuilder()` is called for every non-geometry column in every row group. Each call allocates a new `NoopBuilder` along with a `WKBReader`, `BoundingBox`, and `GeospatialTypes` that are never used. Each subsequent `build()` allocates a new `GeospatialStatistics(null, null)` that is equally throwaway. On wide schemas with many columns, this creates unnecessary GC pressure on the write path.

- Make `NoopBuilder` a singleton via a static `INSTANCE` field returned by `noopBuilder()`
- Cache the `build()` result as a static `EMPTY` instance, since it is stateless (null fields, with merge/abort guarded by null checks)
- Add a package-private `Builder(boolean)` constructor that skips field initialization so the singleton does not allocate unused `WKBReader`, `BoundingBox`, and `GeospatialTypes` objects
- Move `WKBReader` initialization from the field declaration into the public `Builder()` constructor, so only real builders pay the cost

### Are these changes tested?

Yes. Two new tests were added to `TestGeospatialStatistics`:
- `testNoopBuilderIsSingleton`: verifies `noopBuilder()` returns the same builder instance and `build()` returns the same cached result
- `testNoopBuilderUpdateAndAbortAreNoOps`: exercises `update(nonNull)`, `update(null)`, and `abort()` on the noop builder to confirm none throw, and asserts the built result remains invalid with null fields

No. This is an internal optimization with no API or behavioral changes.

Closes #3507

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/GeospatialStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/GeospatialStatistics.java
@@ -43,7 +43,7 @@ public class GeospatialStatistics {
   public static class Builder {
     private BoundingBox boundingBox;
     private GeospatialTypes geospatialTypes;
-    private final WKBReader reader = new WKBReader();
+    private WKBReader reader;
 
     /**
      * Create a builder to create a GeospatialStatistics.
@@ -51,6 +51,14 @@ public class GeospatialStatistics {
     public Builder() {
       this.boundingBox = new BoundingBox();
       this.geospatialTypes = new GeospatialTypes();
+      this.reader = new WKBReader();
+    }
+
+    /**
+     * Internal constructor that skips field initialization for subclasses that don't need it.
+     */
+    Builder(boolean noop) {
+      // Fields intentionally left null for NoopBuilder
     }
 
     public void update(Binary value) {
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/GeospatialStatistics.java
```diff
@@ -43,14 +43,22 @@ public class GeospatialStatistics {
...
     public Builder() {
       this.boundingBox = new BoundingBox();
       this.geospatialTypes = new GeospatialTypes();
+      this.reader = new WKBReader();
+    }
+
+    /**
+     * Internal constructor that skips field initialization for subclasses that don't need it.
+     */
+    Builder(boolean noop) {
```
Fokko, line 60: I don't think this is proper use of overloading

arouel: Since it works technically, what would you expect or understand as proper?

steveloughran: It's a workaround to add second empty constructor, isn't it?
what about having a protected builder(box, types, reader) which the base builder invokes with those newly created types, and which the subclass calls with (null, null, null)? That makes clear what it is doing: creating an instance with no values

PR #3506, opened 2026-04-19
GH-3505: Optimize ByteStreamSplitValuesReader page transposition

iemejia:
### Rationale for this change

`ByteStreamSplitValuesReader` is the symmetric reader for `BYTE_STREAM_SPLIT`-encoded `FLOAT`, `DOUBLE`, `INT32`, and `INT64` columns. On `initFromPage` it eagerly transposes the entire page from stream-split layout (`elementSizeInBytes` separate streams of `valuesCount` bytes each) back to interleaved layout. The current loop is:

```java
private byte[] decodeData(ByteBuffer encoded, int valuesCount) {
  byte[] decoded = new byte[encoded.limit()];
  int destByteIndex = 0;
  for (int srcValueIndex = 0; srcValueIndex < valuesCount; ++srcValueIndex) {
    for (int stream = 0; stream < elementSizeInBytes; ++stream, ++destByteIndex) {
      decoded[destByteIndex] = encoded.get(srcValueIndex + stream * valuesCount);
    }
  }
  return decoded;
}
```

Two issues on the hot path:

1. Every read goes through `ByteBuffer.get(int)` (per-call bounds checks + virtual dispatch through `HeapByteBuffer`/`DirectByteBuffer`).
2. The inner stream offset (`stream * valuesCount`) is recomputed on every iteration even though it depends only on the outer loop.

For a 100k-value `FLOAT` page that is 400k `ByteBuffer.get(int)` calls; for `DOUBLE`/`LONG` it is 800k.

### What changes are included in this PR?

Rewrite `decodeData` in three steps:

1. **Drop down to a `byte[]` view** of the encoded buffer. When `encoded.hasArray()` is true (the typical case), use the backing array directly with the correct base offset; otherwise copy once with a single `get(byte[])` call. Eliminates the per-byte `ByteBuffer.get(int)` bounds check and virtual dispatch.

2. **Specialize loops for the common element sizes (4 and 8)**. Hoist all `stream * valuesCount` offsets into local ints (`s0..s3` for floats/ints, `s0..s7` for doubles/longs) and write each output slot exactly once in a single sequential pass. Reads come from `elementSizeInBytes` concurrent sequential streams, which modern hardware prefetchers handle well.

3. **Generic fallback** for arbitrary element sizes (`FIXED_LEN_BYTE_ARRAY` of any width).

### Benchmark

[...]

PR #3504, opened 2026-04-19
GH-3503: Optimize ByteStreamSplitValuesWriter with batched scatter writes

iemejia:
### Rationale for this change

`ByteStreamSplitValuesWriter` is the primary writer for `BYTE_STREAM_SPLIT`-encoded `FLOAT`, `DOUBLE`, `INT32`, and `INT64` columns. Each value goes through a hot path that performs both an unnecessary allocation and N single-byte virtual dispatches.

For `FloatByteStreamSplitValuesWriter.writeFloat(float v)`:

```java
super.scatterBytes(BytesUtils.intToBytes(Float.floatToIntBits(v)));
```

`BytesUtils.intToBytes` allocates a fresh `byte[4]` on every call. `scatterBytes` then loops:

```java
for (int i = 0; i < bytes.length; ++i) {
  this.byteStreams[i].write(bytes[i]);   // CapacityByteArrayOutputStream.write(int)
}
```

So **per value**: 1 `byte[4]` allocation + 4 single-byte virtual dispatches. For a 100k-value `FLOAT` page that is 100k allocations and 400k single-byte writes. `DOUBLE`/`LONG` are even worse (`byte[8]`, 800k single-byte writes).

### What changes are included in this PR?

Two stacked changes in `ByteStreamSplitValuesWriter`:

1. **Eliminate per-value allocation**: replace `super.scatterBytes(BytesUtils.intToBytes(v))` with `bufferInt(v)` / `bufferLong(v)` that perform the little-endian decomposition with bit shifts directly, no temporary `byte[]`.

2. **Batch single-byte writes**: accumulate `BATCH_SIZE = 128` values in a small per-instance scratch buffer and flush them as `N` bulk `write(byte[], off, len)` calls (one per stream), replacing `BATCH_SIZE * elementSizeInBytes` single-byte virtual dispatches with `elementSizeInBytes` bulk writes per flush. The constant was chosen by sweeping 16/32/64/128/256/512/1024 — 128 is the sweet spot for `FLOAT` throughput while still capturing most of the `DOUBLE`/`LONG` gains.

Pending values are included in `getBufferedSize()` (so page-sizing decisions remain correct) and flushed in `getBytes()`. `reset()` and `close()` clear pending state. Only the four numeric subclasses (Float/Double/Integer/Long) use the batching path; `FixedLenByteArrayByteStreamSplitValuesWriter` continues to use `scatterBytes(byte[])` since its values arrive as already-laid-out byte arrays.

### Benchmark

[...]

PR #3510, opened 2026-04-19
GH-3509: Optimize BinaryPlainValuesReader by reading directly from ByteBuffer

### LINKED ISSUE #3509: Optimize BinaryPlainValuesReader by reading directly from ByteBuffer

`BinaryPlainValuesReader.readBytes()` is the hot-path decoder for `BINARY` (and `STRING`) columns using `PLAIN` encoding. The current implementation funnels every length read through `BytesUtils.readIntLittleEndian(InputStream)` and every value slice through `ByteBufferInputStream.slice(int)`:

```java
public Binary readBytes() {
  try {
    int length = BytesUtils.readIntLittleEndian(in);
    return Binary.fromConstantByteBuffer(in.slice(length));
  } catch (IOException | RuntimeException e) {
    throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
  }
}
```

Two issues per value:

1. `BytesUtils.readIntLittleEndian(InputStream)` calls `in.read()` four times. Each call goes through a `try` / `IOException` plumbing path and a virtual dispatch on `ByteBufferInputStream` (typically resolved as either `SingleBufferInputStream` or `MultiBufferInputStream`).
2. `in.slice(length)` is also a virtual dispatch on `ByteBufferInputStream` for every value.

If the page is materialised as a `MultiBufferInputStream` the cost is even higher because each slice may have to walk a buffer list.

JMH (`BinaryEncodingBenchmark.decodePlain`, 100k values per invocation, JDK 18, `-wi 5 -i 10 -f 3`, 30 samples) on master:

| cardinality | stringLength | ops/s   |
|------------:|-------------:|--------:|
| HIGH        | 10           | 23.11M  |
| HIGH        | 100          | 20.52M  |
| HIGH        | 1000         | 7.07M   |
| LOW         | 10           | 22.89M  |
| LOW         | 100          | 20.35M  |
| LOW         | 1000         | 6.28M   |

### Proposal

Replace the `ByteBufferInputStream` field with a single `ByteBuffer` set up once in `initFromPage`:

```java
@Override
  int available = stream.available();
  this.buffer = available > 0
      ? stream.slice(available).order(ByteOrder.LITTLE_ENDIAN)
      : ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
}

@Override
public Binary readBytes() {
  int length = buffer.getInt();
  ByteBuffer valueSlice = buffer.slice();
  valueSlice.limit(length);
  buffer.position(buffer.position() + length);
  return Binary.fromConstantByteBuffer(valueSlice);
}
```

The length prefix is now a single `ByteBuffer.getInt()` (one bounds check, no `IOException` plumbing, JIT-friendly intrinsic on little-endian buffers) and each value slice is a direct `ByteBuffer.slice()` instead of a virtual `ByteBufferInputStream.slice(int)`.

The trade-off: when the input is a `MultiBufferInputStream` the upfront `stream.slice(available)` call may consolidate the page into a single fresh `ByteBuffer`. This is one allocation per page in exchange for inlined per-value reads, which is a clear win whenever the page contains more than a handful of values.

Expected speedup (same JMH config):

| cardinality | stringLength | Before  | After   | Δ       |
|------------:|-------------:|--------:|--------:|--------:|
| HIGH        | 10           | 23.11M  | 27.13M  | **+17.4% (1.17x)** |
| HIGH        | 100          | 20.52M  | 22.20M  | **+8.2% (1.08x)**  |
| HIGH        | 1000         | 7.07M   | 7.68M   | **+8.6% (1.09x)**  |
| LOW         | 10           | 22.89M  | 26.46M  | **+15.6% (1.16x)** |
| LOW         | 100          | 20.35M  | 22.16M  | **+8.9% (1.09x)**  |
| LOW         | 1000         | 6.28M   | 7.50M   | **+19.4% (1.19x)** |

Allocation per op is unchanged (~88 B/op = the returned `Binary` + the per-value `ByteBuffer` slice).

The improvement is largest at small string lengths because the per-value fixed cost (length read + slice) dominates more there; at 1000-byte values the cost is increasingly dominated by the value-bytes copy/compare downstream rather than the read itself, but the gain is still ~9–19% even there.

### Scope

- Single file change to `parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java`.
- No public-API change; only the implementation of `readBytes()`, `skip()`, and `initFromPage()` is rewritten.
- All 573 `parquet-column` tests pass.

### Relation

Part of a small series of focused performance PRs from work in [parquet-perf](https://github.com/iemejia/parquet-perf). Previous: #3494 (PlainValuesReader), #3496 (PlainValuesWriter), #3500 (Binary.hashCode cache), #3504 (BSS writer), #3506 (BSS reader).

iemejia:
### Rationale for this change

`BinaryPlainValuesReader.readBytes()` is the hot-path decoder for `BINARY` (and `STRING`) columns using `PLAIN` encoding. The current implementation funnels every length read through `BytesUtils.readIntLittleEndian(InputStream)` and every value slice through `ByteBufferInputStream.slice(int)`:

```java
public Binary readBytes() {
  try {
    int length = BytesUtils.readIntLittleEndian(in);
    return Binary.fromConstantByteBuffer(in.slice(length));
  } catch (IOException | RuntimeException e) {
    throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
  }
}
```

Two issues per value:

1. `BytesUtils.readIntLittleEndian(InputStream)` calls `in.read()` four times — full `IOException` plumbing and virtual dispatch on `ByteBufferInputStream` (`SingleBufferInputStream` / `MultiBufferInputStream`).
2. `in.slice(length)` is also a virtual dispatch on `ByteBufferInputStream` for every value.

If the page is materialised as a `MultiBufferInputStream` the cost is even higher because each slice may have to walk a buffer list.

Replace the `ByteBufferInputStream` field with a single `ByteBuffer` set up once in `initFromPage`. The length prefix is then a single `ByteBuffer.getInt()` (one bounds check, JIT-friendly little-endian intrinsic, no `IOException` plumbing) and each value slice is a direct `ByteBuffer.slice()` instead of a virtual `ByteBufferInputStream.slice(int)`.

Trade-off: when the input is a `MultiBufferInputStream` the upfront `stream.slice(available)` call may consolidate the page into a single fresh `ByteBuffer`. This is one allocation per page in exchange for inlined per-value reads — a clear win whenever the page contains more than a handful of values.

### Benchmark

`BinaryEncodingBenchmark.decodePlain` (100k values per invocation, JDK 18, JMH `-wi 5 -i 10 -f 3`, 30 samples per row):

| cardinality | stringLength | Before  | After   | Δ                  |
|------------:|-------------:|--------:|--------:|-------------------:|
| HIGH        | 10           | 23.11M  | 27.13M  | **+17.4% (1.17x)** |
| HIGH        | 100          | 20.52M  | 22.20M  | **+8.2% (1.08x)**  |
| HIGH        | 1000         | 7.07M   | 7.68M   | **+8.6% (1.09x)**  |
| LOW         | 10           | 22.89M  | 26.46M  | **+15.6% (1.16x)** |
| LOW         | 100          | 20.35M  | 22.16M  | **+8.9% (1.09x)**  |
| LOW         | 1000         | 6.28M   | 7.50M   | **+19.4% (1.19x)** |

Per-op allocation is unchanged (~88 B/op = the returned `Binary` + the per-value `ByteBuffer` slice).

The improvement is largest at small string lengths because the per-value fixed cost (length read + slice) dominates more there; at 1000-byte values the cost is increasingly dominated by the value-bytes work downstream rather than the length read itself, but the gain is still ~9–19% even there.

### Are these changes tested?

Yes. All 573 `parquet-column` tests pass. No new test was added because the returned `Binary` values are byte-identical to before (covered by existing `BinaryPlainValuesReader` round-trip tests).

No. Only an internal reader optimization. No public API, file format, or configuration change.

### Closes #3509

Part of a small series of focused performance PRs from work in [parquet-perf](https://github.com/iemejia/parquet-perf). Previous: #3494, #3496, #3500, #3504, #3506.

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar 'BinaryEncodingBenchmark.decodePlain' \
    -wi 5 -i 10 -f 3
```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
@@ -19,36 +19,38 @@
 package org.apache.parquet.column.values.plain;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import org.apache.parquet.bytes.ByteBufferInputStream;
-import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.ValuesReader;
-import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.parquet.io.api.Binary;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Plain encoding reader for BINARY values.
+ *
+ * <p>Reads directly from a {@link ByteBuffer} with {@link ByteOrder#LITTLE_ENDIAN} byte order,
+ * using {@link ByteBuffer#getInt()} for the 4-byte length prefix instead of 4 individual
+ * {@code InputStream.read()} calls through {@link org.apache.parquet.bytes.BytesUtils#readIntLittleEndian}.
+ */
 public class BinaryPlainValuesReader extends ValuesReader {
   private static final Logger LOG = LoggerFactory.getLogger(BinaryPlainValuesReader.class);
-  private ByteBufferInputStream in;
+  private ByteBuffer buffer;
 
   @Override
   public Binary readBytes() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      return Binary.fromConstantByteBuffer(in.slice(length));
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
-    }
+    int length = buffer.getInt();
+    ByteBuffer valueSlice = buffer.slice();
+    valueSlice.limit(length);
+    buffer.position(buffer.position() + length);
+    return Binary.fromConstantByteBuffer(valueSlice);
   }
 
   @Override
   public void skip() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      in.skipFully(length);
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not skip bytes at offset " + in.position(), e);
-    }
+    int length = buffer.getInt();
+    buffer.position(buffer.position() + length);
   }
 
   @Override
@@ -57,6 +59,11 @@ public class BinaryPlainValuesReader extends ValuesReader {
         "init from page at offset {} for length {}",
         stream.position(),
         (stream.available() - stream.position()));
-    this.in = stream.remainingStream();
+    int available = stream.available();
+    if (available > 0) {
+      this.buffer = stream.slice(available).order(ByteOrder.LITTLE_ENDIAN);
+    } else {
+      this.buffer = ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
+    }
   }
 }


parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -57,6 +59,11 @@ public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IO
         "init from page at offset {} for length {}",
         stream.position(),
         (stream.available() - stream.position()));
-    this.in = stream.remainingStream();
+    int available = stream.available();
+    if (available > 0) {
+      this.buffer = stream.slice(available).order(ByteOrder.LITTLE_ENDIAN);
+    } else {
+      this.buffer = ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
```
Fokko, line 66: Why would we allocate an empty buffer?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -19,36 +19,38 @@
...
-  private ByteBufferInputStream in;
+  private ByteBuffer buffer;
 
   @Override
   public Binary readBytes() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      return Binary.fromConstantByteBuffer(in.slice(length));
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
```
Fokko, line 40: Should we keep those `ParquetDecodingException`?

iemejia:
Closing in favor of #3565.

PR #3512, opened 2026-04-19
GH-3511: Add JMH encoding benchmarks and fix parquet-benchmarks shaded jar

### LINKED ISSUE #3511: Add JMH benchmarks for encoding/decoding paths and fix parquet-benchmarks shaded jar
## Background

The 6 perf-optimization PRs currently open (#3494, #3496, #3500, #3504, #3506, #3510) report headline numbers (12x decode speedup, 7193% Binary.hashCode improvement, etc.) but cite JMH benchmarks that **do not exist on master**. Reviewers cannot reproduce the numbers without manually copying benchmark sources from elsewhere.

This issue tracks contributing the JMH benchmarks themselves so reviewers can reproduce, validate, and continue measuring across future changes.

### 1. `parquet-benchmarks` shaded jar is broken on master

A build of `parquet-benchmarks` from the current master produces a jar that is **non-functional**:

```
$ java -jar parquet-benchmarks/target/parquet-benchmarks.jar
Exception in thread "main" java.lang.RuntimeException: ERROR: Unable to find the resource: /META-INF/BenchmarkList
```

The `parquet-benchmarks/pom.xml` is missing two pieces of configuration:

- The `maven-compiler-plugin` lacks the `annotationProcessorPaths` / `annotationProcessors` config for `jmh-generator-annprocess`. As a result the JMH annotation processor never runs, and `META-INF/BenchmarkList` and `META-INF/CompilerHints` are never generated. (Workaround: pass `-Dmaven.compiler.proc=full`, but this is undiscoverable.)
- The `maven-shade-plugin` lacks `AppendingTransformer` entries for `META-INF/BenchmarkList` and `META-INF/CompilerHints`. Even if the resources were generated, shading would drop them.

### 2. No benchmarks for the optimizations under review

The 6 open perf PRs touch encode/decode paths in `parquet-column` and `parquet-common` (PlainValuesReader/Writer, Binary.hashCode, ByteStreamSplitValuesReader/Writer, BinaryPlainValuesReader). Master's `parquet-benchmarks` covers only file-level read/write, not these CPU-bound encoding paths.

## Proposal

Land the following in a single PR against `parquet-benchmarks`:

1. **pom.xml fix**: add JMH annotation-processor config + `AppendingTransformer` entries so the shaded jar is runnable.
2. **11 new JMH benchmark files** covering the encoding/decoding paths under optimization, plus supporting infrastructure:
   - `IntEncodingBenchmark` — encode/decode with PLAIN, DELTA_BINARY_PACKED, BYTE_STREAM_SPLIT, RLE, and dictionary, parameterized on value count and data distribution
   - `BinaryEncodingBenchmark` — Binary write/read paths (PLAIN, dictionary), parameterized on length and cardinality
   - `ByteStreamSplitEncodingBenchmark`, `ByteStreamSplitDecodingBenchmark` — BSS encode/decode for float/double/int/long
   - `FixedLenByteArrayEncodingBenchmark` — FLBA encode/decode
   - `FileReadBenchmark`, `FileWriteBenchmark` — CPU-focused file-level benchmarks (minimal I/O via temp files)
   - `RowGroupFlushBenchmark` — flush-path benchmark
   - `ConcurrentReadWriteBenchmark` — multi-threaded read/write throughput
   - `BlackHoleOutputFile` — `OutputFile` that discards bytes, used to isolate CPU work from I/O
   - `TestDataFactory` — shared test-data generation utilities

After this lands, each existing perf PR will be amended with a one-line "How to reproduce" snippet pointing at the relevant `*Benchmark` class.

### Out of scope (deferred)

The existing `ReadBenchmarks`, `WriteBenchmarks`, and `NestedNullWritingBenchmarks` could be modernized (Hadoop-free `LocalInputFile`, parameterized over compression and writer version, JMH-idiomatic state setup). That is a separate concern and will be proposed in a follow-up PR.

## Validation

With the proposed pom changes, the shaded jar contains a populated `META-INF/BenchmarkList` (87 benchmarks registered) and runs cleanly. As a sanity check, `IntEncodingBenchmark.decodePlain` reproduces the ~91M ops/s baseline cited in #3493/#3494 (master JDK 21, JMH 1.37, 3 warmup + 5 measurement iterations).

iemejia:
## Summary

Resolves #3511.

The `parquet-benchmarks` shaded jar built from current master is **non-functional** — it fails at runtime with `RuntimeException: Unable to find the resource: /META-INF/BenchmarkList`. This PR fixes that and adds 11 JMH benchmarks covering the encode/decode paths exercised by the open performance PRs, so reviewers can reproduce the reported numbers.

## What's broken on master

`parquet-benchmarks/pom.xml` is missing two pieces of configuration:

- `maven-compiler-plugin` lacks the `annotationProcessorPaths` / `annotationProcessors` config for `jmh-generator-annprocess`, so the JMH annotation processor never runs and `META-INF/BenchmarkList` / `META-INF/CompilerHints` are never generated.
- `maven-shade-plugin` lacks `AppendingTransformer` entries for those two resources, so even if generated they would be dropped during shading.

Both problems are fixed in this PR.

## Benchmarks added

11 new files in `parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/`:

| Benchmark | Coverage |
|---|---|
| `IntEncodingBenchmark` | int encode/decode: PLAIN, DELTA_BINARY_PACKED, BYTE_STREAM_SPLIT, RLE, DICTIONARY |
| `BinaryEncodingBenchmark` | Binary write/read paths, parameterized on length and cardinality |
| `ByteStreamSplitEncodingBenchmark` / `ByteStreamSplitDecodingBenchmark` | BSS for float / double / int / long |
| `FixedLenByteArrayEncodingBenchmark` | FLBA encode/decode |
| `FileReadBenchmark` / `FileWriteBenchmark` | CPU-focused file-level benchmarks |
| `RowGroupFlushBenchmark` | Flush path |
| `ConcurrentReadWriteBenchmark` | Multi-threaded read/write throughput |
| `BlackHoleOutputFile` | `OutputFile` that discards bytes — isolates CPU from I/O |
| `TestDataFactory` | Shared data-generation utilities |

## Validation

After this PR, the shaded jar is runnable and registers 87 benchmarks:

```
$ ./mvnw clean package -pl parquet-benchmarks -DskipTests \
$ java -jar parquet-benchmarks/target/parquet-benchmarks.jar -l | wc -l
87
```

Sanity check — `IntEncodingBenchmark.decodePlain` reproduces the master baseline cited in #3493/#3494 (~91M ops/s on JDK 21, JMH 1.37, 3 warmup + 5 measurement iterations):

```
Benchmark                            (dataPattern)   Mode  Cnt         Score         Error  Units
IntEncodingBenchmark.decodePlain        SEQUENTIAL  thrpt    5  93528419.575 ± 1472148.214  ops/s
IntEncodingBenchmark.decodePlain            RANDOM  thrpt    5  90908523.483 ± 1978982.394  ops/s
IntEncodingBenchmark.decodePlain   LOW_CARDINALITY  thrpt    5  92672978.255 ± 2071927.851  ops/s
IntEncodingBenchmark.decodePlain  HIGH_CARDINALITY  thrpt    5  90770177.655 ± 2427904.955  ops/s
```

## Out of scope (deferred)

Modernization of the existing `ReadBenchmarks` / `WriteBenchmarks` / `NestedNullWritingBenchmarks` (Hadoop-free `LocalInputFile`, parameterization, JMH-idiomatic state setup) is a separate concern and will be proposed in a follow-up PR.

## Follow-up

Once this lands, each open perf PR (#3494, #3496, #3500, #3504, #3506, #3510) will be updated with a one-line "How to reproduce" snippet referencing the relevant `*Benchmark` class.

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BinaryEncodingBenchmark.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BinaryEncodingBenchmark.java
@@ -20,10 +20,8 @@ package org.apache.parquet.benchmarks;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import org.apache.parquet.bytes.ByteBufferInputStream;
-import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.bytes.HeapByteBufferAllocator;
 import org.apache.parquet.column.Dictionary;
 import org.apache.parquet.column.Encoding;
@@ -61,6 +59,14 @@ import org.openjdk.jmh.infra.Blackhole;
  *
  * <p>Each benchmark invocation processes {@value #VALUE_COUNT} values. Throughput is
  * reported per-value using {@link OperationsPerInvocation}.
+ *
+ * <p>The dictionary encode/decode benchmarks intentionally measure the full path:
+ * the encoder produces both the RLE-encoded indices and a {@link DictionaryPage};
+ * the decoder consumes the indices through a {@link DictionaryValuesReader} backed
+ * by the same dictionary. If the dictionary exceeds {@link #MAX_DICT_BYTE_SIZE}
+ * (which can happen for high-cardinality, long-string parameter combinations) the
+ * writer falls back to plain encoding and dictionary decoding for that combination
+ * is skipped.
  */
 @BenchmarkMode(Mode.Throughput)
 @OutputTimeUnit(TimeUnit.SECONDS)
@@ -87,13 +93,14 @@ public class BinaryEncodingBenchmark {
   private byte[] deltaLengthEncoded;
   private byte[] deltaStringsEncoded;
   private byte[] dictEncoded;
+  private DictionaryPage dictPage;
   private Dictionary binaryDictionary;
+  private boolean dictionaryAvailable;
 
   @Setup(Level.Trial)
   public void setup() throws IOException {
-    Random random = new Random(42);
     int distinct = "LOW".equals(cardinality) ? TestDataFactory.LOW_CARDINALITY_DISTINCT : 0;
-    data = TestDataFactory.generateBinaryData(VALUE_COUNT, stringLength, distinct, random);
+    data = TestDataFactory.generateBinaryData(VALUE_COUNT, stringLength, distinct, TestDataFactory.DEFAULT_SEED);
 
     // Pre-encode data for decode benchmarks
     plainEncoded = encodeBinaryWith(newPlainWriter());
@@ -104,10 +111,13 @@ public class BinaryEncodingBenchmark {
     for (Binary v : data) {
       dictWriter.writeBytes(v);
     }
-    dictEncoded = dictWriter.getBytes().toByteArray();
-    DictionaryPage dictPage = dictWriter.toDictPageAndClose().copy();
-    binaryDictionary = new PlainValuesDictionary.PlainBinaryDictionary(dictPage);
-    dictWriter.close();
+    BenchmarkEncodingUtils.EncodedDictionary encoded = BenchmarkEncodingUtils.drainDictionary(dictWriter);
+    dictEncoded = encoded.dictData;
+    dictPage = encoded.dictPage;
+    dictionaryAvailable = !encoded.fellBackToPlain();
+    if (dictionaryAvailable) {
+      binaryDictionary = new PlainValuesDictionary.PlainBinaryDictionary(dictPage);
+    }
   }
 
   private byte[] encodeBinaryWith(ValuesWriter writer) throws IOException {
@@ -119,22 +129,12 @@ public class BinaryEncodingBenchmark {
     return bytes;
   }
 
-  private byte[] encodeDictionaryWith(DictionaryValuesWriter.PlainBinaryDictionaryValuesWriter writer)
-      throws IOException {
+  private BenchmarkEncodingUtils.EncodedDictionary encodeDictionaryWith(
+      DictionaryValuesWriter.PlainBinaryDictionaryValuesWriter writer) throws IOException {
     for (Binary v : data) {
       writer.writeBytes(v);
     }
-    BytesInput dataBytes = writer.getBytes();
-    DictionaryPage dictPage = writer.toDictPageAndClose();
-    byte[] bytes;
-    if (dictPage == null) {
-      bytes = dataBytes.toByteArray();
-    } else {
-      BytesInput allBytes = BytesInput.concat(dataBytes, dictPage.getBytes());
-      bytes = allBytes.toByteArray();
-    }
-    writer.close();
-    return bytes;
+    return BenchmarkEncodingUtils.drainDictionary(writer);
   }
 
   // ---- Writer factories ----
@@ -178,8 +178,10 @@ public class BinaryEncodingBenchmark {
 
   @Benchmark
   @OperationsPerInvocation(VALUE_COUNT)
-  public byte[] encodeDictionary() throws IOException {
-    return encodeDictionaryWith(newDictWriter());
+  public void encodeDictionary(Blackhole bh) throws IOException {
+    BenchmarkEncodingUtils.EncodedDictionary encoded = encodeDictionaryWith(newDictWriter());
+    bh.consume(encoded.dictData);
+    bh.consume(encoded.dictPage);
   }
 
   // ---- Decode benchmarks ----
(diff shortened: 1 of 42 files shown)


parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BinaryEncodingBenchmark.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(VALUE_COUNT)
+  public byte[] encodeDeltaByteArray() throws IOException {
+    return encodeBinaryWith(newDeltaStringsWriter());
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(VALUE_COUNT)
```
steveloughran, line 180: didn't know of this trick until i saw your PR; adopted in #3452 earlier today

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BlackHoleOutputFile.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.io.IOException;
+import org.apache.parquet.io.OutputFile;
+import org.apache.parquet.io.PositionOutputStream;
+
+/**
+ * A no-op {@link OutputFile} that discards all written data.
+ * Useful for isolating CPU/encoding cost from filesystem I/O in write benchmarks.
+ */
+public final class BlackHoleOutputFile implements OutputFile {
```
steveloughran, line 29: does this act as a black hole for the benchmarks? or would passing in the blackhole to the constructor and having it used on L62 and L67 be best?

parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/FileReadBenchmark.java
```diff
@@ -0,0 +1,124 @@
...
+  public String codec;
+
+  @Param({"PARQUET_1_0", "PARQUET_2_0"})
+  public String writerVersion;
+
+  private File tempFile;
+
+  @Setup(Level.Trial)
+  public void setup() throws IOException {
+    tempFile = File.createTempFile("parquet-read-bench-", ".parquet");
```
steveloughran, line 84: there's a constant `BenchmarkFiles.TARGET_DIR` which defines the dir for benchmarks; it puts them under target/ so maven will clean them up. I used that in my PR so killing a test run in my IDE didn't leave cruft around...I'd recommend the same.

steveloughran (commented):
commented. I'm only just learning effective jmh myself; these all LGTM. The one request is that temp files are put under target/ so that a `./run.sh all` command puts everything in the existing temporary directory tree

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
```diff
@@ -117,6 +117,59 @@ public long readLong() {
     }
   }
 
+  @Override
+  public void readIntegers(int[] dest, int offset, int count) {
+    try {
+      // Batch-decode dictionary IDs, then batch-lookup
+      int[] ids = new int[count];
```
steveloughran, line 124: You could havve a lambda expression to catch and translate all of these, as per org.apache.hadoop.util.functional.FunctionalIO 
If you are happy with UncheckedIOException, you can use that as is, and submit a PR saying "this shouldn't be private", which is shouldn't be.

steveloughran (changes requested):

iemejia:
Closing — the JMH benchmarks and shaded jar fix from this PR have been distributed into the individual encoding/compression PRs, each of which now includes its own benchmarks:

- #3565 — PLAIN (`PlainEncodingBenchmark`, `PlainDecodingBenchmark`)
- #3566 — DICTIONARY (`DictionaryEncodingBenchmark`, `DictionaryDecodingBenchmark`)
- #3567 — DELTA (`DeltaBinaryPackedEncodingBenchmark`, `DeltaBinaryPackedDecodingBenchmark`, `DeltaByteArrayEncodingBenchmark`, `DeltaByteArrayDecodingBenchmark`, `DeltaLengthByteArrayEncodingBenchmark`, `DeltaLengthByteArrayDecodingBenchmark`, `LongDeltaDecodingBenchmark`)
- #3568 — RLE (`RleEncodingBenchmark`, `RleDecodingBenchmark`, `RleDictionaryIndexDecodingBenchmark`)
- #3569 — BYTE_STREAM_SPLIT (`BssEncodingBenchmark`, `BssDecodingBenchmark`)
- #3570 — Compression (`CompressionBenchmark`, `CpuReadBenchmark`, `CpuWriteBenchmark`, `FileReadBenchmark`, `FileWriteBenchmark`, `ConcurrentReadWriteBenchmark`)
- #3571 — Row Group Flush (`RowGroupFlushBenchmark`)

Apologies for the churn. Thank you.

PR #3514, opened 2026-04-20
GH-3513: Optimize dictionary writers with OpenHashMap + ArrayList (up to ~70x encodeDictionary)

### LINKED ISSUE #3513: Optimize dictionary writers by replacing fastutil Linked maps with OpenHashMap + ArrayList
## Background

The dictionary writers in `DictionaryValuesWriter` (binary, long, double, float, int variants) use fastutil's `*2IntLinkedOpenHashMap` to assign dictionary indices to distinct values. The "Linked" variant is used because the dictionary page must be emitted in **insertion order** (dictionary index `i` corresponds to the `i`-th distinct value seen), and the linked map preserves that order via a doubly-linked list threaded through the slot array.

This is correct, but the linked variant pays an avoidable cost on every insertion:

- Two extra `long` fields per slot (`prev`, `next` indices) → larger slots, more cache lines per probe
- Pointer fix-up on insert: update the previous tail's `next`, the new entry's `prev`, and the `lastEntry` pointer (3–4 scattered writes per insert)
- Re-stitching the linked list on rehash
- Pure pointer chasing — not vectorizable, not branch-friendly

For high-cardinality columns (hundreds of thousands of distinct values per column chunk), this overhead compounds on a hot path.

## Proposal

Replace each `*2IntLinkedOpenHashMap` with the plain `*2IntOpenHashMap` plus a separate primitive-typed list (`IntArrayList` / `LongArrayList` / `FloatArrayList` / `DoubleArrayList` / `ArrayList<Binary>`) to track insertion order:

- The hash map becomes a pure "have I seen this? what's its id?" lookup with the smallest possible slot.
- The list is append-only, contiguous, and cache-friendly to iterate at flush time.

The two responsibilities (lookup vs ordering) were jammed into one structure; splitting them lets each be optimal. The trade-off is one extra primitive entry per distinct value, which is small compared to the per-slot overhead the linked map was paying.

Note: both candidates are fastutil primitive-keyed maps, so this is **not** a boxing change. The win is structural — eliminating an ordering guarantee that was being paid for on every insert when an explicit append-only list provides it more cheaply.

## Expected impact

From local benchmark runs (`BinaryEncodingBenchmark.encodeDictionary`, `IntEncodingBenchmark.encodeDictionary` — being added in #3512):

- `BinaryEncodingBenchmark.encodeDictionary` (high cardinality, short strings): **+23% to +42%** throughput
- `IntEncodingBenchmark.encodeDictionary` (high cardinality): **~+2x** in some configurations
- Low-cardinality cases are flat or trivially different (the linked-list overhead doesn't matter when there are few inserts)

## Files affected

- `parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java` — the 5 inner classes `PlainBinaryDictionaryValuesWriter`, `PlainLongDictionaryValuesWriter`, `PlainDoubleDictionaryValuesWriter`, `PlainFloatDictionaryValuesWriter`, `PlainIntegerDictionaryValuesWriter`.

No public API change. No file format change. Behavior is identical (dictionary pages emit values in the same order).

iemejia:
## Summary

Resolves #3513.

Replaces fastutil's `*2IntLinkedOpenHashMap` with `*2IntOpenHashMap` plus a separate primitive-typed list (`IntArrayList` / `LongArrayList` / `FloatArrayList` / `DoubleArrayList` / `ArrayList<Binary>`) in the five dictionary writers.

## Why

The dictionary page must be emitted in **insertion order** (dictionary index `i` = `i`-th distinct value seen). The Linked variant provides this via a doubly-linked list threaded through the slot array. That guarantee is paid for on every `put`:

- 2 extra `long` fields per slot (`prev`, `next`) → larger slot footprint, more cache lines per probe
- 3–4 scattered writes per insert to fix up the doubly-linked list
- Re-stitching the linked list on rehash
- Pure pointer chasing — not vectorizable, not branch-friendly

For high-cardinality columns (hundreds of thousands of distinct values per chunk), this overhead compounds on a hot path.

After this change, the hash map is a pure "have I seen this? what's its id?" lookup with the smallest possible slot, and the list is append-only / contiguous / cache-friendly to iterate at flush time. The two responsibilities (lookup vs ordering) were jammed into one structure; splitting them lets each be optimal.

Both candidates are fastutil **primitive-keyed** maps, so this is **not** a boxing change. The win is purely structural.

## Benchmark results

From `BinaryEncodingBenchmark.encodeDictionary` and `IntEncodingBenchmark.encodeDictionary` (added in #3512):

| Benchmark | Configuration | master | this PR | speedup |
|---|---|---:|---:|---:|
| `BinaryEncodingBenchmark.encodeDictionary` | LOW card, len=1000 | 3.3291 µs | 0.0469 µs | **70.97x** |
| `BinaryEncodingBenchmark.encodeDictionary` | LOW card, len=100 | 0.3416 µs | 0.0545 µs | **6.26x** |
| `BinaryEncodingBenchmark.encodeDictionary` | HIGH card, len=10 | 1.2991 µs | 0.5116 µs | **2.54x** |
| `BinaryEncodingBenchmark.encodeDictionary` | LOW card, len=10 | 0.0768 µs | 0.0490 µs | **1.57x** |
| `IntEncodingBenchmark.encodeDictionary` | RANDOM | 0.4125 µs | 0.2000 µs | **2.06x** |
| `IntEncodingBenchmark.encodeDictionary` | SEQUENTIAL | 0.4260 µs | 0.2076 µs | **2.05x** |
| `IntEncodingBenchmark.encodeDictionary` | HIGH_CARDINALITY | 0.4217 µs | 0.2069 µs | **2.04x** |

Note: the 70x outlier on LOW cardinality / 1000-char strings is consistent with eliminating linked-list pointer chasing through hash-table slots when many duplicates accumulate at the head.

The Binary cache from #3500 also contributes here; this PR is additive on top of that win.

## How to reproduce

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar \
    'BinaryEncodingBenchmark.encodeDictionary|IntEncodingBenchmark.encodeDictionary' \
    -wi 5 -i 10 -f 3
```

## Validation

## User-facing changes

None. No public API change. No file format change. Dictionary pages emit values in the same order as before.

### Closes #3513

Part of a small series of focused performance PRs from work in [parquet-perf](https://github.com/iemejia/parquet-perf). Previous: #3494, #3496, #3500, #3504, #3506, #3510. Companion benchmarks contribution: #3512.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
@@ -20,23 +20,22 @@ package org.apache.parquet.column.values.dictionary;
 
 import static org.apache.parquet.bytes.BytesInput.concat;
 
-import it.unimi.dsi.fastutil.doubles.Double2IntLinkedOpenHashMap;
 import it.unimi.dsi.fastutil.doubles.Double2IntMap;
-import it.unimi.dsi.fastutil.doubles.DoubleIterator;
-import it.unimi.dsi.fastutil.floats.Float2IntLinkedOpenHashMap;
+import it.unimi.dsi.fastutil.doubles.Double2IntOpenHashMap;
+import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
 import it.unimi.dsi.fastutil.floats.Float2IntMap;
-import it.unimi.dsi.fastutil.floats.FloatIterator;
-import it.unimi.dsi.fastutil.ints.Int2IntLinkedOpenHashMap;
+import it.unimi.dsi.fastutil.floats.Float2IntOpenHashMap;
+import it.unimi.dsi.fastutil.floats.FloatArrayList;
 import it.unimi.dsi.fastutil.ints.Int2IntMap;
-import it.unimi.dsi.fastutil.longs.Long2IntLinkedOpenHashMap;
+import it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.longs.Long2IntMap;
-import it.unimi.dsi.fastutil.longs.LongIterator;
-import it.unimi.dsi.fastutil.objects.Object2IntLinkedOpenHashMap;
+import it.unimi.dsi.fastutil.longs.Long2IntOpenHashMap;
+import it.unimi.dsi.fastutil.longs.LongArrayList;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
-import it.unimi.dsi.fastutil.objects.ObjectIterator;
+import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import org.apache.parquet.bytes.ByteBufferAllocator;
 import org.apache.parquet.bytes.BytesInput;
@@ -231,7 +230,8 @@ public abstract class DictionaryValuesWriter extends ValuesWriter implements Req
   public static class PlainBinaryDictionaryValuesWriter extends DictionaryValuesWriter {
 
     /* type specific dictionary content */
-    protected Object2IntMap<Binary> binaryDictionaryContent = new Object2IntLinkedOpenHashMap<>();
+    protected Object2IntMap<Binary> binaryDictionaryContent = new Object2IntOpenHashMap<>();
+    protected List<Binary> dictionaryValues = new ArrayList<>();
 
     public PlainBinaryDictionaryValuesWriter(
         int maxDictionaryByteSize,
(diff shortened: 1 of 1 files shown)


Fokko (approved):
This looks like a good change to me.

BTW, the link to https://github.com/iemejia/parquet-perf is broken

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesWriter.java
```diff
@@ -246,8 +246,10 @@ public PlainBinaryDictionaryValuesWriter(
     public void writeBytes(Binary v) {
       int id = binaryDictionaryContent.getInt(v);
       if (id == -1) {
-        id = binaryDictionaryContent.size();
-        binaryDictionaryContent.put(v.copy(), id);
+        id = dictionaryValues.size();
```
Fokko, line 249: Looking at it closer, shouldn't be this `binaryDictionaryContent.size()`?

iemejia: @Fokko They should be equivalent because the map and ordered values list are updated and cleared together. I used dictionaryValues.size() because the id is the index into dictionaryValues, which is what dictionary-page writing and fallback use. Using binaryDictionaryContent.size() would also work, but this makes the indexing relationship explicit.

iemejia:
Closing in favor of #3566.

PR #3517, opened 2026-04-21
GH-3516: Optimize DeltaByteArrayWriter and DeltaLengthByteArrayValuesWriter

iemejia:
## Summary

Resolves #3516.

Two related changes in the `DELTA_BYTE_ARRAY` write path:

### 1. `DeltaLengthByteArrayValuesWriter`: drop the unused `LittleEndianDataOutputStream` wrapper

The class wrapped its `CapacityByteArrayOutputStream` with a `LittleEndianDataOutputStream` that was only used by `Binary.writeTo()` — an extra layer of dispatch on every value that never used any LE-specific functionality (`writeInt`/`writeLong`/etc.). `Binary.writeTo(arrayOut)` works directly with the underlying stream.

Also adds a new overload:

```java
public void writeBytes(byte[] data, int offset, int length) {
  lengthWriter.writeInteger(length);
  arrayOut.write(data, offset, length);
}
```

so callers that already have the raw bytes can avoid allocating a `Binary` wrapper.

### 2. `DeltaByteArrayWriter`: eliminate per-value `Binary.slice()` allocation in the suffix path

Tightens the `suffixWriter` field type from `ValuesWriter` to `DeltaLengthByteArrayValuesWriter` (it's always constructed as one) so the new raw-bytes overload is callable. The suffix call becomes:

```java
suffixWriter.writeBytes(vb, i, vb.length - i);
```

instead of `suffixWriter.writeBytes(v.slice(i, vb.length - i))`, eliminating the `ByteArraySliceBackedBinary` allocation per value plus a layer of virtual dispatch.

## Benchmark results

From `BinaryEncodingBenchmark.encodeDeltaByteArray` / `encodeDeltaLengthByteArray` (added in #3512):

| Benchmark | Configuration | master | this PR | speedup |
|---|---|---:|---:|---:|
| `encodeDeltaByteArray` | LOW card, len=10 | 0.1028 µs | 0.0662 µs | **1.55x** |
| `encodeDeltaByteArray` | HIGH card, len=10 | 0.1704 µs | 0.1124 µs | **1.52x** |
| `encodeDeltaByteArray` | LOW card, len=100 | 0.2079 µs | 0.1678 µs | **1.24x** |
| `encodeDeltaLengthByteArray` | LOW card, len=10 | 0.0481 µs | 0.0397 µs | **1.21x** |
| `encodeDeltaLengthByteArray` | LOW card, len=100 | 0.1503 µs | 0.1374 µs | **1.09x** |

[...]

PR #3519, opened 2026-04-21
GH-3518: Bulk write in LittleEndianDataOutputStream.writeInt/writeShort (+35% encodePlain when used)

### LINKED ISSUE #3518: Optimize LittleEndianDataOutputStream.writeInt/writeShort with bulk write (resolves existing TODO)
## Background

`LittleEndianDataOutputStream.writeInt(int)` and `writeShort(int)` decompose the value byte-by-byte and call `out.write(int)` for each byte:

```java
public final void writeInt(int v) throws IOException {
  out.write((v >>> 0) & 0xFF);
  out.write((v >>> 8) & 0xFF);
  out.write((v >>> 16) & 0xFF);
  out.write((v >>> 24) & 0xFF);
}
```

When the underlying stream is `CapacityByteArrayOutputStream` (the typical case in Parquet writers), each `out.write(int)` performs a `hasRemaining` check, a `Math.addExact` for the new size, possibly a slab-grow check, and a single-byte store. For `writeInt`, that's **4 separate trips** through the bookkeeping.

The class already has the right pattern in `writeLong`: build a `writeBuffer[]` and emit a single `out.write(writeBuffer, 0, 8)`. The buffer is even pre-allocated for that purpose. `writeInt` and `writeShort` just don't use it.

There's a `TODO` comment in `writeInt` (lines 147–149 in current master) acknowledging this:

```
// TODO: see note in LittleEndianDataInputStream: maybe faster
// to use Integer.reverseBytes() and then writeInt, or a ByteBuffer
// approach
```

## Proposal

Extend the existing `writeBuffer[]` pattern to `writeInt` and `writeShort`:

```java
public final void writeInt(int v) throws IOException {
  writeBuffer[0] = (byte) (v >>> 0);
  writeBuffer[1] = (byte) (v >>> 8);
  writeBuffer[2] = (byte) (v >>> 16);
  writeBuffer[3] = (byte) (v >>> 24);
  out.write(writeBuffer, 0, 4);
}
```

This collapses 4 `write(int)` calls into 1 `write(byte[], int, int)` call, cutting the bookkeeping overhead by ~4x per int. Matches the existing `writeLong` pattern in the same file.

Resolves the existing `TODO` in the source.

## Expected impact

Standalone JMH benchmark of the class:
- `IntEncodingBenchmark.encodePlain` (when routed through `LittleEndianDataOutputStream`): **~+35%** (~20.9M → ~28.2M ops/s)

## Note on context

PR #3496 deprecates `LittleEndianDataOutputStream` because Parquet's own writers no longer use it (they write directly into `ByteBuffer`-backed slabs, which compiles to a single intrinsic store on little-endian architectures and is strictly faster than any wrapper).

This PR is therefore a **purely external-caller benefit**: any third-party Parquet-format producer still using the class will get the speedup until they migrate. The change is minimal (~10 lines), obviously correct (matches the existing `writeLong` pattern), and resolves a long-standing `TODO` in the source.

## Files affected

- `parquet-common/src/main/java/org/apache/parquet/bytes/LittleEndianDataOutputStream.java`

No public API change.

iemejia:
## Summary

Resolves #3518.

`LittleEndianDataOutputStream.writeInt(int)` and `writeShort(int)` decompose the value byte-by-byte and call `out.write(int)` for each byte:

```java
public final void writeInt(int v) throws IOException {
  out.write((v >>> 0) & 0xFF);
  out.write((v >>> 8) & 0xFF);
  out.write((v >>> 16) & 0xFF);
  out.write((v >>> 24) & 0xFF);
}
```

When the underlying stream is `CapacityByteArrayOutputStream` (the typical case in Parquet writers), each `out.write(int)` performs a `hasRemaining` check, a `Math.addExact`, possibly a slab-grow check, and a single-byte store. For `writeInt`, that's **4 trips** through the bookkeeping per value.

The class already has the right pattern in `writeLong`: build the `writeBuffer[]` and emit a single `out.write(writeBuffer, 0, 8)`. The buffer is even pre-allocated for that purpose. This PR extends the same pattern to `writeInt` and `writeShort`.

Resolves the long-standing `TODO` in `writeInt`:

```
// TODO: see note in LittleEndianDataInputStream: maybe faster
// to use Integer.reverseBytes() and then writeInt, or a ByteBuffer
// approach
```

## Benchmark

`IntEncodingBenchmark.encodePlain` when routed through `LittleEndianDataOutputStream`:

```
master:  ~20.9M ops/s
this PR: ~28.2M ops/s   (+35%)
```

## Note on context

PR #3496 deprecates `LittleEndianDataOutputStream` because Parquet's own writers no longer use it (they write directly into `ByteBuffer`-backed slabs, which compiles to a single intrinsic store on little-endian and is strictly faster than any wrapper).

After #3496 lands, no Parquet code in any module instantiates `LittleEndianDataOutputStream`. This PR therefore benefits **external Parquet-format producers** that still use the class — they get the speedup until they migrate. The change is minimal (~10 lines), obviously correct (matches the existing `writeLong` pattern in the same file), and resolves the existing `TODO`.

If the maintainers prefer to leave a deprecated class untouched, this PR is easy to drop. I'm flagging the option because the change is small enough that it costs almost nothing to land and helps anyone outside the Parquet codebase still on the class.

## Validation

- `parquet-common`: 308 tests pass

## User-facing changes

None. No public API change. Behavior of `writeInt`, `writeShort`, and `writeLong` is identical bit-for-bit.

### Closes #3518

Part of a small series of focused performance PRs from work in [parquet-perf](https://github.com/iemejia/parquet-perf). Previous: #3494, #3496, #3500, #3504, #3506, #3510, #3514, #3517. Companion benchmarks contribution: #3512.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/LittleEndianDataOutputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/LittleEndianDataOutputStream.java
@@ -30,6 +30,7 @@ public class LittleEndianDataOutputStream extends OutputStream {
 
   private static final Logger LOG = LoggerFactory.getLogger(LittleEndianDataOutputStream.class);
   private final OutputStream out;
+  private byte writeBuffer[] = new byte[8];
 
   /**
    * Creates a new data output stream to write data to the specified
@@ -130,8 +131,9 @@ public class LittleEndianDataOutputStream extends OutputStream {
    * @see java.io.FilterOutputStream#out
    */
   public final void writeShort(int v) throws IOException {
-    out.write((v >>> 0) & 0xFF);
-    out.write((v >>> 8) & 0xFF);
+    writeBuffer[0] = (byte) (v >>> 0);
+    writeBuffer[1] = (byte) (v >>> 8);
+    out.write(writeBuffer, 0, 2);
   }
 
   /**
@@ -144,16 +146,12 @@ public class LittleEndianDataOutputStream extends OutputStream {
    * @see java.io.FilterOutputStream#out
    */
   public final void writeInt(int v) throws IOException {
-    // TODO: see note in LittleEndianDataInputStream: maybe faster
-    // to use Integer.reverseBytes() and then writeInt, or a ByteBuffer
-    // approach
-    out.write((v >>> 0) & 0xFF);
-    out.write((v >>> 8) & 0xFF);
-    out.write((v >>> 16) & 0xFF);
-    out.write((v >>> 24) & 0xFF);
+    writeBuffer[0] = (byte) (v >>> 0);
+    writeBuffer[1] = (byte) (v >>> 8);
+    writeBuffer[2] = (byte) (v >>> 16);
+    writeBuffer[3] = (byte) (v >>> 24);
+    out.write(writeBuffer, 0, 4);
   }
-
-  private byte writeBuffer[] = new byte[8];
 
   /**
    * Writes a <code>long</code> to the underlying output stream as eight


iemejia:
Folding this change into #3496 (the PR that deprecates LittleEndianDataOutputStream) since both touch the same class and are best reviewed together.

PR #3523, opened 2026-04-21
GH-3522: Reuse intermediate buffers in RunLengthBitPackingHybridDecoder PACKED path (~22% throughput on dictionary-id decode)

### LINKED ISSUE #3522: Reuse intermediate buffers in RunLengthBitPackingHybridDecoder PACKED path (~22% throughput on dictionary-id decode)
## Description

`RunLengthBitPackingHybridDecoder` allocates two new arrays per PACKED run during decode:

```java
case PACKED:
    int numGroups = header >>> 1;
    currentCount = numGroups * 8;
    currentBuffer = new int[currentCount]; // TODO: reuse a buffer
    byte[] bytes = new byte[numGroups * bitWidth];
    ...
```

The TODO comment in the code itself flags this as a known improvement opportunity. Dictionary-id pages encoded with the RLE/bit-packed hybrid encoder produce many PACKED runs per page, so each `readInt()` traversal of a page allocates O(numGroups) intermediate arrays that are immediately discarded — pure GC pressure on a hot read path.

## Proposed change

Reuse the two buffers across PACKED runs within the same decoder instance, growing them only when a larger run is encountered:

- Add `int[] packedValuesBuffer` and `byte[] packedBytesBuffer` instance fields, initialized to `new int[0]` / `new byte[0]`.
- In the PACKED branch, grow each buffer only when its length is below the required size.
- Add `currentBufferLength` to track the logical length of the active region (since `packedValuesBuffer.length` may now exceed the current run).

This matches the pattern used by other decoders in the codebase that amortize buffer allocations.

## Benchmark

A new `RleDictionaryIndexDecodingBenchmark` (added in #3511 / #3512) isolates the RLE/bit-packed dictionary-id decode path. Three index patterns over 100k INT32 values, BIT_WIDTH=10, JMH `-wi 5 -i 10 -f 2` (20 measurement iterations):

| Pattern         | master (ops/s) | optimized (ops/s) | Improvement |
|-----------------|---------------:|------------------:|:-----------:|
| SEQUENTIAL      |     93,061,521 |       113,856,860 | **+22.3%**  |
| RANDOM          |     92,929,824 |       114,238,638 | **+22.9%**  |
| LOW_CARDINALITY |     92,813,229 |       115,271,347 | **+24.2%**  |

End-to-end file read (`FileReadBenchmark`) sees a much smaller ~2% improvement because RLE decoding is only one of many pipeline stages, but it confirms the signal is real.

## Validation

- 573 `parquet-column` tests pass
- 9 `TestRunLengthBitPackingHybridEncoder` tests pass (these round-trip through the decoder)

## Scope

~22 LOC change to a single file (`RunLengthBitPackingHybridDecoder.java`). Self-contained and obviously correct (resolves the existing TODO).

## Related

Part of the focused performance PR series from https://github.com/iemejia/parquet-perf. The companion ByteStreamSplit writer/reader changes from the same source commit have already been submitted as #3504 (writer) and #3506 (reader).

iemejia:
## Summary

Closes #3522.

`RunLengthBitPackingHybridDecoder` allocates a new `int[]` and `byte[]` on every PACKED run during decode. The code itself flagged this with a `// TODO: reuse a buffer` comment. This PR resolves the TODO by reusing the buffers across runs within the same decoder instance, growing them lazily only when a larger run is encountered.

Also adds a `currentBufferLength` field to track the logical active-region length in `packedValuesBuffer` (since `packedValuesBuffer.length` may now exceed the current run's size after a prior larger run grew it).

## Benchmark

`RleDictionaryIndexDecodingBenchmark` (added in #3512) isolates the RLE/bit-packed dictionary-id decode path. 100k INT32 dictionary IDs, BIT_WIDTH=10, JMH `-wi 5 -i 10 -f 2` (20 measurement iterations):

| Pattern         | master (ops/s) | optimized (ops/s) | Improvement |
|-----------------|---------------:|------------------:|:-----------:|
| SEQUENTIAL      |     93,061,521 |       113,856,860 | **+22.3%**  |
| RANDOM          |     92,929,824 |       114,238,638 | **+22.9%**  |
| LOW_CARDINALITY |     92,813,229 |       115,271,347 | **+24.2%**  |

End-to-end `FileReadBenchmark` sees a much smaller ~2% improvement because RLE decoding is only one of many pipeline stages; the isolated micro-benchmark shows the true magnitude on the affected code path.

## Validation

- `TestRunLengthBitPackingHybridEncoder`: 9 tests pass (these round-trip values through the decoder)

## Scope

17 LOC change to a single file. Self-contained and obviously correct (resolves the existing TODO).

## Related

Part of the focused performance PR series from https://github.com/iemejia/parquet-perf. The companion ByteStreamSplit writer/reader changes from the same source commit (`ba52f82c3`) have already been submitted as #3504 and #3506.

## How to reproduce

The benchmark is added in #3512. Once that lands, reproduce with:

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar \
    'RleDictionaryIndexDecodingBenchmark' -wi 5 -i 10 -f 2
```

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
@@ -18,9 +18,8 @@
  */
 package org.apache.parquet.column.values.rle;
 
-import java.io.DataInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.bitpacking.BytePacker;
@@ -42,7 +41,7 @@ public class RunLengthBitPackingHybridDecoder {
 
   private final int bitWidth;
   private final BytePacker packer;
-  private final InputStream in;
+  private final ByteBuffer buffer;
 
   private MODE mode;
   private int currentCount;
@@ -54,16 +53,22 @@ public class RunLengthBitPackingHybridDecoder {
   private int[] packedValuesBuffer = new int[0];
   private byte[] packedBytesBuffer = new byte[0];
 
-  public RunLengthBitPackingHybridDecoder(int bitWidth, InputStream in) {
+  public RunLengthBitPackingHybridDecoder(int bitWidth, ByteBuffer buffer) {
     LOG.debug("decoding bitWidth {}", bitWidth);
 
     Preconditions.checkArgument(bitWidth >= 0 && bitWidth <= 32, "bitWidth must be >= 0 and <= 32");
     this.bitWidth = bitWidth;
     this.packer = Packer.LITTLE_ENDIAN.newBytePacker(bitWidth);
-    this.in = in;
+    this.buffer = buffer.order(ByteOrder.LITTLE_ENDIAN);
   }
 
-  public int readInt() throws IOException {
+  /**
+   * Reads the next int value from the RLE/Bit-Packing hybrid stream.
+   *
+   * @return the next decoded integer value
+   * @throws ParquetDecodingException if a decoding error occurs
+   */
+  public int readInt() {
     if (currentCount == 0) {
       readNext();
     }
@@ -82,15 +87,15 @@ public class RunLengthBitPackingHybridDecoder {
     return result;
   }
 
-  private void readNext() throws IOException {
-    Preconditions.checkArgument(in.available() > 0, "Reading past RLE/BitPacking stream.");
-    final int header = BytesUtils.readUnsignedVarInt(in);
+  private void readNext() {
+    Preconditions.checkArgument(buffer.hasRemaining(), "Reading past RLE/BitPacking stream.");
+    final int header = BytesUtils.readUnsignedVarInt(buffer);
     mode = (header & 1) == 0 ? MODE.RLE : MODE.PACKED;
     switch (mode) {
       case RLE:
         currentCount = header >>> 1;
         LOG.debug("reading {} values RLE", currentCount);
-        currentValue = BytesUtils.readIntLittleEndianPaddedOnBitWidth(in, bitWidth);
+        currentValue = BytesUtils.readIntLittleEndianPaddedOnBitWidth(buffer, bitWidth);
         break;
       case PACKED:
         int numGroups = header >>> 1;
@@ -107,8 +112,8 @@ public class RunLengthBitPackingHybridDecoder {
         }
         // At the end of the file RLE data though, there might not be that many bytes left.
         int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
-        bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(packedBytesBuffer, 0, bytesToRead);
+        bytesToRead = Math.min(bytesToRead, buffer.remaining());
+        buffer.get(packedBytesBuffer, 0, bytesToRead);
         for (int valueIndex = 0, byteIndex = 0;
             valueIndex < currentCount;
             valueIndex += 8, byteIndex += bitWidth) {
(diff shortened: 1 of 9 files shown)


arouel (commented):
Have you seen the PR #3467 which addresses the same issue? I would prefer if we help each other in reviewing instead of creating more PRs of the same kind.

parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -90,17 +95,24 @@ private void readNext() throws IOException {
...
+        currentBuffer = packedValuesBuffer;
+        int bytesRequired = numGroups * bitWidth;
+        if (packedBytesBuffer.length < bytesRequired) {
+          packedBytesBuffer = new byte[bytesRequired];
+        }
         // At the end of the file RLE data though, there might not be that many bytes left.
         int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
         bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(bytes, 0, bytesToRead);
+        new DataInputStream(in).readFully(packedBytesBuffer, 0, bytesToRead);
```
Fokko: I think it is on purpose to not close this resource, should we add a comment to it?

parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -90,17 +95,24 @@ private void readNext() throws IOException {
...
         for (int valueIndex = 0, byteIndex = 0;
             valueIndex < currentCount;
             valueIndex += 8, byteIndex += bitWidth) {
-          packer.unpack8Values(bytes, byteIndex, currentBuffer, valueIndex);
+          packer.unpack8Values(packedBytesBuffer, byteIndex, currentBuffer, valueIndex);
```
Fokko: I noticed that `unpack8Values(final byte[] input, ...)` is deprecated, should we switch to the `ByteBuffer` overload while we're reworking this?

```
diff --git a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
index da9db00c7..14ec1a87d 100644
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
@@ -21,6 +21,7 @@ package org.apache.parquet.column.values.rle;
 import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.bitpacking.BytePacker;
@@ -52,7 +53,7 @@ public class RunLengthBitPackingHybridDecoder {
 
   // Reusable buffers to avoid per-run allocation in PACKED mode
   private int[] packedValuesBuffer = new int[0];
-  private byte[] packedBytesBuffer = new byte[0];
+  private ByteBuffer packedBytesBuffer = ByteBuffer.allocate(0);
 
   public RunLengthBitPackingHybridDecoder(int bitWidth, InputStream in) {
     LOG.debug("decoding bitWidth {}", bitWidth);
@@ -102,13 +103,13 @@ public class RunLengthBitPackingHybridDecoder {
         }
         currentBuffer = packedValuesBuffer;
         int bytesRequired = numGroups * bitWidth;
-        if (packedBytesBuffer.length < bytesRequired) {
-          packedBytesBuffer = new byte[bytesRequired];
+        if (packedBytesBuffer.capacity() < bytesRequired) {
+          packedBytesBuffer = ByteBuffer.allocate(bytesRequired);
         }
         // At the end of the file RLE data though, there might not be that many bytes left.
         int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
         bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(packedBytesBuffer, 0, bytesToRead);
+        new DataInputStream(in).readFully(packedBytesBuffer.array(), 0, bytesToRead);
         for (int valueIndex = 0, byteIndex = 0;
             valueIndex < currentCount;
             valueIndex += 8, byteIndex += bitWidth) {
```

Fokko (approved):
Thanks for fixing this outstanding TODO @iemejia 🙌 

I left one comment because I think it would good to update to the non-deprecated API of parquet-encoding while at it. WDYT?

iemejia:
Closing in favor of #3568.

Apologies for the churn. If you've been reviewing this PR, please continue the discussion on #3568 which supersedes it. Thank you.

PR #3524, opened 2026-04-21
PARQUET-3364: Allow hyphens in column names in AvroSchemaConverter

### LINKED ISSUE #3364: [parquet-cli] Illegal character in column name

I have a Parquet file with a column that contains a hyphen character and get the following error:

> Illegal character in: Creation-Time

I guess Parquet officially supports only alphanumeric plus underscore, but I have retrieved this Parquet file from a [blob inventory scan](https://docs.azure.cn/en-us/storage/blobs/blob-inventory-how-to) on Azure.

Might there be an option to allow such a column name in `parquet cat` for example?

### Component(s)

CLI

yadavay-amzn:
### What changes were proposed?

Widen the catch in `CatCommand.run()` from `SchemaParseException` to `RuntimeException` so that Avro-related failures (including name validation errors during record reading) fall back to the `GroupReader` path, which handles non-standard field names correctly.

The `parquet cat` CLI command fails on valid Parquet files with hyphenated column names (e.g., `Creation-Time`). The Avro path throws because Avro rejects hyphens in field names. The `GroupReader` fallback already handles these files correctly, but the catch was too narrow (`SchemaParseException` only) to trigger the fallback when the failure occurs during record reading rather than schema conversion.

### How was this tested?

The existing `CatCommandTest.testCatCommandWithHyphenatedFieldNames` test validates this fix — it creates a Parquet file with hyphenated field names and verifies `parquet cat` reads it successfully.

Closes #3364

### DIFF
--- a/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
+++ b/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
@@ -351,11 +351,38 @@ public class AvroSchemaConverter {
   }
 
   public Schema convert(MessageType parquetSchema) {
-    return convertFields(parquetSchema.getName(), parquetSchema.getFields(), new HashMap<>());
+    return withDisabledNameValidation(
+        () -> convertFields(parquetSchema.getName(), parquetSchema.getFields(), new HashMap<>()));
   }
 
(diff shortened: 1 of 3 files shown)


yadavay-amzn:
I would like to work on this. Will submit a PR shortly.

Issue #3261, opened 2025-08-05
`integer overflow` in CapacityByteArrayOutputStream


manuzhang:
### Describe the bug, including details regarding any error messages, version, and platform.

The following exception was thrown when we read a column of `ARRAY<STRING>` in Spark 3.5.0 and Parquet 1.15.2

```
Caused by: java.lang.ArithmeticException: integer overflow
	at java.base/java.lang.Math.addExact(Math.java:883)
	at org.apache.parquet.bytes.CapacityByteArrayOutputStream.addSlab(CapacityByteArrayOutputStream.java:198)
	at org.apache.parquet.bytes.CapacityByteArrayOutputStream.write(CapacityByteArrayOutputStream.java:220)
	at org.apache.parquet.bytes.LittleEndianDataOutputStream.write(LittleEndianDataOutputStream.java:76)
	at java.base/java.io.OutputStream.write(OutputStream.java:127)
	at org.apache.parquet.io.api.Binary$ByteArrayBackedBinary.writeTo(Binary.java:319)
	at org.apache.parquet.column.values.plain.PlainValuesWriter.writeBytes(PlainValuesWriter.java:55)
	at org.apache.parquet.column.values.fallback.FallbackValuesWriter.writeBytes(FallbackValuesWriter.java:178)
	at org.apache.parquet.column.impl.ColumnWriterBase.write(ColumnWriterBase.java:196)
	at org.apache.parquet.io.MessageColumnIO$MessageColumnIORecordConsumer.addBinary(MessageColumnIO.java:473)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.$anonfun$makeWriter$9(ParquetWriteSupport.scala:212)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.$anonfun$makeWriter$9$adapted(ParquetWriteSupport.scala:210)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.$anonfun$makeArrayWriter$5(ParquetWriteSupport.scala:354)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.consumeField(ParquetWriteSupport.scala:490)
	at org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport.$anonfun$makeArrayWriter$4(ParquetWriteSupport.scala:354)
[...]

PR #3525, opened 2026-04-22
GH-3261: Fix integer overflow in `CapacityByteArrayOutputStream`

yadavay-amzn:
## What changes were made?

Fix integer overflow in `CapacityByteArrayOutputStream.addSlab()` that causes `ArithmeticException` when writing large `ARRAY<STRING>` columns (issue #3261).

### Root cause (identified by @Kimahriman)

The overflow check in `addSlab()` used `bytesUsed` to detect overflow, but `bytesUsed` is not updated until *after* `addSlab()` returns in `write()`. Additionally, `nextSlabSize` can be larger than `minimumSize` (due to the doubling strategy), so checking only `bytesUsed + minimumSize` was insufficient.

This meant `bytesAllocated = Math.addExact(this.bytesAllocated, nextSlabSize)` could overflow without being caught by the guard, throwing an uncaught `ArithmeticException` instead of the intended `OutOfMemoryError`.

### Fix

1. **Use `bytesAllocated` instead of `bytesUsed`** for the overflow check — `bytesAllocated` is always up to date when `addSlab()` is called.
2. **Cap `nextSlabSize`** when it would cause `bytesAllocated` to overflow `Integer.MAX_VALUE`, preventing the uncaught `ArithmeticException` on the `Math.addExact` call.

### Tests

Added `TestCapacityByteArrayOutputStreamOverflow` with two tests:
- Verifies that slab allocation near `Integer.MAX_VALUE` succeeds (previously threw `ArithmeticException`)
- Verifies that a true overflow still throws `OutOfMemoryError` as intended

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/CapacityByteArrayOutputStream.java
@@ -167,9 +167,10 @@ public class CapacityByteArrayOutputStream extends OutputStream {
   private void addSlab(int minimumSize) {
     int nextSlabSize;
 
-    // check for overflow
+    // check for overflow using bytesAllocated which is always up to date (unlike bytesUsed which
+    // is updated after addSlab returns in write())
     try {
-      Math.addExact(bytesUsed, minimumSize);
+      Math.addExact(bytesAllocated, minimumSize);
     } catch (ArithmeticException e) {
       // This is interpreted as a request for a value greater than Integer.MAX_VALUE
       // We throw OOM because that is what java.io.ByteArrayOutputStream also does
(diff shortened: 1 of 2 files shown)


steveloughran (commented):
LGTM

yadavay-amzn:
Friendly ping to @wgtmac @Fokko, this is a small doc/parser change that's been LGTM'd by @steveloughran. Could one of you please take a look when you have a moment? Thanks!

wgtmac (approved):

wgtmac:
Thanks @yadavay-amzn for fixing it and @steveloughran for the review! This makes sense to me.

Fokko (approved):

Fokko:
Let's move this forward, thanks @yadavay-amzn for working on this, and thanks @steveloughran and @wgtmac for the review 🚀

PR #3526, opened 2026-04-23
PARQUET-3459: Per column compression

mengna-lin:
### Rationale for this change
 The Parquet format supports per-column compression at the spec level, but parquet-java has always forced a single codec across all columns —  
  this PR exposes that existing capability.                                                                                                   

  - CompressionCodecFactory — new default getCompressor(codec, level) interface method                                                
  - CodecFactory — level-aware compressor instantiation and validation (ZSTD: -131072–22, GZIP: 0–9 or -1, BROTLI: 0–11)                    
  - ParquetProperties — new builder methods withCompressionCodec(col, codec) / withCompressionLevel(col, level)                       
  - ColumnChunkPageWriteStore — resolves compressor per column at write time, falls back to job-level default                         
  - ParquetWriter — new builder methods withCompressionCodec(col, codec) / withCompressionLevel(col, level)                           
  - ParquetOutputFormat — wires per-column codec/level from Hadoop Configuration into ParquetProperties  

### Are these changes tested?
  Yes.                                                                                                                                
  Unit tests cover ParquetProperties getters/copy behavior (TestParquetProperties),                                                   
  CodecFactory level-aware caching and invalid level rejection (TestDirectCodecFactory),                                              
  and ColumnChunkPageWriteStore codec resolution and invalid level rejection (TestColumnChunkPageWriteStore).                         
  Integration tests cover end-to-end data round-trips and footer metadata verification                                                
  through both the ParquetWriter builder API and ParquetOutputFormat (TestParquetWriter).      
Also test with spark job
```Scala
/**
 * Local Spark job that writes a Parquet file with per-column compression:
 *   col_1 (string) -> ZSTD level 9
 *   col_2 (int)    -> UNCOMPRESSED
 *   col_3 (double) -> SNAPPY (default)
 *
 * Build:
 *   cd test-spark-job && mvn package -DskipTests
 *
 * Run:
 *   spark-submit \
 *     --master local[2] \
 *     target/parquet-per-column-compression-test-1.0-SNAPSHOT-job.jar
 *
 * Inspect output with parquet-cli:
 *   hadoop jar ../parquet-cli/target/parquet-cli-1.18.0-SNAPSHOT-runtime.jar \
 *     org.apache.parquet.cli.Main meta \
 *     /tmp/per_column_compression_test/part-*.parquet
 */
public class PerColumnCompressionJob {

  private static final String OUTPUT_PATH = "/tmp/per_column_compression_test";

  public static void main(String[] args) {
    SparkSession spark = SparkSession.builder()
        .master("local[2]")
        .appName("PerColumnCompressionTest")
        // Default codec for all columns
        .config("spark.hadoop.parquet.compression", "SNAPPY")
        // col_1: ZSTD at level 9
        .config("spark.hadoop.parquet.compression#col_1", "ZSTD")
        .config("spark.hadoop.parquet.compression.level#col_1", "9")
        // col_2: no compression
        .config("spark.hadoop.parquet.compression#col_2", "UNCOMPRESSED")
        .getOrCreate();

    spark.sparkContext().setLogLevel("WARN");

    StructType schema = DataTypes.createStructType(new StructField[]{
        DataTypes.createStructField("col_1", DataTypes.StringType, false),
        DataTypes.createStructField("col_2", DataTypes.IntegerType, false),
        DataTypes.createStructField("col_3", DataTypes.DoubleType, false),
    });

    List<Row> rows = Arrays.asList(
        RowFactory.create("alice",   1, 1.1),
        RowFactory.create("bob",     2, 2.2),
        RowFactory.create("charlie", 3, 3.3),
        RowFactory.create("dave",    4, 4.4),
        RowFactory.create("eve",     5, 5.5)
    );

    Dataset<Row> df = spark.createDataFrame(rows, schema);
    System.out.println("\n=== Input DataFrame ===");
    df.printSchema();
    df.show();

    df.coalesce(1).write().mode("overwrite").parquet(OUTPUT_PATH);
    System.out.println("Wrote Parquet to: " + OUTPUT_PATH);

    System.out.println("\n=== Read Back ===");
    spark.read().parquet(OUTPUT_PATH).show();

    System.out.println("\nDone. Inspect compression with:");
    System.out.println("  hadoop jar ../parquet-cli/target/parquet-cli-1.18.0-SNAPSHOT-runtime.jar \\");
    System.out.println("    org.apache.parquet.cli.Main meta " + OUTPUT_PATH + "/part-*.parquet");

    spark.stop();
  }
}
```

Result
```Shell
mengnalin@Mengnas-MacBook-Pro apache-parquet-java-mengna % hadoop jar parquet-cli/target/parquet-cli-1.18.0-SNAPSHOT-runtime.jar meta /tmp/per_column_compression_test/part-*.parquet

File path:  /tmp/per_column_compression_test/part-00000-017e9683-14be-4539-85ce-92f3904a744d-c000.snappy.parquet
Created by: *
Properties:
                   org.apache.spark.version: 4.1.1
  org.apache.spark.sql.parquet.row.metadata: {"type":"struct","fields":[{"name":"col_1","type":"string","nullable":false,"metadata":{}},{"name":"col_2","type":"integer","nullable":false,"metadata":{}},{"name":"col_3","type":"double","nullable":false,"metadata":{}}]}
Schema:
message spark_schema {
  required binary col_1 (STRING);
  required int32 col_2;
  required double col_3;
}


Row group 0:  count: 5  35.20 B records  start: 4  total(compressed): 176 B total(uncompressed):171 B 
--------------------------------------------------------------------------------
       type      encodings count     avg size   nulls   min / max
col_1  BINARY    Z   _     5         15.40 B    0       "alice" / "eve"
col_2  INT32     _   _     5         8.60 B     0       "1" / "5"
col_3  DOUBLE    S   _     5         11.20 B    0       "1.1" / "5.5"
```

  Two new APIs, fully backwards compatible:
     - ParquetWriter.Builder.withCompressionCodec(col, codec)                                                                         
     - ParquetWriter.Builder.withCompressionLevel(col, level)                                                                         
     (Also accessible at the lower level via ParquetProperties.Builder.)

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -39,6 +39,7 @@ import org.apache.parquet.column.values.factory.DefaultValuesWriterFactory;
 import org.apache.parquet.column.values.factory.ValuesWriterFactory;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWriter;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.schema.MessageType;
 
 /**
@@ -135,6 +136,8 @@ public class ParquetProperties {
   private final Map<String, String> extraMetaData;
   private final ColumnProperty<Boolean> statistics;
   private final ColumnProperty<Boolean> sizeStatistics;
+  private final ColumnProperty<CompressionCodecName> columnCodecs;
+  private final ColumnProperty<Integer> columnCompressionLevels;
 
   private ParquetProperties(Builder builder) {
     this.pageSizeThreshold = builder.pageSize;
@@ -167,6 +170,8 @@ public class ParquetProperties {
     this.extraMetaData = builder.extraMetaData;
     this.statistics = builder.statistics.build();
     this.sizeStatistics = builder.sizeStatistics.build();
+    this.columnCodecs = builder.columnCodecs.build();
+    this.columnCompressionLevels = builder.columnCompressionLevels.build();
   }
 
   public static Builder builder() {
@@ -370,6 +375,28 @@ public class ParquetProperties {
     return sizeStatisticsEnabled;
   }
 
+  /**
+   * Returns the compression codec configured for the given column, or {@code null} if no
+   * column-specific codec has been set (the caller should fall back to the job-level codec).
+   *
+   * @param column the column descriptor
+   * @return the per-column codec, or {@code null} if not set
+   */
+  public CompressionCodecName getColumnCodec(ColumnDescriptor column) {
+    return columnCodecs.getValue(column);
+  }
+
+  /**
+   * Returns the compression level configured for the given column, or {@code null} if no
+   * column-specific level has been set.
+   *
+   * @param column the column descriptor
+   * @return the per-column compression level, or {@code null} if not set
+   */
+  public Integer getColumnCompressionLevel(ColumnDescriptor column) {
+    return columnCompressionLevels.getValue(column);
+  }
+
   @Override
   public String toString() {
     return "Parquet page size to " + getPageSizeThreshold() + '\n'
@@ -388,7 +415,9 @@ public class ParquetProperties {
         + "Page row count limit to " + getPageRowCountLimit() + '\n'
         + "Writing page checksums is: " + (getPageWriteChecksumEnabled() ? "on" : "off") + '\n'
         + "Statistics enabled: " + statisticsEnabled + '\n'
-        + "Size statistics enabled: " + sizeStatisticsEnabled;
+        + "Size statistics enabled: " + sizeStatisticsEnabled + '\n'
+        + "Per-column codecs: " + columnCodecs + '\n'
+        + "Per-column compression levels: " + columnCompressionLevels;
   }
 
   public static class Builder {
@@ -419,6 +448,8 @@ public class ParquetProperties {
     private Map<String, String> extraMetaData = new HashMap<>();
     private final ColumnProperty.Builder<Boolean> statistics;
     private final ColumnProperty.Builder<Boolean> sizeStatistics;
+    private final ColumnProperty.Builder<CompressionCodecName> columnCodecs;
+    private final ColumnProperty.Builder<Integer> columnCompressionLevels;
 
     private Builder() {
       enableDict = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_IS_DICTIONARY_ENABLED);
@@ -436,6 +467,8 @@ public class ParquetProperties {
           ColumnProperty.<Integer>builder().withDefaultValue(DEFAULT_BLOOM_FILTER_CANDIDATES_NUMBER);
       statistics = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_STATISTICS_ENABLED);
       sizeStatistics = ColumnProperty.<Boolean>builder().withDefaultValue(DEFAULT_SIZE_STATISTICS_ENABLED);
+      columnCodecs = ColumnProperty.<CompressionCodecName>builder().withDefaultValue(null);
+      columnCompressionLevels = ColumnProperty.<Integer>builder().withDefaultValue(null);
     }
 
     private Builder(ParquetProperties toCopy) {
@@ -460,6 +493,8 @@ public class ParquetProperties {
       this.extraMetaData = toCopy.extraMetaData;
       this.statistics = ColumnProperty.builder(toCopy.statistics);
       this.sizeStatistics = ColumnProperty.builder(toCopy.sizeStatistics);
+      this.columnCodecs = ColumnProperty.builder(toCopy.columnCodecs);
+      this.columnCompressionLevels = ColumnProperty.builder(toCopy.columnCompressionLevels);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


wgtmac:
Does this duplicate https://github.com/apache/parquet-java/pull/3396?

mengna-lin:
> Does this duplicate #3396?

Hi @wgtmac , thanks for your reviews on this feature — I really appreciate the guidance.
I understand the preference for https://github.com/apache/parquet-java/pull/3396 since it was opened first. I just wanted to gently note it's been inactive ~2 months with several review comments still open. In the meantime, I've done my best to incorporate all of that feedback into this PR.
For context, this is currently a blocker for downstream work in Apache Iceberg https://github.com/apache/iceberg/pull/16094. 
Whenever you have a chance, I'd be grateful if you could take a review pass. Thank you!

wgtmac:
@mengna-lin Thank you for the patience! I think now it makes sense to proceed with your PR. Let me take a look then.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -252,10 +260,22 @@ public CompressionCodecName getCodecName() {
...
     if (comp == null) {
       comp = createCompressor(codecName);
-      compressors.put(codecName, comp);
+      compressors.put(key, comp);
+    }
+    return comp;
+  }
+
+  @Override
+  public BytesCompressor getCompressor(CompressionCodecName codecName, int level) {
```
wgtmac, line 274: This path bypasses `DirectCodecFactory#createCompressor(...)`. With a direct codec factory, per-column levels for ZSTD/SNAPPY can fall back to the heap/Hadoop compressor path; please add a direct-factory override or test.

(reviewed by Codex)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -29,9 +29,12 @@
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.compress.CodecPool;
 import org.apache.hadoop.io.compress.CompressionCodec;
+import org.slf4j.Logger;
```
wgtmac: These imports are out of Spotless import order; the same pattern appears in a few added test imports too, so `spotless:check` will likely fail.

(reviewed by Codex)

mengna-lin:
@wgtmac I addressed those comments. Please take another look when you get a chance. Thanks.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -388,7 +415,9 @@ public String toString() {
         + "Page row count limit to " + getPageRowCountLimit() + '\n'
         + "Writing page checksums is: " + (getPageWriteChecksumEnabled() ? "on" : "off") + '\n'
         + "Statistics enabled: " + statisticsEnabled + '\n'
-        + "Size statistics enabled: " + sizeStatisticsEnabled;
+        + "Size statistics enabled: " + sizeStatisticsEnabled + '\n'
+        + "Per-column codecs: " + columnCodecs + '\n'
```
wgtmac: Does this produce too much noise if not enabled? Perhaps we can shorten them into a single line or even omit them when disabled.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -569,6 +575,32 @@ public SELF withCompressionCodec(CompressionCodecName codecName) {
...
     }
 
+    /**
+     * Override the compression codec for a specific column.
+     *
+     * @param columnPath dot-string path of the column (e.g. {@code "my_col"})
+     * @param codecName  the codec to use for that column
+     * @return this builder for method chaining.
+     */
+    public SELF withCompressionCodec(String columnPath, CompressionCodecName codecName) {
```
wgtmac, line 585: Should we provide a convenience method to set codec name and level in a single shot?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestDirectCodecFactory.java
```diff
@@ -285,4 +290,158 @@ public void cachingKeysZstd() {
     Assert.assertEquals(codec_2_1, codec_2_2);
     Assert.assertNotEquals(codec_2_1, codec_5_1);
   }
+
+  @Test
+  public void levelAwareCompressor_sameLevel_returnsCachedInstance() {
```
wgtmac: These test case names are not consistent with this repo.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -271,15 +291,124 @@ public BytesDecompressor getDecompressor(CompressionCodecName codecName) {
...
+  protected BytesCompressor createCompressorAtLevel(CompressionCodecName codecName, int level) {
+    return compressorForCodec(codecName, getCodecAtLevel(codecName, level));
+  }
+
+  private BytesCompressor compressorForCodec(CompressionCodecName codecName, CompressionCodec codec) {
+    return codec == null ? NO_OP_COMPRESSOR : new HeapBytesCompressor(codecName, codec);
+  }
+
+  static void validateZstdLevel(int level) {
+    if (level < 1 || level > 22) {
```
wgtmac: IIRC, ZSTD supports negative levels to be even faster?

wgtmac (commented):
Sorry for the long delay since I've been busy these days. Mostly look good to me. I've left some minor comments.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
mengna-lin: Thanks for the suggestion! My concern is that codec and level are independent (a user may want to set only
  one), and a combined method alongside the existing setters could make it ambiguous which call wins when both configure the same column. What do you think?

wgtmac: I'm fine with current design. Please rebase to fix ci failure.

parquet-common/src/main/java/org/apache/parquet/compression/CompressionCodecFactory.java
```diff
@@ -46,6 +46,30 @@ public interface CompressionCodecFactory {
    */
   BytesInputCompressor getCompressor(CompressionCodecName codecName);
 
+  /**
+   * Returns a {@link BytesInputCompressor} instance for the specified codec name and compression level.
+   * <p>
+   * The compression level controls the trade-off between compression speed and ratio. The valid range
+   * and meaning of the level is codec-specific:
+   * <ul>
+   *   <li>ZSTD: 1 (fastest) to 22 (best compression), default 3</li>
```
wgtmac: The ZSTD range here is now broader because `validateZstdLevel` uses the zstd library min/max levels. Please update this wording to avoid documenting `1-22` only.

(reviewed by Codex)

wgtmac (approved):
Looks good to me.

mengna-lin:
@wgtmac I have addressed all comments and all checks have passed. Thanks for the review.

wgtmac:
Thank you for the hard work, @mengna-lin!

PR #3533, opened 2026-05-01
GH-3522: Optimize IntList.size() from O(slabs) to O(1) with running counter

iemejia:
## Summary

- Replace `IntList.size()` slab-iteration with a simple `totalSize` counter incremented on each `add()`
- Eliminates O(slabs) overhead from dictionary encoding hot paths where `size()` is called frequently

## Details

`IntList.size()` was iterating over all slabs to sum their lengths on every call. For delta encoding writers that check size thresholds frequently, this becomes O(slabs) per check. With a running counter it's O(1).

The improvement is primarily relevant for large row groups where the number of slabs grows.

All 576 parquet-column tests pass.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
@@ -99,6 +99,7 @@ public class IntList {
   // not be added
   private int[] currentSlab;
   private int currentSlabPos;
+  private int totalSize;
 
   private void allocateSlab() {
     currentSlab = new int[currentSlabSize];
@@ -129,6 +130,7 @@ public class IntList {
 
     currentSlab[currentSlabPos] = i;
     ++currentSlabPos;
+    ++totalSize;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
Fokko, line 106: Should we also set `totalSize` to 0 here?

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -129,6 +130,7 @@ public void add(int i) {
 
     currentSlab[currentSlabPos] = i;
     ++currentSlabPos;
+    ++totalSize;
```
Fokko, line 133: Looking at the code, totalSize will always be the same as currentSlabSize

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/IntList.java
```diff
@@ -150,11 +152,6 @@ public IntIterator iterator() {
    * @return the current size of the list
    */
   public int size() {
-    int size = currentSlabPos;
-    for (int[] slab : slabs) {
-      size += slab.length;
-    }
```
Fokko, line 156: I don't think this change is correct. We're returning the slab length, and this one can grow as well (double until MAX_SLAB_SIZE)

iemejia:
Closing in favor of #3566.

PR #3535, opened 2026-05-01
GH-3522: Add batch read APIs to ValuesReader hierarchy

iemejia:
## Summary

- Add `readIntegers()`, `readLongs()`, `readFloats()`, `readDoubles()` batch methods to `ValuesReader` with default loop-based implementations
- Override in specialized readers to amortize per-value overhead across batches

## Overrides

- **RunLengthBitPackingHybridDecoder.readInts()**: batch across RLE runs and packed groups using `Arrays.fill`/`System.arraycopy`
- **DictionaryValuesReader**: batch-decode dictionary IDs first, then batch-lookup values (eliminates per-value IOException try/catch)
- **DeltaBinaryPackingValuesReader**: `System.arraycopy` from pre-decoded buffer
- **PlainValuesReader** (all types): loop over LittleEndianDataInputStream
- **ByteStreamSplitValuesReader** (all types): indexed ByteBuffer bulk read

## Rationale

These APIs enable callers to amortize per-value overhead (virtual dispatch, bounds checks, mode switches) across batches. Combined with other optimizations in this series (ByteBuffer-based RLE decoder, etc.), batch reads yield significant throughput improvements over per-value loops.

All 576 parquet-column tests pass.

### DIFF
--- a/parquet-benchmarks/pom.xml
+++ b/parquet-benchmarks/pom.xml
@@ -94,6 +94,18 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <annotationProcessorPaths>
+            <path>
+              <groupId>org.openjdk.jmh</groupId>
+              <artifactId>jmh-generator-annprocess</artifactId>
+              <version>${jmh.version}</version>
+            </path>
+          </annotationProcessorPaths>
+          <annotationProcessors>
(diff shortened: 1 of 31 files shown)


iemejia:
Closing — the batch read APIs will be superseded by forthcoming PRs for column I/O (v2-par7-column-io) and level write batching (v2-par9-level-write-batching), which are not yet opened because they depend on #3565 (PLAIN), #3568 (RLE), and other encoding PRs being merged first.

Apologies for the churn. The replacement PRs will be opened once their dependencies land. Thank you.

PR #3536, opened 2026-05-01
GH-3522: Eliminate unnecessary page-size copies in compressed page assembly and CRC checksums

iemejia:
## Summary

- Eliminate full-page `toByteArray()` copy in `BAOSBytesInput.writeInto(ByteBuffer)` by streaming through a `ByteBufferBackedOutputStream` adapter
- Eliminate full-page `toByteArray()` copy in CRC32 checksum computation by streaming through a `CRC32OutputStream` adapter

## Details

Two sources of unnecessary full-page-size `byte[]` allocations:

1. `BAOSBytesInput.writeInto(ByteBuffer)` called `toByteArray()` which copies the entire internal buffer. Replace with `writeTo()` using a thin `ByteBufferBackedOutputStream` adapter that writes directly from the internal `buf[]` without allocation.

2. CRC32 checksum computation in `ColumnChunkPageWriteStore` called `toByteArray()` on each `BytesInput` to pass to `crc.update(byte[])`. Replace with `writeAllTo(CRC32OutputStream)` that streams bytes directly to `crc.update()` without intermediate copies.

For a typical 1MB page, this eliminates 1-3 full-page allocations per page during write (one per CRC computation + one for the ByteBuffer assembly). The benefit is reduced GC pressure and peak memory rather than steady-state throughput.

All TestBytesInput tests pass. Compiles clean.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
@@ -608,12 +608,42 @@ public abstract class BytesInput {
 
     @Override
     void writeInto(ByteBuffer buffer) {
-      buffer.put(arrayOut.toByteArray());
+      // Use writeTo() which writes directly from the internal buf[] array,
+      // avoiding the toByteArray() copy that would allocate a full-size byte[]
+      try {
+        arrayOut.writeTo(new ByteBufferBackedOutputStream(buffer));
+      } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


iemejia:
Closing — this will be superseded by a forthcoming PR for column I/O and page assembly optimizations (v2-par7-column-io), which is not yet opened because it depends on #3565 (PLAIN) and #3568 (RLE) being merged first.

Apologies for the churn. The replacement PR will be opened once its dependencies land. Thank you.

PR #3543, opened 2026-05-04
GH-3542 Fix `LocalInputFile` `ByteBuffer` read methods to support direct and non-zero-position buffers

### LINKED ISSUE #3542: Fix `LocalInputFile` `ByteBuffer` read methods to support direct and non-zero-position buffers

`LocalInputFile.readFully(ByteBuffer)` and `LocalInputFile.read(ByteBuffer)` in `parquet-common` are broken for any `ByteBuffer` that either (a) does not expose an accessible backing array or (b) has a non-zero `position()` when passed in. In practice this means any call to `ParquetFileReader.readFooter` against an `InputFile` obtained from new `LocalInputFile(path)` can fail, Parquet itself passes buffer shapes that trigger the bug.

### Root cause

Both methods end with:
`buf.put(buffer, buf.position() + buf.arrayOffset(), buf.remaining());`

Two independent defects:
1. Wrong argument semantics. `ByteBuffer.put(byte[] src, int offset, int length)` treats offset as an offset into the source array. The source here is the freshly-allocated local buffer, whose indices have nothing to do with `buf.position()` or `buf.arrayOffset()`. It happens to work when both are zero; any other state either reads from the wrong offset or throws `IndexOutOfBoundsException`.
2. `arrayOffset()` is not universally defined. Direct buffers, memory-mapped buffers, and read-only views all throw `UnsupportedOperationException` from `arrayOffset()`, so the call explodes before the put is even attempted.

`read(ByteBuffer)` has an additional bug: it copies `buf.remaining()` bytes into the destination regardless of how many bytes `read(byte[])` actually returned, corrupting the buffer on short reads and advancing `position` past the EOF boundary.

### Stack trace
```
java.lang.UnsupportedOperationException
    at java.base/java.nio.ByteBuffer.arrayOffset(ByteBuffer.java:1558)
    at org.apache.parquet.io.LocalInputFile$1.readFully(LocalInputFile.java:93)
    at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:642)
    at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:578)
    at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:971)
    at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:961)
```

### Minimal reproducer
```java
Path path = /* any existing Parquet file */;
try (SeekableInputStream s = new LocalInputFile(path).newStream()) {
    s.readFully(ByteBuffer.allocateDirect(8));   // throws UnsupportedOperationException
}
```
```java
try (SeekableInputStream s = new LocalInputFile(path).newStream()) {
    ByteBuffer heap = ByteBuffer.allocate(8);
    heap.put(new byte[] {0, 0});                 // position=2
    s.readFully(heap);                           // reads from wrong offset in source array
}
```

### Version
- parquet-common 1.17.0
- Introduced by PARQUET-1822 (commit 7c4cb42a, "Avoid requiring Hadoop installation for reading/writing", #1111), which added `LocalInputFile`.

### Component(s)

Core

arouel:
### Rationale for this change

`LocalInputFile`'s `read(ByteBuffer)` and `readFully(ByteBuffer)` are broken for two independent reasons:
1. They pass `buf.position() + buf.arrayOffset()` as the source offset to `ByteBuffer.put(byte[] src, int offset, int length)`. The source is a freshly-allocated local byte[] whose indices have no relationship to `buf.position()` or `buf.arrayOffset()`, this reads from the wrong offset of the source array whenever the destination buffer's position is non-zero.
2. They call `buf.arrayOffset()`, which throws `UnsupportedOperationException` on direct buffers, memory-mapped buffers, and read-only views. `ParquetFileReader.readFooter` passes exactly such buffer shapes, so any consumer wrapping a Path with new `LocalInputFile(path)` can hit this bug.
`read(ByteBuffer)` has an additional defect: it advances the destination by `buf.remaining()` regardless of how many bytes `read(byte[])` actually returned, corrupting the buffer on short reads and at EOF.

- parquet-common/src/main/java/org/apache/parquet/io/LocalInputFile.java:
  - `readFully(ByteBuffer)` now uses `buf.put(buffer)`, a straight array-to-buffer copy that works for every ByteBuffer shape and never calls arrayOffset().
  - `read(ByteBuffer)` copies only the bytes actually read (`buf.put(buffer, 0, read)` when `read > 0`) and returns the underlying read result, so `-1` at EOF propagates correctly and the destination's position is left untouched on EOF.

### Are these changes tested?

Yes. `TestLocalInputOutput` gains eight regression tests that fail against the previous implementation and pass with the fix.

No API signatures changed.

Closes #3542

### DIFF
--- a/parquet-common/src/test/java/org/apache/parquet/io/TestLocalInputOutput.java
+++ b/parquet-common/src/test/java/org/apache/parquet/io/TestLocalInputOutput.java
@@ -18,11 +18,15 @@
  */
 package org.apache.parquet.io;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 
+import java.io.EOFException;
 import java.io.File;
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ReadOnlyBufferException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


parquet-common/src/test/java/org/apache/parquet/io/TestLocalInputOutput.java
```diff
@@ -89,4 +92,144 @@ private File createTempFile() throws IOException {
...
+    }
+  }
+
+  @Test
+  public void readFullyIntoReadOnlyByteBuffer() throws IOException {
+    Path path = writeBytes(new byte[] {7, 8, 9});
+    try (SeekableInputStream stream = new LocalInputFile(path).newStream()) {
+      ByteBuffer backing = ByteBuffer.allocate(3);
+      ByteBuffer buf = backing.asReadOnlyBuffer();
+      assertThrows(java.nio.ReadOnlyBufferException.class, () -> stream.readFully(buf));
```
steveloughran: nit: use an import here.

steveloughran (commented):
explanation makes sense and code seems good.

tests seem thorough, especially `readIntoByteBufferWithNonZeroPosition`.

wgtmac (approved):
Looks reasonable to me

parquet-common/src/test/java/org/apache/parquet/io/TestLocalInputOutput.java
arouel: I have updated the branch to have an explicit import for `ReadOnlyBufferException`.

Fokko (approved):

Fokko:
Let's move this forward! Thanks @arouel for working on this, and thanks @steveloughran and @wgtmac for the review!

PR #3546, opened 2026-05-06
GH-3545: Use parquet-java in SCM block

### LINKED ISSUE #3545: Pom.xml SCM block has the wrong repository (parquet-mr) and not (parquet-java)

Minor issue but we are currently relying on Github redirecting and forwarding for some of the maven plugin functions to work properly. I think it's worth updating. 

### Component(s)

Build

RussellSpitzer:
### Rationale for this change
The repository was renamed from apache/parquet-mr to apache/parquet-java but the SCM block was never updated. Everything has been working because Github maintains the redirect from one url to the other but it's probably better that we don't rely on the redirect.

I noticed this while I was prototyping some release automation.

The parquet-mr is replaced with parquet-java

### Are these changes tested?
Checked that mvn validate still works and it does.

No

Closes #3545

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,9 +35,9 @@
   <description>Parquet is a columnar storage format that supports nested data. This provides the java implementation.</description>
 
   <scm>
-    <connection>scm:git:git@github.com:apache/parquet-mr.git</connection>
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Good one @RussellSpitzer 👍

PR #3550, opened 2026-05-06
GH-3549: Leverage JDK11 `readNBytes`

### LINKED ISSUE #3549: Leverage JDK11 `readNBytes`

This can be used for improved efficiency.

### Component(s)

_No response_

Fokko:
Introduced in JDK11 and is more efficient.

Closes #3549

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
@@ -366,18 +366,9 @@ public abstract class BytesInput {
     @Override
     public void writeAllTo(OutputStream out) throws IOException {
       LOG.debug("write All {} bytes", byteCount);
-      // Transfer in chunks to avoid allocating a byteCount-sized intermediate buffer
-      byte[] buffer = new byte[Math.min(byteCount, 8192)];
-      int remaining = byteCount;
-      while (remaining > 0) {
-        int toRead = Math.min(remaining, buffer.length);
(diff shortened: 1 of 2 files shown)


parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -395,8 +404,11 @@ void writeInto(ByteBuffer buffer) {
 
     public byte[] toByteArray() throws IOException {
       LOG.debug("read all {} bytes", byteCount);
-      byte[] buf = new byte[byteCount];
-      new DataInputStream(in).readFully(buf);
+      byte[] buf = in.readNBytes(byteCount);
```
steveloughran: hadn't noticed this before. Doesn't look like anything in the JDK has actually subclassed it ... that is there's no AFAIK there's no faster native code version. That's probably because java.nio is the way to go there

steveloughran (commented):
commented

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
Fokko: I think it is nice to avoid the wrapper `DataInputStream` here. The biggest win here is in `writeAllTo` where we don't allocate all the bytes directly into memory, but respecting `byteCount`.

Fokko:
There seems to be an issue with `NonBlockedDecompressorStream` reads, which renders this PR more-or-less useless :)

Issue #3479, opened 2026-04-14
`RequiresFallback.isCompressionSatisfying` is too aggressive with current default page size


etseidl:
### Describe the enhancement requested

An issue recently was brought up in arrow-rs (https://github.com/apache/arrow-rs/pull/9700) which brought to my attention the existence of `isCompressionSatisfying` in the `RequiresFallback` interface. In short, after accumulating a page worth of data, `isCompressionSatisfying` is called to see if dictionary encoding is actually compressing the data at all, and if not, then the encoder falls back immediately to the fallback encoder. As far as I could determine, this behavior was introduced very early on, before the advent of the page indexes, so IIRC the page size would have been significantly larger. With page indexes, however, this function is now called after only 20000 rows have been processed. A column with a moderate cardinality might not yet have produced enough repeating values to lead this function to conclude it's best to continue using a dictionary.

For example, a dataframe with an int64 column consisting of one million values mod'd with 32768 will end up ditching dictionary encoding completely, and produce a column chunk of 8.4MB. If the page row count is bumped up to 128k, then dictionary encoding is used throughout and the resultant column chunk is only 2.2MB.

Sadly, it does not appear that this behavior is configurable, so short of increasing the page row count, its behavior cannot be modified.

I can see the need for this type of heuristic, but I think it needs to be modified in light of the current defaults resulting in far too few samples with which to determine if dictionary encoding is beneficial or not. If collecting more samples before falling back is not practical, there should at least be a configuration setting to disable this check.

### Component(s)

Core

yadavay-amzn:
I'd like to work on this. Will open a PR with a fix to make the compression-satisfying check less aggressive with the current default page size.

PR #3556, opened 2026-05-11
PARQUET-3479: Add configuration to disable early dictionary compression check

yadavay-amzn:
## Problem

`FallbackValuesWriter` calls `isCompressionSatisfying()` after the first page to decide whether dictionary encoding is worthwhile. With modern page-index defaults (~20k rows per page), this check fires too early for moderate-cardinality columns — dictionary encoding gets abandoned before enough data has accumulated to show its benefit, resulting in significantly larger files.

As reported in #3479, a column with 1M int64 values mod 32768 produces 8.4MB with the premature fallback vs 2.2MB when dictionary encoding is preserved.

## Fix

Add a configurable property `ParquetProperties.isDictionaryEarlyCheckEnabled()` (default: `true` for backward compatibility) that controls whether the first-page compression check is performed in `FallbackValuesWriter.getBytes()`.

When disabled, dictionary encoding is only abandoned when the dictionary itself exceeds size limits (`shouldFallBack()`), not based on the first-page compression ratio.

## Changes

- `ParquetProperties`: added `dictionaryEarlyCheckEnabled` field, getter, and builder method
- `FallbackValuesWriter`: added overloaded `of()` factory and constructor accepting the flag; guarded the `isCompressionSatisfying` call
- `DefaultValuesWriterFactory`: passes the config through to `FallbackValuesWriter.of()`
- New test `TestFallbackValuesWriter`: verifies dictionary encoding is preserved when the check is disabled

## Testing

- New unit tests pass (2/2)
- Existing `parquet-column` tests unaffected (default `true` preserves existing behavior)

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -327,8 +327,8 @@ public class ParquetProperties {
 
   /**
    * Returns the byte threshold after which the dictionary compression check is performed.
-   * A value of 0 means check on the first page (backward compatible default). Higher values
-   * delay the check until that many raw bytes have been accumulated across pages.
+   * A value of 0 means check on the first page. Higher values delay the check until that
+   * many raw bytes have been accumulated across pages.
    *
    * @return the byte threshold for the dictionary compression check
    */
@@ -727,8 +727,8 @@ public class ParquetProperties {
 
     /**
      * Set the raw data byte threshold after which the dictionary compression check is performed.
-     * A value of 0 means check on the first page (backward compatible default). Higher values
-     * delay the check until that many raw bytes have been accumulated across pages.
+     * A value of 0 means check on the first page. Higher values delay the check until that
+     * many raw bytes have been accumulated across pages.
      *
      * @param val byte threshold (default: 0)
      * @return this builder for method chaining
--- a/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
@@ -67,6 +67,7 @@ public class ParquetProperties {
   public static final boolean DEFAULT_STATISTICS_ENABLED = true;
   public static final boolean DEFAULT_SIZE_STATISTICS_ENABLED = true;
 
+  public static final long DEFAULT_DICTIONARY_CHECK_THRESHOLD_RAW_SIZE_BYTES = 0;
   public static final boolean DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED = true;
 
   /**
@@ -131,6 +132,7 @@ public class ParquetProperties {
   private final int rowGroupRowCountLimit;
   private final int pageRowCountLimit;
   private final boolean pageWriteChecksumEnabled;
+  private final long dictionaryCheckThresholdRawSizeBytes;
   private final ColumnProperty<ByteStreamSplitMode> byteStreamSplitEnabled;
   private final Map<String, String> extraMetaData;
   private final ColumnProperty<Boolean> statistics;
@@ -163,6 +165,7 @@ public class ParquetProperties {
     this.rowGroupRowCountLimit = builder.rowGroupRowCountLimit;
     this.pageRowCountLimit = builder.pageRowCountLimit;
     this.pageWriteChecksumEnabled = builder.pageWriteChecksumEnabled;
+    this.dictionaryCheckThresholdRawSizeBytes = builder.dictionaryCheckThresholdRawSizeBytes;
     this.byteStreamSplitEnabled = builder.byteStreamSplitEnabled.build();
     this.extraMetaData = builder.extraMetaData;
     this.statistics = builder.statistics.build();
@@ -322,6 +325,17 @@ public class ParquetProperties {
     return pageWriteChecksumEnabled;
   }
 
+  /**
+   * Returns the byte threshold after which the dictionary compression check is performed.
+   * A value of 0 means check on the first page (backward compatible default). Higher values
+   * delay the check until that many raw bytes have been accumulated across pages.
+   *
+   * @return the byte threshold for the dictionary compression check
+   */
+  public long getDictionaryCheckThresholdRawSizeBytes() {
+    return dictionaryCheckThresholdRawSizeBytes;
+  }
+
   public OptionalLong getBloomFilterNDV(ColumnDescriptor column) {
     Long ndv = bloomFilterNDVs.getValue(column);
     return ndv == null ? OptionalLong.empty() : OptionalLong.of(ndv);
@@ -415,6 +429,7 @@ public class ParquetProperties {
     private int rowGroupRowCountLimit = DEFAULT_ROW_GROUP_ROW_COUNT_LIMIT;
     private int pageRowCountLimit = DEFAULT_PAGE_ROW_COUNT_LIMIT;
     private boolean pageWriteChecksumEnabled = DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED;
+    private long dictionaryCheckThresholdRawSizeBytes = DEFAULT_DICTIONARY_CHECK_THRESHOLD_RAW_SIZE_BYTES;
     private final ColumnProperty.Builder<ByteStreamSplitMode> byteStreamSplitEnabled;
     private Map<String, String> extraMetaData = new HashMap<>();
     private final ColumnProperty.Builder<Boolean> statistics;
@@ -450,6 +465,7 @@ public class ParquetProperties {
       this.allocator = toCopy.allocator;
       this.pageRowCountLimit = toCopy.pageRowCountLimit;
       this.pageWriteChecksumEnabled = toCopy.pageWriteChecksumEnabled;
+      this.dictionaryCheckThresholdRawSizeBytes = toCopy.dictionaryCheckThresholdRawSizeBytes;
       this.bloomFilterNDVs = ColumnProperty.builder(toCopy.bloomFilterNDVs);
       this.bloomFilterFPPs = ColumnProperty.builder(toCopy.bloomFilterFPPs);
       this.bloomFilterEnabled = ColumnProperty.builder(toCopy.bloomFilterEnabled);
@@ -706,6 +722,20 @@ public class ParquetProperties {
 
     public Builder withPageWriteChecksumEnabled(boolean val) {
       this.pageWriteChecksumEnabled = val;
+      return this;
+    }
+
+    /**
+     * Set the raw data byte threshold after which the dictionary compression check is performed.
+     * A value of 0 means check on the first page (backward compatible default). Higher values
+     * delay the check until that many raw bytes have been accumulated across pages.
+     *
+     * @param val byte threshold (default: 0)
+     * @return this builder for method chaining
+     */
+    public Builder withDictionaryCheckThresholdRawSizeBytes(long val) {
+      Preconditions.checkArgument(val >= 0, "dictionaryCheckThresholdRawSizeBytes must be >= 0");
+      this.dictionaryCheckThresholdRawSizeBytes = val;
       return this;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
@@ -55,6 +55,10 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
    * Overflow is not a concern: a long would require writing over 9.2 exabytes to a single
    * column chunk, which is physically impossible. */
   private long cumulativeRawBytes = 0;
+  /** Accumulates dictionary-encoded size across pages (only reset in resetDictionary) so the
+   * compression decision compares like-scoped cumulative quantities — the column-chunk
+   * dictionary cost is amortized over all pages it covers, not charged against a single page. */
+  private long cumulativeEncodedBytes = 0;
 
   /**
    * writer currently written to
@@ -99,13 +103,30 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
     if (!fellBackAlready && !compressionChecked && cumulativeRawBytes >= dictionaryCheckThresholdRawSizeBytes) {
       compressionChecked = true;
       BytesInput bytes = initialWriter.getBytes();
-      if (!initialWriter.isCompressionSatisfying(rawDataByteSize, bytes.size())) {
+      try {
+        cumulativeEncodedBytes = Math.addExact(cumulativeEncodedBytes, bytes.size());
+      } catch (ArithmeticException e) {
+        // overflow, keep the previous value
+      }
+      // Compare cumulative raw vs cumulative encoded so the column-chunk dictionary
+      // (which is itself cumulative) is amortized over all pages it covers, not charged
+      // against a single page.
+      if (!initialWriter.isCompressionSatisfying(cumulativeRawBytes, cumulativeEncodedBytes)) {
         fallBack();
       } else {
         return bytes;
       }
     }
-    return currentWriter.getBytes();
+    BytesInput result = currentWriter.getBytes();
+    if (!fellBackAlready && !compressionChecked) {
+      // Accumulate dictionary-encoded size for pages flushed before the check fires.
+      try {
+        cumulativeEncodedBytes = Math.addExact(cumulativeEncodedBytes, result.size());
+      } catch (ArithmeticException e) {
+        // overflow, keep the previous value
+      }
+    }
+    return result;
   }
 
   @Override
@@ -149,6 +170,7 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
     fellBackAlready = false;
     compressionChecked = false;
     cumulativeRawBytes = 0;
+    cumulativeEncodedBytes = 0;
     initialUsedAndHadDictionary = false;
   }
 
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
@@ -30,12 +30,12 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
 
   public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter> FallbackValuesWriter<I, F> of(
       I initialWriter, F fallBackWriter) {
-    return new FallbackValuesWriter<>(initialWriter, fallBackWriter, /*checkAfterBytes=*/ 0);
+    return new FallbackValuesWriter<>(initialWriter, fallBackWriter, /*dictionaryCheckThresholdRawSizeBytes=*/ 0);
   }
 
   public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter> FallbackValuesWriter<I, F> of(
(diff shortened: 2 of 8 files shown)


yadavay-amzn:
@Fokko Could you take a look? This adds a config (`parquet.dictionary.early.check.enabled`) to disable the first-page compression check in FallbackValuesWriter. With modern page-index defaults (~20k rows/page), the check fires too early for moderate-cardinality columns, abandoning dictionary encoding prematurely. Includes unit test + E2E integration test writing real Parquet files. Thanks!

wgtmac:
Thanks for looking into this! 

While the problem is real, introducing a boolean flag `dictionaryEarlyCheckEnabled` feels like a band-aid fix that pushes the burden to users. Most users won't know when to manually toggle this to prevent storage inflation.

Instead of a new config, could we make the heuristic more adaptive? For example, we could delay the compression check until we've accumulated a certain amount of raw data (e.g., `1MB`), or evaluate it over the first `N` pages rather than just the first one. 

This would solve the issue out of the box without hurting usability. Thoughts?

yadavay-amzn:
Thanks for taking a look @wgtmac !

I like the idea of having a threshold on the raw data to delay the compression check. Updated the check in `FallbackValuesWriter.getBytes()` to only fire if rawDataByteSize >= threshold (defaulting to 1MB). 

I think the page count approach (checking after N pages) is sensitive to page size configurations. So with 1KB pages you'd need many pages to accumulate meaningful data, while with 1MB pages a single page might be enough. A byte threshold adapts naturally regardless of page size settings.

That being said I'd like to know which approach you'd prefer and iterate based on it.

wgtmac:
Thanks for the quick update. Before diving into the code detail, I have some additional questions:

- `parquet.dictionary.check.after.bytes` is confusing. Does "bytes" refer to raw (uncompressed) bytes or compressed/encoded bytes? Please rename it to be explicit (e.g., parquet.dictionary.check.after.raw.bytes) and update the documentation accordingly.
- In FallbackValuesWriter, rawDataByteSize is reset to 0 at the end of every page via reset(). With modern configurations (e.g., 20k rows per page), a single page's size will almost never reach the 1MB threshold. This effectively disables the compression check entirely. Did you intend for this data size to accumulate across pages, or is permanently bypassing the check for small pages your actual design?
- Since nulls are encoded in definition levels and do not contribute to rawDataByteSize, a page with a high null ratio will have an extremely small raw data size. How does this algorithm account for heavily null-populated columns? Will it also indefinitely bypass the check?
- It would be good to make the default behavior unchanged since many downstream environments (e.g. Apache Spark) unfortunately check produced file metadata (including encoding, file size, etc.) in their integration tests so behavior change may break them.

cc @gszadovszky @Fokko for additional advice.

yadavay-amzn:
@wgtmac thanks for spotting that. Updated:

1. **Renamed** config to `parquet.dictionary.check.after.raw.bytes` to clarify it refers to uncompressed bytes.

2. **Fixed the reset issue.** You were right about `rawDataByteSize` resets per page via `reset()`, so it would never reach a 1MB threshold. Added a separate `cumulativeRawBytes` counter that accumulates across pages and only resets in `resetDictionary()` (between column chunks). The threshold gate uses `cumulativeRawBytes`, the actual compression comparison still uses the current page's `rawDataByteSize` vs encoded size — same comparison as before.

3. **Default is now 0** (backward compatible, check fires on first page, same as old `firstPage` behavior). Users can opt in to a higher threshold to delay the check.

4. **Null-heavy columns:** with default 0 the check fires on the first page regardless of null ratio, same as before. With a higher threshold, nulls don't contribute to `cumulativeRawBytes` (they're in definition levels), so the threshold takes longer to reach, but, the check still eventually fires once enough non-null values accumulate. For all-null columns the check fires immediately since `cumulativeRawBytes >= 0` is trivially true.

Let me know if this direction works or if you'd prefer a different approach. Thanks again for the prompt reviews!

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -79,8 +92,8 @@ public long getBufferedSize() {
 
   @Override
   public BytesInput getBytes() {
-    if (!fellBackAlready && firstPage) {
-      // we use the first page to decide if we're going to use this encoding
+    if (!fellBackAlready && !compressionChecked && cumulativeRawBytes >= checkAfterBytes) {
```
wgtmac: Could we add a test where the threshold is crossed only after a `reset()`? This is the exact case we want to protect here, since `rawDataByteSize` is reset per page while `cumulativeRawBytes` should keep accumulating across pages. The current tests with `0` and `Long.MAX_VALUE` would not catch a regression back to per-page counting.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -771,6 +771,17 @@ public SELF withPageWriteChecksumEnabled(boolean enablePageWriteChecksum) {
       return self();
     }
 
+    /**
+     * Set the raw data byte threshold after which the dictionary compression check is performed.
+     *
+     * @param val byte threshold (0 means check on every page)
```
wgtmac: Nit: this says `0 means check on every page`, but the implementation only checks once per column chunk, same as the old first-page behavior. Could we update this to say `0` means check on the first page / preserve the previous behavior?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
```diff
@@ -161,6 +161,7 @@ public static enum JobSummaryLevel {
   public static final String BLOCK_ROW_COUNT_LIMIT = "parquet.block.row.count.limit";
   public static final String PAGE_ROW_COUNT_LIMIT = "parquet.page.row.count.limit";
   public static final String PAGE_WRITE_CHECKSUM_ENABLED = "parquet.page.write-checksum.enabled";
+  public static final String DICTIONARY_CHECK_AFTER_BYTES = "parquet.dictionary.check.after.raw.bytes";
```
wgtmac: Could we also document `parquet.dictionary.check.after.raw.bytes` in the configuration list above? It would be useful to mention that this is based on raw value bytes, and nulls encoded in definition levels do not contribute to this threshold.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -709,6 +725,19 @@ public Builder withPageWriteChecksumEnabled(boolean val) {
...
+    /**
+     * Set the raw data byte threshold after which the dictionary compression check is performed.
+     * A value of 0 means check on the first page (backward compatible default). Higher values
+     * delay the check until that many raw bytes have been accumulated across pages.
+     *
+     * @param val byte threshold (default: 0)
+     * @return this builder for method chaining
+     */
+    public Builder withDictionaryCheckAfterBytes(long val) {
+      this.dictionaryCheckAfterBytes = val;
```
wgtmac: Should we reject negative values here? A negative threshold effectively behaves like `0`, but accepting it silently seems a bit confusing for a size-like config. Most nearby size/count options validate the input, so `val >= 0` would be clearer.

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -30,7 +30,12 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
 
   public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter> FallbackValuesWriter<I, F> of(
       I initialWriter, F fallBackWriter) {
-    return new FallbackValuesWriter<>(initialWriter, fallBackWriter);
+    return new FallbackValuesWriter<>(initialWriter, fallBackWriter, 0);
```
wgtmac:
```suggestion
    return new FallbackValuesWriter<>(initialWriter, fallBackWriter, /*checkAfterBytes=*/0);
```

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -44,6 +49,14 @@ public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter
 
   private boolean fellBackAlready = false;
 
+  private boolean compressionChecked = false;
+
+  private final long checkAfterBytes;
+
+  /* Accumulates raw bytes across pages (only reset in resetDictionary) so the
+   * threshold check works even when individual pages are smaller than checkAfterBytes. */
+  private long cumulativeRawBytes = 0;
```
wgtmac, line 57:
```suggestion
  private boolean compressionChecked = false;
  private final long checkAfterBytes;
  /* Accumulates raw bytes across pages (only reset in resetDictionary) so the
   * threshold check works even when individual pages are smaller than checkAfterBytes. */
  private long cumulativeRawBytes = 0;
```

How about removing these blank lines?

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -57,16 +70,16 @@ public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter
    */
   private long rawDataByteSize = 0;
 
-  /**
-   * indicates if this is the first page being processed
-   */
-  private boolean firstPage = true;
-
   public FallbackValuesWriter(I initialWriter, F fallBackWriter) {
+    this(initialWriter, fallBackWriter, 0);
```
wgtmac:
```suggestion
    this(initialWriter, fallBackWriter, /*checkAfterBytes=*/0);
```

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -167,6 +180,7 @@ private void fallBack() {
   @Override
   public void writeByte(int value) {
     rawDataByteSize += 1;
+    cumulativeRawBytes += 1;
```
wgtmac: Should we use `cumulativeRawBytes + rawDataByteSize` for checking and only increase `cumulativeRawBytes` once a page to save some cycles?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputFormat.java
wgtmac: TBH, the flag name still looks a little bit unclear to me. How about renaming it to `parquet.dictionary.check.threshold.raw.size.bytes` and also change associated variable and function names?

gszadovszky:
I don't think that integration tests validating the result files metadata should be a concern. Parquet-java should have the freedom to change it's default behavior related to encoding/page limits etc. until according to the specification and the configured properties. These are not breaking changes.

I think we should make our changes as automatic and configuration free as possible. Introducing new configuration with staying with the current default behavior would lead to nobody actually using it. I'm OK with having the default value so the default behavior is unchanged until we want to benchmark the system in prod so we can come up with better defaults.

WDYT?

yadavay-amzn:
@wgtmac Addressed all comments:

- Applied code suggestions (named arg comments, field grouping, Javadoc on cumulativeRawBytes)
- Renamed config to `parquet.dictionary.check.threshold.raw.size.bytes`
- Performance: accumulate `cumulativeRawBytes` once per page (in `getBytes()`) instead of per value
- Added multi-page test proving threshold is crossed after `reset()`
- Javadoc fixed ("0 = check on first page")
- Negative value validation added

Thanks for the thorough review!

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -30,7 +30,12 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
 
   public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter> FallbackValuesWriter<I, F> of(
       I initialWriter, F fallBackWriter) {
-    return new FallbackValuesWriter<>(initialWriter, fallBackWriter);
+    return new FallbackValuesWriter<>(initialWriter, fallBackWriter, /*checkAfterBytes=*/ 0);
```
wgtmac: `checkAfterBytes` now should be replaced by a more consistent name with `dictionaryCheckThresholdRawSizeBytes`

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -30,7 +30,12 @@ public class FallbackValuesWriter<I extends ValuesWriter & RequiresFallback, F e
...
+  }
+
+  public static <I extends ValuesWriter & RequiresFallback, F extends ValuesWriter> FallbackValuesWriter<I, F> of(
+      I initialWriter, F fallBackWriter, long checkAfterBytes) {
```
wgtmac: ditto and same for all below

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -79,8 +89,9 @@ public long getBufferedSize() {
 
   @Override
   public BytesInput getBytes() {
-    if (!fellBackAlready && firstPage) {
-      // we use the first page to decide if we're going to use this encoding
+    cumulativeRawBytes += rawDataByteSize;
```
wgtmac: Do we need to check overflow just in case?

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -322,6 +325,17 @@ public boolean getPageWriteChecksumEnabled() {
     return pageWriteChecksumEnabled;
   }
 
+  /**
+   * Returns the byte threshold after which the dictionary compression check is performed.
+   * A value of 0 means check on the first page (backward compatible default). Higher values
```
wgtmac: IMO, we don't need to mention `backward compatible default` everywhere.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java
```diff
@@ -771,6 +771,17 @@ public SELF withPageWriteChecksumEnabled(boolean enablePageWriteChecksum) {
       return self();
     }
 
+    /**
+     * Set the raw data byte threshold after which the dictionary compression check is performed.
+     *
+     * @param val byte threshold (0 means check on the first page / preserve previous behavior)
```
wgtmac:
```suggestion
     * @param val byte threshold (0 means checking on the first page for every column chunk)
```

wgtmac (commented):
Thanks for addressing the feedback! Left some nits.

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
yadavay-amzn: Renamed `checkAfterBytes` to `dictionaryCheckThresholdRawSizeBytes` throughout

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
yadavay-amzn: Renamed `checkAfterBytes` to `dictionaryCheckThresholdRawSizeBytes` throughout

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
yadavay-amzn: Practically i dont think we'll hit overflow but i added a check just in case

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
yadavay-amzn: Removed

yadavay-amzn:
@wgtmac Fixed the nits. Thanks again!

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -79,8 +91,11 @@ public long getBufferedSize() {
 
   @Override
   public BytesInput getBytes() {
-    if (!fellBackAlready && firstPage) {
-      // we use the first page to decide if we're going to use this encoding
+    if (cumulativeRawBytes + rawDataByteSize >= 0) {
+      cumulativeRawBytes += rawDataByteSize;
```
wgtmac: Why not calling `Math.addExact` just like others?

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
```diff
@@ -79,8 +91,13 @@ public long getBufferedSize() {
...
-      // we use the first page to decide if we're going to use this encoding
+    try {
+      cumulativeRawBytes = Math.addExact(cumulativeRawBytes, rawDataByteSize);
+    } catch (ArithmeticException e) {
+      // overflow, keep the previous value
+    }
+    if (!fellBackAlready && !compressionChecked && cumulativeRawBytes >= dictionaryCheckThresholdRawSizeBytes) {
+      compressionChecked = true;
       BytesInput bytes = initialWriter.getBytes();
       if (!initialWriter.isCompressionSatisfying(rawDataByteSize, bytes.size())) {
```
wgtmac: I think the threshold gate now accumulates raw bytes across pages, but the compression check still compares the current page raw size against the current page encoded size plus the whole accumulated dictionary size. This seems biased toward fallback once the check is delayed beyond the first page. Should this comparison use cumulative raw/encoded sizes, or otherwise avoid charging the full column-chunk dictionary size to a single page? A finite-threshold test with repeated/moderate-cardinality values across pages would help verify the intended PARQUET-3479 case.

yadavay-amzn:
Thanks @wgtmac. You're right, the trigger became cumulative but the comparison was still per-page against the cumulative dictionary, which biases toward fallback the later the check fires.

I've pushed a fix that tracks `cumulativeEncodedBytes` alongside `cumulativeRawBytes` and changes the decision to `isCompressionSatisfying(cumulativeRawBytes, cumulativeEncodedBytes)`, so the column-chunk dictionary is amortized over all the pages it covers rather than charged against a single page. The encoded size is accumulated from the `BytesInput` already produced each page (no extra `getBytes()` calls), and both accumulators now use `Math.addExact` consistently and reset together in `resetDictionary()`.

I added a test (`TestFallbackCumulativeBias`) that reproduces exactly the case you described; repeated/moderate-cardinality values across multiple pages with a finite threshold that fires on a later page. The arithmetic witness:
- **Before (per-page):** `encoded(93) + dict(400) = 493 >= pageRaw(400)` -> fallback
- **After (cumulative):** `totalEncoded(173) + dict(400) = 573 < totalRaw(800)` -> dictionary kept

It fails on the previous commit and passes now. I also added an adversarial test confirming a genuinely high-cardinality column still correctly falls back (the fix amortizes the dictionary cost, it doesn't disable fallback), plus a `resetDictionary` isolation test so one chunk's history doesn't leak into the next.

parquet-column/src/main/java/org/apache/parquet/column/values/fallback/FallbackValuesWriter.java
yadavay-amzn: fixed this

PR #3559, opened 2026-05-13
GH-3558: Properly close buffers

### LINKED ISSUE #3356: TestParquetReader fails with hadoop 3.4.2

### LINKED ISSUE #3558: Properly close buffers

As @pan3793 pointed out https://github.com/apache/parquet-java/pull/3360#issuecomment-3588399639, when updating to a later version of Hadoop, it starts failing with an error of lingering buffers.

### Component(s)

_No response_

Fokko:
### Rationale for this change

With updating to a later version of Hadoop, we got the error that there are lingering buffers that are not closed properly. This PR addresses this issue by properly releasing the acquired buffers.

Also, closes some of the readers using try-with-resource pattern.

Initially caught by @pan3793 in https://github.com/apache/parquet-java/pull/3360#issuecomment-3588399639

Closes #3558
Closes #3356


pom.xml
```diff
@@ -83,7 +83,7 @@
     <spotless.version>2.46.1</spotless.version>
     <shade.prefix>shaded.parquet</shade.prefix>
     <!-- Guarantees no newer classes/methods/constants are used by parquet. -->
-    <hadoop.version>3.3.0</hadoop.version>
+    <hadoop.version>3.4.0</hadoop.version>
```
wgtmac: Do we want to reflect this in the PR title?

parquet-common/src/main/java/org/apache/parquet/bytes/TrackingByteBufferAllocator.java
```diff
@@ -154,12 +171,12 @@ public boolean isDirect() {
   }
 
   @Override
-  public void close() throws LeakedByteBufferException {
-    if (!allocated.isEmpty()) {
-      LeakedByteBufferException ex = new LeakedByteBufferException(
```
wgtmac, line 159: This looks like a behavior change?

wgtmac (commented):
It would be good if @gszadovszky could take a look at this as this is related to `TrackingByteBufferAllocator`.

pom.xml
Fokko: This is the underbound of the supported Hadoop version, so I think it is best to set this to 3.3.0 before merging this. Unless @steveloughran has an opinion on a reasonable lower-bound :)

parquet-common/src/main/java/org/apache/parquet/bytes/TrackingByteBufferAllocator.java
Fokko: Yes, it is, but it is not covered in any tests. WDYT @gszadovszky ?

Fokko: I don't think this was working properly since Hadoop did raise exception around not properly closing buffers, while this exception wasn't being thrown. Let me dig a bit further here

gszadovszky: Yeah. The whole point of this allocator implementation is to fail if something was not released. It may make sense to actually release the buffers so the tests won't leak, be we need to throw that exception, otherwise this class is pointless.

gszadovszky: When I've created this class, I've caught a couple of leaks in parquet-java code with it. Not having that exception thrown during our unit tests means the leaks are fixed not that the monitoring class is not working :)

steveloughran:
so for all cluster filesystems with vector io, it's fine as is. HDFS doesn't support it and for the cloud stores it's all ranged reads straight into allocated buffers.

I think maybe in hadoop we should just cut the attempt to be clever and merge ranges, and just do the parallel reads. On clusterfs work with Owen O'Malley and claude to do the right thing here. 

what would be the perceived penalty of reading the whole file block into one allocated buffer, copying the requested pieces into two separate buffers, and then releasing the larger one (a release function is now returned down after all).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -2327,6 +2327,10 @@ public void readFromVectoredRange(ParquetFileRange currRange, ChunkListBuilder b
         LOG.error(error, e);
         throw new IOException(error, e);
       }
+      // Release the vectored-read buffer back to the allocator when the row group is closed.
+      // Requires fs.file.checksum.verify=false so the returned buffer is the allocator buffer
+      // rather than a sliced subset (see Hadoop's fs.file.checksum.verify docs).
+      builder.addBuffersToRelease(Collections.singletonList(buffer));
```
wgtmac: This assumes the future returns the exact buffer allocated by our allocator. With local checksum verification enabled, Hadoop can return a sliced buffer, so strict allocators like `TrackingByteBufferAllocator` or `ReusingByteBufferAllocator` may fail on release or still report a leak. The new test `core-site.xml` disables that path globally, so this should either guard against sliced vectored IO or have a test with checksum verification left on. (reviewed by Codex)

wgtmac: If this range is registered and a later vectored range times out or fails before `rowGroup.setReleaser(builder.releaser)`, the builder is dropped without closing the releaser. We should close the builder releaser on the failure path, or transfer ownership earlier so partially-read row groups still release their buffers. (reviewed by Codex)

Fokko: Disabling it globally is exactly what where trying to do. There is an open issue on Hadoop that needs to be fixed: https://issues.apache.org/jira/browse/HADOOP-19901 That's not on Parquet to fix or test.

Fokko: Agreed — if a later range times out or fails, the exception propagates before `rowGroup.setReleaser(builder.releaser)` runs, so the already-acquired buffers leak.

Fixed by transferring ownership on the failure path:

- Extracted a `readChunkPagesForBlock(...)` helper (shared by `internalReadRowGroup` and `internalReadFilteredRowGroup`) that closes `builder.releaser` if anything throws before ownership is transferred to the row group.
- In `readVectored`, moved the buffer registration into a `finally` block so a partially-completed vectored read still registers every buffer it acquired, ensuring they're released even when a later range fails.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -2368,6 +2384,10 @@ public void readFromVectoredRange(ParquetFileRange currRange, ChunkListBuilder b
...
```
wgtmac: With checksum verification enabled by default, `ChecksumFileSystem` returns a sliced buffer rather than the allocator-owned buffer. Releasing it here fails with strict allocators and leaves the original allocations unreleased; `core-site.xml` only hides this path in tests. Please either fall back when `VECTOREDIO_BUFFERS_SLICED` is advertised or track and release the actual allocated buffers.

wgtmac: Please register the buffer immediately after `awaitFuture` succeeds. `setReadMetrics` invokes user callback code before the buffer is registered, so an exception there leaks the current buffer despite the outer cleanup.

wgtmac (commented):
My Codex review still finds two outstanding issues posted as inline comments.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
Fokko: Thanks, this is a great one. I've moved it up

PR #3562, opened 2026-05-14
GH-3561 Harden variants against malformed metadata.

### LINKED ISSUE #3561: Harden variant decoding

org.apache.parquet.variant.Variant is brittle against malformed variants, whereas the arrow variant is very strict and includes an optional deep validate phase

* Harden the java implementation, where it doesn't impose any performance penalty
* add tests which can also generate malformed files to add to the parquet test files

The work in #3481 should include some of this or this should be an immediate followon (preferred)

### Component(s)

parquet-variant

steveloughran:
### Rationale for this change

Malformed parquet files could be distruptive enough to not only affect the execution of a single worker thread (which will ultimately reject it), but other threads on the same process.  This can be disruptive.

- reject oversized metadata/value declarations
- reject oversize dictSize in objects
- range checking

Only low cost checks are made, equivalent to arrow variant `try_new_with_metadata_and_shallow_validation()`

There's no equivalent `with_full_validation()` logic is omitted. The caching logic of #3481 may be able to do this when it builds a dictionary, as range checking the increasing dictionary offsets is the key work there.

There's also a depth check consistent with the json parser; it's arguable as to whether that is needed. It will defend against StackOverflowExceptions by anything trying to treewalk, but shouldn't that code be the place to do the checks?

### Are these changes tested?

The new test suite TestHardenedReader can be configured to actually emit the malformed files, to see how applications deal with them.

Existing tests needed changes because they'd been getting away with malformed byte arrays; now this stuff is being checked the test cases need valid variants.

No

Closes #3561

### DIFF
--- a/parquet-variant/pom.xml
+++ b/parquet-variant/pom.xml
@@ -69,6 +69,17 @@
       <version>${slf4j.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-hadoop</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-client</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
@@ -118,12 +118,8 @@ public final class Variant {
     this.depth = 0;
     // A metadata buffer must contain at least the version byte
     Preconditions.checkArgument(this.metadata.remaining() >= 1, "variant metadata is empty");
-    // There is currently only one allowed version.
-    if ((metadata.get(metadata.position()) & VariantUtil.VERSION_MASK) != VariantUtil.VERSION) {
-      throw new UnsupportedOperationException(String.format(
-          "Unsupported variant metadata version: %d",
-          metadata.get(metadata.position()) & VariantUtil.VERSION_MASK));
-    }
+    // There is no checking of version until a policy of how to handle mismatched versions
+    // is defined.
 
     // Pre-compute dictionary size for lazy metadata cache allocation.
     int pos = this.metadata.position();
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
@@ -61,6 +61,8 @@ public final class Variant {
    * Lazy cache for the parsed array header.
    */
   private VariantUtil.ArrayInfo cachedArrayInfo;
+  /** Nesting depth of this Variant relative to the top-level value (0 = top-level). */
+  private final int depth;
 
   /**
    * The threshold to switch from linear search to binary search when looking up a field by key in
@@ -69,10 +71,31 @@ public final class Variant {
    */
   static final int BINARY_SEARCH_THRESHOLD = 32;
 
+  /**
+   * Create a Variant from a tuple of (value, metadata) byte arrays.
+   * Includes validation of the version and the top-level structure.
+   * @param value value buffer
+   * @param metadata metadata buffer
+   * @throws UnsupportedOperationException if there is a version mismatch
+   * @throws IllegalArgumentException for any validation failure.
+   */
   public Variant(byte[] value, byte[] metadata) {
     this(value, 0, value.length, metadata, 0, metadata.length);
   }
 
+  /**
+   * Create a Variant from a subset of the (value, metadata) buffers
+   * supplied.
+   * Includes validation of the version and the toplevel structure.
+   * @param value value buffer
+   * @param valuePos offset where the value data begins
+   * @param valueLength length of value data
+   * @param metadata metadata buffer
+   * @param metadataPos offset where the metadata begins.
+   * @param metadataLength length of the metadata.
+   * @throws UnsupportedOperationException if there is a version mismatch
+   * @throws IllegalArgumentException for any validation failure.
+   */
   public Variant(byte[] value, int valuePos, int valueLength, byte[] metadata, int metadataPos, int metadataLength) {
     this(ByteBuffer.wrap(value, valuePos, valueLength), ByteBuffer.wrap(metadata, metadataPos, metadataLength));
   }
@@ -81,10 +104,20 @@ public final class Variant {
     this(value, metadata.getEncodedBuffer());
   }
 
+  /**
+   * Create a Variant from a tuple of (value, metadata) buffers.
+   * Includes validation of the version and the toplevel structure.
+   * @param value value buffer
+   * @param metadata metadata buffer
+   * @throws UnsupportedOperationException if there is a version mismatch
+   * @throws IllegalArgumentException for any validation failure.
+   */
   public Variant(ByteBuffer value, ByteBuffer metadata) {
     this.value = value.asReadOnlyBuffer().order(ByteOrder.LITTLE_ENDIAN);
     this.metadata = metadata.asReadOnlyBuffer().order(ByteOrder.LITTLE_ENDIAN);
-
+    this.depth = 0;
+    // A metadata buffer must contain at least the version byte
+    Preconditions.checkArgument(this.metadata.remaining() >= 1, "variant metadata is empty");
     // There is currently only one allowed version.
     if ((metadata.get(metadata.position()) & VariantUtil.VERSION_MASK) != VariantUtil.VERSION) {
       throw new UnsupportedOperationException(String.format(
@@ -95,29 +128,39 @@ public final class Variant {
     // Pre-compute dictionary size for lazy metadata cache allocation.
     int pos = this.metadata.position();
     int metaOffsetSize = ((this.metadata.get(pos) >> 6) & 0x3) + 1;
-    if (this.metadata.remaining() > 1) {
-      Preconditions.checkArgument(
-          this.metadata.remaining() >= 1 + metaOffsetSize,
-          "variant metadata truncated: offsetSize=" + metaOffsetSize);
-      this.dictSize = VariantUtil.readUnsignedLittleEndian(this.metadata, pos + 1, metaOffsetSize);
-      long dictTableEnd = 1L + metaOffsetSize + ((long) this.dictSize + 1) * metaOffsetSize;
-      Preconditions.checkArgument(
-          dictTableEnd <= this.metadata.remaining(),
-          "variant metadata dictionary extends past buffer: dictSize=" + this.dictSize);
-    } else {
-      this.dictSize = 0;
-    }
+    Preconditions.checkArgument(
+        this.metadata.remaining() >= 1 + metaOffsetSize,
+        "variant metadata truncated: offsetSize=" + metaOffsetSize);
+    this.dictSize = VariantUtil.readUnsignedLittleEndian(this.metadata, pos + 1, metaOffsetSize);
+    long dictTableEnd = 1L + metaOffsetSize + ((long) this.dictSize + 1) * metaOffsetSize;
+    Preconditions.checkArgument(
+        dictTableEnd <= this.metadata.remaining(),
+        "variant metadata dictionary extends past buffer: dictSize=" + this.dictSize);
     this.metadataCache = null;
+    VariantUtil.validateValueShallow(this.value, dictSize);
   }
 
   /**
-   * Package-private constructor that shares pre-parsed metadata state from a parent Variant.
+   * Package-private constructor that shares pre-parsed metadata state from a parent Variant, performing shallow validation of the structure.
+   * @param value value buffer
+   * @param metadata metadata buffer
+   * @param metadataCache shared metadata cache.
+   * @param dictSize shared dictionary size.
+   * @param depth depth of this variant in a recursive structure.
+   * @throws IllegalArgumentException if the depth of variants is too high or the structure
+   * invalid in some other form.
    */
-  Variant(ByteBuffer value, ByteBuffer metadata, String[] metadataCache, int dictSize) {
+  Variant(ByteBuffer value, ByteBuffer metadata, String[] metadataCache, int dictSize, int depth) {
+    Preconditions.checkArgument(
+        depth <= VariantUtil.MAX_VARIANT_DEPTH,
+        "variant nesting depth exceeds maximum %s",
+        VariantUtil.MAX_VARIANT_DEPTH);
     this.value = value.asReadOnlyBuffer().order(ByteOrder.LITTLE_ENDIAN);
     this.metadata = metadata.asReadOnlyBuffer().order(ByteOrder.LITTLE_ENDIAN);
     this.metadataCache = metadataCache;
     this.dictSize = dictSize;
+    this.depth = depth;
+    VariantUtil.validateValueShallow(this.value, dictSize);
   }
 
   public ByteBuffer getValueBuffer() {
@@ -359,9 +402,10 @@ public final class Variant {
 
   /**
    * Creates a child Variant that shares this instance's metadata cache.
+   * @throws IllegalArgumentException validation error during construction.
    */
   private Variant childVariant(ByteBuffer childValue) {
-    return new Variant(childValue, metadata, metadataCache, dictSize);
+    return new Variant(childValue, metadata, metadataCache, dictSize, depth + 1);
   }
 
   /**
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
@@ -16,6 +16,8 @@
  */
 package org.apache.parquet.variant;
 
+import static org.apache.parquet.variant.VariantUtil.MAX_VARIANT_DEPTH;
+
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
@@ -37,7 +39,7 @@ public final class VariantJsonParser {
 
   private static final JsonFactory JSON_FACTORY = JsonFactory.builder()
       .streamReadConstraints(StreamReadConstraints.builder()
-          .maxNestingDepth(500)
+          .maxNestingDepth(MAX_VARIANT_DEPTH)
           .maxStringLength(10_000_000)
           .maxDocumentLength(50_000_000L)
           .build())
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
@@ -877,7 +877,10 @@ class VariantUtil {
     // Extracts the highest 2 bits in the metadata header to determine the integer size of the
     // offset list.
     int offsetSize = ((metadata.get(pos) >> 6) & 0x3) + 1;
+    // dictionary size
     int dictSize = readUnsigned(metadata, pos + 1, offsetSize);
+    // range check the dictionary, including from integer overflow
+    checkIndex(Math.toIntExact((long) dictSize * offsetSize), metadata.limit());
     HashMap<String, Integer> result = new HashMap<>();
     int offset = readUnsigned(metadata, pos + 1 + offsetSize, offsetSize);
     for (int id = 0; id < dictSize; id++) {
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
@@ -195,7 +195,7 @@ class VariantUtil {
    * same limit as in VariantJsonParser: {@value}.
    * <p>This depth is not part of the specification and was chosen after discussion on the
    * Parquet developer mailing list to be high enough to reduce the impact of parsing malicious files
-   * on a multi-tenant, multi-threaded system, while still supporting complex structurres.
+   * on a multi-tenant, multi-threaded system, while still supporting complex structures.
    *
    * @see <a href="https://lists.apache.org/thread/q6wbom1q9pndv2nj6wcynxjcjxxkc1hm">how deep is a realistic variant depth?</a>
    */
@@ -769,7 +769,8 @@ class VariantUtil {
         dataStart <= (long) value.limit() - value.position(),
         "variant object offset table extends past buffer: numElements=%s",
         numElements);
-    return new ObjectInfo((int) numElements, idSize, offsetSize, idStartOffset, (int) offsetStart, (int) dataStart);
+    return new ObjectInfo(
+        Math.toIntExact(numElements), idSize, offsetSize, idStartOffset, (int) offsetStart, (int) dataStart);
   }
 
   /**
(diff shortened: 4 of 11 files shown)


steveloughran:
@rdblue why not review this PR before worrying about the iceberg one, as its where some things could be clarified and then repeated.

* replicate arrow's "full verification"? they have a verified flag and do that on demand. Primarily one of verifying that the dictionary keys are all good. More expensive. If you are just scanning through records doing filtering and projection, I don't see it's needed.
* what is a good depth limit for children?

steveloughran:
@Fokko could you look at this while I work on that thrift update?

parquet-variant/src/test/java/org/apache/parquet/variant/VariantTestUtil.java
```diff
@@ -35,11 +37,12 @@ public class VariantTestUtil {
   /** Random number generator for generating random strings */
   private static SecureRandom random = new SecureRandom(new byte[] {1, 2, 3, 4, 5});
 
-  static final ByteBuffer EMPTY_METADATA = ByteBuffer.wrap(new byte[] {0b1});
+  // version=1, offsetSize=1, dictSize=0, single end-offset=0 — minimum well-formed empty metadata
+  static final ByteBuffer EMPTY_METADATA = ByteBuffer.wrap(new byte[] {0b1, 0x00, 0x00});
```
Fokko, line 40: I assume that we don't accept the single byte anymore? Should we still allow that, or make it explicit to the users?

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -108,16 +110,23 @@ public Variant(ByteBuffer value, ByteBuffer metadata) {
       this.dictSize = 0;
     }
     this.metadataCache = null;
+    VariantUtil.validateValueShallow(this.value, dictSize);
```
Fokko, line 136: Since we now validate in the constructor, should we add some Javadoc to this public function? Including the `@throws IllegalArgumentException`

Fokko (commented):
This look good to me, thanks @steveloughran for working on this. I left a few comments. Should we also test the performance implications for list/dicts? Maybe a small JMH microbenchmark.

parquet-variant/src/test/java/org/apache/parquet/variant/VariantTestUtil.java
steveloughran: good point. my reading of the spec says header includes 'offset size -1`
set offset-size -1 to 0 and you have

* dictionary size = 1
* metadata size = 1

This is something maybe to call out/clarify in the docs, but I don't see it being possible to declare a zero byte dictionary

steveloughran:
Latest status. Ran benchmarks unattended; still lot of variation where even some tests which only write data are being statisticially significantly faster to serialize. That's a codepath which isn't being updated, and implies that system/other work is interfering. 

Results: https://github.com/steveloughran/benchmarking-variants/tree/main/json/hardening
[final hardened json](https://github.com/steveloughran/benchmarking-variants/blob/main/json/hardening/2026-06-01-parquet-hardened-01.json) and [baseline master](https://github.com/steveloughran/benchmarking-variants/blob/main/json/hardening/2026-05-31-parquet-master-02.json)

Comparing these with [JMH tabulate](https://github.com/steveloughran/jmh-tabulate) (important: use my fork as it strictly enforces a safe version of the charting.js lib from npm):
Screenshot 2026-06-01 at 10 18 59


This looks like a slowdown but filter on statistical significance and most vanish and of the three which are significant, *two are writing data not reading it*
Screenshot 2026-06-01 at 10 17 32

the two `write` variants are just serializing the prebuilt variant, so not stressing the code. the reader does on a deeply nested structure, but there I'm not sure anything is showing
Screenshot 2026-06-01 at 10 39 29

Summary: even though minor some statistically significant slowdown is being reported, the fact that speedups in unmodified codepaths are also observed tells me the results aren't reliable.

Whatever changes are being made here, they aren't actually measurable in the new dataset and general os/execution/jvm noise is more of a factor


+ updated all the constructor javadocs

parquet-variant/src/test/java/org/apache/parquet/variant/VariantTestUtil.java
steveloughran: update: no, it's impossible to generate a valid zero byte variant
* fail fast with an explicit message (allows a branch to be cut later)
* test

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
steveloughran: done

steveloughran:
latest status
* rebase to master
* explicit zero byte metadata rejection and test
* new test for invalid version
* javadocs review
* switch to `readUnsignedLittleEndian()` to read the metadata during validation

steveloughran:
@Fokko any chance of a final review + merge? there's nothing left to do that I can see here...any else would be feature creep.

steveloughran:
feature creep: handle unrecognised primitive type as per spec SHOULD "an implementation should be able to read the rest of the Variant value"

steveloughran:
@laskoviymishka can you review this alongside the iceberg PR https://github.com/apache/iceberg/pull/16568 ? we are trying to harden them together

steveloughran:
squash and rebase to deal with the assertj move. happy with the move though!

parquet-variant/pom.xml
```diff
@@ -69,6 +69,17 @@
       <version>${slf4j.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-hadoop</artifactId>
```
laskoviymishka, line 74: The Iceberg tests for this stayed at direct ByteBuffer construction, so I was a little surprised to see the round-trip harness pull parquet-hadoop and hadoop-client into parquet-variant's test scope. Since Parquet's binary column encoding is just a length-prefixed byte copy, I don't think most of these cases exercise a different construction path than wrapping the bytes directly. Might be worth keeping the round-trip for the one or two cases where encode/decode genuinely matters — but you'll know the parquet-java test conventions better than I do.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -61,6 +61,8 @@ public final class Variant {
    * Lazy cache for the parsed array header.
    */
   private VariantUtil.ArrayInfo cachedArrayInfo;
+  /** Nesting depth of this Variant relative to the top-level value (0 = top-level). */
+  private final int depth;
```
laskoviymishka, line 65: Minor: depth here counts navigation depth through Variant objects, so it also ticks up for leaf primitives and short strings, not just container nesting. Harmless — a one-line note on the field might just save the next reader a double-take.

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -95,29 +128,39 @@ public Variant(ByteBuffer value, ByteBuffer metadata) {
...
-      this.dictSize = VariantUtil.readUnsignedLittleEndian(this.metadata, pos + 1, metaOffsetSize);
-      long dictTableEnd = 1L + metaOffsetSize + ((long) this.dictSize + 1) * metaOffsetSize;
-      Preconditions.checkArgument(
-          dictTableEnd <= this.metadata.remaining(),
-          "variant metadata dictionary extends past buffer: dictSize=" + this.dictSize);
-    } else {
-      this.dictSize = 0;
-    }
+    Preconditions.checkArgument(
+        this.metadata.remaining() >= 1 + metaOffsetSize,
```
laskoviymishka, line 128: This looks like a genuine fix, but it's also a public-API behavior change that might be worth a release note. A 1-byte metadata buffer (just the version byte) used to construct successfully with dictSize=0 via the old `else` branch, and now throws — so any downstream caller relying on that leniency would start seeing IllegalArgumentException.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
```diff
@@ -37,7 +39,7 @@ public final class VariantJsonParser {
 
   private static final JsonFactory JSON_FACTORY = JsonFactory.builder()
       .streamReadConstraints(StreamReadConstraints.builder()
-          .maxNestingDepth(500)
+          .maxNestingDepth(MAX_VARIANT_DEPTH)
```
laskoviymishka, line 42: This bumps the JSON parse-depth boundary from 500 to 1000 as a side effect of sharing the constant, so documents that used to overflow at 501 now parse. Since buildJson/buildJsonArray/buildJsonObject recurse on the real JVM stack, I wasn't sure whether 1000 had been validated as safe on this path specifically (as opposed to the descent path). Might be worth a release note, and maybe a test that parses ~1000-level JSON on a realistic stack. wdyt?

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -190,6 +190,12 @@ class VariantUtil {
   // The size (in bytes) of a UUID.
   static final int UUID_SIZE = 16;
 
+  /**
+   * Maximum permitted nesting depth of a Variant value.
+   * same limit as in VariantJsonParser.
+   */
+  static final int MAX_VARIANT_DEPTH = 1000;
```
laskoviymishka, line 202: I checked and the spec doesn't define a max nesting depth — the Iceberg side treats this as a safety limit too, so I think it's a reasonable cap. The one thing I'd flag is interop: a spec-valid variant nested deeper than 1000, written by Spark/PyIceberg/Arrow, would be rejected here. Might be worth a Javadoc note that it's an implementation limit rather than a spec rule, so that expectation is clear.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -880,6 +886,158 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+   * is not penalized by sub-trees the caller never inspects.
+   *
+   * @param valueBuffer the variant value buffer (position/limit define the extent of this node's slot)
+   * @param dictSize the metadata dictionary size, used to bound object field ids
+   * @throws IllegalArgumentException if the value header or container table does not fit within
+   *     the buffer slot, or if any object field id is out of range
+   */
+  static void validateValueShallow(final ByteBuffer valueBuffer, final int dictSize) {
+    int pos = valueBuffer.position();
+    Preconditions.checkArgument(pos >= 0 && pos < valueBuffer.limit(), "variant value is empty");
```
laskoviymishka: Small thing: pos is `valueBuffer.position()`, which is never negative by ByteBuffer's contract, so the `pos >= 0` half can't fire. Might read more clearly dropped, or with a comment on why it's kept.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -880,6 +886,158 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+    long slot = (long) valueBuffer.limit() - pos;
```
laskoviymishka, line 931: One subtlety worth a doc note (not a change, I think): `slot` is bounded by the top-level buffer's limit, not this node's parent-declared extent, since slice() doesn't set a limit. So a child can — truthfully per its own size — extend into bytes that logically belong to a sibling or the parent's offset table. It's memory-safe since it's the same backing array, and this matches the shallow-not-full split on the Iceberg side, so I'd just note the limitation rather than tighten it.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -880,6 +886,158 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+    }
+  }
+
+  /**
+   * Shallow validation of a container.
+   *
+   * @param valueBuffer buffer with the variant data
+   * @param typeInfo type information from the metadata
+   * @param pos buffer read position
+   * @param length length of slot for this primitive.
```
laskoviymishka: Looks like a copy-paste from validatePrimitiveShallow: the @param says "length of slot for this primitive" but this is a container.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -880,6 +886,158 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+      idSize = 0;
+    }
+    int offsetSize = (typeInfo & 0x3) + 1;
+    int sizeBytes = largeSize ? U32_SIZE : 1;
+    Preconditions.checkArgument(1L + sizeBytes <= length, "variant container header truncated");
+    int numElements = readUnsignedLittleEndian(valueBuffer, pos + 1, sizeBytes);
+    long idStart = 1L + sizeBytes;
+    long idBytes = isObject ? (long) numElements * idSize : 0L;
+    long offsetStart = idStart + idBytes;
+    long offsetBytes = (long) (numElements + 1) * offsetSize;
```
laskoviymishka: This is the one I'd most want a second look at, since it's where the port seems to have drifted from the Iceberg original.

`(numElements + 1)` here is computed as int arithmetic before the `(long)` widening applies, so a largeSize container with `numElements == Integer.MAX_VALUE` wraps `offsetBytes` negative — `dataStart` goes negative, the `dataStart <= length` guard passes, and the loop below then reads ~2^31 entries out of bounds (reachable for an ARRAY, where idBytes is 0). It'd surface as IndexOutOfBoundsException rather than the IllegalArgumentException the Javadoc promises.

The line two above already widens the right way (`(long) numElements * idSize`), as does Variant.java (`((long) this.dictSize + 1) * metaOffsetSize`), so it may just be:

```java
long offsetBytes = ((long) numElements + 1) * offsetSize;
```

The Iceberg original also capped numElements with a MAX_ELEMENTS bound to make this class unreachable, which doesn't seem to have come across in the port. I could be missing a parquet-java-side reason it can't be hit — a regression test at `Integer.MAX_VALUE` for ARRAY and OBJECT would confirm either way. wdyt?

parquet-variant/src/test/java/org/apache/parquet/variant/TestHardenedReader.java
```diff
@@ -0,0 +1,746 @@
...
+   * Reject Offset arithmetic that would wrap java's signed 32-bit integer.
+   */
+  @Test
+  public void testOffsetArithmeticBoundsCheck() throws IOException {
+    // Object header that claims numElements and offsetSize values whose product overflows int
+    // unless validated with widened arithmetic.
+    // Layout: [objectHeader(large=1, idSize=1, offsetSize=4) = 0x4E,
+    //          numElements (4 bytes, little-endian) = 0x33333333,
+    //          three filler bytes]
+    byte[] value = new byte[] {0x4E, 0x33, 0x33, 0x33, 0x33, (byte) 0xFF, (byte) 0xFF, 0x3F};
```
laskoviymishka: Comparing against the Iceberg tests, I don't think this one quite reaches the overflow it's named for. numElements here (0x33333333, ~858M) doesn't wrap on `+1` — so it passes via the offset-table-too-big path, which was already correct, rather than the widened arithmetic. Passing "" for the message substring also means the failure text isn't asserted. If the overflow in the line above is real, this test would stay green.

If it turns out to be a real gap, retargeting at `numElements == Integer.MAX_VALUE` with a largeSize ARRAY header (and one for OBJECT), asserting the specific exception type, would pin it down.

laskoviymishka (commented):
I reviewed the original Iceberg PR this was ported from, apache/iceberg#16568, so I mostly looked at this through a parity lens rather than reopening the design. The shallow validation on descent, depth threading through `childVariant`, and primitive size table all match what landed in Iceberg.

I’m not a committer here, so please weigh these against parquet-java conventions. Some differences may be intentional.

The one thing I’d want another look at before merge is in `validateContainerShallow`:

`(long) (numElements + 1) * offsetSize`

The cast happens after `numElements + 1`, so that addition is still `int` arithmetic. With `numElements == Integer.MAX_VALUE`, it wraps negative, `dataStart` goes negative, the bounds check passes, and the offset loop can read out of bounds. This is reachable for arrays.

The Iceberg version handled this class with a `MAX_ELEMENTS` cap, which doesn’t seem to have made it into the port. The line above already widens correctly with `(long) numElements * idSize`, so this may just need:

`((long) numElements + 1) * offsetSize`

plus the cap from the original. I may be missing a parquet-java-specific reason this is unreachable, though.

`testOffsetArithmeticBoundsCheck` also doesn’t seem to hit the wrap. `0x33333333` fails because the offset table is too large, not because `numElements + 1` overflows, and the test doesn’t assert the error message. I’d retarget it at `Integer.MAX_VALUE` for both array and object cases.

Smaller parity/doc notes:

* `MAX_VARIANT_DEPTH` is an implementation safety limit, not a spec limit. Worth saying that in Javadoc, since deeper variants from other engines would be rejected here.
* The same change raises JSON parser depth from 500 to 1000, which may be worth a release note.
* A one-byte metadata buffer that used to construct now throws. Probably correct, but still a behavior change.
* The round-trip test pulls in `parquet-hadoop`; the Iceberg tests used direct `ByteBuffer` construction. Not blocking, just checking whether the extra harness is needed here.

I’ll defer the per-descent validation cost to Fokko’s JMH thread. Other than the overflow line and test coverage, this looks close.

parquet-variant/pom.xml
steveloughran: I wanted to create the test files for https://github.com/apache/parquet-testing/pull/113 ; so while I agree it's overkill, it's a test only dependency and an easy way to generate test data for interop testing.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantJsonParser.java
steveloughran: That original 500 was completely chosen without any data as a general resilience number, the goal being "a malicious file can damage the stack of the thread processing it, but not the rest of the process". I went with the same number for the variant for consistency, and after a discussion in the mailing list went to 1k.

we haven't shipped the json parser yet, so no need for a release note. I will look at #3415 notes to see if we should update it though

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
steveloughran: will add to the javadocs, with a link to the discussion https://lists.apache.org/thread/q6wbom1q9pndv2nj6wcynxjcjxxkc1hm

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
steveloughran: cut it; all that matters is limit > pos, and as there's a limit - pos calculation on the line below, do a check on that value to make clear what is really being checked is "is there data"

parquet-variant/src/test/java/org/apache/parquet/variant/TestHardenedReader.java
steveloughran: ok, let's try to expand this, as it's the key one you are flagging up.

steveloughran:
...going through this, looks like claude wasn't that good, but it's highlighted that other bits of the existing code in `getArrayInfo()` and `getObjectInfo()` and flag the same issue. 

Rather than inserting (long) wherever I think it's needed, I'm actually putting readUnsigned() into a long wherever that value is coming from the stream, and only casting to an int after the validation is done.

now I'm going through all the variant logic looking for uses of the output of readUnsigned where the immediate action is some addition. And yes, it's pretty common, such as in 
`getMetadataKey()`.

```java
    int dataPos = offsetListPos + (dictSize + 1) * offsetSize;
    int offset = readUnsigned(metadata, offsetListPos + (id) * offsetSize, offsetSize);
    int nextOffset = readUnsigned(metadata, offsetListPos + (id + 1) * offsetSize, offsetSize);
    if (offset > nextOffset) {
      throw new IllegalStateException(String.format("Invalid offset: %d. next offset: %d", offset, nextOffset));
    }
    checkIndex(dataPos + nextOffset - 1, metadata.limit());
```

This then is going to be the code policy here

* all reads of data from that source are treated as malicious and potentially overflowing
* I'll have to audit them all
* using long as the int is actually a fairly effective way of tainting them in the codebase so they don't get used as array offsets or similar without validation

steveloughran:
went through the other uses, seems good. 

The one area not covered in `VariantUtil.getMetadataMap(ByteBuffer)`.

This is only used in the ImmutableMetadata(ByteBuffer) ctor, which is used in some parts of AvroVariantConverter.

* ignore that, assume it'll be zero risk (risk: being wrong)
* ignore that, deal with later (risk: forgetting. this is inevitable)
* cover it too.

I think I should cover it as well, just for completeness

steveloughran:
...that last patch (04f79) fixes ImmutableMetadata too; it's adjacent code and tests. I reviewed all uses of readUnsigned() and made sure they're being range checked before (Math.toIntExact()) for the casting, with > 0 check for signing. having just checked, this can be done in exactly one opcode in arm assembler (`tst  x0, #0xffffffff00000000`), so this java code with multiple reads and two chances to mispredict branches is woeful, but at least the normal branch will be to not fail.

steveloughran:
we had a discussion on this in the parquet meetup, then a slack chat with myself, ryan and neelesh afterwards

- unknown version behaviour reverted that of 1.17: no check, no rejection. This isn't great, but without a better versioning plan, at least it's consistent.
- unknown primary key, behaviour in this impl depends on how it is stored. Unshredded can be skipped, shredded triggers a rejection of the entire load as there's no way to reconstruct things

steveloughran:
conflicts to resolve.

steveloughran:
bad news for all reviewers; big rebase gone in to sync with junit and dependency changes. no squashing of commits, so the final changes are those from the recent review feedback.

steveloughran:
this is ready for final review and merge, btw

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -81,43 +104,59 @@ public Variant(byte[] value, int valuePos, int valueLength, byte[] metadata, int
...
-    // There is currently only one allowed version.
-    if ((metadata.get(metadata.position()) & VariantUtil.VERSION_MASK) != VariantUtil.VERSION) {
-      throw new UnsupportedOperationException(String.format(
-          "Unsupported variant metadata version: %d",
-          metadata.get(metadata.position()) & VariantUtil.VERSION_MASK));
-    }
+    this.depth = 0;
+    // A metadata buffer must contain at least the version byte
+    Preconditions.checkArgument(this.metadata.remaining() >= 1, "variant metadata is empty");
+    // There is no checking of version until a policy of how to handle mismatched versions
```
wgtmac, line 121: The Javadoc still says unsupported versions throw `UnsupportedOperationException`.

parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
```diff
@@ -880,6 +910,161 @@ static HashMap<String, Integer> getMetadataMap(ByteBuffer metadata) {
...
+    if (isObject) {
+      for (int i = 0; i < numElements; i++) {
+        int id = readUnsignedLittleEndian(valueBuffer, pos + (int) idStart + i * idSize, idSize);
+        Preconditions.checkArgument(
+            id < dictSize, "variant object key id %s out of range (dictSize=%s)", id, dictSize);
+      }
+    }
+    // Each child offset must lie within the data region. Children may overlap or leave gaps;
+    // the trailing terminator offset is range-checked for the same reason.
+    for (int i = 0; i <= numElements; i++) {
```
wgtmac, line 998: Could we bound this cost? Construction now scans every container entry, so single-element access to a large array becomes O(N), and repeated child access rescans.

wgtmac (commented):
I just have two comments about this change

Issue #3530, opened 2026-04-28
Apache Parquet Java Performance Improvements


iemejia:
Targeted performance optimizations to parquet-java's encoding, decoding, and compression layers. Each PR addresses an independent codec or subsystem, replacing stream-wrapper abstractions with direct ByteBuffer I/O, eliminating per-value allocations, and adding JMH benchmarks to validate the improvements.

All benchmarks: JDK 25.0.3 (Temurin), JMH 1.37, Linux x86_64, 100K values/iteration.

## Pull requests

| PR | Status | Scope | Headline result |
|----|--------|-------|-----------------|
| #3565 | Merged | PLAIN encoding/decoding | Decode 2.6–13x, encode 2–3.8x |
| #3566 | Open | DICTIONARY encoding/decoding | Encode 1.5–100x (high-card / long binary) |
| #3567 | Open | DELTA_BINARY_PACKED, DELTA_LENGTH_BYTE_ARRAY, DELTA_BYTE_ARRAY | Decode +22–31%, encode +3–7% |
| #3568 | Open | RLE/Bit-Packing hybrid codec | Decode +30%, encode +5% |
| #3569 | Merged | BYTE_STREAM_SPLIT | Decode 1.4–5.1x, encode 1.6–6.9x |
| #3570 | Open | Compression (Snappy, ZSTD, LZ4, GZIP, LZO, Brotli) | Isolated decompress +15–25% (Snappy/Zstd/Gzip), Snappy compress up to ~2.9x; enables LZ4/Brotli/LZO without native libs |
| #3571 | Merged | Row group flush buffer lifecycle | Correct resource management (not a perf win) |

PRs 1–6 and 8 are independent and can be reviewed/merged in parallel. Additional PRs for column I/O (par7) and level write batching (par9) will follow once their dependencies (#3565, #3568) land.

## Common optimization patterns

- **InputStream → ByteBuffer**: Direct `ByteBuffer` (LITTLE_ENDIAN) access for `getInt()`/`getLong()`/`getFloat()`/`getDouble()` JVM intrinsics instead of per-byte stream reads.
- **Buffer reuse**: Allocate once, grow lazily — no fresh allocation per page or per run.
- **pack32/unpack32 fast paths**: Batch 32 values per packer/unpacker call (4x fewer invocations).
- **Eliminate intermediate copies**: Write directly to output buffers, bypass stream wrappers.

### PLAIN (#3565)

| Benchmark | Master | Optimized | Speedup |
[...]

PR #3565, opened 2026-05-17
GH-3530: Optimize PLAIN encoding and decoding with direct ByteBuffer I/O

iemejia:

## Summary

Replace `ByteBufferInputStream` and `LittleEndianDataInputStream` wrappers with direct `ByteBuffer` access for all PLAIN value readers and writers.

**Readers** (`PlainValuesReader`, `BooleanPlainValuesReader`, `BinaryPlainValuesReader`, `FixedLenByteArrayPlainValuesReader`): hold a little-endian `ByteBuffer` from `initFromPage()` and call `getInt`/`getLong`/`getFloat`/`getDouble` directly, eliminating per-value stream overhead.

**Writers** (`PlainValuesWriter`, `BooleanPlainValuesWriter`, `FixedLenByteArrayPlainValuesWriter`): write through `CapacityByteArrayOutputStream`'s new `writeInt`/`writeLong` methods which put values directly into the NIO slab buffer in little-endian order, avoiding temporary byte-array allocation.

**Supporting changes**:
- `CapacityByteArrayOutputStream`: allocate slabs with `ByteOrder.LITTLE_ENDIAN`, add `writeInt(int)` and `writeLong(long)` for single-value NIO writes.
- `BytesInput`: add zero-copy `writeTo(ByteBuffer)` and `toByteArray()` using bulk `ByteBuffer.get()` instead of stream copy.
- `LittleEndianDataOutputStream`: batch single-byte writes into single `write(buf, 0, N)` calls for `writeShort`/`writeInt`.

Includes JMH benchmarks (`PlainEncodingBenchmark`, `PlainDecodingBenchmark`) covering all 7 primitive types for both encoding and decoding.

## Benchmark results

**Environment**: JDK 25.0.3 (Temurin), OpenJDK 64-Bit Server VM, JMH 1.37, Linux x86_64.

Decoding (100K values/iteration, 3 forks x 5 iterations, throughput mode):

| Benchmark | Master (M ops/s) | Branch (M ops/s) | Speedup |
|---|---:|---:|---:|
| decodeInt | 425 | 5,427 | **12.8x** |
| decodeFloat | 416 | 5,440 | **13.1x** |
| decodeLong | 119 | 4,720 | **39.5x** (\*) |
| decodeDouble | 116 | 6,026 | **51.8x** (\*) |
| decodeBoolean | 639 | 1,642 | **2.6x** |
| decodeFlba (len=2,12,16) | 188 | 680 | **3.6x** |
| decodeBinary (len=10,100,1000) | 142 | 225-230 | **1.6x** |

Encoding:

| Benchmark | Master (M ops/s) | Branch (M ops/s) | Speedup |
|---|---:|---:|---:|
| encodeInt | 148 | 559 | **3.8x** |
| encodeFloat | 150 | 532 | **3.5x** |
| encodeLong | 193 | 478 | **2.5x** |
| encodeDouble | 179 | 439 | **2.4x** |
| encodeBoolean | 850 | 1,692 | **2.0x** |
| encodeBinary (len=10) | 76 | 150 | **2.0x** |
| encodeFlba (len=2-16) | 156-184 | 178-224 | **1.1-1.2x** |

(\*) decodeLong/Double show JIT variance across forks (error bars >20%); true steady-state likely ~13x consistent with INT32/FLOAT.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
@@ -23,6 +23,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.column.values.ValuesReader;
+import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.parquet.io.api.Binary;
 
 /**
@@ -33,21 +34,32 @@ import org.apache.parquet.io.api.Binary;
  * {@code InputStream.read()} calls through {@link org.apache.parquet.bytes.BytesUtils#readIntLittleEndian}.
  */
 public class BinaryPlainValuesReader extends ValuesReader {
+  private static final ByteBuffer EMPTY_LE_BUFFER =
+      ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN).asReadOnlyBuffer();
+
   private ByteBuffer buffer;
 
   @Override
   public Binary readBytes() {
-    int length = buffer.getInt();
-    ByteBuffer valueSlice = buffer.slice();
-    valueSlice.limit(length);
-    buffer.position(buffer.position() + length);
-    return Binary.fromConstantByteBuffer(valueSlice);
+    try {
+      int length = buffer.getInt();
+      ByteBuffer valueSlice = buffer.slice();
+      valueSlice.limit(length);
+      buffer.position(buffer.position() + length);
+      return Binary.fromConstantByteBuffer(valueSlice);
+    } catch (RuntimeException e) {
+      throw new ParquetDecodingException("could not read bytes at offset " + buffer.position(), e);
+    }
   }
 
   @Override
   public void skip() {
-    int length = buffer.getInt();
-    buffer.position(buffer.position() + length);
+    try {
+      int length = buffer.getInt();
+      buffer.position(buffer.position() + length);
+    } catch (RuntimeException e) {
+      throw new ParquetDecodingException("could not skip bytes at offset " + buffer.position(), e);
+    }
   }
 
   @Override
@@ -56,7 +68,7 @@ public class BinaryPlainValuesReader extends ValuesReader {
     if (available > 0) {
       this.buffer = stream.slice(available).order(ByteOrder.LITTLE_ENDIAN);
     } else {
-      this.buffer = ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
+      this.buffer = EMPTY_LE_BUFFER.duplicate();
     }
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
@@ -19,44 +19,44 @@
 package org.apache.parquet.column.values.plain;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import org.apache.parquet.bytes.ByteBufferInputStream;
-import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.ValuesReader;
-import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.parquet.io.api.Binary;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
+/**
+ * Plain encoding reader for BINARY values.
+ *
+ * <p>Reads directly from a {@link ByteBuffer} with {@link ByteOrder#LITTLE_ENDIAN} byte order,
+ * using {@link ByteBuffer#getInt()} for the 4-byte length prefix instead of 4 individual
+ * {@code InputStream.read()} calls through {@link org.apache.parquet.bytes.BytesUtils#readIntLittleEndian}.
+ */
 public class BinaryPlainValuesReader extends ValuesReader {
-  private static final Logger LOG = LoggerFactory.getLogger(BinaryPlainValuesReader.class);
-  private ByteBufferInputStream in;
+  private ByteBuffer buffer;
 
   @Override
   public Binary readBytes() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      return Binary.fromConstantByteBuffer(in.slice(length));
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
-    }
+    int length = buffer.getInt();
+    ByteBuffer valueSlice = buffer.slice();
+    valueSlice.limit(length);
+    buffer.position(buffer.position() + length);
+    return Binary.fromConstantByteBuffer(valueSlice);
   }
 
   @Override
   public void skip() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      in.skipFully(length);
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not skip bytes at offset " + in.position(), e);
-    }
+    int length = buffer.getInt();
+    buffer.position(buffer.position() + length);
   }
 
   @Override
   public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IOException {
-    LOG.debug(
-        "init from page at offset {} for length {}",
-        stream.position(),
-        (stream.available() - stream.position()));
-    this.in = stream.remainingStream();
+    int available = stream.available();
+    if (available > 0) {
+      this.buffer = stream.slice(available).order(ByteOrder.LITTLE_ENDIAN);
+    } else {
+      this.buffer = ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
+    }
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BooleanPlainValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/plain/BooleanPlainValuesReader.java
@@ -23,6 +23,7 @@ import java.nio.ByteBuffer;
 import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.ValuesReader;
+import org.apache.parquet.io.ParquetDecodingException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,6 +41,7 @@ public class BooleanPlainValuesReader extends ValuesReader {
   private byte[] pageData;
   private int pageOffset;
   private int bitIndex;
+  private int bitCount;
 
   @Override
   public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IOException {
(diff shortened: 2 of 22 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -19,44 +19,44 @@
...
-  private ByteBufferInputStream in;
+  private ByteBuffer buffer;
 
   @Override
   public Binary readBytes() {
-    try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      return Binary.fromConstantByteBuffer(in.slice(length));
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
```
Fokko, line 40: Should we keep the `ParquetDecodingException`? Otherwise we're throwing the raw `{IOException,RuntimeException}` which is a behavioral change.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -19,44 +19,44 @@
...
-    LOG.debug(
-        "init from page at offset {} for length {}",
-        stream.position(),
-        (stream.available() - stream.position()));
-    this.in = stream.remainingStream();
+    int available = stream.available();
+    if (available > 0) {
+      this.buffer = stream.slice(available).order(ByteOrder.LITTLE_ENDIAN);
+    } else {
+      this.buffer = ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);
```
Fokko: Should we create a constant for the `ByteBuffer.allocate(0).order(ByteOrder.LITTLE_ENDIAN);`?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/FixedLenByteArrayPlainValuesReader.java
```diff
@@ -19,52 +19,56 @@
...
   public FixedLenByteArrayPlainValuesReader(int length) {
     this.length = length;
   }
 
   @Override
   public Binary readBytes() {
-    try {
-      return Binary.fromConstantByteBuffer(in.slice(length));
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not read bytes at offset " + in.position(), e);
```
Fokko, line 47: Same as above, should we keep the wrapped `ParquetDecodingException`?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
```diff
@@ -19,120 +19,91 @@
...
-  }
-
   public static class DoublePlainValuesReader extends PlainValuesReader {
 
     @Override
     public void skip(int n) {
-      try {
-        skipBytesFully(n * 8);
-      } catch (IOException e) {
-        throw new ParquetDecodingException("could not skip " + n + " double values", e);
```
Fokko, line 62: Same here, do we want to keep the `ParquetDecodingException`?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
```diff
@@ -19,120 +19,91 @@
...
+ *
+ * <p>Reads directly from a {@link ByteBuffer} with {@link ByteOrder#LITTLE_ENDIAN} byte order,
+ * bypassing the deprecated {@code LittleEndianDataInputStream} wrapper to avoid per-value virtual
+ * dispatch overhead. The underlying page data is obtained as a single contiguous {@link ByteBuffer}
+ * via {@link ByteBufferInputStream#slice(int)}.
  */
 public abstract class PlainValuesReader extends ValuesReader {
   private static final Logger LOG = LoggerFactory.getLogger(PlainValuesReader.class);
 
-  protected LittleEndianDataInputStream in;
```
Fokko, line 35: We should go through the deprecation cycle here, but is anything using this outside of the project itself?

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -643,6 +673,20 @@ public ByteBuffer toByteBuffer() throws IOException {
...
     }
 
+    /**
+     * Zero-copy override: returns the backing array directly when fully used,
+     * skipping the base-class BAOS allocation + copy on every decompressor call.
+     * Returning the mutable array is safe — the base class already exposes a
+     * mutable {@code BAOS.getBuf()}.
+     */
+    @Override
+    public byte[] toByteArray() {
```
Fokko, line 701: This overrides a deprecated API, as a follow-up we probably should move the internal calls to the new API: 

```java
@deprecated Use {@link #toByteBuffer(ByteBufferAllocator, Consumer)}
```

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BooleanPlainValuesReader.java
```diff
@@ -18,57 +18,65 @@
...
+  public void skip() {
+    bitIndex++;
   }
 
-  @Deprecated
   @Override
-  public int getNextOffset() {
-    return in.getNextOffset();
+  public void skip(int n) {
+    bitIndex += n;
```
Fokko, line 87: Should we check for bounds, and throw a `ParquetDecodingException` in case of out of bounds?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
```diff
@@ -19,120 +19,91 @@
...
   public static class DoublePlainValuesReader extends PlainValuesReader {
 
     @Override
     public void skip(int n) {
-      try {
-        skipBytesFully(n * 8);
-      } catch (IOException e) {
-        throw new ParquetDecodingException("could not skip " + n + " double values", e);
-      }
+      buffer.position(buffer.position() + n * 8);
```
Fokko: Should we use `Math.multiplyExact` here and below?

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
iemejia: Done. Added try/catch wrapping `RuntimeException` (which covers `BufferUnderflowException`, `IllegalArgumentException`, etc.) into `ParquetDecodingException` in both `readBytes()` and `skip()`.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
iemejia: Done. Extracted `EMPTY_LE_BUFFER` as a `private static final` read-only constant (`ByteBuffer.allocate(0).order(LITTLE_ENDIAN).asReadOnlyBuffer()`), used via `.duplicate()` in `initFromPage`. Same pattern applied in `PlainValuesReader`.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/FixedLenByteArrayPlainValuesReader.java
iemejia: Done. Added try/catch wrapping in `readBytes()`, `skip()`, and `skip(int n)`. Also using `Math.multiplyExact(n, length)` in `skip(int n)` to detect overflow.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
iemejia: Done. All `readXxx()` and `skip(int n)` methods across DoublePlainValuesReader, FloatPlainValuesReader, IntegerPlainValuesReader, and LongPlainValuesReader now wrap `RuntimeException` in `ParquetDecodingException` with descriptive messages matching the original error contract.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
iemejia: Good point. The old `protected LittleEndianDataInputStream in` field is now `protected ByteBuffer buffer` — the type change is binary-incompatible regardless, so a deprecation cycle wouldn't help external subclasses (they'd get a compile error either way). I searched the project and only internal subclasses (the 4 inner classes) access this field. I think this is acceptable given this class was never annotated `@Public` and the field type change makes deprecation impractical. WDYT?

parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
iemejia: Addressed in f0bdac633. The base-class `toInputStream()` now tries `getInternalByteBuffer()` first (zero-copy fast path) before falling back to the deprecated `toByteBuffer()`. Also added `getInternalByteBuffer()` and `toInputStream()` overrides to `ByteArrayBytesInput` so the byte-array-backed path is zero-copy too. Added `@SuppressWarnings("deprecation")` on the intentional overrides.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BooleanPlainValuesReader.java
iemejia: Done. Added a `bitCount` field (set to `length * 8` in `initFromPage`) and an explicit bounds check in `readBoolean()` that throws `ParquetDecodingException` with a descriptive message when attempting to read beyond the page boundary.

parquet-column/src/main/java/org/apache/parquet/column/values/plain/PlainValuesReader.java
iemejia: Done. Applied `Math.multiplyExact` in all `skip(int n)` methods: `Math.multiplyExact(n, 8)` for double/long, `Math.multiplyExact(n, 4)` for float/int, and `Math.multiplyExact(n, length)` in `FixedLenByteArrayPlainValuesReader`. Overflow now produces an `ArithmeticException` which gets caught and wrapped in `ParquetDecodingException`.

iemejia:
@Fokko another one ready for a last review / approval :)

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
```diff
@@ -19,44 +19,56 @@
...
   @Override
   public void skip() {
     try {
-      int length = BytesUtils.readIntLittleEndian(in);
-      in.skipFully(length);
-    } catch (IOException | RuntimeException e) {
-      throw new ParquetDecodingException("could not skip bytes at offset " + in.position(), e);
+      int length = buffer.getInt();
+      buffer.position(buffer.position() + length);
+    } catch (RuntimeException e) {
```
Fokko, line 60: Should we include IOException as well?

Fokko (approved):

parquet-column/src/main/java/org/apache/parquet/column/values/plain/BinaryPlainValuesReader.java
iemejia: The original code caught `IOException | RuntimeException` because it read through `ByteBufferInputStream` / `BytesUtils.readIntLittleEndian(in)` which throw checked `IOException`. The new code uses `ByteBuffer.getInt()` and `ByteBuffer.position()` directly -- these only throw `RuntimeException` subclasses (`BufferUnderflowException`, `IllegalArgumentException`), so `IOException` is no longer possible. Adding `catch (IOException e)` would actually be a compile error since nothing in the try block declares it. The same `catch (RuntimeException e)` pattern is used consistently across all the other readers (`PlainValuesReader`, `FixedLenByteArrayPlainValuesReader`).

Fokko:
Let's move this forward, thanks @iemejia for working on this! This is an impressive speed-up!

iemejia:
Thank you @Fokko ! Great to see this landing and benefitting everyone.

PR #3569, opened 2026-05-17
GH-3530: Optimize BYTE_STREAM_SPLIT encoding/decoding

iemejia:

## Summary

Optimize scalar encode/decode for the BYTE_STREAM_SPLIT encoding.

**Reader**: Specialized transpose loops for element sizes 2/4/8/12/16 bytes plus generic fallback. Bulk array access when backing array is available.

**Writer**: Batched scatter buffers (`int[]`/`long[]` batches of 64) replacing per-value `scatterBytes()` which allocated temp `byte[]` and issued N single-byte writes.

Includes unit tests for transpose specializations, batch-boundary crossing, `getBufferedSize` with partial batches, direct ByteBuffer decode paths, and close/reset with pending unflushed batches.

JMH benchmarks: `BssEncodingBenchmark`, `BssDecodingBenchmark` covering FLOAT, DOUBLE, INT32, INT64, and FIXED_LEN_BYTE_ARRAY.

## Benchmark results

**Environment**: JDK 25.0.3 (Temurin), OpenJDK 64-Bit Server VM, JMH 1.37, Linux x86_64, `-wi 3 -i 5 -f 2`.

Decoding:

| Benchmark | Baseline (M ops/s) | Optimized (M ops/s) | Speedup |
|---|---:|---:|---:|
| decodeInt | 203 | 956 | **4.7x** |
| decodeFloat | 263 | 965 | **3.7x** |
| decodeFloatDirect | 263 | 693 | **2.6x** |
| decodeDouble | 132 | 341 | **2.6x** |
| decodeLong | 133 | 323 | **2.4x** |
| decodeLongDirect | 149 | 282 | **1.9x** |
| decodeFlba(2) | 286 | 449 | **1.6x** |
| decodeFlba(7) | 125 | 169 | **1.4x** |
| decodeFlba(12) | 95 | 176 | **1.9x** |
| decodeFlba(16) | 78 | 132 | **1.7x** |

Encoding:

| Benchmark | Baseline (M ops/s) | Optimized (M ops/s) | Speedup |
|---|---:|---:|---:|
| encodeLong | 52 | 342 | **6.6x** |
| encodeDouble | 53 | 305 | **5.8x** |
| encodeFloat | 101 | 452 | **4.5x** |
| encodeInt | 99 | 439 | **4.4x** |
| encodeFlba(12) | 41 | 97 | **2.4x** |
| encodeFlba(16) | 32 | 70 | **2.2x** |
| encodeFlba(7) | 69 | 146 | **2.1x** |
| encodeFlba(2) | 192 | 276 | **1.4x** |

Every benchmark shows clear improvement with no regressions. 8-byte types benefit most from the batched scatter (6.6x) since the baseline scattered 8 bytes per value into 8 separate streams. The new `Direct` benchmarks validate the direct-buffer decode path.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
@@ -52,7 +52,7 @@ public abstract class ByteStreamSplitValuesReader extends ValuesReader {
   // Decode an entire data page by transposing from stream-split layout to interleaved layout.
   private byte[] decodeData(ByteBuffer encoded, int valuesCount) {
     int totalBytes = valuesCount * elementSizeInBytes;
-    assert encoded.remaining() >= totalBytes;
+    assert encoded.remaining() == totalBytes;
 
     // Bulk access: use the backing array directly if available, otherwise copy once.
     byte[] src;
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
@@ -49,17 +49,138 @@ public abstract class ByteStreamSplitValuesReader extends ValuesReader {
     return offset;
   }
 
-  // Decode an entire data page
+  // Decode an entire data page by transposing from stream-split layout to interleaved layout.
   private byte[] decodeData(ByteBuffer encoded, int valuesCount) {
-    assert encoded.limit() == valuesCount * elementSizeInBytes;
-    byte[] decoded = new byte[encoded.limit()];
-    int destByteIndex = 0;
-    for (int srcValueIndex = 0; srcValueIndex < valuesCount; ++srcValueIndex) {
-      for (int stream = 0; stream < elementSizeInBytes; ++stream, ++destByteIndex) {
-        decoded[destByteIndex] = encoded.get(srcValueIndex + stream * valuesCount);
+    int totalBytes = valuesCount * elementSizeInBytes;
+    assert encoded.remaining() >= totalBytes;
+
+    // Bulk access: use the backing array directly if available, otherwise copy once.
+    byte[] src;
+    int srcBase;
+    if (encoded.hasArray()) {
+      src = encoded.array();
+      srcBase = encoded.arrayOffset() + encoded.position();
+    } else {
+      src = new byte[totalBytes];
+      encoded.get(src);
+      srcBase = 0;
+    }
+
+    byte[] decoded = new byte[totalBytes];
+
+    // Specialized single-pass loops for common element sizes.
+    if (elementSizeInBytes == 2) {
+      int s0 = srcBase, s1 = srcBase + valuesCount;
+      for (int i = 0; i < valuesCount; ++i) {
+        int di = i * 2;
+        decoded[di] = src[s0 + i];
+        decoded[di + 1] = src[s1 + i];
+      }
+    } else if (elementSizeInBytes == 4) {
+      int s0 = srcBase,
+          s1 = srcBase + valuesCount,
+          s2 = srcBase + 2 * valuesCount,
+          s3 = srcBase + 3 * valuesCount;
+      for (int i = 0; i < valuesCount; ++i) {
+        int di = i * 4;
+        decoded[di] = src[s0 + i];
+        decoded[di + 1] = src[s1 + i];
+        decoded[di + 2] = src[s2 + i];
+        decoded[di + 3] = src[s3 + i];
+      }
+    } else if (elementSizeInBytes == 8) {
+      int s0 = srcBase,
+          s1 = srcBase + valuesCount,
+          s2 = srcBase + 2 * valuesCount,
+          s3 = srcBase + 3 * valuesCount,
+          s4 = srcBase + 4 * valuesCount,
+          s5 = srcBase + 5 * valuesCount,
+          s6 = srcBase + 6 * valuesCount,
+          s7 = srcBase + 7 * valuesCount;
+      for (int i = 0; i < valuesCount; ++i) {
+        int di = i * 8;
+        decoded[di] = src[s0 + i];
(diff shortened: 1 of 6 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
```diff
@@ -49,17 +49,138 @@ protected int nextElementByteOffset() {
...
+  // Decode an entire data page by transposing from stream-split layout to interleaved layout.
   private byte[] decodeData(ByteBuffer encoded, int valuesCount) {
-    assert encoded.limit() == valuesCount * elementSizeInBytes;
-    byte[] decoded = new byte[encoded.limit()];
-    int destByteIndex = 0;
-    for (int srcValueIndex = 0; srcValueIndex < valuesCount; ++srcValueIndex) {
-      for (int stream = 0; stream < elementSizeInBytes; ++stream, ++destByteIndex) {
-        decoded[destByteIndex] = encoded.get(srcValueIndex + stream * valuesCount);
+    int totalBytes = valuesCount * elementSizeInBytes;
+    assert encoded.remaining() >= totalBytes;
```
Fokko:
```suggestion
    assert encoded.remaining() == totalBytes;
```

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -29,9 +28,23 @@
...
+   * Batch size for buffered scatter writes. Values are accumulated in a batch buffer
+   * and flushed as bulk {@code write(byte[], off, len)} calls to each stream, replacing
+   * N individual single-byte writes with one bulk write per stream per flush.
+   */
+  private static final int BATCH_SIZE = 64;
+
   protected final int numStreams;
   protected final int elementSizeInBytes;
-  private final CapacityByteArrayOutputStream[] byteStreams;
+  protected final CapacityByteArrayOutputStream[] byteStreams;
```
Fokko, line 40: Why not keep this one private?
```suggestion
  private final CapacityByteArrayOutputStream[] byteStreams;
```

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -187,7 +260,49 @@ public FixedLenByteArrayByteStreamSplitValuesWriter(
...
+      if (flbaBatchCount == BATCH_SIZE) {
+        flushFlbaBatch();
+      }
+    }
+
+    private void flushFlbaBatch() {
+      if (flbaBatchCount == 0) return;
+      final int count = flbaBatchCount;
+      for (int stream = 0; stream < length; stream++) {
+        byteStreams[stream].write(batchBufs[stream], 0, count);
```
Fokko: I see, probaly we want to make a method for this one:

```
writeToStream(stream, batchBufs[stream], 0, count);
```

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
```diff
@@ -76,27 +91,83 @@ public Encoding getEncoding() {
...
+    }
+  }
+
+  private void flushBatch() {
+    if (batchCount == 0) return;
+    if (intBatch != null) {
+      flushIntBatch();
+    } else if (longBatch != null) {
+      flushLongBatch();
+    }
```
Fokko, line 147: Should we add an else case that throws?

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesReader.java
iemejia: Done in 689208f99.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
iemejia: Can't make it `private`: `FixedLenByteArrayByteStreamSplitValuesWriter` is a static nested class that extends the parent, and Java's name resolution can't reach private parent members from that context (private methods/fields aren't inherited, and static nested classes have no enclosing-instance reference to use for cross-class access). Additionally, japicmp flags the visibility reduction as an API-breaking change since these fields are `protected` in the released version. Left as-is.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
iemejia: Done in 689208f99. Extracted a package-private `writeToStream(int stream, byte[] buf, int off, int len)` helper. It's package-private (not private) because the `FixedLenByteArrayByteStreamSplitValuesWriter` static nested subclass also needs to call it — private methods aren't inherited and can't be resolved from a static nested class that extends the enclosing class. All three flush methods (`flushIntBatch`, `flushLongBatch`, `flushFlbaBatch`) now use it consistently.

parquet-column/src/main/java/org/apache/parquet/column/values/bytestreamsplit/ByteStreamSplitValuesWriter.java
iemejia: Done in 689208f99. Added an else-throw with `ParquetEncodingException` for fail-fast if `batchCount > 0` but no batch buffer is initialized.

iemejia:
@Fokko All review comments addressed in 689208f99. Benchmark results updated in the PR description (re-run on JDK 25.0.3, same machine). Ready for another round of review or merge.

Fokko (approved):

iemejia:
I had to rebase this one to solve some merge issues.. This should be good to merged once the tests are green again @Fokko

PR #3570, opened 2026-05-17
GH-3530: Bypass Hadoop codec abstraction to optimize compression performance

iemejia:

## Summary

Bypass the Hadoop `CompressionCodec` / `CodecPool` / stream-wrapper machinery for all supported codecs, compressing/decompressing each page by calling the underlying library directly, in both `CodecFactory` (heap) and `DirectCodecFactory` (direct memory).

| Codec | Before | After |
|---|---|---|
| Snappy | Hadoop `SnappyCodec` stream wrappers | snappy-java direct (heap) |
| ZSTD | Streaming `ZstdOutputStream`/`ZstdInputStream` | Reusable `ZstdCompressCtx`/`ZstdDecompressCtx` one-shot |
| LZ4_RAW | Hadoop codec abstraction | aircompressor LZ4, reusable direct ByteBuffers |
| LZ4 (legacy) | Hadoop `Lz4Codec` (native) | aircompressor Hadoop-framed LZ4 streams |
| GZIP | Hadoop `GzipCodec` + codec-pool | JDK `GZIPOutputStream`/`GZIPInputStream`, 64 KB buffers |
| LZO | GPL `com.hadoop.compression.lzo.LzoCodec` | aircompressor `LzoHadoopStreams` (Apache 2.0, wire-compatible) |
| Brotli | abandoned `brotli-codec` (x86-only) | `brotli4j` (aarch64 + x86, reflection-loaded, runtime-optional) |

Per-column compression levels (PARQUET-3459) are integrated into the direct path. Unsupported codecs now throw instead of silently falling back. japicmp-verified with deprecated shims retained.

**Portability / licensing:** LZ4, Brotli, LZO now work without native Hadoop codecs or GPL deps (previously threw `native lz4 library not available` / `ClassNotFoundException: BrotliCodec` / `LzoCodec`); Brotli gains aarch64.

## Benchmarks

- **`CompressionBenchmark`** isolates the codec hot path per codec / page size / heap-direct factory. Data shapes are **real Parquet pages** — realistic typed values run through Parquet's own `ValuesWriter` encoders — so the compressor sees production bytes. 16 shapes span every non-deprecated encoding (PLAIN, dictionary+RLE, DELTA_BINARY_PACKED, DELTA_BYTE_ARRAY, DELTA_LENGTH_BYTE_ARRAY, BYTE_STREAM_SPLIT, RLE) across int32/int64/float/double/FLBA/binary/boolean; selection is by distinct byte distribution (documented in-class).
- **`WriteBenchmarks`/`ReadBenchmarks`** — end-to-end 1 M-row write/read per codec; write also reports file size via `@AuxCounters`.

### Results

*Isolated `CompressionBenchmark` (JMH 1.37, JDK 17, Linux x86_64, HEAP, 1 MB pages).* Throughput vs the Hadoop path, geomean over the 16 encoding shapes (>1 = faster):

| Codec | compress | decompress |
|---|--:|--:|
| SNAPPY | **1.38×** | **1.25×** |
| ZSTD | 1.00× | **1.18×** |
| GZIP | 1.00× | **1.15×** |
| LZ4_RAW | 0.96× | 1.06× |

- Snappy peaks near **2.9× compress / 1.8× decompress** on small, highly compressible pages where wrapper/pool overhead dominated.
- Zstd and Gzip decompress are uniformly faster; the Gzip 64 KB buffer removes the JDK 512 B per-page overhead.
- LZ4_RAW compress has a small direct-buffer trade-off on high-throughput pages; parity/faster elsewhere.

**How much of end-to-end write/read is compression?** Measured on the `WriteBenchmarks`/`ReadBenchmarks` dataset, the codec accounts for roughly:

| Codec | % of write | % of read |
|---|--:|--:|
| SNAPPY | ~1% | ~0% |
| LZ4_RAW | ~3% | ~3% |
| ZSTD | ~10% | ~2% |
| GZIP | ~46% | ~25% |

so the isolated codec gains translate into a **small end-to-end impact** for the fast codecs, with more room to move on GZIP- and ZSTD-heavy workloads.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
@@ -19,44 +19,58 @@
 package org.apache.parquet.hadoop;
 
 import com.github.luben.zstd.Zstd;
+import com.github.luben.zstd.ZstdCompressCtx;
+import com.github.luben.zstd.ZstdDecompressCtx;
+import io.airlift.compress.lz4.Lz4Compressor;
+import io.airlift.compress.lz4.Lz4Decompressor;
+import io.airlift.compress.lz4.Lz4HadoopStreams;
+import io.airlift.compress.lzo.LzoHadoopStreams;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
+import java.util.zip.Deflater;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.io.compress.CodecPool;
 import org.apache.hadoop.io.compress.CompressionCodec;
-import org.apache.hadoop.io.compress.CompressionOutputStream;
-import org.apache.hadoop.io.compress.Compressor;
-import org.apache.hadoop.io.compress.Decompressor;
-import org.apache.hadoop.io.compress.zlib.ZlibCompressor;
 import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.ByteBufferAllocator;
+import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.compression.CompressionCodecFactory;
 import org.apache.parquet.conf.HadoopParquetConfiguration;
 import org.apache.parquet.conf.ParquetConfiguration;
-import org.apache.parquet.hadoop.codec.Lz4RawCodec;
 import org.apache.parquet.hadoop.codec.ZstandardCodec;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.hadoop.util.ConfigurationUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.xerial.snappy.Snappy;
 
 public class CodecFactory implements CompressionCodecFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
 
+  /**
+   * Cache of Hadoop {@link CompressionCodec} instances keyed by codec class name (and optional
+   * compression level). Retained for backwards compatibility with subclasses that override
+   * {@link #createCompressor(CompressionCodecName)} / {@link #createDecompressor(CompressionCodecName)}
+   * and rely on {@link #getCodec(CompressionCodecName)}. The default implementation no longer uses
+   * the Hadoop codec abstraction and does not populate this map.
+   *
+   * @deprecated the default implementation no longer resolves codecs through Hadoop; this field is
+   *     kept only for binary/source compatibility and will be removed in 2.0.0.
+   */
+  @Deprecated
   protected static final Map<String, CompressionCodec> CODEC_BY_NAME =
       Collections.synchronizedMap(new HashMap<String, CompressionCodec>());
-
-  static final String GZIP_COMPRESS_LEVEL = "zlib.compress.level";
-  static final String BROTLI_COMPRESS_QUALITY = "compression.brotli.quality";
 
   private final Map<String, BytesCompressor> compressors = new HashMap<>();
   private final Map<CompressionCodecName, BytesDecompressor> decompressors = new HashMap<>();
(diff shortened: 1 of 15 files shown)


iemejia:
Hi @wgtmac,

Sorry to bother you. I noticed you merged PARQUET-3459, so I thought you might be a good person to review this PR given your familiarity with Parquet's compression paths. It is a bit of a big PR but I tried to do my best to explain it clearly and document it in detail.

The benchmarks look promising. Besides the performance improvements, this change also reduces direct Hadoop dependencies, which has been discussed as an important maintenance goal by @julienledem in Parquet maintainer meetings. Of course I am not removing those because of backwards compatibility but it will be easy to do if we ever decide to.

If you have a chance to take a look, I'd really appreciate it. It's been waiting for review for quite some time, and I'd value your feedback. Thanks!

wgtmac:
@iemejia Sorry I didn't make it to review this earlier but I agree it looks promising. Could you help rebase it to resolve merge conflict first?

iemejia:
@wgtmac Just rebased it and added some correctness fixes. Please let me know on the best way to proceed. This is quite a big review, and sometimes I hesitated to split it but I don't know if that would be a lot more work also for the reviewer. Just let me know your preference.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -503,4 +564,764 @@ public abstract void decompress(ByteBuffer input, int compressedSize, ByteBuffer
...
+
+    @Override
+    public void decompress(ByteBuffer input, int compressedSize, ByteBuffer output, int decompressedSize)
+        throws IOException {
+      ByteBuffer inputSlice = input.slice();
+      inputSlice.limit(compressedSize);
+      ByteBuffer outputSlice = output.slice();
+      outputSlice.limit(decompressedSize);
+      int decompressed = Snappy.uncompress(inputSlice, outputSlice);
+      if (decompressed != decompressedSize) {
```
wgtmac, line 712: This ByteBuffer overload also accepts heap buffers, but Snappy requires direct buffers here. The ZSTD path below has the same issue. Could we keep heap buffers working and add a test for both codecs?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/DirectCodecFactory.java
```diff
@@ -103,8 +104,14 @@ protected BytesCompressor createCompressor(final CompressionCodecName codecName)
         return new SnappyCompressor();
       case ZSTD:
         return new ZstdCompressor();
-      // todo: create class similar to the SnappyCompressor for zlib and exclude it as
-      // snappy is above since it also generates allocateDirect calls.
+      case LZ4_RAW:
+        return new Lz4RawCompressor();
+      case BROTLI:
+        if (Brotli4j.AVAILABLE) {
+          return new BrotliDirectCompressor(conf.getInt("compression.brotli.quality", 1));
```
wgtmac, line 111: The level-aware BROTLI path falls back to the heap compressor, while the no-level path uses the direct compressor. Could we keep this path direct when brotli4j is available?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -292,22 +314,67 @@ public BytesDecompressor getDecompressor(CompressionCodecName codecName) {
...
+      case GZIP:
+        validateGzipLevel(level);
+        return new GzipBytesCompressor(level, pageSize);
+      case BROTLI:
+        if (Brotli4j.AVAILABLE) {
+          validateBrotliLevel(level);
+          return new BrotliBytesCompressor(level);
+        }
+        throw new UnsupportedOperationException(
+            "BROTLI codec requires brotli4j on the classpath (com.aayushatharva.brotli4j)");
```
wgtmac, line 369: The no-level BROTLI path can still use the Hadoop codec when brotli4j is missing. This level-aware path throws instead. Could we keep the fallback for callers that set a Brotli level?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/CodecFactory.java
```diff
@@ -503,4 +564,764 @@ public abstract void decompress(ByteBuffer input, int compressedSize, ByteBuffer
...
+
+    LzoBytesCompressor(int pageSize) {
+      this.baos = new ByteArrayOutputStream(pageSize);
+    }
+
+    @Override
+    public BytesInput compress(BytesInput bytes) throws IOException {
+      baos.reset();
+      try (OutputStream los = LZO_STREAMS.createOutputStream(baos)) {
+        bytes.writeAllTo(los);
```
wgtmac, line 1079: This path ignores the existing Hadoop LZO strategy, level, and buffer-size settings. Could we preserve those settings where possible, or document the change and test old files made with non-default settings?

wgtmac (commented):
I just used Codex to perform the first round of review and spotted these issues. Let me know what you think.

PR #3571, opened 2026-05-17
GH-3530: Eagerly release column buffers during row group flush

iemejia:

## Summary

Eagerly release column buffers during row group flush for correct resource management.

Call `pageWriter.close()` after each column in `flushToFileWriter()` instead of holding all buffers until the entire flush completes. Add `writeAllToAndRelease()` to `ConcatenatingByteBufferCollector` for progressive slab-by-slab memory release during write. Make `close()` idempotent (safe after eager release or double-close).

### Changes
- Call `pageWriter.close()` after each column in `flushToFileWriter()`
- Add `writeAllToAndRelease()` to `ConcatenatingByteBufferCollector` for progressive slab-by-slab memory release during write
- Make `close()` idempotent (safe to call after eager release or double-close)
- Add `RowGroupFlushBenchmark` (20-column wide schema, `PeakTrackingAllocator`) and `BlackHoleOutputFile` for measuring flush performance and peak memory
- Add tests for eager release, double-close safety, and output equivalence

## Benchmark results

**Environment**: JDK 25.0.3 (Temurin), OpenJDK 64-Bit Server VM, JMH 1.37, Linux x86_64, `-wi 3 -i 5 -f 2`, `-Xms512m -Xmx1g`.

`RowGroupFlushBenchmark` (100K rows, 20 BINARY columns, 200 bytes each, UNCOMPRESSED):

| Row Group Size | Metric | Baseline | Optimized | Delta |
|---:|---|---:|---:|---:|
| 8 MB | Flush time (ms/op) | 319 ± 11 | 322 ± 10 | ~0% (noise) |
| 8 MB | Peak allocator (MB) | 9.4 | 9.4 | same |
| 64 MB | Flush time (ms/op) | 330 ± 12 | 327 ± 11 | ~0% (noise) |
| 64 MB | Peak allocator (MB) | 66.0 | 66.0 | same |

No throughput regression. Peak memory is byte-for-byte identical with and without this change — the peak is reached during page compression (as pages are accumulated into `ConcatenatingByteBufferCollector`), not during flush. This is correct resource management that makes buffers GC-eligible sooner and guarantees release on error paths via try-with-resources, but is not a memory optimization.

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
@@ -26,6 +26,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -64,10 +65,14 @@ public class ConcatenatingByteBufferCollector extends BytesInput implements Auto
 
   @Override
   public void close() {
+    if (slabs.isEmpty()) {
+      return;
+    }
     for (ByteBuffer slab : slabs) {
       allocator.release(slab);
     }
     slabs.clear();
+    size = 0;
   }
 
   @Override
@@ -76,6 +81,30 @@ public class ConcatenatingByteBufferCollector extends BytesInput implements Auto
     for (ByteBuffer buffer : slabs) {
       channel.write(buffer.duplicate());
     }
+  }
+
+  /**
+   * Writes all collected slabs to the given output stream, releasing each slab's
+   * {@link ByteBuffer} back to the allocator immediately after it has been written.
+   * This progressively frees memory during the write rather than holding all slabs
+   * until {@link #close()} is called.
+   *
+   * <p>After this method returns, the collector is empty and {@link #size()} returns 0.
+   * Calling {@link #close()} afterwards is safe but has no additional effect.
+   *
+   * @param out the output stream to write to
+   * @throws IOException if an I/O error occurs
+   */
+  public void writeAllToAndRelease(OutputStream out) throws IOException {
+    WritableByteChannel channel = Channels.newChannel(out);
+    Iterator<ByteBuffer> it = slabs.iterator();
+    while (it.hasNext()) {
+      ByteBuffer slab = it.next();
+      channel.write(slab.duplicate());
+      allocator.release(slab);
+      it.remove();
+    }
+    size = 0;
   }
 
   @Override
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
@@ -75,14 +75,6 @@ public class ConcatenatingByteBufferCollector extends BytesInput implements Auto
     size = 0;
   }
 
-  @Override
-  public void writeAllTo(OutputStream out) throws IOException {
-    WritableByteChannel channel = Channels.newChannel(out);
-    for (ByteBuffer buffer : slabs) {
-      channel.write(buffer.duplicate());
-    }
-  }
-
   /**
    * Writes all collected slabs to the given output stream, releasing each slab's
    * {@link ByteBuffer} back to the allocator immediately after it has been written.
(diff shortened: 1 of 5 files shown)


parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
```diff
@@ -78,6 +83,30 @@ public void writeAllTo(OutputStream out) throws IOException {
...
+   * until {@link #close()} is called.
+   *
+   * <p>After this method returns, the collector is empty and {@link #size()} returns 0.
+   * Calling {@link #close()} afterwards is safe but has no additional effect.
+   *
+   * @param out the output stream to write to
+   * @throws IOException if an I/O error occurs
+   */
+  public void writeAllToAndRelease(OutputStream out) throws IOException {
+    WritableByteChannel channel = Channels.newChannel(out);
```
Fokko, line 95: Should we wrap a try-with-resource around this channel?

parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
```diff
@@ -78,6 +83,30 @@ public void writeAllTo(OutputStream out) throws IOException {
...
+   * This progressively frees memory during the write rather than holding all slabs
+   * until {@link #close()} is called.
+   *
+   * <p>After this method returns, the collector is empty and {@link #size()} returns 0.
+   * Calling {@link #close()} afterwards is safe but has no additional effect.
+   *
+   * @param out the output stream to write to
+   * @throws IOException if an I/O error occurs
+   */
+  public void writeAllToAndRelease(OutputStream out) throws IOException {
```
Fokko: Wait, are we just using this in tests?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -692,6 +692,11 @@ public void flushToFileWriter(ParquetFileWriter writer) throws IOException {
     for (ColumnDescriptor path : schema.getColumns()) {
       ColumnChunkPageWriter pageWriter = writers.get(path);
```
Fokko: Should we use the try-with-resource pattern here? (Fanboy of the pattern talking here)

parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
iemejia: Good catch — it was indeed only used in tests. I've merged `writeAllToAndRelease` into `writeAllTo` so the progressive buffer release is now the production implementation (called via `ParquetFileWriter.writeColumnChunk` → `bytes.writeAllTo(out)`). The separate method is gone.

parquet-common/src/main/java/org/apache/parquet/bytes/ConcatenatingByteBufferCollector.java
iemejia: No, closing the channel would close the underlying `OutputStream` which is owned by the caller. The channel from `Channels.newChannel(out)` is just a stateless wrapper with no independent resources to release. I've added a comment explaining this in the updated code.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
iemejia: Done. I am also a big fan of try-with-resources. This also ensures the column's buffers are released even if `writeToFileWriter` throws.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java
```diff
@@ -465,7 +466,7 @@ public void writeToFileWriter(ParquetFileWriter writer) throws IOException {
       if (LOG.isDebugEnabled()) {
         LOG.debug(String.format(
                 "written %,dB for %s: %,d values, %,dB raw, %,dB comp, %d pages, encodings: %s",
-                buf.size(),
+                bytesWritten,
```
Fokko, line 469: Why this change? We only need this when we have debug log enabled

iemejia: `buf.size()` is captured before the write because `writeColumnChunk` internally calls `buf.writeAllTo(out)` which progressively releases each slab and resets `size` to 0 (see `ConcatenatingByteBufferCollector.writeAllTo`, line 103). If we called `buf.size()` inside the debug block after the write, it would always report 0.

iemejia:
@Fokko All review comments addressed. Added baseline vs optimized benchmark comparison to the PR description — no throughput regression and identical peak memory. Ready for another look.

Fokko (approved):

iemejia:
@Fokko this one also has been there for a while. PTAL and tell me if anything else is missing. Otherwise it is probably good to merge.

PR #3575, opened 2026-05-22
GH-3574: parquet-hadoop: Statistics.toParquetStatistics: always set null_count

### LINKED ISSUE #3574: null_count is omitted for large columns in parquet files

Currently in [ParquetMetadataConverter.java](https://github.com/apache/parquet-java/blob/7be05b4702df78ae0c0c6b44adc6b7b7af2d931f/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java), there is a guard that prevents the writing of statistics such as min/max AND null_count when the stats are larger than the max allowed size under truncation. The rationale for this makes sense for omitting min/max, however null_count can be written on the file despite the size of its content. See the code below:

https://github.com/apache/parquet-java/blob/7be05b4702df78ae0c0c6b44adc6b7b7af2d931f/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L800-L807

The missing `null_count` metadata sometimes causes downstream consumers of the parquet files to error. For example in Snowflake we are seeing the following kind of error:

```
non-nullable column without default has null values according to file statistics
```

### Component(s)

Core

### LINKED ISSUE #4535: Iceberg null_counts bug for downstream consumers
See https://github.com/apache/parquet-java/issues/3574

mdibaiee:
### Rationale for this change

Missing `null_count` statistics for columns in parquet files can cause issues with downstream consumers of these files. It is not necessary to omit this statistic for columns which are larger than the truncation configuration, since despite the truncation, their nullability can be asserted with confidence. It is reasonable to keep omitting min/max statistics due to the rationale explained in the comment in the code.

Always add `null_count` statistics for columns in parquet files, unconditional of their size.

### Are these changes tested?

Yes, `TestParquetMetadataConverter.java` has been updated to reflect these changes

I think we can consider the additional `null_count` statistic's appearance as a user-facing change

Closes #3574

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -800,12 +800,14 @@ public class ParquetMetadataConverter {
   public static Statistics toParquetStatistics(
       org.apache.parquet.column.statistics.Statistics stats, int truncateLength) {
     Statistics formatStats = new Statistics();
+    if (!stats.isEmpty()) {
+      formatStats.setNull_count(stats.getNumNulls());
+    }
     // Don't write stats larger than the max size rather than truncating. The
     // rationale is that some engines may use the minimum value in the page as
     // the true minimum for aggregations and there is no way to mark that a
     // value has been truncated and is a lower bound and not in the page.
     if (!stats.isEmpty() && withinLimit(stats, truncateLength)) {
-      formatStats.setNull_count(stats.getNumNulls());
       if (stats.hasNonNullValue()) {
         byte[] min;
         byte[] max;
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -800,12 +800,12 @@ public static Statistics toParquetStatistics(org.apache.parquet.column.statistic
   public static Statistics toParquetStatistics(
       org.apache.parquet.column.statistics.Statistics stats, int truncateLength) {
     Statistics formatStats = new Statistics();
+    formatStats.setNull_count(stats.getNumNulls());
```
wgtmac: Do we need to preserve the check `stats.isEmpty()` for this?

mdibaiee: Good point, addressed 👍🏽

mdibaiee:
@wgtmac hey, thanks for the initial review. any chance of getting the workflows approved for running so the checks also run and we are closer to merging?

If any help from our side can help expedite this please let us know, as we have customers blocked on this and are happy to help as much as possible. Appreciate your time 🙇🏽

wgtmac (approved):
+1 Thanks @mdibaiee!

I'll leave it open for a few days just in case others have different opinion.

mdibaiee:
Thanks @wgtmac -- do you think sufficient time has passed and this can be merged now?

PR #3576, opened 2026-05-22
GH-3573: Fix VariantUtil string decoding to use explicit UTF-8 charset

### LINKED ISSUE #3573: VariantUtil decodes string bytes using platform default charset instead of UTF-8

I was playing with Variant implementation and trying to run corresponding tests, but got them failing. Turns out that VariantUtil.getString(), VariantUtil.getMetadataKey(), and VariantUtil.getMetadataMap() construct Java strings from raw bytes without specifying a charset. So this relies on the JVM platform default charset, which is strictly speaking not guaranteed to be UTF-8.

For example, on JDK <= 17, setting `LC_ALL=C` (the invocation recommended in the project README) causes the JVM to use ASCII. Under this encoding, multi-byte UTF-8 sequences - such as `é` (encoded as `0xC3 0xA9`) - are decoded as two separate characters instead of one, corrupting non-ASCII string values read from Variant columns.

The [Variant binary encoding spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) explicitly requires UTF-8 for all string data. As far as I can see the write path already uses `StandardCharsets.UTF_8` explicitly (in VariantBuilder.appendString() and MetadataBuilder.getOrInsert()), so this makes the mismatch on the read path both inconsistent and incorrect.

**To reproduce**, run the existing `testParseUnicodeString` test with `LC_ALL=C`:

```bash
LC_ALL=C ./mvnw test -pl parquet-variant -Dtest=TestVariantParseJson#testParseUnicodeString
```

Expected: test passes. 
Actual:
```
org.junit.ComparisonFailure: expected:<[él è]ve> but was:<[ÃlÃ¨]ve>
```

CI run tests on JDK 11 and 17 but does not set `LC_ALL=C`, so I guess `file.encoding` stays UTF-8 by default. JDK 18+ is also unaffected because JEP 400 makes UTF-8 the default regardless of locale.

**Proposed Fix:** add `StandardCharsets.UTF_8` to all six call sites, consistent with `Binary.toStringUsingUTF8()`, `KeyMetadata`, and the rest of the codebase.

### Component(s)

Core

mikhail-melnik:
### Rationale for this change

`VariantUtil.getString()`, `getMetadataKey()`, and `getMetadataMap()` used `new String(byte[], ...)` without a charset, relying on the JVM platform default. On JDK <= 17 with `LC_ALL=C` (the build invocation recommended in the README), the platform default becomes `US-ASCII`, causing multi-byte UTF-8 sequences in Variant string values and metadata keys to be decoded as multiple characters instead of one.

The [Variant binary encoding spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) mandates UTF-8 for all string data. The write path already uses `StandardCharsets.UTF_8` explicitly, this PR aligns the read path to match.

`VariantUtil.java`: add `StandardCharsets.UTF_8` to six `new String(byte[], ...)` call sites across `getString()`, `getMetadataKey()`, and `getMetadataMap()`.

### Are these changes tested?

Yes, the existing `testParseUnicodeString` test in `TestVariantParseJson` covers this. It was already written to catch this exact case but only failed when running under `LC_ALL=C` (i.e. on JDK <= 17 with the README-recommended build invocation). With this fix, `LC_ALL=C ./mvnw test -pl parquet-variant` passes cleanly.

No API changes. Behavior change only for JVMs running with a non-UTF-8 default charset, where string values containing non-ASCII characters were previously silently corrupted on read.

Closes #3573

### DIFF
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantUtil.java
@@ -20,6 +20,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.HashMap;
 import org.apache.parquet.Preconditions;
@@ -657,12 +658,12 @@ class VariantUtil {
       checkIndex(start + length - 1, value.limit());
       if (value.hasArray()) {
         // If the buffer is backed by an array, we can use the array directly.
-        return new String(value.array(), value.arrayOffset() + start, length);
+        return new String(value.array(), value.arrayOffset() + start, length, StandardCharsets.UTF_8);
       } else {
         // If the buffer is not backed by an array, we need to copy the bytes into a new array.
         byte[] valueArray = new byte[length];
         slice(value, start).get(valueArray);
-        return new String(valueArray);
+        return new String(valueArray, StandardCharsets.UTF_8);
       }
     }
     throw unexpectedType(Variant.Type.STRING, value);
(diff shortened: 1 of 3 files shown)


steveloughran (commented):
ooh, this is bad!

Looks like org.apache.iceberg.variants.VariantUtil uses UTF_8, but it's worth worrying about the others.

Have you considered adding a test which uses a non-ascii key in a variant and verify it then resolves?

mikhail-melnik:
Hi @steveloughran, thanks for looking into this PR! I've added 2 tests - one for getMetadataKey (via getFieldByKey call) and one for getMetadataMap (via creating ImmutableMetadata), should have more or less complete coverage now.

wgtmac (approved):

steveloughran (approved):
LGTM, needs a review by a committer though

Fokko (approved):
Oof, great catch @mikhail-melnik

Fokko:
Thanks for the review @wgtmac, @steveloughran and @copilot

steveloughran:
You don't have to thank copilot. It's like being grateful to a washing machine for cleaning your clothes

PR #3579, opened 2026-05-23
Bump minimum Java version to 17 and remove pre-17 workarounds

iemejia:
## Summary

Raise the minimum build/runtime requirement to **Java 17** and remove workarounds and tooling that are no longer needed. The Hadoop 3.4.x bump and the vectored-I/O buffer fix that previously lived here have been split out (the Hadoop upgrade will come via #3559).

### Java 17 baseline
- Set `maven.compiler.release` to 17 and add an enforcer rule requiring Java 17+ to build.
- Remove the `extraJavaTestArgs` (`--add-opens` / `-XX` flags) that were needed to run the tests on newer JDKs from a Java 11 baseline.
- Drop the per-module `maven.compiler.release=17` override in `parquet-encoding-vector`; it is now the project-wide default.
- Simplify the `CleanUtil` javadoc.
- CI: build only on JDK 17 (drop the JDK 11 matrix entry) and stop exporting `extraJavaTestArgs`.
- Update the README to state the Java 17 requirement.

### Spotless upgrade (2.46.1 -> 3.8.0)
Spotless 2.46.1 calls `com.sun.tools.javac.util.Log` methods that were removed in JDK 25, causing `NoSuchMethodError` during formatting. Spotless 3.8.0 is compatible with JDK 25. The switch/case comment indentation changes are produced by the new formatter version.

### Remove obsolete japicmp exclude for `IndexIterator`
The exclude worked around a false-positive source incompatibility reported by japicmp 0.23.1 on the synthetic/bridge methods of `PrimitiveIterator.OfInt` (#3388). japicmp was upgraded to 0.26.1 (#3592), which handles those methods correctly, so the workaround is no longer needed. Verified that `japicmp:cmp` (with `breakBuildOnSourceIncompatibleModifications=true`) passes with the exclude removed.

### Split out (deferred to #3559)
The Hadoop `3.3.0 -> 3.4.3` bump and the related vectored-I/O `ByteBuffer` leak fix have been removed from this PR. The Java 17 build passes without them (vectored-I/O tests simply skip on Hadoop 3.3.x, with no buffer leaks), so they are not required here and are better handled alongside #3559.

### Testing
- Verified on JDK 17: full `spotless:check`, `japicmp:cmp` (with break-on-incompatible), and `parquet-hadoop` compile/tests all pass.
- Verified the Java 17 build passes with Hadoop 3.3.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
   </mailingLists>
 
   <properties>
-    <maven.compiler.release>11</maven.compiler.release>
+    <maven.compiler.release>17</maven.compiler.release>
     <github.global.server>github</github.global.server>
     <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
     <jackson.datatype.groupId>com.fasterxml.jackson.datatype</jackson.datatype.groupId>
@@ -80,7 +80,7 @@
     <jackson-annotations.version>2.22</jackson-annotations.version>
     <japicmp.version>0.26.1</japicmp.version>
     <javax.annotation.version>1.3.2</javax.annotation.version>
-    <spotless.version>2.46.1</spotless.version>
+    <spotless.version>3.8.0</spotless.version>
     <shade.prefix>shaded.parquet</shade.prefix>
     <!-- Guarantees no newer classes/methods/constants are used by parquet. -->
     <hadoop.version>3.3.0</hadoop.version>
@@ -122,23 +122,6 @@
 
     <!-- Resource intesive tests are enabled by default but disabled in the CI envrionment -->
     <enableResourceIntensiveTests>true</enableResourceIntensiveTests>
-
-    <extraJavaTestArgs>
-      -XX:+IgnoreUnrecognizedVMOptions
-      --add-opens=java.base/java.lang=ALL-UNNAMED
-      --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
-      --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
-      --add-opens=java.base/java.io=ALL-UNNAMED
-      --add-opens=java.base/java.net=ALL-UNNAMED
-      --add-opens=java.base/java.nio=ALL-UNNAMED
-      --add-opens=java.base/java.util=ALL-UNNAMED
-      --add-opens=java.base/java.util.concurrent=ALL-UNNAMED
-      --add-opens=java.base/java.util.concurrent.atomic=ALL-UNNAMED
-      --add-opens=java.base/sun.nio.ch=ALL-UNNAMED
-      --add-opens=java.base/sun.nio.cs=ALL-UNNAMED
-      --add-opens=java.base/sun.security.action=ALL-UNNAMED
-      --add-opens=java.base/sun.util.calendar=ALL-UNNAMED
-    </extraJavaTestArgs>
   </properties>
 
   <modules>
@@ -300,6 +283,21 @@
           <artifactId>maven-enforcer-plugin</artifactId>
           <executions>
             <execution>
+              <id>enforce-java-version</id>
+              <goals>
+                <goal>enforce</goal>
+              </goals>
+              <configuration>
+                <rules>
+                  <requireJavaVersion>
+                    <version>[17,)</version>
+                    <message>Java 17 or higher is required to build parquet-java.</message>
+                  </requireJavaVersion>
+                </rules>
+                <fail>true</fail>
+              </configuration>
+            </execution>
+            <execution>
               <id>enforce-banned-dependencies</id>
               <goals>
                 <goal>enforce</goal>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -80,10 +80,10 @@
     <jackson-annotations.version>2.21</jackson-annotations.version>
     <japicmp.version>0.25.7</japicmp.version>
     <javax.annotation.version>1.3.2</javax.annotation.version>
-    <spotless.version>2.46.1</spotless.version>
+    <spotless.version>3.5.1</spotless.version>
     <shade.prefix>shaded.parquet</shade.prefix>
     <!-- Guarantees no newer classes/methods/constants are used by parquet. -->
-    <hadoop.version>3.3.0</hadoop.version>
+    <hadoop.version>3.4.3</hadoop.version>
```
nastra: we should probably get the hadoop version bump merged first. There's already a PR for that: https://github.com/apache/parquet-java/pull/3559

pom.xml
```diff
@@ -80,10 +80,10 @@
     <jackson-annotations.version>2.21</jackson-annotations.version>
     <japicmp.version>0.25.7</japicmp.version>
     <javax.annotation.version>1.3.2</javax.annotation.version>
-    <spotless.version>2.46.1</spotless.version>
+    <spotless.version>3.5.1</spotless.version>
```
nastra:
```suggestion
    <spotless.version>3.8.0</spotless.version>
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1361,12 +1362,42 @@ private void readVectored(List<ConsecutivePartList> allParts, ChunkListBuilder b
       totalSize += len;
     }
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
-    // Request a vectored read;
-    f.readVectored(ranges, options.getAllocator());
-    int k = 0;
-    for (ConsecutivePartList consecutivePart : allParts) {
-      ParquetFileRange currRange = ranges.get(k++);
-      consecutivePart.readFromVectoredRange(currRange, builder);
+    // Use a capturing allocator to track all buffers allocated by Hadoop during vectored reads.
```
nastra: https://github.com/apache/parquet-java/pull/3559 has an alternative approach for this.

pom.xml
```diff
@@ -605,7 +600,7 @@
             </excludeModules>
             <excludes>
               <exclude>${shade.prefix}</exclude>
-              <!-- JDK 11 adds interface methods/bridges on PrimitiveIterator.OfInt; ignore japicmp source incompatibility -->
+              <!-- JDK adds interface methods/bridges on PrimitiveIterator.OfInt; ignore japicmp source incompatibility -->
```
kevinjqliu: is this comment still correct without the `11`? or is it a java 11 specific comment?

kevinjqliu (commented):
Thanks for the PR! 

Looks like we can split out some some changes in other PRs, that would be helpful for reviews. I think we should keep this PR as light as possible and only deals changes required for java 17. Anything else we can either split into separate PR and merge first, or drop as a follow up

pom.xml
iemejia: Good call. I dropped the Hadoop bump from this PR — I verified the Java 17 build (compile + tests) passes fine on Hadoop 3.3.0 (vectored-I/O tests just skip, no buffer leaks), so it isn't needed here. Let's have the bump land via #3559.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
iemejia: Removed from this PR as well — it's only relevant once vectored I/O is actually exercised (Hadoop 3.4.x). Deferring to your approach in #3559.

pom.xml
iemejia: Done — bumped to `3.8.0` and ran `spotless:apply` (no formatting changes beyond the version bump). Thanks for the suggestion.

pom.xml
iemejia: Good question — I dug into it. It turned out not to be JDK-11-specific behavior at all: the exclude was a workaround for a false-positive source incompatibility that japicmp **0.23.1** reported on the synthetic/bridge methods of `PrimitiveIterator.OfInt` (added in #3388). japicmp was upgraded to **0.26.1** in #3592 (already on master), which handles those methods correctly. I verified that `japicmp:cmp` (with `breakBuildOnSourceIncompatibleModifications=true`) passes with the exclude removed, so I removed the exclude and the comment entirely rather than reword it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
Fokko: Thanks 👍  I don't think we should add code to Parquet to patch open issues on Hadoop. Just cleaned up the PR, I think it is good to go now.

Fokko (approved):
Based on the discussion on the devlist, let's move this forward for 1.19.0

iemejia:
YES !!! Thanks @Fokko !

PR #3584, opened 2026-05-24
chore: Deprecate `H1SeekableInputStream`

Fokko:
### Rationale for this change

Less is more. Hadoop 1 has been deprecated since a while, and the capability checks are not needed from Hadoop 3.3 onward.


dossett:
+1 non-binding, if needed.  The test failures seem unrelated?

PR #3586, opened 2026-05-25
GH-3585: Fix DataPageHeaderV2.num_nulls=-1 when column statistics are disabled

### LINKED ISSUE #3585: DataPageHeaderV2.num_nulls=-1 when column statistics are disabled

num_nulls is a required field in parquet v2 header (https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L729). However current implementation tracks this field only when statistics is enabled. When statistics is disabled, this required field will be set to -1, a default value in the tracker, causing incorrect num_nulls and will fail in some reader which do explicit sanity check like Rust arrow-rs.

This bug can be reproduced using:

```java
package com.keuin.testparquet;

import org.apache.avro.generic.GenericRecord;
import org.apache.parquet.avro.AvroParquetWriter;
import org.apache.parquet.column.page.DataPage;
import org.apache.parquet.column.page.DataPageV2;
import org.apache.parquet.column.page.PageReadStore;
import org.apache.parquet.column.page.PageReader;
import org.apache.parquet.hadoop.ParquetFileReader;
import org.apache.parquet.hadoop.ParquetWriter;
import org.apache.parquet.hadoop.util.HadoopInputFile;
import org.apache.parquet.io.InputFile;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.io.TempDir;

import java.nio.file.Path;

import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_1_0;
import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_2_0;
import static org.junit.jupiter.api.Assertions.*;

/**
 * Minimal reproduction of parquet-java bug:
 *
 * When using WriterVersion.PARQUET_2_0 (which produces DataPageV2 pages) combined with
 * .withStatisticsEnabled("column", false), the DataPageHeaderV2.num_nulls field is
 * written as -1 (0xFFFFFFFF).
 *
 * Root cause: ColumnWriterV2.writePage() calls Math.toIntExact(statistics.getNumNulls()),
 * and NoopStatistics.getNumNulls() unconditionally returns -1 (meaning "not set").
 * But DataPageHeaderV2.num_nulls is a required i32 field in the Thrift schema,
 * so -1 gets serialized as-is.
 *
 * Affected versions: parquet-java 1.15.2 (and likely all versions with NoopStatistics)
 * Bug location: ColumnWriterV2.java line ~107, NoopStatistics.java line ~97
 */
class ParquetV2NumNullsBugTest {

    /**
     * Minimal Avro schema: one non-null int field + one nullable string field.
     */
    private static final String SCHEMA_JSON = "{\n"
            + "  \"type\": \"record\",\n"
            + "  \"name\": \"TestRecord\",\n"
            + "  \"fields\": [\n"
            + "    {\"name\": \"id\", \"type\": \"int\"},\n"
            + "    {\"name\": \"value\", \"type\": [\"null\", \"string\"], \"default\": null}\n"
            + "  ]\n"
            + "}";

    /**
     * BUG REPRODUCTION: PARQUET_2_0 + withStatisticsEnabled(false) → num_nulls = -1
     *
     * This test demonstrates the bug. With statistics disabled on a nullable column,
     * the DataPageV2 header's num_nulls field is incorrectly set to -1.
     */
    @Test
    void bugRepro_parquet2_statisticsDisabled_numNullsIsMinusOne(@TempDir Path tempDir) throws Exception {
        Schema schema = new Schema.Parser().parse(SCHEMA_JSON);
        org.apache.hadoop.fs.Path parquetPath =
                new org.apache.hadoop.fs.Path(tempDir.resolve("bug.parquet").toUri());

        // Write: PARQUET_2_0 + statistics disabled on "value" column
        try (ParquetWriter<GenericRecord> writer = AvroParquetWriter.<GenericRecord>builder(parquetPath)
                .withSchema(schema)
                .withCompressionCodec(CompressionCodecName.UNCOMPRESSED)
                .withWriterVersion(PARQUET_2_0)
                .withStatisticsEnabled("value", false)  // ← triggers NoopStatistics
                .withPageSize(1024 * 1024) // large page to keep all records in one page
                .build()) {

            for (int i = 0; i < 10; i++) {
                GenericRecord record = new GenericData.Record(schema);
                record.put("id", i);
                // 4 nulls: i = 0, 3, 6, 9
                record.put("value", i % 3 == 0 ? null : "hello-" + i);
                writer.write(record);
            }
        }

        // Read back using low-level page API
        InputFile inputFile = HadoopInputFile.fromPath(parquetPath, conf);
        try (ParquetFileReader reader = ParquetFileReader.open(inputFile)) {
            MessageType fileSchema = reader.getFileMetaData().getSchema();

            // "value" is the second leaf column in the schema
            // For union ["null","string"], Avro maps to optional binary field
            var columns = fileSchema.getColumns();
            var valueColumn = columns.stream()
                    .filter(c -> c.getPath()[0].equals("value"))
                    .findFirst()
                    .orElseThrow(() -> new AssertionError("Column 'value' not found"));

            PageReadStore rowGroup = reader.readNextRowGroup();
            assertNotNull(rowGroup, "Should have at least one row group");

            PageReader pageReader = rowGroup.getPageReader(valueColumn);
            DataPage page = pageReader.readPage();
            assertNotNull(page, "Should have at least one data page");

            // Verify it's a V2 page (because we used PARQUET_2_0)
            assertInstanceOf(DataPageV2.class, page,
                    "PARQUET_2_0 writer should produce DataPageV2 pages");

            DataPageV2 pageV2 = (DataPageV2) page;
            int numNulls = pageV2.getNullCount();

            System.out.println("=== BUG REPRODUCTION ===");
            System.out.println("WriterVersion:        PARQUET_2_0");
            System.out.println("Statistics enabled:   false (for 'value' column)");
            System.out.println("Expected num_nulls:   4 (records with i=0,3,6,9)");
            System.out.println("Actual num_nulls:     " + numNulls);
            System.out.println("Bug present:          " + (numNulls == -1 ? "YES" : "NO"));
            System.out.println();

            // This assertion demonstrates the bug exists:
            // num_nulls should be 4, but is actually -1
            assertEquals(-1, numNulls,
                    "BUG: NoopStatistics.getNumNulls() returns -1, which gets written to "
                    + "DataPageHeaderV2.num_nulls. This value is the 'required i32 num_nulls' "
                    + "field in the Thrift page header.");
        }
    }

    /**
     * CONTROL: PARQUET_2_0 + statistics ENABLED → num_nulls is correct
     *
     * When statistics are enabled (default), num_nulls correctly reflects the actual null count.
     */
    @Test
    void control_parquet2_statisticsEnabled_numNullsIsCorrect(@TempDir Path tempDir) throws Exception {
        Schema schema = new Schema.Parser().parse(SCHEMA_JSON);
        org.apache.hadoop.fs.Path parquetPath =
                new org.apache.hadoop.fs.Path(tempDir.resolve("control.parquet").toUri());

        // Write: PARQUET_2_0 + statistics enabled (default)
        try (ParquetWriter<GenericRecord> writer = AvroParquetWriter.<GenericRecord>builder(parquetPath)
                .withSchema(schema)
                .withCompressionCodec(CompressionCodecName.UNCOMPRESSED)
                .withWriterVersion(PARQUET_2_0)
                // No withStatisticsEnabled(false) → default statistics are collected
                .withPageSize(1024 * 1024)
                .build()) {

            for (int i = 0; i < 10; i++) {
                GenericRecord record = new GenericData.Record(schema);
                record.put("id", i);
                record.put("value", i % 3 == 0 ? null : "hello-" + i);
                writer.write(record);
            }
        }

        // Read back
        InputFile inputFile = HadoopInputFile.fromPath(parquetPath, conf);
        try (ParquetFileReader reader = ParquetFileReader.open(inputFile)) {
            MessageType fileSchema = reader.getFileMetaData().getSchema();
            var valueColumn = fileSchema.getColumns().stream()
                    .filter(c -> c.getPath()[0].equals("value"))
                    .findFirst()
                    .orElseThrow();

            PageReadStore rowGroup = reader.readNextRowGroup();
            PageReader pageReader = rowGroup.getPageReader(valueColumn);
            DataPage page = pageReader.readPage();

            assertInstanceOf(DataPageV2.class, page);
            DataPageV2 pageV2 = (DataPageV2) page;
            int numNulls = pageV2.getNullCount();

            System.out.println("=== CONTROL (statistics enabled) ===");
            System.out.println("WriterVersion:        PARQUET_2_0");
            System.out.println("Statistics enabled:   true (default)");
            System.out.println("Expected num_nulls:   4");
            System.out.println("Actual num_nulls:     " + numNulls);
            System.out.println();

            assertEquals(4, numNulls,
                    "With statistics enabled, num_nulls should correctly be 4");
        }
    }

    /**
     * CONTROL: PARQUET_1_0 + statistics disabled → V1 page, no num_nulls in header
     *
     * DataPageV1 does not have a required num_nulls field, so this combination
     * does not trigger the bug. Statistics are simply omitted from the optional field.
     */
    @Test
    void control_parquet1_statisticsDisabled_noIssue(@TempDir Path tempDir) throws Exception {
        Schema schema = new Schema.Parser().parse(SCHEMA_JSON);
        org.apache.hadoop.fs.Path parquetPath =
                new org.apache.hadoop.fs.Path(tempDir.resolve("v1.parquet").toUri());

        // Write: PARQUET_1_0 + statistics disabled
        try (ParquetWriter<GenericRecord> writer = AvroParquetWriter.<GenericRecord>builder(parquetPath)
                .withSchema(schema)
                .withCompressionCodec(CompressionCodecName.UNCOMPRESSED)
                .withWriterVersion(PARQUET_1_0)
                .withStatisticsEnabled("value", false)
                .withPageSize(1024 * 1024)
                .build()) {

            for (int i = 0; i < 10; i++) {
                GenericRecord record = new GenericData.Record(schema);
                record.put("id", i);
                record.put("value", i % 3 == 0 ? null : "hello-" + i);
                writer.write(record);
            }
        }

        // Read back — V1 pages don't have a separate num_nulls field in the header
        InputFile inputFile = HadoopInputFile.fromPath(parquetPath, conf);
        try (ParquetFileReader reader = ParquetFileReader.open(inputFile)) {
            MessageType fileSchema = reader.getFileMetaData().getSchema();
            var valueColumn = fileSchema.getColumns().stream()
                    .filter(c -> c.getPath()[0].equals("value"))
                    .findFirst()
                    .orElseThrow();

            PageReadStore rowGroup = reader.readNextRowGroup();
            PageReader pageReader = rowGroup.getPageReader(valueColumn);
            DataPage page = pageReader.readPage();

            // V1 writer produces DataPageV1, not DataPageV2
            assertFalse(page instanceof DataPageV2,
                    "PARQUET_1_0 should NOT produce DataPageV2 pages");

            System.out.println("=== CONTROL (PARQUET_1_0) ===");
            System.out.println("WriterVersion:        PARQUET_1_0");
            System.out.println("Statistics enabled:   false");
            System.out.println("Page type:            " + page.getClass().getSimpleName());
            System.out.println("Result:               No issue — V1 pages don't have required num_nulls");
            System.out.println();
        }
    }
}
```

### Component(s)

_No response_

keuin:
### Rationale for this change

Fix the num_nulls field in DataPageHeaderV2, which will be unexpectedly set to -1 when statistics is disabled.

Add a counter tracker in ColumnValueCollector, which is always enabled, to track required field num_nulls.

### Are these changes tested?

Yes, tested. The test will fail before this fix, and will succeed with this fix.

No user-facing or API changes.

Closes #3585

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
@@ -51,6 +51,9 @@ abstract class ColumnWriterBase implements ColumnWriter {
   private ValuesWriter definitionLevelColumn;
   private ValuesWriter dataColumn;
   private int valueCount;
+  // track the required field DataPageHeaderV2.num_nulls
+  // https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift
+  protected int nullCount;
 
   private long rowsWrittenSoFar = 0;
   private int pageRowCount;
@@ -115,6 +118,7 @@ abstract class ColumnWriterBase implements ColumnWriter {
       definitionLevel(definitionLevel);
       collector.writeNull(repetitionLevel, definitionLevel);
       ++valueCount;
+      ++nullCount;
     } catch (Throwable e) {
       statusManager.abort();
       throw e;
@@ -392,6 +396,7 @@ abstract class ColumnWriterBase implements ColumnWriter {
       definitionLevelColumn.reset();
       dataColumn.reset();
       valueCount = 0;
+      nullCount = 0;
       collector.resetPageStatistics();
       pageRowCount = 0;
     } catch (Throwable t) {
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
@@ -58,9 +58,14 @@ import org.apache.hadoop.fs.Path;
 import org.apache.parquet.ParquetReadOptions;
 import org.apache.parquet.bytes.HeapByteBufferAllocator;
 import org.apache.parquet.bytes.TrackingByteBufferAllocator;
+import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.ParquetProperties;
 import org.apache.parquet.column.ParquetProperties.WriterVersion;
+import org.apache.parquet.column.page.DataPage;
+import org.apache.parquet.column.page.DataPageV2;
+import org.apache.parquet.column.page.PageReadStore;
+import org.apache.parquet.column.page.PageReader;
 import org.apache.parquet.column.values.bloomfilter.BloomFilter;
 import org.apache.parquet.crypto.AesCipher;
 import org.apache.parquet.crypto.ColumnEncryptionProperties;
@@ -857,5 +862,69 @@ public class TestParquetWriter {
     // After closing, check whether file exists or is empty
     FileSystem fs = file.getFileSystem(conf);
     assertTrue(!fs.exists(file) || fs.getFileStatus(file).getLen() == 0);
+  }
+
+  @Test
+  public void testV2PageNullCountWithStatisticsDisabled() throws Exception {
+    // Regression test: when using PARQUET_2_0 with statistics disabled on a nullable column,
+    // DataPageHeaderV2.num_nulls must still contain the correct null count (not -1).
+    MessageType schema = Types.buildMessage()
+        .required(INT32)
+        .named("id")
+        .optional(BINARY)
+        .as(stringType())
+        .named("value")
+        .named("test_schema");
+
+    File file = temp.newFile();
+    file.delete();
+    Path path = new Path(file.getAbsolutePath());
+
+    int totalRecords = 10;
+    int expectedNulls = 4; // records where i % 3 == 0: i=0,3,6,9
+
+    // Write with PARQUET_2_0 and statistics disabled on the nullable "value" column
+    try (ParquetWriter<Group> writer = ExampleParquetWriter.builder(path)
+        .withType(schema)
+        .withWriterVersion(PARQUET_2_0)
+        .withStatisticsEnabled("value", false)
+        .withPageSize(1024 * 1024) // large page to keep all records in one page
+        .build()) {
+      SimpleGroupFactory factory = new SimpleGroupFactory(schema);
+      for (int i = 0; i < totalRecords; i++) {
+        Group group = factory.newGroup().append("id", i);
+        if (i % 3 != 0) {
+          group.append("value", "hello-" + i);
+        }
+        writer.write(group);
+      }
+    }
+
+    // Read back the page-level metadata and verify num_nulls
+    try (ParquetFileReader reader = ParquetFileReader.open(HadoopInputFile.fromPath(path, new Configuration()))) {
+      MessageType fileSchema = reader.getFooter().getFileMetaData().getSchema();
+
+      // Find the "value" column descriptor
+      ColumnDescriptor valueColumn = fileSchema.getColumns().stream()
+          .filter(c -> c.getPath()[0].equals("value"))
+          .findFirst()
+          .orElseThrow(() -> new AssertionError("Column 'value' not found"));
+
+      PageReadStore rowGroup = reader.readNextRowGroup();
+      PageReader pageReader = rowGroup.getPageReader(valueColumn);
+      DataPage page = pageReader.readPage();
+
+      // Verify it's a V2 page (because we used PARQUET_2_0)
+      assertTrue(
+          "PARQUET_2_0 writer should produce DataPageV2 pages, got: "
+              + page.getClass().getSimpleName(),
+          page instanceof DataPageV2);
+
+      DataPageV2 pageV2 = (DataPageV2) page;
+      assertEquals(
+          "DataPageV2.num_nulls should be the actual null count even when statistics are disabled",
+          expectedNulls,
+          pageV2.getNullCount());
+    }
   }
 }
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV2.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV2.java
@@ -100,7 +100,7 @@ final class ColumnWriterV2 extends ColumnWriterBase {
     Encoding encoding = values.getEncoding();
     pageWriter.writePageV2(
         rowCount,
-        Math.toIntExact(statistics.getNumNulls()),
+        nullCount,
         valueCount,
         repetitionLevels.getBytes(),
         definitionLevels.getBytes(),


parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
```diff
@@ -54,6 +58,7 @@ class ColumnValueCollector {
   }
 
   void resetPageStatistics() {
+    this.nullCount = 0;
     this.statistics = statisticsEnabled
         ? Statistics.createStats(path.getPrimitiveType())
         : Statistics.noopStats(path.getPrimitiveType());
```
wgtmac, line 59: Is it possible to change `Statistics.noopStats` to make it possible to ignore min/max stats but still collects null count?

keuin: It's possible, but putting it outside `Statistics` is a design choice. By reading code I think the `Statistics` object is a 1:1 mapping to parquet header's `statistics` object, which is optional and may be disabled by user.

Putting counter of `num_nulls` outside `statistics` object makes the code clearer, because the code maps clearly to parquet header structure.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
```diff
@@ -45,6 +45,10 @@ class ColumnValueCollector {
   private SizeStatistics.Builder sizeStatisticsBuilder;
   private GeospatialStatistics.Builder geospatialStatisticsBuilder;
 
+  // track the required `num_nulls` field in DataPageHeaderV2
+  // https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift
+  private int nullCount;
```
wgtmac: Could we keep this counter in ColumnWriterBase instead? nullCount is page-level metadata, like valueCount and pageRowCount, while ColumnValueCollector is otherwise about optional statistics and bloom filters. Keeping it with the other page counters would make the ownership clearer.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterBase.java
```diff
@@ -403,6 +404,7 @@ void writePage() {
   abstract void writePage(
       int rowCount,
       int valueCount,
+      int nullCount,
```
wgtmac: This adds a V2-only value to the shared writePage signature, so ColumnWriterV1 has to accept an unused parameter. Consider keeping nullCount on the V2 path only, either through a V2-specific helper or a protected getter from the base class.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
```diff
@@ -858,4 +863,68 @@ public void testNoFlushAfterException() throws Exception {
...
+    MessageType schema = Types.buildMessage()
+        .required(INT32)
+        .named("id")
+        .optional(BINARY)
+        .as(stringType())
+        .named("value")
+        .named("test_schema");
+
+    File file = temp.newFile();
+    temp.delete();
```
wgtmac: This should delete only the placeholder file, not the TemporaryFolder root. file.delete() is enough to free the output path and avoids changing the test fixture lifecycle.

parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnValueCollector.java
keuin: Yes, placing it inside ColumnWriterBase is the correct way. Fixed.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetWriter.java
keuin: Fixed.

wgtmac (approved):

wgtmac:
Thank you, @keuin!

Issue #3587, opened 2026-05-26
Setting parquet.thrift.string.size.limit = -1 incorrectly throws


divjotarora:
### Describe the bug, including details regarding any error messages, version, and platform.

https://github.com/apache/parquet-java/pull/3359 added a `parquet.thrift.string.size.limit` configuration to allow specifying a custom max Thrift message size. If unspecified or explicitly set to -1, the default (100 MB) is used. The config validation code introduced a minor bug [here](https://github.com/apache/parquet-java/blob/8931c1c55f1fba399dd75139f75bcde0b84137c0/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java#L420): if the value is explicitly set to -1, `maxMessageSize` is correctly set to the default, but the `configuredMaxMessageSize <= 0` branch will still throw an exception.

PR #3588, opened 2026-05-26
GH-3587: Fix `parquet.thrift.string.size.limit` validation

divjotarora:
### Rationale for this change
This PR fixes a small bug in `parquet.thrift.string.size.limit` validation [here](https://github.com/apache/parquet-java/blob/8931c1c55f1fba399dd75139f75bcde0b84137c0/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java#L420). If this config is explicitly set to -1, the code indicates that the default max message size should be used, but also throws an exception.

If `parquet.thrift.string.size.limit == -1`, use the default value and do not fail validation.

### Are these changes tested?
Yes, there are new regression tests added to `TestParquetFileReaderMaxMessageSize`.

No.

Closes #GH-3587

### DIFF
--- a/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
+++ b/parquet-format-structures/src/main/java/org/apache/parquet/format/Util.java
@@ -417,8 +417,8 @@ public class Util {
       // Set to default 100 MB
       maxMessageSize = DEFAULT_MAX_MESSAGE_SIZE;
     }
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Great catch @divjotarora

wgtmac (approved):

Fokko:
Thanks for fixing this @divjotarora and thanks for the review @wgtmac 🙌

PR #3589, opened 2026-05-26
GH-3572: Bump thrift to 0.23

### LINKED ISSUE #3572: Update thrift to 0.23.0 to eliminate warnings about CVE-2026-43870

There'a a new thrift release with a CVE associated with the RPC code. 

Parquet is *not* exposed to this, but thrift gets onto the classpath and all CVE scanners will be complaining about the version being vulnerable to CVE-2026-43870.

### Component(s)

Build

steveloughran:
### Rationale for this change

There's a new Thrift release out.

Changes include a fix for the CVE https://github.com/advisories/GHSA-526f-jxpj-jmg2
This is server side and only affect thrift javascript code. While parquet is unaffected, security scanner tools aren't necessarily going to be that nuanced.

* updated build files/scripts with thrift version declarations
* updated references in README.md
* Added instructions in README as to where to find the gpg/sha signatures and a link to the thrift team KEYS file.

### Are these changes tested?

* Expecting PR CI to do the tests.
* It compiles!
* I manually ran the modified wget command in the README to verify the path to the tarball is valid.

No

Closes #3572

### DIFF
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
@@ -26,14 +26,14 @@ import org.apache.thrift.transport.TTransportException;
  * A stub transport which implements the minimum amount needed for range/depth
  * validation within the thrift library to succeed.
  */
-public final class StubTTransport extends TTransport {
+final class StubTTransport extends TTransport {
 
-  public static final StubTTransport INSTANCE = new StubTTransport();
+  static final StubTTransport INSTANCE = new StubTTransport();
 
   /**
    * There's no limits on recursion depth.
    */
-  private final TConfiguration conf = new TConfiguration(
+  private static final TConfiguration CONFIGURATION = new TConfiguration(
       TConfiguration.DEFAULT_MAX_FRAME_SIZE, TConfiguration.DEFAULT_MAX_MESSAGE_SIZE, Integer.MAX_VALUE);
 
   private StubTTransport() {}
@@ -66,7 +66,7 @@ public final class StubTTransport extends TTransport {
 
   @Override
   public TConfiguration getConfiguration() {
-    return conf;
+    return CONFIGURATION;
   }
 
   @Override
--- a/parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
+++ b/parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
@@ -34,16 +34,12 @@ public final class StubTTransport extends TTransport {
    * There's no limits on recursion depth.
    */
   private final TConfiguration conf = new TConfiguration(
-      TConfiguration.DEFAULT_MAX_FRAME_SIZE,
-      TConfiguration.DEFAULT_MAX_MESSAGE_SIZE,
-      Integer.MAX_VALUE);
+      TConfiguration.DEFAULT_MAX_FRAME_SIZE, TConfiguration.DEFAULT_MAX_MESSAGE_SIZE, Integer.MAX_VALUE);
 
-  private StubTTransport() {
-  }
+  private StubTTransport() {}
 
   @Override
-  public void checkReadBytesAvailable(long l) {
-  }
+  public void checkReadBytesAvailable(long l) {}
 
   @Override
   public boolean isOpen() {
@@ -56,8 +52,7 @@ public final class StubTTransport extends TTransport {
   }
 
   @Override
-  public void close() {
-  }
+  public void close() {}
 
   @Override
   public int read(byte[] bytes, int i, int i1) throws TTransportException {
@@ -75,6 +70,5 @@ public final class StubTTransport extends TTransport {
   }
 
   @Override
-  public void updateKnownMessageSize(long l) {
-  }
+  public void updateKnownMessageSize(long l) {}
 }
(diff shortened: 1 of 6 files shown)


steveloughran:
new thrift is triggering NPEs in tests. That is *bad*
```
Error:  org.apache.parquet.hadoop.thrift.TestParquetToThriftReadWriteAndProjection.testPullInRequiredLists -- Time elapsed: 0.027 s <<< ERROR!
org.apache.parquet.io.ParquetDecodingException: Can not read value at 1 in block 0 in file file:/home/runner/work/parquet-java/parquet-java/parquet-thrift/target/test/TestParquetToThriftReadWriteAndProjection/file.parquet
	at org.apache.parquet.hadoop.InternalParquetRecordReader.nextKeyValue(InternalParquetRecordReader.java:280)
	at org.apache.parquet.hadoop.ParquetRecordReader.nextKeyValue(ParquetRecordReader.java:207)
	at org.apache.parquet.hadoop.thrift.TestParquetToThriftReadWriteAndProjection.shouldDoProjection(TestParquetToThriftReadWriteAndProjection.java:375)
	at org.apache.parquet.hadoop.thrift.TestParquetToThriftReadWriteAndProjection.shouldDoProjectionWithThriftColumnFilter(TestParquetToThriftReadWriteAndProjection.java:337)
	at org.apache.parquet.hadoop.thrift.TestParquetToThriftReadWriteAndProjection.testPullInRequiredLists(TestParquetToThriftReadWriteAndProjection.java:301)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
Caused by: java.lang.NullPointerException
	at org.apache.thrift.protocol.TProtocol.incrementRecursionDepth(TProtocol.java:59)
	at org.apache.parquet.thrift.test.RequiredListFixture$RequiredListFixtureStandardScheme.read(RequiredListFixture.java:414)
	at org.apache.parquet.thrift.test.RequiredListFixture$RequiredListFixtureStandardScheme.read(RequiredListFixture.java:410)
	at org.apache.parquet.thrift.test.RequiredListFixture.read(RequiredListFixture.java:345)
	at org.apache.parquet.thrift.TBaseRecordConverter$1.readOneRecord(TBaseRecordConverter.java:63)
	at org.apache.parquet.thrift.TBaseRecordConverter$1.readOneRecord(TBaseRecordConverter.java:58)
	at org.apache.parquet.thrift.ThriftRecordConverter.getCurrentRecord(ThriftRecordConverter.java:945)
	at org.apache.parquet.io.RecordReaderImplementation.read(RecordReaderImplementation.java:439)
```

steveloughran:
NPE cause is [THRIFT-5916](https://issues.apache.org/jira/browse/THRIFT-5916) _Bring Java recursion limit support at parity with C++_;
https://github.com/apache/thrift/pull/3287

commit #66dae3f is the "least change to existing code" patch, not necessarily the most elegant. It provides enough of a stub class to stop thrift blowing up.

The alternative is for ParquetProtocol and BufferedProtocolReadToWrite.NullProtocol to
* Implement incrementRecursionDepth()/decrementRecursionDepth() as no-ops
* Override all the implementations of checkReadBytesAvailable() to be no-ops.

Good: less stubbing of thrift internals
Bad:
- duplicate implementation of no-op methods. They're only no-op methods though...
- the transport is still null, so risk of NPEs surfacing if new checks are added.

Either way there's a risk of regressions if they add more checks in future releases.

There's also the open issue "should there be a depth limit?"; we're adding one when hardening variants after all. But there we can be fairly confident there are no large datasets using the type.

If thrift cpp enforces a depth of 64, then so will parquet-cpp, won't it? In which case modifying this PR to use the default depth of 64 would be consistent.

parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.apache.thrift.TConfiguration;
+import org.apache.thrift.transport.TTransport;
+import org.apache.thrift.transport.TTransportException;
+
+/**
+ * A stub transport which implements the minimum amount needed for range/depth
+ * validation within the thrift library to succeed.
+ */
+public final class StubTTransport extends TTransport {
```
Fokko: Should we make this package private?

parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public static final StubTTransport INSTANCE = new StubTTransport();
+
+  /**
+   * There's no limits on recursion depth.
+   */
+  private final TConfiguration conf = new TConfiguration(
```
Fokko: We can make this one static

Fokko (approved):
Thanks @steveloughran for fixing this 🙌

parquet-thrift/src/main/java/org/apache/parquet/thrift/StubTTransport.java
steveloughran: if it's possible, yes.

steveloughran:
@Fokko I think this one is ready to go in;

Fokko:
@steveloughran thanks for the ping, and I agree :) Thanks for fixing this!

steveloughran:
@Fokko 
thanks, not actually needed, but it'll stop the scanning tools worrying.

PR #3591, opened 2026-05-28
MINOR: Remove Java 8 direct buffer cleaner path

dossett:
Since Java 8 is no longer supported, we don't need to support its specific reflection requirements here.

(Codex was used to create this PR)


Fokko (approved):
I love it! LGTM

Fokko:
Thanks @dossett

PR #3597, opened 2026-06-05
GH-3596: Add `RowRanges.Builder` for incremental construction from selected row indices

### LINKED ISSUE #3596: Add a `RowRanges.Builder` for incremental construction from selected row indices

`org.apache.parquet.internal.filter2.columnindex.RowRanges` is currently only constructible internally — instances are produced by column-index filtering (`ColumnIndexFilter.calculateRowRanges(...)`) or by the package-private factory/union helpers. All of its constructors are `private`, so there is no supported way for code outside the package to build a `RowRanges` from an arbitrary set of selected rows.

This is limiting for external readers that determine which rows to read from some other source than column-index filtering. A concrete motivating case is a materialization path that receives a stream of selected row indices from a downstream operator (for example, the row positions surviving a filter or join) and needs to turn that stream into a `RowRanges` to drive reads — without knowing page boundaries ahead of time.

#### Proposed change

Add a small `RowRanges.Builder` that lets callers append selected row indices incrementally and coalesces consecutive indices into `Range` entries:

```java
RowRanges.Builder builder = RowRanges.builder();
for (long row : selectedRowsInOrder) {
  builder.addSelected(row);
}
RowRanges ranges = builder.build();
```

Semantics:

- `addSelected(long)` must be called in **strictly increasing** order. Consecutive indices are merged into a single `Range`; a gap closes the current run and starts a new one.
- Calling `addSelected` with a value `<=` the previous value throws `IllegalArgumentException` (rejects out-of-order and duplicate indices).
- `build()` returns `RowRanges.EMPTY` when no rows were selected.

#### Why a builder (vs. exposing a constructor)

A builder keeps `RowRanges` immutable and its internal `Range` list encapsulated, while still letting callers feed rows one at a time. The coalescing logic lives in one place rather than being re-implemented by each external caller, and the strictly-increasing contract is enforced at the point of construction.

#### Scope

- Additive, `Core` only. No change to existing `RowRanges` behavior or to the column-index filtering path.
- No user-facing API removal or behavioral change to existing methods.

This is the first of two related enhancements opening up `RowRanges`/reader APIs needed by the materialization feature described above; a follow-up (#3597) will expose per-row-range reader APIs on `ParquetFileReader`.

### Component(s)
**Core**

peter-toth:
### Rationale for this change

This PR is based on @mbutrovich's previous work.

Opening up APIs needed by a later materialization feature in Spark. External readers need to assemble a `RowRanges` incrementally from a stream of selected row indices (e.g. produced by a downstream filter or join) without having to know page boundaries ahead of time.

Since `RowRanges` is the type external readers must work with, this PR also promotes it out of the `org.apache.parquet.internal.*` package into a supported location.

- Adds a `RowRanges.Builder` for incremental construction. It accepts a strictly-increasing sequence of selected rows via `addSelectedRow(long rowIndex)` — or ranges via `addSelectedRange(long from, long to)`, with `addSelectedRow` delegating to it — and coalesces consecutive/adjacent entries into `Range`s. Out-of-order, overlapping, duplicate, negative, or `from > to` calls throw `IllegalArgumentException`. `build()` returns an independent snapshot, so the builder can be reused afterwards. `Builder` is `final` with a private constructor.
- Moves `RowRanges` (with `Range` and `Builder`) from `org.apache.parquet.internal.filter2.columnindex` to `org.apache.parquet.filter2.columnindex`.
- Preserves backward compatibility for the one released public API that referenced the old type: a deprecated `org.apache.parquet.internal.filter2.columnindex.RowRanges` subclass is kept, and `ParquetFileReader.readFilteredRowGroup(int, RowRanges)` gains a deprecated overload (taking the old internal type) that delegates to the new one. Both are marked for removal in 2.0.
- Reimplements `RowRanges.create(...)` on top of the new `Builder` and keeps it on the deprecated `internal.RowRanges` (preserving the released `internal.RowRanges.create(...)` signature), so the public `RowRanges` no longer references the internal `OffsetIndex`; `ColumnIndexFilter` now builds via the public builder.
- Other `internal`-only surfaces (`RowRanges` statics, `Range`, `ColumnIndexFilter` return types) are relocated without a bridge, as the `internal` package carries no compatibility guarantee.

### Are these changes tested?

Yes. `TestRowRanges` covers single/multiple/coalesced ranges, the empty builder case, the out-of-order/duplicate/negative rejection paths, `addSelectedRange` coalescing and overlap/`from > to` rejection, the `Long.MAX_VALUE` edge case, and that `build()` returns a snapshot independent of the builder. Existing `TestColumnIndexFilter` and `TestParquetFileReaderRowRanges` continue to pass against the relocated type.

Yes. `RowRanges` (and `Range`/`Builder`) is now available at the supported package `org.apache.parquet.filter2.columnindex`. The old `org.apache.parquet.internal.filter2.columnindex.RowRanges` and `ParquetFileReader.readFilteredRowGroup(int, <internal> RowRanges)` remain as deprecated shims and will be removed in 2.0.

Closes #3596


peter-toth:
@wgtmac, could you please take a look at this PR?

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
```diff
@@ -316,4 +316,72 @@ public List<Range> getRanges() {
...
+            "addSelected requires strictly increasing row indices; got " + blockRow + " after " + runEnd);
+      }
+      return this;
+    }
+
+    /**
+     * @return the constructed {@link RowRanges}, or {@link RowRanges#EMPTY} when no rows were
+     *     selected.
+     */
+    public RowRanges build() {
```
wgtmac, line 380: `build()` should return a snapshot or become terminal. As written, the returned `RowRanges` shares the builder’s mutable `ranges` list, so later calls on the same builder can mutate a previously built result and even break the sorted-ranges invariant.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
```diff
@@ -316,4 +316,72 @@ public List<Range> getRanges() {
...
+
+    /**
+     * Marks {@code blockRow} as selected. Must be called in strictly increasing order; calling
+     * with a value less than or equal to the previous call's value throws
+     * {@link IllegalArgumentException}.
+     *
+     * @param blockRow the row index to mark selected (must be {@code >} the last value passed)
+     * @return this builder for chaining
+     */
+    public Builder addSelected(long blockRow) {
```
wgtmac: `addSelected` should reject negative row indexes explicitly. `-1` also collides with the builder’s sentinel state, so invalid input can be silently swallowed or corrupt the active run.

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
```diff
@@ -316,4 +316,72 @@ public List<Range> getRanges() {
   public String toString() {
     return ranges.toString();
   }
+
+  /**
```
wgtmac, line 295: This needs a clearer API commitment. The motivation is external readers/Spark, but the type lives under `internal` and the PR says there are no user-facing changes. If this is meant to be supported externally, the package/API contract should say so; otherwise downstream users may depend on an unsupported API.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
```diff
@@ -316,4 +316,72 @@ public List<Range> getRanges() {
...
+   * <p>Usage:
+   * <pre>{@code
+   * RowRanges.Builder builder = RowRanges.builder();
+   * for (long row : selectedRowsInOrder) {
+   *   builder.addSelected(row);
+   * }
+   * RowRanges ranges = builder.build();
+   * }</pre>
+   */
+  public static class Builder {
```
wgtmac: Consider making `Builder` `final` and adding an explicit constructor. Otherwise subclassing and `new RowRanges.Builder()` become part of the public API surface by accident.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
wgtmac: The naming could be clearer before this API ships. `addSelected(long blockRow)` is ambiguous; something like `addSelectedRow(long rowIndex)` would better communicate that the value is a 0-based row index within the current row group.

wgtmac (commented):
Codex review, focused only on API design.

wgtmac:
Do you have any advice on this new API? @gszadovszky

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
gszadovszky: Yeah, at the time, it was yet another approach to tighten the amount of publicly available API in parquet-java. It quickly turned out that 3rd parties want to use it. For the next major release it is a good candidate to be moved to a more public package. Maybe deprecate it here, and also create at another location?

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
peter-toth: Fixed in https://github.com/apache/parquet-java/pull/3597/commits/17cce6912afcc3836d96f703fae0462b67b5dff2.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
peter-toth: Added a check in https://github.com/apache/parquet-java/pull/3597/commits/17cce6912afcc3836d96f703fae0462b67b5dff2.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
peter-toth: Good idea, fixed in https://github.com/apache/parquet-java/pull/3597/commits/17cce6912afcc3836d96f703fae0462b67b5dff2.

parquet-column/src/main/java/org/apache/parquet/internal/filter2/columnindex/RowRanges.java
peter-toth: Renamed in https://github.com/apache/parquet-java/pull/3597/commits/17cce6912afcc3836d96f703fae0462b67b5dff2.

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
peter-toth: @wgtmac, @gszadovszky, I wonder which package could be the right place for `RowRanges` and the new builder? Shall I try to move/deprecate them in this PR or separate one?

gszadovszky: A proper target package would be the same as now just skip the `internal` part. I am fine moving/deprecating in this PR. Theoretically, we can simply move it without deprecation since it is "internal" currently, so we have no guarantees for backward compatibility. WDYT, @wgtmac?

wgtmac: Agreed.

peter-toth: Ok, moved in https://github.com/apache/parquet-java/pull/3597/commits/8d8e2dd71d51a639cd61f27b9e340b3937bc3f51.

One thing to flag: the relocation changes a public signature `ParquetFileReader.readFilteredRowGroup(int, RowRanges)` now takes the new public `RowRanges` instead of the internal one. japicmp reports this as `METHOD_REMOVED` (and the inherited form on `CompressionConverter$TransParquetFileReader`), so I added excludes for:
- `org.apache.parquet.internal.filter2.columnindex.RowRanges` (+ `$Range`)
- `org.apache.parquet.internal.filter2.columnindex.ColumnIndexFilter` (return-type changes)
- `org.apache.parquet.hadoop.ParquetFileReader#readFilteredRowGroup(int, …internal…RowRanges)` (+ the `TransParquetFileReader` inherited form)

gszadovszky: Oh, I haven't noticed that, sorry. Then, we probably need to get back to the deprecated/move pattern. :disappointed: 
Let's move to the new place, extend the class as required at the old place and deprecate. Also deprecate the public API that references the old place and create new methods for the new one. Mark 2.0 as the removal version. WDYT?

peter-toth: I looked at the released (1.17.0) surface: the only genuinely-public, cross-package method exposing `RowRanges` is `ParquetFileReader.readFilteredRowGroup(int, RowRanges)`. `ParquetFileReader.getRowRanges(int)`, that returns `RowRanges` was only made public on `master` (1.18.0).
Everything else is under `org.apache.parquet.internal.*`. 

Is it ok if we preserve compatibility for just `readFilteredRowGroup(int, RowRanges)` and treat the `internal.*` relocations as internal changes with no compat guarantee?

wgtmac: > Is it ok if we preserve compatibility for just readFilteredRowGroup(int, RowRanges) and treat the internal.* relocations as internal changes with no compat guarantee?

It really depends on whether this relocation will break any existing users. As you've said that `readFilteredRowGroup` has been made public since 1.17.0 so we have to add a shim layer for the internal RowRanges anyway?

peter-toth: Yes we do. Move and deprecation is done in https://github.com/apache/parquet-java/pull/3597/commits/fa56ff25b15d4c106dea50210fffaac035db757b.

Please note that `org.apache.parquet.internal.*` excludes still remained.

gszadovszky: Thanks, @peter-toth. LGTM.

peter-toth:
@wgtmac, @gszadovszky, please let me know how to proceed with this PR.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1268,6 +1268,22 @@ public ColumnChunkPageReadStore readFilteredRowGroup(int blockIndex, RowRanges r
...
+   * @param rowRanges  the row ranges to be read from the requested block
+   * @return the PageReadStore which can provide PageReaders for each column or null if there are no rows in this block
+   * @throws IOException              if an error occurs while reading
+   * @throws IllegalArgumentException if the {@code blockIndex} is invalid or the {@code rowRanges} is null
+   * @deprecated use {@link #readFilteredRowGroup(int, RowRanges)} with
+   *     {@link org.apache.parquet.filter2.columnindex.RowRanges} instead. This overload is retained
+   *     for backward compatibility and will be removed in 2.0.
+   */
+  @Deprecated
+  public ColumnChunkPageReadStore readFilteredRowGroup(
```
wgtmac, line 1282: Shouldn't we keep and deprecate original `readFilteredRowGroup` with internal RowRanges API and add a new one with the public RowRanges API? Then we don't need to change the pom file?

peter-toth:
Thanks @wgtmac. I'm a bit confused, so let me lay out how things stand now:

**How old and new relate.** The real `RowRanges` implementation now lives at the new public location `org.apache.parquet.filter2.columnindex.RowRanges`. The old `org.apache.parquet.internal.filter2.columnindex.RowRanges` is kept as a thin **deprecated subclass** of it (`Old extends New`), purely so the released signatures keep linking.

**The method you commented on.** That overload *is* the deprecated one — `readFilteredRowGroup(int, internal.RowRanges)` is retained (and `@Deprecated`) as a shim that delegates to the new `readFilteredRowGroup(int, public.RowRanges)`. So we already keep the original method; it's not going away until 2.0.

**Why the excludes exist.** They're not caused by `readFilteredRowGroup` — that signature is preserved, so japicmp is happy with it. The excludes are for the `internal` package itself: making `Old` an empty subclass moved its members (statics, `Range`) into `New`, and `ColumnIndexFilter.calculateRowRanges` now returns the public `RowRanges`. Since `internal.*` carries no backward-compat guarantee, those are intentional and that's what the three excludes cover.

**On dropping the excludes.** Reversing the inheritance (`New extends Old`) would indeed let us avoid the excludes now — but I think it just moves the cost to 2.0: when we delete the old parent, `New` loses its superclass and all the inherited members, so we'd have to pull them up into `New` and add fresh excludes *for the class we actually keep*. The current direction makes `Old` a deletable leaf, so in 2.0 these three excludes just get removed along with it — no new ones.

The other option is to fully separate `New` from `Old` (no shared type), but then `ColumnIndexFilter` and the range-algebra helpers would have to be duplicated to produce the new type (or we add an `Old → New` conversion layer), which felt worse than three internal-package excludes.

Do you have a concrete approach in mind? I'm happy to explore it.

peter-toth:
@wgtmac , any suggestion on the above? Do you think the current implementation is fine or shall try to fully separate old and new `RowRanges`?

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
```diff
@@ -0,0 +1,400 @@
...
+     * @return this builder for chaining
+     */
+    public Builder addSelectedRow(long rowIndex) {
+      if (rowIndex < 0) {
+        throw new IllegalArgumentException("addSelectedRow requires a non-negative row index; got " + rowIndex);
+      }
+      if (runStart < 0) {
+        runStart = rowIndex;
+        runEnd = rowIndex;
+      } else if (rowIndex == runEnd + 1) {
```
wgtmac: nit: we may have problem when `rowIndex` is Long.MAX_VALUE

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
```diff
@@ -0,0 +1,400 @@
...
+   * <p>
+   * The union of the ranges are calculated so the result ranges always contain the disjunct ranges. See union for
+   * details.
+   *
+   * @param rowCount    row count
+   * @param pageIndexes pageIndexes
+   * @param offsetIndex offsetIndex
+   * @return a mutable RowRanges
+   */
+  public static RowRanges create(long rowCount, PrimitiveIterator.OfInt pageIndexes, OffsetIndex offsetIndex) {
```
wgtmac: This function needs the above `import org.apache.parquet.internal.column.columnindex.OffsetIndex;` and exposes internal API so it looks odd. That said, `OffsetIndex` and `ColumnIndexFilter` are already internal classes that are public to users. I'm not sure if we need to do the same thing for these classes to remove their `internal` path, or just move `create` to the internal RowRanges class without introducing confusion to this class.

@gszadovszky Do you have a preference?

wgtmac (commented):
Sorry for the delay. Now it overall looks good. I just have two more comments. Let me know what you think. Thanks!

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
peter-toth: Good catch. Fixed in https://github.com/apache/parquet-java/pull/3597/commits/a30ea848eaf9151ef24040618934a167b5b498db.

parquet-column/src/main/java/org/apache/parquet/filter2/columnindex/RowRanges.java
peter-toth: Thanks @wgtmac. Laying out the options as I see them:

**1. Make `OffsetIndex` public (drop its `internal`).** Likely a bigger change with its own compat issues — it repeats the same signature juggling we're doing here for `RowRanges`, but on a type with a much wider blast radius (more implementors and usages). I'd rather not pull that into this PR; better as its own change for 2.0.

**2. Move `create()` to `internal.RowRanges` or into its only caller `ColumnIndexFilter`, as-is.** This doesn't work cleanly today: `create()` builds ranges via the private `add()` and the package-private `Range` constructor, so either target (both in a different package) would force `add()` and `Range`'s constructor to become `protected`/`public`. That'd be odd now that we have the public `RowRanges.Builder` as the intended construction API — we'd be exposing a second, lower-level way to build ranges.

**3. Reimplement `create()` on top of the `Builder`.** I think this is the cleanest. It needs one new public builder method, `addSelectedRange(from, to)`, and `addSelectedRow` just delegates to it. `create()` then builds purely through the public API, so `add()`/`Range` stay private. One behavior difference: today's `add()` silently unions *overlapping* ranges, whereas the builder only coalesces *adjacent* ones and rejects overlaps. For the filter path pages never overlap, so the result is identical — and rejecting overlaps actually looks like an improvement to me, since an overlap there would mean a real bug.

I went with option 3 in commit https://github.com/apache/parquet-java/pull/3597/commits/68c0556d75b9a23a3160c38c6f912b68fd2ffe73: `create()` is reimplemented on the builder and kept on `internal.RowRanges` (so the released `internal.RowRanges.create(...)` signature stays), and the public `RowRanges` no longer references the internal `OffsetIndex`. Let me know if that works for you.

wgtmac (approved):
Thanks @peter-toth, now it looks good to me. It would be good to wait for @gszadovszky to take a look in case I missed anything.

gszadovszky (approved):
It looks good to me, too.
Thank you @peter-toth and @wgtmac!

peter-toth:
Thank you @wgtmac and @gszadovszky for the review!

PR #3599, opened 2026-06-05
GH-3598: Expose `getRowRanges(int)`

### LINKED ISSUE #3598: Expose row-range planning APIs on `ParquetFileReader` (`getRowRanges`, `getCompressedBytesForRowRanges`)

`ParquetFileReader` already computes, per row group, the `RowRanges` that may pass the configured filter — but only as a `private` step on the internal read path (`getRowRanges(int)` feeding `readFilteredRowGroup`). There is also no supported way to ask how many compressed bytes a given set of row ranges corresponds to without actually reading the column data.

This makes it hard for an external reader to plan I/O up front. A concrete motivating case is a materialization path (e.g. a Spark-side scanner) that wants to (a) obtain the column-index-derived row ranges that may pass the filter for a row group, and (b) get a metadata-only estimate of the on-disk compressed bytes those ranges correspond to for the currently requested columns — so it can size buffers and schedule reads without first touching column data.

#### Proposed change

Two additive APIs on `ParquetFileReader`, both metadata-only (they consult the column/offset indexes from the footer; no column data is read):

- **`public RowRanges getRowRanges(int blockIndex)`** — promote the existing private method to public. It returns the row ranges within the row group that may pass the configured filter. When no filter is configured it short-circuits to a `RowRanges` covering all rows in the row group (the previous `private` version asserted a filter was present, since it was only ever reached on the filtering path).

- **`public long getCompressedBytesForRowRanges(int blockIndex, RowRanges rowRanges)`** — sum of compressed page sizes (`OffsetIndex.getCompressedPageSize`, which includes the page header) across the reader's currently requested columns, for pages whose row range intersects `rowRanges`. Returns `0` for empty ranges or when no columns are requested. Throws `MissingOffsetIndexException` if a requested column lacks an offset index.

  Note: dictionary pages are not represented in `OffsetIndex`, so they are **excluded** from the sum — the result is a lower bound on actual on-disk bytes for those columns/rows by exactly the dictionary-page contribution.

#### Scope

- Additive, `Core` only. The only behavioral change is that `getRowRanges` now handles the no-filter case (returning all rows) instead of asserting; all existing callers already guard with a filter check, so they are unaffected.
- No user-facing API removal.

This is the second of two related enhancements opening up `RowRanges`/reader APIs needed by the materialization feature described above. The first (#3596) added `RowRanges.Builder` for incremental construction from selected row indices.

### Component(s):
**Core**

peter-toth:
### Rationale for this change

This PR is based on @mbutrovich's previous work.

Opening up an API needed by a later materialization feature in Spark. External readers (e.g. a Spark-side scanner) need the column-index-derived row ranges that may pass the configured filter for a row group, so they can plan which pages to read without reading column data themselves.

`getRowRanges(int)` already existed as a private helper; this change makes it public and gives it a well-defined behavior when no filter is configured.

- `getRowRanges(int blockIndex)`: made public; returns the row ranges within the row group that may pass the configured filter. The computation is metadata-only (consults the column index from the footer; no column data is read). With no filter configured, it shortcuts to a `RowRanges` covering all rows of the row group rather than asserting that a filter is present.

### Are these changes tested?

Yes. `TestParquetFileReaderRowRanges` verifies that, with no filter configured, `getRowRanges` returns ranges covering all rows of the row group.

No.

Closes #3598

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -1518,46 +1518,6 @@ public class ParquetFileReader implements Closeable {
     return rowRanges;
   }
 
-  /**
-   * Returns the total compressed byte count of this reader's requested columns' pages whose
-   * row ranges intersect {@code rowRanges} within the given row group. The set of columns is
-   * taken from the reader's currently configured requested schema (see
-   * {@link #setRequestedSchema}). Metadata-only: consults each column's {@link OffsetIndex}
-   * from the file footer; no column data is read.
-   *
-   * <p>Page size here is {@link OffsetIndex#getCompressedPageSize} (includes page header).
-   * Dictionary pages are not represented in {@link OffsetIndex} and are therefore excluded
-   * from the sum.
-   *
-   * @param blockIndex row group index
-   * @param rowRanges  row ranges to intersect against pages
-   * @return sum of compressed page sizes across requested columns for pages overlapping
-   *         {@code rowRanges}
-   * @throws ColumnIndexStore.MissingOffsetIndexException if any requested column lacks an
-   *         offset index
-   */
-  public long getCompressedBytesForRowRanges(int blockIndex, RowRanges rowRanges) {
-    if (rowRanges.rowCount() == 0 || paths.isEmpty()) {
-      return 0L;
-    }
-    BlockMetaData block = blocks.get(blockIndex);
-    long blockRowCount = block.getRowCount();
-    ColumnIndexStore ciStore = getColumnIndexStore(blockIndex);
-    long total = 0L;
-    for (ColumnPath path : paths.keySet()) {
-      OffsetIndex offsetIndex = ciStore.getOffsetIndex(path);
-      int pageCount = offsetIndex.getPageCount();
-      for (int i = 0; i < pageCount; i++) {
-        long from = offsetIndex.getFirstRowIndex(i);
-        long to = offsetIndex.getLastRowIndex(i, blockRowCount);
-        if (rowRanges.isOverlapping(from, to)) {
-          total += offsetIndex.getCompressedPageSize(i);
-        }
-      }
-    }
-    return total;
-  }
-
   public boolean skipNextRowGroup() {
     return advanceToNextBlock();
   }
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1504,6 +1518,46 @@ private RowRanges getRowRanges(int blockIndex) {
...
+   * from the sum.
+   *
+   * @param blockIndex row group index
+   * @param rowRanges  row ranges to intersect against pages
+   * @return sum of compressed page sizes across requested columns for pages overlapping
+   *         {@code rowRanges}
+   * @throws ColumnIndexStore.MissingOffsetIndexException if any requested column lacks an
+   *         offset index
+   */
+  public long getCompressedBytesForRowRanges(int blockIndex, RowRanges rowRanges) {
```
wgtmac: Do we really want to maintain this? It seems that all methods used in it are publicly available. Application code can just write this up there.

peter-toth: @wgtmac, I can move this part to Spark.
What's your take on the `getRowRanges()` change in this PR and the [other](https://github.com/apache/parquet-java/pull/3597) PR? Shall I combine them into one?

wgtmac: I‘m in favor of keeping them as separate PRs.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1489,9 +1489,23 @@ public ColumnIndexStore getColumnIndexStore(int blockIndex) {
...
+   *
+   * @param blockIndex the row group (block) index
+   * @return row ranges within the block that may pass the configured filter
+   */
+  public RowRanges getRowRanges(int blockIndex) {
+    if (!FilterCompat.isFilteringRequired(options.getRecordFilter())) {
+      return RowRanges.createSingle(blocks.get(blockIndex).getRowCount());
+    }
     RowRanges rowRanges = blockRowRanges.get(blockIndex);
     if (rowRanges == null) {
```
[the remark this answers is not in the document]
peter-toth: Fixed in https://github.com/apache/parquet-java/pull/3599/commits/b04cbb81e8a26eda15f2aa2d82e67f0d0e4deeca.

wgtmac (approved):
+1

peter-toth:
Thank you @wgtmac for the approval.
I wonder how we can take this PR forward and what is your take on https://github.com/apache/parquet-java/pull/3597?

peter-toth:
Thank you for the review @wgtmac.

PR #3607, opened 2026-06-07
GH-3601: Avoid repeated created_by parsing in footer metadata conversion

### LINKED ISSUE #3601: Optimize repeated shouldIgnoreStatistics calls during footer reading in ParquetMetadataConverter

`CorruptStatistics.shouldIgnoreStatistics(String createdBy, PrimitiveTypeName columnType)` performs `VersionParser.parse() `and `SemanticVersion.parse()` on every invocation. The `createdBy` string is constant per file (from `FileMetaData.created_by`), but this method is called once per column chunk per row group during `ParquetMetadataConverter.fromParquetMetadata()`.

For a file with R row groups and C columns, the version string is parsed R×C times during file open — all yielding the same result.

**Impact**
High CPU during `ParquetFileReader` construction on files with many high groups/columns.

**Proposed fix**
Compute `shouldIgnoreStatistics` once per file before the row group loop in `fromParquetMetadata`, and pass the pre-computed boolean through `buildColumnChunkMetaData` → `fromParquetStatisticsInternal`.

Since `buildColumnChunkMetaData` and `fromParquetStatisticsInternal` are public/package-level API methods enforced by japicmp-maven-plugin, we will add overloaded methods that accept a boolean `shouldIgnoreBinaryStats` parameter rather than changing existing signatures. The existing String `createdBy` signatures remain for backward compatibility and delegate to the new overloads.

The page-level path in `ParquetFileReader.Chunk.readAllPages()` also calls `shouldIgnoreStatistics` via `fromParquetStatisticsInternal`, but is not in scope for this fix — at read time the cost is masked by I/O, decompression, and decoding. The footer path during file open is where R×C calls happen in a tight loop with no I/O to amortize the cost.

Alternatively, caching on `createdBy` inside `shouldIgnoreStatistics` would also eliminate redundant parsing — with the advantage that parsing occurs only once per distinct version string across the entire process lifetime (rather than once per file open). 

However, the caller-side fix is preferred because it avoids per-call hash/equals overhead on the hot path (R×C times per file),  and keeps the optimization local to the scope where the redundancy occurs.

**Profiling data**
`shouldIgnoreStatistics` accounts for ~65% of `fromParquetMetadata` CPU time across multiple samples.
Image

### Component(s)

Core

yadavay-amzn:
## Summary

Footer conversion previously evaluated `CorruptStatistics.shouldIgnoreStatistics(createdBy, columnType)` per (row-group x column), re-parsing the `created_by` version string on every call.

This PR factors `CorruptStatistics` into two orthogonal checks:
- `fileHasCorruptStatistics(createdBy)` - file-level, parsed once
- `isCorruptStatisticsColumnType(columnType)` - per-column type gate

The file-level corrupt-stats flag is precomputed once in `fromParquetMetadata`, gated on the schema containing at least one affected column type (BINARY or FIXED_LEN_BYTE_ARRAY), so binary-free files skip the `created_by` parsing entirely and avoid spurious PARQUET-251 warnings. The boolean is threaded through `buildColumnChunkMetaData`/`fromParquetStatisticsInternal` instead of re-parsing `created_by` per column.

Note: an earlier process-wide cache approach was dropped per review feedback.

## Tests
- `CorruptStatisticsTest`: unit tests for the factored methods (`fileHasCorruptStatistics`, `isCorruptStatisticsColumnType`)
- `TestParquetMetadataConverter#testCorruptStatsPerColumnGate`: verifies per-column gate behavior with corrupt/non-corrupt files
- `TestParquetMetadataConverter#testSchemaGateSkipsCorruptStatsCheckForNonBinarySchema`: verifies that an INT-only schema with a known-corrupt `created_by` preserves statistics (schema gate prevents unnecessary parsing)

Closes #3601

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -47,20 +47,25 @@ public class CorruptStatistics {
   private static final SemanticVersion CDH_5_PARQUET_251_FIXED_END = new SemanticVersion(1, 5, 0);
 
   /**
-   * Decides if the statistics from a file created by createdBy (the created_by field from parquet format)
-   * should be ignored because they are potentially corrupt.
+   * Returns whether the given column type is one of the types affected by the PARQUET-251 bug
+   * (BINARY or FIXED_LEN_BYTE_ARRAY).
    *
-   * @param createdBy  the created-by string from a file footer
-   * @param columnType the type of the column that this is checking
-   * @return true if the statistics may be invalid and should be ignored, false otherwise
+   * @param columnType the primitive type of the column
+   * @return true if this column type could have corrupt statistics
    */
-  public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName columnType) {
+  public static boolean isCorruptStatisticsColumnType(PrimitiveTypeName columnType) {
+    return columnType == PrimitiveTypeName.BINARY || columnType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
+  }
 
-    if (columnType != PrimitiveTypeName.BINARY && columnType != PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
-      // the bug only applies to binary columns
-      return false;
-    }
-
+  /**
+   * Determines whether a file (identified by its created_by string) was written by a version of
+   * parquet-mr that had the PARQUET-251 statistics bug. This is a file-level check that does not
+   * consider column type.
+   *
+   * @param createdBy the created-by string from a file footer
+   * @return true if the file was written by a version with the corrupt statistics bug
+   */
+  public static boolean fileHasCorruptStatistics(String createdBy) {
     if (Strings.isNullOrEmpty(createdBy)) {
       // created_by is not populated, which could have been caused by
       // parquet-mr during the same time as PARQUET-251, see PARQUET-297
@@ -101,6 +106,22 @@ public class CorruptStatistics {
       warnParseErrorOnce(createdBy, e);
       return true;
     }
+  }
+
+  /**
+   * Decides if the statistics from a file created by createdBy (the created_by field from parquet format)
+   * should be ignored because they are potentially corrupt.
+   *
+   * @param createdBy  the created-by string from a file footer
+   * @param columnType the type of the column that this is checking
+   * @return true if the statistics may be invalid and should be ignored, false otherwise
+   */
+  public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName columnType) {
+    if (!isCorruptStatisticsColumnType(columnType)) {
+      // the bug only applies to binary columns
+      return false;
+    }
+    return fileHasCorruptStatistics(createdBy);
   }
 
   private static void warnParseErrorOnce(String createdBy, Throwable e) {
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -49,23 +49,20 @@ public class CorruptStatistics {
   /**
    * Returns whether the given column type is one of the types affected by the PARQUET-251 bug
    * (BINARY or FIXED_LEN_BYTE_ARRAY).
-   *
-   * @param columnType the primitive type of the column
-   * @return true if this column type could have corrupt statistics
    */
-  public static boolean isCorruptStatisticsColumnType(PrimitiveTypeName columnType) {
+  private static boolean isCorruptStatisticsColumnType(PrimitiveTypeName columnType) {
     return columnType == PrimitiveTypeName.BINARY || columnType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
   }
 
   /**
-   * Determines whether a file (identified by its created_by string) was written by a version of
-   * parquet-mr that had the PARQUET-251 statistics bug. This is a file-level check that does not
-   * consider column type.
+   * Determines whether a file (identified by its created_by string) may have been written by a
+   * version of parquet-mr that had the PARQUET-251 statistics bug. This is a file-level check
+   * that does not consider column type.
    *
    * @param createdBy the created-by string from a file footer
-   * @return true if the file was written by a version with the corrupt statistics bug
+   * @return true if the file may have been written by a version with the corrupt statistics bug
    */
-  public static boolean fileHasCorruptStatistics(String createdBy) {
+  public static boolean mayHaveCorruptStatistics(String createdBy) {
     if (Strings.isNullOrEmpty(createdBy)) {
       // created_by is not populated, which could have been caused by
       // parquet-mr during the same time as PARQUET-251, see PARQUET-297
@@ -121,7 +118,7 @@ public class CorruptStatistics {
       // the bug only applies to binary columns
       return false;
     }
-    return fileHasCorruptStatistics(createdBy);
+    return mayHaveCorruptStatistics(createdBy);
   }
 
   private static void warnParseErrorOnce(String createdBy, Throwable e) {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -947,6 +947,13 @@ public class ParquetMetadataConverter {
   // Visible for testing
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
       String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
+    boolean fileHasCorruptStats = CorruptStatistics.fileHasCorruptStatistics(createdBy);
+    return fromParquetStatisticsInternal(formatStats, type, typeSortOrder, fileHasCorruptStats);
+  }
+
+  // Core implementation using pre-computed file-level corrupt stats flag
+  static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
+      Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder, boolean fileHasCorruptStats) {
     // create stats object based on the column type
     org.apache.parquet.column.statistics.Statistics.Builder statsBuilder =
         org.apache.parquet.column.statistics.Statistics.getBuilderForReading(type);
@@ -969,8 +976,10 @@ public class ParquetMetadataConverter {
         // valid with the type's sort order. In previous releases, all stats were
         // aggregated using a signed byte-wise ordering, which isn't valid for all the
         // types (e.g. strings, decimals etc.).
-        if (!CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName())
-            && (sortOrdersMatch || maxEqualsMin)) {
+        // The fileHasCorruptStats flag applies only to BINARY and FIXED_LEN_BYTE_ARRAY columns.
+        boolean ignoreForThisColumn =
+            fileHasCorruptStats && CorruptStatistics.isCorruptStatisticsColumnType(type.getPrimitiveTypeName());
+        if (!ignoreForThisColumn && (sortOrdersMatch || maxEqualsMin)) {
           if (isSet) {
             statsBuilder.withMin(formatStats.min.array());
             statsBuilder.withMax(formatStats.max.array());
@@ -992,46 +1001,6 @@ public class ParquetMetadataConverter {
       String createdBy, Statistics statistics, PrimitiveType type) {
     SortOrder expectedOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
     return fromParquetStatisticsInternal(createdBy, statistics, type, expectedOrder);
-  }
-
-  // Overload that uses a pre-computed shouldIgnoreCorruptStats flag to avoid redundant parsing
-  private org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
-      String createdBy, Statistics formatStats, PrimitiveType type, boolean shouldIgnoreCorruptStats) {
-    SortOrder typeSortOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
-    org.apache.parquet.column.statistics.Statistics.Builder statsBuilder =
-        org.apache.parquet.column.statistics.Statistics.getBuilderForReading(type);
-
-    if (formatStats != null) {
-      if (formatStats.isSetMin_value() && formatStats.isSetMax_value()) {
-        byte[] min = formatStats.min_value.array();
-        byte[] max = formatStats.max_value.array();
-        if (isMinMaxStatsSupported(type) || Arrays.equals(min, max)) {
-          statsBuilder.withMin(min);
-          statsBuilder.withMax(max);
-        }
-      } else {
-        boolean isSet = formatStats.isSetMax() && formatStats.isSetMin();
-        boolean maxEqualsMin = isSet ? Arrays.equals(formatStats.getMin(), formatStats.getMax()) : false;
-        boolean sortOrdersMatch = SortOrder.SIGNED == typeSortOrder;
-        // The shouldIgnoreCorruptStats flag applies only to BINARY and FIXED_LEN_BYTE_ARRAY.
-        // For other types, shouldIgnoreStatistics always returns false, so we only guard those.
-        PrimitiveTypeName primitiveTypeName = type.getPrimitiveTypeName();
-        boolean ignoreForThisColumn = shouldIgnoreCorruptStats
-            && (primitiveTypeName == PrimitiveTypeName.BINARY
-                || primitiveTypeName == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY);
-        if (!ignoreForThisColumn && (sortOrdersMatch || maxEqualsMin)) {
-          if (isSet) {
-            statsBuilder.withMin(formatStats.min.array());
-            statsBuilder.withMax(formatStats.max.array());
-          }
-        }
-      }
-
-      if (formatStats.isSetNull_count()) {
-        statsBuilder.withNumNulls(formatStats.null_count);
-      }
-    }
-    return statsBuilder.build();
   }
 
   GeospatialStatistics toParquetGeospatialStatistics(
@@ -1858,24 +1827,23 @@ public class ParquetMetadataConverter {
 
   public ColumnChunkMetaData buildColumnChunkMetaData(
       ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, String createdBy) {
-    boolean shouldIgnoreCorruptStats =
-        CorruptStatistics.shouldIgnoreStatistics(createdBy, PrimitiveTypeName.BINARY);
-    return buildColumnChunkMetaData(metaData, columnPath, type, createdBy, shouldIgnoreCorruptStats);
+    boolean fileHasCorruptStats = CorruptStatistics.fileHasCorruptStatistics(createdBy);
+    return buildColumnChunkMetaData(metaData, columnPath, type, fileHasCorruptStats);
   }
 
   ColumnChunkMetaData buildColumnChunkMetaData(
       ColumnMetaData metaData,
       ColumnPath columnPath,
       PrimitiveType type,
-      String createdBy,
-      boolean shouldIgnoreCorruptStats) {
+      boolean fileHasCorruptStats) {
+    SortOrder typeSortOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
     return ColumnChunkMetaData.get(
         columnPath,
         type,
         fromFormatCodec(metaData.codec),
         convertEncodingStats(metaData.getEncoding_stats()),
         fromFormatEncodings(metaData.encodings),
-        fromParquetStatisticsInternal(createdBy, metaData.statistics, type, shouldIgnoreCorruptStats),
+        fromParquetStatisticsInternal(metaData.statistics, type, typeSortOrder, fileHasCorruptStats),
         metaData.data_page_offset,
         metaData.dictionary_page_offset,
         metaData.num_values,
@@ -1904,10 +1872,11 @@ public class ParquetMetadataConverter {
     MessageType messageType = fromParquetSchema(parquetMetadata.getSchema(), parquetMetadata.getColumn_orders());
     List<BlockMetaData> blocks = new ArrayList<BlockMetaData>();
     List<RowGroup> row_groups = parquetMetadata.getRow_groups();
-    // Compute once per file: the result is the same for BINARY and FIXED_LEN_BYTE_ARRAY
-    // (the only types affected by PARQUET-251), and always false for other types.
-    boolean shouldIgnoreCorruptStats =
-        CorruptStatistics.shouldIgnoreStatistics(parquetMetadata.getCreated_by(), PrimitiveTypeName.BINARY);
+    // Compute once per file: whether this file was written by a version with the PARQUET-251 bug.
+    // Only parse created_by if the schema has columns affected by the bug (BINARY/FIXED_LEN_BYTE_ARRAY).
+    // The per-column type check is applied later when statistics are actually read.
+    boolean fileHasCorruptStats = schemaHasCorruptStatisticsColumnType(messageType)
+        && CorruptStatistics.fileHasCorruptStatistics(parquetMetadata.getCreated_by());
 
     if (row_groups != null) {
       for (RowGroup rowGroup : row_groups) {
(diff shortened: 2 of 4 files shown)


parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -35,6 +36,8 @@
  */
 public class CorruptStatistics {
   private static final AtomicBoolean alreadyLogged = new AtomicBoolean(false);
+  private static final ConcurrentHashMap<String, Boolean> CACHE = new ConcurrentHashMap<>();
```
wgtmac: Instead of caching the result, is it better to parse this once and cache a flag somewhere in the reader so that it is shared by all columns associated with that reader?

asifsmohammed (commented):
I generally like the approach @yadavay-amzn, here's why I would not use cache alone as I mentioned in the issue. It's better than the current approach of checking it RGxCC times

- In high workload scenarios we are going to call `shouldIgnoreStatistics` RGxCC times i.e. we are going to call `.size` and `.computeIfAbsent`(hash + equals) that many times which is still not ideal.

The cleaner fix would be at the caller side in `ParquetMetadataConverter.fromParquetMetadata()`: compute `shouldIgnoreStatistics` once before the row group loop using the file-level `createdBy`, then pass the pre-computed boolean through `buildColumnChunkMetaData` → `fromParquetStatisticsInternal`. This eliminates all per-column overhead — no hash, no lookup, just a boolean flowing through the call chain.

This eliminates the need to check `shouldIgnoreStatistics` for every RGxCC. Both approaches could coexist (cache helps other callers), but the caller-side fix is where we see real improvements

yadavay-amzn:
@asifsmohammed Thanks for reviewing! I am an agreement with your suggestions. 

Removed the global static cache entirely. Now computes `shouldIgnoreStatistics` once per file in `fromParquetMetadata` before the row-group loop and passes the pre-computed flag through `buildColumnChunkMetaData` to `fromParquetStatisticsInternal`.

cc @wgtmac

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
yadavay-amzn: Removed the global static cache entirely. Now computes `shouldIgnoreStatistics` once per file in `fromParquetMetadata` before the row-group loop and passes the pre-computed flag through `buildColumnChunkMetaData` to `fromParquetStatisticsInternal`.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -970,6 +970,46 @@ public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
     return fromParquetStatisticsInternal(createdBy, statistics, type, expectedOrder);
   }
 
+  // Overload that uses a pre-computed shouldIgnoreCorruptStats flag to avoid redundant parsing
+  private org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
```
asif-moh: Instead of duplicating the entire `fromParquetStatisticsInternal` body, the existing method can simply delegate to a new overload and this eliminates duplicate code

```
static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
      String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
    return fromParquetStatisticsInternal(
        formatStats, 
        type, 
        typeSortOrder, 
        CorruptStatistics.fileHasCorruptStatistics(createdBy) // This is a new method in CorruptStatistics
    );
}


// overloaded method
static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
      Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder, boolean fileHasCorruptStats) {
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1829,6 +1880,10 @@ public ParquetMetadata fromParquetMetadata(
     MessageType messageType = fromParquetSchema(parquetMetadata.getSchema(), parquetMetadata.getColumn_orders());
     List<BlockMetaData> blocks = new ArrayList<BlockMetaData>();
     List<RowGroup> row_groups = parquetMetadata.getRow_groups();
+    // Compute once per file: the result is the same for BINARY and FIXED_LEN_BYTE_ARRAY
+    // (the only types affected by PARQUET-251), and always false for other types.
+    boolean shouldIgnoreCorruptStats =
+        CorruptStatistics.shouldIgnoreStatistics(parquetMetadata.getCreated_by(), PrimitiveTypeName.BINARY);
```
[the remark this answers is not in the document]
asif-moh: This is calling `shouldIgnoreStatistics` with `PrimitiveTypeName.BINARY` hardcoded which is incorrect.

Instead we can refactor `shouldIgnoreStatistics` by adding public methods. 
```
public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName columnType) {
    if (!isCorruptStatisticsColumnType(columnType)) {
      // the bug only applies to binary columns
      return false;
    }
    return fileHasCorruptStatistics(createdBy);
}

public static boolean isCorruptStatisticsColumnType(PrimitiveTypeName columnType) {
    return columnType == PrimitiveTypeName.BINARY || columnType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
}

public static boolean fileHasCorruptStatistics(String createdBy) {
    // rest of the logic from shouldIgnoreStatistics
}
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
asif-moh, line 1952: We no longer need to pass `createdBy` downstream

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -1794,13 +1834,24 @@ public FileMetaDataAndRowGroupOffsetInfo visit(RangeMetadataFilter filter) throw
 
   public ColumnChunkMetaData buildColumnChunkMetaData(
       ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, String createdBy) {
+    boolean shouldIgnoreCorruptStats =
+        CorruptStatistics.shouldIgnoreStatistics(createdBy, PrimitiveTypeName.BINARY);
+    return buildColumnChunkMetaData(metaData, columnPath, type, createdBy, shouldIgnoreCorruptStats);
+  }
+
+  ColumnChunkMetaData buildColumnChunkMetaData(
```
asif-moh, line 1835: `buildColumnChunkMetaData` can delegate to a package-private overload that takes the boolean similar to what you have done but with few changes,
```
public ColumnChunkMetaData buildColumnChunkMetaData(
      ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, String createdBy) {
    return buildColumnChunkMetaData(
        metaData, columnPath, type, CorruptStatistics.fileHasCorruptStatistics(createdBy));
}

ColumnChunkMetaData buildColumnChunkMetaData(
      ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, boolean fileHasCorruptStats) {
    SortOrder expectedOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
    return ColumnChunkMetaData.get(...,
        fromParquetStatisticsInternal(metaData.statistics, type, expectedOrder, fileHasCorruptStats), ...);
}
```
No need to pass createdBy downstream, the boolean is all the internal overload needs. SortOrder computation moves here since we bypass `fromParquetStatistics` to avoid re-parsing createdBy as you have already done by replacing `fromParquetStatisticsInternal` with `fromParquetStatistics`.

Also notice how the new public methods we extracted in `CorruptStatistics` are being used in each delegate method here

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -970,6 +970,46 @@ public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
...
+        boolean isSet = formatStats.isSetMax() && formatStats.isSetMin();
+        boolean maxEqualsMin = isSet ? Arrays.equals(formatStats.getMin(), formatStats.getMax()) : false;
+        boolean sortOrdersMatch = SortOrder.SIGNED == typeSortOrder;
+        // The shouldIgnoreCorruptStats flag applies only to BINARY and FIXED_LEN_BYTE_ARRAY.
+        // For other types, shouldIgnoreStatistics always returns false, so we only guard those.
+        PrimitiveTypeName primitiveTypeName = type.getPrimitiveTypeName();
+        boolean ignoreForThisColumn = shouldIgnoreCorruptStats
+            && (primitiveTypeName == PrimitiveTypeName.BINARY
+                || primitiveTypeName == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY);
+        if (!ignoreForThisColumn && (sortOrdersMatch || maxEqualsMin)) {
```
asif-moh: This check in `shouldIgnoreStatistics` is dead code with current changes as we always pass `BINARY`
```
if (columnType != PrimitiveTypeName.BINARY && columnType !=PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY)
```
We can utilize the new methods here. Please refer to below comments for context. 

```
if (!(fileHasCorruptStats && CorruptStatistics.isCorruptStatisticsColumnType(type.getPrimitiveTypeName()))
```

Instead of calling or moving the `PrimitiveTypeName` checks to `ParquetMetadataConverter` and leave it as the responsibility of `CorruptStatistics`
```
if (!CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName())
```

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
yadavay-amzn: Done. Removed the duplicated body; the `createdBy` overload now delegates to the boolean overload via `CorruptStatistics.fileHasCorruptStatistics(createdBy)`.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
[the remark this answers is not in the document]
yadavay-amzn: Done. Added `isCorruptStatisticsColumnType(columnType)` and `fileHasCorruptStatistics(createdBy)`, with `shouldIgnoreStatistics` delegating to both. The file-level flag is computed via `fileHasCorruptStatistics` (no hardcoded BINARY), and the per-column BINARY/FIXED_LEN_BYTE_ARRAY check is applied where stats are converted.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
yadavay-amzn: Done. The precomputed boolean is threaded instead; `createdBy` is no longer passed downstream.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
yadavay-amzn: Done. `buildColumnChunkMetaData` now delegates to a package-private overload taking the boolean, with the SortOrder computation moved into it; the public `(..., String createdBy)` overload is preserved for back-compat.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
yadavay-amzn: Done. Removed the now-dead column-type check; the gate is expressed via `CorruptStatistics.isCorruptStatisticsColumnType` in `fromParquetStatisticsInternal`.

yadavay-amzn:
@asif-moh @wgtmac Addressed the feedback: factored `CorruptStatistics` into `fileHasCorruptStatistics` + `isCorruptStatisticsColumnType` (with `shouldIgnoreStatistics` delegating), precompute the file-level flag once in `fromParquetMetadata` (schema-gated so binary-free files do no `created_by` parsing), thread a boolean through `buildColumnChunkMetaData`/`fromParquetStatisticsInternal`, drop the duplicated body and dead code, and fix the per-column gate so non-binary columns are not ignored. Updated the title/description to match. PTAL.

asif-moh (approved):
LGTM! Thanks for making the changes

asifsmohammed (approved):

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -47,20 +47,25 @@ public class CorruptStatistics {
...
+   * @param columnType the primitive type of the column
+   * @return true if this column type could have corrupt statistics
    */
-  public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName columnType) {
-
-    if (columnType != PrimitiveTypeName.BINARY && columnType != PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
-      // the bug only applies to binary columns
-      return false;
-    }
+  public static boolean isCorruptStatisticsColumnType(PrimitiveTypeName columnType) {
```
wgtmac: IMHO, we need to be cautious when adding a new public method because we have to maintain it forever. Can we remove it from here and just let downstream where calls it to use a private method instead?

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -47,20 +47,25 @@ public class CorruptStatistics {
...
 
+  /**
+   * Determines whether a file (identified by its created_by string) was written by a version of
+   * parquet-mr that had the PARQUET-251 statistics bug. This is a file-level check that does not
+   * consider column type.
+   *
+   * @param createdBy the created-by string from a file footer
+   * @return true if the file was written by a version with the corrupt statistics bug
+   */
+  public static boolean fileHasCorruptStatistics(String createdBy) {
```
wgtmac: Rename it to `mayHaveCorruptStatistics` ?

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
asif-moh: We need `fileHasCorruptStatistics` (or `mayHaveCorruptStatistics` after rename) to remain public, otherwise we'd have to duplicate the version-parsing logic in `ParquetMetadataConverter`. We can make `isCorruptStatisticsColumnType` private and inline the column-type check downstream. 
Or we can overload `shouldIgnoreStatistics(String createdBy)` but either way we will have 1 new public method.

@yadavay-amzn @wgtmac wdyt?

yadavay-amzn: Went with your suggestion - kept mayHaveCorruptStatistics public (ParquetMetadataConverter needs it) and made isCorruptStatisticsColumnType private, inlining the column-type check (BINARY || FIXED_LEN_BYTE_ARRAY) at its call sites. Net new public API is just mayHaveCorruptStatistics(String). Behavior unchanged; parquet-column + parquet-hadoop tests pass.

yadavay-amzn:
@wgtmac addressed the API-minimization and rename: fileHasCorruptStatistics is renamed to mayHaveCorruptStatistics and kept public (ParquetMetadataConverter relies on it, otherwise the version-parsing logic would be duplicated). isCorruptStatisticsColumnType is now private with the column-type check inlined at its call sites, so the only new public method this PR adds is mayHaveCorruptStatistics(String). Behavior is unchanged; tests pass. PTAL.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
yadavay-amzn: Renamed to mayHaveCorruptStatistics.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
yadavay-amzn: Minimized the new public surface: isCorruptStatisticsColumnType is now private with the column-type check inlined at its call sites, so mayHaveCorruptStatistics(String) is the only new public method. It has to stay public because ParquetMetadataConverter relies on it (otherwise the version-parsing logic would be duplicated).

wgtmac:
Could you resolve the conflict?

yadavay-amzn:
@wgtmac rebased onto latest master and resolved the conflict (it was only in TestParquetMetadataConverter - kept master's new NaN/IEEE754 tests plus the two GH-3601 tests). The GH-3601 change is intact; parquet-column + parquet-hadoop tests pass and spotless is clean.

asifsmohammed:
@wgtmac, @yadavay-amzn has addressed your feedback and rebased onto master. Is there anything else 
pending on this PR, or is it ready to merge? We'd love to get this included in the next release if possible. TY!

wgtmac (changes requested):
Thanks for the updates! I took another look and I still do not think this is the right shape yet.

I have to admit that my earlier suggestion may have pushed this in the wrong direction. I suggested factoring out a file-level corrupt-stats check, but I do not think we should merge a partial solution with side effect. My concerns with the current implementation are:

1. `CorruptStatistics.mayHaveCorruptStatistics` adds a new public API for an internal optimization. It is tied to `PARQUET-251` semantics, so we would need to maintain that specific meaning as public API.

2. `mayHaveCorruptStatistics` is not a pure predicate. It can log warnings and mutate `alreadyLogged`. Precomputing it can therefore have user-visible side effects.

3. The fix is still partial. The main footer path avoids the row-group x column parse, but other production paths still parse `created_by` on each call. Examples include `fromParquetStatisticsInternal(String createdBy, ...)`, `buildColumnChunkMetaData(..., String createdBy)`, page-level stats in `ParquetFileReader`, and lazy encrypted metadata in `ColumnChunkMetaData`.

4. The current `fileHasCorruptStats` boolean leaks too much PARQUET-251-specific logic into `ParquetMetadataConverter`. The converter now knows about the schema gate, BINARY / FIXED_LEN_BYTE_ARRAY, and the file-level corrupt-stats flag. This pattern will not scale well if we later need another writer-version-based statistics workaround.

5. The schema-gate test does not prove the schema gate. `testSchemaGateSkipsCorruptStatsCheckForNonBinarySchema` only proves INT stats are preserved. That would still pass because of the later per-column gate. It does not prove that parsing or the corrupt-stats check was skipped.

I have not had enough time to fully design the final API, so please treat the following as a possible direction, not a complete design. A cleaner approach may be to cache the parsed `created_by` result, not a PARQUET-251-specific boolean.

Java already has `VersionParser.ParsedVersion` in `parquet-common`. `parquet-hadoop` already depends on `parquet-common`. There is also precedent in `CorruptDeltaByteArrays.requiresSequentialReads(ParsedVersion, Encoding)`.

The converter could parse `created_by` once per file into a `ParsedVersion`, pass that `ParsedVersion` through the converter, and keep the PARQUET-251 decision inside `CorruptStatistics`.

For example:

```java
    fromParquetStatisticsInternal(writerVersion, metaData.statistics, type, typeSortOrder)
```

And `CorruptStatistics` could have:

```java
    shouldIgnoreStatistics(ParsedVersion writerVersion, PrimitiveTypeName columnType)
```

The existing String-based API can stay as a compatibility wrapper. This keeps the optimization in `ParquetMetadataConverter`, keeps the version-specific logic in `CorruptStatistics`, and avoids exposing `mayHaveCorruptStatistics` or threading a `PARQUET-251` boolean through generic metadata conversion code.

asif-moh:
@wgtmac  Thanks for the review. I agree with the overall direction toward `ParsedVersion`. Let me share findings from my investigation here,
  
  1. mayHaveCorruptStatistics as public API

  Agreed, exposing a PARQUET-251-specific predicate as public API is the wrong abstraction. It ties us to maintaining that exact semantic contract. The ParsedVersion-based approach keeps this internal.

  2. mayHaveCorruptStatistics is not a pure predicate

  I looked at this more closely and I believe the logging behavior is actually unchanged between master and this PR. The schema gate (`schemaHasCorruptStatisticsColumnType`) ensures `mayHaveCorruptStatistics` is only called when the schema has `BINARY`/`FIXED_LEN_BYTE_ARRAY` columns, the same precondition that would trigger the warning on master. In all 4 cases (corrupt/non-corrupt × has-binary/no-binary), the warning fires or doesn't fire in identical scenarios. The only difference is microsecond-level timing within the same `fromParquetMetadata` call.

That said, I agree the `ParsedVersion` approach is cleaner here regardless parsing is side-effect-free, and the warning stays inside `shouldIgnoreStatistics` where it's always been.

  3. The fix is partial

  I investigated this further and the problem is broader than I initially realized. `VersionParser.parse(createdBy)` is called from 6 distinct production sites, all parsing the same constant string from `FileMetaData.getCreatedBy()`:

  | # | Call site | Frequency per file | When |
  |---|-----------|-------------------|------|
  | 1 | `CorruptStatistics.shouldIgnoreStatistics` via `buildColumnChunkMetaData` | R × C | Footer decode |
  | 2 | `CorruptStatistics.shouldIgnoreStatistics` via `ParquetFileReader.readAllPages` | Pages per column chunk | Page read |
  | 3 | `CorruptStatistics.shouldIgnoreStatistics` via `ParquetRewriter.convertStatistics` | Pages per rewritten chunk | Rewrite |
  | 4 | `CorruptStatistics.shouldIgnoreStatistics` via `EncryptedColumnChunkMetaData.decryptIfNeeded` | 1 per encrypted column | Lazy decrypt |
  | 5 | `CorruptDeltaByteArrays.requiresSequentialReads(String, Encoding)` in `ParquetRecordReader` | 1 per reader init (parse only if DELTA_BYTE_ARRAY present) | Reader init |
  | 6 | `ColumnReadStoreImpl` constructor via `MessageColumnIO.getRecordReader` | R (once per row group read) | Row group materialization |
  | 7 | `ColumnReadStoreImpl` constructor via `ParquetRewriter.nullifyColumn` | 1 per nullified column | Rewrite with nullification |

Notably, `ColumnReadStoreImpl` already parses `createdBy` into a `ParsedVersion` and stores it as a field but it does so R times (once per row group) because nobody upstream caches the parsed result. The `CorruptDeltaByteArrays.requiresSequentialReads(ParsedVersion, Encoding)` overload already exists and is used from `ColumnReaderBase` this is exactly the pattern you suggested for CorruptStatistics.

  4. PARQUET-251 logic leaking into the converter

  Agreed. Threading a boolean specific to one bug doesn't scale. The ParsedVersion approach keeps the converter generic and it just says "here's the writer version" and lets `CorruptStatistics` decide.

 Proposed approach:
  
I'd like to propose addressing the root cause separately: cache a `ParsedVersion` in `FileMetaData`. Since `FileMetaData` is constructed once per file and already stores the `createdBy` string, it's the natural place to parse once and cache:

```
 // In FileMetaData:
private final transient ParsedVersion writerVersion;
  
public FileMetaData(
.... {
this.createdBy = createdBy;
this.writerVersion = parseVersion(createdBy);  // ← just add this
...
}

public ParsedVersion getWriterVersion() {
    return writerVersion;
}
  
private static ParsedVersion parseVersion(String createdBy) {
  if (Strings.isNullOrEmpty(createdBy)) return null;
  try {
      return VersionParser.parse(createdBy);
  } catch (RuntimeException | VersionParseException e) {
      return null;
}
```
  
This is purely additive (new field + getter), doesn't break serialization (transient), and enables all 7 call sites to stop re-parsing. `ColumnReadStoreImpl` could accept `ParsedVersion` directly instead of re-parsing in its constructor, and `CorruptStatistics` would gain a `shouldIgnoreStatistics(ParsedVersion, PrimitiveTypeName)` overload following the established `CorruptDeltaByteArrays.requiresSequentialReads(ParsedVersion, Encoding)` pattern. 
  
Would you be open to a separate issue/PR for adding `ParsedVersion` caching to `FileMetaData`? That would be a small, zero-behavior-change change just parsing `createdBy` in the existing constructor body and exposing a  getter. The ParsedVersion-based overloads for `CorruptStatistics`, `ColumnReadStoreImpl`, etc. can be added incrementally after that, and this PR could then rebase onto the foundation cleanly.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -2159,4 +2159,110 @@ public void testColumnIndexNanCountsRoundTrip() {
...
+    byte[] maxBytes = new byte[] {4, 5, 6, 7};
+    formatStats.setMin(minBytes);
+    formatStats.setMax(maxBytes);
+    formatStats.setNull_count(5);
+
+    // For BINARY column: stats should be ignored (only null_count preserved)
+    PrimitiveType binaryType = new PrimitiveType(Repetition.OPTIONAL, PrimitiveTypeName.BINARY, "bin_col");
+    Statistics binaryStats = ParquetMetadataConverter.fromParquetStatisticsInternal(
+        corruptCreatedBy, formatStats, binaryType, ParquetMetadataConverter.SortOrder.SIGNED);
+    assertFalse("BINARY min/max should be ignored for corrupt file", binaryStats.hasNonNullValue());
```
nastra, line 2180: you might need to rebase the PR because we switched all assertions to AssertJ and removed all JUnit4 usages.

PR #3608, opened 2026-06-09
Introduce implementation for FileType

brkyvz:
### Rationale for this change

Introduces the reference implementation for the new LogicalTypeAnnotation File as discussed in this [design document](https://docs.google.com/document/d/1AiwrstqkwkBoOZqgOkm9JGwSMcNeHyLR7EEj1CVqpZQ/edit?tab=t.0#heading=h.k8qyue4jj4rn)

Introduces a new LogicalTypeAnnotation called FILE. It enforces that a group with this type annotation have the following fields by name:
```
uri 		 		STRING	
offset				INT64      
size	 	 		INT64	
content_type		 STRING
checksum		 	STRING
inline 		 		BINARY
```

### Are these changes tested?

Yes, unit tested

Introduces a new LogicalTypeAnnotation called FILE


parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -821,12 +821,34 @@ public THIS addFields(Type... types) {
...
+        if (LogicalTypeAnnotation.FileLogicalTypeAnnotation.PATH_FIELD.equals(fieldName)) {
+          Preconditions.checkArgument(
+              field.getRepetition() == Type.Repetition.REQUIRED,
+              "FILE type field 'path' must be REQUIRED in group '%s'",
+              name);
+          hasPath = true;
+        } else if (!LogicalTypeAnnotation.FileLogicalTypeAnnotation.OPTIONAL_FIELD_NAMES.contains(fieldName)) {
+          throw new IllegalArgumentException(
+              "FILE type group '" + name + "' contains unrecognized field '" + fieldName
+                  + "'. Valid fields are: path, size, offset, etag");
```
divjotarora: nit: stringify `FileLogicalTypeAnnotation.OPTIONAL_FIELD_NAMES` and use it in the error to avoid drift if we add new fields?

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -821,12 +821,34 @@ public THIS addFields(Type... types) {
...
+        if (logicalTypeAnnotation instanceof LogicalTypeAnnotation.FileLogicalTypeAnnotation) {
+          validateFileTypeFields(name, fields);
+        }
         return new GroupType(repetition, name, logicalTypeAnnotation, fields, id);
       } else {
         return new GroupType(repetition, name, getOriginalType(), fields, id);
       }
     }
 
+    private static void validateFileTypeFields(String name, List<Type> fields) {
```
divjotarora, line 833: Where do we validate:

1. If `offset` is present, `size` must be present
2. `offset` and `size` must be non-negative (note spec doesn't technically say `offset` must be >= 0, but I left a suggestion on that change to add it

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuildersWithLogicalTypes.java
divjotarora, line 1: We should add tests for validity of size/offset fields, both their values (>= 0) and valid/invalid combinations

divjotarora (changes requested):

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
brkyvz: addressed

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuildersWithLogicalTypes.java
brkyvz: where can we add value checks during reads and writes? Can you provide any code references please?

emkornfield (approved):

alkis:
I had a look at this against the revised self-reference semantics in apache/parquet-format#603 and ended up writing some of it, so rather than leave it as review comments I opened it as a PR against this branch: brkyvz/parquet-java#1 (delta is just two commits on top of `fileType`).

The main piece is the writer-side API you'd need anyway: `FileValueWriter.write(payload)` returns either inline bytes or an `offset`/`size` pair based on a configurable threshold (`ParquetProperties.withFileSelfReferenceThreshold`), so object models don't choose between the two forms themselves. It has to write the payload eagerly, mid-record — `offset`/`size` are ordinary column values, so once one reaches a `ColumnWriter` it's encoded into a buffered page and a placeholder can't be patched up later.

Three things worth flagging regardless of whether you take the PR:

1. **The AAD uses a per-chunk counter, but the spec defines field 6 as the file offset.** The counter isn't recoverable from the file — nothing stores it, so a reader has to walk the preceding values to rebuild it, which defeats exactly the property the offset-keyed AAD is there for ("available to a reader without counting preceding values... may therefore resolve a self-reference without decoding the pages it skips"). Switching to `offset` also drops the parameter from five signatures, since the value already carries it.

2. **`decompressUnknownSize` doesn't work for any codec, including SNAPPY.** The drain loop ends on `read() == -1`, but `NonBlockedDecompressorStream` throws `IOException("Corrupt file: Zero bytes read during decompression.")` when its block is consumed rather than returning `-1`. It was never caught because CI dies at compile. Replaced with grow-and-retry into a dynamically sized buffer, which the spec explicitly allows; that covers LZ4_RAW too.

3. **`parquet-hadoop` has two test compile errors that block the module** independent of anything I changed (they fail on `be266d31`): `TestParquetMetadataConverter` calls `assertEquals`/`assertTrue` with no JUnit import, and `TestSelfReferenceFileWrite` is missing the `ParquetReadOptions` import. Fixing those is what let me actually run things — currently 679/679 in `parquet-column` and 761/762 in `parquet-hadoop`.

One thing I deliberately did **not** change: I'd tightened schema validation to require `inline` whenever `offset` is declared, on the reasoning that `uri` is optional per value so a `uri`+`offset`+`size` schema can still emit a self-reference with no `inline` column chunk to inherit from. But your tests assert that schema is valid, so I reverted it — your call which way it should go.

Also worth noting for sequencing: this can't go green until a parquet-format release carries `FileType` (2.13.0's `LogicalType` union stops at 18 — that's the current CI failure), and #603 is still open, so the AAD detail could still move. To build locally I pinned `parquet.format.version` to an artifact holding parquet-format master's `parquet.thrift`.

Happy to fold any of it in differently, or split it up, if that's easier for you.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestFileValueWriter.java
```diff
@@ -0,0 +1,274 @@
...
+    writer.endBlock();
+    writer.end(new java.util.HashMap<>());
+
+    // The out-of-line payload resolves back to the original bytes.
+    InputFile inputFile = HadoopInputFile.fromPath(path, conf);
+    try (ParquetFileReader reader =
+        ParquetFileReader.open(inputFile, ParquetReadOptions.builder().build())) {
+      BlockMetaData block = reader.getFooter().getBlocks().get(0);
+      ColumnChunkMetaData inlineMeta = findColumn(block, INLINE_COLUMN);
+      BytesInput resolved = reader.resolveSelfReference(inlineMeta, outOfLine.getOffset(), outOfLine.getSize());
```
emkornfield: this isn't really showing a round trip and we are using the value from memory, not the one persisted.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestFileValueWriter.java
```diff
@@ -0,0 +1,274 @@
...
+    writer.endBlock();
+    writer.end(new java.util.HashMap<>());
+
+    InputFile inputFile = HadoopInputFile.fromPath(path, conf);
+    try (ParquetFileReader reader =
+        ParquetFileReader.open(inputFile, ParquetReadOptions.builder().build())) {
+      BlockMetaData block = reader.getFooter().getBlocks().get(0);
+      ColumnChunkMetaData inlineMeta = findColumn(block, INLINE_COLUMN);
+      for (int i = 0; i < payloads.length; i++) {
+        FileValueWriter.Placement placement = placements.get(i);
```
emkornfield: again, this seems to be going through an in-memory structure.

emkornfield (changes requested):
I think the read side of self-reference vs inline needs work here, it seems that we aren't actually doing a full round trip but relying on in-memory structures to retrieve the bytes.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -609,6 +609,50 @@ public InternalFileEncryptor getEncryptor() {
...
+   * @param compressor the compressor for the {@code inline} column chunk's codec
+   * @param pageBlockEncryptor the data-module encryptor of the {@code inline} column chunk, or
+   *     {@code null} if the column chunk is not encrypted
+   * @param columnOrdinal the ordinal of the {@code inline} column the self-reference inherits from
+   * @return the offset and size of the stored representation
+   * @throws IOException if writing or compression fails
+   */
+  public SelfReferenceStorage.StoredRange writeSelfReference(
+      BytesInput resolvedBytes,
+      CodecFactory.BytesCompressor compressor,
```
emkornfield: this seems like an awkward API, shold compressor and pageBlockEncryptor be abstracted in some way (so all that is needed is column Ordinal)? or should all three be grouped together?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/FileValueWriter.java
```diff
@@ -0,0 +1,197 @@
...
+   *     schema must declare {@code inline}: it is the reference point whose compression and
+   *     encryption a self-reference inherits, so a schema without it can only store payloads inline
+   *     or as external references. Note that the schema builder does not require {@code inline} for
+   *     groups that declare {@code uri}, so an external-reference schema may reach here; pair such a
+   *     schema with a threshold of {@link Integer#MAX_VALUE} so nothing is stored out of line.
+   * @param selfReferenceThreshold payloads of at most this many bytes are stored inline; larger ones
+   *     become self-references. See
+   *     {@code ParquetProperties.Builder#withFileSelfReferenceThreshold(int)}.
+   */
+  public FileValueWriter(
```
emkornfield: Seems like this API might be error prone should it be package private and instead users rely on a static factory on ParquetFileWriter?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/FileValueWriter.java
```diff
@@ -0,0 +1,197 @@
...
+   * file body immediately as self-references.
+   *
+   * <p>Must be called while a block is open on the underlying writer, and before that block's column
+   * chunks are flushed.
+   *
+   * @param payload the resolved (logical) bytes of the value
+   * @return the placement describing which fields to write
+   * @throws IOException if writing the self-reference payload fails
+   */
+  public Placement write(Binary payload) throws IOException {
```
emkornfield: Why are we leaving this to only inline? should we have another API for external references to keep these consistent?

emkornfield: Also, it seems like this should also take in media type?

PR #3610, opened 2026-06-11
[GH-3609]: Add new sort order for int96 timestamps

### LINKED ISSUE #3609: Add new sort order for int96 timestamps

Implement the proposal in https://github.com/apache/parquet-format/pull/584 to add a new `INT96_TIMESTAMP_ORDER`

### Component(s)

Core, Thrift

divjotarora:
### Rationale for this change
See parquet-format proposal for rationale: https://github.com/apache/parquet-format/pull/584

Add new `INT96_TIMESTAMP_ORDER` sort order and related parsing.

### Are these changes tested?
Yes, new unit tests.

Two new flags and a Thrift change.

Closes #3609

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/ColumnOrder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/ColumnOrder.java
@@ -41,11 +41,17 @@ public class ColumnOrder {
      * The column order is defined by the IEEE 754 standard.
      */
     IEEE_754_TOTAL_ORDER,
+    /**
+     * Chronological order for INT96 timestamps.
+     */
+    INT96_TIMESTAMP_ORDER
   }
 
   private static final ColumnOrder UNDEFINED_COLUMN_ORDER = new ColumnOrder(ColumnOrderName.UNDEFINED);
   private static final ColumnOrder TYPE_DEFINED_COLUMN_ORDER = new ColumnOrder(ColumnOrderName.TYPE_DEFINED_ORDER);
   private static final ColumnOrder IEEE_754_TOTAL_ORDER = new ColumnOrder(ColumnOrderName.IEEE_754_TOTAL_ORDER);
+  private static final ColumnOrder INT96_TIMESTAMP_COLUMN_ORDER =
+      new ColumnOrder(ColumnOrderName.INT96_TIMESTAMP_ORDER);
 
   /**
    * @return a {@link ColumnOrder} instance representing an undefined order
@@ -69,6 +75,14 @@ public class ColumnOrder {
    */
   public static ColumnOrder ieee754TotalOrder() {
     return IEEE_754_TOTAL_ORDER;
+  }
+
+  /**
+   * @return a {@link ColumnOrder} instance representing the chronological order of INT96 timestamps
+   * @see ColumnOrderName#INT96_TIMESTAMP_ORDER
+   */
+  public static ColumnOrder int96TimestampOrder() {
+    return INT96_TIMESTAMP_COLUMN_ORDER;
   }
 
   private final ColumnOrderName columnOrderName;
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
@@ -20,6 +20,7 @@ package org.apache.parquet.schema;
 
 import java.io.Serializable;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Comparator;
 import org.apache.parquet.io.api.Binary;
 
@@ -354,4 +355,49 @@ public abstract class PrimitiveComparator<T> implements Comparator<T>, Serializa
           return "BINARY_AS_FLOAT16_IEEE_754_TOTAL_ORDER_COMPARATOR";
         }
       };
+
+  /**
+   * Comparator for two timestamps encoded as INT96 (12-byte little-endian) binary.
+   * Layout: first 8 bytes = nanoseconds within the day, last 4 bytes = Julian day.
+   *
+   * Two-level comparison, matching the INT96 timestamp sort order:
+   * 1. Compare the last 4 bytes (Julian day) as a signed little-endian int32.
+   * 2. If equal, compare the first 8 bytes (nanos) as a signed little-endian int64.
+   */
+  static final PrimitiveComparator<Binary> BINARY_AS_INT96_TIMESTAMP_COMPARATOR = new BinaryComparator() {
+    private static final long NANOSECONDS_PER_DAY = 86_400_000_000_000L;
+
+    @Override
+    int compareBinary(Binary b1, Binary b2) {
+      if (b1.length() != 12 || b2.length() != 12) {
+        throw new IllegalArgumentException(
+            "INT96 binary length must be 12, got " + b1.length() + " and " + b2.length());
+      }
+
+      ByteBuffer bb1 = b1.toByteBuffer().slice();
+      ByteBuffer bb2 = b2.toByteBuffer().slice();
+      bb1.order(ByteOrder.LITTLE_ENDIAN);
+      bb2.order(ByteOrder.LITTLE_ENDIAN);
+
+      int result = Integer.compare(bb1.getInt(8), bb2.getInt(8));
+      if (result != 0) return result;
+
+      long nanos1 = bb1.getLong(0);
+      long nanos2 = bb2.getLong(0);
+      if (nanos1 < 0 || nanos1 > NANOSECONDS_PER_DAY) {
+        throw new IllegalArgumentException(
+            "Invalid nanos value (must be positive and less than 1 day): " + nanos1);
+      }
+      if (nanos2 < 0 || nanos2 > NANOSECONDS_PER_DAY) {
+        throw new IllegalArgumentException(
+            "Invalid nanos value (must be positive and less than 1 day): " + nanos2);
+      }
+      return Long.compare(nanos1, nanos2);
+    }
+
+    @Override
+    public String toString() {
+      return "BINARY_AS_INT96_TIMESTAMP_COMPARATOR";
+    }
+  };
 }
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
@@ -385,7 +385,9 @@ public final class PrimitiveType extends Type {
 
       @Override
       PrimitiveComparator<?> comparator(LogicalTypeAnnotation logicalType, ColumnOrder columnOrder) {
-        return PrimitiveComparator.BINARY_AS_SIGNED_INTEGER_COMPARATOR;
+        return columnOrder != null && columnOrder.getColumnOrderName() == ColumnOrderName.INT96_TIMESTAMP_ORDER
+            ? PrimitiveComparator.BINARY_AS_INT96_TIMESTAMP_COMPARATOR
+            : PrimitiveComparator.BINARY_AS_SIGNED_INTEGER_COMPARATOR;
       }
     },
     FIXED_LEN_BYTE_ARRAY("getBinary", Binary.class) {
@@ -578,9 +580,14 @@ public final class PrimitiveType extends Type {
     this.decimalMeta = decimalMeta;
 
     if (columnOrder == null) {
-      columnOrder = primitive == PrimitiveTypeName.INT96 || originalType == OriginalType.INTERVAL
-          ? ColumnOrder.undefined()
-          : ColumnOrder.typeDefined();
+      if (primitive == PrimitiveTypeName.INT96) {
+        // INT96 is only used for deprecated timestamps and supports no other semantics.
+        columnOrder = ColumnOrder.int96TimestampOrder();
+      } else if (originalType == OriginalType.INTERVAL) {
+        columnOrder = ColumnOrder.undefined();
+      } else {
+        columnOrder = ColumnOrder.typeDefined();
+      }
     } else if (columnOrder.getColumnOrderName() == ColumnOrderName.IEEE_754_TOTAL_ORDER) {
       Preconditions.checkArgument(
           primitive == PrimitiveTypeName.FLOAT || primitive == PrimitiveTypeName.DOUBLE,
@@ -629,10 +636,16 @@ public final class PrimitiveType extends Type {
     }
 
     if (columnOrder == null) {
-      columnOrder = primitive == PrimitiveTypeName.INT96
-              || logicalTypeAnnotation instanceof LogicalTypeAnnotation.IntervalLogicalTypeAnnotation
-          ? ColumnOrder.undefined()
-          : ColumnOrder.typeDefined();
+      if (primitive == PrimitiveTypeName.INT96) {
+        // A plain INT96 is the legacy timestamp encoding; default it to the chronological order.
+        // An annotated INT96 carries other semantics, so leave its order undefined.
+        columnOrder =
+            logicalTypeAnnotation == null ? ColumnOrder.int96TimestampOrder() : ColumnOrder.undefined();
+      } else if (logicalTypeAnnotation instanceof LogicalTypeAnnotation.IntervalLogicalTypeAnnotation) {
+        columnOrder = ColumnOrder.undefined();
+      } else {
+        columnOrder = ColumnOrder.typeDefined();
+      }
     } else if (columnOrder.getColumnOrderName() == ColumnOrderName.IEEE_754_TOTAL_ORDER) {
       Preconditions.checkArgument(
           primitive == PrimitiveTypeName.FLOAT
@@ -651,8 +664,14 @@ public final class PrimitiveType extends Type {
   private ColumnOrder requireValidColumnOrder(ColumnOrder columnOrder) {
     if (primitive == PrimitiveTypeName.INT96) {
       Preconditions.checkArgument(
-          columnOrder.getColumnOrderName() == ColumnOrderName.UNDEFINED,
+          columnOrder.getColumnOrderName() == ColumnOrderName.UNDEFINED
+              || columnOrder.getColumnOrderName() == ColumnOrderName.INT96_TIMESTAMP_ORDER,
           "The column order %s is not supported by INT96",
+          columnOrder);
+    } else {
+      Preconditions.checkArgument(
+          columnOrder.getColumnOrderName() != ColumnOrderName.INT96_TIMESTAMP_ORDER,
+          "The column order %s is only supported by INT96",
           columnOrder);
     }
     if (getLogicalTypeAnnotation() != null) {
--- a/parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
+++ b/parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
@@ -88,7 +88,18 @@ public class TestBinaryTruncator {
     testTruncator(
         Types.required(FIXED_LEN_BYTE_ARRAY).length(12).as(INTERVAL).named("test_fixed_interval"), false);
     testTruncator(Types.required(BINARY).as(DECIMAL).precision(10).scale(2).named("test_binary_decimal"), false);
-    testTruncator(Types.required(INT96).named("test_int96"), false);
+  }
+
+  @Test
+  public void testInt96() {
+    // INT96 has a fixed 12-byte width and a chronological comparator (so it is excluded from the
+    // variable-length checks above, like FLOAT16). Its truncator is a no-op: verify it returns the
+    // value unchanged regardless of the requested length.
+    BinaryTruncator int96Truncator =
+        BinaryTruncator.getTruncator(Types.required(INT96).named("test_int96"));
+    Binary int96Value = Binary.fromConstantByteArray(new byte[] {0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4});
+    assertThat(int96Truncator.truncateMin(int96Value, 4)).isSameAs(int96Value);
+    assertThat(int96Truncator.truncateMax(int96Value, 4)).isSameAs(int96Value);
   }
 
   @Test
--- a/parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
+++ b/parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
@@ -40,7 +40,6 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
-import org.apache.parquet.TestUtils;
 import org.apache.parquet.example.data.simple.NanoTime;
 import org.apache.parquet.io.api.Binary;
 import org.junit.jupiter.api.Test;
@@ -391,10 +390,9 @@ public class TestPrimitiveComparator {
     Binary valid = int96(0, 0);
     for (long invalidNanos : new long[] {-1L, Long.MIN_VALUE, 86_400_000_000_001L, Long.MAX_VALUE}) {
       Binary invalid = int96(0, invalidNanos);
-      TestUtils.assertThrows(
-          "Expected IllegalArgumentException for nanos=" + invalidNanos,
-          IllegalArgumentException.class,
-          () -> BINARY_AS_INT96_TIMESTAMP_COMPARATOR.compare(valid, invalid));
+      assertThatThrownBy(() -> BINARY_AS_INT96_TIMESTAMP_COMPARATOR.compare(valid, invalid))
+          .as("Expected IllegalArgumentException for nanos=" + invalidNanos)
+          .isInstanceOf(IllegalArgumentException.class);
     }
   }
 
--- a/parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
+++ b/parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
@@ -20,6 +20,7 @@ package org.apache.parquet.schema;
 
 import static org.apache.parquet.schema.PrimitiveComparator.BINARY_AS_FLOAT16_COMPARATOR;
 import static org.apache.parquet.schema.PrimitiveComparator.BINARY_AS_FLOAT16_IEEE_754_TOTAL_ORDER_COMPARATOR;
+import static org.apache.parquet.schema.PrimitiveComparator.BINARY_AS_INT96_TIMESTAMP_COMPARATOR;
 import static org.apache.parquet.schema.PrimitiveComparator.BINARY_AS_SIGNED_INTEGER_COMPARATOR;
 import static org.apache.parquet.schema.PrimitiveComparator.BOOLEAN_COMPARATOR;
 import static org.apache.parquet.schema.PrimitiveComparator.DOUBLE_COMPARATOR;
@@ -39,6 +40,8 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import org.apache.parquet.TestUtils;
+import org.apache.parquet.example.data.simple.NanoTime;
 import org.apache.parquet.io.api.Binary;
 import org.junit.jupiter.api.Test;
 
@@ -350,6 +353,48 @@ public class TestPrimitiveComparator {
             .usingComparator(BINARY_AS_SIGNED_INTEGER_COMPARATOR)
             .isEqualTo(v2);
       }
+    }
+  }
+
+  private static Binary int96(int julianDay, long nanosOfDay) {
+    return new NanoTime(julianDay, nanosOfDay).toBinary();
+  }
+
+  @Test
+  public void testInt96TimestampComparator() {
+    Binary[] valuesInAscendingOrder = {
+      int96(Integer.MIN_VALUE, 0), // most negative julian day
+      int96(-1, 86_399_999_999_999L), // negative julian days sort before day 0
+      int96(0, 0), // start of the julian period
+      int96(0, 86_399_999_999_999L), // same day, later time of day
+      int96(2440000, 123L), // 1968-05-23T00:00:00.000000123, pre-epoch but positive julian day
+      int96(2458850, 43_200_000_000_000L), // 2020-01-01T12:00:00
+      int96(2458881, 39_600_000_000_000L), // 2020-02-01T11:00:00, later day even though earlier time of day
+      int96(2458881, 39_600_000_000_001L), // 2020-02-01T11:00:00.000000001, nanos tie-break
+      int96(Integer.MAX_VALUE, 86_399_999_999_999L)
+    };
+
+    for (int i = 0; i < valuesInAscendingOrder.length; ++i) {
+      for (int j = 0; j < valuesInAscendingOrder.length; ++j) {
+        assertThat(Integer.signum(BINARY_AS_INT96_TIMESTAMP_COMPARATOR.compare(
+                valuesInAscendingOrder[i], valuesInAscendingOrder[j])))
+            .as("comparing value " + i + " to value " + j)
+            .isEqualTo(Integer.signum(Integer.compare(i, j)));
+      }
+    }
+  }
+
+  @Test
+  public void testInt96TimestampComparatorRejectsInvalidNanos() {
+    // Same Julian day so the comparator reaches the nanos validation instead of
+    // returning early on the day comparison.
+    Binary valid = int96(0, 0);
+    for (long invalidNanos : new long[] {-1L, Long.MIN_VALUE, 86_400_000_000_001L, Long.MAX_VALUE}) {
+      Binary invalid = int96(0, invalidNanos);
+      TestUtils.assertThrows(
+          "Expected IllegalArgumentException for nanos=" + invalidNanos,
+          IllegalArgumentException.class,
+          () -> BINARY_AS_INT96_TIMESTAMP_COMPARATOR.compare(valid, invalid));
     }
   }
 
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -88,6 +88,7 @@ import org.apache.parquet.format.GeographyType;
 import org.apache.parquet.format.GeometryType;
 import org.apache.parquet.format.GeospatialStatistics;
 import org.apache.parquet.format.IEEE754TotalOrder;
+import org.apache.parquet.format.Int96TimestampOrder;
 import org.apache.parquet.format.IntType;
 import org.apache.parquet.format.KeyValue;
 import org.apache.parquet.format.LogicalType;
@@ -146,6 +147,7 @@ public class ParquetMetadataConverter {
 
   private static final TypeDefinedOrder TYPE_DEFINED_ORDER = new TypeDefinedOrder();
   private static final IEEE754TotalOrder IEEE_754_TOTAL_ORDER = new IEEE754TotalOrder();
+  private static final Int96TimestampOrder INT96_TIMESTAMP_ORDER = new Int96TimestampOrder();
   public static final MetadataFilter NO_FILTER = new NoFilter();
   public static final MetadataFilter SKIP_ROW_GROUPS = new SkipMetadataFilter();
   public static final long MAX_STATS_SIZE = 4096; // limit stats to 4k
@@ -289,6 +291,9 @@ public class ParquetMetadataConverter {
           break;
         case IEEE_754_TOTAL_ORDER:
           columnOrder.setIEEE_754_TOTAL_ORDER(IEEE_754_TOTAL_ORDER);
+          break;
+        case INT96_TIMESTAMP_ORDER:
+          columnOrder.setINT96_TIMESTAMP_ORDER(INT96_TIMESTAMP_ORDER);
           break;
         case UNDEFINED:
           // Use TypeDefinedOrder if some types (e.g. INT96) have undefined column orders.
@@ -911,8 +916,16 @@ public class ParquetMetadataConverter {
   }
 
   private static boolean isMinMaxStatsSupported(PrimitiveType type) {
-    return type.columnOrder().getColumnOrderName() == ColumnOrderName.TYPE_DEFINED_ORDER
-        || type.columnOrder().getColumnOrderName() == ColumnOrderName.IEEE_754_TOTAL_ORDER;
+    switch (type.columnOrder().getColumnOrderName()) {
+      case TYPE_DEFINED_ORDER:
+      case IEEE_754_TOTAL_ORDER:
+      case INT96_TIMESTAMP_ORDER:
+        return true;
+      case UNDEFINED:
+        return false;
+      default:
+        throw new IllegalArgumentException("Unknown column order: " + type.columnOrder());
+    }
   }
 
   /**
@@ -2062,7 +2075,17 @@ public class ParquetMetadataConverter {
                   || schemaElement.converted_type == ConvertedType.INTERVAL)) {
             columnOrder = org.apache.parquet.schema.ColumnOrder.undefined();
           }
+          // INT96_TIMESTAMP_ORDER is only valid for INT96 columns, ignore it anywhere else.
+          if (columnOrder.getColumnOrderName() == ColumnOrderName.INT96_TIMESTAMP_ORDER
+              && schemaElement.type != Type.INT96) {
+            columnOrder = org.apache.parquet.schema.ColumnOrder.undefined();
+          }
           primitiveBuilder.columnOrder(columnOrder);
+        } else if (schemaElement.type == Type.INT96) {
+          // A footer without column orders predates INT96_TIMESTAMP_ORDER, so an INT96 column here
+          // must not inherit the (chronological) construction-time default: its stats, if any, were
+          // written under the legacy order and must be ignored.
+          primitiveBuilder.columnOrder(org.apache.parquet.schema.ColumnOrder.undefined());
         }
         childBuilder = primitiveBuilder;
       } else {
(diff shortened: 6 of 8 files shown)


parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
```diff
@@ -370,6 +373,10 @@ public boolean getSizeStatisticsEnabled(ColumnDescriptor column) {
     return sizeStatisticsEnabled;
   }
 
+  public boolean getInt96TimestampStatisticsEnabled() {
```
rdblue: Why make this an option? Since we don't use int96 stats otherwise, I think it would be perfectly fine to keep it simple and just produce the new stats all the time.

parquet-column/src/main/java/org/apache/parquet/schema/ColumnOrder.java
```diff
@@ -36,11 +36,18 @@ public enum ColumnOrderName {
     /**
      * Type defined order meaning that the comparison order of the elements are based on its type.
      */
-    TYPE_DEFINED_ORDER
+    TYPE_DEFINED_ORDER,
+    /**
+     * Chronological order for INT96 timestamps: values are compared by the Julian day (the last 4
+     * bytes, as a little-endian signed int32), then by the nanoseconds within the day (the first 8
+     * bytes, as a little-endian signed int64). Only supported for the INT96 physical type.
```
rdblue: I don't think that this Javadoc is the right place for documenting the format and how to compare. This is part of the spec so the spec needs to be clear and this needs to state what the order means.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -293,4 +294,36 @@ public String toString() {
       return "BINARY_AS_FLOAT16_COMPARATOR";
     }
   };
+
+  /*
```
rdblue: Did you intend for this to be Javadoc?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -655,6 +661,15 @@ public PrimitiveType withLogicalTypeAnnotation(LogicalTypeAnnotation logicalType
     return new PrimitiveType(getRepetition(), primitive, length, getName(), logicalType, getId());
   }
 
+  /**
+   * @param columnOrder the column order
+   * @return a new PrimitiveType with the same fields and the given column order
+   */
+  public PrimitiveType withColumnOrder(ColumnOrder columnOrder) {
```
rdblue: If we always produce INT96 stats using the timestamp order, would we need this addition to the API? I think we could always produce a `PrimitiveType` with the right order when constructing types. We would just need to make sure that deserialization correctly distinguishes between unordered and timestamp order.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2086,14 +2101,22 @@ Repetition fromParquetRepetition(FieldRepetitionType repetition) {
...
+    if (columnOrder.isSetINT96_TIMESTAMP_ORDER() && readInt96TimestampStatisticsEnabled()) {
+      return org.apache.parquet.schema.ColumnOrder.int96TimestampOrder();
+    }
     // The column order is not yet supported by this API
     return org.apache.parquet.schema.ColumnOrder.undefined();
   }
 
+  private boolean readInt96TimestampStatisticsEnabled() {
+    return options == null
+        || options.isEnabled(ParquetInputFormat.INT96_TIMESTAMP_STATISTICS_READING_ENABLED, true);
```
rdblue: Why would we have an option for reading INT96 stats? Wouldn't this always be true if we know the int96 stats are using the timestamp order?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java
```diff
@@ -89,7 +89,7 @@ public InternalParquetRecordWriter(
       ParquetProperties props) {
     this.parquetFileWriter = parquetFileWriter;
     this.writeSupport = Objects.requireNonNull(writeSupport, "writeSupport cannot be null");
-    this.schema = schema;
+    this.schema = ParquetFileWriter.applyInt96TimestampOrder(schema, props);
```
rdblue: I don't think this is the right place to fixup order. It would be better to always use the new order for INT96 when constructing schemas. When converting from file schemas, we would need to detect timestamp vs unordered, but anything going through the write path should automatically use timestamp order because the write path should always produce it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
```diff
@@ -447,6 +451,35 @@ public ParquetFileWriter(
         props.getAllocator());
   }
 
+  /**
+   * Returns the schema with INT96 columns tagged with the INT96_TIMESTAMP_ORDER column order if
+   * INT96 timestamp statistics are enabled, so that statistics are accumulated with the
+   * chronological comparator and the proper column order is written to the footer.
+   */
+  static MessageType applyInt96TimestampOrder(MessageType schema, ParquetProperties props) {
```
rdblue: See my comments above about this, but I'm skeptical that we should fixup the schema.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputFormat.java
```diff
@@ -122,6 +122,12 @@ public class ParquetInputFormat<T> extends FileInputFormat<Void, T> {
    */
   public static final String STATS_FILTERING_ENABLED = "parquet.filter.stats.enabled";
 
+  /**
+   * key to configure whether INT96 min/max statistics written with the INT96_TIMESTAMP_ORDER
+   * column order are read (enabled by default)
+   */
+  public static final String INT96_TIMESTAMP_STATISTICS_READING_ENABLED = "parquet.int96.timestamp.statistics.read.enabled";
```
rdblue: I don't think this should be an option.

parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java
divjotarora: Removed the flag

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
divjotarora: I was copying the formatting from [BINARY_AS_SIGNED_INTEGER_COMPARATOR](https://github.com/apache/parquet-java/blob/65884b558785da1c05f9398a83a12390e3b2d7f5/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java#L209). Is this not the right way?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
divjotarora: Removed the writer-side flag and all of this schema-fixing code.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
divjotarora: Removed the flag

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
divjotarora: Yep, all of this was removed along with removing the writer-side flag because we can just construct the `PrimitiveType` with the order now

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
rdblue: [Javadoc](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format) starts with `/**` so I don't think this would produce Javadoc and is instead a regular comment.

divjotarora: Changed to Javadoc

wgtmac:
Please resolve the conflicts.

divjotarora:
> Please resolve the conflicts.

@wgtmac Sorry for the delay, I was on holiday last week. Conflicts are resolved now

parquet-column/src/main/java/org/apache/parquet/schema/ColumnOrder.java
rdblue: Sorry if my comment wasn't very clear. I can see how it would be misread. I think this should state "Chronological order of INT96 timestamps" and not cover serialization.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -354,4 +355,36 @@ public String toString() {
...
         }
       };
+
+  /**
+   * Comparator for two timestamps encoded as INT96 (12-byte little-endian) binary.
+   * Layout: first 8 bytes = nanoseconds within the day, last 4 bytes = Julian day.
+   *
+   * Two-level comparison, matching the INT96 timestamp sort order:
+   * 1. Compare the last 4 bytes (Julian day) as a signed little-endian int32.
+   * 2. If equal, compare the first 8 bytes (nanos) as a signed little-endian int64.
```
rdblue, line 365: The correctness of this depends on the two fields so I think we have to be more careful since the caller may pass in something unexpected.

I think that we need to validate that the nanosecond value is positive and less than the number of nanoseconds per day. If either of these is violated, then this sort order is no longer correct.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -578,9 +580,15 @@ public PrimitiveType(
     this.decimalMeta = decimalMeta;
 
     if (columnOrder == null) {
-      columnOrder = primitive == PrimitiveTypeName.INT96 || originalType == OriginalType.INTERVAL
-          ? ColumnOrder.undefined()
-          : ColumnOrder.typeDefined();
+      if (primitive == PrimitiveTypeName.INT96) {
+        // A plain INT96 is the legacy timestamp encoding; default it to the chronological order.
+        // An annotated INT96 carries other semantics, so leave its order undefined.
+        columnOrder = originalType == null ? ColumnOrder.int96TimestampOrder() : ColumnOrder.undefined();
```
rdblue: I don't think that there are any original types that can be applied to INT96. We stopped adding them years ago. I think it would be safe to just return `ColumnOrder.int96TimestampOrder()` without the check. The comment would be "INT96 is only used for (deprecated) timestamps".

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -93,7 +93,16 @@ public void testContractNonStringTypes() {
     testTruncator(
         Types.required(FIXED_LEN_BYTE_ARRAY).length(12).as(INTERVAL).named("test_fixed_interval"), false);
     testTruncator(Types.required(BINARY).as(DECIMAL).precision(10).scale(2).named("test_binary_decimal"), false);
-    testTruncator(Types.required(INT96).named("test_int96"), false);
+
+    // INT96 has a fixed 12-byte width and a chronological comparator (so it is excluded from the
+    // variable-length checks above, like FLOAT16). Its truncator is a no-op: verify it returns the
+    // value unchanged regardless of the requested length.
+    BinaryTruncator int96Truncator = BinaryTruncator.getTruncator(
```
rdblue: This is replacing the INT96 handling above, but this expansion deserves its own new test case because it is not running the standard `testTrunctator` logic. This should be in a new `testInt96`.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
```diff
@@ -354,6 +359,60 @@ public void testBinaryAsSignedIntegerComparatorWithEquals() {
...
+      timestampToInt96("2020-02-01T11:00:00"), // later day even though earlier time of day
+      timestampToInt96("2020-02-01T11:00:00.000000001"), // nanos tie-break
+      int96(Integer.MAX_VALUE, 86_399_999_999_999L)
+    };
+
+    // The same value in different Binary representations must compare identically; the offset
+    // variant guards against absolute reads not being relative to the value's start
+    List<Function<Binary, Binary>> representations = List.of(
+        b -> b,
+        b -> Binary.fromReusedByteArray(b.getBytes()),
```
rdblue: There's no functional difference between reused and constant here. It is just a way to inform Parquet that the value may be modified (and should be copied) or not.

I also don't think there is much value in these cases more generally. This is testing that `Binary#toByteBuffer` does the right thing and that is not the responsibility of this test to validate.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
```diff
@@ -354,6 +359,60 @@ public void testBinaryAsSignedIntegerComparatorWithEquals() {
     }
   }
 
+  private static Binary int96(int julianDay, long nanosOfDay) {
+    return new NanoTime(julianDay, nanosOfDay).toBinary();
+  }
+
+  private static Binary timestampToInt96(String timestamp) {
```
rdblue: I don't see the value of complicating the tests with string parsing. I think you can test what you need to with Julian day and nanos. No need to make your tests depend on correct conversion from epoch to Julian.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
```diff
@@ -354,6 +359,60 @@ public void testBinaryAsSignedIntegerComparatorWithEquals() {
...
+    LocalDateTime dt = LocalDateTime.parse(timestamp);
+    int julianDay = (int) (dt.toLocalDate().toEpochDay() + 2440588);
+    return new NanoTime(julianDay, dt.toLocalTime().toNanoOfDay()).toBinary();
+  }
+
+  @Test
+  public void testInt96TimestampComparator() {
```
rdblue, line 363: This also needs to cover cases like negative nanos and nanos >= 86_400_000_000_000. Use a separate test case if the intent is to fail.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -911,8 +916,10 @@ private static byte[] tuncateMax(BinaryTruncator truncator, int truncateLength,
   }
 
   private static boolean isMinMaxStatsSupported(PrimitiveType type) {
-    return type.columnOrder().getColumnOrderName() == ColumnOrderName.TYPE_DEFINED_ORDER
-        || type.columnOrder().getColumnOrderName() == ColumnOrderName.IEEE_754_TOTAL_ORDER;
+    ColumnOrderName name = type.columnOrder().getColumnOrderName();
+    return name == ColumnOrderName.TYPE_DEFINED_ORDER
+        || name == ColumnOrderName.IEEE_754_TOTAL_ORDER
+        || name == ColumnOrderName.INT96_TIMESTAMP_ORDER;
```
rdblue: This is getting too complicated. Either use a switch or use `name != UNDEFINED`.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1171,17 +1171,14 @@ public void testMissingValuesFromStats() {
 
   @Test
   public void testSkippedV2Stats() {
+    // INTERVAL has an undefined column order, so its stats are skipped.
```
rdblue: Please remove this. It is unrelated and doesn't need to be part of this PR.

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1171,17 +1171,14 @@ public void testMissingValuesFromStats() {
...
         Types.optional(PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY)
             .length(12)
             .as(OriginalType.INTERVAL)
             .named(""),
         new BigInteger("12345678"),
         new BigInteger("12345679"));
-    testSkippedV2Stats(
-        Types.optional(PrimitiveTypeName.INT96).named(""),
-        new BigInteger("-75687987"),
-        new BigInteger("45367657"));
```
rdblue, line 1282: There are other tests that also need to be updated. For example, `testV2StatsEqualMinMax` tests cases where the min and max are equal and stats are preserved. This also uses `BigInteger` to pass the values, which is suspicious.

I think this should update all of the INT96 cases to use `NanoTime`. Looking at the code path, I don't think that NanoTime is plumbed to work since I don't see a `Statistics` implementation for it. Maybe this should actually pass a Binary produced by NanoTime instead.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -2057,7 +2064,17 @@ private void buildChildren(
...
+          if (columnOrder.getColumnOrderName() == ColumnOrderName.INT96_TIMESTAMP_ORDER
+              && schemaElement.type != Type.INT96) {
+            columnOrder = org.apache.parquet.schema.ColumnOrder.undefined();
+          }
           primitiveBuilder.columnOrder(columnOrder);
+        } else if (schemaElement.type == Type.INT96) {
+          // A footer without column orders predates INT96_TIMESTAMP_ORDER, so an INT96 column here
+          // must not inherit the (chronological) construction-time default: its stats, if any, were
+          // written under the legacy order and must be ignored.
+          primitiveBuilder.columnOrder(org.apache.parquet.schema.ColumnOrder.undefined());
```
rdblue, line 2093: There isn't a test in `TestParquetMetadataConverter` that validates Parquet metadata for an INT96 without a column order. There is a test that a new schema will produce a column order, but not one to verify that a schema without one will be read correctly.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+  private static final Binary NEXT_DAY = int96(2440589, 0L); // 1970-01-02 00:00:00
+
+  private static final List<Binary> VALUES = List.of(LATE_IN_DAY, NEXT_DAY, EARLY, SAME_DAY_EARLY);
+  private static final Binary EXPECTED_MIN = EARLY;
+  private static final Binary EXPECTED_MAX = NEXT_DAY;
+
+  @Rule
+  public TemporaryFolder tmp = new TemporaryFolder();
+
+  private static Binary int96(int julianDay, long nanosOfDay) {
```
rdblue: Please don't add new serializations. Use `NanoTime#toBinary()` and `Binary#toByteBuffer()`.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+    } finally {
+      writer.close();
+    }
+    return file;
+  }
+
+  private static ParquetMetadata readFooter(File file, Configuration conf) throws IOException {
+    Path path = new Path(file.getAbsolutePath());
+    try (ParquetFileReader reader = ParquetFileReader.open(
+        HadoopInputFile.fromPath(path, conf),
```
rdblue: I recommend using the InputFile and OutputFile paths rather than having a dependency on Hadoop. You should use LocalInputFile and LocalOutputFile.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+    byte[] bytes = Files.readAllBytes(file.toPath());
+    int footerLen = ByteBuffer.wrap(bytes, bytes.length - 8, 4)
+        .order(ByteOrder.LITTLE_ENDIAN)
+        .getInt();
+    int footerStart = bytes.length - 8 - footerLen;
+    return Util.readFileMetaData(new ByteArrayInputStream(bytes, footerStart, footerLen));
+  }
+
+  /** Rewrites the footer of src into a new file, keeping all data pages byte-identical. */
+  private File rewriteFooter(File src, FileMetaData footer, String name)
```
rdblue: This method seems misnamed. It doesn't match the comment.

What about `copyWithNewFooter`?

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+  private static Binary int96(int julianDay, long nanosOfDay) {
+    return Binary.fromConstantByteArray(ByteBuffer.allocate(12)
+        .order(ByteOrder.LITTLE_ENDIAN)
+        .putLong(nanosOfDay)
+        .putInt(julianDay)
+        .array());
+  }
+
+  private File writeFile() throws IOException {
+    File file = new File(tmp.getRoot(), "int96.parquet");
```
rdblue: This should use `tmp.newFile("int96.parquet")` instead of constructing one manually.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+    return dst;
+  }
+
+  /**
+   * Retags the INT96 column orders in the footer as TYPE_ORDER while leaving the (already written)
+   * min/max statistics in place. This simulates a legacy writer that emitted INT96 stats under
+   * TYPE_ORDER before INT96_TIMESTAMP_ORDER existed; such stats must be ignored by the reader.
+   */
+  private static void downgradeInt96ToTypeOrder(FileMetaData footer) {
```
rdblue: Can you add a couple more cases? I'd like to see UNDEFINED, TYPE_DEFINED_ORDER, and unset cases.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+    // Values are written in non-chronological order; the writer must still produce the
+    // chronological min/max, not first/last or byte-wise extremes.
+    File file = writeFile();
+    FileMetaData rawFooter = readRawFooter(file);
+    // schema[0] is the message root; column_orders are indexed by leaf order: ts=0, id=1
+    assertTrue(rawFooter.getColumn_orders().get(0).isSetINT96_TIMESTAMP_ORDER());
+    assertTrue(rawFooter.getColumn_orders().get(1).isSetTYPE_ORDER());
+
+    for (RowGroup rowGroup : rawFooter.getRow_groups()) {
+      for (ColumnChunk chunk : rowGroup.getColumns()) {
```
rdblue: Use `get(0)` like above. There's no need to search.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+
+  @Test
+  public void testWriterEmitsInt96StatsAndColumnOrder() throws IOException {
+    // Values are written in non-chronological order; the writer must still produce the
+    // chronological min/max, not first/last or byte-wise extremes.
+    File file = writeFile();
+    FileMetaData rawFooter = readRawFooter(file);
+    // schema[0] is the message root; column_orders are indexed by leaf order: ts=0, id=1
+    assertTrue(rawFooter.getColumn_orders().get(0).isSetINT96_TIMESTAMP_ORDER());
+    assertTrue(rawFooter.getColumn_orders().get(1).isSetTYPE_ORDER());
```
rdblue: No need to check this.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+        if (chunk.getMeta_data().getType() == Type.INT96) {
+          Statistics stats = chunk.getMeta_data().getStatistics();
+          assertTrue(stats != null && stats.isSetMin_value());
+          assertArrayEquals(EXPECTED_MIN.getBytes(), stats.getMin_value());
+          assertArrayEquals(EXPECTED_MAX.getBytes(), stats.getMax_value());
+        }
+      }
+    }
+
+    // Column index should be present for both columns.
```
rdblue: But only validated for INT96.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
```diff
@@ -0,0 +1,291 @@
...
+        HadoopInputFile.fromPath(path, conf), HadoopReadOptions.builder(conf, path).build()
+      )
+    ) {
+      assertNotNull(reader.readColumnIndex(getColumn(reader.getFooter(), "id")));
+
+      ColumnIndex columnIndex = reader.readColumnIndex(getColumn(reader.getFooter(), "ts"));
+      assertNotNull(columnIndex);
+      assertArrayEquals(EXPECTED_MIN.getBytes(), toArray(columnIndex.getMinValues().get(0)));
+      assertArrayEquals(EXPECTED_MAX.getBytes(),
+        toArray(columnIndex.getMaxValues().get(columnIndex.getMaxValues().size() - 1)));
```
rdblue: There is only one page, right?

rdblue:
I think there are a few things to be fixed, but overall I trust that this works.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
divjotarora: Added validation for both cases

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
divjotarora: done

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
divjotarora: Added a separate `testInt96`

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
divjotarora: Removed this function

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
divjotarora: Added a separate test case

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
divjotarora: Removed, now we're just testing using `valuesInAscendingOrder`

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
divjotarora: Switched to `switch`

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Removed

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
divjotarora: Added such a test

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Fixed `testV2StatsEqualMinMax` to use `NanoTime#toBinary()` and added the `Statistics` implementation for `Binary` as well. Also added an INT96 test case to `testV2OnlyStats`

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Removed this function

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Renamed `copyWithNewFooter`

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Removed this helper and added both `testReaderIgnoresInt96StatsWithTypeDefinedOrder` and `testReaderIgnoresInt96StatsWhenColumnOrdersAbsent`

divjotarora: I wasn't able to test `UNDEFINED` as I didn't find a good way to serialize an undefined value for the union

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Removed

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Changed to `Column index should be present for the INT96 column`

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Done

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
divjotarora: Changed to `get(0)`

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -354,4 +355,47 @@ public String toString() {
...
+      bb1.order(ByteOrder.LITTLE_ENDIAN);
+      bb2.order(ByteOrder.LITTLE_ENDIAN);
+
+      int result = Integer.compare(bb1.getInt(8), bb2.getInt(8));
+      if (result != 0) return result;
+
+      long nanos1 = bb1.getLong(0);
+      long nanos2 = bb2.getLong(0);
+      if (nanos1 < 0 || nanos1 > NANOSECONDS_PER_DAY) {
+        throw new IllegalArgumentException("Invalid nanos value: " + nanos1);
```
rdblue: I like that this is clear and direct, but we usually want to also state why the value is invalid. In this case, it is beyond the number of nanos per day. I'd change this to `"Invalid nanos value (must be positive and less than 1 day): "`

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
```diff
@@ -354,6 +356,51 @@ public void testBinaryAsSignedIntegerComparatorWithEquals() {
...
+  @Test
+  public void testInt96TimestampComparatorRejectsInvalidNanos() {
+    // Same Julian day so the comparator reaches the nanos validation instead of
+    // returning early on the day comparison.
+    Binary valid = int96(0, 0);
+    for (long invalidNanos : new long[] {-1L, Long.MIN_VALUE, 86_400_000_000_001L, Long.MAX_VALUE}) {
+      Binary invalid = int96(0, invalidNanos);
+      try {
+        BINARY_AS_INT96_TIMESTAMP_COMPARATOR.compare(valid, invalid);
+        fail("Expected IllegalArgumentException for nanos=" + invalidNanos);
```
rdblue: You can use `TestUtils.assertThrows` for this. It's in parquet-common: parquet-common/src/test/java/org/apache/parquet/TestUtils.java

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -911,8 +916,16 @@ private static byte[] tuncateMax(BinaryTruncator truncator, int truncateLength,
...
-        || type.columnOrder().getColumnOrderName() == ColumnOrderName.IEEE_754_TOTAL_ORDER;
+    switch (type.columnOrder().getColumnOrderName()) {
+      case TYPE_DEFINED_ORDER:
+      case IEEE_754_TOTAL_ORDER:
+      case INT96_TIMESTAMP_ORDER:
+        return true;
+      case UNDEFINED:
+        return false;
+      default:
+        throw new IllegalArgumentException("Unknown column order: " + type.columnOrder());
```
rdblue, line 925: Nit: I think we would normal throw `UnsupportedOperationException` since this is a valid argument (it's an enum after all).

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1258,10 +1266,12 @@ public void testV2StatsEqualMinMax() {
             .named(""),
         new BigInteger("-8752832"),
         new BigInteger("-8752832"));
+    Binary int96 = new NanoTime(2458850, 43_200_000_000_000L)
+      .toBinary(); // 2020-01-01T12:00:00
```
rdblue: Why does this convert to `Binary` when the decimal test uses `BigInteger`? Should this use `NanoTime` instead?

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1281,6 +1291,8 @@ private static <T> Statistics<?> createStats(PrimitiveType type, T min, T max) {
       return createStatsTyped(type, (Long) min, (Long) max);
     } else if (c == BigInteger.class) {
       return createStatsTyped(type, (BigInteger) min, (BigInteger) max);
+    } else if (min instanceof Binary) {
+      return createStatsTyped(type, (Binary) min, (Binary) max);
```
rdblue: Should this accept `NanoTime`?

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -1413,7 +1434,29 @@ public void testColumnOrders() throws IOException {
     assertEquals(
         ColumnOrder.typeDefined(), columns.get(0).getPrimitiveType().columnOrder());
     assertEquals(ColumnOrder.undefined(), columns.get(1).getPrimitiveType().columnOrder());
-    assertEquals(ColumnOrder.undefined(), columns.get(2).getPrimitiveType().columnOrder());
+    assertEquals(ColumnOrder.int96TimestampOrder(), columns.get(2).getPrimitiveType().columnOrder());
+  }
+
+  @Test
+  public void testLegacyINT96ColumnOrder() throws IOException {
```
rdblue: Can we also check the conversion with TYPE_DEFINED_ORDER? I think you could just create the primitive type with that specific order and then round trip to ensure it is preserved.

I think this is a better test than the file ones that replace the footer, since it's easier than manually editing the footer contents.

parquet-hadoop/src/test/java/org/apache/parquet/statistics/TestInt96TimestampStatistics.java
rdblue: I think this tests the same thing as `testLegacyINT96ColumnOrder` in `TestParquetMetadataConverter`. That's probably the right place for this, so test here are good even if we can't test undefined.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
divjotarora: Prefer to leave `IllegalArgumentException` to match `getColumnOrders` above

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
divjotarora: Done

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveComparator.java
divjotarora: Done

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Sure, changed to just pass down `NanoTime` and modified `createStats`/`createStatsTyped` to accept `NanoTime` rather than `Binary`. At the bottom layer in `createStatsTyped`, we still call `toBinary()` to serialize the `NanoTime` values, but the callers are cleaner now

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Changed this and added a `createStatsTyped` override for `NanoTime`

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Sure, split this into `testUndefinedINT96ColumnOrder` and `testTypeDefinedOrderINT96ColumnOrder` to test the two cases

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
rdblue: That's fair, I guess this is matching existing convention in this file so it's fine.

rdblue (approved):

rdblue:
This looks good to me, but there are CI failures from spotless. When CI is passing, I'll commit this.

divjotarora:
> This looks good to me, but there are CI failures from spotless. When CI is passing, I'll commit this.

Fixed the spotless checks + the merge conflicts from the master branch

divjotarora:
Note: merging this is blocked because a parquet-format release containing https://github.com/apache/parquet-format/pull/584 hasn't been cut yet, so we cannot pull a parquet.thrift definition with this change and the linked CI jobs fail during compilation.

wgtmac:
What is the next step for this? Are we still waiting for the next release of parquet-format or merging https://github.com/apache/parquet-java/pull/3709? @divjotarora

divjotarora:
> What is the next step for this? Are we still waiting for the next release of parquet-format or merging #3709? @divjotarora

Yes, either one of those would unblock this PR

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -354,4 +355,49 @@ public String toString() {
...
+
+      int result = Integer.compare(bb1.getInt(8), bb2.getInt(8));
+      if (result != 0) return result;
+
+      long nanos1 = bb1.getLong(0);
+      long nanos2 = bb2.getLong(0);
+      if (nanos1 < 0 || nanos1 > NANOSECONDS_PER_DAY) {
+        throw new IllegalArgumentException(
+            "Invalid nanos value (must be positive and less than 1 day): " + nanos1);
+      }
```
Jiayi-Wang-db, line 390: I think there's a bug here, for example:
```
b1: day , NANOSECONDS_PER_DAY
b2: day + 1, 0
```
The code will return b1<b2 but they are the same.

divjotarora: Changed the code to error on `>= NANOSECONDS_PER_DAY`. The [spec](https://github.com/apache/parquet-format/blob/219e3f12a62f9476e830c21e26d030d231f7c017/src/main/thrift/parquet.thrift#L1143-L1153) describes comparing INT96 values in this way. The `arrow-rs` [code](https://github.com/apache/arrow-rs/blob/e3a21bf4083909ab2f2ae89b03cdd6b5642d24e2/parquet/src/data_type.rs#L137-L157) does the same as parquet-java, but without any checks. Erroring out seems like the best option to me.

Jiayi-Wang-db (approved):

Issue #3615, opened 2026-06-19
Introduce AssertJ for testing


nastra:
### Describe the enhancement requested

AssertJ gives us readable, fluent assertions that make tests easier to write and failures easier to analyze. With JUnit's `assertEquals(expected, actual)` it's easy to mix up argument order, and you end up with confusing failure messages. AssertJ's `assertThat(actual).isEqualTo(expected)` reads naturally and always gets the order right. Additionally, it provides a lot of flexibility when asserting things.

Compound checks that need multiple JUnit assertions collapse into a single chain:

```
// JUnit
assertNotNull(list);
assertEquals(3, list.size());
assertTrue(list.contains("foo"));
```
```
// AssertJ
assertThat(list).hasSize(3).contains("foo");
```

When something fails, AssertJ tells you exactly what went wrong and makes debugging much easier, because the failure message usually contains enough details to understand the issue. For example, an assertion on a list would fail with `expected list to contain 'foo' but was ['bar', 'baz']`. JUnit assertions on the other hand would only contain `expected true but was false`.

It also covers cases where JUnit assertions are just clunky: exception messages, collection contents, map entries, string patterns, comparisons with custom comparators. All without needing Hamcrest matchers or custom helper methods.

I'm planning to add it as a project-wide test dependency so new tests can use it immediately. Existing tests don't need to be migrated, and we can move them over incrementally where it makes sense.

PR #3617, opened 2026-06-19
GH-3615: Use assertThat checks in parquet-variant tests

nastra:
### Rationale for this change

This introduces AssertJ for testing, which enables more fluent assertion checks, easier debugging and overall easier testing. Additional details about the motivation can be found in https://github.com/apache/parquet-java/issues/3615

added the assertj library to the test scope. Also updated parquet-variant to use the new `assertThat` checks to make test code more readable

### Are these changes tested?
yes, existing tests

no

### DIFF
--- a/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArray.java
+++ b/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArray.java
@@ -18,10 +18,12 @@
  */
 package org.apache.parquet.variant;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.time.LocalDate;
-import org.junit.Assert;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -77,7 +79,7 @@ public class TestVariantArray {
     Variant value = new Variant(ByteBuffer.wrap(new byte[] {0b0011, 0x00}), VariantTestUtil.EMPTY_METADATA);
     VariantTestUtil.testVariant(value, v -> {
       VariantTestUtil.checkType(v, VariantUtil.ARRAY, Variant.Type.ARRAY);
-      Assert.assertEquals(0, v.numArrayElements());
+      assertThat(v.numArrayElements()).isEqualTo(0);
     });
   }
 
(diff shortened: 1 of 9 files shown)


parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantArray.java
```diff
@@ -171,13 +171,10 @@ public void testArrayFourByteOffset() {
...
-      // An object header
-      Variant value = new Variant(ByteBuffer.wrap(new byte[] {0b1000010}), VariantTestUtil.EMPTY_METADATA);
-      value.numArrayElements();
-      Assert.fail("Expected exception not thrown");
-    } catch (Exception e) {
-      Assert.assertEquals("Cannot read OBJECT value as ARRAY", e.getMessage());
-    }
+    // An object header
+    Variant value = new Variant(ByteBuffer.wrap(new byte[] {0b1000010}), VariantTestUtil.EMPTY_METADATA);
+    assertThatThrownBy(value::numArrayElements)
```
nastra, line 176: Previously with `Assert.fail` we would have a larger block where it wasn't clear code line of code would actually throw the exception and what error msg would be returned. We're more explicit now in terms of testing error conditions, because we check the underlying exception and the error msg.

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantParseJson.java
```diff
@@ -18,191 +18,189 @@
...
 
   @Test
   public void testParseDecimalFloat() throws IOException {
     Variant v = VariantJsonParser.parseJson("3.14");
     Variant.Type type = v.getType();
-    Assert.assertTrue(
-        "Expected decimal type, got " + type,
-        type == Variant.Type.DECIMAL4 || type == Variant.Type.DECIMAL8 || type == Variant.Type.DECIMAL16);
-    Assert.assertEquals(0, new BigDecimal("3.14").compareTo(v.getDecimal()));
+    assertThat(type).isIn(Variant.Type.DECIMAL4, Variant.Type.DECIMAL8, Variant.Type.DECIMAL16);
```
nastra, line 92: if this check ever fails, then it will me much clearer what the issue is, because previously it would only say `false is not true`. I've modified the check to show how such an error msg would look:

```
Expecting actual:
  DECIMAL16
to be in:
  [DECIMAL4, DECIMAL8]
```

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantParseJson.java
```diff
@@ -212,59 +210,59 @@ public void testParseComplexDocument() throws IOException {
...
-  public void testParseMalformedJson() throws IOException {
-    VariantJsonParser.parseJson("{invalid");
+  @Test
+  public void testParseMalformedJson() {
+    assertThatThrownBy(() -> VariantJsonParser.parseJson("{invalid"))
+        .isInstanceOf(IOException.class)
+        .hasMessageContaining("was expecting double-quote to start field name");
   }
 
-  @Test(expected = IOException.class)
```
nastra, line 265: we're also more explicit here in terms of testing failure conditions

emkornfield:
@nastra I think there are several libraries that provide this functionality  (e.g. I think there is a Google Truth library) maybe start a thread on the mailing list to just make sure nobody has any strong technical reasons for preferring one over the other.

wgtmac (approved):

PR #3622, opened 2026-06-19
GH-3621: Replace the previous field in the queue instead of appending to the queue

### LINKED ISSUE #3621: Memory leak in previousField of ValidatingRecordConsumer

I stumbled on this when an MQTT to parquet program that I'm working on kept getting OOM errors much sooner than I expected.  I know the data needs to be retained in memory until it's written to disk, but the JVM was showing significantly more memory growth than the data.  I probably wouldn't have noticed this except my data itself is largely uint8 and uint16 data from a PLC, so an extra ~24 bytes per field is a lot (relatively) in this case.

The .hprof showed:

Image

Tested on 1.17.1 and master@f280d557509da65df7e205dce6898bce9b4a396e.

It looks like when `endField()` is called, it pushes the current field onto the `previousField` queue instead of replacing the last previous field, so as records are processed, the `previousField` queue grows indefinitely in $O(\text{records} × \text{fields})$

https://github.com/apache/parquet-java/blob/f280d557509da65df7e205dce6898bce9b4a396e/parquet-column/src/main/java/org/apache/parquet/io/ValidatingRecordConsumer.java#L112-L116

### Component(s)

Core

njlaw:
### Rationale for this change

The `endField` method pushes the current field onto the `previousField` queue, which is used to maintain track of the previous field for each level of nesting with groups; however, the pushed fields are never removed from the queue, so it grows without bound.

When `endField()` is called, it now replaces the previous field for the current nesting depth instead of appending a field.

### Are these changes tested?

I added a test that verifies the `previousField` size is 0 after processing a message; however, the test requires introspection, which feels wrong.  I'm open to suggestions!

No, there are no user-facing changes.

Closes #3621

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/ValidatingRecordConsumer.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/ValidatingRecordConsumer.java
@@ -112,6 +112,7 @@ public class ValidatingRecordConsumer extends RecordConsumer {
   public void endField(String field, int index) {
     delegate.endField(field, index);
     fieldValueCount.pop();
+    previousField.pop();
     previousField.push(fields.pop());
   }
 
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

wgtmac:
Thanks for fixing this!

Fokko (approved):
Great catch!

PR #3626, opened 2026-06-22
GH-3625: Compute dataSize correctly when dealing with duplicate keys

### LINKED ISSUE #3625: dataSize is wrongly computed for duplicate keys in VariantBuilder

When a Variant object is built with the same key appended more than once, `VariantBuilder.endObject()` deduplicates the entries and keeps the last-written value. However, the total `dataSize` of the object's value region was computed from each key's **first** occurrence rather than the value that was actually retained.

When duplicate keys had values of different encoded sizes, this size mismatch corrupted the output:

* If the retained value was larger (e.g. `{"a": 1, "a": "hello"}`), the buffer was under-reserved, the offset width could be undersized, and the final offset entry understated the real data length. Reading the field then threw `IllegalArgumentException: Invalid byte-array offset` or returned truncated data.
* If the retained value was smaller, the object reserved stale trailing bytes and overstated its size.

### Component(s)

Core

nastra:
### Rationale for this change
When a Variant object is built with the same key appended more than once, `VariantBuilder.endObject()` deduplicates the entries and keeps the last-written value. However, the total `dataSize` of the object's value region was computed from each key's **first** occurrence rather than the value that was actually retained.

When duplicate keys had values of different encoded sizes, this size mismatch corrupted the output:

* If the retained value was larger (e.g. `{"a": 1, "a": "hello"}`), the buffer was under-reserved, the offset width could be undersized, and the final offset entry understated the real data length. Reading the field then threw `IllegalArgumentException: Invalid byte-array offset` or returned truncated data.
* If the retained value was smaller, the object reserved stale trailing bytes and overstated its size.

Closes https://github.com/apache/parquet-java/issues/3625

### Are these changes tested?
yes, added tests that reproduce the underlying issue

### DIFF
--- a/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObjectBuilder.java
+++ b/parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObjectBuilder.java
@@ -278,6 +278,72 @@ public class TestVariantObjectBuilder {
   }
 
   @Test
+  public void testDuplicateKeysKeptValueLarger() {
+    // The retained (last-written) value is larger than the first occurrence. The data size must be
+    // computed from the retained value, otherwise the encoded object is truncated/corrupt.
+    VariantBuilder b = new VariantBuilder();
+    VariantObjectBuilder objBuilder = b.startObject();
+    objBuilder.appendKey("duplicate");
+    objBuilder.appendInt(1); // 5 bytes
+    objBuilder.appendKey("duplicate");
+    objBuilder.appendString("hello"); // 6 bytes
+    b.endObject();
+    VariantTestUtil.testVariant(b.build(), v -> {
+      VariantTestUtil.checkType(v, VariantUtil.OBJECT, Variant.Type.OBJECT);
+      Assert.assertEquals(1, v.numObjectElements());
+      Variant variant = v.getFieldByKey("duplicate");
+      VariantTestUtil.checkType(variant, VariantUtil.SHORT_STR, Variant.Type.STRING);
+      Assert.assertEquals("hello", variant.getString());
+    });
+  }
+
+  @Test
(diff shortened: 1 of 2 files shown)


parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObjectBuilder.java
```diff
@@ -277,6 +277,72 @@ public void testDuplicateKeys() {
...
+      VariantTestUtil.checkType(v, VariantUtil.OBJECT, Variant.Type.OBJECT);
+      Assert.assertEquals(1, v.numObjectElements());
+      Variant variant = v.getFieldByKey("duplicate");
+      VariantTestUtil.checkType(variant, VariantUtil.PRIMITIVE, Variant.Type.INT);
+      Assert.assertEquals(1, variant.getInt());
+    });
+  }
+
+  @Test
+  public void testDuplicateKeysDifferentSizesAcrossMultipleKeys() {
```
nastra, line 321: without the fix this test would fail with `java.lang.IllegalArgumentException: Invalid byte-array offset (25). length: 22`

parquet-variant/src/test/java/org/apache/parquet/variant/TestVariantObjectBuilder.java
```diff
@@ -277,6 +277,72 @@ public void testDuplicateKeys() {
     Assert.assertEquals(1, v.getFieldByKey("duplicate").getLong());
   }
 
+  @Test
+  public void testDuplicateKeysKeptValueLarger() {
```
nastra, line 281: without the fix this test would fail with `java.lang.IllegalArgumentException: Invalid byte-array offset (10). length: 10`

wgtmac (approved):

Fokko (approved):

PR #3616, opened 2026-06-19
GH-3615: Introduce AssertJ for testing

nastra:
### Rationale for this change

This introduces AssertJ for testing, which enables more fluent assertion checks, easier debugging and overall easier testing. Additional details about the motivation can be found in https://github.com/apache/parquet-java/issues/3615

I've also opened https://github.com/apache/parquet-java/pull/3617 where I'm using AssertJ checks in the `parquet-variant` module to show how those assertions make the test code more readable and more explicit when it comes to testing error conditions.

### What changes are included in this PR?
added the assertj library to the test scope

PR #3629, opened 2026-06-22
GH-3628: prevent NPE & unclosed releaser

### LINKED ISSUE #3628: ColumnChunkPageReadStore.close() NPEs if setReleaser() was never called, skipping buffer cleanup in finally

ColumnChunkPageReadStore.close() NPEs if setReleaser() was never called. Also it's possible that the `releaser` itself isn't closed when releasing the buffers of the readers fails anywhere in the loop

### Component(s)

Core

nastra:
### Rationale for this change
ColumnChunkPageReadStore.close() NPEs if setReleaser() was never called. Also it's possible that the `releaser` itself isn't closed when releasing the buffers of the readers fails anywhere in the loop

Closes https://github.com/apache/parquet-java/issues/3628

### Are these changes tested?
added a test that reproduces both cases

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -410,9 +410,14 @@ class ColumnChunkPageReadStore implements PageReadStore, DictionaryPageReadStore
 
   @Override
   public void close() {
-    for (ColumnChunkPageReader reader : readers.values()) {
-      reader.releaseBuffers();
+    try {
+      for (ColumnChunkPageReader reader : readers.values()) {
+        reader.releaseBuffers();
+      }
+    } finally {
+      if (releaser != null) {
+        releaser.close();
+      }
     }
-    releaser.close();
   }
 }
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
@@ -21,6 +21,7 @@ package org.apache.parquet.hadoop;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -46,6 +47,7 @@ import org.apache.parquet.format.BlockCipher;
 import org.apache.parquet.internal.column.columnindex.OffsetIndex;
 import org.apache.parquet.internal.filter2.columnindex.RowRanges;
 import org.apache.parquet.io.ParquetDecodingException;
+import org.apache.parquet.util.AutoCloseables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageReadStore.java
```diff
@@ -0,0 +1,149 @@
...
+  public void closeWithoutSetReleaserDoesNotThrow() {
+    try (TrackingByteBufferAllocator allocator = TrackingByteBufferAllocator.wrap(new HeapByteBufferAllocator())) {
+      ParquetReadOptions options =
+          ParquetReadOptions.builder().withAllocator(allocator).build();
+
+      ColumnChunkPageReadStore store = new ColumnChunkPageReadStore(0L);
+      store.addColumn(COLUMN, newReaderWithoutPages(options));
+
+      // setReleaser() is intentionally NOT called here.
+      store.close();
```
nastra, line 72: having AssertJ (https://github.com/apache/parquet-java/pull/3616) would make testing this a bit easier and more explicit, because we could then just say `assertThatCode(store::close).doesNotThrowAnyException()`

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestColumnChunkPageReadStore.java
```diff
@@ -0,0 +1,149 @@
...
+      store.addColumn(COLUMN, newReaderWithQueuedBuffer(throwingAllocator));
+
+      // The store-level releaser holds a tracked buffer that must be released by close()'s finally block.
+      ByteBufferReleaser storeReleaser = new ByteBufferReleaser(storeAllocator);
+      storeReleaser.releaseLater(storeAllocator.allocate(8));
+      store.setReleaser(storeReleaser);
+
+      try {
+        store.close();
+        throw new AssertionError("Expected close() to propagate the reader failure");
```
nastra, line 94: AssertJ would also simplify testing this error condition. We would end up having `assertThatThrownBy(store::close).isInstanceof(...).hasMessage(...)`

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageReadStore.java
```diff
@@ -410,9 +410,14 @@ void setReleaser(ByteBufferReleaser releaser) {
 
   @Override
   public void close() {
-    for (ColumnChunkPageReader reader : readers.values()) {
-      reader.releaseBuffers();
+    try {
+      for (ColumnChunkPageReader reader : readers.values()) {
```
wgtmac: Can we keep closing the remaining readers if one `releaseBuffers()` call fails? As written, a failure still leaves later readers unreleased, and a `releaser.close()` failure would hide the original exception.

nastra: that's a good point. I've slightly updated the code so that any occurring failure is being kept and shows up in the final exception that is being thrown

nastra: also note that `AutoCloseables` will correctly deal with a `null`

wgtmac (approved):

Fokko (approved):
Nice one! 👍

PR #3631, opened 2026-06-23
GH-3632: Fix record-level notIn evaluation

### LINKED ISSUE #3632: Record-level notIn filter can return matching values
### Describe the bug
The record-level filter implementation for `notIn` can accept a row after comparing the row value with only the first non-matching literal in the set.

For example, `notIn(col, {1.0, 3.0})` should reject `3.0`, but the current generated predicate can accept it after seeing that `3.0 != 1.0`.

### Expected behavior
`notIn` should scan the full literal set and accept a row only when no set value equals the row value.

### Affected area
Record-level filter predicate evaluation.

wgtmac:
### Rationale for this change

The generated record-level filter for `notIn` can accept a row after checking only the first non-matching value in the literal set. For example, `notIn(col, {1.0, 3.0})` should reject `3.0`, but the current logic can accept it after seeing `3.0 != 1.0`.

- Evaluate `notIn` by checking whether any literal equals the row value before deciding the result.
- Add a regression test for multi-value `notIn` record-level filtering.

### Are these changes tested?

- `./mvnw -q -Drat.skip=true -DskipTests -pl parquet-generator,parquet-column install`
- `./mvnw -q -Drat.skip=true -pl parquet-hadoop -Dtest=TestRecordLevelFilters -Dsurefire.failIfNoSpecifiedTests=false test`

This fixes record-level filtering for `notIn` predicates with multiple literal values.

Closes #3632

### DIFF
--- a/parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
+++ b/parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
@@ -281,51 +281,50 @@ public class IncrementallyUpdatedFilterPredicateGenerator {
   }
 
   private void addInNotInCase(TypeInfo info, boolean isEq, boolean expectMultipleResults) throws IOException {
-    add("    if (clazz.equals(" + info.className + ".class)) {\n" + "      if (pred.getValues().contains(null)) {\n"
-        + "        valueInspector = new ValueInspector() {\n"
-        + "          @Override\n"
-        + "          public void updateNull() {\n"
-        + "            setResult("
-        + isEq + ");\n" + "          }\n"
-        + "\n"
-        + "          @Override\n"
-        + "          public void update("
-        + info.primitiveName + " value) {\n" + "            setResult("
-        + !isEq + ");\n" + "          }\n"
-        + "        };\n"
-        + "      } else {\n"
-        + "        final Set<"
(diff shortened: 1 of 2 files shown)


wgtmac:
Just caught this bug while working on https://github.com/apache/parquet-java/pull/3393. Could you please take a look at this? @gszadovszky @Fokko

parquet-generator/src/main/java/org/apache/parquet/filter2/IncrementallyUpdatedFilterPredicateGenerator.java
gszadovszky: This branch still looks incorrect for `notIn` sets that contain null plus non-null values. For example, `notIn(col, {null, 3.0})` will emit `setResult(true)` for any non-null value, so a row where `col == 3.0` is still kept.

wgtmac: Fixed in the latest push. The generated evaluator now handles null membership only in updateNull(), and update(value) still scans the non-null values with the column comparator. Added a regression test for in/notIn with {null, 3.0}; notIn now rejects col == 3.0 and keeps only the non-matching non-null value.

gszadovszky (approved):
Thanks @wgtmac for fixing this!
LGTM.

wgtmac:
Thank you, @gszadovszky!

Issue #3652, opened 2026-07-09
Convert JUnit4-style asserts to AssertJ


nastra:
### Describe the enhancement requested

Convert JUnit4-style asserts to AssertJ as this makes it easier to migrate to JUnit5 later on

### Component(s)

Core

PR #3654, opened 2026-07-09
GH-3652: Use assertThat checks in parquet-common tests

nastra:
### Rationale for this change
Switching assertions from JUnit4-style Assert to AssertJ makes code easier to read and debug when tests fail.

This updates parquet-common tests to use AssertJ checks, which provide more fluent assertions and more detailed information when assertions fails on CI.

The changes in this PR were done with the help of Claude but I manually reviewed every single LOC

### Are these changes tested?
yes, existing tests

no


parquet-common/src/test/java/org/apache/parquet/TestUtils.java
```diff
@@ -1,68 +0,0 @@
...
-public class TestUtils {
-
-  /**
-   * A convenience method to avoid a large number of @Test(expected=...) tests
-   *
-   * @param message  A String message to describe this assertion
-   * @param expected An Exception class that the Runnable should throw
-   * @param callable A Callable that is expected to throw the exception
-   */
-  public static void assertThrows(String message, Class<? extends Exception> expected, Callable callable) {
```
nastra, line 34: I've replaced all usages of these methods because using assertThatThrownBy() directly gives you greater flexibility in testing

parquet-common/src/test/java/org/apache/parquet/SemanticVersionTest.java
```diff
@@ -18,34 +18,29 @@
...
   @Test
   public void testCompare() {
-    assertTrue(new SemanticVersion(1, 8, 1).compareTo(new SemanticVersion(1, 8, 1)) == 0);
-    assertTrue(new SemanticVersion(1, 8, 0).compareTo(new SemanticVersion(1, 8, 1)) < 0);
-    assertTrue(new SemanticVersion(1, 8, 2).compareTo(new SemanticVersion(1, 8, 1)) > 0);
+    assertThat(new SemanticVersion(1, 8, 1)).isEqualByComparingTo(new SemanticVersion(1, 8, 1));
+    assertThat(new SemanticVersion(1, 8, 0)).isLessThan(new SemanticVersion(1, 8, 1));
+    assertThat(new SemanticVersion(1, 8, 2)).isGreaterThan(new SemanticVersion(1, 8, 1));
 
-    assertTrue(new SemanticVersion(1, 8, 1).compareTo(new SemanticVersion(1, 8, 1)) == 0);
```
nastra, line 34: these 3 checks were checking the same thing and were duplicates

parquet-common/src/test/java/org/apache/parquet/bytes/TestByteBufferInputStreams.java
```diff
@@ -155,12 +175,11 @@ public void testReadByte() throws Exception {
...
     for (int i = 0; i < length; i += 1) {
-      Assert.assertEquals("Position should increment", i, stream.position());
-      Assert.assertEquals(i, stream.read());
+      assertThat(stream.position()).as("Position should increment").isEqualTo(i);
+      assertThat(stream.read()).isEqualTo(i);
     }
 
-    assertThrows(
-        "Should throw EOFException at end of stream", EOFException.class, (Callable<Integer>) stream::read);
+    assertThatThrownBy(stream::read).isInstanceOf(EOFException.class);
```
nastra, line 182: the exception message is null so we're omitting the hasMessage() check here and in a few other places. Depending on the JDK version, it's also possible that the exception either has a null message or has some message, so it's safer to omit the check

parquet-common/src/test/java/org/apache/parquet/bytes/TestByteBufferInputStreams.java
```diff
@@ -155,12 +175,11 @@ public void testReadByte() throws Exception {
     int length = stream.available();
 
     for (int i = 0; i < length; i += 1) {
-      Assert.assertEquals("Position should increment", i, stream.position());
-      Assert.assertEquals(i, stream.read());
+      assertThat(stream.position()).as("Position should increment").isEqualTo(i);
+      assertThat(stream.read()).isEqualTo(i);
     }
 
-    assertThrows(
```
nastra, line 162: we're replacing all usages of `assertThrows` because using `assertThatThrownBy` gives you greater flexibility for localized testing, in case a test case would want to add some additional checks on top of what already is being checked

parquet-common/src/test/java/org/apache/parquet/bytes/TestDeprecatedBufferInputStream.java
```diff
@@ -96,57 +96,57 @@ public void testSliceData() throws Exception {
...
 
-    Assert.assertEquals("Position should be at end", length, stream.position());
-    Assert.assertEquals("Should produce 5 buffers", 5, buffers.size());
+    assertThat(stream.position()).as("Position should be at end").isEqualTo(length);
+    assertThat(buffers).as("Should produce 5 buffers").hasSize(5);
 
     int i = 0;
 
     ByteBuffer one = buffers.get(0);
-    Assert.assertSame("Should use the same backing array", one.array(), data.array());
```
nastra, line 105: this is an example where the actual/expected parameters are flipped in the JUnit4 Assert. We're expecting `one` to be the actual object and compare it against `data`

parquet-common/src/test/java/org/apache/parquet/bytes/TestMultiBufferInputStream.java
```diff
@@ -60,67 +61,65 @@ public void testSliceData() throws Exception {
...
-    Assert.assertEquals("Position should be at end", length, stream.position());
-    Assert.assertEquals("Should produce 5 buffers", 5, buffers.size());
+    assertThat(stream.position()).as("Position should be at end").isEqualTo(length);
+    assertThat(buffers).as("Should produce 5 buffers").hasSize(5);
 
     int i = 0;
 
     // one is a view of the first buffer because it is smaller
     ByteBuffer one = buffers.get(0);
-    Assert.assertSame(
```
nastra, line 70: actual/expected were flipped here as well

parquet-common/src/test/java/org/apache/parquet/glob/TestGlob.java
```diff
@@ -31,45 +29,44 @@ public class TestGlob {
...
-    assertEquals(List.of("", ""), Strings.expandGlob("{,}"));
-    assertEquals(List.of("ab", "a", "ac"), Strings.expandGlob("a{b,{},c}"));
+    assertThat(Strings.expandGlob("")).containsExactly("");
+    assertThat(Strings.expandGlob("{}")).containsExactly("");
+    assertThat(Strings.expandGlob("a{}")).containsExactly("a");
+    assertThat(Strings.expandGlob("a{}b")).containsExactly("ab");
+    assertThat(Strings.expandGlob("{}a")).containsExactly("a");
+    assertThat(Strings.expandGlob("a{}")).containsExactly("a");
+    assertThat(Strings.expandGlob("{,}")).containsExactly("", "");
+    assertThat(Strings.expandGlob("a{b,{},c}")).containsExactly("ab", "a", "ac");
```
nastra, line 44: we don't need to compare against actual lists. It's easier to just verify that the `actual` contains those elements in that specified order

parquet-common/src/test/java/org/apache/parquet/glob/TestWildcardPath.java
```diff
@@ -18,121 +18,172 @@
...
-    }
-  }
-
   @Test
   public void testNoWildcards() {
     WildcardPath wp = new WildcardPath("", "foo", '.');
-    assertMatches(wp, "foo", "foo.x", "foo.x.y");
-    assertDoesNotMatch(wp, "xfoo", "xfoox", "fooa.x.y");
+
+    assertThat(wp.matches("foo")).isTrue();
```
nastra, line 31: it is easier to debug a single line if the test ever fails

wgtmac (approved):

Fokko (approved):

PR #3662, opened 2026-07-14
GH-3667: Close input readers when ParquetRewriter setup fails

### LINKED ISSUE #3667: Close input readers when ParquetRewriter setup fails

`ParquetRewriter.getFileReaders` opens each input reader into a local list. If opening a later input throws `IOException`, the exception is propagated without closing readers already opened for earlier inputs. Because construction fails, callers cannot close those readers, which leaks their input streams.

Expected behavior: readers opened before the failure are closed before the setup exception is propagated.

### Component(s)

Core

anxkhn:
### Rationale for this change

`ParquetRewriter.getFileReaders` opens a `TransParquetFileReader` for each input
file into a local list. If opening a later file throws `IOException`, it is
rethrown as `IllegalArgumentException` without closing the readers already opened
for the earlier files, so their `SeekableInputStream` handles leak. A single
corrupt, missing, or permission-denied file among several valid inputs can
therefore leak the input streams opened before the partial-open failure.

In `getFileReaders`, close the readers already opened before rethrowing, using
`org.apache.parquet.util.AutoCloseables.close(...)` so every reader is released
and any close failures are aggregated as suppressed exceptions on the original
`IllegalArgumentException`. This matches the existing cleanup idiom in the module.
The change is limited to the `catch (IOException)` block; the success path is
unchanged.

### Are these changes tested?

Yes. A new `ParquetRewriterTest` case
(`testInputFileReadersClosedWhenLaterInputFileFailsToOpen`) supplies a valid
input file (wrapped so it records whether the stream it hands out is closed)
followed by an `InputFile` whose `newStream()` throws, asserts that constructing
the `ParquetRewriter` throws `IllegalArgumentException`, and asserts the first
reader's stream was closed. It fails on the pre-fix code (the first stream is
left open) and passes with the fix. The full `ParquetRewriterTest` class passes
(`Tests run: 96, Failures: 0, Errors: 0`).

No public API changes. On a partial-open failure while opening input files,
`ParquetRewriter` now closes the input streams it had already opened.

Closes #3667

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
@@ -46,6 +46,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -88,6 +89,7 @@ import org.apache.parquet.hadoop.util.HadoopOutputFile;
 import org.apache.parquet.hadoop.util.TestFileBuilder;
 import org.apache.parquet.internal.column.columnindex.ColumnIndex;
 import org.apache.parquet.internal.column.columnindex.OffsetIndex;
+import org.apache.parquet.io.DelegatingSeekableInputStream;
 import org.apache.parquet.io.InputFile;
 import org.apache.parquet.io.InvalidRecordException;
 import org.apache.parquet.io.OutputFile;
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Seams reasonable to me 👍

wgtmac:
I would recommend creating an issue for this. Or at least do not mark it as `MINOR`.

anxkhn:
Created #3667 and updated the PR title and body to reference it. Thanks for the guidance.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
```diff
@@ -689,6 +691,84 @@ public void testMergeTwoFilesNullifyAndRenamedSameColumn() throws Exception {
...
+      public SeekableInputStream newStream() throws IOException {
+        throw new IOException("simulated open failure");
+      }
+    };
+
+    OutputFile output = HadoopOutputFile.fromPath(new Path(outputFile), conf);
+    RewriteOptions options =
+        new RewriteOptions.Builder(parquetConf, Arrays.asList(openable, failing), output).build();
+
+    assertThrows(IllegalArgumentException.class, () -> new ParquetRewriter(options));
```
nastra: can you please rebase the PR and use AssertJ assertions? This would be something like `assertThatThrownBy(() -> new ParquetRewriter(options)).isInstanceOf(IllegalArgumentException.class).hasMessage(...)`.
The other one would be a simple `assertThat(openable.isStreamClosed()).isTrue()`

anxkhn:
Rebased onto current master and updated the regression test to use AssertJ assertions as requested. I also narrowed the PR description to the partial-open failure path covered by this change. The targeted regression test passes (4 tests).

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/rewrite/ParquetRewriterTest.java
anxkhn: rebased and updated both assertions to use assertj. thanks!

anxkhn:
@fokko thanks for the approval. is there anything else needed from me before this can be merged?

PR #3663, opened 2026-07-14
MINOR: Close output stream when ParquetFileWriter construction fails on encryption validation

anxkhn:
### Rationale for this change

The shared private `ParquetFileWriter` constructor opens the output stream early
(`OutputFile.create` / `createOrOverwrite`), and only afterwards validates that
every column named in the encryption properties actually exists in the file
schema. When a configured encrypted column is missing, it throws
`ParquetCryptoRuntimeException`.

That validation runs after the stream is opened and is not guarded, so on the
throw the half-constructed writer is never returned to the caller. Its `close()`
is the only place that closes the stream, and it can never run, so the open
stream (and its underlying file/DFS handle and buffers) leaks. A service that
repeatedly constructs writers against a misconfigured encryption schema
accumulates leaked handles until exhaustion.

This is the same acquire-then-throw hazard the sibling `ParquetFileReader`
constructor already guards against by closing its stream before rethrowing, so
the writer should behave consistently.

- In `ParquetFileWriter`, wrap the encryption-setup and encrypted-column-in-schema
  validation block (which runs after `this.out` is assigned) in a
  `try { ... } catch (Exception e) { out.close(); throw e; }`. On any exception
  the already-opened stream is closed before the exception propagates, mirroring
  the existing guard in the `ParquetFileReader` constructor. The early return for
  the no-encryption case is left outside the guard (nothing to clean up there).
- Add a regression test, `testConstructorClosesStreamWhenEncryptedColumnMissing`,
  that constructs a writer with an encrypted column absent from the schema,
  asserts the expected `ParquetCryptoRuntimeException`, and asserts the stream was
  closed. It uses a small `RecordingOutputFile` whose `PositionOutputStream`
  flips an `AtomicBoolean` on `close()`, so it needs no disk I/O.

Two files change: `ParquetFileWriter.java` (the guard) and
`TestParquetFileWriter.java` (the test plus helper).

### Are these changes tested?

Yes.

- The new test is red without the source fix (the stream is left open) and green
  with it. It runs parameterized (vectored true/false), 2/2 passing.
- Regression sweep passes: `TestParquetFileWriter` (38), `TestParquetWriter`
  (19), `TestEncryptionOptions` (1), 0 failures / 0 errors.
- Spotless is clean on both changed files.

Commands used:

```
mvn -q -pl parquet-hadoop -am -DskipTests -Dspotless.check.skip=true install
mvn -pl parquet-hadoop -Dtest=TestParquetFileWriter#testConstructorClosesStreamWhenEncryptedColumnMissing test
mvn -pl parquet-hadoop -Dtest=TestParquetFileWriter,TestParquetWriter,TestEncryptionOptions test
```

No. Behavior on the success path is unchanged; the only difference is that a
failed writer construction (invalid encrypted-column configuration) now releases
the output stream it opened instead of leaking it. No public API changes.

### DIFF
--- a/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
+++ b/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
@@ -48,6 +48,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileStatus;
@@ -76,6 +77,9 @@ import org.apache.parquet.column.statistics.BinaryStatistics;
 import org.apache.parquet.column.statistics.LongStatistics;
 import org.apache.parquet.column.values.bloomfilter.BlockSplitBloomFilter;
 import org.apache.parquet.column.values.bloomfilter.BloomFilter;
+import org.apache.parquet.crypto.ColumnEncryptionProperties;
+import org.apache.parquet.crypto.FileEncryptionProperties;
+import org.apache.parquet.crypto.ParquetCryptoRuntimeException;
 import org.apache.parquet.example.data.Group;
 import org.apache.parquet.example.data.simple.SimpleGroup;
 import org.apache.parquet.format.Statistics;
@@ -84,6 +88,7 @@ import org.apache.parquet.hadoop.example.GroupReadSupport;
 import org.apache.parquet.hadoop.example.GroupWriteSupport;
 import org.apache.parquet.hadoop.metadata.BlockMetaData;
 import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.hadoop.metadata.ConcatenatingKeyValueMetadataMergeStrategy;
 import org.apache.parquet.hadoop.metadata.FileMetaData;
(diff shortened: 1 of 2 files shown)


parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
```diff
@@ -1498,6 +1505,94 @@ public void testMergeMetadataWithNoConflictingKeyValues() {
...
+          file,
+          SCHEMA,
+          CREATE,
+          DEFAULT_BLOCK_SIZE,
+          MAX_PADDING_SIZE_DEFAULT,
+          ParquetProperties.DEFAULT_COLUMN_INDEX_TRUNCATE_LENGTH,
+          ParquetProperties.DEFAULT_STATISTICS_TRUNCATE_LENGTH,
+          ParquetProperties.DEFAULT_PAGE_WRITE_CHECKSUM_ENABLED,
+          encryptionProperties);
+      fail("Expected ParquetCryptoRuntimeException for an encrypted column absent from the schema");
```
nastra: we're migrating the codebase to AssertJ assertions, so you might want to change this to `assertThatThrownBy(() -> ...).isInstanceOf(ParquetCryptoRuntimeException.class).hasMessage(...)`

anxkhn: Updated the exception assertion to AssertJ and included the expected message. Thanks.

anxkhn:
Addressed the review feedback, rebased onto the latest master, and reran the targeted test suite successfully.

dossett:
+1 (non-binding)

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestParquetFileWriter.java
anxkhn: rebased onto current master and retained the assertj assertion. the targeted test and full TestParquetFileWriter suite pass. thanks!

PR #3665, opened 2026-07-14
Fix CI caused by a silent merge conflict

nastra:
### Rationale for this change
This fixes CI which was caused by a silent merge conflict

### Are these changes tested?
yes, existing tests


Fokko (approved):
Thanks for the quick follow-up @nastra 🙌

Issue #3618, opened 2026-06-19
Migrate from JUnit4 to JUnit5


nastra:
### Describe the enhancement requested

The last version of JUnit4 was released on Feb 13, 2021 and I think it makes sense to upgrade the codebase gradually to JUnit5. As a first step, we can upgrade the dependency without upgrading any of the tests. Those will continue to run as JUnit4 tests. 
Then gradually we could migrate modules one-by-one to JUnit5

PR #3673, opened 2026-07-17
GH-3618: Migrate multiple modules to JUnit5

nastra:
### Rationale for this change
Now that these modules have been migrated to AssertJ assertions, migrating them to JUnit5 is trivial and mostly requires import changes. 
While this touches 180+ files, there are actually < 10 files that have substantial changes (mostly around parameterized testing). All the other files have pure import changes. JUnit5 doesn't support built-in class-level test parameterization and so those tests were converted to parameterize at the method level.

I'll have a second PR that will migrate parquet-hadoop to JUnit5 because that requires few more substantial changes

### Are these changes tested?
yes

no

### DIFF
--- a/parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWriteOldListBehavior.java
+++ b/parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWriteOldListBehavior.java
@@ -34,7 +34,6 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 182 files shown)


parquet-column/src/test/java/org/apache/parquet/column/values/delta/benchmark/BenchmarkReadingRandomIntegers.java
```diff
@@ -33,10 +33,12 @@
...
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesReader;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridValuesWriter;
-import org.junit.BeforeClass;
 import org.junit.Rule;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.migrationsupport.rules.EnableRuleMigrationSupport;
 
+@EnableRuleMigrationSupport
```
nastra, line 41: these JUnit benchmarks depend on `com.carrotsearch.junitbenchmarks.BenchmarkRule` which is only available for JUnit4, so we can't really convert that Rule to JUnit5. We're making sure that this uses JUnit5 imports but runs with the migration module, which supports JUnit4 rule execution

parquet-avro/src/test/java/org/apache/parquet/avro/TestReadWriteOldListBehavior.java
```diff
@@ -53,32 +52,23 @@
 import org.apache.parquet.io.api.Binary;
 import org.apache.parquet.io.api.RecordConsumer;
 import org.apache.parquet.schema.MessageTypeParser;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
-@RunWith(Parameterized.class)
```
nastra, line 60: JUnit5 doesn't have class-level parameterization unfortunately, so everything is parameterized at the method level

parquet-common/src/test/java/org/apache/parquet/bytes/TestDeprecatedBufferInputStream.java
```diff
@@ -22,141 +22,191 @@
...
-@RunWith(Parameterized.class)
-public class TestDeprecatedBufferInputStream extends TestByteBufferInputStreams {
-  @Parameter(0)
-  public ByteBuffer data;
-
-  @Parameter(1)
-  public Integer offset;
-
-  @Parameters
-  public static List<Object[]> parameters() {
```
nastra, line 45: this refactors the test slightly to have nested test classes that use their own parameters

wgtmac (approved):

PR #3678, opened 2026-07-20
MINOR: Allow callers to provide known Hadoop file lengths

dossett:
In an earlier PR @steveloughran suggested that callers who already knew the length of a file (such as Iceberg) could supply that value when opening a file and save at least on metadata lookup.  This is the result of me working through that idea with codex, not sure if it's what was originally intended.

It does bump the hadoop dependency from 3.3.0 to 3.3.5.  I assume that's safe, but calling it out just in case.

CODEX summary follows
------
Metadata-aware callers may already know a Parquet file's exact length from a catalog, manifest, or prior listing. `HadoopInputFile.fromPath(Path, Configuration)` currently performs `getFileStatus()` unconditionally, discarding that knowledge and potentially adding a remote metadata request before the file can be read.

This adds a length-aware factory that trusts the caller-provided value and passes it to Hadoop's `openFile()` API. The API is format-neutral, but it enables table formats and other metadata-aware readers to avoid the status lookup when they can guarantee the length. Existing path- and status-based construction retain their current behavior. This follows up on the known-length optimization discussed in #3395.

The change also raises the Hadoop baseline from 3.3.0 to 3.3.5, where `FS_OPTION_OPENFILE_LENGTH` was standardized and supported by S3A.

Summary:

- Add `HadoopInputFile.fromPath(Path, long, Configuration)`.
- Preserve `FileStatus` as the preferred open hint when one is already available.
- Otherwise pass the known length through `FS_OPTION_OPENFILE_LENGTH`.
- Reject negative lengths and document that the supplied length is trusted without validation.

Testing:

- `./mvnw -pl parquet-hadoop -Dtest=TestHadoopOpenFile test`
- `./mvnw -q -pl parquet-hadoop -DskipTests spotless:check`

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
@@ -69,7 +69,8 @@ public class HadoopInputFile implements InputFile {
    * Creates an input file using a caller-supplied file length.
    *
    * <p>The length is trusted and no file status lookup is performed. Callers must provide the exact
-   * length of the file that will be opened. The file is not validated until it is opened.
+   * length of the file that will be opened. The file system may defer checking whether the file
+   * exists until the first read. If the supplied length is incorrect, reads may fail.
    *
    * @param path file path
    * @param length exact file length in bytes
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
@@ -19,6 +19,7 @@
 
 package org.apache.parquet.hadoop.util;
 
+import static org.apache.hadoop.fs.Options.OpenFileOptions.FS_OPTION_OPENFILE_LENGTH;
 import static org.apache.parquet.hadoop.util.wrapped.io.FutureIO.awaitFuture;
 
 import java.io.IOException;
@@ -28,6 +29,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FutureDataInputStreamBuilder;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.io.InputFile;
 import org.apache.parquet.io.SeekableInputStream;
@@ -53,12 +55,35 @@ public class HadoopInputFile implements InputFile {
   private static final String PARQUET_READ_POLICY = "parquet, vector, random, adaptive";
 
   private final FileSystem fs;
+  private final Path path;
+  private final long length;
   private final FileStatus stat;
   private final Configuration conf;
 
   public static HadoopInputFile fromPath(Path path, Configuration conf) throws IOException {
     FileSystem fs = path.getFileSystem(conf);
     return new HadoopInputFile(fs, fs.getFileStatus(path), conf);
+  }
+
+  /**
+   * Creates an input file using a caller-supplied file length.
+   *
+   * <p>The length is trusted and no file status lookup is performed. Callers must provide the exact
+   * length of the file that will be opened. The file is not validated until it is opened.
+   *
+   * @param path file path
+   * @param length exact file length in bytes
+   * @param conf configuration used to resolve the file system
+   * @return an input file that uses the supplied length
+   * @throws IllegalArgumentException if {@code length} is negative
+   * @throws IOException if the file system cannot be resolved
+   */
+  public static HadoopInputFile fromPath(Path path, long length, Configuration conf) throws IOException {
+    if (length < 0) {
+      throw new IllegalArgumentException("Invalid file length: " + length);
+    }
+    FileSystem fs = path.getFileSystem(conf);
+    return new HadoopInputFile(fs, path, length, conf);
   }
 
   public static HadoopInputFile fromPathUnchecked(Path path, Configuration conf) {
@@ -76,7 +101,17 @@ public class HadoopInputFile implements InputFile {
 
   private HadoopInputFile(FileSystem fs, FileStatus stat, Configuration conf) {
     this.fs = fs;
+    this.path = stat.getPath();
+    this.length = stat.getLen();
     this.stat = stat;
+    this.conf = conf;
+  }
+
+  private HadoopInputFile(FileSystem fs, Path path, long length, Configuration conf) {
+    this.fs = fs;
+    this.path = path;
+    this.length = length;
+    this.stat = null;
     this.conf = conf;
   }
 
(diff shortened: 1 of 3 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -61,6 +65,27 @@ public static HadoopInputFile fromPath(Path path, Configuration conf) throws IOE
...
+   * @return an input file that uses the supplied length
+   * @throws IllegalArgumentException if {@code length} is negative
+   * @throws IOException if the file system cannot be resolved
+   */
+  public static HadoopInputFile fromPath(Path path, long length, Configuration conf) throws IOException {
+    if (length < 0) {
+      throw new IllegalArgumentException("Invalid file length: " + length);
+    }
+    FileSystem fs = path.getFileSystem(conf);
+    return new HadoopInputFile(fs, path, length, conf);
```
steveloughran, line 91: this is the only place the new ctor is called, isn't it? so it can never be invoked with a length that is unknown.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -61,6 +65,27 @@ public static HadoopInputFile fromPath(Path path, Configuration conf) throws IOE
     return new HadoopInputFile(fs, fs.getFileStatus(path), conf);
   }
 
+  /**
+   * Creates an input file using a caller-supplied file length.
+   *
+   * <p>The length is trusted and no file status lookup is performed. Callers must provide the exact
+   * length of the file that will be opened. The file is not validated until it is opened.
```
steveloughran: declare that the existence of the file may not be checked until the first read, and the length may be trusted; if it is wrong then reads may fail.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestHadoopOpenFile.java
```diff
@@ -101,10 +102,41 @@ public void testOpenFileGoodPath() throws Throwable {
...
+
     // The fallback call of open(path) never took place.
     Mockito.verify(mockFS, never()).open(path);
   }
 
+  /**
+   * A caller-supplied length avoids a status lookup and is passed to openFile().
+   */
+  @Test
+  public void testOpenFileWithKnownLength() throws Throwable {
```
steveloughran, line 120: add a test opening a nonexistent file with a declared length. For cloud storage, it may not fail until the first read(); for local fs it will fail immediately. In both cases it must fail with FileNotFoundException.

steveloughran (changes requested):
commented, suggested one more test and a change to the javadocs; no production code changes

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
dossett: Yes, does that suggest another change?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/util/TestHadoopOpenFile.java
dossett: Added a test for local fs, it's not clear how to test the cloud store behavior strictly with a unit test.  Neither GCS nor S3 provides a `MiniDFSCluster`-type tool as best I can tell

divjotarora (commented):
Do callers lose any functionality if they use the new length-based constructor? Codex claims you lose etag validation in this case because we don't have a full file status and don't call `withFileStatus()` when opening a stream. The Hadoop S3 [code](https://github.com/apache/hadoop/blob/rel/release-3.3.5/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/impl/OpenFileSupport.java#L217-L246) seems to confirm that. Is this a concern? Should we at least document it?

dossett:
@divjotarora Thank you, that's a good catch.  It seems like a tradeoff we can allow callers to make, I added some explicit documentation to that effect.

dossett:
I believe the CI failure is [a bug in japicmp](https://github.com/siom79/japicmp/issues/526). A class had only a private constructor and I added a second private constructor.  japicmp incorrectly concluded that the class became non-extendable.

I added a japicmp exclusion that can later be removed.

dossett:
My PR to japicmp was approved, confirming the false positive failure above

Fokko:
@dossett I wasn't aware that this was blocked by https://github.com/apache/parquet-java/pull/3764

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -109,20 +149,24 @@ public long getLength() {
...
     FSDataInputStream stream;
     try {
-      // this method is async so that implementations may do async HEAD head
+      // this method is async so that implementations may do async HEAD
       // requests, such as S3A/ABFS when a file status is passed down.
-      final CompletableFuture<FSDataInputStream> future = fs.openFile(stat.getPath())
-          .withFileStatus(stat)
-          .opt(OPENFILE_READ_POLICY_KEY, PARQUET_READ_POLICY)
-          .build();
+      FutureDataInputStreamBuilder builder = fs.openFile(path).opt(OPENFILE_READ_POLICY_KEY, PARQUET_READ_POLICY);
```
Fokko, line 154: The [`FutureDataInputStreamBuilder`](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FutureDataInputStreamBuilder.html) is marked as unstable. I think we might want to pick @steveloughran brain here :)

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
```diff
@@ -76,28 +106,38 @@ public static HadoopInputFile fromStatus(FileStatus stat, Configuration conf) th
 
   private HadoopInputFile(FileSystem fs, FileStatus stat, Configuration conf) {
     this.fs = fs;
+    this.path = stat.getPath();
+    this.length = stat.getLen();
```
Fokko, line 110: Wouldn't this potentially already trigger a HEAD operation?

dossett: I don't believe so, this is the constructor for the code path that's NOT using the length file hint and this is just extracting these values fro the already existing `FileStatus` object passed in.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/util/HadoopInputFile.java
dossett: I agree and FWIW this PR doesn't introduce the use of this class.  `fs.openFile(path)` returns an object of this type, which we're doing on `master` we just immediately make the next calls to `withFileStatus`, `opt`, and `build` so it never has to be declared.

dossett:
Thank you for the comments @Fokko 

re: #3764, yes I coded in anger when this PR led me to find that bug :-) I've rebased against master and removed the temporary exclusion from this PRs POM.

PR #3680, opened 2026-07-21
GH-3679: Allow TimestampType to annotate FLBA(12)

### LINKED ISSUE #3679: Support extended precision for nanosecond timestamps

Implement the proposal in https://github.com/apache/parquet-format/issues/600 and the linked [proposal document](https://docs.google.com/document/d/1H43RQZhWKcg9c4tJO5W87YxhbUxN7qGSFxJBrrwWOfY/edit?tab=t.0#heading=h.ccmrsmcbig6b) to allow the `TimestampType` logical type to annotate `FIXED_LEN_BYTE_ARRAY` physical type with `type_length = 12`.

### Component(s)

Core

divjotarora:
### Rationale for this change
See parquet-format issue https://github.com/apache/parquet-format/issues/600 and the linked proposal document for rationale

This PR implements support for the `TimestampType` logical type annotation on `FIXED_LEN_BYTE_ARRAY` physical type values with `type_length=12` (96 bits).

### Are these changes tested?
Yes, several test files are updated along with the source.

Users will be able to declare schemas with this physical/logical type combination. No API changes.

Closes #3679

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
@@ -20,6 +20,7 @@ package org.apache.parquet.schema;
 
 import java.io.Serializable;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Comparator;
 import org.apache.parquet.io.api.Binary;
 
@@ -288,29 +289,16 @@ public abstract class PrimitiveComparator<T> implements Comparator<T>, Serializa
         throw new IllegalArgumentException(
             "Timestamp binary length must be 12 bytes, got " + b1.length() + " and " + b2.length());
       }
-
-      ByteBuffer bb1 = b1.toByteBuffer();
-      ByteBuffer bb2 = b2.toByteBuffer();
-      // The buffers may be slices with a non-zero position (e.g. a ByteBuffer-backed Binary), so
-      // index relative to position() rather than absolute offset 0. Byte 11 (position + 11) is the
-      // most significant byte and carries the sign; byte 0 (position) is least significant.
-      int p1 = bb1.position();
-      int p2 = bb2.position();
-
-      // If one value is negative and the other is positive, one is trivially greater.
-      boolean neg1 = bb1.get(p1 + 11) < 0;
-      boolean neg2 = bb2.get(p2 + 11) < 0;
-      if (neg1 != neg2) {
-        return neg1 ? -1 : 1;
+      ByteBuffer bb1 = b1.toByteBuffer().slice();
+      bb1.order(ByteOrder.LITTLE_ENDIAN);
+      ByteBuffer bb2 = b2.toByteBuffer().slice();
+      bb2.order(ByteOrder.LITTLE_ENDIAN);
+      // Signed comparison of the high 4 bytes followed by unsigned comparison of the low 8 bytes.
+      int hiResult = Integer.compare(bb1.getInt(8), bb2.getInt(8));
+      if (hiResult != 0) {
+        return hiResult;
       }
-
-      for (int i = 11; i >= 0; --i) {
-        int diff = toUnsigned(bb1.get(p1 + i)) - toUnsigned(bb2.get(p2 + i));
-        if (diff != 0) {
-          return diff;
-        }
-      }
-      return 0;
+      return Long.compareUnsigned(bb1.getLong(0), bb2.getLong(0));
     }
 
     @Override
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
@@ -279,6 +279,47 @@ public abstract class PrimitiveComparator<T> implements Comparator<T>, Serializa
   };
 
   /**
+   * Comparator for timestamps encoded as 12-byte little-endian two's-complement values.
+   */
+  static final PrimitiveComparator<Binary> BINARY_AS_SIGNED_TIMESTAMP_COMPARATOR = new BinaryComparator() {
+    @Override
+    int compareBinary(Binary b1, Binary b2) {
+      if (b1.length() != 12 || b2.length() != 12) {
+        throw new IllegalArgumentException(
+            "Timestamp binary length must be 12 bytes, got " + b1.length() + " and " + b2.length());
+      }
+
+      ByteBuffer bb1 = b1.toByteBuffer();
+      ByteBuffer bb2 = b2.toByteBuffer();
+      // The buffers may be slices with a non-zero position (e.g. a ByteBuffer-backed Binary), so
+      // index relative to position() rather than absolute offset 0. Byte 11 (position + 11) is the
+      // most significant byte and carries the sign; byte 0 (position) is least significant.
+      int p1 = bb1.position();
+      int p2 = bb2.position();
+
+      // If one value is negative and the other is positive, one is trivially greater.
+      boolean neg1 = bb1.get(p1 + 11) < 0;
+      boolean neg2 = bb2.get(p2 + 11) < 0;
+      if (neg1 != neg2) {
+        return neg1 ? -1 : 1;
+      }
+
+      for (int i = 11; i >= 0; --i) {
+        int diff = toUnsigned(bb1.get(p1 + i)) - toUnsigned(bb2.get(p2 + i));
+        if (diff != 0) {
+          return diff;
+        }
+      }
+      return 0;
+    }
+
+    @Override
+    public String toString() {
+      return "BINARY_AS_SIGNED_TIMESTAMP_COMPARATOR";
+    }
+  };
+
+  /**
    * This comparator is for comparing two float16 values represented in 2 bytes binary.
    */
   static final PrimitiveComparator<Binary> BINARY_AS_FLOAT16_COMPARATOR = new BinaryComparator() {
--- a/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
@@ -29,6 +29,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
@@ -249,7 +250,7 @@ public abstract class PrimitiveStringifier {
       return toFormattedString(getInstant(value));
     }
 
-    private String toFormattedString(Instant instant) {
+    final String toFormattedString(Instant instant) {
       return formatter.format(instant);
     }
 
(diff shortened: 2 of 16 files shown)


stevomitric (approved):
Should the `converted_type` be suppressed for the FLBA(12) carrier, the same way NANOS already returns empty? Those legacy types are defined as INT64-only, so emitting them here gives a contradictory footer.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -563,6 +563,9 @@ public Optional<Boolean> visit(
               @Override
               public Optional<Boolean> visit(
                   LogicalTypeAnnotation.TimestampLogicalTypeAnnotation timestampLogicalType) {
+                if (primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
+                  return checkFixedPrimitiveType(12, timestampLogicalType);
+                }
                 return checkInt64PrimitiveType(timestampLogicalType);
```
stevomitric: Will this throw an error saying timestamps only support int64 physical type - i.e. when we pass in int32? Should we extend the error message to include FLBA as well?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveComparator.java
```diff
@@ -278,6 +278,47 @@ public String toString() {
...
+      int p2 = bb2.position();
+
+      // If one value is negative and the other is positive, one is trivially greater.
+      boolean neg1 = bb1.get(p1 + 11) < 0;
+      boolean neg2 = bb2.get(p2 + 11) < 0;
+      if (neg1 != neg2) {
+        return neg1 ? -1 : 1;
+      }
+
+      for (int i = 11; i >= 0; --i) {
```
emkornfield: Can this be done more efficiently by retrieving int/long values from the underlying type and comparing those?

divjotarora: Good idea, changed to signed comparison of higher 4 bytes and unsigned comparison of lower 8 bytes

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: Nice catch, fixed to inline the `Preconditions` check against both INT64 and FLBA(12) combined rather than using `checkInt64PrimitiveType`

divjotarora (commented):
> Should the `converted_type` be suppressed for the FLBA(12) carrier, the same way NANOS already returns empty? Those legacy types are defined as INT64-only, so emitting them here gives a contradictory footer.

Thanks for catching @stevomitric, the existing tests didn't catch this because they were focusing on using FLBA(12) with NANOS and it doesn't have a converted type already. The code is now fixed and there are more tests to validate FLBA(12) with MILLIS and MICROS as well.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -274,56 +310,96 @@ Instant getInstant(int value) {
     ;
   };
 
+  // Converts a count of time units since epoch, held as a BigInteger (the 96-bit FLBA(12) carrier),
+  // into an Instant. The 96-bit count does not fit in a long, but the whole-second count does, so
```
emkornfield: doesn't this depend on TimeUnit?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -266,6 +267,41 @@ Instant getInstant(long value) {
...
+      }
+      try {
+        BigInteger units = new BigInteger(bigEndian);
+        try {
+          return toFormattedString(getInstant(units));
+        } catch (ArithmeticException | DateTimeException e) {
+          return units.toString();
+        }
+      } catch (NumberFormatException e) {
+        return BINARY_INVALID;
```
emkornfield: what cases does this throw?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
```diff
@@ -266,6 +267,41 @@ Instant getInstant(long value) {
...
+    private TimestampStringifier(String name, String format) {
+      super(name, format);
+    }
+
+    @Override
+    public String stringify(Binary value) {
+      if (value == null) {
+        return BINARY_NULL;
+      }
+      byte[] littleEndian = value.getBytesUnsafe();
```
emkornfield: consider refactoring to toBigEndian method.  I think there might already be some byte utils someplace.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
```diff
@@ -91,6 +93,17 @@ public void testContractNonStringTypes() {
     testTruncator(Types.required(INT96).named("test_int96"), false);
   }
 
+  @Test
+  public void testFlba12Timestamp() {
+    BinaryTruncator truncator = BinaryTruncator.getTruncator(Types.required(FIXED_LEN_BYTE_ARRAY)
+        .length(12)
+        .as(LogicalTypeAnnotation.timestampType(true, TimeUnit.NANOS))
+        .named("test_fixed_timestamp"));
+    Binary value = Binary.fromConstantByteArray(new byte[] {0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4});
```
emkornfield: why are only the last 4 bytes non-zero?

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -417,6 +446,18 @@ private Binary toBinary(int... bytes) {
     return Binary.fromConstantByteArray(array);
   }
 
+  // Encodes a unit count as the extended-precision timestamp carrier: a 12-byte signed
```
emkornfield: don't we need tests for out of range values?

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
```diff
@@ -417,6 +446,18 @@ private Binary toBinary(int... bytes) {
...
+  // two's-complement little-endian value (sign-extended, so negatives fill the high bytes with
+  // 0xFF).
+  private Binary flba12(long units) {
+    byte[] le = new byte[12];
```
emkornfield: nit: might be clearer if we can use BigInteger to byte array and flip the bytes?

emkornfield (changes requested):
One thing that I think we want to at least look at is if we build an FLBA(12) will common wrappers (e.g. avro) correctly write the correct bytes to the field or will they fail?

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
divjotarora: Good catch, changed the comment to

> Converts a count of time units since epoch, held as a BigInteger (the 96-bit FLBA(12) carrier),
   into an Instant. Throws ArithmeticException if the sections portion overflows long.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
divjotarora: The `BigInteger` ctor throws if the array is empty, but that's not possible as it's guaranteed to be size 12 higher up. Removed the try/catch

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveStringifier.java
divjotarora: Added a general `reverse()` function to `BytesUtils`, this can be used for bidirectional LE <--> BE swaps, but let me know if you'd prefer `toBigEndian` specifically.

parquet-column/src/test/java/org/apache/parquet/internal/column/columnindex/TestBinaryTruncator.java
divjotarora: No specific reason. I changed it to `bytes[i] = i` to make it look less intentional.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
divjotarora: It's about the same amount of code, but probably conceptually easier to read. Done.

parquet-column/src/test/java/org/apache/parquet/schema/TestPrimitiveStringifier.java
divjotarora: Added an override that directly accepts a `BigInteger` + tests that use it for values larger than the int64 range.

divjotarora (commented):
> Should the `converted_type` be suppressed for the FLBA(12) carrier, the same way NANOS already returns empty? Those legacy types are defined as INT64-only, so emitting them here gives a contradictory footer.

Thanks for catching @stevomitric, the existing tests didn't catch this because they were focusing on using FLBA(12) with NANOS and it doesn't have a converted type already. The code is now fixed and there are more tests to validate FLBA(12) with MILLIS and MICROS as well.

emkornfield (approved):

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInterOpReadFlba12Timestamp.java
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.parquet.schema.LogicalTypeAnnotation;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.PrimitiveStringifier;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
+import org.junit.jupiter.api.Test;
+
+public class TestInterOpReadFlba12Timestamp {
+  private static final String FILE = "flba12_timestamp.parquet";
+  // TODO: update after parquet-testing PR merges
```
Fokko: https://github.com/apache/parquet-testing/pull/123 has been merged :)

Fokko, line 43: I think you ment https://github.com/apache/parquet-testing/pull/123

Fokko (approved):
@divjotarora Can you fix the merge conflicts?

divjotarora:
> @divjotarora Can you fix the merge conflicts?

Updated the PR, should be clean now @Fokko

Fokko:
@divjotarora We need to move to JUnit5:

```
[ERROR] Banned imports detected in TEST code:
[ERROR] 
[ERROR] Reason: Use JUnit 5 (org.junit.jupiter.*) and AssertJ instead of JUnit 4
[ERROR] 	in file:///home/runner/work/parquet-java/parquet-java/parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestInputFormatFileStatusReuse.java
[ERROR] 		static org.junit.Assert.assertEquals 	(Line: 20, Matched by: org.junit.**)
[ERROR] 		static org.junit.Assert.assertSame   	(Line: 21, Matched by: org.junit.**)
[ERROR] 		static org.junit.Assert.assertTrue   	(Line: 22, Matched by: org.junit.**)
[ERROR] 		org.junit.Rule                       	(Line: 50, Matched by: org.junit.**)
[ERROR] 		org.junit.Test                       	(Line: 51, Matched by: org.junit.**)
[ERROR] 		org.junit.rules.TemporaryFolder      	(Line: 52, Matched by: org.junit.**)
[ERROR] 
[ERROR] Analysis of 257 files took less than 1 second
[ERROR] -> [Help 1]
```

divjotarora:
> @divjotarora We need to move to JUnit5:

@Fokko The `TestInputFormatFileStatusReuse.java` file is not modified by this PR. It seems spotless checks are failing due to https://github.com/apache/parquet-java/pull/3674 being merged. I've opened https://github.com/apache/parquet-java/pull/3813 to fix this on the master branch, PTAL

divjotarora:
@Fokko I've rebased this PR on top of the merged https://github.com/apache/parquet-java/pull/3813, should be good to go now

PR #3688, opened 2026-07-26
Revert "GH-3574: Statistics.toParquetStatistics always set null_count (#3575)"

Fokko:
This reverts commit 63aebcc0cf3684ba6c9dc2b7243c6007d3a1269f.


kevinjqliu (approved):
LGTM 

The change is valid based on `parquet-format`, which [recommends writing null_count even when it’s zero](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L286-L287), but it doesn’t preserve backward compatibility and can cause problems for downstream consumers like Iceberg.

We should revert for now until we find a safer way to introduce this change.

wgtmac (approved):
I agree we need to revert that PR to unblock the release. But I do think that fix is valid in essence.

kevinjqliu:
lets reopen https://github.com/apache/parquet-java/issues/3574 to track this

Fokko:
I've re-opened the issue, and I agree @wgtmac, the change on its own makes sense, but currently we implicitly have either all the stats, or no stats at all. So we should evolve that contract 👍

PR #3698, opened 2026-07-30
GH-3697: Reuse hadoop file status and footer in ParquetRecordReader

### LINKED ISSUE #3697: Reuse hadoop file status and footer in ParquetRecordReader

This is actually [PARQUET-2415](https://issues.apache.org/jira/browse/PARQUET-2415)
_Reuse hadoop file status and footer in ParquetRecordReader_ moved to a github issue.

That has a stale pr #1242 by @wankunde; I've got claude to rebase it and update to junit5/assertj

### Component(s)

Core

steveloughran:
Contains contributions by Claude.ai (rebasing; junit5/assertj migration)

Contributed by @wankunde; made current by claude.

### Rationale for this change

HTTP HEAD is expensive, slow and comes out of your quota of a few thousand reads/second

The InputFile used when reading a file footer is cached in the input split, so all splits read in the same process *only* will use that and any cached FileStatus when opening the file. This saves a HEAD against s3, azure and gcs on every split read within that process.

### Are these changes tested?

new test, otherwise all internal.

no

Closes #3697

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
@@ -18,6 +18,7 @@
  */
 package org.apache.parquet.hadoop;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.DataInput;
@@ -36,6 +37,7 @@ import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 import org.apache.parquet.hadoop.metadata.BlockMetaData;
 import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData;
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.MessageTypeParser;
 
@@ -54,6 +56,13 @@ public class ParquetInputSplit extends FileSplit implements Writable {
 
   private long end;
   private long[] rowGroupOffsets;
+
+  /**
+   * Footer of the file, if the split was built by a caller which had already read it.
+   * Not written by {@link #write(DataOutput)}, so it is only visible within the JVM which set it.
+   */
+  @JsonIgnore
+  private volatile ParquetMetadata footer;
 
   /**
    * Writables must have a parameterless constructor
@@ -220,6 +229,24 @@ public class ParquetInputSplit extends FileSplit implements Writable {
    */
   public long[] getRowGroupOffsets() {
     return rowGroupOffsets;
+  }
+
+  /**
+   * @return the footer of the file, if it was passed in by whoever built the split, else null.
+   */
+  public ParquetMetadata getFooter() {
+    return footer;
+  }
+
+  /**
+   * Pass in a footer already read from the file, so that a reader created from this split does not have to read it
+   * again. As the footer is not serialized with the split, this only has an effect on readers created in the same
+   * JVM.
+   *
+   * @param footer footer of the file this split refers to
+   */
+  public void setFooter(ParquetMetadata footer) {
+    this.footer = footer;
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


wankunde:
AFAK, this changed has been ported into several cloud products. 
@wgtmac @amousavigourabi @gszadovszky 
Could you take a look at this PR ?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -222,6 +231,24 @@ public long[] getRowGroupOffsets() {
...
+
+  /**
+   * Pass in a footer already read from the file, so that a reader created from this split does not have to read it
+   * again. As the footer is not serialized with the split, this only has an effect on readers created in the same
+   * JVM.
+   *
+   * @param footer footer of the file this split refers to
+   */
+  public void setFooter(ParquetMetadata footer) {
+    this.footer = footer;
```
wgtmac, line 249: `footer` is omitted from `write()`, but `readFields()` never clears it. Hadoop may reuse Writable instances, so a split can retain the footer from another file and `ParquetRecordReader` may open the wrong `InputFile`. Please clear `footer` in `readFields()` and cover object reuse in the round-trip test.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetRecordReader.java
```diff
@@ -154,9 +156,16 @@ private void initializeInternalReader(ParquetInputSplit split, Configuration con
...
-        ParquetFileReader.open(HadoopInputFile.fromPath(path, configuration), optionsBuilder.build());
+    // open a reader with the metadata filter, reusing the footer of the split and the file it was read from
+    // when the split was built by a caller which had already read them
+    ParquetMetadata footer = split.getFooter();
+    InputFile inputFile = footer != null && footer.getInputFile() != null
+        ? footer.getInputFile()
+        : HadoopInputFile.fromPath(path, configuration);
+    ParquetReadOptions options = optionsBuilder.build();
+    ParquetFileReader reader = footer != null
+        ? ParquetFileReader.open(inputFile, footer, options, inputFile.newStream())
```
wgtmac, line 167: Readers opened from the same encrypted footer share its mutable `InternalFileDecryptor` and JCE `Cipher` state, so concurrent splits in one JVM can race. Please use reader-local decryption state or skip footer reuse for encrypted files, and add a concurrent encrypted test.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
```diff
@@ -55,6 +57,13 @@ public class ParquetInputSplit extends FileSplit implements Writable {
   private long end;
   private long[] rowGroupOffsets;
 
+  /**
+   * Footer of the file, if the split was built by a caller which had already read it.
+   * Not written by {@link #write(DataOutput)}, so it is only visible within the JVM which set it.
+   */
+  @JsonIgnore
```
wgtmac, line 64: This `JsonIgnore` is relocated in the published jar and is invisible to an external `ObjectMapper`, as the new CLI mix-in notes. The current test runs before shading. Please use a serialization-neutral exclusion and test the shaded artifact; `ParquetMetadata.inputFile` has the same issue.

wgtmac (commented):
These review comments are generated by Codex. I've just checked them and all of these seem non-blocking to me. Let me know what you think @steveloughran @wankunde.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
steveloughran: Good one. Needs fix.

steveloughran:
I will try to fix but can't set a timeline. If you haven't heard I left cloudera on Friday, I do want to wrap up the recent PRs but I need to set up my mac mini as a dev box first, with the pi5 rebuilt as the docker server. That's where the hadoop 3.4.3 jars on maven central came from, FWIW

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetInputSplit.java
steveloughran: @wgtmac what do you suggest as a good strategy here? the xml ones have their own issues ( https://github.com/FasterXML/jackson-modules-base ); otherwise I'd use the javax.xml.bind ones.

wgtmac: Unfortunately I don't have a good answer even after consulting AI. It seems that there are already similar issues before this PR so I think this is not blocking?

Do you have any suggestion? @gszadovszky @Fokko

gszadovszky: I agree this should not be blocking. We should create a separate issue for this. I do not have a good solution either.

PR #3699, opened 2026-07-31
PARQUET-2249: Write IEEE 754 total order by default for floating-point columns

Jiayi-Wang-db:
### Rationale for this change

Under the new writer behavior, a floating-point column containing NaN gets finite min/max computed over the non-NaN values, together with a `nan_count`. A reader that predates `nan_count` ignores it, accepts the finite bounds, and can incorrectly prune row groups that actually contain NaN. e.g. [1.0, NaN] is written as min = max = 1.0, so an old reader drops the row group for greater(x, 1.0) even though NaN matches.

By contrast, readers ignore statistics written under an unknown sort order. Writing IEEE_754_TOTAL_ORDER by default is therefore the safer, forward-compatible behavior: older readers simply decline to use the stats they can't interpret rather than misinterpreting them. 

FLOAT, DOUBLE and FLOAT16 columns built without an explicit column order now default to `IEEE_754_TOTAL_ORDER`

### Are these changes tested?
yes

Yes. Floating-point columns are now written with IEEE_754_TOTAL_ORDER by default instead of TYPE_DEFINED_ORDER.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
@@ -227,7 +227,7 @@ public class MessageTypeParser {
     }
     if (t.equalsIgnoreCase(PrimitiveType.COLUMN_ORDER_KEYWORD)) {
       check(st.nextToken(), "(", "column order followed by (", st);
-      childBuilder.columnOrder(parseColumnOrder(st.nextToken(), st));
+      childBuilder.columnOrder(parseColumnOrder(st.nextToken()));
       check(st.nextToken(), ")", "column order ended by )", st);
       t = st.nextToken();
     }
@@ -246,24 +246,18 @@ public class MessageTypeParser {
     }
   }
 
-  private static ColumnOrder parseColumnOrder(String t, Tokenizer st) {
-    ColumnOrder.ColumnOrderName name;
-    try {
-      name = ColumnOrder.ColumnOrderName.valueOf(t.toUpperCase(Locale.ENGLISH));
-    } catch (IllegalArgumentException e) {
-      throw new IllegalArgumentException("Unknown column order: " + t + " at " + st.getLocationString(), e);
+  private static ColumnOrder parseColumnOrder(String t) {
+    // An unrecognized order degrades to UNDEFINED rather than failing, matching
+    // ParquetMetadataConverter.fromParquetColumnOrder, so a schema written by a newer API with an
+    // order this version does not know stays parseable. Statistics under an unknown order are
+    // ignored by readers anyway.
+    if (t.equalsIgnoreCase(ColumnOrder.ColumnOrderName.TYPE_DEFINED_ORDER.name())) {
+      return ColumnOrder.typeDefined();
     }
-    switch (name) {
-      case UNDEFINED:
-        return ColumnOrder.undefined();
-      case TYPE_DEFINED_ORDER:
-        return ColumnOrder.typeDefined();
-      case IEEE_754_TOTAL_ORDER:
-        return ColumnOrder.ieee754TotalOrder();
-      default:
-        throw new IllegalArgumentException(
-            "Unsupported column order: " + name + " at " + st.getLocationString());
+    if (t.equalsIgnoreCase(ColumnOrder.ColumnOrderName.IEEE_754_TOTAL_ORDER.name())) {
+      return ColumnOrder.ieee754TotalOrder();
     }
+    return ColumnOrder.undefined();
   }
 
   private static boolean isLogicalType(String t) {
--- a/parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
+++ b/parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
@@ -225,6 +225,12 @@ public class MessageTypeParser {
       check(t, ")", "logical type ended by )", st);
       t = st.nextToken();
     }
+    if (t.equalsIgnoreCase(PrimitiveType.COLUMN_ORDER_KEYWORD)) {
+      check(st.nextToken(), "(", "column order followed by (", st);
+      childBuilder.columnOrder(parseColumnOrder(st.nextToken(), st));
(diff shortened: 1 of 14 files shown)


parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
```diff
@@ -648,6 +662,20 @@ public PrimitiveType(
...
   }
 
+  /**
+   * Whether a {@code FLOAT} or {@code DOUBLE} column with no explicit column order should default to
+   * IEEE 754 total order. This is skipped when a logical type annotation is present but does not
+   * accept IEEE 754 total order (e.g. an unknown annotation), so the type stays constructible with
+   * the type-defined fallback.
+   */
+  private boolean defaultsToIeee754TotalOrder(PrimitiveTypeName primitive) {
+    if (primitive != PrimitiveTypeName.FLOAT && primitive != PrimitiveTypeName.DOUBLE) {
```
wgtmac: Is this correct? FLOAT16 has FLBA primitive type which will be rejected here. Should we pass LogicalTypeAnnotation explicitly instead of calling `getLogicalTypeAnnotation()` so these logics are in a single place?

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
```diff
@@ -775,7 +776,9 @@ private void testBuilder(PrimitiveType type, Object min, byte[] minBytes, Object
 
   @Test
   public void testSpecBuilderForFloat() {
-    PrimitiveType type = Types.required(FLOAT).named("test_float");
+    // NaN/+-0 read semantics are specific to type-defined order (float now defaults to IEEE 754).
```
wgtmac: There are so many similar verbose comments. Should we just remove them? I believe it is clear when test cases have set explicit column orders.

parquet-column/src/main/java/org/apache/parquet/schema/PrimitiveType.java
Jiayi-Wang-db: Thanks for catching this. Refactored into a single static defaultColumnOrder(primitive, originalType, logicalTypeAnnotation) that recognizes FLOAT16 .

parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java
Jiayi-Wang-db: Removed.

wgtmac (approved):
LGTM, thanks!

wgtmac:
@Fokko @gszadovszky Do you want to take a look?

parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
```diff
@@ -240,6 +246,26 @@ private static void addPrimitiveType(
...
+    switch (name) {
+      case UNDEFINED:
+        return ColumnOrder.undefined();
+      case TYPE_DEFINED_ORDER:
+        return ColumnOrder.typeDefined();
+      case IEEE_754_TOTAL_ORDER:
+        return ColumnOrder.ieee754TotalOrder();
+      default:
+        throw new IllegalArgumentException(
+            "Unsupported column order: " + name + " at " + st.getLocationString());
```
Fokko: Should we throw here, or emit a null/unknown `ColumnOrder`?

Fokko (approved):
Based on the dev-list discussion thread, this makes sense to me 👍

parquet-column/src/main/java/org/apache/parquet/schema/MessageTypeParser.java
Jiayi-Wang-db: Good point, I think it should be degraded to Undefined, and not throw explicit exception. Made the change.

gszadovszky:
It looks good to me. There is however a potential follow up issue from codex:

>[P2] Keep mixed-version floating schemas mergeable
>
>PrimitiveType.java:661-667 makes new floating schemas default to IEEE_754_TOTAL_ORDER, while ParquetMetadataConverter.java:2064-2070 reads legacy footers as TYPE_DEFINED_ORDER. PrimitiveType.java:910-913 rejects those differing orders during strict schema union, and ParquetInputFormat.java:378-385 uses strict global-metadata merging for client-side split planning. Consequently, metadata aggregation over otherwise identical pre- and post-upgrade files can fail with IncompatibleSchemaModificationException. Please add a mixed-footer regression test and reconcile differing floating orders conservatively during aggregation, for example as UNDEFINED.

Jiayi-Wang-db:
> It looks good to me. There is however a potential follow up issue from codex:
> 
> > [P2] Keep mixed-version floating schemas mergeable
> > PrimitiveType.java:661-667 makes new floating schemas default to IEEE_754_TOTAL_ORDER, while ParquetMetadataConverter.java:2064-2070 reads legacy footers as TYPE_DEFINED_ORDER. PrimitiveType.java:910-913 rejects those differing orders during strict schema union, and ParquetInputFormat.java:378-385 uses strict global-metadata merging for client-side split planning. Consequently, metadata aggregation over otherwise identical pre- and post-upgrade files can fail with IncompatibleSchemaModificationException. Please add a mixed-footer regression test and reconcile differing floating orders conservatively during aggregation, for example as UNDEFINED.


Hi Gobor, thanks for flagging this. It makes sense to me that we should use the `UNDEFINED` column order in the merged schema when there is a conflict instead of throwing exception.
AFAIK, when reading files, the reader ignores the column order in the merged schema and uses the column order from each file’s footer instead. So I don’t see any issue with changing it to `UNDEFINED`.
cc @wgtmac @Fokko , I’d appreciate another pair of eyes on this.

Jiayi-Wang-db:
Could anyone help to merge this PR? Thanks!

wgtmac:
Just noticed that `PrimitiveType.union()` is also used by the deprecated `_metadata` writing APIs. Since `_metadata` is deprecated, I don't think this needs to block the PR.

Fokko:
Let's move this forward and let me cut another RC. Thanks everyone!

Issue #3696, opened 2026-07-29
Cache ParsedVersion in FileMetaData to eliminate redundant created_by parsing


asif-moh:
## Summary

`VersionParser.parse(createdBy)` is called from 7 distinct production sites, all parsing the same constant string from `FileMetaData.getCreatedBy()`. Since `FileMetaData` is constructed once per file and already stores the `createdBy` string, it is the natural place to parse once and cache.

## Problem

The `created_by` string is re-parsed into a `ParsedVersion` at every call site independently:

| # | Call site | Frequency per file | When |
|---|-----------|-------------------|------|
| 1 | `CorruptStatistics.shouldIgnoreStatistics` via `buildColumnChunkMetaData` | R × C | Footer decode |
| 2 | `CorruptStatistics.shouldIgnoreStatistics` via `ParquetFileReader.readAllPages` | Pages per column chunk | Page read |
| 3 | `CorruptStatistics.shouldIgnoreStatistics` via `ParquetRewriter.convertStatistics` | Pages per rewritten chunk | Rewrite |
| 4 | `CorruptStatistics.shouldIgnoreStatistics` via `EncryptedColumnChunkMetaData.decryptIfNeeded` | 1 per encrypted column | Lazy decrypt |
| 5 | `CorruptDeltaByteArrays.requiresSequentialReads(String, Encoding)` in `ParquetRecordReader` | 1 per reader init | Reader init |
| 6 | `ColumnReadStoreImpl` constructor via `MessageColumnIO.getRecordReader` | R (once per row group read) | Row group materialization |
| 7 | `ColumnReadStoreImpl` constructor via `ParquetRewriter.nullifyColumn` | 1 per nullified column | Rewrite with nullification |

`ColumnReadStoreImpl` already parses `createdBy` into a `ParsedVersion` and stores it as a field — but does so R times (once per row group) because nobody upstream caches the parsed result.

## Proposed Change

Add a cached `ParsedVersion` field to `FileMetaData`:

```java
// In FileMetaData:
private final transient ParsedVersion writerVersion;

public FileMetaData(...) {
    ...
    this.createdBy = createdBy;
    this.writerVersion = parseVersion(createdBy);
    ...
}

public ParsedVersion getWriterVersion() {
    return writerVersion;
}

private static ParsedVersion parseVersion(String createdBy) {
    if (Strings.isNullOrEmpty(createdBy)) return null;
    try {
        return VersionParser.parse(createdBy);
[...]

PR #3700, opened 2026-08-01
GH-3696: Cache ParsedVersion in FileMetaData and use it in fromParquetMetadata

### LINKED ISSUE #3601: Optimize repeated shouldIgnoreStatistics calls during footer reading in ParquetMetadataConverter

asifsmohammed:
### Rationale for this change

`VersionParser.parse(createdBy)` is called from 7 production sites, all parsing the same constant string from `FileMetaData.getCreatedBy()`. In `fromParquetMetadata`, this happens R×C times (once per column per row group) during footer metadata conversion. Since `FileMetaData` is constructed once per file and already stores the `createdBy` string, it is the natural place to parse once and cache the result.

This PR caches the parsed version and migrates the first (and hottest) call site — `ParquetMetadataConverter.fromParquetMetadata` — to use the cache, eliminating redundant `VersionParser.parse` and `SemanticVersion.parse` calls from the R×C inner loop.

Fixes the 1st item in this issue https://github.com/apache/parquet-java/issues/3696

- Add `getWriterVersion()` to `FileMetaData` with lazy-init via an immutable `WriterVersionResult` holder (thread-safe, double-checked locking)
- Retain the semantic-version parse failure in `ParsedVersion` so cached callers can reuse the original exception without reparsing
- Add `shouldIgnoreStatistics(ParsedVersion, String, PrimitiveTypeName)` overload to `CorruptStatistics` that uses the cached `SemanticVersion` directly and keeps the original `createdBy` string for logging
- Refactor `ParquetMetadataConverter.fromParquetMetadata` to construct `FileMetaData` before the row-group loop and use the cached `ParsedVersion` in `buildColumnChunkMetaData`
- Share the statistics conversion implementation between the existing String path and the cached `ParsedVersion` path
- Fall back to the String-based path when `getWriterVersion()` throws `VersionParseException` to preserve the existing logging behavior

### Are these changes tested?

Yes.

- `FileMetaDataTest` — 6 tests covering valid, null, empty, unparseable version strings, and caching
- `CorruptStatisticsTest.testShouldIgnoreStatisticsWithParsedVersion` — covers all branches of the new `ParsedVersion` overload, including null, non-parquet-mr, empty version, invalid semver, retained parse failure, corrupt, and fixed versions
- `TestParquetMetadataConverter.testV2StatsDoNotTriggerCorruptStatisticsCheck` — verifies that V2 min/max statistics bypass the corrupt-statistics check

No breaking changes. Adds new public methods:

- `FileMetaData.getWriterVersion()` — returns cached `ParsedVersion`, throws `VersionParseException` for unparseable strings
- `CorruptStatistics.shouldIgnoreStatistics(ParsedVersion, String, PrimitiveTypeName)` — for callers that already have a parsed version; the String parameter preserves the original `createdBy` value for logging
- `ParquetMetadataConverter.fromParquetStatistics(ParsedVersion, String, Statistics, PrimitiveType)` — converts statistics using a cached writer version
- `ParquetMetadataConverter.buildColumnChunkMetaData(ColumnMetaData, ColumnPath, PrimitiveType, ParsedVersion, String)` — builds column metadata using a cached writer version

Closes #3601

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -19,6 +19,7 @@
 package org.apache.parquet;
 
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.parquet.SemanticVersion.SemanticVersionParseException;
 import org.apache.parquet.VersionParser.ParsedVersion;
 import org.apache.parquet.VersionParser.VersionParseException;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
@@ -69,8 +70,10 @@ public class CorruptStatistics {
 
     try {
       ParsedVersion version = VersionParser.parse(createdBy);
-      return shouldIgnoreStatistics(version, columnType);
+      return shouldIgnoreStatistics(version, createdBy, columnType);
     } catch (RuntimeException | VersionParseException e) {
+      // couldn't parse the created_by field, log what went wrong, don't trust the
+      // stats, but don't make this fatal.
       warnParseErrorOnce(createdBy, e);
       return true;
     }
@@ -78,13 +81,15 @@ public class CorruptStatistics {
 
   /**
    * Decides if the statistics from a file should be ignored because they are potentially corrupt.
-   * Use this overload when the writer version has already been parsed to avoid redundant parsing.
+   * Use this when the writer version has already been parsed to avoid redundant parsing.
    *
    * @param writerVersion the pre-parsed writer version, or {@code null} if unknown/unparseable
+   * @param createdBy     the original created-by string from the file footer (used for logging)
    * @param columnType    the type of the column that this is checking
    * @return true if the statistics may be invalid and should be ignored, false otherwise
    */
-  public static boolean shouldIgnoreStatistics(ParsedVersion writerVersion, PrimitiveTypeName columnType) {
+  public static boolean shouldIgnoreStatistics(
+      ParsedVersion writerVersion, String createdBy, PrimitiveTypeName columnType) {
 
     if (columnType != PrimitiveTypeName.BINARY && columnType != PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
       return false;
@@ -100,13 +105,16 @@ public class CorruptStatistics {
     }
 
     if (Strings.isNullOrEmpty(writerVersion.version)) {
-      warnOnce("Ignoring statistics because created_by did not contain a semver (see PARQUET-251): "
-          + writerVersion);
+      warnOnce("Ignoring statistics because created_by did not contain a semver (see PARQUET-251): " + createdBy);
       return true;
     }
 
     if (!writerVersion.hasSemanticVersion()) {
-      warnOnce("Ignoring statistics because created_by could not be parsed (see PARQUET-251): " + writerVersion);
+      try {
+        SemanticVersion.parse(writerVersion.version);
+      } catch (SemanticVersionParseException e) {
+        warnParseErrorOnce(createdBy, e);
+      }
       return true;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -19,7 +19,6 @@
 package org.apache.parquet;
 
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.parquet.SemanticVersion.SemanticVersionParseException;
 import org.apache.parquet.VersionParser.ParsedVersion;
 import org.apache.parquet.VersionParser.VersionParseException;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
@@ -110,11 +109,7 @@ public class CorruptStatistics {
     }
 
     if (!writerVersion.hasSemanticVersion()) {
-      try {
-        SemanticVersion.parse(writerVersion.version);
-      } catch (SemanticVersionParseException e) {
-        warnParseErrorOnce(createdBy, e);
-      }
+      warnParseErrorOnce(createdBy, writerVersion.getSemanticVersionParseFailure());
       return true;
     }
 
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -19,7 +19,6 @@
 package org.apache.parquet;
 
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.parquet.SemanticVersion.SemanticVersionParseException;
 import org.apache.parquet.VersionParser.ParsedVersion;
 import org.apache.parquet.VersionParser.VersionParseException;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
@@ -70,37 +69,60 @@ public class CorruptStatistics {
 
     try {
       ParsedVersion version = VersionParser.parse(createdBy);
-
-      if (!"parquet-mr".equals(version.application)) {
-        // assume other applications don't have this bug
-        return false;
-      }
-
-      if (Strings.isNullOrEmpty(version.version)) {
-        warnOnce("Ignoring statistics because created_by did not contain a semver (see PARQUET-251): "
-            + createdBy);
-        return true;
-      }
-
-      SemanticVersion semver = SemanticVersion.parse(version.version);
-
-      if (semver.compareTo(PARQUET_251_FIXED_VERSION) < 0
-          && !(semver.compareTo(CDH_5_PARQUET_251_FIXED_START) >= 0
-              && semver.compareTo(CDH_5_PARQUET_251_FIXED_END) < 0)) {
-        warnOnce("Ignoring statistics because this file was created prior to "
-            + PARQUET_251_FIXED_VERSION
-            + ", see PARQUET-251");
-        return true;
-      }
-
-      // this file was created after the fix
-      return false;
-    } catch (RuntimeException | SemanticVersionParseException | VersionParseException e) {
-      // couldn't parse the created_by field, log what went wrong, don't trust the stats,
-      // but don't make this fatal.
+      return shouldIgnoreStatistics(version, columnType);
+    } catch (RuntimeException | VersionParseException e) {
       warnParseErrorOnce(createdBy, e);
       return true;
     }
+  }
+
+  /**
+   * Decides if the statistics from a file should be ignored because they are potentially corrupt.
+   * Use this overload when the writer version has already been parsed to avoid redundant parsing.
+   *
+   * @param writerVersion the pre-parsed writer version, or {@code null} if unknown/unparseable
+   * @param columnType    the type of the column that this is checking
+   * @return true if the statistics may be invalid and should be ignored, false otherwise
+   */
+  public static boolean shouldIgnoreStatistics(ParsedVersion writerVersion, PrimitiveTypeName columnType) {
+
+    if (columnType != PrimitiveTypeName.BINARY && columnType != PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
+      return false;
+    }
+
+    if (writerVersion == null) {
+      warnOnce("Ignoring statistics because created_by is null or empty! See PARQUET-251 and PARQUET-297");
+      return true;
+    }
+
+    if (!"parquet-mr".equals(writerVersion.application)) {
+      return false;
+    }
+
+    if (Strings.isNullOrEmpty(writerVersion.version)) {
+      warnOnce("Ignoring statistics because created_by did not contain a semver (see PARQUET-251): "
+          + writerVersion);
+      return true;
+    }
+
+    if (!writerVersion.hasSemanticVersion()) {
+      warnOnce("Ignoring statistics because created_by could not be parsed (see PARQUET-251): " + writerVersion);
+      return true;
+    }
+
+    SemanticVersion semver = writerVersion.getSemanticVersion();
+
+    if (semver.compareTo(PARQUET_251_FIXED_VERSION) < 0
+        && !(semver.compareTo(CDH_5_PARQUET_251_FIXED_START) >= 0
+            && semver.compareTo(CDH_5_PARQUET_251_FIXED_END) < 0)) {
+      warnOnce("Ignoring statistics because this file was created prior to "
+          + PARQUET_251_FIXED_VERSION
+          + ", see PARQUET-251");
+      return true;
+    }
+
+    // this file was created after the fix
+    return false;
   }
 
   private static void warnParseErrorOnce(String createdBy, Throwable e) {
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -1003,7 +1003,7 @@ public class ParquetMetadataConverter {
   public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
       String createdBy, Statistics statistics, PrimitiveType type) {
     SortOrder expectedOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
-    return fromParquetStatisticsInternal(createdBy, statistics, type, expectedOrder);
+    return fromParquetStatisticsInternal(null, createdBy, statistics, type, expectedOrder);
   }
 
   public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -47,6 +47,8 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.parquet.CorruptStatistics;
 import org.apache.parquet.ParquetReadOptions;
 import org.apache.parquet.Preconditions;
+import org.apache.parquet.VersionParser.ParsedVersion;
+import org.apache.parquet.VersionParser.VersionParseException;
 import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.column.EncodingStats;
 import org.apache.parquet.column.ParquetProperties;
@@ -945,7 +947,25 @@ public class ParquetMetadataConverter {
   // Visible for testing
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
       String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
-    // create stats object based on the column type
+    return fromParquetStatisticsInternal(
+        CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName()),
+        formatStats,
+        type,
+        typeSortOrder);
+  }
+
+  // Visible for testing
+  static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
+      ParsedVersion writerVersion, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
+    return fromParquetStatisticsInternal(
+        CorruptStatistics.shouldIgnoreStatistics(writerVersion, type.getPrimitiveTypeName()),
+        formatStats,
+        type,
+        typeSortOrder);
+  }
+
+  private static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
+      boolean shouldIgnoreStatistics, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
     org.apache.parquet.column.statistics.Statistics.Builder statsBuilder =
         org.apache.parquet.column.statistics.Statistics.getBuilderForReading(type);
 
@@ -962,13 +982,7 @@ public class ParquetMetadataConverter {
         boolean isSet = formatStats.isSetMax() && formatStats.isSetMin();
         boolean maxEqualsMin = isSet ? Arrays.equals(formatStats.getMin(), formatStats.getMax()) : false;
         boolean sortOrdersMatch = SortOrder.SIGNED == typeSortOrder;
-        // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
-        // The sort order is checked to avoid returning min/max stats that are not
-        // valid with the type's sort order. In previous releases, all stats were
-        // aggregated using a signed byte-wise ordering, which isn't valid for all the
-        // types (e.g. strings, decimals etc.).
-        if (!CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName())
-            && (sortOrdersMatch || maxEqualsMin)) {
+        if (!shouldIgnoreStatistics && (sortOrdersMatch || maxEqualsMin)) {
           if (isSet) {
             statsBuilder.withMin(formatStats.min.array());
             statsBuilder.withMax(formatStats.max.array());
@@ -990,6 +1004,12 @@ public class ParquetMetadataConverter {
       String createdBy, Statistics statistics, PrimitiveType type) {
     SortOrder expectedOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
     return fromParquetStatisticsInternal(createdBy, statistics, type, expectedOrder);
+  }
+
+  public org.apache.parquet.column.statistics.Statistics fromParquetStatistics(
+      ParsedVersion writerVersion, Statistics statistics, PrimitiveType type) {
+    SortOrder expectedOrder = overrideSortOrderToSigned(type) ? SortOrder.SIGNED : sortOrder(type);
+    return fromParquetStatisticsInternal(writerVersion, statistics, type, expectedOrder);
   }
 
   GeospatialStatistics toParquetGeospatialStatistics(
@@ -1837,6 +1857,24 @@ public class ParquetMetadataConverter {
         fromParquetStatistics(metaData.geospatial_statistics, type));
   }
 
+  public ColumnChunkMetaData buildColumnChunkMetaData(
+      ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, ParsedVersion writerVersion) {
+    return ColumnChunkMetaData.get(
+        columnPath,
+        type,
+        fromFormatCodec(metaData.codec),
+        convertEncodingStats(metaData.getEncoding_stats()),
+        fromFormatEncodings(metaData.encodings),
+        fromParquetStatistics(writerVersion, metaData.statistics, type),
+        metaData.data_page_offset,
+        metaData.dictionary_page_offset,
+        metaData.num_values,
+        metaData.total_compressed_size,
+        metaData.total_uncompressed_size,
+        fromParquetSizeStatistics(metaData.size_statistics, type),
+        fromParquetStatistics(metaData.geospatial_statistics, type));
+  }
+
   public ParquetMetadata fromParquetMetadata(FileMetaData parquetMetadata) throws IOException {
     return fromParquetMetadata(parquetMetadata, null, false);
   }
@@ -1854,6 +1892,17 @@ public class ParquetMetadataConverter {
       Map<RowGroup, Long> rowGroupToRowIndexOffsetMap)
       throws IOException {
     MessageType messageType = fromParquetSchema(parquetMetadata.getSchema(), parquetMetadata.getColumn_orders());
+    org.apache.parquet.hadoop.metadata.FileMetaData fileMetaData =
+        buildFileMetaData(parquetMetadata, messageType, encryptedFooter, fileDecryptor);
+    String createdBy = fileMetaData.getCreatedBy();
+    ParsedVersion writerVersion = null;
+    boolean useWriterVersion = false;
+    try {
+      writerVersion = fileMetaData.getWriterVersion();
+      useWriterVersion = true;
+    } catch (VersionParseException e) {
+      // Fall back to String-based path which logs the parse error with full context
+    }
     List<BlockMetaData> blocks = new ArrayList<BlockMetaData>();
     List<RowGroup> row_groups = parquetMetadata.getRow_groups();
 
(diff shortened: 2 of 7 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -80,6 +83,7 @@ public FileMetaData(
     this.keyValueMetaData =
         unmodifiableMap(Objects.requireNonNull(keyValueMetaData, "keyValueMetaData cannot be null"));
     this.createdBy = createdBy;
+    this.writerVersion = parseVersion(createdBy);
```
wgtmac: Seems reasonable. Do we want to initialize this field lazily?

wgtmac (approved):
Thanks for fixing this! If this gets checked in, the PR that fixes malformed stats checking can be closed, right?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -42,6 +44,7 @@ public enum EncryptionType {
   private final MessageType schema;
   private final Map<String, String> keyValueMetaData;
```
wgtmac, line 59: This cached field is `transient final`. After Java deserialization it will stay `null` even when `createdBy` is valid. Recompute lazily in `getWriterVersion()`, or add serialization handling and tests.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -118,4 +122,23 @@ public InternalFileDecryptor getFileDecryptor() {
   public EncryptionType getEncryptionType() {
     return encryptionType;
   }
+
+  /**
+   * @return the parsed writer version, or {@code null} if {@code createdBy} is null, empty, or unparseable
+   */
+  @JsonIgnore
+  public ParsedVersion getWriterVersion() {
```
wgtmac: This adds the cached value, but no production call site uses it yet. Please migrate the callers listed in the issue, especially footer stats, page stats, reader init, rewrite, and lazy encrypted metadata paths.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -118,4 +122,23 @@ public InternalFileDecryptor getFileDecryptor() {
...
+
+  private static ParsedVersion parseVersion(String createdBy) {
+    if (createdBy == null || createdBy.isEmpty()) {
+      return null;
+    }
+    try {
+      return VersionParser.parse(createdBy);
+    } catch (RuntimeException | VersionParser.VersionParseException e) {
+      return null;
+    }
```
wgtmac, line 172: Returning `null` loses the difference between missing `createdBy` and parse failure. `CorruptStatistics` currently logs different warnings for these cases. Preserve the parse failure state, or keep enough context for migrated callers to preserve existing behavior.

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/metadata/FileMetaDataTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private static final MessageType SCHEMA = new MessageType(
+      "test", new PrimitiveType(Type.Repetition.REQUIRED, PrimitiveType.PrimitiveTypeName.INT32, "id"));
+
+  @Test
+  void validCreatedByIsParsed() {
+    FileMetaData meta =
+        new FileMetaData(SCHEMA, Collections.emptyMap(), "parquet-mr version 1.12.0 (build abc123)");
+
+    assertThat(meta.getWriterVersion()).isNotNull();
```
wgtmac, line 41: These tests only cover the getter. They do not prove the repeated parsing problem is fixed. Please add coverage for at least one migrated production path using the cached `ParsedVersion` instead of reparsing `createdBy`.

wgtmac (changes requested):
Thanks for pushing this cleaner direction. I think caching the parsed writer version in FileMetaData is the right foundation, but this PR is not a complete fix yet. It currently adds the cache, but does not migrate the production call sites that still parse created_by repeatedly. Please update the hot/footer/page/reader/rewrite paths to consume the cached ParsedVersion, and cover that behavior in tests.

asifsmohammed:
> This adds the cached value, but no production call site uses it yet. Please migrate the callers listed in the issue, especially footer stats, page stats, reader init, rewrite, and lazy encrypted metadata paths.

> These tests only cover the getter. They do not prove the repeated parsing problem is fixed. Please add coverage for at least one migrated production path using the cached ParsedVersion instead of reparsing createdBy.

@wgtmac I'd prefer to keep this PR focused on the caching foundation and migrate callers in a follow-up. The migration touches multiple files which is a larger change that's easier to review separately. The follow-up will include tests proving the production path uses the cached version.
If needed I can add fixes for 1 caller using the `ParsedVersion` in this PR or I create a separate PR to fix all callers using `ParsedVersion` in parallel to this PR. Wdyt? I have no concerns on fixing all of them in this PR itself.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
```diff
@@ -42,6 +44,8 @@ public enum EncryptionType {
   private final MessageType schema;
   private final Map<String, String> keyValueMetaData;
   private final String createdBy;
+  private transient volatile ParsedVersion writerVersion;
+  private transient volatile boolean writerVersionParsed;
```
wgtmac: Please replace `writerVersion` and `writerVersionParsed` with one private immutable `WriterVersionResult` field. Use a null field only for “not initialized”. The result should represent valid, missing, or invalid and retain the original parse exception. Initialize it once in a synchronized block. Keep only `getWriterVersion()` public: return the parsed version, return null for missing `createdBy`, and rethrow the cached exception for invalid `createdBy` so callers can preserve the existing fallback without reparsing.

wgtmac (commented):
Thanks for the quick response!

I agree with keeping this PR small and moving the full migration to follow-ups. However, I do think it is worth migrating at least one production caller so this PR provides a concrete fix, not only an unused cache.

Please also narrow the title and description and remove “Closes #3696”, since the remaining paths still parse created_by repeatedly.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
...
+          + writerVersion);
+      return true;
+    }
+
+    if (!writerVersion.hasSemanticVersion()) {
+      warnOnce("Ignoring statistics because created_by could not be parsed (see PARQUET-251): " + writerVersion);
+      return true;
+    }
+
+    SemanticVersion semver = writerVersion.getSemanticVersion();
```
asifsmohammed, line 116: `ParsedVersion` eagerly parses and caches the `SemanticVersion` in its constructor, so `getSemanticVersion()` avoids the redundant `SemanticVersion.parse(version.version)` that the String-based overload previously performed on every call. The left and right spikes in flame graph are for parsing `SemanticVersion` twice. 

Image

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
asifsmohammed: I have a couple of callouts on this approach, happy to get your input on this @wgtmac 

The `shouldIgnoreStatistics(ParsedVersion, PrimitiveTypeName)` overload achieves full behavioral parity
  with the String-based overload for all real-world inputs. There are two theoretical edge cases where
  the log output differs (return value is always identical):

  1. **Empty version field** (e.g., `"parquet-mr version (build abc)"`): The log message appends
     `ParsedVersion.toString()` instead of the raw `createdBy` string, since the ParsedVersion overload
     doesn't have access to the original string.

  2. **Non-empty but invalid semver** (e.g., `"parquet-mr version xyz (build abc)"`): The old code threw
     `SemanticVersionParseException` caught by the outer catch block, logging via `warnParseErrorOnce`
     with a stack trace. The new code logs via `warnOnce` without a stack trace, since `ParsedVersion`'s
     constructor already caught and discarded the exception internally.

  Neither case occurs in practice as no known parquet writer produces such strings. For truly unparseable
  `createdBy` strings (where `VersionParser.parse` itself fails), `fromParquetMetadata` falls back to
  the String-based path via the `useWriterVersion` flag, preserving exact logging parity including the
  stack trace.

asifsmohammed:
Hi @wgtmac, I've addressed your feedback.
Would you have a chance to take another look when you get a moment? Happy to address any further feedback. Thanks!

charlessumo:
Hi @wgtmac is it possible to speed up process to get this and subsequent PRs merged? It will bring significantly cost savings to our company

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
 
     try {
       ParsedVersion version = VersionParser.parse(createdBy);
+      return shouldIgnoreStatistics(version, columnType);
+    } catch (RuntimeException | VersionParseException e) {
```
wgtmac: Why `SemanticVersionParseException` is gone?

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
...
-      // this file was created after the fix
+    if (columnType != PrimitiveTypeName.BINARY && columnType != PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
       return false;
-    } catch (RuntimeException | SemanticVersionParseException | VersionParseException e) {
-      // couldn't parse the created_by field, log what went wrong, don't trust the stats,
-      // but don't make this fatal.
-      warnParseErrorOnce(createdBy, e);
+    }
+
+    if (writerVersion == null) {
```
wgtmac: Shouldn't we throw a NPE? It could't happen before this PR.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
 
     try {
       ParsedVersion version = VersionParser.parse(createdBy);
+      return shouldIgnoreStatistics(version, columnType);
+    } catch (RuntimeException | VersionParseException e) {
+      warnParseErrorOnce(createdBy, e);
```
wgtmac, line 76:
```suggestion
      // couldn't parse the created_by field, log what went wrong, don't trust the      
      // stats, but don't make this fatal.
      warnParseErrorOnce(createdBy, e);
```

Let's keep the original comment.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -962,13 +982,7 @@ static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInte
         boolean isSet = formatStats.isSetMax() && formatStats.isSetMin();
         boolean maxEqualsMin = isSet ? Arrays.equals(formatStats.getMin(), formatStats.getMax()) : false;
         boolean sortOrdersMatch = SortOrder.SIGNED == typeSortOrder;
-        // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
```
wgtmac, line 965: Could we preserve these comments?

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
...
+
+  /**
+   * Decides if the statistics from a file should be ignored because they are potentially corrupt.
+   * Use this overload when the writer version has already been parsed to avoid redundant parsing.
+   *
+   * @param writerVersion the pre-parsed writer version, or {@code null} if unknown/unparseable
+   * @param columnType    the type of the column that this is checking
+   * @return true if the statistics may be invalid and should be ignored, false otherwise
+   */
+  public static boolean shouldIgnoreStatistics(ParsedVersion writerVersion, PrimitiveTypeName columnType) {
```
wgtmac: This overload makes calls like `shouldIgnoreStatistics(null, type)` ambiguous; the cast in the new test demonstrates the source incompatibility. Could we use a distinct method name for the `ParsedVersion` path, including the new converter overloads?

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
```diff
@@ -70,37 +69,60 @@ public static boolean shouldIgnoreStatistics(String createdBy, PrimitiveTypeName
...
+      return false;
+    }
+
+    if (Strings.isNullOrEmpty(writerVersion.version)) {
+      warnOnce("Ignoring statistics because created_by did not contain a semver (see PARQUET-251): "
+          + writerVersion);
+      return true;
+    }
+
+    if (!writerVersion.hasSemanticVersion()) {
```
wgtmac, line 111: `ParsedVersion` has already swallowed `SemanticVersionParseException` here, so this no longer preserves the old `warnParseErrorOnce(createdBy, e)` behavior. Could we keep the original string and parse exception for this path?

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
```diff
@@ -945,7 +947,25 @@ public static org.apache.parquet.column.statistics.Statistics fromParquetStatist
   // Visible for testing
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
       String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
-    // create stats object based on the column type
+    return fromParquetStatisticsInternal(
+        CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName()),
```
wgtmac: This evaluates `shouldIgnoreStatistics` even when stats are null or V2 min/max is used, so it may log “Ignoring statistics” and consume the one-shot warning when nothing is ignored. Could we keep this check inside the legacy min/max branch and add a regression test?

wgtmac (changes requested):
Sorry for the delay! Thanks @asifsmohammed for improving this! I still have some comments. Will merge it after all those have been addressed.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/metadata/FileMetaData.java
asifsmohammed: **Update**: Both edge cases are now resolved. Added `createdBy` as a parameter so log messages use the raw string, and the `!hasSemanticVersion()` branch re-parses to recreate the exception with stack trace. Full parity achieved for all scenarios.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
asifsmohammed: Added String `createdBy` as a parameter to the `ParsedVersion` overload, so this signature (`ParsedVersion`, `String`, `PrimitiveTypeName`) is no longer ambiguous with (`String`, `PrimitiveTypeName`). The `createdBy` parameter also solves the logging parity issue mentioned below comment. Converter overloads follow the same pattern.

parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
asifsmohammed: `createdBy` string is now passed as a parameter, and the `!hasSemanticVersion()` branch re-parses `writerVersion.version` to recreate the `SemanticVersionParseException` for `warnParseErrorOnce(createdBy, e)`. This gives exact log parity (original string + stack trace). The re-parse only fires when the `ParsedVersion` fails to parse it, so zero performance impact on the hot path.

parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
asifsmohammed: Moved `shouldIgnoreStatistics` evaluation inside the V1 legacy min/max branch, V2 stats now bypass it entirely. Added `testV2StatsDoNotTriggerCorruptStatisticsCheck` regression test that verifies a corrupt writer version with V2 stats still gets valid min/max without consuming the one-shot warning.

asif-moh:
Hi @wgtmac, I've addressed your feedback. Please take a look when possible. Once we merge this I can work on other 6 code paths mentioned in this issue https://github.com/apache/parquet-java/issues/3696

asifsmohammed:
Hi @wgtmac, I've addressed your feedback. Please take a look when possible. Once we merge this I can work on other 6 code paths mentioned in this issue https://github.com/apache/parquet-java/issues/3696

wgtmac:
Sorry for the delay. I still think the current approach is not clean. To avoid more roundtrip discussion, I've put up a draft as the diff below. The basic idea is to make `ParsedVersion` retain the parse failure, so callers can reuse the original exception instead of parsing the same version again just to reconstruct the warning.  
In `ParquetMetadataConverter`, both the legacy String path and the cached ParsedVersion path now share one statistics-conversion implementation. The corrupt-statistics check is selected lazily only for legacy min/max statistics, which removes the duplicated conversion logic and the `useWriterVersion` branch while keeping the existing String entry point intact. Let me know what you think. Thanks!

```
diff --git a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
index 546b9bdf..85493efc 100644
--- a/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
+++ b/parquet-column/src/main/java/org/apache/parquet/CorruptStatistics.java
@@ -19,7 +19,6 @@
 package org.apache.parquet;

 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.parquet.SemanticVersion.SemanticVersionParseException;
 import org.apache.parquet.VersionParser.ParsedVersion;
 import org.apache.parquet.VersionParser.VersionParseException;
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName;
@@ -110,11 +109,7 @@ public class CorruptStatistics {
     }

     if (!writerVersion.hasSemanticVersion()) {
-      try {
-        SemanticVersion.parse(writerVersion.version);
-      } catch (SemanticVersionParseException e) {
-        warnParseErrorOnce(createdBy, e);
-      }
+      warnParseErrorOnce(createdBy, writerVersion.getSemanticVersionParseFailure());
       return true;
     }

diff --git a/parquet-column/src/test/java/org/apache/parquet/CorruptStatisticsTest.java b/parquet-column/src/test/java/org/apache/parquet/CorruptStatisticsTest.java
index ff950a25..ba4668ad 100644
--- a/parquet-column/src/test/java/org/apache/parquet/CorruptStatisticsTest.java
+++ b/parquet-column/src/test/java/org/apache/parquet/CorruptStatisticsTest.java
@@ -162,6 +162,8 @@ public class CorruptStatisticsTest {
     // version field present but not a valid semantic version
     ParsedVersion invalidSemver = new ParsedVersion("parquet-mr", "not-a-semver", "abc");
     assertThat(invalidSemver.hasSemanticVersion()).isFalse();
+    assertThat(invalidSemver.getSemanticVersionParseFailure())
+        .isInstanceOf(SemanticVersion.SemanticVersionParseException.class);
     assertThat(CorruptStatistics.shouldIgnoreStatistics(
             invalidSemver, "parquet-mr version not-a-semver (build abc)", PrimitiveTypeName.BINARY))
         .isTrue();
diff --git a/parquet-common/src/main/java/org/apache/parquet/VersionParser.java b/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
index 07feb0d3..82e45eb8 100644
--- a/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
+++ b/parquet-common/src/main/java/org/apache/parquet/VersionParser.java
@@ -41,6 +41,7 @@ public class VersionParser {

     private final boolean hasSemver;
     private final SemanticVersion semver;
+    private final Exception semanticVersionParseFailure;

     public ParsedVersion(String application, String version, String appBuildHash) {
       checkArgument(!Strings.isNullOrEmpty(application), "application cannot be null or empty");
@@ -48,17 +49,20 @@ public class VersionParser {
       this.version = Strings.isNullOrEmpty(version) ? null : version;
       this.appBuildHash = Strings.isNullOrEmpty(appBuildHash) ? null : appBuildHash;

-      SemanticVersion sv;
-      boolean hasSemver;
-      try {
-        sv = SemanticVersion.parse(version);
-        hasSemver = true;
-      } catch (RuntimeException | SemanticVersionParseException e) {
-        sv = null;
-        hasSemver = false;
+      SemanticVersion sv = null;
+      boolean hasSemver = false;
+      Exception parseFailure = null;
+      if (this.version != null) {
+        try {
+          sv = SemanticVersion.parse(this.version);
+          hasSemver = true;
+        } catch (RuntimeException | SemanticVersionParseException e) {
+          parseFailure = e;
+        }
       }
       this.semver = sv;
       this.hasSemver = hasSemver;
+      this.semanticVersionParseFailure = parseFailure;
     }

     public boolean hasSemanticVersion() {
@@ -69,6 +73,14 @@ public class VersionParser {
       return semver;
     }

+    /**
+     * Returns the exception captured when parsing the semantic version failed, or {@code null} if
+     * parsing succeeded.
+     */
+    Exception getSemanticVersionParseFailure() {
+      return semanticVersionParseFailure;
+    }
+
     @Override
     public boolean equals(Object o) {
       if (this == o) return true;
diff --git a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
index 1f2fd6e4..8f2dd852 100644
--- a/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java
@@ -947,44 +947,7 @@ public class ParquetMetadataConverter {
   // Visible for testing
   static org.apache.parquet.column.statistics.Statistics fromParquetStatisticsInternal(
       String createdBy, Statistics formatStats, PrimitiveType type, SortOrder typeSortOrder) {
-    org.apache.parquet.column.statistics.Statistics.Builder statsBuilder =
-        org.apache.parquet.column.statistics.Statistics.getBuilderForReading(type);
-
-    if (formatStats != null) {
-      // Use the new V2 min-max statistics over the former one if it is filled
-      if (formatStats.isSetMin_value() && formatStats.isSetMax_value()) {
-        byte[] min = formatStats.min_value.array();
-        byte[] max = formatStats.max_value.array();
-        if (isMinMaxStatsSupported(type) || Arrays.equals(min, max)) {
-          statsBuilder.withMin(min);
-          statsBuilder.withMax(max);
-        }
-      } else {
-        boolean isSet = formatStats.isSetMax() && formatStats.isSetMin();
-        boolean maxEqualsMin = isSet ? Arrays.equals(formatStats.getMin(), formatStats.getMax()) : false;
-        boolean sortOrdersMatch = SortOrder.SIGNED == typeSortOrder;
-        // NOTE: See docs in CorruptStatistics for explanation of why this check is needed
-        // The sort order is checked to avoid returning min/max stats that are not
-        // valid with the type's sort order. In previous releases, all stats were
-        // aggregated using a signed byte-wise ordering, which isn't valid for all the
-        // types (e.g. strings, decimals etc.).
-        if (!CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName())
-            && (sortOrdersMatch || maxEqualsMin)) {
-          if (isSet) {
-            statsBuilder.withMin(formatStats.min.array());
-            statsBuilder.withMax(formatStats.max.array());
-          }
-        }
-      }
-
-      if (formatStats.isSetNull_count()) {
-        statsBuilder.withNumNulls(formatStats.null_count);
-      }
-      if (formatStats.isSetNan_count()) {
-        statsBuilder.withNanCount(formatStats.getNan_count());
-      }
-    }
-    return statsBuilder.build();
+    return fromParquetStatisticsInternal(null, createdBy, formatStats, type, typeSortOrder);
   }

   // Visible for testing
@@ -1015,7 +978,10 @@ public class ParquetMetadataConverter {
         // valid with the type's sort order. In previous releases, all stats were
         // aggregated using a signed byte-wise ordering, which isn't valid for all the
         // types (e.g. strings, decimals etc.).
-        if (!CorruptStatistics.shouldIgnoreStatistics(writerVersion, createdBy, type.getPrimitiveTypeName())
+        boolean shouldIgnoreStatistics = writerVersion == null
+            ? CorruptStatistics.shouldIgnoreStatistics(createdBy, type.getPrimitiveTypeName())
+            : CorruptStatistics.shouldIgnoreStatistics(writerVersion, createdBy, type.getPrimitiveTypeName());
+        if (!shouldIgnoreStatistics
             && (sortOrdersMatch || maxEqualsMin)) {
           if (isSet) {
             statsBuilder.withMin(formatStats.min.array());
@@ -1875,20 +1841,7 @@ public class ParquetMetadataConverter {

   public ColumnChunkMetaData buildColumnChunkMetaData(
       ColumnMetaData metaData, ColumnPath columnPath, PrimitiveType type, String createdBy) {
-    return ColumnChunkMetaData.get(
-        columnPath,
-        type,
-        fromFormatCodec(metaData.codec),
-        convertEncodingStats(metaData.getEncoding_stats()),
-        fromFormatEncodings(metaData.encodings),
-        fromParquetStatistics(createdBy, metaData.statistics, type),
-        metaData.data_page_offset,
-        metaData.dictionary_page_offset,
-        metaData.num_values,
-        metaData.total_compressed_size,
-        metaData.total_uncompressed_size,
-        fromParquetSizeStatistics(metaData.size_statistics, type),
-        fromParquetStatistics(metaData.geospatial_statistics, type));
+    return buildColumnChunkMetaData(metaData, columnPath, type, null, createdBy);
   }

   public ColumnChunkMetaData buildColumnChunkMetaData(
@@ -1934,10 +1887,8 @@ public class ParquetMetadataConverter {
         buildFileMetaData(parquetMetadata, messageType, encryptedFooter, fileDecryptor);
     String createdBy = fileMetaData.getCreatedBy();
     ParsedVersion writerVersion = null;
-    boolean useWriterVersion = false;
     try {
       writerVersion = fileMetaData.getWriterVersion();
-      useWriterVersion = true;
     } catch (VersionParseException e) {
       // Fall back to String-based path which logs the parse error with full context
     }
@@ -2020,10 +1971,8 @@ public class ParquetMetadataConverter {
           if (!lazyMetadataDecryption) { // full column metadata (with stats) is available
             PrimitiveType primitiveType =
                 messageType.getType(columnPath.toArray()).asPrimitiveType();
-            column = useWriterVersion
-                ? buildColumnChunkMetaData(
-                    metaData, columnPath, primitiveType, writerVersion, createdBy)
-                : buildColumnChunkMetaData(metaData, columnPath, primitiveType, createdBy);
+            column = buildColumnChunkMetaData(
+                metaData, columnPath, primitiveType, writerVersion, createdBy);
             column.setRowGroupOrdinal(rowGroup.getOrdinal());
             if (metaData.isSetBloom_filter_offset()) {
               column.setBloomFilterOffset(metaData.getBloom_filter_offset());
```

asifsmohammed:
@wgtmac thanks for the feedback. Please take a look again and let me know if you have further concerns or questions.

> The basic idea is to make `ParsedVersion` retain the parse failure, so callers can reuse the original exception instead of parsing the same version again just to reconstruct the warning.

Thats a great suggestion, added `getSemanticVersionParseFailure`

> In `ParquetMetadataConverter`, both the legacy String path and the cached ParsedVersion path now share one statistics-conversion implementation. The corrupt-statistics check is selected lazily only for legacy min/max statistics, which removes the duplicated conversion logic and the `useWriterVersion` branch while keeping the existing String entry point intact.

Removed duplicate implementation in `fromParquetStatisticsInternal`, `buildColumnChunkMetaData` and also removed  `useWriterVersion` flag. Now we just call `shouldIgnoreStatistics` based on parsed version value.

wgtmac (approved):
Thanks for bearing with me! @asifsmohammed

asif-moh:
@wgtmac Any idea when a next release will be or is there a release cycle documented somewhere? I'll try to plan the remaining 6 fixes accordingly or will see if someone else is interested in contributing.

wgtmac:
We've just released 1.18.0 a while ago so I don't expect 1.19 to be released very soon. Feel free to create followup PRs when you have time. I don't think they should be tied to any release though.

PR #3701, opened 2026-08-01
MINOR: Respect ByteBuffer offset in Binary equality

fallintoplace:
### Rationale for this change

`ByteBufferBackedBinary.equals(ByteBuffer, int, int)` compares from index 0 instead of the binary’s stored offset. For direct buffers with a non-zero position, this can make equal values compare unequal and can also make different values compare equal when prefix bytes happen to match.

Use the stored offset when comparing a `ByteBufferBackedBinary` with another `ByteBuffer`. Add regression coverage for equality symmetry, matching hash codes, and unequal values whose prefix bytes match.

### Are these changes tested?

Yes. `TestBinary` passes with 16 tests, and the Spotless check passes.

Equality for direct `ByteBuffer`-backed binaries now uses the bytes represented by the binary slice.


Fokko (approved):
This makes sense to me, not sure why it is 0 🤔

PR #3702, opened 2026-08-01
MINOR: Preserve ByteBuffer order in little-endian reads

fallintoplace:
### Rationale for this change

`ByteBufferBackedBinary.get2BytesLittleEndian()` currently changes the byte order of its backing `ByteBuffer` to little endian. Because the buffer can be caller-owned, reading the binary can unexpectedly affect subsequent reads through the original buffer.

Read the short through a duplicate buffer so the original buffer order remains unchanged.

### Are these changes tested?

Yes. The regression test verifies the returned value and preserves the original big-endian order for heap, direct, and read-only buffers. The complete `TestBinary` class and Spotless check pass.

Calling `get2BytesLittleEndian()` no longer changes the order of the caller-owned backing buffer.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -24,7 +24,6 @@ import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
 import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
```diff
@@ -587,7 +587,7 @@ public short get2BytesLittleEndian() {
         throw new IllegalArgumentException("length must be 2");
       }
 
-      return value.order(ByteOrder.LITTLE_ENDIAN).getShort(offset);
+      return value.duplicate().order(ByteOrder.LITTLE_ENDIAN).getShort(offset);
```
divjotarora: Can we do something similar as `ByteArraySliceBackedBinary` and avoid duplicating?

```
return (short) (((value.get(offset + 1) & 0xff) << 8) | (value.get(offset) & 0xff));
```

divjotarora (changes requested):
Nice catch! Seems this is a tricky area of the code, several other bugs due to not calling `duplicate()` were recently fixed 😬

divjotarora (approved):

wgtmac (approved):
Thanks @fallintoplace for fixing this and @divjotarora for the review!

PR #3707, opened 2026-08-04
GH-3706: Mark writer aborted on end() failure to avoid flushing incomplete files

### LINKED ISSUE #3706: ParquetFileWriter.end() flushes an incomplete file to storage when finalizing fails

close() in end()'s finally runs before the writer is marked aborted, so a failure serializing indexes/bloom filters/footer flushes the incomplete stream. And withAbortOnFailure only catches IOException, so a RuntimeException/Error (e.g. OOM) never aborts at all. On non-rename storage this leaves a corrupt file at the final path that a retried task silently reuses. Affects the parquet-hadoop writer path on master, any platform.

### Component(s)

Core

arg3t:
### Rationale for this change

Makes ParquetFileWriter#end() mark the writer aborted before flushing, so an incomplete file is never committed on failure. Previously close() ran before the catch block in withAbortOnFailure and withAbortOnFailure only caught IOException, causing:

 - an incomplete file being flushed to storage when serializing indexes/stats/footer throws any exception
 - a RuntimeException (e.g. OOM) to be missed, leaving the writer un-aborted and still flushing

- withAbortOnFailure catches Throwable (not just IOException) and marks the writer aborted.
- end() runs close() in an outer finally, so abort happens before close() and the flush is skipped on failure.

### Are these changes tested?
Yes. 2 new tests in TestParquetFileWriter inject an IOException/RuntimeException during end() and assert the stream is never flushed.

## Are there any user-facing changes?
No. Successful writes are unchanged; only the failure path is affected.

Closes #3706

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java
@@ -344,7 +344,7 @@ public class ParquetFileWriter implements AutoCloseable {
   private <T> T withAbortOnFailure(IOCallable<T> action) throws IOException {
     try {
       return action.call();
-    } catch (IOException e) {
+    } catch (Throwable e) {
       aborted = true;
       throw e;
     }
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Makes sense, thanks @arg3t for fixing this!

PR #3709, opened 2026-08-05
GH-3708: Inline parquet.thrift

### LINKED ISSUE #3708: Inline parquet.thrift instead of dynamically pulling in parquet-format

Tracking issue for mailing list [thread](https://lists.apache.org/thread/nkpkqz4fn9t9pcg7d16t8kwgmth86pn7) to remove the dynamic Maven-based dependency on `apache/parquet-format` (specifically the `parquet.thrift` file) and instead inlining the `parquet.thrift` file in the repo. This would allow for faster development as POCs/reference implementations could be completed with passing CI checks without waiting for a parquet-format release.

### Component(s)

Thrift

divjotarora:
### Rationale for this change
It is currently not possible to build working POCs for unreleased parquet-format changes in parquet-java because the parquet-format dependency must be updated using a released version. This blocks reference implementations from merging and will become a bigger issue if the current versioning proposal goes through and we start having more "preview" features where writes are expected to support unreleased spec changes behind feature flags.  

This PR removes the dependency on github.com/apache/parquet-format and instead adds an inlined copy of `parquet.thrift` (`parquet-format-structures/src/main/thrift/parquet.thrift`) as well as a "sidecar" metadata file (`parquet-format-structures/src/main/thrift/parquet-format.version`) to indicate the parquet-format commit that's being inlined. There is a new `dev/update-parquet-thrift.sh [$REF]` script that takes in a parquet-format ref (commit SHA or tag) and updates the inlined `parquet.thrift` file to match that version. The script downloads from github.com/apache/parquet-format by default but can be pointed at a fork by setting the `PARQUET_FORMAT_REPO` environment variable (e.g. for POC implementations of unmerged format changes).

### Are these changes tested?
Edit inlined parquet.thrift with a meaningless change and then validate it gets overwritten:
```
-- Validate diff
> git diff
--- a/parquet-format-structures/src/main/thrift/parquet.thrift
+++ b/parquet-format-structures/src/main/thrift/parquet.thrift
...

-- Force an update
> ./dev/update-parquet-thrift.sh apache-parquet-format-2.13.0
Resolving apache-parquet-format-2.13.0 ...
  -> commit: c47e2a66e88943fc46fde1b028a9432f14fdf5c0
Fetching parquet.thrift at c47e2a66e88943fc46fde1b028a9432f14fdf5c0 ...
  -> written to /home/div.arora/parquet-java-inline-thrift/parquet-java/parquet-format-structures/src/main/thrift/parquet.thrift
  -> sidecar updated: /home/div.arora/parquet-java-inline-thrift/parquet-java/parquet-format-structures/src/main/thrift/parquet-format.version

Update complete:
  old commit: c47e2a66e88943fc46fde1b028a9432f14fdf5c0
  new commit: c47e2a66e88943fc46fde1b028a9432f14fdf5c0

-- Validate diff
> git diff
-- empty

-- Update to parquet-format commit containing FILE type
> ./dev/update-parquet-thrift.sh e94a5d090b324a0c0ee1adbb8ea6b099852dc3cc
Fetching parquet.thrift at e94a5d090b324a0c0ee1adbb8ea6b099852dc3cc ...
  -> written to /home/div.arora/parquet-java-inline-thrift/parquet-java/parquet-format-structures/src/main/thrift/parquet.thrift
  -> sidecar updated: /home/div.arora/parquet-java-inline-thrift/parquet-java/parquet-format-structures/src/main/thrift/parquet-format.version

Update complete:
  old commit: c47e2a66e88943fc46fde1b028a9432f14fdf5c0
  new commit: e94a5d090b324a0c0ee1adbb8ea6b099852dc3cc

-- Validate diff
> git diff
--- a/parquet-format-structures/src/main/thrift/parquet-format.version
+++ b/parquet-format-structures/src/main/thrift/parquet-format.version
@@ -1,2 +1,2 @@
 # Provenance of the inlined parquet.thrift. Maintained by dev/update-parquet-thrift.sh.
-parquet-format.commit=c47e2a66e88943fc46fde1b028a9432f14fdf5c0
+parquet-format.commit=e94a5d090b324a0c0ee1adbb8ea6b099852dc3cc
...
diff --git a/parquet-format-structures/src/main/thrift/parquet.thrift b/parquet-format-structures/src/main/thrift/parquet.thrift
...
+struct FileType {
+}
+

-- Download from a fork, commit hash is before IEEE sort order + nan_count changes
> PARQUET_FORMAT_REPO="https://github.com/divjotarora/parquet-format" ./dev/update-parquet-thrift.sh f4288e602a41deba78c58e8bd116b36076a03588
Update complete:
  old commit: c47e2a66e88943fc46fde1b028a9432f14fdf5c0
  new commit: f4288e602a41deba78c58e8bd116b36076a03588

> git diff
--- a/parquet-format-structures/src/main/thrift/parquet.thrift
+++ b/parquet-format-structures/src/main/thrift/parquet.thrift
...
-   9: optional i64 nan_count;
...
```

No

Closes #3708


dev/prepare-release.sh
```diff
@@ -38,6 +38,9 @@ new_development_version="$release_version-SNAPSHOT"
 
 tag="apache-parquet-$release_version-rc$2"
 
+# Ensure the inlined parquet.thrift matches an official parquet-format release.
+"$(dirname "$0")/check-parquet-thrift-release.sh"
```
rdblue: I think this is an unnecessary requirement. This assumes that we can only release Parquet Java if there has also been a release of Parquet Format, which is beyond the scope of what we are trying to do here.

The primary purpose of this change is to use a local copy of parquet.thrift. A secondary concern is helping us to update that file easily and know when it has changed. I think we should remove changes to the release script unless we decide to make this a requirement.

dev/check-parquet-thrift-release.sh
```diff
@@ -0,0 +1,82 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Verifies that the parquet-format.version sidecar file is tracking a released parquet-format
+# version and that the inlined parquet.thrift byte-matches the upstream file at that version.
```
rdblue: I think that this should be removed, or at least changed to check against the `parquet-format.version` file and nothing else. We do need to be able to check the state of the thrift file, but that has little to do with releases. Plus, the file is version controlled, so we should be able to easily tell when it was modified and where the content came from by looking at parquet-format.version.

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,161 @@
...
+
+if [[ $# -eq 0 ]]; then
+  usage
+fi
+
+mode=""
+commit_arg=""
+version_arg=""
+
+if [[ "$1" == "--version" ]]; then
```
rdblue: Can't we just use a ref? If it is a tag then we get that version, otherwise we get the commit hash? That seems simpler than implementing CLI parsing and validation here.

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,161 @@
...
+    else
+      echo "ERROR: git/network failure resolving tag apache-parquet-format-${version_arg}" >&2
+    fi
+    exit 1
+  }
+  resolved_version="$version_arg"
+  echo "  -> commit: $resolved_sha"
+else
+  resolved_sha="$commit_arg"
+  resolved_version="UNRELEASED"
```
rdblue: This isn't necessarily true. I think we should omit the resolved version instead of saying the content is not released. This could be a commit hash referring to a commit that is in a released version.

dev/parquet-thrift-lib.sh
```diff
@@ -0,0 +1,64 @@
...
+  url="${PARQUET_FORMAT_RAW}/${ref}/${THRIFT_PATH_IN_FORMAT}"
+  for attempt in 1 2 3; do
+    if curl -fsSL -o "$dest" "$url" && [[ -s "$dest" ]]; then
+      return 0
+    fi
+    if [[ "$attempt" -eq 3 ]]; then
+      echo "ERROR: failed to fetch ${url} after ${attempt} attempts" >&2
+      return 1
+    fi
+    sleep $((attempt * 2))
```
rdblue: Seems overcomplicated to have retry logic unless we actually need it.

divjotarora: I added this because I saw one failure during testing, but Github was later confirmed to be down during that time, so realistically retries wouldn't have helped much. Removed.

divjotarora:
@rdblue Thanks for the review! You raised a few big points, I'll recap them all here rather than replying to individual comments:

1. Release validation: the release check to ensure parquet-java tracks a released version of parquet-format was a suggestion from the mailing list [thread](https://lists.apache.org/thread/nkpkqz4fn9t9pcg7d16t8kwgmth86pn7). I audited the C++ and Rust implementations and didn't find a matching check there, so I've removed it.
2. Update script: I've removed the `--version` argument and now the update script just takes a single ref (either a full commit SHA or a tag) and it just downloads the corresponding `parquet.thrift`. I've also removed the retry logic. 
3. Sidecar file: the `parquet-format.version` sidecar file now only stores a commit SHA, not a semantic version. In the case that the update script is invoked with a tag ref, it's first resolved to the corresponding SHA and that's stored in the sidecar.

Fokko:
Thanks @divjotarora for working on this. I'm wondering if it wouldn't be easier to pull the structures from the snapshots repository: https://repository.apache.org/content/groups/snapshots/org/apache/parquet/parquet-format-structures/. It looks like it is missing some versions, but I think it would make sense to just take the snapshot from there and avoid a shell script.

divjotarora:
> Thanks @divjotarora for working on this. I'm wondering if it wouldn't be easier to pull the structures from the snapshots repository: https://repository.apache.org/content/groups/snapshots/org/apache/parquet/parquet-format-structures/. It looks like it is missing some versions, but I think it would make sense to just take the snapshot from there and avoid a shell script.

@Fokko thanks for taking a look. The suggestion to pull in parquet-format via Maven solves the issue for local development and unblocks building reference implementations against a local copy of `parquet.thrift`. However, it does not make CI pass because CI will still reference a commit hash of parquet-format.

FWIW the motivation for this change was that the Rust and C++ Parquet implementations both vendor this file (Rust technically has hand-rolled Thrift structs and a custom parser, so they don't vendor parquet.thrift but they do make custom changes to those structs for reference implementations and it's conceptually the same thing), so there is precedent for this. Based on my investigation, vendoring the file is the most flexible option.

rdblue:
Using snapshot builds is an interesting idea, but I think it's simpler to just vendor the file. I like having the file here so that we can see when it was updated, what version it came from, and whether there have been local changes. With a copy from a snapshot build, I think we'd end up not really knowing exactly what is being used (at least not easily). It would also be harder to test out changes before committing them.

I was also surprised by the size of this PR, though. I think that the reason is we are adding requirements that we didn't in the other implementations, like an easy way to check the version or update it automatically. I think it's nice to have those, but the main goal is just to vendor the file and point to the local one. I wouldn't want to split this PR, but if the scope of the util scripts is too much we could to focus on not being tied to a parquet-format release and, separately, the management scripts in a follow up.

divjotarora:
@rdblue Pretty much all of the management script code has been removed, there is just one script now to pull in a provided ref of parquet-format. This could be removed in favor of copy/pasting in the file, but its < 150 LoC and quite straightforward. The bulk of the PR size comes from the `parquet.thrift` addition (+1,443 out of the total +1,574 LoC).

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,127 @@
...
+#   (e.g. apache-parquet-format-2.13.0).
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+THRIFT_FILE="${REPO_ROOT}/parquet-format-structures/src/main/thrift/parquet.thrift"
+SIDECAR_FILE="${REPO_ROOT}/parquet-format-structures/src/main/thrift/parquet-format.version"
+
+PARQUET_FORMAT_REPO="https://github.com/apache/parquet-format"
```
Fokko: We can do this separately, but I think it would also make sense to run this against a fork.

Fokko:
I'd rather avoid having two copies of `parquet.thrift` living in separate places, but it is not a hill I'm willing to die on. I do agree that it speeds up the development cycle, at the same time, I don't expect anyone to iterate much on the thrift definitions.

The only thing I'm missing here is adding some docs on how to use the script.

dev/update-parquet-thrift.sh
divjotarora: Done, the script honors `PARQUET_FORMAT_REPO` if set and defaults to apache/parquet-format if not. Tested via
```
> PARQUET_FORMAT_REPO="https://github.com/divjotarora/parquet-format" ./dev/update-parquet-thrift.sh f4288e602a41deba78c58e8bd116b36076a03588
```

divjotarora:
> I'd rather avoid having two copies of `parquet.thrift` living in separate places, but it is not a hill I'm willing to die on. I do agree that it speeds up the development cycle, at the same time, I don't expect anyone to iterate much on the thrift definitions.
> 
> The only thing I'm missing here is adding some docs on how to use the script.

@Fokko Can you give me a pointer on where to add such docs? There is no README in the `dev` folder.

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,130 @@
...
+[[ $# -ne 1 ]] && usage
+
+ref="$1"
+
+# Resolve ref to a full commit SHA.
+if [[ "$ref" =~ ^[0-9a-f]{40}$ ]]; then
+  resolved_sha="$ref"
+else
+  echo "Resolving ${ref} ..."
+  ls_out="$(git ls-remote "$PARQUET_FORMAT_REPO" "$ref" "${ref}^{}")" || {
```
rdblue, line 64: A comment to help the reader understand what `${ref}^{}` is doing would help.

> ^{} refers to the unwrapped commit of an annotated tag, if present

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,130 @@
...
+    echo "ERROR: git ls-remote failed for '${ref}' in ${PARQUET_FORMAT_REPO}" >&2
+    exit 1
+  }
+  if [[ -z "$ls_out" ]]; then
+    echo "ERROR: ref '${ref}' not found in ${PARQUET_FORMAT_REPO}" >&2
+    exit 1
+  fi
+  resolved_sha="$(printf '%s\n' "$ls_out" \
+    | awk '{ if ($2 ~ /\^\{\}$/) deref=$1; else plain=$1 }
+           END { print (deref != "" ? deref : plain) }')"
```
rdblue: This is complicated and hard to read. I think you could use `git ls-remote ${ref} ${ref}^{} | tail -1 | cut -f1` instead. `^{}` is returned second if it is present, so the `tail -1` will select it.

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,130 @@
...
+url="${PARQUET_FORMAT_RAW}/${resolved_sha}/${THRIFT_PATH_IN_FORMAT}"
+if ! curl -fsSL -o "$tmp" "$url"; then
+  echo "ERROR: could not fetch parquet.thrift at ${resolved_sha}" >&2
+  echo "       The tracked files were not modified." >&2
+  exit 1
+fi
+
+if [[ ! -s "$tmp" ]]; then
+  echo "ERROR: fetched parquet.thrift is empty" >&2
+  echo "       The tracked files were not modified." >&2
```
rdblue: This is fairly obvious from `git` so I don't know that we need to state it.

rdblue (approved):
It took a bit to understand what the script is doing, but it all look correct to me.

dev/update-parquet-thrift.sh
```diff
@@ -0,0 +1,130 @@
...
+#
+#   Set PARQUET_FORMAT_REPO to pull from a fork instead of apache/parquet-format,
+#   e.g. PARQUET_FORMAT_REPO=https://github.com/<user>/parquet-format.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+THRIFT_FILE="${REPO_ROOT}/parquet-format-structures/src/main/thrift/parquet.thrift"
+SIDECAR_FILE="${REPO_ROOT}/parquet-format-structures/src/main/thrift/parquet-format.version"
```
rdblue, line 37: I'd prefer a name that refers to the thrift file, rather than to the parquet-format repo, but this is minor.

Fokko:
@divjotarora, I think a few lines under https://github.com/apache/parquet-java#building would be sufficient. So folks know how to find it and use it.

Fokko (approved):
In general, this looks good to me 👍

dev/update-parquet-thrift.sh
divjotarora: Added

dev/update-parquet-thrift.sh
divjotarora: Removed here and below

dev/update-parquet-thrift.sh
divjotarora: Thanks for the suggestion, that's definitely simpler 😄

dev/update-parquet-thrift.sh
divjotarora: IMO we should leave this as-is to indicate that the provenance of the `parquet.thrift` file is the parquet-format repo. Can change if you feel strongly though

RussellSpitzer (approved):

PR #3711, opened 2026-08-06
GH-3710: Tolerate unrecognized logical/physical type combinations when reading

### LINKED ISSUE #3710: Specify handling for unrecognized logical/physical type combinations
Tracking issue for a reference implementation of https://github.com/apache/parquet-format/issues/607

### Component(s)

Core

divjotarora:
### Rationale for this change
See https://github.com/apache/parquet-format/issues/607 for rationale.

This PR modifies parquet-java to gracefully handle unrecognized logical/physical type combinations by dropping the logical type during the read and dropping any associated statistics for the relevant columns. Note that unrecognized logical types are already handled gracefully and no changes were required.

### Are these changes tested?
Yes, several unit tests added.

No.

Closes #3710


parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -426,8 +429,40 @@ public THIS columnOrder(ColumnOrder columnOrder) {
...
+      return self();
+    }
+
     @Override
     protected PrimitiveType build(String name) {
+      try {
+        return validateAndBuild(name);
+      } catch (IllegalStateException e) {
+        boolean canDrop = dropUnsupportedLogicalTypeCombinations
+            && !(logicalTypeAnnotation instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation);
```
emkornfield: why is this special cased?

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
```diff
@@ -2279,4 +2283,62 @@ public void testColumnIndexNanCountsRoundTrip() {
     assertThat(roundTrip).isNotNull();
     assertThat(roundTrip.getNanCounts()).containsExactly(1L, 0L, 0L);
   }
+
+  @Test
```
emkornfield, line 2287: not sure where the right file for it is but it would be good to have a more end-to-end test.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -426,8 +429,40 @@ public THIS columnOrder(ColumnOrder columnOrder) {
...
     }
 
+    /**
+     * When set, an unsupported combination results in the logical type annotation being dropped
+     * rather than throwing. The associated statistics are also forcefully ignored by setting the
+     * column order to {@link ColumnOrderName#UNDEFINED}.
+     *
+     * @return this builder for method chaining
+     */
+    public THIS dropUnsupportedLogicalTypeCombinations() {
```
emkornfield, line 448: how is this method expected to be called in practice?

emkornfield (changes requested):
I think most important question is how we make this transition and an end-to-end test.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: This is called directly in `ParquetMetadataConverter.java` [here](https://github.com/apache/parquet-java/pull/3711/changes#diff-1afc9f89a782ddd4e7cd17546ca048954091627d7a31597ab88892eb2a7a76abR2065-R2066)

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: This is leftover from an old version of the implementation, removed the special case

parquet-hadoop/src/test/java/org/apache/parquet/format/converter/TestParquetMetadataConverter.java
divjotarora: Added a new `TestReadInvalidTypeCombination`

divjotarora (commented):
@emkornfield I added an e2e test that reads a golden file with an INT32 column annotated with UUID. We can add this file to parquet-testing as part of this work as well.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
emkornfield: sorry missed this.

parquet-hadoop/src/test/resources/invalid_type_combination.parquet
emkornfield, line 1: is this just a place holder for the until the parquet testing file is merged?

parquet-hadoop/src/test/java/org/apache/parquet/hadoop/TestReadInvalidTypeCombination.java
```diff
@@ -0,0 +1,77 @@
...
+public class TestReadInvalidTypeCombination {
+
+  // Path to a Parquet file that contains an invalid logical/physical type combination.
+  private static final String FILE_PATH = "/invalid_type_combination.parquet";
+
+  private static Path getFilePath() {
+    try {
+      return new Path(TestReadInvalidTypeCombination.class.getResource(FILE_PATH).toURI());
+    } catch (URISyntaxException e) {
+      throw new RuntimeException(e);
```
emkornfield: just make the method throws Exception?

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -426,8 +429,38 @@ public THIS columnOrder(ColumnOrder columnOrder) {
...
+      this.dropUnsupportedLogicalTypeCombinations = true;
+      return self();
+    }
+
     @Override
     protected PrimitiveType build(String name) {
+      try {
+        return validateAndBuild(name);
+      } catch (IllegalStateException e) {
```
emkornfield: it is possible to make this more specific?  Also there is a general spec question on whether we should error for clearly invalid types (e.g. decimal with a precision that is too high on ints).  This is probably also a spec level question.

emkornfield (commented):
main concern is on the breadth of the exception cast.

parquet-hadoop/src/test/resources/invalid_type_combination.parquet
divjotarora: Yes, I'll update the PR once that one is merged

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: Good point, I agree this should be scoped to only throwing on invalid combinations. Right now everything throws `IllegalStateException` so we can't distinguish at this level. I added a new `UnsupportedLogicalTypeAnnotation` exception to distinguish.

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -344,6 +353,9 @@ public abstract static class BasePrimitiveBuilder<P, THIS extends BasePrimitiveB
     private int precision = NOT_SET;
     private int scale = NOT_SET;
     private ColumnOrder columnOrder;
+    // When true and an unsupported logical/physical type combination is encountered, the
+    // annotation is dropped and the column order is forced to "undefined" so stats are ignored.
```
RussellSpitzer: This is a load bearing comment :) I would probably skip all this since it's a implementation detail that can go stale pretty quickly. Instead stick to the behavior.

For example you could say "ignoreUnsupportedLogicalAnnotations" as the variable name and I think that would cover it? Or "dropUnsupportedLogicalAnnotations"?

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -1605,4 +1605,17 @@ public void testGeographyLogicalTypeWithoutEdgeInterpolationAlgorithm() {
         Types.optional(BINARY).as(LogicalTypeAnnotation.geographyType()).named("aGeography");
     assertThat(optionalGeographyActual).isEqualTo(optionalGeographyExpected);
   }
+
+  @Test
+  public void testDropUnsupportedLogicalTypeCombinations() {
+    // Other tests already validate that unsupported type combinations throw by default, so this
```
RussellSpitzer: I'd probably drop this description?

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
```diff
@@ -590,9 +632,15 @@ public Optional<Boolean> visit(
                 return checkBinaryPrimitiveType(geographyLogicalType);
               }
 
+              private void checkAnnotation(boolean valid, String message, Object... args) {
```
RussellSpitzer: I'd generally want to avoid passing a "boolean" into the "checkAnnotation" method. Since it doesn't actually check an annotation here, it just throws an exception if you pass through is false. I think this is a bit over-indexing on the previous "checkState" work and possibly hiding some other holes.

I would probably try to re-think the whole visitor here rather than try to wrap a previous failure differently.


Instead of what we have here maybe something like

AllowedPhysicalTypes visitor

```java
/**
 * The physical types a logical type annotation may annotate. An empty set means the annotation
 * cannot be applied to a primitive type at all.
 */
private static final class AllowedPhysicalTypes {
  private static final AllowedPhysicalTypes NONE =
      new AllowedPhysicalTypes(EnumSet.noneOf(PrimitiveTypeName.class), NOT_SET);
  private static final AllowedPhysicalTypes ANY =
      new AllowedPhysicalTypes(EnumSet.allOf(PrimitiveTypeName.class), NOT_SET);
  private final Set<PrimitiveTypeName> types;
  private final int requiredLength;
  static Optional<AllowedPhysicalTypes> of(PrimitiveTypeName... types) {
    return Optional.of(new AllowedPhysicalTypes(EnumSet.copyOf(asList(types)), NOT_SET));
  }
  static Optional<AllowedPhysicalTypes> fixed(int requiredLength) {
    return Optional.of(new AllowedPhysicalTypes(
        EnumSet.of(PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY), requiredLength));
  }
  boolean accepts(PrimitiveTypeName type, int length) {
    return types.contains(type) && (requiredLength == NOT_SET || length == requiredLength);
  }
  @Override
  public String toString() {
    if (requiredLength != NOT_SET) {
      return PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY + "(" + requiredLength + ")";
    }
    return types.stream().map(Enum::name).collect(Collectors.joining(", "));
  }
}
```

```java
private static final LogicalTypeAnnotationVisitor<AllowedPhysicalTypes> ALLOWED_PHYSICAL_TYPES =
    new LogicalTypeAnnotationVisitor<AllowedPhysicalTypes>() {
      @Override public Optional<AllowedPhysicalTypes> visit(StringLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(JsonLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(BsonLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(EnumLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(GeometryLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(GeographyLogicalTypeAnnotation t) { return of(BINARY); }
      @Override public Optional<AllowedPhysicalTypes> visit(DateLogicalTypeAnnotation t) { return of(INT32); }
      @Override public Optional<AllowedPhysicalTypes> visit(TimestampLogicalTypeAnnotation t) { return of(INT64); }
      @Override public Optional<AllowedPhysicalTypes> visit(UUIDLogicalTypeAnnotation t) { return fixed(UUIDLogicalTypeAnnotation.BYTES); }
      @Override public Optional<AllowedPhysicalTypes> visit(Float16LogicalTypeAnnotation t) { return fixed(Float16LogicalTypeAnnotation.BYTES); }
      @Override public Optional<AllowedPhysicalTypes> visit(IntervalLogicalTypeAnnotation t) { return fixed(12); }
      @Override public Optional<AllowedPhysicalTypes> visit(TimeLogicalTypeAnnotation t) {
        return t.getUnit() == TimeUnit.MILLIS ? of(INT32) : of(INT64);
      }
      @Override public Optional<AllowedPhysicalTypes> visit(IntLogicalTypeAnnotation t) {
        return t.getBitWidth() == 64 ? of(INT64) : of(INT32);
      }
      @Override public Optional<AllowedPhysicalTypes> visit(DecimalLogicalTypeAnnotation t) {
        return of(INT32, INT64, BINARY, FIXED_LEN_BYTE_ARRAY);
      }
      @Override public Optional<AllowedPhysicalTypes> visit(UnknownLogicalTypeAnnotation t) {
        return Optional.of(ANY);
      }
    };
```

Then instead of trying to throw our way to control flow we can do something like

```java
if (logicalTypeAnnotation != null) {
  AllowedPhysicalTypes allowed =
      logicalTypeAnnotation.accept(ALLOWED_PHYSICAL_TYPES).orElse(AllowedPhysicalTypes.NONE);
  if (!allowed.accepts(primitiveType, length)) {
    if (!dropUnsupportedLogicalAnnotations) {
      throw new IllegalStateException(allowed.isEmpty()
          ? logicalTypeAnnotation + " can not be applied to a primitive type"
          : String.format("%s can only annotate %s", logicalTypeAnnotation, allowed));
    }
    LOGGER.warn(
        "Dropping unsupported logical type annotation {} on physical type {}",
        logicalTypeAnnotation, primitiveType);
    return new PrimitiveType(
        repetition, primitiveType, this.length, name, null, null, id, ColumnOrder.undefined());
  }
  validateDecimalPrecision(meta);
}
```

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
RussellSpitzer: This is avoiding the checkValidAnnotation block

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: Removed comment and renamed var to `ignoreUnsupportedLogicalAnnotations`

parquet-column/src/main/java/org/apache/parquet/schema/Types.java
divjotarora: Fair point, I was trying to make the minimal code change but I agree it was a bad abstraction. The structure you're proposing is similar to Arrow C++ and I think it's better. Will take a stab at it.

RussellSpitzer: Yeah, Sometimes I think it's just worth pulling the trigger and doing it the right way. You definitely had the smallest change version here, but I think it becomes dangerous for leaking cases (like the decimal one above) because everything keeps going through different paths.

divjotarora:
Thanks for the suggestions @RussellSpitzer, I've applied the new code structure now

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
```diff
@@ -1605,4 +1601,20 @@ public void testGeographyLogicalTypeWithoutEdgeInterpolationAlgorithm() {
...
+    for (LogicalTypeAnnotation annotation : annotations) {
+      PrimitiveType pt = Types.required(BOOLEAN)
+          .ignoreUnsupportedLogicalAnnotations()
+          .as(annotation)
+          .named("unsupported");
+      assertThat(pt.getPrimitiveTypeName()).isEqualTo(BOOLEAN);
+      assertThat(pt.getLogicalTypeAnnotation()).isNull();
+      assertThat(pt.columnOrder().getColumnOrderName()).isEqualTo(ColumnOrder.ColumnOrderName.UNDEFINED);
+    }
+  }
```
RussellSpitzer, line 1625: Since we are still rejecting Decimals with bad precessions with a hard error (I think this is a good choice), we need a test to make sure that still happens when "ignoreUnsupportedLogicalAnnotations()" is used.

RussellSpitzer (approved):
Looks good to me now! I left one note on the missing test case. 

I'm on board with setup of 

We tolerate wrong logical type on physical type except for those few cases we have already explicitly noted are bad (Decimal with the wrong precision)

parquet-column/src/test/java/org/apache/parquet/schema/TestTypeBuilders.java
divjotarora: Added a test

parquet-arrow/src/test/java/org/apache/parquet/arrow/schema/TestSchemaConverter.java
```diff
@@ -660,6 +660,6 @@ public void testParquetInt32TimestampMicrosToArrow() {
                 .named("a"))
             .named("root")))
         .isInstanceOf(IllegalStateException.class)
-        .hasMessage("TIMESTAMP(MICROS,false) can only annotate INT64");
+        .hasMessage("TIMESTAMP(MICROS,false) can only annotate [INT64]");
```
Fokko, line 663: nit: Maybe we can drop the brackets to avoid having to update these tests

Fokko (approved):
Left one nit, but this looks good to me 👍

parquet-arrow/src/test/java/org/apache/parquet/arrow/schema/TestSchemaConverter.java
divjotarora: This was an intentional change as there might be multiple valid physical types and I felt printing as an array helps readability

PR #3717, opened 2026-08-13
GH-3716: Fix data corruption in ByteBufferBackedBinary.getBytes() for non-array-backed buffers

### LINKED ISSUE #3716: [parquet-java 1.18.0] data-corruption for reading NUMERIC arrays

We have an existing passing test case to write a parquet numeric array and read it back. After upgrading parquet-java from 1.17.1 to 1.18.0, the test failed because of the read mismatch.

For example:
```
# what we write via a parquet writer
{"12345678901234567890.123456789012345678", "12345678901234567890.123456789012345678"},
{"22.2345", "22.2345"},

# what we read from a parquet reader
row 0: 12345678901234567890.123456789012345678 12345678901234567890.123456789012345678
row 1: 12345678901234567890.123456789012345678 22.234500000000000000 <- the first record is wrong
```

I can reproduce this issue with the following test code:
```java
import org.apache.parquet.example.data.Group;
import org.apache.parquet.example.data.simple.SimpleGroupFactory;
import org.apache.parquet.hadoop.ParquetWriter;
import org.apache.parquet.hadoop.example.ExampleParquetWriter;
import org.apache.parquet.hadoop.example.GroupWriteSupport;
import org.apache.parquet.io.api.Binary;
import org.apache.parquet.schema.MessageTypeParser;
import org.apache.parquet.hadoop.ParquetReader;
import org.apache.parquet.hadoop.example.GroupReadSupport;

import java.math.BigDecimal;
import java.math.BigInteger;
import java.util.Arrays;

public class Repro {
    public static void main(String[] args) throws Exception {
        String schemaStr = "message test {\n" +
                "  optional group numeric_array (LIST) {\n" +
                "    repeated group list {\n" +
                "      optional fixed_len_byte_array(16) element (DECIMAL(38,18));\n" +
                "    }\n" +
                "  }\n" +
                "}";
        MessageType schema = MessageTypeParser.parseMessageType(schemaStr);

        Path path = new Path("file:///tmp/parquet-repro/out.parquet");
        path.getFileSystem(conf).delete(path, false);

        // 2 rows, 2 elements per row -- this is the minimal trigger.
        // A single element per row does NOT reproduce the bug.
        String[][] rows = {
            {"12345678901234567890.123456789012345678", "12345678901234567890.123456789012345678"},
            {"22.2345", "22.2345"},
        };

        try (ParquetWriter<Group> writer = ExampleParquetWriter.builder(path)
                .withConf(conf)
                .withType(schema)
                .withCompressionCodec(CompressionCodecName.UNCOMPRESSED)
                .withDictionaryEncoding(false)
                .build()) {
            SimpleGroupFactory factory = new SimpleGroupFactory(schema);
            for (String[] row : rows) {
                Group group = factory.newGroup();
                Group listGroup = group.addGroup(0);
                for (String decVal : row) {
                    Group repeatedGroup = listGroup.addGroup(0);
                    byte[] bytes = toFixedLenBytes(decVal, 18, 16);
                    repeatedGroup.add(0, Binary.fromReusedByteArray(bytes));
                }
                writer.write(group);
            }
        }

        try (ParquetReader<Group> reader = ParquetReader.builder(new GroupReadSupport(), path)
                .withConf(conf)
                .build()) {
            Group g;
            int i = 0;
            while ((g = reader.read()) != null) {
                Group listGroup = g.getGroup(0, 0);
                int count = listGroup.getFieldRepetitionCount(0);
                StringBuilder sb = new StringBuilder("row " + i + ": ");
                for (int j = 0; j < count; j++) {
                    Group repeatedGroup = listGroup.getGroup(0, j);
                    byte[] readBytes = repeatedGroup.getBinary(0, 0).getBytes();
                    BigDecimal bd = new BigDecimal(new BigInteger(readBytes), 18);
                    sb.append(bd).append(" ");
                }
                System.out.println(sb);
                i++;
            }
        }
    }

    private static byte[] toFixedLenBytes(String decStr, int scale, int len) {
        BigDecimal bd = new BigDecimal(decStr).setScale(scale);
        BigInteger unscaled = bd.unscaledValue();
        byte[] full = unscaled.toByteArray();
        byte[] tgt = new byte[len];
        if (unscaled.signum() == -1) {
            Arrays.fill(tgt, (byte) 0xFF);
        }
        int srcStart = Math.max(0, full.length - len);
        int copyLen = full.length - srcStart;
        System.arraycopy(full, srcStart, tgt, len - copyLen, copyLen);
        return tgt;
    }
}
```

### Component(s)

_No response_

yimingli-vmware:
## Rationale for this change

Fixes #3716.

`FixedLenByteArrayPlainValuesReader` hands out `Binary` values that all
share one page-wide `ByteBuffer`, advancing its live position on every
`readBytes()` call. `Binary.ByteBufferBackedBinary.getBytes()` and
`toStringUsingUTF8()` (non-array-backed branch) called
`value.limit(offset + length)` directly on that same shared buffer before
capturing `value.position()`. Since `ByteBuffer.limit()` clamps `position`
down whenever `position > newLimit`, calling `getBytes()` on an earlier
value *after* later values have already advanced the buffer permanently
rewinds the buffer's live position -- corrupting every `readBytes()` call
that follows.

This surfaces as data corruption when reading a repeated (`LIST`)
`FIXED_LEN_BYTE_ARRAY` column with 2+ elements per row across 2+ rows:
record assembly stores each `Binary` and only materializes it once a full
row/group has been built, which is exactly the lazy-after-later-value
pattern that triggers the clamp. Each subsequent row reads back the
previous row's last-written element instead of its own (see #3716 for a
minimal standalone repro).

## What changes are included in this PR?

- `Binary.ByteBufferBackedBinary.getBytes()` and `.toStringUsingUTF8()` now
  `duplicate()` the buffer before adjusting position/limit, so the shared
  buffer's own position is never mutated.
- Added regression tests to
  `TestFixedLenByteArrayPlainValuesWriterReader` that read values out of
  the order they were materialized, matching the lazy-consumption pattern
  from record assembly, and fail against unpatched 1.18.0.

## Are these changes tested?

Yes -- two new tests
(`testLazyGetBytesDoesNotCorruptSubsequentReadsDirectBuffer`,
`testLazyToStringUsingUTF8DoesNotCorruptSubsequentReadsDirectBuffer`) in
`parquet-column/src/test/java/org/apache/parquet/column/values/plain/TestFixedLenByteArrayPlainValuesWriterReader.java`
fail on the unpatched code and pass with this fix. Also verified against
the full-file-roundtrip repro from #3716.

## Are there any user-facing changes?

No API changes. This fixes a silent data-corruption bug introduced in
1.18.0; no user-facing behavior changes other than correct results.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
+++ b/parquet-column/src/main/java/org/apache/parquet/io/api/Binary.java
@@ -442,16 +442,14 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
       if (value.hasArray()) {
         ret = new String(value.array(), value.arrayOffset() + offset, length, StandardCharsets.UTF_8);
       } else {
-        int limit = value.limit();
-        value.limit(offset + length);
-        int position = value.position();
-        value.position(offset);
-        // no corresponding interface to read a subset of a buffer, would have to slice it
-        // which creates another ByteBuffer object or do what is done here to adjust the
-        // limit/offset and set them back after
-        ret = StandardCharsets.UTF_8.decode(value).toString();
-        value.limit(limit);
-        value.position(position);
+        // Duplicate before adjusting position/limit so we never mutate the shared
+        // buffer's own position: readBytes() may have already advanced it past
+        // this value's range (e.g. lazily-consumed values in a repeated field),
+        // and limit(offset + length) would otherwise silently clamp it backwards.
+        ByteBuffer duplicate = value.duplicate();
+        duplicate.position(offset);
+        duplicate.limit(offset + length);
+        ret = StandardCharsets.UTF_8.decode(duplicate).toString();
       }
 
       return ret;
@@ -475,13 +473,18 @@ public abstract class Binary implements Comparable<Binary>, Serializable {
     public byte[] getBytes() {
       byte[] bytes = new byte[length];
 
-      int limit = value.limit();
-      value.limit(offset + length);
-      int position = value.position();
-      value.position(offset);
-      value.get(bytes);
-      value.limit(limit);
-      value.position(position);
+      if (value.hasArray()) {
+        System.arraycopy(value.array(), value.arrayOffset() + offset, bytes, 0, length);
+      } else {
+        // Duplicate before adjusting position/limit so we never mutate the shared
+        // buffer's own position: readBytes() may have already advanced it past
+        // this value's range (e.g. lazily-consumed values in a repeated field),
+        // and limit(offset + length) would otherwise silently clamp it backwards.
+        ByteBuffer duplicate = value.duplicate();
+        duplicate.position(offset);
+        duplicate.limit(offset + length);
+        duplicate.get(bytes);
+      }
       if (!isBackingBytesReused) { // backing buffer might change
         cachedBytes = bytes;
       }
--- a/parquet-column/src/test/java/org/apache/parquet/column/values/plain/TestFixedLenByteArrayPlainValuesWriterReader.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/values/plain/TestFixedLenByteArrayPlainValuesWriterReader.java
@@ -160,4 +160,64 @@ public class TestFixedLenByteArrayPlainValuesWriterReader {
       // Should not throw
     }
   }
+
+  // ---- Lazy getBytes() must not corrupt the shared page buffer's live position ----
+
+  // Regression test: getBytes() on an old value used to shift the buffer's shared
+  // read position backwards, so the next readBytes() call returned stale data.
+  @Test
+  public void testLazyGetBytesDoesNotCorruptSubsequentReadsDirectBuffer() throws IOException {
+    Binary[] expected = {fixedBinary(0), fixedBinary(50), fixedBinary(100), fixedBinary(150)};
+    ByteBuffer direct = writeToDirectBuffer(expected);
+
+    FixedLenByteArrayPlainValuesReader reader = new FixedLenByteArrayPlainValuesReader(FIXED_LEN);
+    reader.initFromPage(expected.length, ByteBufferInputStream.wrap(direct));
+
+    // "row 0": read two values, then materialize them lazily -- getBytes() on the
+    // first value happens after the shared buffer's position has already moved past it.
+    Binary row0v0 = reader.readBytes();
+    Binary row0v1 = reader.readBytes();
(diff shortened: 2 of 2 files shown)


dossett:
Nice catch.  +1 (non-binding). I confirmed the new test code fails under 1.18.0

yimingli-vmware:
> Nice catch. +1 (non-binding). I confirmed the new test code fails under 1.18.0

Hi @dossett, thanks for the confirmation. I wish we could have a patch release for addressing this issue soon, since we are blocked by the Jackson CVEs which are resolved in 1.18.0.

jerolba:
Out of curiosity, was this issue latent in the codebase, or was it introduced around the v1.18.0 changes?"

dossett:
Based on my analysis (with Codex) it was introduced in 1.18.0, specifically in #3565.

I confirmed that 1.18.0 with that PR reverted will pass the new tests.

jerolba:
@dossett Got it, thanks! So the bug in `ByteBufferBackedBinary` was exposed by the updates to `FixedLenByteArrayPlainValuesReader`, right?

dossett:
That's almost a philosophical question about what constitutes a bug :-) But there's definitely some truth to that.

RussellSpitzer (approved):

Fokko (approved):
Thanks for the quick follow up @yimingli-vmware 🙌

yimingli-vmware:
Hi @Fokko and @wgtmac, my PR got a [failed workflow run ](https://github.com/apache/parquet-java/actions/runs/31655606283) since this [Fix Maven parent and JUnit test compatibility](https://github.com/apache/parquet-java/pull/3727) merged into Master but my PR was in the stale state. I've already rebased my PR off Mater, could @wgtmac help me re-trigger my workflow to make sure this PR does not break anything?

And @dossett created another [PR](https://github.com/apache/parquet-java/pull/3729) which solves a similar issue, so may be that PR also needs to be added in the 1.18.1 milestone?

Since i've got 2 approvals and it's my first PR to this project, should i merge this PR, or committers/members will merge this PR? Thanks!

wgtmac (approved):
Thanks @yimingli-vmware for fixing this!

Fokko:
Thanks @yimingli-vmware for fixing this 🙌

dossett:
@Fokko I have a separate PR that, I think, fixes a very similar corruption issue https://github.com/apache/parquet-java/pull/3729

Fokko:
@dossett Let ma take a look 👀  Thanks for pinging me 👍

PR #3726, opened 2026-08-16
GH-3719: Fix vectored read allocation limits and fallback safety

### LINKED ISSUE #3719: Vectored Parquet reads can fall back unsafely after partial asynchronous reads and exceed allocation limits

`ParquetFileReader` can produce unsafe fallback behavior when Hadoop vectored I/O
is enabled and a filesystem partially submits or completes a vectored read before
raising `IllegalArgumentException` or `UnsupportedOperationException`.

The current implementation catches those exceptions around `readVectored(...)`
and retries every range using ordinary reads against the same `ChunkListBuilder`.
If an earlier range already populated the builder, its data is appended again.
For a filtered column with selected pages `P0` and `P2`, the buffered page
sequence can become `[P0, P0, P2]` although the page index still describes
`[P0, P2]`. Depending on the page contents, decoding can fail or silently
associate the wrong page with the selected rows. Even when no page has been
consumed yet, scalar fallback is unsafe once sibling asynchronous reads may still
be operating on the same stream.

Current upstream code:

https://github.com/apache/parquet-java/blob/8e30c4cee3c7e85a8cf2133697f13138509b05b7/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L1293-L1307

The same vectored path also allocates one buffer for an entire contiguous
requested range instead of honoring `parquet.read.allocation.size` (8 MiB by
default), unlike the ordinary read path. Large projected column chunks or
filtered pages can therefore create unexpectedly large heap allocations.

The current `master` branch and Apache Parquet Java 1.18.0 contain this behavior.
The vulnerable path is also present in the 1.15.x, 1.16.x, and 1.17.x release
lines. Vectored I/O defaults to enabled starting in 1.16.0, so supported Hadoop
filesystems can reach this path without an explicit opt-in; in 1.15.2 it is
reachable when explicitly enabled.

Expected behavior:

- Preserve ordinary fallback only when vectored I/O is unavailable or range
  preparation fails before asynchronous submission starts.
- Once submission begins, fail the read safely rather than replaying scalar reads
  against a partially populated builder or an active stream.
- Wait for already-published sibling reads before returning the original failure.
- Split filesystem byte ranges to respect the configured allocation limit without
  changing the logical read plan or decoded results.
- Add regression coverage for partial submission/completion, pending sibling
  futures, filtered pages, oversized columns, and checksum-enabled reads.

### Component(s)

parquet-hadoop

sunchao:
## Why are the changes needed?

Apache Parquet's Hadoop vectored-read path can fall back to ordinary reads after
asynchronous filesystem requests have already been submitted or partially
consumed. Retrying against the same partially populated chunk builder can
duplicate selected page data and return incorrect filtered rows. Outstanding
sibling requests can also continue using a stream while fallback or cleanup
begins.

The same path requests one buffer for an entire contiguous projected range
instead of splitting requests at `parquet.read.allocation.size`, which can cause
unexpectedly large allocations.

Vectored I/O is enabled by default in Apache Parquet 1.16 and later, so these
issues are not limited to applications that explicitly opt in.

## What changes were proposed in this PR?

- Split contiguous requested filesystem ranges at the configured allocation
  limit while preserving the existing logical page and column plan.
- Keep ordinary fallback for failures detected while preparing vectored ranges;
  convert failures after the vectored submission call is attempted into
  `IOException` rather than replaying reads against a partially populated
  builder.
- Await already-published sibling read futures within the existing shared
  timeout before propagating an asynchronous failure.
- Verify dictionary pages and V1/V2 data-page checksums incrementally across
  split buffers instead of allocating another oversized contiguous buffer.
- Add focused vectored-reader and checksum regression coverage.

The current patch still has four known review points that need follow-up:

- Filesystems that align ranges for checksums can allocate more than the
  requested range size, so splitting requests alone does not guarantee that
  every backend allocation respects the configured cap.
- A backend that rejects vectored I/O before submitting work should retain safe
  ordinary-read fallback; the current submission boundary cannot distinguish
  that case from a partially submitted failure.
- Filesystems such as S3A can block while synchronously submitting more ranges
  than their in-flight limit permits, before the existing read timeout applies.
- A backend can start reads and then fail before publishing their futures to
  the wrapper, leaving those reads invisible to the current cleanup path.

## How was this PR tested?

Previously recorded results for this commit on Java 17:

```bash
mvn -pl parquet-hadoop \
  -Dtest=TestParquetFileReaderVectoredIO,TestDataPageChecksums \
  -Dsurefire.failIfNoSpecifiedTests=false \
  test
```

The focused suites reported **37 passing tests**, including **19 vectored-reader
tests** and **18 checksum tests**.

```bash
mvn -pl parquet-hadoop test
mvn -pl parquet-hadoop spotless:check
```

The complete module reported **748 tests, zero failures, zero errors, and 24
existing Hadoop-capability skips**. Apache RAT approved all **256** scanned
licenses, and Spotless passed. The four review points above are not covered by
those recorded passing tests.

Closes #3719.

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -773,6 +773,10 @@ public class ParquetFileReader implements Closeable {
   // not final. in some cases, this may be lazily loaded for backward-compat.
   private ParquetMetadata footer;
 
+  // Some InputFile implementations fetch remote metadata for getLength(). Cache the
+  // vectored range-validation length lazily so ordinary reads need no extra lookup.
+  private long vectoredReadFileLength = -1;
+
   private int currentBlock = 0;
   private ColumnChunkPageReadStore currentRowGroup = null;
   private DictionaryPageReader nextDictionaryReader = null;
@@ -1311,15 +1315,10 @@ public class ParquetFileReader implements Closeable {
 
   /**
    * Should the read use vectored IO?
-   * <p>
-   * This returns true if all necessary conditions are met:
-   * <ol>
-   *   <li> The option is enabled</li>
-   *   <li> The Hadoop version supports vectored IO</li>
-   *   <li> The stream implementation explicitly supports the API; for other streams the classic
-   *         API is always used.</li>
-   *   <li> The allocator is not direct. This is to avoid HADOOP-19101 surfacing.
-   * </ol>
+   * <p>The option must be enabled and the stream's availability probe must accept the
+   * allocator. For Hadoop streams, that probe checks runtime API availability and excludes
+   * direct allocators to avoid HADOOP-19101. It does not guarantee that a particular
+   * vectored-read request will be accepted.
    * @return true or false.
    */
   private boolean shouldUseVectoredIo() {
@@ -1345,7 +1344,10 @@ public class ParquetFileReader implements Closeable {
   private void readVectored(List<ConsecutivePartList> allParts, ChunkListBuilder builder) throws IOException {
     final int maximumAllocation = options.getMaxAllocationSize();
     Preconditions.checkArgument(maximumAllocation > 0, "Invalid maximum allocation size %s", maximumAllocation);
-    final long fileLength = file.getLength();
+    if (vectoredReadFileLength < 0) {
+      vectoredReadFileLength = file.getLength();
+    }
+    final long fileLength = vectoredReadFileLength;
     List<ParquetFileRange> ranges = new ArrayList<>(allParts.size());
     List<Integer> partRangeCounts = new ArrayList<>(allParts.size());
     long totalSize = 0;
@@ -1372,8 +1374,9 @@ public class ParquetFileReader implements Closeable {
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
     final long readStart = System.nanoTime();
     try {
-      // Even a synchronous failure can occur after some reads have been scheduled,
-      // so falling back to normal IO is unsafe once this call has been entered.
+      // Even a synchronous rejection can follow partial submission. The Hadoop bridge
+      // publishes futures only after submission returns, so missing futures do not prove
+      // that no reads started. Once this call is entered, normal-read fallback is unsafe.
       f.readVectored(ranges, options.getAllocator());
       int firstRange = 0;
       for (int partIndex = 0; partIndex < allParts.size(); partIndex++) {
@@ -1382,6 +1385,7 @@ public class ParquetFileReader implements Closeable {
         firstRange = endRange;
       }
     } catch (IllegalArgumentException | UnsupportedOperationException e) {
+      // Consumption may also have populated the builder. Do not replay those chunks.
       IOException failure =
           new IOException("Vectored read failed after asynchronous reads may have been submitted", e);
       awaitRemainingVectoredReads(ranges, readStart, failure);
--- a/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
@@ -36,6 +36,7 @@ import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.io.SequenceInputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
@@ -1293,14 +1294,14 @@ public class ParquetFileReader implements Closeable {
   private void readAllPartsVectoredOrNormal(List<ConsecutivePartList> allParts, ChunkListBuilder builder)
       throws IOException {
 
-    if (shouldUseVectoredIo(allParts)) {
+    if (shouldUseVectoredIo()) {
       try {
         readVectored(allParts, builder);
         return;
       } catch (IllegalArgumentException | UnsupportedOperationException e) {
-        // Either the arguments are wrong or somehow this is being invoked against
-        // a hadoop release which doesn't have the API and yet somehow it got here.
-        LOG.warn("readVectored() failed; falling back to normal IO against {}", f, e);
+        // At this point only range preparation can have failed; exceptions from the
+        // vectored call itself are wrapped below because reads may already be active.
+        LOG.warn("Preparing vectored reads failed; falling back to normal IO against {}", f, e);
       }
     }
     for (ConsecutivePartList consecutiveChunks : allParts) {
@@ -1315,37 +1316,14 @@ public class ParquetFileReader implements Closeable {
    * <ol>
    *   <li> The option is enabled</li>
    *   <li> The Hadoop version supports vectored IO</li>
-   *   <li> The part lengths are all valid for vectored IO</li>
    *   <li> The stream implementation explicitly supports the API; for other streams the classic
    *         API is always used.</li>
    *   <li> The allocator is not direct. This is to avoid HADOOP-19101 surfacing.
    * </ol>
-   * @param allParts all parts to read.
    * @return true or false.
    */
-  private boolean shouldUseVectoredIo(final List<ConsecutivePartList> allParts) {
-    return options.useHadoopVectoredIo()
-        && f.readVectoredAvailable(options.getAllocator())
-        && arePartsValidForVectoredIo(allParts);
-  }
-
-  /**
-   * Validate the parts for vectored IO.
-   * Vectored IO doesn't support reading ranges of size greater than
-   * Integer.MAX_VALUE.
-   * @param allParts all parts to read.
-   * @return true or false.
-   */
-  private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
-    for (ConsecutivePartList consecutivePart : allParts) {
-      if (consecutivePart.length >= Integer.MAX_VALUE) {
-        LOG.debug(
-            "Part length {} greater than Integer.MAX_VALUE thus disabling vectored IO",
-            consecutivePart.length);
-        return false;
-      }
-    }
-    return true;
+  private boolean shouldUseVectoredIo() {
+    return options.useHadoopVectoredIo() && f.readVectoredAvailable(options.getAllocator());
   }
 
   /**
@@ -1357,32 +1335,100 @@ public class ParquetFileReader implements Closeable {
    * If directly implemented by a Filesystem then it is likely to be a more efficient
    * operation such as a scatter-gather read (native IO) or set of parallel
    * GET requests against an object store.
+   * The allocation limit applies to filesystem buffers; decoders can still require a
+   * contiguous buffer for an individual logical value larger than that limit.
    * @param allParts all parts to be read.
    * @param builder used to build chunk list to read the pages for the different columns.
-   * @throws IOException any IOE.
-   * @throws IllegalArgumentException arguments are invalid.
-   * @throws UnsupportedOperationException if the filesystem does not support vectored IO.
+   * @throws IOException if submitting or consuming the vectored reads fails.
+   * @throws IllegalArgumentException if range preparation fails before any reads are submitted.
    */
   private void readVectored(List<ConsecutivePartList> allParts, ChunkListBuilder builder) throws IOException {
-
+    final int maximumAllocation = options.getMaxAllocationSize();
+    Preconditions.checkArgument(maximumAllocation > 0, "Invalid maximum allocation size %s", maximumAllocation);
+    final long fileLength = file.getLength();
     List<ParquetFileRange> ranges = new ArrayList<>(allParts.size());
+    List<Integer> partRangeCounts = new ArrayList<>(allParts.size());
     long totalSize = 0;
     for (ConsecutivePartList consecutiveChunks : allParts) {
       final long len = consecutiveChunks.length;
-      Preconditions.checkArgument(
-          len < Integer.MAX_VALUE,
-          "Invalid length %s for vectored read operation. It must be less than max integer value.",
-          len);
-      ranges.add(new ParquetFileRange(consecutiveChunks.offset, (int) len));
+      final long start = consecutiveChunks.offset;
+      if (start < 0 || len < 0 || start > fileLength || len > fileLength - start) {
+        throw new IOException(String.format(
+            "Invalid vectored read range (offset %d, length %d) for file length %d",
+            start, len, fileLength));
+      }
+      final int firstRange = ranges.size();
+      long remaining = len;
+      long offset = start;
+      do {
+        int rangeLength = (int) Math.min(remaining, maximumAllocation);
+        ranges.add(new ParquetFileRange(offset, rangeLength));
+        offset += rangeLength;
+        remaining -= rangeLength;
+      } while (remaining > 0);
+      partRangeCounts.add(ranges.size() - firstRange);
       totalSize += len;
     }
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
-    // Request a vectored read;
-    f.readVectored(ranges, options.getAllocator());
-    int k = 0;
-    for (ConsecutivePartList consecutivePart : allParts) {
-      ParquetFileRange currRange = ranges.get(k++);
-      consecutivePart.readFromVectoredRange(currRange, builder);
+    final long readStart = System.nanoTime();
+    try {
+      // Even a synchronous failure can occur after some reads have been scheduled,
+      // so falling back to normal IO is unsafe once this call has been entered.
+      f.readVectored(ranges, options.getAllocator());
+      int firstRange = 0;
+      for (int partIndex = 0; partIndex < allParts.size(); partIndex++) {
+        int endRange = firstRange + partRangeCounts.get(partIndex);
+        allParts.get(partIndex).readFromVectoredRanges(ranges.subList(firstRange, endRange), builder);
+        firstRange = endRange;
+      }
+    } catch (IllegalArgumentException | UnsupportedOperationException e) {
+      IOException failure =
+          new IOException("Vectored read failed after asynchronous reads may have been submitted", e);
+      awaitRemainingVectoredReads(ranges, readStart, failure);
+      throw failure;
+    } catch (IOException | RuntimeException e) {
+      awaitRemainingVectoredReads(ranges, readStart, e);
+      throw e;
+    }
+  }
+
+  /**
+   * Wait for submitted reads with published futures to finish before their stream can be
+   * closed. Cancelling result futures does not stop all Hadoop backends from continuing IO.
+   */
+  private void awaitRemainingVectoredReads(List<ParquetFileRange> ranges, long readStart, Throwable failure) {
+    if (Thread.currentThread().isInterrupted()
+        || failure instanceof InterruptedIOException && failure.getCause() instanceof InterruptedException) {
+      return;
+    }
+
+    final long timeoutNanos = TimeUnit.SECONDS.toNanos(HADOOP_VECTORED_READ_TIMEOUT_SECONDS);
+    for (ParquetFileRange range : ranges) {
+      Future<ByteBuffer> future = range.getDataReadFuture();
+      if (future == null || future.isDone()) {
+        continue;
+      }
+
+      long remainingNanos = Math.max(timeoutNanos - (System.nanoTime() - readStart), 0L);
+      try {
+        FutureIO.awaitFuture(future, remainingNanos, TimeUnit.NANOSECONDS);
+      } catch (InterruptedIOException e) {
+        if (failure != e) {
+          failure.addSuppressed(e);
+        }
+        if (e.getCause() instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+          return;
+        }
+      } catch (TimeoutException e) {
+        failure.addSuppressed(e);
+        LOG.warn("Timed out waiting for vectored read {} after another read failed", range, e);
+        return;
+      } catch (IOException | RuntimeException e) {
+        if (failure != e) {
+          failure.addSuppressed(e);
+        }
+      }
     }
   }
 
@@ -1951,10 +1997,12 @@ public class ParquetFileReader implements Closeable {
      * Calculate checksum of input bytes, throw decoding exception if it does not match the provided
      * reference crc
      */
-    private void verifyCrc(int referenceCrc, BytesInput bytes, String exceptionMsg) {
+    private void verifyCrc(int referenceCrc, String exceptionMsg, BytesInput... inputs) throws IOException {
       crc.reset();
-      try (ByteBufferReleaser releaser = crcAllocator.getReleaser()) {
-        crc.update(bytes.toByteBuffer(releaser));
+      for (BytesInput input : inputs) {
+        for (ByteBuffer buffer : input.toInputStream().remainingBuffers()) {
+          crc.update(buffer);
+        }
       }
       if (crc.getValue() != ((long) referenceCrc & 0xffffffffL)) {
         throw new ParquetDecodingException(exceptionMsg);
@@ -2021,8 +2069,8 @@ public class ParquetFileReader implements Closeable {
             if (options.usePageChecksumVerification() && pageHeader.isSetCrc()) {
               verifyCrc(
                   pageHeader.getCrc(),
-                  pageBytes,
-                  "could not verify dictionary page integrity, CRC checksum verification failed");
+                  "could not verify dictionary page integrity, CRC checksum verification failed",
+                  pageBytes);
             }
             DictionaryPageHeader dicHeader = pageHeader.getDictionary_page_header();
             dictionaryPage = new DictionaryPage(
@@ -2041,8 +2089,8 @@ public class ParquetFileReader implements Closeable {
             if (options.usePageChecksumVerification() && pageHeader.isSetCrc()) {
               verifyCrc(
                   pageHeader.getCrc(),
-                  pageBytes,
-                  "could not verify page integrity, CRC checksum verification failed");
+                  "could not verify page integrity, CRC checksum verification failed",
+                  pageBytes);
             }
             DataPageV1 dataPageV1 = new DataPageV1(
                 pageBytes,
@@ -2072,11 +2120,12 @@ public class ParquetFileReader implements Closeable {
                 this.readAsBytesInput(dataHeaderV2.getDefinition_levels_byte_length());
             final BytesInput values = this.readAsBytesInput(dataSize);
             if (options.usePageChecksumVerification() && pageHeader.isSetCrc()) {
-              pageBytes = BytesInput.concat(repetitionLevels, definitionLevels, values);
               verifyCrc(
                   pageHeader.getCrc(),
-                  pageBytes,
-                  "could not verify page integrity, CRC checksum verification failed");
+                  "could not verify page integrity, CRC checksum verification failed",
+                  repetitionLevels,
+                  definitionLevels,
+                  values);
             }
             DataPageV2 dataPageV2 = new DataPageV2(
                 dataHeaderV2.getNum_rows(),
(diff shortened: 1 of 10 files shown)


parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1335,100 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
-   * @throws IllegalArgumentException arguments are invalid.
-   * @throws UnsupportedOperationException if the filesystem does not support vectored IO.
+   * @throws IOException if submitting or consuming the vectored reads fails.
+   * @throws IllegalArgumentException if range preparation fails before any reads are submitted.
    */
   private void readVectored(List<ConsecutivePartList> allParts, ChunkListBuilder builder) throws IOException {
-
+    final int maximumAllocation = options.getMaxAllocationSize();
+    Preconditions.checkArgument(maximumAllocation > 0, "Invalid maximum allocation size %s", maximumAllocation);
+    final long fileLength = file.getLength();
```
dossett: Adding `getLength()` to every vectored read could be expensive on object stores (unlike local HDFS).

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1335,100 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+      // Even a synchronous failure can occur after some reads have been scheduled,
+      // so falling back to normal IO is unsafe once this call has been entered.
+      f.readVectored(ranges, options.getAllocator());
+      int firstRange = 0;
+      for (int partIndex = 0; partIndex < allParts.size(); partIndex++) {
+        int endRange = firstRange + partRangeCounts.get(partIndex);
+        allParts.get(partIndex).readFromVectoredRanges(ranges.subList(firstRange, endRange), builder);
+        firstRange = endRange;
+      }
+    } catch (IllegalArgumentException | UnsupportedOperationException e) {
```
dossett, line 1436: An `UnsupportedOperationException` will now be rethrown as an `IOExeption`.  If I understand correctly, on `master` this will now kill the read entirely instead of resulting in a warning and a fallback to non-vectored IO.  If so, that seems like a meaningful change.

Maybe I'm missing something, or maybe this just wouldn't happen in practice?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Thanks, addressed in [ad623d6](https://github.com/apache/parquet-java/pull/3726/commits/ad623d6f4a44934ee2ccb98ab28bfee7242b0885). `HadoopInputFile` already reads its cached `FileStatus`, but other `InputFile` implementations need not do so. I now cache the length lazily for vectored-range validation, so it adds at most one length lookup per reader instead of one per row-group read. Ordinary reads, including those with a supplied footer, do not acquire an extra lookup.

Added regression tests for repeated filtered/unfiltered reads and both footer paths. All 40 focused tests pass; the full `parquet-hadoop` module reports 751 tests, zero failures or errors, and 24 Hadoop-capability skips. Spotless and RAT also pass.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Yes, this is a deliberate and meaningful behavior change. Even a synchronous `UnsupportedOperationException` can follow partial submission, and the Hadoop bridge only copies the futures back to Parquet after submission returns. During result consumption, the chunk builder may also already contain data. Falling back in either situation can race outstanding reads or replay chunks and return incorrect rows.

Ordinary I/O still handles disabled/unavailable vectored I/O, and preparation-time `IllegalArgumentException`/`UnsupportedOperationException` still permits fallback. I clarified that boundary in [ad623d6](https://github.com/apache/parquet-java/pull/3726/commits/ad623d6f4a44934ee2ccb98ab28bfee7242b0885) and corrected the availability documentation: the Hadoop bridge checks runtime API/allocator support, not per-stream `hasCapability`. The tests cover ordinary-read selection plus synchronous, asynchronous, and partial-submission failures without fallback.

I do not have an observed production rejection from an otherwise supported backend. A genuinely pre-submission rejection should be recoverable, but the current interface does not reliably distinguish it from partial submission. Restoring broader fallback would need an explicit no-work-submitted guarantee; I have kept the conservative behavior for now.

dossett: Thank you for the thoughtful reply.  I'll defer to others more knowledgeable than myself about the tradeoffs here, but I've learned a lot.

sunchao:
cc @wgtmac can you take a look?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -2343,32 +2396,39 @@ private void setReadMetrics(long startNs, long len) {
...
-        buffer = FutureIO.awaitFuture(currRange.getDataReadFuture(), timeoutSeconds, TimeUnit.SECONDS);
-        setReadMetrics(readStart, currRange.getLength());
+        for (ParquetFileRange range : ranges) {
+          currentRange = range;
+          LOG.debug(
+              "Waiting for vectored read to finish for range {} with timeout {} seconds",
+              range,
+              timeoutSeconds);
+          long remainingNanos = Math.max(timeoutNanos - (System.nanoTime() - readStart), 0L);
+          buffers.add(FutureIO.awaitFuture(range.getDataReadFuture(), remainingNanos, TimeUnit.NANOSECONDS));
```
wgtmac: `readFromVectoredRanges()` never registers returned buffers with `builder.addBuffersToRelease(...)`; tracking/pooling allocators therefore leak them.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1334,105 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+      for (int partIndex = 0; partIndex < allParts.size(); partIndex++) {
+        int endRange = firstRange + partRangeCounts.get(partIndex);
+        allParts.get(partIndex).readFromVectoredRanges(ranges.subList(firstRange, endRange), builder);
+        firstRange = endRange;
+      }
+    } catch (IllegalArgumentException | UnsupportedOperationException e) {
+      // Consumption may also have populated the builder. Do not replay those chunks.
+      IOException failure =
+          new IOException("Vectored read failed after asynchronous reads may have been submitted", e);
+      awaitRemainingVectoredReads(ranges, readStart, failure);
```
wgtmac: After partial submission, futures are copied back to `ParquetFileRange` only when `readWrappedRanges()` returns. A backend failure can leave started reads untracked, so cleanup cannot wait for them.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1334,105 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
       totalSize += len;
     }
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
-    // Request a vectored read;
-    f.readVectored(ranges, options.getAllocator());
-    int k = 0;
-    for (ConsecutivePartList consecutivePart : allParts) {
-      ParquetFileRange currRange = ranges.get(k++);
-      consecutivePart.readFromVectoredRange(currRange, builder);
+    final long readStart = System.nanoTime();
```
wgtmac: The timeout starts only after `f.readVectored(...)` returns, so a blocking submission is outside the 300-second limit.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1334,105 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
-   * @throws UnsupportedOperationException if the filesystem does not support vectored IO.
+   * @throws IOException if submitting or consuming the vectored reads fails.
+   * @throws IllegalArgumentException if range preparation fails before any reads are submitted.
    */
   private void readVectored(List<ConsecutivePartList> allParts, ChunkListBuilder builder) throws IOException {
-
+    final int maximumAllocation = options.getMaxAllocationSize();
+    Preconditions.checkArgument(maximumAllocation > 0, "Invalid maximum allocation size %s", maximumAllocation);
+    if (vectoredReadFileLength < 0) {
+      vectoredReadFileLength = file.getLength();
```
wgtmac: `file.getLength()` can fail before submission, but that `IOException` bypasses the ordinary-read fallback.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1334,105 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+    }
+  }
+
+  /**
+   * Wait for submitted reads with published futures to finish before their stream can be
+   * closed. Cancelling result futures does not stop all Hadoop backends from continuing IO.
+   */
+  private void awaitRemainingVectoredReads(List<ParquetFileRange> ranges, long readStart, Throwable failure) {
+    if (Thread.currentThread().isInterrupted()
+        || failure instanceof InterruptedIOException && failure.getCause() instanceof InterruptedException) {
```
wgtmac, line 1485: An interrupted wait returns before sibling futures are drained; `close()` may then race with backend reads.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Fixed in [f82eaa42](https://github.com/apache/parquet-java/pull/3726/commits/f82eaa428528849cf019c99410fe879c4d641875). Vectored reads now track the original allocations and transfer them to the row group's releaser after successful completion. This includes extra checksum buffers and buffers whose returned results are slices, so the allocator receives the original objects it created. The row group is also closed on read/parsing failures.

Added ownership tests for filtered and unfiltered reads, slices, checksum allocations, and parse failures. Real Hadoop 3.4.1 raw/checksum filesystem probes decoded all 128 rows with balanced allocations/releases and a clean tracking allocator.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Addressed in [f82eaa42](https://github.com/apache/parquet-java/pull/3726/commits/f82eaa428528849cf019c99410fe879c4d641875). The bridge now publishes available futures in a `finally` block, without masking the original submission failure. A failed submission still fails the read without replaying ordinary I/O.

Stream closure is queued behind the actual submission call. Buffer release requires every range to have a published, completed, non-cancelled future. Hadoop can publish futures for work it never schedules, so cleanup does not block indefinitely waiting for those; the batch retains its remaining allocations if completion cannot be established.

Added regressions for partial submission and a future-publication failure that must not replace the submission exception.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Fixed in [f82eaa42](https://github.com/apache/parquet-java/pull/3726/commits/f82eaa428528849cf019c99410fe879c4d641875). Submission now runs on a reader-owned worker, and one 300-second budget per vectored-read batch starts before submission and is shared with every range wait. It is no longer reset for each part.

On timeout, the reader is invalidated and cleanup is queued on the same worker, so cancelling the submission `Future` cannot let cleanup overtake a backend call that is still running. The caller can return even if the backend ignores interruption; its resources remain owned by deferred cleanup, which may itself be delayed indefinitely.

Added tests for blocked submission, the shared deadline, and submission that continues after cancellation. These are local regression tests; I have not validated this against live S3.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Fixed in [f82eaa42](https://github.com/apache/parquet-java/pull/3726/commits/f82eaa428528849cf019c99410fe879c4d641875). The length lookup now has a narrow pre-submission fallback: a metadata `IOException` disables vectored I/O for that reader and continues with ordinary reads on the already-open stream. This avoids repeated failed lookups on later row groups. Interruption still propagates, and failures after submission still cannot trigger ordinary-read replay.

Added regression tests with and without a supplied footer, plus an interrupted-lookup case. Both fallback regressions fail on the original commit and pass with this change.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Adjusted in [f82eaa42](https://github.com/apache/parquet-java/pull/3726/commits/f82eaa428528849cf019c99410fe879c4d641875). An interrupted wait still propagates promptly as `InterruptedIOException`; it does not block draining siblings. The failed reader is invalidated and its original stream is handed to deferred cleanup, which cannot overtake a submission call still running after cancellation.

Cleanup may close a stream with asynchronous reads still pending. Close or cancellation is not treated as proof that I/O stopped: buffer release requires every range to have a published, completed, non-cancelled future. Unknown completion retains the batch's allocations.

Added tests for a submission that ignores interruption and for a pending read that completes after stream closure, checking that its buffer is not released early.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1364,153 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
     }
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
-    // Request a vectored read;
-    f.readVectored(ranges, options.getAllocator());
-    int k = 0;
-    for (ConsecutivePartList consecutivePart : allParts) {
-      ParquetFileRange currRange = ranges.get(k++);
-      consecutivePart.readFromVectoredRange(currRange, builder);
+    if (vectoredReadExecutor == null) {
+      vectoredReadExecutor = Executors.newSingleThreadExecutor(task -> {
```
wgtmac: Each reader keeps one worker thread after its first vectored read. Could this be expensive when many readers are open?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1310,42 +1324,35 @@ private void readAllPartsVectoredOrNormal(List<ConsecutivePartList> allParts, Ch
...
-        LOG.debug(
-            "Part length {} greater than Integer.MAX_VALUE thus disabling vectored IO",
-            consecutivePart.length);
-        return false;
+  private boolean prepareVectoredReadFileLength() throws IOException {
+    if (vectoredReadFileLength >= 0) {
+      return true;
+    }
+    try {
+      vectoredReadFileLength = file.getLength();
```
wgtmac: I think this whole length lookup can go away.

The only thing it feeds is the bounds check at L1387. Those offsets come from our own footer. If they are out of range the file is already broken and the read will fail anyway, so the check does not buy us much. The condition we actually needed before was `len < Integer.MAX_VALUE`, and after the split at L1396 that holds by construction.

Dropping `getLength()` also drops `vectoredReadFileLength`, `vectoredIoDisabled`, the interrupt handling here, and a good number of tests. It closes two of the review threads above at the same time. Much less code to reason about.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1364,153 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+      if (start < 0 || len < 0 || start > fileLength || len > fileLength - start) {
+        throw new IOException(String.format(
+            "Invalid vectored read range (offset %d, length %d) for file length %d",
+            start, len, fileLength));
+      }
+      final int firstRange = ranges.size();
+      long remaining = len;
+      long offset = start;
+      do {
+        int rangeLength = (int) Math.min(remaining, maximumAllocation);
```
wgtmac, line 1388: Hadoop checksum reads can merge ranges and allocate larger buffers than `maximumAllocation`. Is the allocation limit only best effort for such filesystems?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1364,153 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+        allParts.get(partIndex)
+            .readFromVectoredRanges(ranges.subList(firstRange, endRange), builder, operation);
+        firstRange = endRange;
+      }
+      operation.transferTo(builder.releaser);
+    } catch (TimeoutException e) {
+      IOException failure = new IOException("Timed out submitting vectored reads", e);
+      abortVectoredRead(operation, failure);
+      throw failure;
+    } catch (IllegalArgumentException | UnsupportedOperationException e) {
```
wgtmac: +1 to @dossett here, and I do not think this is settled yet.

Today a filesystem that rejects vectored IO gets a warning and a normal read. After this change the query fails. You said the current API cannot tell a pre-submission rejection from a partial submission, and I agree. But then we are trading a correctness bug for an availability regression, and users on such a filesystem will only find out in production.

Two things I would like before this goes in:

- a committer says yes to the behavior change explicitly, not just "defer to others"
- the description says it is a behavior change, so it shows up in the release notes

One more idea: keep the fallback, but make it safe by building a fresh `ChunkListBuilder` instead of reusing the dirty one, and only allow it when nothing has been consumed yet. That fixes the reported bug without changing what users see. It does not help when sibling reads are still on the stream, but in a real pre-submission rejection there are no sibling reads. Did you look at that option?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1357,32 +1364,153 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
+      }
+      abortVectoredRead(operation, e);
+      throw e;
+    }
+  }
+
+  private void abortVectoredRead(VectoredReadOperation operation, Throwable failure) {
+    // Deferred cleanup now owns the original stream and executor. Prevent a later
+    // call (including an attempted ordinary read) from racing that cleanup.
+    f =
```
wgtmac: Swapping `f` for a stream that throws on every call is too clever for my taste.

`f` is protected, so subclasses can see it. And after this, `close()` closes the fake stream instead of the real one, which is surprising when you read `close()` on its own.

A plain flag checked at the entry points would do the same job and read much better:

```java
if (vectoredReadFailed) {
  throw new IOException("Cannot reuse a reader after a vectored read failure", vectoredReadFailure);
}
```

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1860,6 +2002,9 @@ public void close() throws IOException {
         f.close();
       }
     } finally {
+      if (vectoredReadExecutor != null) {
```
wgtmac: After `abortVectoredRead()` this field is null, so `close()` does nothing here, and the real stream is only closed by the task queued in `VectoredReadOperation.abort()`. If the backend never returns from `readVectored`, that task never runs, and we hold the file descriptor and the daemon thread until the JVM exits.

Your own comment in `VectoredReadOperation` says the cleanup "may itself be delayed indefinitely", so I think you already know this. Still, an unbounded leak is a lot to accept inside a bug fix. Can we bound the wait, log at warn when we give up, and write it down in the `readVectored` javadoc so users know what they are getting?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1951,10 +2096,12 @@ protected PageHeader readPageHeader(BlockCipher.Decryptor blockDecryptor, byte[]
      * Calculate checksum of input bytes, throw decoding exception if it does not match the provided
      * reference crc
      */
-    private void verifyCrc(int referenceCrc, BytesInput bytes, String exceptionMsg) {
+    private void verifyCrc(int referenceCrc, String exceptionMsg, BytesInput... inputs) throws IOException {
       crc.reset();
-      try (ByteBufferReleaser releaser = crcAllocator.getReleaser()) {
-        crc.update(bytes.toByteBuffer(releaser));
+      for (BytesInput input : inputs) {
+        for (ByteBuffer buffer : input.toInputStream().remainingBuffers()) {
```
wgtmac, line 2099: Is `crcAllocator` still needed? I only find its initialization and close, but no use.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/VectoredReadOperation.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+      });
+    } catch (RejectedExecutionException cleanupFailure) {
+      // The reader owns this executor and must not shut it down before queuing cleanup.
+      // Do not mask the read error or recycle buffers whose IO lifetime is now unknown.
+      failure.addSuppressed(cleanupFailure);
+      LOG.warn("Could not schedule cleanup after a vectored read failure", cleanupFailure);
+    } finally {
+      // shutdownNow would discard the queued cleanup or interrupt its close operation.
+      executor.shutdown();
```
wgtmac: Ownership of the executor is split across two classes and it is hard to follow. `ParquetFileReader` creates it, sets its own field to null in `abortVectoredRead()`, and then this class shuts it down, so the reader's `close()` can no longer touch it.

If we keep the executor at all (see my comment where it is created), can it live entirely in this class? Created here, shut down here.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/VectoredReadBufferAllocator.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Owns the original allocations for one vectored read. Filesystems may return slices of these buffers and allocate
+ * additional buffers for checksums, so the range results do not identify all the allocations that need releasing.
+ *
+ * <p>The caller must wait for submission and all reads to finish before transferring or closing this owner. Stopping
+ * allocations, cancelling a result future, or closing the stream does not establish that the filesystem has stopped
+ * using a previously allocated buffer.
+ */
+final class VectoredReadBufferAllocator implements ByteBufferAllocator, AutoCloseable {
```
wgtmac: I agree we need to track what the filesystem allocates. It can hand back slices and it can allocate extra buffers for checksums, so registering the returned buffers is not enough. A wrapper allocator is the right idea.

But the ordinary path just calls `builder.addBuffersToRelease(buffers)`. Could this class be only the tracking part, pass the tracked buffers to `addBuffersToRelease()` on success, and drop the rest? `stopAllocating`, `transferTo`, `close` and the flags around them exist mostly to support the deferred release after a failure, and that only exists because of the async cleanup path. If the executor goes away, most of this can go away with it.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
wgtmac: I think fail-fast is safer here. We cannot reliably tell a rejection before submission from a partial submission. This may change behavior for some backends, but I am okay with this trade-off.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Addressed in [5d3d6f3e](https://github.com/apache/parquet-java/commit/5d3d6f3eb37f48cb46f28d4f040b2dc2ecfb6828). The executor now has zero core workers and a 60-second idle timeout, so an open reader no longer keeps a worker indefinitely between reads. A later submission recreates the worker.

It still has at most one worker per reader and a FIFO queue: failure cleanup must remain behind a submission that continues running after cancellation. An actively blocked backend call therefore still retains its worker. Added an expiry/recreation regression, and the existing timeout/cancellation cleanup tests now use the production executor factory.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Yes: the strict bound is on the length of each range Parquet requests, not on every buffer the filesystem allocates. Checksum alignment and range merging can produce larger allocations. This is now explicit in the configuration README, `ParquetReadOptions` API documentation, and reader documentation in [5d3d6f3e](https://github.com/apache/parquet-java/commit/5d3d6f3eb37f48cb46f28d4f040b2dc2ecfb6828). Footer/decoder allocations and total memory are outside this setting as well.

Strengthened the merged-buffer regression to verify that requested ranges stay within 128 bytes while the backend allocates a larger merged buffer, all rows decode correctly, and the original merged/checksum buffers are released. The PR description now states this limitation rather than implying a hard filesystem allocation cap.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Removed in [5d3d6f3e](https://github.com/apache/parquet-java/commit/5d3d6f3eb37f48cb46f28d4f040b2dc2ecfb6828), including the field, import, all constructor initialization, and close argument. Incremental checksum verification already updates CRC directly from the existing buffer views, so the scratch allocator is unnecessary.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
sunchao: Kept the fail-fast behavior after submission is attempted. The updated PR description now records this accepted trade-off and distinguishes it from safe pre-submission metadata/preparation fallback.

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -897,10 +901,8 @@ public ParquetFileReader(Configuration conf, Path file, ParquetMetadata footer)
 
     if (options.usePageChecksumVerification()) {
       this.crc = new CRC32();
```
wgtmac: The idle timeout removes the permanent idle thread, but a blocked backend call still uses one worker per reader with no global bound. Is this acceptable for workloads with many concurrent readers?

parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java
```diff
@@ -1358,32 +1355,150 @@ private boolean arePartsValidForVectoredIo(List<ConsecutivePartList> allParts) {
...
     }
     LOG.debug("Reading {} bytes of data with vectored IO in {} ranges", totalSize, ranges.size());
-    // Request a vectored read;
-    f.readVectored(ranges, options.getAllocator());
-    int k = 0;
-    for (ConsecutivePartList consecutivePart : allParts) {
-      ParquetFileRange currRange = ranges.get(k++);
-      consecutivePart.readFromVectoredRange(currRange, builder);
+    if (vectoredReadExecutor == null) {
+      vectoredReadExecutor = VectoredReadOperation.newExecutor();
```
wgtmac, line 1398: A blocked backend call still uses one worker per reader with no global bound. Is this acceptable for workloads with many concurrent readers?

wgtmac:
Do you want to take a look as well? @gszadovszky @Fokko @divjotarora

divjotarora:
@wgtmac I'd be interested in reviewing, but I'm a bit swamped right now. Can get to it tomorrow if that's OK

gszadovszky (approved):
LGTM

PR #3728, opened 2026-08-17
Add AGENTS.md with AssertJ/JUnit5 test migration guidelines

nastra:
### Rationale for this change
This adds guidelines that I've been defining/using for Claude when migrating the entire codebase from JUnit4 to JUnit5 + AssertJ

An AGENTS.md file that contains migration guidelines


AGENTS.md
```diff
@@ -0,0 +1,923 @@
+# AssertJ test migration rules
```
wgtmac, line 20: Do we need to add license header or add a RAT rule to exclude this file?

nastra: oops yes, this needs a license header. added one

dossett:
I'm glad to see an AGENTS.md file added, this is great.  This is a LOT of context about writing unit tests, which is also great.  But it might be too much context for the agent to always have, even when doing things not related to unit tests.  A pattern I've found very useful is to create some agent-specific documentation in a separate file and add a pointer to that in AGENTS.md.  Then the agent always has the context of where to find the information when it needs it.

```
## Unit tests
When writing, updating, or migrating unit tests please read the reference document [e.g., `./agents/docs/unit-testing.md`]
```

Something like that

AGENTS.md
divjotarora, line 1: I agree with @dossett that this is likely too much testing-specific info in a top-level `AGENTS.md`. The suggestion to move it to a separate file that can lazy loaded only when writing tests sounds reasonable. Additionally, we're mixing a lot of information about writing new tests vs. migrating existing ones. Perhaps it would make sense to separate these? Even if all of this info goes into a `TESTING_GUIDELINES.md` file or something but in separate sections, it would be more readable.

Lastly, there's a few "MUST" rules in here (e.g. every `assertThatThrownBy` must contain an `isInstanceOf` chain + msg check). Do we have linter rules for these rather than just putting them in agent guidelines?

divjotarora (commented):
Thanks for working on this @nastra!

AGENTS.md
nastra: I've updated this and moved things into separate files, which are referenced from the main AGENTS.md.

> Lastly, there's a few "MUST" rules in here (e.g. every assertThatThrownBy must contain an isInstanceOf chain + msg check). Do we have linter rules for these rather than just putting them in agent guidelines?

The project doesn't have a linter that would be able to enforce this unfortunately

nastra:
@wgtmac @Fokko can you guys please take another look at this?

divjotarora (approved):

wgtmac (approved):

Fokko (approved):
I'm not super familiar with these kinds of docs (which are mainly for LLMs), but I think this is a good start. Thanks @nastra for working on this 🙌

dossett:
Thank you @nastra, love the split out.  It will be exciting to have a top level AGENTS.md file that we can start to shape.

One additional request: symlinking this to CLAUDE.md can be helpful for Claude Code users.

reference: https://code.claude.com/docs/en/memory#agents-md

PR #3729, opened 2026-08-17
`toByteArray` should return a copy not a reference

dossett:
#3565 introduced several optimizations, but I think it also introduced an issue that PR seeks to address.  Basically `BytesInput.copy(BytesInput)` has a public contract to return a copy, but a change to `BytesInput.toByteArray()` led to returning a reference in some cases.  In addition to breaking the public contract, this can lead to data corruption.

See also #3717

### DIFF
--- a/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
+++ b/parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
@@ -690,12 +690,6 @@ public abstract class BytesInput {
       return java.nio.ByteBuffer.wrap(in, offset, length);
     }
 
-    @SuppressWarnings("deprecation")
-    @Override
-    public byte[] toByteArray() {
-      return Arrays.copyOfRange(in, offset, offset + length);
-    }
-
     @Override
     public long size() {
       return length;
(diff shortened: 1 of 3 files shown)


parquet-common/src/main/java/org/apache/parquet/bytes/BytesInput.java
```diff
@@ -690,18 +690,9 @@ public ByteBuffer toByteBuffer() throws IOException {
       return java.nio.ByteBuffer.wrap(in, offset, length);
     }
 
-    /**
-     * Zero-copy override: returns the backing array directly when fully used,
-     * skipping the base-class BAOS allocation + copy on every decompressor call.
-     * Returning the mutable array is safe — the base class already exposes a
-     * mutable {@code BAOS.getBuf()}.
```
dossett, line 697: I think `Returning the mutable array is safe` is the key.  The array was mutable before, but it was a mutable _copy_ of the original array.

Fokko: Should we revert the whole `toByteArray` method? This effectively reverts the part in:

https://github.com/apache/parquet-java/pull/3565/changes#diff-49155017f14a28333fad01fe81e8c7cd353cdc73824360db5d8ba15b7e7dd6daR693-R707

It is also deprecated

dossett: Done!

Fokko (approved):
Thanks for fixing this issue and removing the override, less is more 👍

PR #3730, opened 2026-08-17
MINOR: Upgrade thrift to 0.24

dossett:
I was testing a PR and couldn't `brew install` thrift 0.23, because only 0.24 is available.  So this PR does two things:

- bump to 0.24
- add an experimental `Dockerfile` that developers can use that removes the dependency on a locally installed copy of thrift (and any other dependency).

I find solutions like the Dockerfile to be really useful, but if maintainers feel its not appropriate to add to the repo I would certainly understand.  The part can be reverted and the thrift upgrade will work just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,23 +43,23 @@ Parquet-Java requires Java 17 or higher and uses Maven to build. It also depends
 To build and install the thrift compiler, run:
 
 ```
-wget -nv https://archive.apache.org/dist/thrift/0.23.0/thrift-0.23.0.tar.gz
-tar xzf thrift-0.23.0.tar.gz
-cd thrift-0.23.0
+wget -nv https://archive.apache.org/dist/thrift/0.24.0/thrift-0.24.0.tar.gz
+tar xzf thrift-0.24.0.tar.gz
+cd thrift-0.24.0
 chmod +x ./configure
 ./configure --disable-libs
 sudo make install -j
 ```
 
 Note: if you wish to verify the signature and checksum of a release:
-1. The GPG and sha checksums can be found under https://archive.apache.org/dist/thrift/0.23.0/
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -43,23 +43,23 @@ Parquet-Java requires Java 17 or higher and uses Maven to build. It also depends
...
+1. The GPG and sha checksums can be found under https://archive.apache.org/dist/thrift/0.24.0/
 2. Validate the signature of the artifact against the [Thrift committer KEYS](https://downloads.apache.org/thrift/KEYS).
 
-If you're on OSX and use homebrew, you can instead install Thrift 0.23.0 with `brew` and ensure that it comes first in your `PATH`.
+If you're on OSX and use homebrew, you can instead install Thrift 0.24.0 with `brew` and ensure that it comes first in your `PATH`.
 
 ```
 brew install thrift
-export PATH="/usr/local/opt/thrift@0.23.0/bin:$PATH"
+export PATH="$(brew --prefix thrift)/bin:$PATH"
```
Fokko, line 62: Unfortunally, this is an improvement since brew does not keep the old Thrift versions

Fokko (approved):
LGTM, We might need something similar in [`parquet-format`](https://github.com/apache/parquet-format/pull/597) :)

explore-with-yasir (approved):
+1 to this upgrade
one additional data point in its favor beyond the brew situation: Thrift 0.24.0 also resolves three CVEs filed against earlier Thrift lines (CVE-2026-43871, CVE-2026-45112, CVE-2026-48586). Because parquet-format-structures shades the Thrift classes into its published jar, SCA scanners attribute those advisories to every application that ships the jar, so this bump quietly clears findings for downstream users that they cannot fix through their own dependency management.

Fokko:
@dossett Thanks for raising this PR, and thanks @explore-with-yasir for the review 👍

Issue #3735, opened 2026-08-22
Variant field names not being ordered correctly via UTF-8 byte order


rayokota:
### Describe the bug, including details regarding any error messages, version, and platform.

The Variant spec requires the field ids in an object's header to be sorted by the                                                                                                 
  **UTF-8 byte order** of the field names, so a reader can binary-search them.                                                                                                      
  `VariantBuilder` sorted the fields — and `Variant.getFieldByKey` binary-searched them —                                                                                           
  using `String.compareTo`, which orders by **UTF-16 code units**, not UTF-8 bytes.                                                                                                 
                                                                                                                                                                                    
  The two orderings are identical for all field names in the Basic Multilingual Plane, but                                                                                          
  they diverge for names containing supplementary-plane characters (U+10000 and above):                                                                                             
  `String.compareTo` orders a leading high surrogate (0xD800–0xDBFF) before code points in
  U+E000..U+FFFF, whereas UTF-8 byte order (and the spec) orders them after. Consequences:                                                                                          
                                                                                                                                                                                    
  - An object parquet-java builds with such keys has field ids sorted in an order that                                                                                              
    violates the spec.                                                                                                                                                              
  - A spec-compliant reader (e.g. the Apache Arrow C++/Rust/Go Variant readers) binary-searching                                                                                    
    that object can fail to find fields.                                                                                                                                            
  - Conversely, parquet-java's own binary search fails to find a supplementary-plane key in an                                                                                      
    object produced by a spec-compliant writer.                                                                                                                                     
                                                                                                                                                                                    
  The bug only surfaces when an object both contains a supplementary-plane key and is large                                                                                         
  enough to take the binary-search path, so it has gone unnoticed.  


### Component(s)
Core

PR #3736, opened 2026-08-22
GH-3735 Fix Variant field name comparisons to use UTF-8 byte order

rayokota:
<!--
Thanks for opening a pull request!

If you're new to Parquet-Java, information on how to contribute can be found here: https://parquet.apache.org/docs/contribution-guidelines/contributing

Please open a GitHub issue for this pull request: https://github.com/apache/parquet-java/issues/new/choose
and format pull request title as below:

    GH-${GITHUB_ISSUE_ID}: ${SUMMARY}

or simply use the title below if it is a minor issue:

    MINOR: ${SUMMARY}

-->

### Rationale for this change

The Variant spec requires the field ids in an object's header to be sorted by the                                                                                                 
  **UTF-8 byte order** of the field names, so a reader can binary-search them.                                                                                                      
  `VariantBuilder` sorted the fields — and `Variant.getFieldByKey` binary-searched them —                                                                                           
  using `String.compareTo`, which orders by **UTF-16 code units**, not UTF-8 bytes.                                                                                                 
                                                                                                                                                                                    
  The two orderings are identical for all field names in the Basic Multilingual Plane, but                                                                                          
  they diverge for names containing supplementary-plane characters (U+10000 and above):                                                                                             
  `String.compareTo` orders a leading high surrogate (0xD800–0xDBFF) before code points in                                                                                          
  U+E000..U+FFFF, whereas UTF-8 byte order (and the spec) orders them after. Consequences:                                                                                          
                                                                                                                                                                                    
  - An object parquet-java builds with such keys has field ids sorted in an order that                                                                                              
    violates the spec.                                                                                                                                                              
  - A spec-compliant reader (e.g. the Apache Arrow C++/Rust/Go Variant readers) binary-searching                                                                                    
    that object can fail to find fields.                                                                                                                                            
  - Conversely, parquet-java's own binary search fails to find a supplementary-plane key in an                                                                                      
    object produced by a spec-compliant writer.                                                                                                                                     
                                                                                                                                                                                    
  The bug only surfaces when an object both contains a supplementary-plane key and is large                                                                                         
  enough to take the binary-search path, so it has gone unnoticed.      


- New `VariantUtil.encodeKey(String)` and `VariantUtil.compareKeys(byte[], byte[])`, which order
  field names by unsigned lexicographic UTF-8 byte order
- `VariantBuilder.FieldEntry.compareTo` and `Variant.getFieldByKey` now use that comparison
- Two new tests: `testObjectKeysSortedByUtf8ByteOrder` and
  `testLargeObjectBinarySearchWithSupplementaryKey`


### Are these changes tested?

  Two new tests in `TestVariantObjectBuilder`:                                                                                                                                      
  - `testObjectKeysSortedByUtf8ByteOrder` — builds an object with keys U+FFFF (`EF BF BF`) and                                                                                      
    U+10000 (`F0 90 80 80`) appended in reverse and asserts the encoded field order is U+FFFF                                                                                       
    then U+10000 (UTF-8 order), which the previous `compareTo` reversed.                                                                                                            
  - `testLargeObjectBinarySearchWithSupplementaryKey` — a 42-field object (above                                                                                                    
    `BINARY_SEARCH_THRESHOLD`) mixing ASCII keys with U+FFFF and U+10000, asserting                                                                                                 
    `getFieldByKey` resolves both through the binary-search path. 

<!-- Please uncomment the line below and replace ${GITHUB_ISSUE_ID} with the actual Github issue id. -->
Closes #GH-3735

### DIFF
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
@@ -282,7 +282,7 @@ public final class Variant {
         int mid = (low + high) >>> 1;
         int midId = VariantUtil.readUnsignedLittleEndian(value, idStart + info.idSize * mid, info.idSize);
         String midKey = getMetadataKeyCached(midId);
-        int cmp = midKey.compareTo(key);
+        int cmp = VariantUtil.compareKeys(midKey, key);
         if (cmp < 0) {
           low = mid + 1;
         } else if (cmp > 0) {
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
@@ -273,6 +273,8 @@ public final class Variant {
         }
       }
     } else {
+      // Encode the lookup key once, outside the loop, rather than on every comparison.
+      byte[] keyBytes = VariantUtil.encodeKey(key);
       int low = 0;
       int high = info.numElements - 1;
       while (low <= high) {
@@ -282,7 +284,7 @@ public final class Variant {
         int mid = (low + high) >>> 1;
         int midId = VariantUtil.readUnsignedLittleEndian(value, idStart + info.idSize * mid, info.idSize);
         String midKey = getMetadataKeyCached(midId);
-        int cmp = VariantUtil.compareKeys(midKey, key);
+        int cmp = VariantUtil.compareKeys(VariantUtil.encodeKey(midKey), keyBytes);
         if (cmp < 0) {
           low = mid + 1;
         } else if (cmp > 0) {
--- a/parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
+++ b/parquet-variant/src/main/java/org/apache/parquet/variant/VariantBuilder.java
@@ -679,6 +679,12 @@ public class VariantBuilder {
     final int offset;
     int valueSize = 0;
 
+    /**
+     * Lazy cache of the UTF-8 encoding of `key`, which sorting an object compares O(log n) times
+     * per entry. Encoded on demand so single-field objects, which are never compared, skip it.
+     */
+    private byte[] keyBytes;
+
     FieldEntry(String key, int id, int offset) {
       this.key = key;
       this.id = id;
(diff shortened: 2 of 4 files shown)


parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -273,6 +273,8 @@ public Variant getFieldByKey(String key) {
         }
       }
```
peterxcli, line 274: I'm thinking if we could just do binary search anyway. as arrow-rs does it.
https://github.com/apache/arrow-rs/blob/7ec3f5ab0c41f10ece65d3745b989ddde510539b/parquet-variant/src/variant/metadata.rs#L378-L397

parquet-variant/src/main/java/org/apache/parquet/variant/Variant.java
```diff
@@ -282,7 +284,7 @@ public Variant getFieldByKey(String key) {
         int mid = (low + high) >>> 1;
         int midId = VariantUtil.readUnsignedLittleEndian(value, idStart + info.idSize * mid, info.idSize);
         String midKey = getMetadataKeyCached(midId);
-        int cmp = midKey.compareTo(key);
+        int cmp = VariantUtil.compareKeys(VariantUtil.encodeKey(midKey), keyBytes);
```
peterxcli, line 287: Should we preserve read compatibility with Variant values written by earlier parquet-java versions? Its writers sorted object entries using Java `String.compareTo`.

peterxcli (commented):
Thanks for this fix, just discovered the same problem few minutes ago 😆. left two comment, one is considering removing the threshold for binary search, the other is for backward compatability.

Parquet spec has regulated the logical field list order:
> https://parquet.apache.org/docs/file-format/types/variantencoding/#:~:text=For%20objects%2C%20field,constructing%20Variant%20values.
> For objects, field IDs and offsets must be listed in the order of the corresponding field names, sorted lexicographically (using unsigned byte ordering for UTF-8). Note that the field values themselves are not required to follow this order. As a result, offsets will not necessarily be listed in ascending order. The field values are not required to be in the same order as the field IDs, to enable flexibility when constructing Variant values.

btw, seems like spark also does the same incorrect behaviour.
- `VariantBuilder.FieldEntry.compareTo` determines the encoded object-field order:
https://github.com/apache/spark/blob/9da9f8d673914d1648514f59d85e3adafb300d1a/common/variant/src/main/java/org/apache/spark/types/variant/VariantBuilder.java#L1009-L1010
- `Variant.getFieldByKey` uses `String.compareTo` while binary-searching objects containing at least 32 fields:
https://github.com/apache/spark/blob/9da9f8d673914d1648514f59d85e3adafb300d1a/common/variant/src/main/java/org/apache/spark/types/variant/Variant.java#L145-L167

rayokota:
@peterxcli , I see you made the equivalent fixes for Spark.  Feel free to adapt this PR to match or to close this PR and open a new one.  Thanks!

rayokota:
Closed in favor of https://github.com/apache/parquet-java/pull/3746

PR #3745, opened 2026-08-24
Add parameter to enable reading KMS URL from key material

ggershinsky:


parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/FileKeyUnwrapper.java
```diff
@@ -157,10 +157,13 @@ KeyToolkit.KmsClientAndDetails getKmsClientFromConfigOrKeyMaterial(KeyMaterial k
...
-      kmsInstanceURL = keyMaterial.getKmsInstanceURL();
+      if (hadoopConfiguration.getBoolean(
+          KeyToolkit.KMS_ENABLE_URL_READ_PROPERTY_NAME, KeyToolkit.KMS_ENABLE_URL_READ_DEFAULT)) {
+        kmsInstanceURL = keyMaterial.getKmsInstanceURL();
+      }
+
       if (null == kmsInstanceURL) {
-        throw new ParquetCryptoRuntimeException(
-            "KMS instance URL is missing both in properties and file key material");
+        kmsInstanceURL = KmsClient.KMS_INSTANCE_URL_DEFAULT;
```
wgtmac, line 166: Why do we need to change this line?

ggershinsky: This will prevent throwing an exception for most of the current users. Today, this exception is never thrown in practice, because the writer always fills the url value (set or default) in the key material, so the reader always gets something, either from properties or from the key material. Now that we skip reading the url from key material by default, the existing readers that don't use the property value, will get an exception. So to prevent it and to preserve the current interface, we drop the exception, and return the KMS_INSTANCE_URL_DEFAULT value (which is given to the readers today, from the key material).

Fokko (approved):
LGTM, although we could throw in a test to avoid chaing the behavior in the future 👍

wgtmac (approved):

ggershinsky:
Thanks. It's somewhat tricky, but I'll check if a reasonable unitest can be built for this.

ggershinsky:
Added a unitest. @Fokko @wgtmac can you have a look at it? I'll merge the patch then.

wgtmac (approved):

Fokko:
@ggershinsky Thanks, LGTM 👍

PR #3748, opened 2026-08-25
GH-3747: Preserve the final selected row when skipping pages

### LINKED ISSUE #3747: Column-index filtering can return a value from an earlier row
## Describe the bug

Column-index filtering can make the record-based Parquet reader return a value from the wrong row. The problem occurs when the reader takes the final selected row position from its iterator while it is still processing an earlier data page. It treats the now-empty iterator as completion, even though that last row has not been reached.

Here is a concrete example using an ordinary equality predicate. A row group has two required INT32 columns:

| Physical row | `predicate` | `payload` |
| --- | ---: | ---: |
| 0 | 0 | 0 |
| 1 | 0 | 10 |
| 2 | 100 | 20 |
| 3 | 100 | 30 |
| 4 | 0 | 40 |

The columns have different, valid page boundaries: `predicate` has pages containing rows **[0,1]**, **[2,3]**, and **[4]**, while `payload` has pages containing rows **[0,1,2]** and **[3,4]**.

Read this file with `ParquetReader<Group>`, `FilterCompat.get(eq(intColumn("predicate"), 0))`, and `useColumnIndexFilter(true)`. The column index excludes predicate page [2,3], leaving candidate row positions **0, 1, and 4**. The expected payloads are **[0, 10, 40]**, but the reader returns **[0, 10, 20]**: the last record combines the predicate value from row 4 with the payload from row 2. Reading the same file with column-index filtering disabled returns the correct result.

## Cause

`SynchronizingColumnReader.isFullyConsumed()` currently checks only `!rowIndexes.hasNext()`. Consuming a position from this iterator means that the row has become the next target, not that its data has been read. In the example, the payload reader selects target row 4 while still on its first page. At the page transition, it exits instead of opening the second page and advancing to row 4.

The completion check needs to account for the pending target as well as iterator exhaustion, while retaining the existing no-more-targets sentinel. This is a reader-state bug; it does not require a file-format change.

## Version and reproduction

Reproduced against Apache Parquet `master` at `60175684378abff1ea001b6541ec38da42a2eff1` (`1.19.0-SNAPSHOT`), using OpenJDK 17.0.19 on macOS arm64. A standalone check writes real Parquet files with the page boundaries above, then reads them through the normal predicate-filtering API. It uses no externally supplied row positions.

The wrong result occurs with both V1 and V2 data pages and required INT32 columns. Changing only the reader's completion check returns the correct result in both cases. Controls with aligned page boundaries, or with two consecutive rows in the final candidate range, already return the correct results on the old reader; not every filtered read triggers the problem.

Focused column-reader regression tests also reproduce the failure with optional and repeated columns in both page versions: all four fail before the fix and pass afterward.

## Component(s)

Core

sunchao:
## Why are the changes needed?

A Parquet file stores each column in separate data pages, and the page boundaries do not have to line up across columns. When column-index filtering rules out pages for a predicate, the record reader must advance every projected column to the same retained row positions. Otherwise it can assemble a record using values from different rows.

The current reader can lose that synchronization at the final selected row. For example, consider a row group with these two required INT32 columns:

| Physical row | `predicate` | `payload` |
| --- | ---: | ---: |
| 0 | 0 | 0 |
| 1 | 0 | 10 |
| 2 | 100 | 20 |
| 3 | 100 | 30 |
| 4 | 0 | 40 |

Suppose `predicate` has pages covering rows **[0,1]**, **[2,3]**, and **[4]**, while `payload` has pages covering **[0,1,2]** and **[3,4]**. Filtering for `predicate = 0` eliminates predicate page [2,3], so the reader needs positions **0, 1, and 4**. It should return payloads **[0, 10, 40]**. With column-index filtering enabled, it instead returns **[0, 10, 20]**: the last payload comes from row 2, not row 4.

The payload reader has already taken row 4 from the selected-position iterator, but has not yet reached that row. When it needs to move to the next page, `SynchronizingColumnReader` mistakes the exhausted iterator for the end of the read and stops on the earlier page. Disabling column-index filtering gives the correct result for the same file.

## What changes were proposed in this PR?

Keep the synchronizing reader active while its final selected row is still pending. Iterator exhaustion is only sufficient to finish once the current target has been reached, or the reader has explicitly recorded that no target remains. This lets the existing page-advance and value-skipping logic reach the last selected row without changing how other rows are read.

The change is confined to that completion check, with regression coverage in the existing column-reader test class. It does not change public APIs, the Parquet format, or how pages are selected for filtering.

## How was this PR tested?

Validated locally against Apache `master` at `60175684378abff1ea001b6541ec38da42a2eff1`, with Java 17 and Thrift 0.23.0. All four added regression tests fail before the fix with an earlier row's value and pass afterward. They cover optional and repeated columns, V1 and V2 pages, null or empty rows, and selections ending at different positions within a page. The three pre-existing tests in the class remain passing.

The full `parquet-column` suite passes **675 tests**, and the existing `TestColumnIndexFiltering` file-reader suite passes **24 tests**. The column reactor's dependency tests and the `spotless:check` formatting check also pass.

An additional standalone check writes real files with the page layouts above and reads them through an ordinary `ParquetReader` predicate, without supplying row positions. It reproduces the wrong payload with the old reader and returns the correct values with this fix, for both V1 and V2 pages. Aligned-page and final-contiguous-range controls remain correct. This check is separate from the committed unit tests.

Closes #3747.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/impl/SynchronizingColumnReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/impl/SynchronizingColumnReader.java
@@ -85,7 +85,9 @@ class SynchronizingColumnReader extends ColumnReaderBase {
 
   @Override
   boolean isFullyConsumed() {
-    return !rowIndexes.hasNext();
+    // The final target may have been taken from the iterator without being reached yet.
+    // Long.MAX_VALUE marks that there are no more target rows.
+    return !rowIndexes.hasNext() && (currentRow >= targetRow || targetRow == Long.MAX_VALUE);
   }
 
   @Override
--- a/parquet-column/src/test/java/org/apache/parquet/column/impl/TestColumnReaderImpl.java
+++ b/parquet-column/src/test/java/org/apache/parquet/column/impl/TestColumnReaderImpl.java
@@ -18,9 +18,12 @@
  */
 package org.apache.parquet.column.impl;
 
+import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_1_0;
 import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_2_0;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.apache.parquet.Version;
 import org.apache.parquet.VersionParser;
@@ -29,7 +32,9 @@ import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.column.ColumnReader;
 import org.apache.parquet.column.Dictionary;
 import org.apache.parquet.column.ParquetProperties;
+import org.apache.parquet.column.ParquetProperties.WriterVersion;
 import org.apache.parquet.column.page.DataPage;
+import org.apache.parquet.column.page.DataPageV1;
 import org.apache.parquet.column.page.DataPageV2;
 import org.apache.parquet.column.page.DictionaryPage;
 import org.apache.parquet.column.page.mem.MemPageReader;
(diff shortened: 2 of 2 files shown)


wgtmac (approved):
LGTM, thanks @sunchao!

erikwright:
@sunchao  @wgtmac  — would it make sense to backport this to parquet-1.19.x, ideally for the pending 1.19.0 release, or a subsequent patch release if it’s too late? Alternatively, would targeting the next 1.18.x patch release be a better path?

I reproduced this failure while exploring page pruning in Iceberg (see related [Iceberg #17596](https://github.com/apache/iceberg/issues/17596)), and verified that this fix resolves it. Happy to validate a backport against that reproducer.

sunchao:
@erikwright yes I think it's a good idea. Let me open another PR for this targeting `parquet-1.19.x` branch.

sunchao:
PR: https://github.com/apache/parquet-java/pull/3797

PR #3749, opened 2026-08-26
MINOR: Reuse RLE decoder buffers for packed runs

divjotarora:
### Rationale for this change

`RunLengthBitPackingHybridDecoder` allocated integer and byte buffers for every packed run. This
decoder is used for repetition and definition levels and dictionary IDs, so repeated packed runs
created avoidable allocation and GC pressure.

The decoder now grows and reuses its packed-run buffers, tracks the active value position
independently of buffer capacity, and reads directly from the input stream. It preserves existing
short-input zero padding and EOF behavior. A focused JMH benchmark and regression tests cover
smaller runs following larger runs and truncated final groups.

### Are these changes tested?

Yes. The existing `parquet-column` tests pass. The new benchmark shows 27-35% speedup and reduced allocations:
```
> ./parquet-benchmarks/run.sh all \
    org.apache.parquet.benchmarks.RunLengthBitPackingHybridDecodingBenchmark \
    -wi 3 -i 5 -f 2 -prof gc

   Bit width                       Before ops/s                        After ops/s    Change    Before B/value    After B/value
  ━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━
           1    125,889,985.415 ± 1,443,739.608      173,074,464.937 ± 470,051.407    +37.5%    13.501 ± 0.001    0.002 ± 0.001
  ───────────  ─────────────────────────────────  ─────────────────────────────────  ────────  ────────────────  ───────────────
           8      130,416,053.717 ± 754,362.833      175,155,689.891 ± 499,245.748    +34.3%    13.751 ± 0.001    0.002 ± 0.001
  ───────────  ─────────────────────────────────  ─────────────────────────────────  ────────  ────────────────  ───────────────
          16    132,141,608.282 ± 5,365,801.318    167,817,248.033 ± 1,318,258.836    +27.0%    13.751 ± 0.797    0.002 ± 0.001

```

No. This is an internal performance improvement with no API or format changes.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
@@ -18,9 +18,10 @@
  */
 package org.apache.parquet.column.values.rle;
 
-import java.io.DataInputStream;
+import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import org.apache.parquet.Preconditions;
 import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.column.values.bitpacking.BytePacker;
@@ -47,7 +48,9 @@ public class RunLengthBitPackingHybridDecoder {
   private MODE mode;
   private int currentCount;
   private int currentValue;
-  private int[] currentBuffer;
+  private int currentBufferPosition;
+  private int[] currentBuffer = new int[0];
+  private byte[] packedBytes = new byte[0];
 
   public RunLengthBitPackingHybridDecoder(int bitWidth, InputStream in) {
     LOG.debug("decoding bitWidth {}", bitWidth);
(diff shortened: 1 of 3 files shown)


parquet-column/src/main/java/org/apache/parquet/column/values/rle/RunLengthBitPackingHybridDecoder.java
```diff
@@ -91,16 +94,25 @@ private void readNext() throws IOException {
...
+        }
+        currentBufferPosition = 0;
+        int bytesRequired = numGroups * bitWidth;
+        if (packedBytes.length < bytesRequired) {
+          packedBytes = new byte[bytesRequired];
+        }
         // At the end of the file RLE data though, there might not be that many bytes left.
         int bytesToRead = (int) Math.ceil(currentCount * bitWidth / 8.0);
         bytesToRead = Math.min(bytesToRead, in.available());
-        new DataInputStream(in).readFully(bytes, 0, bytesToRead);
```
Fokko, line 99: Nice, I always found it weird to wrap the stream with a new stream to read the bytes 👍

Fokko (approved):
Neat improvement @divjotarora Thanks!

PR #3750, opened 2026-08-26
GH-2142: Write fields of empty message types as proto bytes

### LINKED ISSUE #2142: Proto-Parquet cannot write a schema with an empty group
This is valid protobuf:

 
```java

message Something {
    EmptyMessage outerEmptyMessage = 1;
}

message EmptyMessage {
}
```
However when we try to convert this to Parquet we get:
```java

org.apache.parquet.schema.InvalidSchemaException: Cannot write a schema with an empty group: optional group outerEmptyMessage = 1 {
}
```
Is this a limitation of the Parquet format, or just a bug?

If the former, can we break earlier when generating the schema out of Protobuf, and not when trying to write the content to the already generated schema?

 

**Reporter**: [André Pinto](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=andresp)

<sub>**Note**: *This issue was originally created as [PARQUET-1240](https://issues.apache.org/jira/browse/PARQUET-1240). Please see the [migration documentation](https://issues.apache.org/jira/browse/PARQUET-2502) for further details.*</sub>

puskarpeter:
### Rationale for this change

Protobuf allows empty message definitions, but Parquet forbids empty groups. Converting a message
that merely *contains* a field of an empty message type produces a schema with an empty group,
which writer construction rejects with `InvalidSchemaException: Cannot write a schema with an
empty group`. Such fields appear in real-world schemas (deprecated stubs, marker/placeholder
messages), and a single one makes the whole message type unwritable.

`ProtoSchemaConverter.addMessageField` terminates a field whose message type has no fields as a
`BINARY` column holding the serialized message — the same mechanism PARQUET-1711 uses for
recursion beyond `maxRecursion`. Since an empty message serializes to zero bytes, the column is
cheap, and field **presence** still round-trips (null = unset vs empty bytes = set):

- singular field → `optional binary stub` (or `required`, per the field);
- repeated field, parquet-specs mode → LIST-wrapped binary via the existing
  `addRepeatedPrimitive`, so element cardinality survives;
- repeated field, old style → `repeated binary`;
- map value type → `optional binary value` inside the `key_value` group (keys stay typed).

`ProtoWriteSupport.createMessageWriter`'s existing truncated-field check (primitive BINARY where a
message field was declared → `BinaryWriter`) is generalized to look through the LIST/MAP wrapper
(`getGroupType` → `getContentType`), so the writer tree lines up with these schemas.

A message that is empty at the **root** is still rejected — there is no parent field to hold the
bytes, and a Parquet file with zero columns is not representable.

**Read side** (second commit, after review): `ProtoMessageConverter` used to cast every message
field's Parquet type to a group, so a message field stored as `BINARY` failed with a
`ClassCastException` while the converter tree was built — which also means files with
PARQUET-1711 recursion truncation have been unreadable by `ProtoParquetReader` since 1.13.0 (#995
shipped with an explicit "TODO: ReadSupport"). A new `ProtoBinaryMessageConverter` parses the
bytes back through `parentBuilder.newBuilderForField(field)` and hands the message to the existing
parent container, so singular fields, LIST elements, old-style repeated fields and map values all
round-trip without touching `ListConverter`/`MapConverter`. Parse failures surface as
`ParquetDecodingException`.

### Are these changes tested?

Yes. New `ProtoEmptyMessageTest` (new test messages `Stub`/`StubBox` in `Trees.proto`) writes
through the real write path (`ProtoParquetWriter` → `MessageColumnIO`, both specs-compliant and
old style) and reads back with `GroupReadSupport` and with `ProtoParquetReader`:

- singular / repeated / map-value empty-message fields round-trip with correct cardinality and
  zero-byte values;
- presence round-trips (set empty message vs unset field);
- `ProtoParquetReader` reads the written messages back equal to the originals (specs-compliant and
  old style), and the same read path round-trips a `BinaryTree` truncated at `maxRecursion`;
- an empty root message still fails with `InvalidSchemaException` ("Cannot write a schema with an
  empty group").

`ProtoSchemaConverterTest.testEmptyMessageFields` pins the converted schema. The full
parquet-protobuf suite passes (117 tests).

Message types that previously could not be written to Parquet at all now can; fields of empty
message types appear as (possibly LIST/MAP-wrapped) `binary` columns and read back into the
original messages with `ProtoParquetReader`. Files with `maxRecursion` truncation, previously
unreadable by `ProtoParquetReader`, now read back as well. No change for schemas that were
previously writable. Error behavior for an empty root message is unchanged.

Closes #2142

### DIFF
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoMessageConverter.java
@@ -36,6 +36,7 @@ import com.google.protobuf.DoubleValue;
 import com.google.protobuf.FloatValue;
 import com.google.protobuf.Int32Value;
 import com.google.protobuf.Int64Value;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import com.google.protobuf.StringValue;
 import com.google.protobuf.UInt32Value;
@@ -347,6 +348,9 @@ class ProtoMessageConverter extends GroupConverter {
           if (messageType.equals(BytesValue.getDescriptor())) {
             return new ProtoBytesValueConverter(pvc);
           }
+          // Otherwise the column holds the serialized message itself: ProtoSchemaConverter stores
+          // fields of empty message types and recursion beyond maxRecursion as proto bytes.
+          return new ProtoBinaryMessageConverter(pvc, parentBuilder, fieldDescriptor);
         }
         Message.Builder subBuilder = parentBuilder.newBuilderForField(fieldDescriptor);
         return new ProtoMessageConverter(conf, pvc, subBuilder, parquetType.asGroupType(), extraMetadata);
@@ -492,6 +496,39 @@ class ProtoMessageConverter extends GroupConverter {
         Binary binaryValue = dictionary.decodeToBinary(i);
         dict[i] = translateEnumValue(binaryValue);
       }
+    }
+  }
+
+  /**
+   * Reads a message field that {@link ProtoSchemaConverter} stored as the serialized proto bytes
+   * (a field of an empty message type, or recursion truncated at maxRecursion) back into the
+   * message.
+   */
+  static final class ProtoBinaryMessageConverter extends PrimitiveConverter {
+
+    private final ParentValueContainer parent;
+    private final Message.Builder parentBuilder;
+    private final Descriptors.FieldDescriptor fieldDescriptor;
+
+    ProtoBinaryMessageConverter(
+        ParentValueContainer parent,
+        Message.Builder parentBuilder,
+        Descriptors.FieldDescriptor fieldDescriptor) {
+      this.parent = parent;
+      this.parentBuilder = parentBuilder;
+      this.fieldDescriptor = fieldDescriptor;
+    }
+
+    @Override
+    public void addBinary(Binary binary) {
+      Message.Builder builder = parentBuilder.newBuilderForField(fieldDescriptor);
+      try {
+        builder.mergeFrom(ByteString.copyFrom(binary.toByteBuffer()));
+      } catch (InvalidProtocolBufferException e) {
+        throw new ParquetDecodingException(
+            "Cannot parse field " + fieldDescriptor.getFullName() + " from its serialized proto bytes", e);
+      }
+      parent.add(builder.build());
     }
   }
 
--- a/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
+++ b/parquet-protobuf/src/main/java/org/apache/parquet/proto/ProtoSchemaConverter.java
@@ -327,6 +327,18 @@ public class ProtoSchemaConverter {
       return addMapField(descriptor, builder, seen, depth);
     }
 
+    // Parquet forbids empty groups, so a field of an empty message type is terminated as proto
+    // bytes (zero bytes when the message is set - presence still round-trips), preserving the
+    // field's repetition so the write path (Array/Repeated/MapWriter) still matches the schema.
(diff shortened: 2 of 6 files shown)


dossett:
I'm always interested in parquet-protobuf work.  I asked Codex to scrutinize this PR and according to that analysis it's possible to write records that can't be read back yet.

```java
@Test
  public void emptyMessageRoundTripsThroughProtoReader() throws Exception {
    Trees.StubBox expected = Trees.StubBox.newBuilder()
        .setStub(Trees.Stub.getDefaultInstance())
        .build();

    Path file = write(true, expected);

    assertThat(TestUtils.readMessages(file, Trees.StubBox.class))
        .containsExactly(expected);
  }
```

The read fails because when `ProtoMessagerConverter` is eventually called it still expects a Parquet group and cast case the binary column.  I confirmed this unit test fails against this PR branch.

The unit test fails on `master` at the write step, which this PR does address.

puskarpeter:
@dossett Thanks for the good catch :) I never came across this issue in my testing since we are always reading the written outputs through other reader implementations such as the arrow reader when I was reading back the results in Zeppelin in Spark and our use-cases only exercised the write-side.

I added more unit tests and this issue should be fixed now as well.

The description of the PR was also updated with this in mind.

wgtmac:
I’m not very familiar with this module, so please correct me if I’m missing any context. The fix appears to work for parquet-protobuf round-tripping. Could this lead to schema mismatches or otherwise surprising behavior for non-protobuf Parquet readers? Is this representation already considered an established compatibility contract because recursion truncation uses the same encoding? A short documentation note or cross-reader test would help clarify the expected behavior.

wgtmac:
Gentle ping @puskarpeter

puskarpeter:
@wgtmac Thanks for looking into it :) Both are fair points.

  Non-protobuf readers just see an unannotated `BINARY` column with the field's own repetition (or inside the usual `LIST`/`MAP` wrappers), so there is no logical type to misinterpret. For an empty message the value is either zero bytes (set) or null (unset). The tests were already reading the files back through
  `GroupReadSupport`, which knows nothing about protobuf, but that was implicit, so I added tests that assert the footer schema of the written file in both schema styles. Readers that want to interpret the bytes can use the descriptor the writer already stores in the footer, which is what `ProtoParquetReader` does.

  Regarding the contract, PARQUET-1711 has been writing truncated recursion this way since 1.13.0, so this PR reuses an encoding that already exists in files out there rather than adding a new one. You are right that it was only mentioned in the `maxRecursion` Javadoc though, so I added a section to the
  parquet-protobuf README and a note in the `ProtoSchemaConverter` Javadoc covering both cases.

  One caveat I put in the docs as well: if an empty message type later gains fields, new files will have a group where old files have `BINARY`, so schema merging tools will report a type conflict. That is the same thing that already happens today when `maxRecursion` changes.

parquet-protobuf/README.md
```diff
@@ -21,3 +21,45 @@ parquet-protobuf
...
+      optional binary value;
+    }
+  }
+}
+```
+
+Readers that do not know about protobuf simply see opaque bytes (all of them empty for an
+empty message type). Readers that do can parse the bytes with the message's descriptor, which
+the writer stores in the file footer under the `parquet.proto.descriptor` key;
+`ProtoParquetReader` does this automatically and materializes the original message.
```
wgtmac: nit: `ProtoParquetReader` seems to load the protobuf class from `parquet.proto.class` (or the configured class), rather than parse `parquet.proto.descriptor` from the footer. Could we clarify that here so it doesn’t imply the footer descriptor alone is sufficient?

wgtmac (approved):

parquet-protobuf/README.md
puskarpeter: Should be more clear now.

wgtmac:
Thanks for working on this, @puskarpeter!

PR #3752, opened 2026-08-26
GH-3751: Preserve field repetition when truncating recursive proto fields

### LINKED ISSUE #3751: parquet-protobuf: recursion truncation breaks repeated and map fields — ClassCastException in specs-compliant mode, file corruption in the old style
### Describe the bug

`ProtoSchemaConverter` (PARQUET-1711) terminates recursive proto message fields at
`parquet.proto.maxRecursion` depth by replacing them with the serialized proto bytes. The
truncation hardcodes the replacement column as `optional binary`, ignoring the original field's
repetition:

```java
// ProtoSchemaConverter.addMessageField
if (seen.get(typeName).size() > maxRecursion) {
  return builder.primitive(BINARY, Type.Repetition.OPTIONAL).as((LogicalTypeAnnotation) null);
}
```

`ProtoWriteSupport.MessageWriter` however still wraps every repeated field's writer in
`ArrayWriter` (specs-compliant) or `RepeatedWriter` (old style), and map fields in `MapWriter` —
all of which emit record structure the `optional binary` column cannot hold. The mock-based unit
tests (`ProtoWriteSupportTest.testRepeatedRecursion` / `testMapRecursion`) never validate against a
real `MessageColumnIO`, so the mismatch was never caught.

Consequences, each reproduced by an end-to-end write test included in the follow-up PR:

1. **Specs-compliant mode, repeated recursive field, data deeper than maxRecursion:** the write
   crashes with

   ```
   java.lang.ClassCastException: class org.apache.parquet.io.PrimitiveColumnIO cannot be cast to
   class org.apache.parquet.io.GroupColumnIO
   ```

   (`ArrayWriter` calls `startGroup()`/`startField("list", 0)` against a primitive column). This is
   a data-dependent landmine: schema creation and shallow rows succeed; the job dies only when a
   row's data actually nests past the limit.

2. **Old style (writeSpecsCompliant=false), repeated recursive field with more than one element at
   the truncation depth:** no exception — the write emits inconsistent repetition levels for the
   second and following elements, **corrupting the file**. Depending on the data, reading it back
   either fails with `ParquetDecodingException` (`Can not read value at ... in block`,
   EOF/BufferUnderflow underneath) or silently returns a wrong tree (elements lost or attached to
   phantom duplicate parent nodes).

3. **Specs-compliant map field at which the recursion budget runs out** (e.g.
   `google.protobuf.Struct` maps reached through `list_value` branches): the whole MAP — including
   its keys — collapses into a single unreadable binary in the schema, and writing data through
   that branch crashes with the same `ClassCastException` (`MapWriter` navigating `key_value`
   groups over a primitive column). On map paths where the budget happens to trip at a singular
   field first (like `Struct`'s main `fields → struct_value` chain), the schema was already fine —
   the collapse is branch-dependent.

### Reproducer

Any repeated self-recursive message nested deeper than maxRecursion, e.g. the existing test proto
`Trees.WideTree`:

```java
Trees.WideTree deep = ...; // chain of children 5 levels deep, 2 children per node
ProtoWriteSupport.setWriteSpecsCompliant(conf, true);
ProtoSchemaConverter.setMaxRecursion(conf, 2);
try (ParquetWriter<Message> w = ProtoParquetWriter.<Message>builder(path)
    .withMessage(Trees.WideTree.class).withConf(conf).build()) {
  w.write(deep); // ClassCastException
}
```

Affects all released versions since 1.13.0 (PARQUET-1711) through current master (verified on
1.17.1 and master @ e02f65e2).

### Proposed fix (PR follows)

Preserve the field's shape when truncating, mirroring how ordinary repeated primitives are handled:

- `ProtoSchemaConverter.addMessageField`:
  - repeated + specs-compliant → LIST-wrapped binary via the existing `addRepeatedPrimitive`
    (`optional group x (LIST) { repeated group list { required binary element } }`);
  - otherwise `builder.primitive(BINARY, getRepetition(descriptor))` (`repeated binary` in the old
    style; truncated optional fields unchanged, proto2 required fields now keep `required`);
  - specs-compliant map fields keep their MAP structure unconditionally; a recursive **value** type
    is truncated to `optional binary` inside `key_value` when `addMapField` recurses into the value
    field (same recursion budget, applied one level deeper where it belongs).
- `ProtoWriteSupport.createMessageWriter`: look through the LIST/MAP wrapper when detecting a
  truncated-to-binary message field (`getContentType`, introduced by the fix for #2142, which
  terminates empty message types through the same mechanism) so `BinaryWriter` is selected for
  truncated elements/values; the existing `ArrayWriter`/`RepeatedWriter`/`MapWriter` wrapping then
  lines up with the schema.

With the fix, each repeated element / map value at the truncation depth round-trips as one binary
containing the serialized subtree (`X.parseFrom(bytes)` reconstructs it), keys of truncated-value
maps stay queryable, and truncated optional fields are byte-for-byte unchanged.

Existing expected-schema tests were regenerated (`WideTree.par`, `Value.par`, `Struct.par`,
inline schemas, and the `testDeepRecursion` Struct fan-out series changes from `2n+4` to `2n+5`
because a truncated map now retains its key column). New `ProtoRecursionTruncationTest` (5 tests)
writes through a real `MessageColumnIO`: repeated recursion in both modes, a map field exhausting
the recursion budget (fails with the ClassCastException before the fix), and the already-working
map-main-path and optional cases as regression guards.

Note on schema compatibility: files previously written with a truncated *optional* field are
unchanged. A repeated/map truncated field changes its schema shape — but writing more than one
element at the truncation depth crashed (specs) or corrupted the file (old style) before, so no
valid existing files carry the old shape with meaningful multi-element data.

### Related

- PARQUET-1711 / #995 — introduced maxRecursion truncation (optional-field case only).
- #2708 / PARQUET-2181 — read-side ClassCastException in parquet-cli on proto files; note that
  `ProtoParquetReader` itself also cannot read back **any** truncated field (including the optional
  case that writes fine): `ProtoMessageConverter.newScalarConverter` has no binary→message path and
  throws `ClassCastException` at converter-tree construction. That read-side gap is orthogonal to
  this write-side fix and probably deserves its own issue.
- #2142 — empty message types cannot be written at all; the companion PR fixes it with the same
  terminate-as-proto-bytes mechanism, and this fix builds on it.

puskarpeter:
> [!NOTE]
> **Draft — stacked on #3750.** This fix reuses the terminate-as-proto-bytes machinery introduced
> by the empty-message fix (GH-2142), so only the top commit belongs to this PR; the base 2 commits
> are #3750's. I will rebase onto master and mark this ready for review once #3750 merges.

### Rationale for this change

PARQUET-1711 truncates recursive proto fields at `parquet.proto.maxRecursion` depth by replacing
them with the serialized proto bytes, but hardcodes the replacement column as `optional binary`,
ignoring the field's actual repetition. `ProtoWriteSupport` still wraps repeated fields' writers in
`ArrayWriter`/`RepeatedWriter` and map fields in `MapWriter`, which emit record structure the
`optional binary` column cannot hold. As a result (see #3751):

- specs-compliant mode: writing data that actually nests past `maxRecursion` through a *repeated*
  recursive field crashes with `ClassCastException: PrimitiveColumnIO cannot be cast to
  GroupColumnIO` — a data-dependent failure that passes schema creation and shallow rows;
- old style: writing more than one repeated element at the truncation depth emits inconsistent
  repetition levels and **corrupts the file** — depending on the data, reading it back either
  fails with `ParquetDecodingException` or silently returns a wrong tree (elements lost or
  attached to phantom duplicate nodes);
- when the recursion budget runs out at a *map* field itself (e.g. `google.protobuf.Struct` maps
  reached through `list_value` branches), the whole MAP — including its keys — collapses into one
  binary in the schema, and writing data through it crashes with the same `ClassCastException`.

The existing mock-based tests (`ProtoWriteSupportTest.testRepeatedRecursion`/`testMapRecursion`)
never validate against a real `MessageColumnIO`, which is why the mismatch went unnoticed.

`ProtoSchemaConverter.addMessageField` keeps the field's shape when truncating, reusing the
terminate-as-proto-bytes path introduced for empty message types (#3750, which this builds on):

- repeated + specs-compliant → LIST-wrapped binary via the existing `addRepeatedPrimitive`;
- otherwise `builder.primitive(BINARY, getRepetition(descriptor))` — `repeated binary` in the old
  style; truncated optional fields stay `optional binary`, byte-for-byte identical to before
  (proto2 `required` fields in a recursion cycle now keep their `required` repetition instead of
  being forced `optional`);
- the specs-compliant map branch now runs *before* the recursion check, so the MAP structure
  (typed key) is always preserved and a recursive value type is truncated to `optional binary`
  inside `key_value` when `addMapField` recurses into the value field — same recursion budget,
  applied at the level where the recursion actually is.

The writer side needs no further changes: the `getContentType` check from #3750 already
selects `BinaryWriter` behind LIST/MAP wrappers, so the existing
`ArrayWriter`/`RepeatedWriter`/`MapWriter` wrapping then lines up with the schema.

### Are these changes tested?

Yes. New `ProtoRecursionTruncationTest` (6 tests) writes recursive data deeper than `maxRecursion`
through the real write path (`ProtoParquetWriter` → `MessageColumnIO`) and reads it back:

- repeated recursion (`Trees.WideTree`), specs-compliant **and** old style — previously the
  `ClassCastException` and the file-corrupting write, respectively; now every element at the
  truncation depth round-trips as the serialized subtree;
- a map field that exhausts the recursion budget (`google.protobuf.Struct` behind `list_value`) —
  previously the whole-MAP collapse plus the same `ClassCastException` when data reached it; now
  keys stay typed and queryable, values round-trip as serialized protos;
- map-value recursion on the main `Struct` path and optional recursion (`Trees.BinaryTree`) as
  regression guards for the shapes that already worked;
- `ProtoParquetReader` round trip: a `WideTree` deeper than `maxRecursion` (both modes) and a
  map-recursive `Struct` read back **equal to the original messages** — with the binary-to-message
  read path from #3750, truncation is now lossless end to end.

Expected-schema fixtures were regenerated for the new truncation shape: `WideTree.par`,
`Value.par`, `Struct.par`, the inline schemas in `ProtoSchemaConverterTest`, and the
`testDeepRecursion` Struct fan-out series (now `2n+5` — a truncated map keeps its key column). The
full parquet-protobuf suite passes.

Schemas containing *repeated* or *map* recursive fields change shape at the truncation depth
(LIST-of-binary / MAP-with-binary-value instead of a single `optional binary`) — but writing more
than one element at that depth previously crashed (specs mode) or corrupted the file (old style),
so no valid existing files carry meaningful multi-element data in the old shape. Truncated
optional fields are unchanged; proto2 `required` fields in a recursion cycle now map to `required
binary` (previously forced `optional`). Data past the truncation depth now round-trips losslessly:
each binary cell is the serialized subtree, recoverable with `X.parseFrom(bytes)`.

Closes #3751


puskarpeter:
@wgtmac Now that #3750 is merged I rebased this one on master and marked it ready. It is the follow-up mentioned in #3751: the same terminate-as-proto-bytes path, applied to recursion truncation so repeated and
  map fields keep their shape instead of collapsing into a single `optional binary`. Only `ProtoSchemaConverter` changes, the writer and reader side are already covered by #3750.

wgtmac (approved):

Issue #3760, opened 2026-09-02
Clean up hadoop README.md


ggershinsky:
1. Remove a duplicate
2. Fix format of a number of parameters (no line breaks, because spaces are missing)

zhangxinyao88:
Hi, I’d like to take this. I’ll clean up the duplicate content and formatting in the Hadoop README, then verify the rendered Markdown.

ggershinsky:
Sorry, I should have closed this as completed.

PR #3761, opened 2026-09-02
GH-3760: Clean up hadoop README.md

ggershinsky:

In the Hadoop README.md doc, fix format of a number of parameters (no line breaks, because spaces are missing)

### DIFF
--- a/parquet-hadoop/README.md
+++ b/parquet-hadoop/README.md
@@ -222,7 +222,7 @@ conf.set("parquet.bloom.filter.enabled#column.path", false);
 If it is true, the bloom filter will be generated with the optimal bit size 
 according to the number of real data distinct values. If it is false, it will not take effect.
 Note that the maximum bytes of the bloom filter will not exceed `parquet.bloom.filter.max.bytes` configuration (if it is 
-set too small, the generated bloom filter will not be efficient).
+set too small, the generated bloom filter will not be efficient).  
 **Default value:** `false`
 
 ---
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Thanks @ggershinsky for cleaning this up. This is the reason why I prefer to use soft-line wraps in the editor, instead of line-breaks :D

parquet-hadoop/README.md
```diff
@@ -288,12 +288,6 @@ true if the reader is using a `DirectByteBufferAllocator`
 **Description:** Whether to write out page level checksums.  
 **Default value:** `true`
 
----
-
-**Property:** `parquet.crypto.factory.class`  
-**Description:** Class implementing EncryptionPropertiesFactory.  
-**Default value:** None. If not set, the file won't be encrypted by a crypto factory.
```
divjotarora, line 295: Is this actually a duplicate? This parameter is under section `Class: ParquetOutputFormat` and references `EncryptionPropertiesFactory`. The duplicate below is under section `Class: HadoopReadOptions` and references `DecryptionPropertiesFactory`.  I'm fine removing it if there's precedent to not duplicate, but it does seem to be used for two different things.

ggershinsky: Thanks. Yep, it makes sense to list the same parameter twice, in both output (encrypt) and read (decrypt) sections. I'll update the patch.

divjotarora (approved):

PR #3762, opened 2026-09-03
Fix/prepare rc self ci check

RussellSpitzer:
The "Prepare Release" job now ignores github actions on a commit that are related to the release. Previously we would check that no actions were running (to make sure CI had finished) but by running the release job we started a new action on that commit. To fix this we just ignore all jobs that start with release-*.

### Are these changes tested?

New Bats tests checking this situation

No


Fokko (approved):
The LLM approves 👍 Thanks @RussellSpitzer

PR #3763, opened 2026-09-03
Prepare RC: write generated Maven settings outside the repo

RussellSpitzer:
### Rationale for this change

Before we were making a temporary .release-settings.xml in the working directory. This caused RAT to fail since the temp file was not licensed correctly. 

Now we use mktemp under RUNNER_TEMP (with TMPDIR / /tmp fallbacks) so the file lives outside the tree RAT scans, and guard the EXIT trap with ${settings_file:-} in case the script exits before the variable is set.

Changes the position of the temporary file

### Are these changes tested?

Only locally, made sure the file writes outside of the working directory

No

### DIFF
--- a/release/bin/prepare-rc.sh
+++ b/release/bin/prepare-rc.sh
@@ -255,7 +255,7 @@ step_summary "Created tag \`${rc_tag}\` at \`${tag_commit}\`"
 step_summary ""
 step_summary "### Nexus Deployment"
 
-settings_file=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/parquet-release-settings.XXXXXX")
+settings_file=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/parquet-release-settings.xml")
 generate_maven_settings "${settings_file}"
 
 maven_deploy "${settings_file}"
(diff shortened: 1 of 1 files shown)


release/bin/prepare-rc.sh
```diff
@@ -255,7 +255,7 @@ step_summary "Created tag \`${rc_tag}\` at \`${tag_commit}\`"
 step_summary ""
 step_summary "### Nexus Deployment"
 
-settings_file=".release-settings.xml"
+settings_file=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/parquet-release-settings.XXXXXX")
```
Fokko: Is the extension on purpose?
```suggestion
settings_file=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/parquet-release-settings.xml")
```

Fokko (approved):

release/bin/prepare-rc.sh
RussellSpitzer: Honeslty I just thought it looked cool when the llm suggested it so I kept it. I don't think there is a real reason for the extension except that it's temporary so it doesn't matter

Fokko:
Thanks again @RussellSpitzer

PR #3765, opened 2026-09-04
Release: Keep random suffix in settings file

Fokko:
### Rationale for this change

TIL: mktemp requires its template to end with at least three trailing X characters (which it replaces with random chars to make a unique name), and using --suffix=.ext lets you keep a file extension after that random run — a template like parquet-release-settings.xml fails because it has no Xs.

Revert "Update release/bin/prepare-rc.sh"

This reverts commit 784c1533f0d5920da10b7f8273b462c77884d21d.


RussellSpitzer (approved):
I have learned something today!

wgtmac (approved):

wgtmac:
> I have learned something today!

Same here :)

PR #3770, opened 2026-09-04
GH-3769: Improve RunLengthBitPackingHybridDecoder.readNext hot path

### LINKED ISSUE #3769: Improve RunLengthBitPackingHybridDecoder.readNext hot path

RunLengthBitPackingHybridDecoder.readNext() allocates a fresh int[], a fresh byte[], and a DataInputStream wrapper on every bit-packed run it decodes. Since this decoder is the primary path for definition levels, repetition levels, and dictionary indices on every column read, those allocations directly inflate young-generation GC pressure on scan-heavy workloads. The existing code even flags it: // TODO: reuse a buffer.

### Component(s)

_No response_

abstractdog:
### Rationale for this change

`RunLengthBitPackingHybridDecoder.readNext()` allocates a fresh `int[]`, a fresh `byte[]`, and a new `DataInputStream` wrapper on every bit-packed run. It's the primary decode path for def/rep levels and dictionary indices on every column read, so those allocations directly inflate young-gen GC on scan-heavy workloads. The code has even flagged it: `// TODO: reuse a buffer`.

- `RunLengthBitPackingHybridDecoder`: reuse `int[] currentBuffer` and `byte[] packedBuffer` as grow-only instance fields; track the current run size in a new `packedRunSize` field (so `readInt()` no longer uses `currentBuffer.length` as the boundary); zero-fill the tail on short end-of-stream reads to preserve the old `new byte[]` implicit-zero semantics; read directly from the underlying `InputStream` instead of wrapping in `DataInputStream`.
- New `RleDecodingBenchmark` in `parquet-benchmarks` to lock in the improvement and guard against regressions.

### Are these changes tested?

Yes. All tests in `parquet-column` pass, including `TestRunLengthBitPackingHybridEncoder` (round-trips through the decoder) and `TestDictionary` (exercises dictionary-index decoding, which is the primary caller). The new JMH benchmark also verifies the behavior end-to-end via encode → decode round trips.

JMH results (100K values decoded, JDK 17, `bitWidth ∈ {1, 3, 10}`):

```
java -jar parquet-benchmarks/target/parquet-benchmarks.jar RleDecodingBenchmark -f 1 -wi 3 -w 1 -i 3 -r 2 -prof gc
```

| Metric | Before | After |
|---|---:|---:|
| Alloc/value | 4.9–5.9 B/op | 0.055–0.060 B/op (~90× less) |
| Alloc rate | ~4.2–4.7 GB/s | ~55–62 MB/s (~70× less) |
| Young-gen GCs/trial | 55–63 | ≈ 0 |
| Throughput | 830–910 M vals/s | 1.06–1.18 B vals/s (+21–28%) |

No. Purely internal implementation change to a single decoder class.

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/RleDecodingBenchmark.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/RleDecodingBenchmark.java
@@ -1,0 +1,126 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.benchmarks;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.apache.parquet.bytes.HeapByteBufferAllocator;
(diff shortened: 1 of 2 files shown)


Fokko:
@abstractdog Thanks for raising this PR @abstractdog. I think this is _very_ similar to https://github.com/apache/parquet-java/pull/3749. Maybe you could help review that one?

abstractdog:
> @abstractdog Thanks for raising this PR @abstractdog. I think this is _very_ similar to #3749. Maybe you could help review that one?

oh, kinda duplicate, I'll close this I guess

abstractdog:
closing this, already being addressed by https://github.com/apache/parquet-java/pull/3749/

PR #3773, opened 2026-09-04
GH-3772: Bulk skip in RunLengthBitPackingHybridDecoder / DictionaryValuesReader

### LINKED ISSUE #3772: Bulk skip in RunLengthBitPackingHybridDecoder / DictionaryValuesReader
### Motivation

Make Hive leverage bulk skip when implementing probe decode for Parquet, similarly to https://issues.apache.org/jira/browse/HIVE-22731, which was about ORC.

### Problem

`ValuesReader.skip(int n)` ships with a naive default:

```java
public void skip(int n) {
  for (int i = 0; i < n; i++) skip();
}
```

For dictionary-encoded columns (the common case), each `skip()` bottoms
out in `RunLengthBitPackingHybridDecoder.readInt()` — a mode switch,
array-index arithmetic, and a value the caller immediately discards.

Any filter-then-skip path (column-index row ranges, hash-join probe
filtering, runtime filters) pays this cost per skipped row.

### Proposal

1. Add `RunLengthBitPackingHybridDecoder.skipInts(int n)` — re-use
   `readNext()` per run, then advance `currentCount` by
   `min(n, currentCount)` instead of walking every value through
   `readInt()`.
2. Override `skip(int)` on `DictionaryValuesReader` and
   `RunLengthBitPackingHybridValuesReader` to call `decoder.skipInts(n)`.

### Component(s)

Core

abstractdog:
### Rationale for this change

`ValuesReader.skip(int n)` defaults to a naive loop of `skip()` calls. For dictionary-encoded columns each `skip()` bottoms out in `RunLengthBitPackingHybridDecoder.readInt()` — a mode switch, array indexing, and a value the caller immediately discards. Any filter-then-skip path (column-index row ranges, hash-join probe filtering, runtime filters) pays this cost per skipped row.

`RleSkipBenchmark`, JMH throughput, JDK 17, 100 k values/op:

| pattern | bitWidth | `readInt()` loop | `skipInts` | speedup |
|---------|---------:|-----------------:|-----------:|--------:|
| rle     |        8 |        1.38 B/s  | 104.9 B/s  |   ~76×  |
| packed  |        8 |        0.89 B/s  |   4.12 B/s |   ~4.6× |
| mixed   |        8 |        0.96 B/s  |   6.99 B/s |   ~7.2× |

- New `RunLengthBitPackingHybridDecoder.skipInts(int)` — re-uses `readNext()` per run, then advances `currentCount` by `min(n, currentCount)` instead of walking every value through `readInt()`.
- `skip(int)` overrides on `DictionaryValuesReader` and `RunLengthBitPackingHybridValuesReader` delegating to `decoder.skipInts(n)`.
- `parquet-benchmarks / RleSkipBenchmark` JMH benchmark.

`readNext()` and the `ValuesReader.skip(int)` default are unchanged.

### Are these changes tested?

`TestRunLengthBitPackingHybridDecoderSkip` covers RLE-only, PACKED-only, mixed, zero-skip, full-skip, mid-run partial skip, `bitWidth = 0`, and a randomised 4 K-value read/skip interleaving cross-checked against a `readInt()` reference. All `parquet-column` tests pass.

No. Additive and binary-compatible: no signatures change; `skipInts(n)` is semantically identical to N discarded `readInt()`s. Existing callers of `ValuesReader.skip(int)` pick up the fast path with no code change.

### DIFF
--- a/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
+++ b/parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
@@ -125,4 +125,15 @@ public class DictionaryValuesReader extends ValuesReader {
       throw new ParquetDecodingException(e);
     }
   }
+
+  @Override
+  public void skip(int n) {
+    // Bulk-skip dictionary keys without decoding them or looking them up in the dictionary.
+    // See RunLengthBitPackingHybridDecoder#skipInts for the fast-path details.
+    try {
+      decoder.skipInts(n);
+    } catch (IOException e) {
+      throw new ParquetDecodingException(e);
+    }
+  }
 }
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/RleSkipBenchmark.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/RleSkipBenchmark.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.benchmarks;
+
+import java.nio.ByteBuffer;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.apache.parquet.bytes.ByteBufferInputStream;
+import org.apache.parquet.bytes.DirectByteBufferAllocator;
+import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridDecoder;
+import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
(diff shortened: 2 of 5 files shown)


abstractdog:
can I ask for a review of this?

made Hive-side benchmark with this change, wonderful results for dictionary columns, see details in https://github.com/apache/hive/pull/6758#issuecomment-5567313232

Fokko (approved):
Nice concise optimization :)

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
divjotarora, line 1: `DictionaryValuesReader#initFromPage` has this code:
```
    this.in = stream.remainingStream();
    if (in.available() > 0) {
      LOG.debug("init from page at offset {} for length {}", stream.position(), stream.available());
      int bitWidth = BytesUtils.readIntLittleEndianOnOneByte(in);
      LOG.debug("bit width {}", bitWidth);
      decoder = new RunLengthBitPackingHybridDecoder(bitWidth, in);
    } else {
      decoder = new RunLengthBitPackingHybridDecoder(1, in) {
        @Override
        public int readInt() throws IOException {
          throw new IOException("Attempt to read from empty page");
        }
      };
    }
  }
```

Specifically in the `else`, we create a special `RunLengthBitPackingHybridDecoder` for empty dictionary pages. Should we override that decoder's `skipInts` to preserve the same `IOException` behavior when skipping? Otherwise, we bypass `readInt` and call `readNext` in the skip path, which throws `IllegalArgumentException`.

divjotarora (approved):
Nice optimization @abstractdog! One small comment to maintain today's exception behavior in the empty dictionary case.

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
abstractdog: makes sense @divjotarora , good catch, thanks! addressed it

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
```diff
@@ -57,7 +59,21 @@ public void initFromPage(int valueCount, ByteBufferInputStream stream) throws IO
...
-          throw new IOException("Attempt to read from empty page");
+          throw new IOException(EMPTY_PAGE_MESSAGE);
+        }
+
+        /**
+         * Skipping goes through the decoder directly instead of {@link #readInt()}, so it needs the
+         * same guard: without it, readNext() would hit the empty stream and throw a raw
+         * IllegalArgumentException that {@link DictionaryValuesReader#skip(int)} does not wrap.
+         * Skipping zero values reads nothing and stays silent, matching the loop that
+         * {@link org.apache.parquet.column.values.ValuesReader#skip(int)} used to run.
```
divjotarora: I prefer not to have comments like "used to run" in code as it's codifying _old_ behavior that's been replaced. Is this actually necessary here? Seems fine to just say "Skipping zero values does nothing and stays silent" without any mention of the old loop.

This is personal preference, so feel free to ignore if you disagree / defer to the maintainers of the repo.

divjotarora (approved):
Nice change @abstractdog!

parquet-column/src/main/java/org/apache/parquet/column/values/dictionary/DictionaryValuesReader.java
abstractdog: agree, tend to take care of the same, missed this time:
https://github.com/apache/parquet-java/pull/3773/changes/80ecda728ef07446d1d0a3dcadacd5b730d1dc7c

PR #3781, opened 2026-09-06
GH-3780: Decouple ParquetReadOptions from hadoop-mapreduce-client-core

### LINKED ISSUE #3780: Decouple parquet-hadoop module from hadoop-mapreduce-client-core

Decouple `ParquetReadOptions` from the legacy Hadoop `ParquetInputFormat`/`FileInputFormat` classes, which currently force pulling in the [`hadoop-mapreduce-client-core`](https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-mapreduce-client-core) dependency (and its transitive JARs).

## Summary

To instantiate a `org.apache.parquet.hadoop.ParquetReader` we need to use `org.apache.parquet.ParquetReadOptions`, which references a set of keys located in `org.apache.parquet.hadoop.ParquetInputFormat` and calls a static `getFilter` method declared also on `ParquetInputFormat`, which `extends org.apache.hadoop.mapreduce.lib.input.FileInputFormat`. 

`ParquetInputFormat` is part of the `parquet-hadoop` module, while `FileInputFormat` is declared in the `hadoop-mapreduce-client-core` JAR from the Hadoop project.

Because `ParquetReader` (via `ParquetReadOptions`) needs to call `ParquetInputFormat.getFilter(...)`, the JVM is forced to initialize `ParquetInputFormat`, and initializing a class triggers the loading and initialization of its superclass (`FileInputFormat`) along with its entire transitive dependency graph (`org.apache.hadoop.mapreduce.*`). For code that only needs to read a plain Parquet file (and not the MapReduce input-format machinery), this pulls an unwanted, heavy Hadoop-`mapreduce` dependency into the classpath and link set. `AvroParquetReader` and `ProtoParquetReader` extend from `ParquetReader` and have the same issue.

`ParquetInputFormat` has three responsibilities:
* define a set of property keys as constants
* deserialize the filter predicates from a configuration value
* support the integration of Parquet files into Hadoop MapReduce

This issue proposes to extract the first two responsibilities into two new Hadoop-agnostic classes, in the `org.apache.parquet.conf` package:
* `ParquetInputProperties`: the property-key constants only
* `ParquetInputFilters`: the filter deserialization logic
so that the configuration can be used without ever loading `FileInputFormat` and its transitive dependencies.

```
ParquetReader                                   ← org.apache.parquet.hadoop (parquet-hadoop)
        │  uses
        ▼
ParquetReadOptions                              ← org.apache.parquet (parquet-hadoop)
        │  (static import keys + getFilter call)
        ▼
ParquetInputFormat.getFilter(...)               ← org.apache.parquet.hadoop (parquet-hadoop)
        │  extends
        ▼
FileInputFormat<Void, T>                        ← org.apache.hadoop.mapreduce.lib.input (hadoop-mapreduce-client-core)
        │  extends
        ▼
InputFormat<K, V>                               ← org.apache.hadoop.mapreduce.lib.input (hadoop-mapreduce-client-core)
        │  (transitive)
        ▼
{ InputSplit, JobContext, TaskAttemptContext,
  RecordReader, ... }                           ← org.apache.hadoop.mapreduce* (hadoop-mapreduce-client-core)
```

With the new `ParquetInputProperties` / `ParquetInputFilters`, building `ParquetReadOptions` no longer touches any `org.apache.hadoop.mapreduce` type, so consumers not related to Hadoop avoid transitively including the MapReduce dependency.

## Public API / Behavioral change

No behavior change. This is a refactoring:

- New classes `org.apache.parquet.conf.ParquetInputProperties` and
  `org.apache.parquet.conf.ParquetInputFilters` (in `parquet-hadoop`) hold the constants and the
  `ParquetConfiguration`-based filter resolution respectively.
- The legacy `org.apache.hadoop.conf.Configuration`-based entry points remain on
  `ParquetInputFormat` (they are used only via the legacy MapReduce path) and are kept for binary /
  source compatibility.
- All pre-existing constants on `ParquetInputFormat` are now `@Deprecated` and delegate to the new
  classes; source and binary compatibility are preserved.

## Acceptance criteria

- `ParquetReadOptions` (used for plain file reads) references only `ParquetInputProperties` / `ParquetInputFilters`, never `ParquetInputFormat` / `org.apache.hadoop.mapreduce.InputFormat`.
- Loading `ParquetInputProperties` / `ParquetInputFilters` / `ParquetReadOptions` does not initialize `FileInputFormat`.
- All existing tests still pass (`./mvnw test`).

### Component(s)

Core

jerolba:
### Rationale for this change

Reading the Parquet read configuration via `ParquetReadOptions` previously referenced
`ParquetInputFormat.getFilter(...)` and statically imported its constants. `ParquetInputFormat` extends
`org.apache.hadoop.mapreduce.lib.input.FileInputFormat`, so using the read configuration forced the JVM
to initialize `ParquetInputFormat` and therefore `FileInputFormat` and its whole
`org.apache.hadoop.mapreduce.*` transitive dependency graph, even though only plain `String`/`boolean`
config properties were needed. The new `ParquetInputProperties` (constants) and `ParquetInputFilters`
(filter resolution) centralise the `ParquetConfiguration`-based read configuration so that read-only
consumers no longer transitively pull in the Hadoop `mapreduce` dependency.

#### New files

- `parquet-hadoop/src/main/java/org/apache/parquet/conf/ParquetInputProperties.java`
  - Declares all Parquet read-configuration constants previously on `ParquetInputFormat`:
    `READ_SUPPORT_CLASS`, `UNBOUND_RECORD_FILTER`, `STRICT_TYPE_CHECKING`, `FILTER_PREDICATE`,
    `RECORD_FILTERING_ENABLED`, `STATS_FILTERING_ENABLED`, `DICTIONARY_FILTERING_ENABLED`,
    `COLUMN_INDEX_FILTERING_ENABLED`, `PAGE_VERIFY_CHECKSUM_ENABLED`, `BLOOM_FILTERING_ENABLED`,
    `OFF_HEAP_DECRYPT_BUFFER_ENABLED`, `HADOOP_VECTORED_IO_ENABLED`, `HADOOP_VECTORED_IO_DEFAULT`.
  - Constants only; no `org.apache.hadoop.mapreduce` dependency.
- `parquet-hadoop/src/main/java/org/apache/parquet/conf/ParquetInputFilters.java`
  - `ParquetConfiguration`-based filter resolution: `getFilter(ParquetConfiguration)`,
    `getUnboundRecordFilter(ParquetConfiguration)` and `getFilterPredicate(ParquetConfiguration)`.
  - No `org.apache.hadoop.mapreduce` dependency.

#### Removed

- `parquet-hadoop/src/main/java/org/apache/parquet/ParquetInputConfiguration.java` (intermediate
  class, split into the two classes above).

#### Modified

- `parquet-hadoop/.../ParquetReadOptions.java`
  - Static imports now resolve to `ParquetInputProperties`, and the `getFilter(...)` call to
    `ParquetInputFilters` instead of `ParquetInputFormat`. `ParquetReadOptions` no longer references
    `ParquetInputFormat`.
- `parquet-hadoop/.../hadoop/InternalParquetRecordReader.java`
  - Static imports of `RECORD_FILTERING_ENABLED` / `STRICT_TYPE_CHECKING` now from
    `ParquetInputProperties`.
- `parquet-hadoop/.../hadoop/ParquetInputFormat.java`
  - Constants are now `@Deprecated`, delegating to `ParquetInputProperties.*`.
  - Kept the legacy Hadoop `Configuration`-based `getFilter(Configuration)` /
    `getUnboundRecordFilter(Configuration)` overloads (used by the MapReduce read path), delegating by
    wrapping into `HadoopParquetConfiguration`. Internal callers use `ParquetInputFilters` /
    `ParquetInputProperties` directly.
- Tests updated accordingly (`DeprecatedInputFormatTest`, `TestParquetFileWriter`,
  `TestInputOutputFormat`, `TestInputFormatColumnProjection`, `TestDataPageChecksums`,
  `TestColumnChunkPageWriteStore`, `TestPropertiesDrivenEncryption`).

### Are these changes tested?

No change in behavior, just refactoring code location. Code compilation and existing tests validate the change.

No. Fully backward and binary compatible: the `ParquetInputFormat` constants and methods are retained
  (deprecated) and delegate to the new class; constant string/boolean values are unchanged, so any
  previously-serialised configuration still works.

Closes #3780

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java
@@ -19,17 +19,17 @@
 
 package org.apache.parquet;
 
+import static org.apache.parquet.conf.ParquetInputFilters.getFilter;
+import static org.apache.parquet.conf.ParquetInputProperties.BLOOM_FILTERING_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.COLUMN_INDEX_FILTERING_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.DICTIONARY_FILTERING_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.HADOOP_VECTORED_IO_DEFAULT;
+import static org.apache.parquet.conf.ParquetInputProperties.HADOOP_VECTORED_IO_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.OFF_HEAP_DECRYPT_BUFFER_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.PAGE_VERIFY_CHECKSUM_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.RECORD_FILTERING_ENABLED;
+import static org.apache.parquet.conf.ParquetInputProperties.STATS_FILTERING_ENABLED;
 import static org.apache.parquet.format.converter.ParquetMetadataConverter.NO_FILTER;
-import static org.apache.parquet.hadoop.ParquetInputFormat.BLOOM_FILTERING_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.COLUMN_INDEX_FILTERING_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.DICTIONARY_FILTERING_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.HADOOP_VECTORED_IO_DEFAULT;
-import static org.apache.parquet.hadoop.ParquetInputFormat.HADOOP_VECTORED_IO_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.OFF_HEAP_DECRYPT_BUFFER_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.PAGE_VERIFY_CHECKSUM_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.RECORD_FILTERING_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.STATS_FILTERING_ENABLED;
-import static org.apache.parquet.hadoop.ParquetInputFormat.getFilter;
 import static org.apache.parquet.hadoop.UnmaterializableRecordCounter.BAD_RECORD_THRESHOLD_CONF_KEY;
 
 import java.util.Collections;
--- a/parquet-hadoop/src/main/java/org/apache/parquet/conf/ParquetInputFilters.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/conf/ParquetInputFilters.java
@@ -1,0 +1,84 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.parquet.conf;
+
+import java.io.IOException;
+import org.apache.parquet.filter.UnboundRecordFilter;
+import org.apache.parquet.filter2.compat.FilterCompat;
+import org.apache.parquet.filter2.compat.FilterCompat.Filter;
+import org.apache.parquet.filter2.predicate.FilterPredicate;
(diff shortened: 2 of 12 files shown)


dossett:
+1 (non-binding), this seems delightful!

It would be really nice to have a test that proves a file can be read without the hadoop libraries in the classpath or otherwise accessible.  I think the easiest way to do that would be to create a new maven module that excludes `hadoop-mapreduce-client-core` entirely and write a unit test in that module that shows `ParquetReadOptions.builder().build()` works.  That seems like a lot for a single unit test, but it would provide a solid guarantee.

jerolba:
I agree that it would be nice to have it tested, but it will add a lot of complexity to the project just for a test setup.

I've tested this change in [Carpet](https://github.com/jerolba/parquet-carpet/) removing `hadoop-mapreduce-client-core` [dependency](https://github.com/jerolba/parquet-carpet/blob/master/carpet-record/build.gradle#L27) and, based on the code path used by the library, it works. The change doesn't ensure that usages of other `parquet-hadoop` classes will not require the dependency.

Unfortunately, we are far from removing the dependency of `hadoop-common`. I think that it will require a complete rewrite of the `parquet-hadoop` module.

dossett:
I was curious how complex the test would be, and it doesn't seem too bad.  What do you think?

https://github.com/dossett/parquet-java/pull/1/changes

(that's a hypothetical PR against this branch)

jerolba:
For me it's also more important to verify that `hadoop-mapreduce-client-core` is not required for writing and reading a local file.
`InternalParquetRecordReader` and `InternalParquetRecordWriter` can involve, directly or indirectly, some Hadoop classes.

On top of your PR I've added a test that reads/writes a file:
https://github.com/dossett/parquet-java/pull/2/changes

I have no strong opinion, but I don't feel comfortable adding a complete Maven module just for testing a special use case.

Let's see what the main committers think.

PR #3566, opened 2026-05-17
GH-3530: Optimize DICTIONARY encoding/decoding data structures and use ByteBuffer

iemejia:
Part of #3530 — Apache Parquet Java Performance Improvements

## Summary

Optimize dictionary encoding and decoding data structures.

**Encoding**:
- Replace `LinkedOpenHashMap` with `OpenHashMap` + `ArrayList` for all `DictionaryValuesWriter` subclasses, eliminating insertion-order linked-list overhead and enabling O(1) indexed access for dictionary page serialization and fallback.
- Make `IntList.size()` O(1) by tracking `totalSize` incrementally instead of summing across slab arrays.

**Decoding**:
- Convert `PlainValuesDictionary` numeric constructors (INT32, INT64, FLOAT, DOUBLE) from `InputStream`-based per-byte reads to direct `ByteBuffer.getInt`/`getLong`/`getFloat`/`getDouble`.

**Binary hashCode caching**:
- Cache `hashCode()` for `Binary` instances not backed by reusable byte arrays, avoiding redundant recomputation during dictionary hash-map probes.

JMH benchmarks: `DictionaryEncodingBenchmark`, `DictionaryDecodingBenchmark` with `TestDataFactory` and `BenchmarkEncodingUtils`.

## Benchmark results

**Environment**: JDK 25.0.3 (Temurin), OpenJDK 64-Bit Server VM, JMH 1.37, Linux x86_64.

Encoding (100K values/iteration, 2 averaged runs):

| Benchmark | Baseline (M ops/s) | Optimized (M ops/s) | Speedup |
|---|---:|---:|---:|
| encodeInt HIGH_CARD | 14.9 | 23.5 | **1.58x** |
| encodeLong HIGH_CARD | 12.0 | 19.2 | **1.60x** |
| encodeFloat HIGH_CARD | 14.4 | 21.9 | **1.52x** |
| encodeDouble HIGH_CARD | 11.7 | 17.9 | **1.53x** |
| encodeBinary LOW len=10 | 75.6 | 125.6 | **1.66x** |
| encodeBinary LOW len=100 | 13.2 | 107.8 | **8.2x** |
| encodeBinary LOW len=1000 | 1.5 | 148.3 | **~100x** |
| encodeBinary HIGH len=10 | 6.4 | 13.2 | **2.1x** |
| encodeFlba HIGH len=12 | 6.3 | 15.4 | **2.4x** |
| encodeFlba HIGH len=16 | 6.1 | 14.6 | **2.4x** |
[...]

PR #3782, opened 2026-09-07
Use JDK intrinsics for Binary equality/comparison/hashing

abstractdog:
### Rationale for this change

`Binary#equals`, `Binary#lexicographicCompare`, and `Binary#hashCode` sit on hot paths for statistics min/max maintenance, dictionary hash-map probing, predicate evaluation, and bloom-filter build — so any per-byte overhead is amortized across every reader and writer. The current implementations use hand-rolled scalar byte loops; the JDK 9+ range overloads of `Arrays.equals` / `Arrays.compareUnsigned` / `Arrays.hashCode` / `ByteBuffer.mismatch` route through `ArraysSupport.vectorizedMismatch`, an `@IntrinsicCandidate` helper HotSpot substitutes with a SIMD byte-scan (SSE / AVX2 / NEON).

- Replace the scalar loops in `Binary.java` for `equals`, `lexicographicCompare`, and `hashCode` with the intrinsic-backed `Arrays.*` / `ByteBuffer.mismatch` APIs across all four `Binary` shape pairs (`byte[]`/`byte[]`, `byte[]`/`ByteBuffer`, `ByteBuffer`/`ByteBuffer`).
- Add a JMH suite `parquet-benchmarks/BinaryComparisonBenchmark` covering `equals` / `compareTo` / `hashCode` at length regimes 8, 64, and 512, both worst-case (full match) and average-case (mid-length mismatch).

Follows the same technique already accepted for `DeltaByteArrayWriter` in #3465. Complementary to the `Binary#hashCode` caching in #3566 — caching reduces call frequency, intrinsics speed up the calls that remain plus `equals` / `compareTo` which caching does not address.

### Are these changes tested?

Yes. Covered by the existing `Binary` unit tests (equality, comparison, hashing across all shape pairs) plus the new JMH suite. `compareTo` semantics are preserved — `Arrays.compareUnsigned` matches the contract of the previous loop (unsigned byte order, shorter-runs-first on prefix match).

No API changes. Behavior is identical; only the implementation is faster.

**Benchmark** — JDK 17, 1 fork, 3×1s warmup, 5×1s measurement, throughput (ops/s):

| bench                       | len | before | after   | speedup |
| --------------------------- | --: | -----: | ------: | ------: |
| `equalsMatch_bytesBytes`    | 512 |  9.1 M |  39.1 M |   4.30× |
| `equalsMatch_bytesBytes`    |  64 | 41.4 M | 160.9 M |   3.89× |
| `equalsMismatch_bytesBytes` | 512 | 14.6 M |  55.5 M |   3.79× |
| `compareTo_bytesBytes`      | 512 | 15.1 M |  58.8 M |   3.91× |
| `compareTo_bytesBytes`      |  64 | 59.0 M | 199.4 M |   3.38× |
| `equalsMismatch_bytesBuf`   |  64 | 63.8 M | 190.2 M |   2.98× |

Wins scale with length — at len=512 the SIMD lane count amortizes fully (≈4×); at len=64 it's ≈3–4×; len=8 sees only 1.1–1.7× because there's barely enough work for one SIMD lane. `hashCode` is unchanged in JDK 17 numbers — the `31*h+b` polynomial has a serial cross-iteration dependency; HotSpot only gained a lane-split intrinsic (`ArraysSupport.vectorizedHashCode`) in JDK 21, so the new `Arrays.hashCode` call is equivalent on 17 and picks up the vectorization automatically on 21+.

For further reference on intrinsics, see:

https://github.com/openjdk/jdk17u/blob/1ed3717f5a226981a1ebeff70dcafadc74569190/src/java.base/share/classes/java/util/Arrays.java#L2582

https://github.com/openjdk/jdk17u/blob/1ed3717f5a226981a1ebeff70dcafadc74569190/src/hotspot/share/opto/library_call.cpp#L5393

### DIFF
--- a/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BinaryComparisonBenchmark.java
+++ b/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/BinaryComparisonBenchmark.java
@@ -1,0 +1,199 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.parquet.benchmarks;
+
+import java.nio.ByteBuffer;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.apache.parquet.io.api.Binary;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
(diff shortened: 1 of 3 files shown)


Fokko:
Thanks for raising this PR @abstractdog, very interesting. Any way to verify using tests that we don't change hashes?

abstractdog:
> Thanks for raising this PR @abstractdog, very interesting. Any way to verify using tests that we don't change hashes?

good catch! addressed in https://github.com/apache/parquet-java/pull/3782/changes/ce1006b67eda198260edb09de309a06ce62216fd

the test cases now validate against a fixed truth (generated by `scalarHashCode`), also checked pre/post patch

PR #3785, opened 2026-09-09
GH-3683: Support programmatic KMS client factories

### LINKED ISSUE #3683: Allow supplying a KmsClient instance/supplier instead of only a reflectively-instantiated class name

The key-tools KMS integration (`org.apache.parquet.crypto.keytools`) instantiates the `KmsClient` purely by reflection from a class name: `KeyToolkit.getKmsClient(...)` reads `parquet.encryption.kms.client.class` and calls `newInstance()`, requiring a public no-arg constructor, with credentials expected to arrive later via `KmsClient.initialize(conf, kmsInstanceID, kmsInstanceURL, accessToken)`. The same reflective no-arg pattern applies to `CryptoFactory` / `DecryptionPropertiesFactory` / `EncryptionPropertiesFactory` via `parquet.crypto.factory.class`.

This works well when a KMS client is stateless and can bootstrap all of its credentials from the `Configuration` plus the access-token string. It does not work for clients that must be *constructed with* their dependencies and can't be reduced to a class name + string token, e.g.:

- clients holding a live, credential-bearing SDK handle (federated / workload-identity credentials that aren't representable as a token string);
- clients created and wired by a dependency-injection container;
- in-memory / fake KMS clients used in tests, which carry per-test state and have no meaningful no-arg form.

For these, users must build a static side-channel: register the real instance in a static map keyed by a UUID written into the `Configuration`, point `parquet.encryption.kms.client.class` at a thin reflective shim that looks the instance back up in `initialize()`, and override `parquet.encryption.kms.instance.id` per instance to avoid colliding on `KeyToolkit`'s per-`(kmsInstanceID, accessToken)` client cache. That's global mutable state with its own lifecycle/leak management and a one-`Configuration`-per-client invariant — boilerplate every such user reinvents.

**Proposal.** Add an opt-in, fully backward-compatible way to supply a pre-built `KmsClient` (or a `Supplier<KmsClient>` / small factory) programmatically, which `KeyToolkit.getKmsClient(...)` prefers over class-name reflection when present. Reflection stays the default, so existing configs are untouched. Rough shape (names TBD):

```java
// today (still works):
conf.set("parquet.encryption.kms.client.class", "com.example.MyKmsClient");

// proposed addition:
KeyToolkit.setKmsClientFactory(conf, () -> myPreBuiltKmsClient); // or a KmsClientFactory
```

`initialize(...)` would still be invoked on the supplied instance, so credential/token plumbing is unchanged.

**Scope / non-goals.** No new dependencies and no vendor-specific code — this is only about *how* a `KmsClient` is provided, not *which* one. `PropertiesDrivenCryptoFactory`, the `KeyMaterial` format, caching, per-column keys, and key-rotation tooling are all unchanged; this only lifts the requirement that the client be reflectively no-arg constructible.

**Question for maintainers.** Would a change along these lines be welcome? And do you prefer (a) a `Supplier<KmsClient>` set on the `Configuration` / read-write options, or (b) a settable `KmsClientFactory` on `KeyToolkit`? Happy to implement it and open a PR (with tests using a non-no-arg client) once there's agreement on direction.

### Component(s)

Core

stevenwarejones:
### What changes?

- Add a public `KmsClientFactory` and `KeyToolkit.setKmsClientFactory(Configuration, KmsClientFactory)`.
- Pass the current configuration and resolved KMS instance ID, URL, and access token to the factory. Each invocation returns a distinct client that `KeyToolkit` initializes and caches.
- Prefer the registered factory while keeping class-name reflection as the fallback.
- Preserve registrations across same-JVM `Configuration` copies using an opaque registration ID.
- Isolate KMS-client and KEK caches by factory registration, and isolate write-side KEKs by KMS instance within a factory registration without changing the pre-existing reflective cache.
- Add `removeKmsClientFactory(Configuration)` and document the API and lifecycle in the README.

### Why?

Some KMS clients need live, constructor-injected dependencies such as SDK clients or dependency-injection state. Those clients cannot be created through a public no-argument constructor and should not need a static registry as a side channel.

### Testing

- Focused `KeyToolkitTest` and `TestKmsUrlRead` suites pass: 23 tests, 0 failures.
- Regression coverage verifies factory registration and lifecycle, copied configurations, missing-factory failure, cache isolation between registrations, KMS-instance isolation within a registration, and valid KEK reuse within the same KMS instance.
- End-to-end reader and writer tests exercise factory-created clients through the public Parquet APIs.

Closes #3683

### DIFF
--- a/parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
+++ b/parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
@@ -22,10 +22,11 @@ package org.apache.parquet.crypto.keytools;
 import java.io.IOException;
 import java.util.Base64;
 import java.util.Collections;
-import java.util.IdentityHashMap;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentMap;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
@@ -48,6 +49,9 @@ public class KeyToolkit {
    * KMS stands for “key management service”.
    */
   public static final String KMS_CLIENT_CLASS_PROPERTY_NAME = "parquet.encryption.kms.client.class";
+
+  private static final String KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME =
+      "parquet.encryption.kms.client.factory.registration.id";
   /**
    * ID of the KMS instance that will be used for encryption (if multiple KMS instances are available).
    */
@@ -123,10 +127,10 @@ public class KeyToolkit {
   private static final KmsClientCacheContext DEFAULT_KMS_CLIENT_CACHE_CONTEXT = new KmsClientCacheContext(
       null, KMS_CLIENT_CACHE_PER_TOKEN, KEK_WRITE_CACHE_PER_TOKEN, KEK_READ_CACHE_PER_TOKEN);
 
-  // Programmatically supplied factories and their caches, scoped to the exact Configuration object.
-  // Callers must remove registrations when the Configuration is no longer in use.
-  private static final Map<Configuration, KmsClientCacheContext> KMS_CLIENT_FACTORY_REGISTRATIONS =
-      Collections.synchronizedMap(new IdentityHashMap<>());
+  // Programmatically supplied factories and their caches, scoped by an ID copied with Configuration.
+  // Callers must remove registrations when the Configuration and its copies are no longer in use.
+  private static final Map<String, KmsClientCacheContext> KMS_CLIENT_FACTORY_REGISTRATIONS =
+      Collections.synchronizedMap(new HashMap<>());
 
   private enum KmsClientCache {
     INSTANCE;
@@ -369,14 +373,15 @@ public class KeyToolkit {
    * KeyToolkit} in the same way as reflectively constructed clients. The KMS client and key
    * encryption key caches are isolated from registrations for other configurations.
    *
-   * <p>The registration is associated with the exact {@code Configuration} object, so changing
-   * that object's properties does not affect the registration. The factory receives the current
-   * configuration and resolved KMS details when it creates a client.
+   * <p>The registration ID is stored in the {@code Configuration}, so copies made in the same JVM
+   * share the factory and caches. The factory receives the current configuration and resolved KMS
+   * details when it creates a client.
    *
-   * <p>The association is not serialized, and configuration copies must register their own
-   * factory. The caller must invoke {@link #removeKmsClientFactory(Configuration)} after all
-   * readers and writers using the configuration have closed. Replacing a factory clears the
-   * previous registration and its caches.
+   * <p>The factory itself is local to this JVM. A configuration deserialized in another JVM must
+   * register its factory before use. The caller must invoke {@link
+   * #removeKmsClientFactory(Configuration)} after all readers and writers using the configuration
+   * or its copies have closed. Replacing a factory clears the previous registration and its
+   * caches.
    *
    * @param configuration Hadoop configuration associated with the factory
    * @param kmsClientFactory factory used to create KMS clients
@@ -384,8 +389,13 @@ public class KeyToolkit {
   public static void setKmsClientFactory(Configuration configuration, KmsClientFactory kmsClientFactory) {
     Objects.requireNonNull(configuration, "configuration");
     Objects.requireNonNull(kmsClientFactory, "kmsClientFactory");
+    String registrationId = configuration.getTrimmed(KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME);
+    if (stringIsEmpty(registrationId)) {
+      registrationId = UUID.randomUUID().toString();
+      configuration.set(KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME, registrationId);
+    }
     KmsClientCacheContext previous =
-        KMS_CLIENT_FACTORY_REGISTRATIONS.put(configuration, new KmsClientCacheContext(kmsClientFactory));
+        KMS_CLIENT_FACTORY_REGISTRATIONS.put(registrationId, new KmsClientCacheContext(kmsClientFactory));
     if (previous != null) {
       previous.clear();
     }
@@ -401,7 +411,12 @@ public class KeyToolkit {
    */
   public static void removeKmsClientFactory(Configuration configuration) {
     Objects.requireNonNull(configuration, "configuration");
-    KmsClientCacheContext registration = KMS_CLIENT_FACTORY_REGISTRATIONS.remove(configuration);
+    String registrationId = configuration.getTrimmed(KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME);
+    if (stringIsEmpty(registrationId)) {
+      return;
+    }
+    configuration.unset(KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME);
+    KmsClientCacheContext registration = KMS_CLIENT_FACTORY_REGISTRATIONS.remove(registrationId);
     if (registration != null) {
       registration.clear();
     }
@@ -479,8 +494,15 @@ public class KeyToolkit {
   }
 
   static KmsClientCacheContext getKmsClientCacheContext(Configuration configuration) {
-    KmsClientCacheContext cacheContext = KMS_CLIENT_FACTORY_REGISTRATIONS.get(configuration);
-    return cacheContext == null ? DEFAULT_KMS_CLIENT_CACHE_CONTEXT : cacheContext;
+    String registrationId = configuration.getTrimmed(KMS_CLIENT_FACTORY_REGISTRATION_ID_PROPERTY_NAME);
+    if (stringIsEmpty(registrationId)) {
+      return DEFAULT_KMS_CLIENT_CACHE_CONTEXT;
+    }
+    KmsClientCacheContext cacheContext = KMS_CLIENT_FACTORY_REGISTRATIONS.get(registrationId);
+    if (cacheContext == null) {
+      throw new ParquetCryptoRuntimeException("No KmsClientFactory is registered for this configuration");
+    }
+    return cacheContext;
   }
 
   private static void clearKekWriteCaches() {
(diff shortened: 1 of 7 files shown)


stevenwarejones:
Hi @ggershinsky, @wgtmac, and @shangxinli — would you be willing to review this change? It implements the KmsClientFactory direction discussed in #3683, including explicit registration cleanup and isolation of the KMS and KEK caches.

ggershinsky:
Sure, thanks

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
```diff
@@ -281,15 +343,64 @@ public static void rotateMasterKeys(String folderPath, Configuration hadoopConfi
...
+   * previous registration and its caches.
+   *
+   * @param configuration Hadoop configuration associated with the factory
+   * @param kmsClientFactory factory used to create KMS clients
+   */
+  public static void setKmsClientFactory(Configuration configuration, KmsClientFactory kmsClientFactory) {
+    Objects.requireNonNull(configuration, "configuration");
+    Objects.requireNonNull(kmsClientFactory, "kmsClientFactory");
+    KmsClientCacheContext previous =
+        KMS_CLIENT_FACTORY_REGISTRATIONS.put(configuration, new KmsClientCacheContext(kmsClientFactory));
```
ggershinsky: What if a shell/notebook user adds some parameter to the config during a session?
They'll need to call `removeKmsClientFactory` and then `setKmsClientFactory` each time?
Are there usecases where config changes are hard to trace?
Maybe there is an alternative approach? (eg using something similar to the kms instance, say "parquet.encryption.kms.factory.instance")

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KmsClientFactory.java
```diff
@@ -0,0 +1,32 @@
...
+/** Factory for creating {@link KmsClient} instances with programmatically supplied dependencies. */
+@FunctionalInterface
+public interface KmsClientFactory {
+
+  /**
+   * Creates a KMS client. {@link KeyToolkit} initializes the returned client before using it.
+   *
+   * @return a new or pre-built KMS client
+   */
+  KmsClient createKmsClient();
```
ggershinsky: should we pass the current Configuration object here? It might have a useful input for creation of custom KMS clients.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
stevenwarejones: Mutating the same `Configuration` does not require deregistration. Registrations use object identity, so changes to its properties do not affect lookup.

I added a test that mutates the configuration after registration and verifies that the factory remains registered and receives the updated configuration. If a client has already been cached, changing a property will not recreate it; that is also the behavior of the existing reflective path. Calling `setKmsClientFactory` again replaces the registration and clears its caches, so a separate remove call is not needed.

A `parquet.encryption.kms.factory.instance` string would require a static ID-to-object registry because `Configuration` cannot contain the live factory itself. That would recreate the side channel this API is intended to remove and would have ambiguous behavior when a copied configuration is serialized to another JVM. Configuration copies therefore still need their own explicit registration.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KmsClientFactory.java
stevenwarejones: Done. I went slightly further and now pass the full creation context: `Configuration`, KMS instance ID, KMS instance URL, and access token. That lets an immutable or constructor-injected implementation select dependencies using the resolved values before `initialize` is called.

I also documented that each factory invocation must return a distinct, uninitialized client, since the factory is invoked separately for uncached token/KMS-instance combinations. Tests now verify the complete context and separate initialized clients for two access tokens.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
ggershinsky: Sounds good, thanks. Indeed, the `Configuration.equals` implementation ignores the content.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KmsClientFactory.java
ggershinsky: Cool, it certainly makes sense to add these three.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
ggershinsky: One more question on this. What if Spark/Flink/etc copies a `Configuration` into another object (today or in the future), some time during a session?

stevenwarejones: Good point. That would be a copied `Configuration`, so the identity-based registration would not be found.

One option is to store an opaque registration ID in the configuration. That would support copies within the same JVM, but it could not carry the live factory into another JVM. Transparent cross-JVM support would require Spark/Flink-side integration to register or create the factory on each worker.

Would same-JVM copy support, with an explicit error when no local factory is registered, be the right scope here? Or would you suggest a different approach?

ggershinsky: Nope, this option sounds good to me. As for an error, can you check that an exception is indeed thrown? To make sure a dataframe is not written unencrypted silently, when no factory or kms class parameter are found.

stevenwarejones: Done.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
```diff
@@ -109,13 +117,21 @@ public class KeyToolkit {
   // KMS client two level cache: token -> KMSInstanceId -> KmsClient
   static final TwoLevelCacheWithExpiration<KmsClient> KMS_CLIENT_CACHE_PER_TOKEN = KmsClientCache.INSTANCE.getCache();
 
-  // KEK two level cache for wrapping: token -> MEK_ID -> KeyEncryptionKey
-  static final TwoLevelCacheWithExpiration<KeyEncryptionKey> KEK_WRITE_CACHE_PER_TOKEN =
+  // KEK cache for wrapping: token -> KMS instance ID -> master key ID -> KeyEncryptionKey
+  static final TwoLevelCacheWithExpiration<ConcurrentMap<String, KeyEncryptionKey>> KEK_WRITE_CACHE_PER_TOKEN =
```
ggershinsky: I'm somewhat uneasy about changing the existing functionality. Is this required for the new feature?

parquet-hadoop/src/test/java/org/apache/parquet/crypto/keytools/KeyToolkitTest.java
```diff
@@ -0,0 +1,487 @@
...
+  @Test
+  public void missingFactoryForConfigurationCopyFailsEncryptionPropertiesCreation() {
+    Configuration configuration = new Configuration(false);
+    configuration.set(PropertiesDrivenCryptoFactory.UNIFORM_KEY_PROPERTY_NAME, MASTER_KEY_ID);
+    setKmsClientFactory(
+        configuration, (conf, kmsId, kmsUrl, token) -> new ConstructorInjectedKmsClient("dependency"));
+    Configuration copy = new Configuration(configuration);
+    KeyToolkit.removeKmsClientFactory(configuration);
+
+    assertThatThrownBy(() -> new PropertiesDrivenCryptoFactory()
```
ggershinsky, line 131: can you add a basic unitest that asserts an exception when neither factory nor client property are set?

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
stevenwarejones: What if I scope the KMS-instance cache change to factory registrations and leave the pre-existing reflective cache unchanged, or do you have concerns about the factory-specific change as well?

parquet-hadoop/src/test/java/org/apache/parquet/crypto/keytools/KeyToolkitTest.java
stevenwarejones: Done. The new test requests encryption without a factory or client class and verifies that it fails with `ParquetCryptoRuntimeException`.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
stevenwarejones: I went ahead and restored the pre-existing cache. Now, only factory registrations scope KEKs by KMS instance.

parquet-hadoop/src/main/java/org/apache/parquet/crypto/keytools/KeyToolkit.java
```diff
@@ -143,6 +160,80 @@ private TwoLevelCacheWithExpiration<byte[]> getCache() {
     }
   }
 
+  static final class KmsClientCacheContext {
+    private final KmsClientFactory factory;
+    private final TwoLevelCacheWithExpiration<KmsClient> kmsClientCache;
+    private final TwoLevelCacheWithExpiration<KeyEncryptionKey> defaultKekWriteCache;
+    private final TwoLevelCacheWithExpiration<ConcurrentMap<String, KeyEncryptionKey>> factoryKekWriteCache;
```
ggershinsky, line 167: just curious why the write-side KEKs are separated by KMS client instance id?
(e.g., the same is not done for read-side KEKs).

stevenwarejones: The write-side cache is keyed by master-key ID, and the same ID can refer to different keys in different KMS instances. Without the KMS instance in the cache key, one factory-created client could reuse a KEK wrapped by another.

The read-side cache is keyed by the randomly generated 128-bit KEK ID stored in the key material, which identifies the specific KEK. It is also already isolated by factory registration, so I left its existing structure unchanged.

ggershinsky: Sounds good, thanks.

ggershinsky (approved):

ggershinsky:
Let's leave this open for a week, in case there are other comments from the community.

ggershinsky:
Thanks @stevenwarejones

PR #3787, opened 2026-09-10
Prepare RC: push the RC tag only after staging succeeds

RussellSpitzer:
### Rationale for this change
In the release automation we pushed the tag before we were sure the staging artifacts were successfully pushed. This means we were burning an RC even if the stage failed.

Tag is still created before staging, but not pushed until after staging.

### Are these changes tested?
No

No

### DIFF
--- a/release/bin/prepare-rc.sh
+++ b/release/bin/prepare-rc.sh
@@ -237,17 +237,16 @@ else
 fi
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wgtmac (approved):

Fokko (approved):
Makes sense, thanks @RussellSpitzer and thanks @wgtmac for the review 🚀

PR #3397, opened 2026-02-17
Parquet Java ALP Implementation

vinooganesh:
cc @julienledem @alamb @emkornfield @prtkgaur

### Rationale for this change

Reworks the ALP encoding implementation to address emkornfield's architectural feedback on PR #3390. The original buffered all values in memory and decoded eagerly. This makes the writer incremental (encode per-vector as values arrive) and the reader lazy (decode on demand), matching how other Parquet encodings work.

Builds on Julien Le Dem's original implementation (https://github.com/apache/parquet-java/pull/3390). File structure, integration points, core math, and interop test infrastructure all come from his work. The rework focused on the internal writer/reader plumbing.

### What changes are included in this PR?

**Architecture (addressing review feedback):**
- Incremental writer. Values buffer in a fixed-size vector, each full vector encodes and flushes immediately.
- Lazy reader. Vectors decode on first access via offset array, skip() is O(1).
- Interleaved page layout so each vector is self-contained.
- Extracted `AlpValuesReader` abstract base class for shared decode logic (float/double readers only implement `decodeBody`).
- Preset caching. Full parameter search for first 8 vectors, top 5 combos cached for the rest.

**Spec compliance:**
- Fixed packed data size formula to `ceil(n * bitWidth / 8)` (reusing `BytesUtils.paddedByteCountFromBits`).
- Signed frame-of-reference range in the size estimator (a prior unsigned `max - min` overstated the span on mixed-sign vectors).
- Reads little-endian through `ByteBuffer`'s typed getters on a `LITTLE_ENDIAN`-ordered buffer.
- Uses parquet-encoding's `BytePacker` instead of custom bit-packing.
- Capped max vector size at 32768 to prevent uint16 overflow in `num_exceptions`.
- bitWidth bounds checks in the readers (`> 32` float / `> 64` double throw).

**Configuration:**
- `withAlp()`, `withAlp(AlpConfig)`, `withAlp(columnPath)` and `withAlp(columnPath, AlpConfig)` on `ParquetProperties.Builder` and the Hadoop `ParquetWriter.Builder`, plus `withoutAlp()` / `withoutAlp(columnPath)` to turn it back off globally or for one column.
- `AlpConfig` carries only the vector size (default 1024, validated in its constructor). A column is ALP encoded when it has a config at all.
- Threaded through `DefaultV1ValuesWriterFactory` and `DefaultV2ValuesWriterFactory` so per-column overrides work.
[...]

PR #3791, opened 2026-09-13
MINOR: Update inlined parquet.thrift to parquet-format 2.14.0

vinooganesh:
Brings the inlined `parquet.thrift` up from parquet-format 2.13.0 to 2.14.0, via `dev/update-parquet-thrift.sh apache-parquet-format-2.14.0`. The file is byte-identical to upstream at that tag, and the sidecar is repinned accordingly.

That picks up three format additions: the FILE logical type, INT96 chronological ordering, and the ALP encoding.

Two small Java changes come with it:

- `FILE` now exists in the `LogicalType` union. Before the bump a file using it fell into the unrecognised-logical-type branch and was ignored, preserving the physical type; afterwards it was a recognised member with no case and would have thrown. It is now ignored explicitly, so reader behaviour is unchanged. Covered by a test.

- `ALP` is in the format enum but has no `org.apache.parquet.column.Encoding` counterpart yet, so `testEnumEquivalence` has nothing to round trip it through. It skips ALP for now. The skip is temporary and is removed by #3397, which adds the Java side.

No behaviour changes otherwise. Full parquet-hadoop suite passes (746 tests), clean build, RAT and spotless clean.

@RussellSpitzer this is the precursor you asked for on #3397, so that the ALP entry comes from upstream rather than being a local addition. It turned out it cannot be thrift only: bumping on a clean master fails `testEnumEquivalence` with `No enum constant org.apache.parquet.column.Encoding.ALP`, because the generated enum gains ALP(10) while the hand written one has none, and adding that constant pulls in the ALP readers. Hence the temporary skip here, which #3397 removes along with the local block.


wgtmac (approved):

RussellSpitzer (approved):

divjotarora (approved):

Fokko (approved):
Thanks @vinooganesh for bumping this 👍

Fokko:
Thanks @wgtmac @RussellSpitzer and @divjotarora for the quick review 🚀

PR #3793, opened 2026-09-14
GH-3792: Bind parquet fields to Avro names before aliases

### LINKED ISSUE #3792: AvroIndexedRecordConverter resolves aliases before the exact field name

The alias loop returns before the exact-name match is considered. If field b declares alias a, reading column a returns field b and thus values are mapped into the wrong field.
`AvroRecordConverter` does it correctly

### Component(s)

Avro

nastra:
### Rationale for this change
This makes sure that the actual Avro field name is considered before looking at aliases. 

Fixes https://github.com/apache/parquet-java/issues/3792

### Are these changes tested?
yes

no


parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -126,16 +126,18 @@ private static <T> Class<T> getDatumClass(GenericData model, Schema schema) {
 
   private Schema.Field getAvroField(String parquetFieldName) {
     Schema.Field avroField = avroSchema.getField(parquetFieldName);
+    if (avroField != null) {
```
nastra, line 129: this also aligns with how `AvroRecordConverter` does it. I also checked other places in the codebase and that seems to be the only one affected

parquet-avro/src/test/java/org/apache/parquet/avro/TestAvroFieldNameAndAlias.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.hadoop.ParquetReader;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class TestAvroFieldNameAndAlias {
```
nastra, line 37: there wasn't a good fit for this in existing tests, so I decided to create a separate class

nastra:
@Fokko @wgtmac please take a look when you get a chance

wgtmac (approved):

Issue #3354, opened 2025-11-07
The error message "is not a Parquet file" doesn't show the Parquet path anymore


felipepessoto:
### Describe the bug, including details regarding any error messages, version, and platform.

Version 1.13.1, used by Spark 3.5: https://github.com/apache/parquet-java/blob/db4183109d5b734ec5930d870cdae161e408ddba/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L548

used to show the parquet file path: `Cause: java.lang.RuntimeException: file:/tmp/spark-37..../00000000000000000000.checkpoint.parquet is not a Parquet file (length is too low: 10)`

In version 1.15.2, used by Spark 4: https://github.com/apache/parquet-java/blob/859eac165b08f927fa14590c33bc5f476405fb68/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileReader.java#L584

The error message says: `Cause: java.lang.RuntimeException: io.delta.kernel.defaults.internal.parquet.ParquetIOUtils$1@63d4ffeb is not a Parquet file (length is too low: 10)`

dheerajdwivedi:
The bug in introduced by changes in https://github.com/delta-io/delta.
Full details are here: https://github.com/delta-io/delta/pull/6611

PR #3804, opened 2026-09-20
MINOR: Include local file paths in reader diagnostics

1fanwang:
### Rationale for this change

When a local Parquet file is invalid, the error shows a Java object ID instead of the filename. Callers cannot identify the failing input.

Related to https://github.com/apache/parquet-java/issues/3354. This fixes the built-in local input implementation, not the third-party input in that report.

Return the stored path from the local input's string representation, as Hadoop-backed inputs already do.

#### Testing Done

On Java 17, the same program opens a real ten-byte file through the Parquet reader.

Before, at https://github.com/apache/parquet-java/commit/2df8d02678dab4bb8b926a0d3221cc652984c7ab:

```text
org.apache.parquet.io.LocalInputFile@6bffbc6d is not a Parquet file (length is too low: 10)
```

After:

```text
invalid.parquet is not a Parquet file (length is too low: 10)
```

Run from either checkout:

```bash
mvn -q -pl parquet-hadoop -am -DskipTests package dependency:build-classpath -Dmdep.outputFile=target/runtime-classpath
java -cp "parquet-common/target/classes:parquet-hadoop/target/classes:$(cat parquet-hadoop/target/runtime-classpath)" target/InvalidLocalFileProbe.java invalid.parquet
```

<details>
<summary>Reproducer source</summary>

Save this as target/InvalidLocalFileProbe.java after the build:

```java
import java.nio.file.Files;
import java.nio.file.Path;
import org.apache.parquet.hadoop.ParquetFileReader;
import org.apache.parquet.io.LocalInputFile;

class InvalidLocalFileProbe {
  public static void main(String[] args) throws Exception {
    Path path = Files.write(Path.of(args[0]), new byte[10]);
    try (ParquetFileReader ignored = ParquetFileReader.open(new LocalInputFile(path))) {
      throw new AssertionError("Invalid file was accepted");
    } catch (RuntimeException exception) {
      System.out.println(exception.getMessage());
    }
  }
}
```

</details>

Local-file errors and reader file identifiers now show the supplied path. File contents and read behavior are unchanged.


divjotarora (approved):

wgtmac (approved):

wgtmac:
Thanks @1fanwang for fixing this and @divjotarora for the review!

PR #3811, opened 2026-09-20
GH-3751: Fix recursive protobuf collection handling

makarandhinge:
#3751 
### Rationale for this change

Recursive protobuf messages can contain collection fields whose element types recursively reference protobuf messages. The current handling does not correctly support these recursive collection structures.

This change fixes the recursive protobuf collection handling so that these schemas can be processed correctly.

- Fix recursive handling of protobuf collection types.
- Ensure recursively nested protobuf collection structures are handled correctly.
- Add/update test coverage for the affected recursive protobuf collection case.

### Are these changes tested?

Yes.

The affected protobuf collection handling is covered by tests, and the relevant test suite passes.

No user-facing CLI/API changes are intended. This is a correctness fix for protobuf schema/collection handling.


puskarpeter:
Hi @makarandhinge, thanks for looking at this. Just a heads-up that this issue already has a fix in progress: #3752, linked from #3751 since August. It is stacked on #3750 (now approved), which introduces the
  terminate-as-proto-bytes read path that this PR reimplements, so the two will conflict once #3750 merges.

makarandhinge:
Thanks for the heads-up @puskarpeter! I wasn't aware that #3752 was already in progress. I had independently worked on the recursive protobuf collection handling and the related tests.

Given that #3752 is already stacked on #3750, what would you recommend I do with my PR #3811? Should I close it and contribute any missing changes/tests to #3752, or would you prefer that I rework #3811 based on #3752?

